Issue #44, opened 2020-06-08
Expose x_field, y_field arguments in proximity tools


brendancol:
Right now proximity requires `x` `y` named fields in the xarray.DataArray.  Lets expose `x_field` and `y_field` optional arguments as overrides.

PR #45, opened 2020-06-10
Proximity hot fix_issue44

kristinepetrosyan:
#44


brendancol:
@kristinepetrosyan Nice! small change
- [ ] change `xcord`, `ycord` to just `x='x'` and `y='y'` in the proximity func sig (and update docstring / tests accordingly)

PR #47, opened 2020-06-14
slope: validate output

thuydotm:


brendancol:
@thuydotm make sure there are no big files committed here.   Instead of needing an external `tif`, create a small fixture (30  x 20) or something not square to run tests on.

brendancol:
- you can use open_rasterio...to read the images in
- take a 5x4 chunk out of DEM and copy into `test_slope`
- take a 5x4 chunk out of calculated (`planar_slope....tif`) version and copy into `test_slope`
- remove large files from PR
-

brendancol:
Make tests look something like: https://github.com/makepath/xarray-spatial/blob/master/xrspatial/tests/test_zonal.py#L395-L404

brendancol:
@thuydotm looks like there is close to all that is left here

thuydotm:
@brendancol conflict resolved

brendancol:
- [ ] drop the notebook from PR
- [ ] drop txt file

PR #85, opened 2020-07-22
List all the dependencies

ocefpaf:
I used `depfinder` to list all the dependencies. Maybe most of them are optional and we may need to adjust them. I know that at least `scikit-learn` is needed at import time.


brendancol:
@ocefpaf Yes...we are about to drop that dependency soon. But this is great thank you! I can push another release

ocefpaf:
> @ocefpaf Yes...we are about to drop that dependency soon. But this is great thank you! I can push another release

Thanks! That would be nice so we can get the conda package going. We can drop the unused deps kater.

brendancol:
@ocefpaf hey I just cut a `0.0.8` release which includes this PR.  What else can I do to help get the package out?

ocefpaf:
Let's watch https://github.com/conda-forge/staged-recipes/pull/12177 and see if that passes. It worked for me locally so I think 0.0.8 is a winner :stuck_out_tongue:

PR #117, opened 2020-08-05
Add A* pathfinding

thuydotm:


brendancol:
@thuydotm super exciting!

- [ ] add example to `user_guide.ipynb`
- [ ] benchmark and numba-ize

brendancol:
@thuydotm Let's also prepare a separate example using the City of Austin Road Network:

### Download Data
https://data.austintexas.gov/Locations-and-Maps/Street-Centerline/m5w3-uea6

### Create Lines:
```python
streets = geopandas.read_file('zip://./data/Street Centerline.zip')
streets = streets.to_crs({'init': 'epsg:4326'})

xs = []
ys = []
for s in streets.geometry.values:
    try:
        coords = s.coords.xy
        xs += coords[0].tolist()
        ys += coords[1].tolist()
        
        xs.append(np.nan)
        ys.append(np.nan)
    except:
        continue
        
street_df = pd.DataFrame(dict(x=xs, y=ys))
street_df
```

### Rasterize using datashader.Canvas.lines

### Pick two locations and run A* between them using the roads as the only valid values.   # Note that the start or destination may need to be snapped to the grid so they are not fully surrounded by barrier pixels

thuydotm:
@brendancol ok, I'm working on it

brendancol:
@thuydotm also, let's change the output to be an equal sized Xarray with NaNs for non-path pixels, and the value of the path pixels being the current cost up to that point...

jbednar:
It looks to me like this is doing pathfinding on a pixel basis, reporting a binary mask of pixels that are on the path? If so, would there be a nice way to extract a vector path out of this or provide that as a separate output, which would encode the directionality explicitly?  At least in a central-city road grid network, simplifying adjacent pixels that are on a line (according to some fuzzyness measure) might also be useful.

brendancol:
@jbednar yes totally.  I think it would be nice to have a generic "vectorize (as it raster -> vector geometry)" which could handle points, lines, and polygons. Then the result of a_star_search could be a xr.DataArray shoulding cumulative travel cost...but which could easily be turn to a vector path using the "vectorize" (something with a ambiguous name than that)
.

jbednar:
Sounds good, as long as the A* search is guaranteed not to create a crossing path (which I _think_ it is?). I think a crossing path would be ambiguous about directionality at the pixel level, so that it would be unclear how to convert it into a vector path.

PR #118, opened 2020-08-07
A*: an example of Austin road network

thuydotm:


thuydotm:
- [] prepare a separate example using the City of Austin Road Network
- [] add snap option
- [] benchmark and numba-ize
- [] add Pathfinding Supported Spatial Functions in README.md
- [] add example to user_guide.ipynb

thuydotm:
@brendancol snap option has been added, also the return of a_star_search is now a xarray DataArray. I'm working on numba-ization. I think it would require changes in data structure.

PR #122, opened 2020-08-18
Allow all numpy float data types, not just numpy.float64

jthetzel:
Xarray-spatial checks if data array data types are a subclass
of numpy.float. However, numpy.float is an alias of
builtins.float, of which only numpy.float64 is subclassed.
Changing numpy.float to numpy.floating allows all numpy float
data types.


brendancol:
@jthetzel Thanks for the PR.  We are reviewing and will get it merged ASAP.  So great to have another contributor!

Issue #125, opened 2020-08-31
Add annulus kernel to focal statistics


chase-dwelle:
In addition to the circular kernel, I would like to have the ability to use an [annulus](https://en.wikipedia.org/wiki/Annulus_(mathematics)) as a mask for calculating spatial statistics.

I am dealing with high resolution datasets and radii in the [km] range. Don't think there would be issues with converting from `circle->ellipse` at larger radii for an annulus, but correct me if I'm wrong.

Happy to contribute this if the authors feel it's useful.

brendancol:
@chase-dwelle hey thanks for the suggestion

Would love an annulus kernel.  We also just generally need more eyes on the focal module and would love more contributors.

chase-dwelle:
Thanks, opened up a PR to add the kernel. Probably worth considering abstracting out kernel generation into its own class or set of functions, though I am not sure what other kernels are popular with geospatial analysis.

Issue #124, opened 2020-08-31
Implications of using absolute cellsize after conversion from latlng to meters?


chase-dwelle:
I'm using focal statistics with GeoTIFFs in EPSG:4326, I used the `lnglat_to_meters` to convert the coordinates to meters. However, when doing this [I get a negative `cellsize_y`](https://github.com/makepath/xarray-spatial/blob/f6cb493678a8a5bc78dc6fac3aebe63bb224d644/xrspatial/focal.py#L90), which causes downstream issues.

Is there any issue with using `return cellsize_x, np.abs(cellsize_y)` to address this, or will their be ramifications I'm not accounting for?

chase-dwelle:
This change was merged, code change is here: https://github.com/chase-dwelle/xarray-spatial/blob/644a37be6a241939ccdd7ac0962e2a0c0deb0124/xrspatial/focal.py#L97

PR #126, opened 2020-08-31
Add annulus focal kernel

chase-dwelle:
Addresses #125, #124

Using the circle kernel to extend to an annulus kernel. Implementation excludes the cells on the bound of the inner radius of the annulus.

Also implements the ability to return the raw z-scores from `hotspots`.

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -332,12 +332,12 @@ def _hotspots(z_array):
     return out
 
 
-def hotspots(raster, x='x', y='y', kernel_shape='circle', kernel_radius=10000, zscore=False, r2=None):
+def hotspots(raster, kernel, x='x', y='y'):
     """Identify statistically significant hot spots and cold spots in an input
     raster. To be a statistically significant hot spot, a feature will have a
     high value and be surrounded by other features with high values as well.
     Neighborhood of a feature defined by the input kernel, which currently
-    support a shape of circle and a radius in meters.
+    support a shape of circle, annulus, or custom kernel.
 
     The result should be a raster with the following 7 values:
     90 for 90% confidence high value cluster
@@ -353,7 +353,6 @@ def hotspots(raster, x='x', y='y', kernel_shape='circle', kernel_radius=10000, z
     raster: xarray.DataArray
         Input raster image with shape=(height, width)
     kernel: Kernel
-    zscore: If True, return z-scores instead of hotspots
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


chase-dwelle:
Thinking the kernel should be "pure" and not include the focal point. Will push a change.

thuydotm:
Thanks for the PR, I would love to have some tests with this annulus kernel. With the `zscore` option added to `hotspots`, I think we should have a separate function for it. 

The ongoing stuff with focal is to allow some other kernel shapes (eg: cone, wedge) and also custom kernel. We'd better supply kernel value to a function instead of creating them inside the function, I believe. So basically, the interface should look be something like this:

```python
def _validate_kernel_shape(values, shape, radius, outer_radius, inner_radius, ...):
    if (shape == 'circle'  and radius is valid):
        if values is not None or outer_radius is not None or inner_radius is not None:
             raise ValueError("""circular kernel must be specified by shape='circle' and a valid `radius`, recieved: ...""")
    # check other kernel shapes
    ....

def create_kernel(cellsize_x, cellsize_y, values, shape, radius, outer_radius, inner_radius):
      _validate_kernel_shape()
      if shape == 'circle':
           kernel = _gen_circular_kernel(cellsize_x, cellsize_y, radius)
      elif shape == 'annulus':
           kernel = _gen_annulus_kernel(cellsize_x, cellsize_y, outer_radius, inner_radius)
      ....
      return kernel

def apply(raster, kernel, func):
      ....
```

chase-dwelle:
Thanks, will make these changes. Would it make sense to extend the kernel by passing `**kwargs` considering the expansion of kernel shapes? Could also pass it as a dictionary to make definition and declaration of kernels more explicit.

thuydotm:
@chase-dwelle with the expansion of kernel shapes, I'm thinking that we can provide a bunch of functions that explicitly tell the shape in their names instead of having 1 function with a long list of args.

```python
def circular_kernel(cellsize_x, cellsize_y, radius):
      # first validate args
      # create and return 2d array of kernel

def annulus_kernel(cellsize_x, cellsize_y, outer_radius, inner_radius):
      # first validate args
      # create and return 2d array of kernel 

def custom_kernel(kernel_values):
     ...
```

How do you think?

thuydotm:
Anyway, these are some next steps. We're happy to have this PR merged once having tests updated.

chase-dwelle:
Added some tests for the kernels and made kernels a required argument to be passed to functions. The data flows go: `raster -> cellsize -> kernel -> function/convolution`, where `cellsize` and `kernel` are returned from functions and are used as arguments instead of passing kernel constructor information to each of the function/convolution methods.

This adds a couple more steps, but also is more straightforward from a code readability standpoint (in my opinion). 

I think it might be reasonable to refactor kernel construction to a `xrspatial/kernels.py` in a future PR as kernel information is going to only grow more. Would that be reasonable @thuydotm ?

thuydotm:
@chase-dwelle hey the PR looks good and we'll merge it now. 

The data flow sounds good. Next step is to remove the function `create_kernel`. A kernel of any shape can be created directly by calling its corresponding function. It makes sense to assume that user knows which shape they want and which params related.

We should still have kernel functions in focal in my opinion since it's only used for focal stats currently.

PR #127, opened 2020-09-04
Moved kernel creation to name-specific functions.

chase-dwelle:
Per @thuydotm , simplified some kernel creation.

- Removed `create_kernel()`, now kernels are created by calling their name-specific functions. These are only `circle_kernel()` and `annulus_kernel()` at the moment. Custom kernels are any numpy array and therefore don't have a function, e.g., `kernel = np.ones((3, 3))`.
- Removed `_validate_kernel_shape()` and replaced with `validate_kernel()`, which checks that the kernel is a numpy array and has odd dimensions. This isn't a strict requirement, but I think it makes sense to have the focal point be at the center of a kernel. Perhaps in the future we can accept custom kernels where the focal point is not centered and then pad the kernel in order to center it.
- Updated tests to reflect these changes.

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -110,7 +110,7 @@ def _gen_ellipse_kernel(half_w, half_h):
     return ellipse.astype(float)
 
 
-def circular_kernel(cellsize_x, cellsize_y, radius):
+def circle_kernel(cellsize_x, cellsize_y, radius):
     # validate radius, convert radius to meters
     r = _get_distance(str(radius))
 
@@ -119,6 +119,7 @@ def circular_kernel(cellsize_x, cellsize_y, radius):
 
     kernel = _gen_ellipse_kernel(kernel_half_w, kernel_half_h)
     return kernel
+
 
 def annulus_kernel(cellsize_x, cellsize_y, outer_radius, inner_radius):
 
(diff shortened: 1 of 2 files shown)


brendancol:
Hell yeah!  @thuydotm let me know what you think

thuydotm:
> * Removed `create_kernel()`, now kernels are created by calling their name-specific functions. These are only `circle_kernel()` and `annulus_kernel()` at the moment. 

`circle_kernel()` and `annulus_kernel()` look good to me.

> * Custom kernels are any numpy array and therefore don't have a function, e.g., `kernel = np.ones((3, 3))`.
> * Removed `_validate_kernel_shape()` and replaced with `validate_kernel()`, which checks that the kernel is a numpy array and has odd dimensions. 

@chase-dwelle We should have a function name `custom_kernel` which accepts a numpy array as an argument (we may consider other array-like objects as long as we can validate them). The logic in `validate_kernel()` should be inside `custom_kernel` also. For now, the function either raises an error or returns the array itself.

chase-dwelle:
> @chase-dwelle We should have a function name `custom_kernel` which accepts a numpy array as an argument (we may consider other array-like objects as long as we can validate them). The logic in `validate_kernel()` should be inside `custom_kernel` also. For now, the function either raises an error or returns the array itself.

Changed. Agree on accepting any array-like object, but need to figure out if it would break any downstream dependencies.

thuydotm:
@chase-dwelle the return of custom_kernel needs to be a numpy array for sure.

@brendancol this PR looks good and ready to be merged.

Issue #130, opened 2020-09-15
Allow convolution using CUDA


chase-dwelle:
For focal statistics, would be useful to allow convolution via CUDA for very large data arrays or with large convolution kernels.

- [X] 2D convolution to start with. Is there an application where 3D would be useful (across time/z-axis)?
- [X] Automatically detect if CUDA-enabled GPU is present, if not use Numba's parallel computation
- [X] Do convolution based on passed kernels
- [ ] Make kernel creation a focus in `focal.py`, with the default being the mean kernel.
- [X] Automatically reflect across boundaries, this is not an issue for the mean kernel but could be an issue for other custom kernels. Will need to be addressed in the future.

chase-dwelle:
@brendancol I've written the above checked issues. A couple things I'd want to iron out before opening a PR:

- I think it makes sense to have a separate `convolution.py` as I think it's cleaner to have it separated, though you may want to contain it within `focal.py`.
- Are you okay with replacing `focal._apply` with convolution functions?

brendancol:
@chase-dwelle this sounds great!  I like the approach and would be happy to help test.  I have 1 CUDA GPU which could help.  I'm fine replacing `focal._apply`.  @thuydotm let's us know what you think.

brendancol:
@chase-dwelle I would love to hear your thoughts on a general spelling for dealing with edge effects.

`xwrap`: wrap around xdim
`ywarp`: wrap around ydim
`warp`: wrap both x, y
`pad`: return np.nan on edges of incompletely represented kernels
`valid`: change shape of output to only include valid kernel areas
`doublecount`: double count edge pixels to have complete neighborhood

Just brainstorming...

thuydotm:
Sounds good to me. I'm ok with having kernel/convolution separated from focal also.

PR #131, opened 2020-09-16
Added convolution module for use in focal statistics.

chase-dwelle:
Addresses #130 

A few things need to be addressed, as mentioned in #130 

- How to handle boundaries. This in an issue on non-symmetrical mean kernels, as their weighted averages can be affected. For my use cases, what is coded works for mean kernels: the values of the raster will be reflected across the boundary so the appropriate weightings are applied (given a symmetrical kernel). Sum kernels don't require padding so `pad=False` works just fine.
- After coding this, there could be some validity in making an easier kernel-creation function to include common functions like weighted means, but it's also very simple to just do `kernel / kernel.sum()` to get this.
- I am a little confused how the original `focal._apply()` behavior was supposed to work from the old `test_apply()`, in one test `np.nan == 0` (the sums) and in the other it was supposed to be `==1` (the means).
- There could be some way to deal with `np.nan` in the raster for the kernel creation, but I honestly always deal with them at the source so I'm not sure if there is a clean way to handle them, or if we _should_ allow that behavior. I personally like the amplification of `NaN`s to alert me to underlying data issues.

### DIFF
--- a/xrspatial/convolution.py
+++ b/xrspatial/convolution.py
@@ -1,0 +1,117 @@
+import datashader.transfer_functions._cuda_utils as cu
+from numba import cuda, float32, int32, prange, jit
+from xrspatial.utils import ngjit
+import numpy as np
+
+def _cuda_present():
+    """Check for supported CUDA device. If none found, return False"""
+    try:
+        cuda.cudadrv.devices.gpus.current
+        local_cuda = True
+    except cuda.cudadrv.error.CudaSupportError:
+        local_cuda = False
+
+    return local_cuda
+
+
+def convolve_2d(image, kernel, pad=True, use_cuda=True):
+    """Function to call the 2D convolution via Numba.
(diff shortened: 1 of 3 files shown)


brendancol:
@chase-dwelle hey this looks awesome.  Any chance you could add a section to `examples/user_guide/focal.ipynb` with some simple examples of using the new  module?

chase-dwelle:
> @chase-dwelle sorry for delay on getting this merged. this looks awesome. Any chance you could add a section to `examples/user_guide/focal.ipynb` with some simple examples of using the new module?

Happy to, but won't get to it until next week. I can include the example as part of this PR or open a new PR.

chase-dwelle:
@brendancol I added an example notebook for the focal and convolution kernels. Very simplistic but shows basic usage. The numbering of the notebooks seemed somewhat arbitrary so I labeled it as `4`. Let me know if it should be numbered differently.

PR #137, opened 2020-10-21
Reclassify cross tabulate notebook

brendancol:
- [ ] Add descriptions for each cell


brendancol:
@lex-c this is an older notebook which shows a common analysis pattern using Xarray-Spatial.  Let's make it into another example and happy to chat about it.

brendancol:
@lex-c this is an older notebook which shows a common analysis pattern using Xarray-Spatial.  Let's make it into another example and happy to chat about it.

giancastro:
@brendancol, will this PR still be merged?

PR #174, opened 2021-01-12
docs enhancements

SapirLastimoza-Dooley:


xrspatial/aspect.py
```diff
@@ -154,21 +154,52 @@ def _run_dask_numpy(data:da.Array) -> da.Array:
...
 def aspect(agg: xr.DataArray, name:str ='aspect'):
-    """Returns downward slope direction in compass degrees (0 - 360) with 0 at 12 o'clock.
+    """Calculates, for all cells in the array, the downward slope direction of each cell 
+        based on the elevation of its neighbors in a 3x3 grid. The value is measured 
+        clockwise in degrees with 0 and 360 at due north. Flat areas are given a value of -1.
 
     Parameters
     ----------
     agg : DataArray
+        Numpy, CuPy, Dask, or CuPy GPU data array with agg.data = elevation
```
thuydotm, line 164: The explanation for param `agg` and the return of the function should be more clear. Ex: agg: xarray.DataArray. The data dimension of input aggregate is 2D and can be either a numpy array, a cupy array, a dask array with numpy backed or a dask array with cupy backed...

xrspatial/aspect.py
```diff
@@ -154,21 +154,52 @@ def _run_dask_numpy(data:da.Array) -> da.Array:
...
 
     Returns
-    -------
+    ----------
     data: DataArray
+        Numpy, CuPy, Dask, or CuPy GPU data array with agg.out = aspect
```
thuydotm, line 169: You may want to add some more details here. The output is a DataArray where its data dimension is the calculated aspect. Output data has the same type as of input. If we pass a DataArray with numpy-backed xarray, the output is also numpy-backed. Similarly with other input types.  Other attributes of the output aggregate are preserved as of input. Note that we don't have `.out` attribute in a DataArray.

xrspatial/aspect.py
```diff
@@ -154,21 +154,52 @@ def _run_dask_numpy(data:da.Array) -> da.Array:
...
+    Imports
+    >>> import numpy as np
+    >>> import xarray as xr
+    >>> import xrspatial
+
+    Single Cell
+    >>> elevation = np.array([[101, 92, 85], [101, 92, 85], [101, 91, 84]])
+    >>> elev_array = xr.DataArray(elevation)
+    >>> elev_array
+    <xarray.DataArray (dim_0: 3, dim_1: 3)>
```
thuydotm, line 188: Would be great if we specify the dimensions here then we can show that the coords of input and output are the same. Ex:
```
    agg = xr.DataArray(np.array([[0, 1, 0, 0],
                                 [1, 1, 0, 0],
                                 [0, 1, 2, 2],
                                 [1, 0, 2, 0],
                                 [0, 2, 2, 2]]),
                       dims=['lat', 'lon'])

    height, width = agg.shape
    _lon = np.linspace(0, width - 1, width)
    _lat = np.linspace(0, height - 1, height)
    agg['lon'] = _lon
    agg['lat'] = _lat
```

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
thuydotm, line 1: Can you remove the .vscode folder from this PR? 
You can still keep it in your local repo `echo ".vscode/" >> .gitignore`

thuydotm (changes requested):

brendancol:
@SapirLastimoza-Dooley this is looking great!  Let me know when you think its good to merge and we'll merge it.

thuydotm:
@SapirLastimoza-Dooley Everything looks good to me except for a typo with pd.DataFrame. Can you please fix this and we'll be good to merge.

PR #263, opened 2021-03-29
Edit proximity

lex-c:
Quick edit to get absolute value rather than sum of squares for the manhattan (taxicab) distance


brendancol:
@lex-c recreate PR with only changes to the proximity notebook. It should have only 1 commit and not touch any core xarray-spatial functions (multispectral touched here)

PR #294, opened 2021-04-02
add spatial_ref tests to appropriate module tests

lex-c:
- add private module and method to add hard-coded EPSG4326 to da's:
    - written and added in format compatible with rioxarray and read as crs
    - add while preserving attrs and coords needed for other tests
 - add tests for base case of every function accepting and returning da's:
    - test attrs and coords preserved: (no tests for crs validity, would need to use at least pyproj to test that)
    (- convolution convolve_2d left out because no da return.)
 - run pytest to verify all pass (with some skipped because no cuda on my machine)


brendancol (commented):
@lex-c 
- [x] let's change _add_EPSG4326 -> add_crs_metadata(epsg=4326) and add to xrspatial/utils.py
- [x] make sure to be using 4 SPACES for indentation
- [x] run flake8 and find any style errors
- [ ] remove any VSCode related files

lex-c:
I fixed all the points above and for fun I also flaked all the test modules and edited everything I was absolutely sure would not be code breaking 😊

giancastro:
@lex-c, could you take a look at the branch's conflicts?

PR #306, opened 2021-04-06
add vscode and pycharm in gitignore

lex-c:


brendancol:
@lex-c thanks...looks like this already made it in in another PR.  Closing for now but we can reopen if I'm mistaken

lex-c:
Yes, I've been trying to add it in every PR for now that still has the
issue because I understand it's really bad to have it show up 😊

PR #308, opened 2021-04-06
Fix examples cli

lex-c:
Tested with conda build and upload package to anaconda 
```conda install -c chlochlo xarray-spatial```
xrspatial examples works


lex-c:
I added back in exclude xrspatial/examples in the manifest.in to exclude dir generated in build process (perhaps) to make sdist and vsc match.
Used check-manifest package to check
Works now and passed check-manifest test

brendancol:
@lex-c
- [ ] remove `venv.bak/`

Also, you should use `conda` instead of `virtualenv` in the future...

lex-c:
@brendancol I'm working on this now but I can't find any venv.bak. What directory was that in?

PR #309, opened 2021-04-07
fix manhattan distance to sum of abs

lex-c:


thuydotm:
@lex-c let's change this to sum of absolute instead. 
```python
mahattan_distance(x1, y1, x2, y2) = abs(x1 - x2) + abs(y1 - y2)
```

thuydotm:
This looks good to me, let's get this merged!

PR #324, opened 2021-04-09
Save cogs clear nb

lex-c:


lex-c:
@brendancol it says "package pytest is not installed; ask your administrator to install..." a set of packages. Is this on my end? I don't think I've had this kind of error before and I have pytest in my local conda

lex-c:
I added save from memory and some overviews, but it's set correctly yet.

giancastro:
@lex-c, what issue are you solving in this PR? Could you link it?

PR #335, opened 2021-04-12
Add git pre-commit to clear output and metadata from nbs

lex-c:
@brendan I added the pre-commit git hook and it removes output and metadata. (Pretested in a sample git repo, and then retested with xrs)


brendancol:
@lex-c I don't see any files changed here...probably because its a githook, but are there any files changed here?

lex-c:
No , just the base .git; I guess it's not necessary to merge

PR #339, opened 2021-04-13
true_color: sigmoid contrast enhancement

thuydotm:


thuydotm:
Adding back the sigmoid contrast enhancement when doing true color. Function drop_clouds is removed and will be re-added with a new PR.

thuydotm:
@brendancol I think this is ready to merge. Here are output samples (with and w/o clouds) after enhance contrast with sigmoid function where c=10, th=0.125

image
image

brendancol:
@thuydotm nice! looking great!

PR #358, opened 2021-04-15
add get_data module and example sentinel-2 data

SapirLastimoza-Dooley:
added `xrspatial.datasets.get_data()` so that users can load example sentinel-2 data to use in multispectral docstring examples


brendancol:
@SapirLastimoza-Dooley This looks great!

brendancol (commented):
nice style!

PR #364, opened 2021-04-16
true_color returns a DataArray

thuydotm:


thuydotm:
True color now returns a DataArray (dask or numpy backed depending on the input array type) instead of a PIL image. Some fixes also included to improve the performace. 

@brendancol this is ready to merge.

PR #370, opened 2021-04-18
Only after PR#349: Test NBs onJupyterLab

lex-c:
Edited sentinel NB for user ease and compatibility in JupyterLabs; still 2 caveats in sentinel NB.
I also edited the Pathfinding NB and removed the unnecessary line parsing and replaced it with an explode to expand multiline rows; works with spd now.


lex-c:
Ran all NBs on MS JLabs; there are still some issues, especially with the ones outside user_guide.

PR #381, opened 2021-04-21
add deprecation warning to tiles module

lex-c:
Tested that it only shows up on import of the tiles module and not base package. @brendancol did you want the warning each time someone calls any of the functions?


lex-c:
@brendancol I put the deprecation warning globally for importing the module. Would you rather have decorators on each of the functions so each of them calls the warning?

brendancol:
@lex-c sweet looking good.  Through a `Ready to Merge` label on there when you feel its good to go

PR #434, opened 2021-05-17
Optl reproj

lex-c:
Added reprojection with one blank function to test optional modules install. Can be extended to other modules as well.
Install with
```pip install -e . --install-option='--reprojection'```
to get reprojection module.


lex-c:
Added reprojection using rio including support for dask using built-in map_blocks for parallel:
This involves calculating the shapes for each chunk reprojected and adjusting the chunk shapes.
It is still slower than the non-parallel rio.reproject because it uses as part of the iteration
We'll have to go a bit lower-level to rasterio to get better speeds I think. I'm working on getting that in.

lex-c:
Added the reproject by xarray.DataArray.map_blocks; setting coords on the template is still required for multi chunk dask array data. Tests are set up for numpy and single chunk dask and -- commented out for now -- multi chunk dask.

lex-c:
@brendancol this is ready to merge. I updated the setup for extra modules to be much more extendable to other modules we put in the future and much easier to use. 
```pip install '.[gdal]'```
from the xarray-spatial dir works and installs the extra deps
I also downgraded the dask version because of a spatialpandas issue with Dask imports.

brendancol:
@lex-c thats great. Let's get the CI passing and review together

lex-c:
@brendancol checks have passed. Still looking into the examples download issue.

PR #439, opened 2021-05-24
add make terrain sample data function

giancastro:
This PR solves the issue #437.


giancastro:
@brendancol

brendancol (commented):
- [x] make `noise` optional dependency
- [x] try: except ImportError: `must install noise to use terrain pip install noise`
- [ ] remove noise from setup.py

PR #477, opened 2021-06-18
add classify notebook

giancastro:


giancastro:
@marcozimmermannpm @brendancol, could you take a look at the notebook and see if we are good to move forward?

PR #484, opened 2021-06-22
true_color: added output name and attrs

thuydotm:
- [x] Add name for output data array
- [x] Keep attrs
- [x] Add docs with example


thuydotm:
Let's get this PR merged and release a new version so that we can have https://github.com/microsoft/PlanetaryComputerExamples/pull/24 run smoothly

giancastro (changes requested):

giancastro (approved):

Issue #488, opened 2021-06-24
Naming scheme for notebooks


tcmetzger:
While working on the example notebooks, I got the impression that there is no consistent scheme for naming the .ipynb files.

Since the notebooks we are currently working on seem to be focused on highlighting certain functionalities of certain tools, I'd propose a naming scheme like this:

`[tool name]_[functionality].ipynb`

For example, the classification notebook that I'm currently working on with @giancastro highlights different classification methods with xarray-spatial. The filename in this case would then be: `xarray-spatial_classification-methods.ipynb`

Thoughts/suggestions?

giancastro:
The naming scheme looks very nice. It'll make easier to understand what the notebook is about without open it.

Issue #475, opened 2021-06-17
Refine MS Notebook Style / Text


brendancol:
The goal is to do a full review of all MS notebooks and refine grammar, style, and layout.  Could you send a half day on each of these (others coming down the pipeline...).

On https://github.com/microsoft/PlanetaryComputerExamples/tree/main/tutorials:

- [ ] cloudless-mosaic-sentinel2.ipynb
- [ ] local-tools.ipynb
- [ ] ndvi_hotspots.ipynb

marcozimmermannpm:
@tcmetzger this is the repository where the notebooks are: https://github.com/microsoft/PlanetaryComputerExamples

PR #489, opened 2021-06-25
Edits to classification notebook

tcmetzger:
As described in #475, this PR contains various edits to the classification example notebook.

Most notably, I have updated wording and structure to clarify the goal of the notebook and edited all text to conform to the Microsoft Writing Style Guide (https://docs.microsoft.com/en-us/style-guide/welcome/). I have also updated the filename according to the scheme proposed in #488.

cc @giancastro, @brendancol, @marcozimmermannpm


tcmetzger:
@giancastro @brendancol Three ideas/requests:

- For the introductory paragraph, it would be good to briefly mention why xarray-spatial is a good tool to use for classification in this context. So far, I only have a sentence saying that xarray-spatial "contains different classification methods" - it would be nice to be a little more specific about why one would want to use xarray-spatial to classify data.
- I think it would be helpful to include some more detail about the basic pros and cons of each of the four classification methods. I have tried to add some background, but would certainly appreciate your input.
- I would like to see some kind of concluding paragraph after the last example in the notebook. I would suggest either a "Next steps" section that could suggest other data sets/areas of interest to try with the four classification methods - or a summary section that highlights once again the pros and cons of each of the four classification methods and links to some other, more advanced methods (Geometric Interval, Standard Deviation, etc.).

brendancol:
@tcmetzger @giancastro this is looking great!

I a bit torn on the use of matplotlib, we can also use the `Elevation` cmap from `datashader.colors` and avoid the extra dependency.  Let's go ahead and get MS PR going and we can switch that out later if needed

PR #502, opened 2021-07-07
add Dask support for zonal_stats and zonal_crosstab

thuydotm:
- [x] add Dask support for zonal stats
- [x] add Dask support for zonal crosstab 2d
- [x] add Dask support for zonal crosstab 3d
- [x] allow floating point agg for zones
- [x] allow zone with id 0 (current behavior considers zone 0 as nodata region of the image so it is excluded from calculation)
- [x] add param `nodata_zones` with a default of `None` (or `nan`?) to specify value that will be excluded from `zones` raster
- [x] add param `nodata_values` with a default of `None` to specify value that will be excluded from `values` raster
- [x] update tests
- [x] update docstrings

### DIFF
--- a/xrspatial/tests/test_zonal.py
+++ b/xrspatial/tests/test_zonal.py
@@ -29,7 +29,7 @@ def create_zones_values(backend):
     ])
     values_2d = xr.DataArray(values_val_2d)
 
-    values_val_3d = np.ones(4 * 3 * 6).reshape(3, 6, 4)
+    values_val_3d = np.ones(4*3*8).reshape(3, 8, 4)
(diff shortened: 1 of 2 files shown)


brendancol:
@thuydotm hey this is super exciting.  Can you add a mini-checklist to cover remaining tasks here?

thuydotm:
@brendancol absolutely, checklist added. Can you review the last 2 items in the checklist? At the moment, we consider zone 0 as background of the image that will be excluded from calculation. I think we're better explicitly specify `nodata_zones` and `nodata_values` as inputs of the function.

PR #539, opened 2021-08-04
canvas_like

thuydotm:


thuydotm:
This PR changes the name of `resample` function in xrspatial.utils to `canvas_like` and modifies its behavior related to canvas size setting. If no `height` provided, it will be implied from the canvas width and canvas bounds. 

@brendancol There is an example notebook using `resample` function: https://github.com/makepath/xarray-spatial/blob/master/examples/resample.ipynb. As this notebook use data and resources from the Planetary Computer, should we remove it or update it w.r.t these new changes?

brendancol:
@thuydotm is this ready to merge?

Issue #541, opened 2021-08-16
Refine Hillshade GPU accelerated version


a7az0th:
Currently the Hillshade GPU accelerated version runs ~3x faster than numpy, but it can do better.
We can also try to minimize dependencies. It could work without cupy

PR #542, opened 2021-08-16
Added a pure numba hillshade that is 10x faster compared to numpy

a7az0th:
The new code uses less context switching and data transfers, minimizes communication overhead and removes the cupy dependency.  I've made it run by default, if cuda is available. Otherwise if falls back to numpy

This change addresses issue #541

### DIFF
--- a/xrspatial/hillshade.py
+++ b/xrspatial/hillshade.py
@@ -1,6 +1,6 @@
 # std lib
 from functools import partial
-from math import sqrt
+import math
 
 # 3rd-party
 try:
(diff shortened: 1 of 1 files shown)


brendancol:
- [ ] change to I/O to cupy arrays

brendancol:
@thuydotm can you review this and get back with any comments?

thuydotm:
Thanks @a7az0th!
I think we can now remove the old `run_cupy` function and replace it with the `run_numba` since the old code is no longer called. `run_numba` should be renamed to `run_cupy` to be consistent with our coding convention. Otherwise looks good to me.

a7az0th:
@thuydotm I've updated the source according to your feedback. I believe we can merge this.

thuydotm:
Merged, thanks!

a7az0th:
Thank you, too!

PR #552, opened 2021-09-15
Fast terrain

kiliakis:
This PR brings all the terrain generation related development and optimization. In brief:

- adds a `fast_perlin.py` module that contains a `fast_perlin()` method that can be used to generate perlin noise. It offers higher performance compared to the previous implementation of perlin noise and has a Numpy, CuPy and Dask implementation. *Note that the interface is now slightly different. The user needs to pass an xr.DataArray (can be based on numpy, cupy or dask array) instead of width, height.* 
- adds a `fast_terrain.py` module that contains a `generate_fast_terrain()` method that can be used to generate a randomized terrain. It is faster compare to the existing implementation of terrain generation. Uses the `fast_perlin()` method and has a Numpy, CuPy and Dask implementation. *Note that the interface is now slightly different. The user needs to pass an xr.DataArray (can be based on numpy, cupy or dask array) instead of a Canvas.* 
- Adds two jupyter notebooks in the examples directory, one for each of the two newly added methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,9 +144,9 @@ In the GIS world, rasters are used for representing continuous phenomena (e.g. e
 | [Curvature](xrspatial/curvature.py) | ✅️ | | | ⚠️  |
 | [Hillshade](xrspatial/hillshade.py) | ✅️ | ✅️  | | |
 | [Slope](xrspatial/slope.py) | ✅️  | ✅️  | ✅️ | ⚠️  |
-| [Terrain Generation](xrspatial/terrain.py) | ✅️ |  | | |
+| [Terrain Generation](xrspatial/terrain.py) | ✅️ | ✅️ | ✅️ | |
 | [Viewshed](xrspatial/viewshed.py) | ✅️ |  | | |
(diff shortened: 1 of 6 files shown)


kiliakis:
@thuydotm , @brendancol please review it and let me know of your feedback.

thuydotm:
Thanks @kiliakis
I have tested those new functions and they are working well. I'm going to add some tests to `test_terrain.py` and do some refactor so that all the terrain generation methods are in one file of `terrain.py`

kiliakis:
hi @thuydotm , I saw that the flak8 linting test failed. Shall I re-format it with flake8 and push the new commit? Is there any project specific configuration file for liniting?

thuydotm:
@kiliakis we can solve that later, let's get the tests for `generate_terrain` passed first. `perlin` is working well though. Last time, I tested using the fast_terrain benchmark notebook directly, and since the dask chunk size is bigger than the actual data size, we got indentical results for numpy case and dask case. You can change the chunk size to be smaller to see the difference. Please see PR https://github.com/makepath/xarray-spatial/pull/555 where I've wrote some unit tests.

PR #555, opened 2021-09-17
generate_terrain: cupy case, dask numpy case

thuydotm:


thuydotm:
@kiliakis I've refactored the code and added some tests. It happened that the results when running with vanilla numpy, dask numpy, and cupy backed DataArrays are not the same. Similar things happened when running with the original code you added. Can you investigate and make sure the results match? I can help to take a look into the dask case.

kiliakis:
@thuydotm  The perlin noise function (that is also used by generate terrain) uses a random number generator (see here for example: https://github.com/makepath/xarray-spatial/pull/555/files?file-filters%5B%5D=.md&file-filters%5B%5D=.py#diff-4599a7f9c05a7872e774fd7d4fa503515a3e5eecbed9d36ed379bb2266491a62R71). 
The numpy implementation uses `numpy.random` and Cupy uses `cupy.random`.  These functions do no return the same sequence of random numbers. 
So my guess is this is why we don't see the same results. To see the same results, we will have to modify the code to not shuffle the p array (see here: https://github.com/makepath/xarray-spatial/pull/555/files?file-filters%5B%5D=.md&file-filters%5B%5D=.py#diff-4599a7f9c05a7872e774fd7d4fa503515a3e5eecbed9d36ed379bb2266491a62R76). But this is something that can be used for testing only.

PR #562, opened 2021-09-29
Optimize natural_break on large inputs

a7az0th:
Optimized natural break on CPU.
* Removed slow random sample generation with a much faster one.
* Removed a slow and unnecessary nan removal
* Optimized the _run_numpy_jenks_matrices by making it multithreaded.

Initial performance I benchmarked took 11.27s to calculate
New version takes 3.2s


thuydotm:
Thanks! LGTM

PR #572, opened 2021-11-10
zonal stats: speed up dask case

thuydotm:
This PR uses the same approach as https://github.com/makepath/xarray-spatial/pull/568 to improve performance for zonal stats when input data arrays are dask-backed. It computes stats chunk by chunk and then summarizes all the results and return output as a dask DataFrame. 

This also limits stats that supported in dask case to a subset of default stats, which is safer since a custom statistics would not be always element-wise thus can produce unexpected results.

`nodata_zones` is removed as we already support `zone_ids`, and exclude invalid values (nan, inf) from our calculations.

### DIFF
--- a/xrspatial/zonal.py
+++ b/xrspatial/zonal.py
@@ -1,12 +1,14 @@
 from math import sqrt
 from typing import Optional, Callable, Union, Dict, List
 
-import dask.array as da
-import dask.dataframe as dd
 import numpy as np
 import pandas as pd
 import xarray as xr
 from xarray import DataArray
+
+import dask.array as da
+import dask.dataframe as dd
+from dask import delayed
 
 from xrspatial.utils import ngjit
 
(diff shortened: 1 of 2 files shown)


xrspatial/zonal.py
```diff
@@ -106,105 +122,191 @@ def _stats(
...
+    return strides
+
+
+@delayed
+def _stats_func_dask_numpy(
+    zones_block: np.array,
+    values_block: np.array,
+    unique_zones: np.array,
+    zone_ids: np.array,
+    func: callable,
```
ianthomas23, line 152: I think `callable` should have a capital C so that it is the type rather than the function.

xrspatial/zonal.py
```diff
@@ -106,105 +122,191 @@ def _stats(
...
+    nodata_values: Union[int, float] = None,
+) -> pd.DataFrame:
+
+    sorted_zones = np.sort(zones_block.flatten())
+    sored_indices = np.argsort(zones_block.flatten())
```
ianthomas23, line 157: `sored_indices` should be `sorted_indices`?

xrspatial/zonal.py
```diff
@@ -106,105 +122,191 @@ def _stats(
...
+    values_by_zones = values_block.flatten()[sored_indices]
+
+    # exclude nans from calculation
+    # flatten_zones is already sorted, NaN elements (if any) are at the end
+    # of the array, removing them will not affect data before them
+    sorted_zones = sorted_zones[np.isfinite(sorted_zones)]
+    zone_breaks = _strides(sorted_zones, unique_zones)
+
+    start = 0
+    results = np.zeros(unique_zones.shape) * np.nan
```
ianthomas23, line 167: I think that
```python
results = np.full(unique_zones.shape, np.nan)
```
might be better (and faster).

xrspatial/zonal.py
```diff
@@ -106,105 +122,191 @@ def _stats(
...
+    for i in range(len(unique_zones)):
+        end = zone_breaks[i]
+        if unique_zones[i] in zone_ids:
+            zone_values = values_by_zones[start: end]
```
ianthomas23, line 171: Unnecessary space in `[start: end]`.

ianthomas23:
Just a few minor comments, otherwise it looks good to merge.

thuydotm:
Thanks Ian, I just updated the code. I'll merge into master once the tests all passed.

PR #581, opened 2021-11-18
Error in zonal_stas when zone_values is empty

g2giovanni:
# Fixes 
stats_dict is not properly populated when zone_values is empty

## Proposed Changes

  - moved check of empty zone_values inside the for loop

### DIFF
--- a/xrspatial/zonal.py
+++ b/xrspatial/zonal.py
@@ -322,10 +322,10 @@ def _stats_numpy(
             # filter out non-finite and nodata_values
             zone_values = zone_values[
                 np.isfinite(zone_values) & (zone_values != nodata_values)]
-            if len(zone_values) == 0:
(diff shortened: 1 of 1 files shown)


g2giovanni:
@thuydotm I think with the last commits it was introduced a bug when zone_values is empty. I made a simple fix. 
Anyway, zonal_stats with the last modifications are impressive. Thank you very much. I passed from a computation time of 30 minutes to just 2 minutes 👍

thuydotm:
@g2giovanni thanks for pointing it out. Can you please add a small test for this case to `test_zonal`?

In addition, I'm curious about the computation you did with `zonal_stats` as I want to know how much it got improved with different data other than the data I used to test its performance. Did you use dask backed or numpy backed data array? What size/shape was it and how many zones were there? How was the computation time compared to QGIS?

PR #585, opened 2021-11-25
polygonize function on cpu for numpy-backed xarray DataArrays

### LINKED ISSUE #471: Port polygonize to Numba

ianthomas23:
This PR adds a cpu-based `polygonize` function for numpy-backed `DataArray`s.

The algorithm is slightly different to GDAL's in that it is simpler and has fewer memory reallocations, at the cost of requiring more RAM. It is placed in a new `xrspatial/experimental` directory as the API has not been finalised and in particular there are some optional dependencies (e.g, `geopandas`) that we don't want to be required dependencies.

The API follows that of `rasterio.features.shapes` but there are 4 possible return types that are specified using the `return_type` kwarg. Possibilities are:
  * `numpy` (the default) - return polygons as numpy arrays, which is the fastest option.
  * `spatialpandas` - return a `spatialpandas.GeoDataFrame`.
  * `geopandas` - return a `geopandas.GeoDataFrame`. This option is only available if `geopandas` and `shapely` optional dependencies have been installed.
  * `awkward` - return polygons as an `awkward` array (https://awkward-array.readthedocs.io/en/latest/index.html). This option is only available if `awkward` is installed.

Here is some example code to run the same `raster` and `mask` arrays through this algorithm and that of rasterio/GDAL, and display the results side-by-side using Matplotlib.

```python
import matplotlib.path as mpath
import matplotlib.pyplot as plt
import matplotlib.patches as mpatches
import numpy as np
import rasterio.features
from shapely.geometry import shape
import xarray as xr
from xrspatial.experimental import polygonize

shape = (30, 40)
rng = np.random.default_rng(7423)
raster = rng.integers(low=0, high=3, size=shape, dtype=np.int32)

mask = rng.uniform(0, 1, size=shape) < 0.8

# xarray-spatial polygonize.
xrsp_values, xrsp_polygons = polygonize(xr.DataArray(raster), mask=xr.DataArray(mask))

# rasterio/gdal polygonize.
rasterio_values = []
rasterio_polygons = []
for shape, z in rasterio.features.shapes(raster, mask=mask):
    rasterio_values.append(z)
    rasterio_polygons.append(shape['coordinates'])

def polygon_to_mpl_path(poly, is_rasterio):
    npts = sum([len(boundary) for boundary in poly])
    verts = np.empty((npts, 2))
    codes = np.full(npts, 2)  # LINETO
    i = 0
    for j, boundary in enumerate(poly):
        n = len(boundary)
        # rasterio/gdal boundaries do not always have correct orientation.
        if is_rasterio and j == 0:
            verts[i:i+n] = boundary[::-1]
        else:
            verts[i:i+n] = boundary
        codes[i] = 1  # MOVETO
        i += n
    codes[i-1] = 79  # CLOSEPOLY
    return mpath.Path(verts, codes)

# Map of value -> color index to color by value.
color_map = {}
for i, val in enumerate(np.unique(xrsp_values)):
    color_map[val] = i

fig, axes = plt.subplots(ncols=2, figsize=(12, 6), sharex=True, sharey=True)
for j, (ax, values, polygons, title) in enumerate(zip(
    axes, [xrsp_values, rasterio_values], [xrsp_polygons, rasterio_polygons],
    ['xarray-spatial', 'rasterio'])):

    for value, polygon in zip(values, polygons):
        i = color_map[value]
        c = f"C{i}"
        path = polygon_to_mpl_path(polygon, j==1)
        patch = mpatches.PathPatch(path, facecolor=c, edgecolor=c, alpha=0.5)
        ax.add_patch(patch)

    ax.set_title(title)
    ax.autoscale_view(tight=True)

plt.show()
```
Example output:
ex

Closes #471

### DIFF
--- a/xrspatial/experimental/polygonize.py
+++ b/xrspatial/experimental/polygonize.py
@@ -544,9 +544,8 @@ def polygonize(
 
     return_type: str, default="numpy"
         Format of returned data.  Allowed values are "numpy", "spatialpandas",
-        "geopandas" and "awkward".  Both "numpy" and "spatialpandas" are
-        always available whereas "geopandas" and "awkward" require optional
-        dependencies.
+        "geopandas" and "awkward".  Only "numpy" is always available, the
+        others require optional dependencies.
 
     Returns
     -------
--- a/xrspatial/experimental/polygonize.py
+++ b/xrspatial/experimental/polygonize.py
@@ -1,0 +1,591 @@
+# Polygonize algorithm creates vector polygons for connected regions of pixels
+# that share the same pixel value in a raster.  It is a raster to vector
+# converter.
+#
+# Algorithm here uses compass directions for clarity, so +x direction is East
+# and +y direction is North.  2D arrays are flattened to 1D to make maths of
+# moving through the grid easier/faster to calculate.  2D array have shape
+# (ny, nx) so 1D flattened array has length n = nx*ny.  Position within a 1D
+# array is denoted by index ij = i + j*nx, so i = ij % nx and j = ij // nx.
+# Direction of motion through the grid is represented by forward and left
+# which are the values to add to ij to move one pixel in the forward and left
+# direction respectively.  forward is +1 for E, -1 for W, +nx for N and -nx
+# for S.
+#
+# There are two main stages in the algorithm.  Firstly the raster is divided
+# into connected regions that contain adjacent pixels of the same value.  Each
+# region is labelled with a unique integer ID starting at 1.  Regions
+# corresponding to masked out pixels are all given the same region ID of 0.
+#
+# The second stage identifies where polygon exteriors and holes start, and
+# follows each of these boundaries around the raster, keeping to the edge of
+# the region.  Holes are grouped together with their enclosing exterior
+# boundary.
+#
+# The points of exterior boundaries are ordered in an anticlockwise manner,
+# those of hole boundaries in a clockwise manner.  This assumes that both the
+# x and y coordinates are monotonically increasing or decreasing.
+
+from enum import Enum
+import numba as nb
+import numpy as np
+from typing import Dict, List, Optional, Tuple, Union
+import xarray as xr
+
+from ..utils import ngjit
+
+
+_regions_dtype = np.uint32
+_visited_dtype = np.uint8
+
+
+class Turn(Enum):
+    Left = -1
+    Straight = 0
+    Right = 1
+
+
+@ngjit
+def _diff_row(ij0: int, ij1: int, nx: int) -> bool:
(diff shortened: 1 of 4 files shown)


ianthomas23:
I have made `spatialpandas` an optional dependency. It was listed as required in `setup.py` but not in `requirements.txt` which is what CI uses. It is only used in one example as well as `polygonize`. I have created issue #587 to deal with the duplication of requirements.

Now the only `return_type` of `polygonize` that is always available is `numpy`. The other 3 options require optional dependencies.

ianthomas23:
Note that this PR includes a function to calculate the unique regions of a raster, i.e. the area connectivity. There is already a function that does this in `zonal.py`, but the algorithm here also considers an optional mask. I expect that we will combine the two functions into one in the future.

ianthomas23:
Note to self: In the example code above the rendering of rasterio polygons isn't always correct as the winding orientation is not consistent. So if we ever need to plot rasterio output for comparison this needs to be improved.

xrspatial/experimental/polygonize.py
```diff
@@ -0,0 +1,590 @@
...
+    converter.
+
+    Parameters
+    ----------
+    raster: xr.DataArray
+        Input raster.
+
+    mask: xr.DataArray, optional
+        Optional input mask.  Pixels to include should have mask values of 1
+        or True, pixels to exclude should have 0 or True.  This is the
```
thuydotm, line 527: I think it should be "pixels to exclude should have 0 or False".

xrspatial/experimental/polygonize.py
```diff
@@ -0,0 +1,590 @@
...
+    if return_type == "numpy":
+        return column, polygon_points
+    elif return_type == "awkward":
+        return _to_awkward(column, polygon_points)
+    elif return_type == "geopandas":
+        return _to_geopandas(column, polygon_points, column_name)
+    elif return_type == "spatialpandas":
+        return _to_spatialpandas(column, polygon_points, column_name)
+    else:
+        raise ValueError(f"Invalid return_type '{return_type}'")
```
thuydotm, line 504: It would be better to validate `return_type` in `polygonize()` IMO

thuydotm (changes requested):
Just a few minor comments, otherwise look good to me!

xrspatial/experimental/polygonize.py
ianthomas23: Good catch!

xrspatial/experimental/polygonize.py
ianthomas23: OK, will do. I'll also add a new test passing an invalid string for `return_type` and checking that an appropriate exception is raised.

PR #586, opened 2021-11-26
zonal crosstab: speedup numpy case

thuydotm:


xrspatial/zonal.py
```diff
@@ -657,14 +654,155 @@ def _crosstab_dask(
...
+            cat_ids = unique_cats
+        # remove cats that does not exist in `values` raster
+        cat_ids = [c for c in cat_ids if c in unique_cats]
+    return unique_cats, cat_ids
+
+
+def _sort_values(sorted_indices, values, cats):
+    if len(values.shape) == 2:
+        # 2D case
+        result = values.flatten()[sorted_indices]
```
ianthomas23, line 680: Would it be better to use `ravel` here, which only copies if necessary, rather than `flatten` which always copies?

xrspatial/zonal.py
```diff
@@ -657,14 +654,155 @@ def _crosstab_dask(
...
+    else:
+        # 3D case
+        num_cats = len(cats)
+        h, w = values.shape[1:]
+        result = np.zeros(shape=(num_cats, h * w), dtype=values.dtype)
+        for i, cat in enumerate(cats):
+            result[i] = values[i].flatten()[sorted_indices]
```
ianthomas23, line 687: Same question about `ravel` vs `flatten`.

xrspatial/zonal.py
```diff
@@ -657,14 +654,155 @@ def _crosstab_dask(
...
+    cat_ids: Union[List, np.ndarray],
+    nodata_values: Union[int, float],
+    agg: str,
+) -> pd.DataFrame:
+    # find ids for all zones
+    unique_zones = np.unique(zones[np.isfinite(zones)])
+    # selected zones to do analysis
+    if zone_ids is None:
+        zone_ids = unique_zones
+    # remove zones that does not exist in `zones` raster
```
ianthomas23, line 762: `does` should be `do`.

xrspatial/zonal.py
```diff
@@ -657,14 +654,155 @@ def _crosstab_dask(
...
+    zone_ids = [z for z in zone_ids if z in unique_zones]
```
ianthomas23, line 763: Could this be in an `else` block of the previous `if` statement? (Like first review comment).

xrspatial/zonal.py
```diff
@@ -657,14 +654,155 @@ def _crosstab_dask(
...
+
+    crosstab_dict = {}
+    # zone column
+    crosstab_dict["zone"] = zone_ids
+    # stats columns
+    for cat in cat_ids:
+        crosstab_dict[cat] = []
+
+    flatten_zones = zones.flatten()
```
ianthomas23, line 772: `ravel` rather than `flatten` to only copy if necessary?

ianthomas23 (changes requested):

thuydotm:
Changed as suggested. Thanks @ianthomas23!

PR #591, opened 2021-12-01
Update release docs

ianthomas23:
Update docs on release process, as followed for release 0.3.0.


ianthomas23:
@thuydotm I am going to merge this. We can check it is correct at time of next release, early in January.

PR #596, opened 2021-12-14
crosstab: speedup dask case

thuydotm:


xrspatial/zonal.py
```diff
@@ -752,14 +600,133 @@ def _crosstab_numpy(
...
-                nodata_values, crosstab_dict, agg
+                nodata_values, crosstab_dict
             )
         start = end
 
+    if TOTAL_COUNT in crosstab_dict:
+        crosstab_dict[TOTAL_COUNT] = np.array(
+            crosstab_dict[TOTAL_COUNT], dtype=np.float32
+        )
+    for j, cat in enumerate(cat_ids):
```
ianthomas23, line 611: `j` not used here.

xrspatial/zonal.py
```diff
@@ -752,14 +600,133 @@ def _crosstab_numpy(
...
+
+    # construct output dataframe
+    if agg == 'percentage':
+        # replace 0s with nans to avoid dividing by 0 error
+        crosstab_dict[TOTAL_COUNT][crosstab_dict[TOTAL_COUNT] == 0] = np.nan
+        for cat in cat_ids:
+            crosstab_dict[cat] = crosstab_dict[cat] / crosstab_dict[TOTAL_COUNT] * 100  # noqa
+
     crosstab_df = pd.DataFrame(crosstab_dict)
+    crosstab_df = crosstab_df[['zone'] + [c for c in cat_ids]]
```
ianthomas23, line 622: If `cat_ids` is always a `list` or `ndarray`, can we just replace `[c for c in cat_ids]` with `list(cat_ids)`?

xrspatial/zonal.py
```diff
@@ -752,14 +600,133 @@ def _crosstab_numpy(
...
+            zone_values = _get_zone_values(values_by_zones, start, end)
+            _single_zone_crosstab(
+                zone_values, unique_cats, cat_ids,
+                nodata_values, results
+            )
+        start = end
+
+    if TOTAL_COUNT in results:
+        results[TOTAL_COUNT] = np.array(results[TOTAL_COUNT], dtype=np.float32)
+    for j, cat in enumerate(cat_ids):
```
ianthomas23, line 660: `j` not used.

xrspatial/zonal.py
```diff
@@ -752,14 +600,133 @@ def _crosstab_numpy(
...
+
+    if agg == 'percentage':
+        # replace 0s with nans to avoid dividing by 0 error
+        result[TOTAL_COUNT][result[TOTAL_COUNT] == 0] = np.nan
+        for cat in cat_ids:
+            result[cat] = result[cat] / result[TOTAL_COUNT] * 100
+
+    df = pd.DataFrame(result)
+    df['zone'] = zone_ids
+    columns = ['zone'] + [c for c in cat_ids]
```
ianthomas23, line 690: Use `list(cat_ids)` again here?

ianthomas23 (changes requested):
Just a few comments.

Also, `_sort_values` function at line 499 is no longer used so can be removed.

ianthomas23:
In future, it would be really good if performance improvement PRs could include an ASV benchmark as the output of that would show a side-by-side comparison of the performance before and after the changes.

PR #602, opened 2021-12-17
Test python 3.9 on CI

ianthomas23:
Trying use of python 3.9 on CI.


ianthomas23:
I've checked the test outputs, everything ran fine.

PR #604, opened 2021-12-17
Docs enhancement

thuydotm:
This PR refactors code, update tests and examples for the following modules:
- aspect
- classify
- focal

Function `general_output_checks()` is added to test that output attributes are preserved.

### DIFF
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -149,6 +149,13 @@ def _run_dask_numpy(data: da.Array) -> da.Array:
                            depth=(1, 1),
                            boundary=np.nan,
                            meta=np.array(()))
+
+    # Fill borders with nans to ensure nan edge effect.
+    out[0, :] = np.nan
+    out[-1, :] = np.nan
+    out[:,  0] = np.nan
+    out[:, -1] = np.nan
+
     return out
 
 
@@ -160,8 +167,19 @@ def aspect(agg: xr.DataArray,
     Calculates, for all cells in the array, the downward slope direction
     of each cell based on the elevation of its neighbors in a 3x3 grid.
     The value is measured clockwise in degrees with 0 (due north), and 360
-    (again due north). Flat areas are given a value of -1.
-    Values along the edges are not calculated.
+    (again due north). Values along the edges are not calculated.
+
+    Direction of the aspect can be determined by its value:
+    From 0     to 22.5:  North
+    From 22.5  to 67.5:  Northeast
+    From 67.5  to 112.5: East
+    From 112.5 to 157.5: Southeast
+    From 157.5 to 202.5: South
+    From 202.5 to 247.5: West
+    From 247.5 to 292.5: Northwest
+    From 337.5 to 360:   North
+
+    Note that values of -1 denote flat areas.
 
     Parameters
     ----------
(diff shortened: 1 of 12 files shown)


xrspatial/aspect.py
```diff
@@ -183,113 +196,90 @@ def aspect(agg: xr.DataArray,
...
+        >>> aspect_da = aspect(raster_da)
+        >>> print(aspect_da)
+        <xarray.DataArray 'aspect' (y: 6, x: 5)>
+        dask.array<_trim, shape=(6, 5), dtype=float64, chunksize=(3, 3), chunktype=numpy.ndarray>
+        Dimensions without coordinates: y, x
+        >>> aspect_da.compute()  # compute the results
+        <xarray.DataArray 'aspect' (y: 5, x: 5)>
+        array([[nan,     nan,        nan,        nan,          nan],
+               [nan, 120.96375653, 104.03624347, 135.        , nan],
+               [nan, 101.30993247, 108.43494882, 123.69006753, nan],
```
ianthomas23, line 252: I get values similar to the `numpy` values above. Maybe this is a copy and paste error?

xrspatial/tests/general_checks.py
```diff
@@ -0,0 +1,32 @@
...
+        # dask case
+        assert isinstance(
+            output_agg.data.compute(), type(input_agg.data.compute()))
+
+    # shape and other attributes remain the same
+    assert output_agg.shape == input_agg.shape
+    assert output_agg.dims == input_agg.dims
+    assert output_agg.attrs == input_agg.attrs
+    for coord in input_agg.coords:
+        assert np.all(output_agg[coord] == input_agg[coord])
```
ianthomas23, line 20: There is a `np.testing.assert_array_equal` that would be better here. But this applies throughout the codebase so maybe it should be in a separate issue/PR.

xrspatial/tests/general_checks.py
```diff
@@ -0,0 +1,32 @@
...
+
+    if expected_results is not None:
+        if isinstance(input_agg.data, da.Array):
+            assert np.isclose(
```
ianthomas23, line 24: Again for future reference `np.testing.assert_allclose` is a better alternative here.

xrspatial/utils.py
```diff
@@ -98,6 +98,10 @@ def is_dask_cupy(agg: xr.DataArray):
     return isinstance(agg.data, da.Array) and is_cupy_backed(agg)
 
 
+def not_implemented_func(agg, *args):
+    raise NotImplementedError('Not yet implemented.')
```
ianthomas23, line 102: The disadvantage here is that the exception is raised from a different point in the code than where the problem originally occurred. But the stack trace shows this, and the combined use of `not_implemented_func` and `ArrayTypeFunctionMapping` simplifies the code a lot so this is clearly a net improvement.

ianthomas23 (changes requested):
A few comments, but only one thing to check and possibly change.

Thanks @thuydotm!

xrspatial/aspect.py
thuydotm: Thanks for catching this, just fixed!

xrspatial/tests/general_checks.py
thuydotm: sounds good, let's make this change in a separate PR

xrspatial/utils.py
thuydotm: We can add a `messages` param to the `not_implemented_func()` to clearly state the reason if it fails.

PR #608, opened 2021-12-21
Add rtx gpu hillshade with shadows

ianthomas23:
This PR adds an `rtxpy` implementation of `hillshade` with full shadows. Output looks like this for UK elevation (not very well clipped at sea level but it is only an example):

out2

To use, call `hillshade` with a `cupy`-backed `xarray` and the keyword argument `shadows=True`. The algorithm aims to produce very similar results to the CPU algorithm where there aren't shadows, so follows the usual simple maths of conventional hillshade algorithms.

There are a few improvements to the rtxpy `viewshed` function too, to make it more robust. There is some similar code for both algorithms relating to generation of primary rays; they will be combined later when when I correctly deal with the presence or absence of `x,y` coordinates.

Thanks to @a7az0th who did most of the heavy lifting here.

### DIFF
--- a/xrspatial/gpu_rtx/hillshade.py
+++ b/xrspatial/gpu_rtx/hillshade.py
@@ -1,0 +1,197 @@
+# Only call functions in this file if has_rtx() returns True as this checks
+# that the required dependent libraries are installed.
+
+import cupy
+import numba as nb
+import numpy as np
+from rtxpy import RTX
+from scipy.spatial.transform import Rotation as R
+import xarray as xr
+
+from .cuda_utils import add, dot, invert, make_float3, mul
+from .mesh_utils import create_triangulation
+
+from ..utils import calc_cuda_dims
+
+
+@nb.cuda.jit
+def _generate_primary_rays_kernel(data, H, W):
+    """
+    A GPU kernel that given a set of x and y discrete coordinates on a raster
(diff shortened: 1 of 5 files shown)


xrspatial/gpu_rtx/hillshade.py
```diff
@@ -114,7 +114,7 @@ def _shade_lambert_kernel(hits, normals, output, H, W, sun_dir, cast_shadows):
         temp = (cos_theta + 1) / 2
 
         if cast_shadows and hits[i, j, 0] >= 0:
-            temp = temp / 2
+            temp = temp / 2.5
```
a7az0th, line 117: Please add a comment why we're multiplying by this constant here. To me dividing by 2 makes sense but 2.5 is odd

ianthomas23: Yes, the number is somewhat arbitrary. I think it will be best if I switch it back to being 2.

ianthomas23:
@thuydotm I've reverted the recent addition of `ArrayTypeFunctionMapping` in `hillshade.py` as the new `shadows` argument only exists on the rtx version of the function. We could get this to work by adding support for optional arguments or keyword arguments to `ArrayTypeFunctionMapping`, but I didn't want to potentially mess this up a day before release.

Issue #620, opened 2022-01-02
Add High-Level Analytics module


brendancol:
The purpose of the `analytics` module would be to provide compositions of existing xarray-spatial functions as high-order geospatial raster functions.

As an example, we could have `summarize_terrain` which would take an elevation `xr.DataArray` and return an `xr.Dataset` with `slope`, `aspect`, `curvature` variables with default naming convention like `input_arr.name + '-slope'`

ianthomas23:
Closed by #621.

PR #621, opened 2022-01-02
Stubbed out function for Analytics module

brendancol:
#620

### DIFF
--- a/xrspatial/analytics.py
+++ b/xrspatial/analytics.py
@@ -1,0 +1,17 @@
+from xrspatial import slope
+from xrspatial import curvature
+from xrspatial import aspect
+
+import xarray as xr
+
+
+def summarize_terrain(raster: xr.DataArray):
+
+    if raster.name is None:
+        raise NameError('Requires xr.DataArray.name property to be set')
+
+    ds = raster.to_dataset()
+    ds[f'{raster.name}-slope'] = slope(raster)
+    ds[f'{raster.name}-curvature'] = curvature(raster)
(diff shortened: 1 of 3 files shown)


ianthomas23:
I like this idea. Normally I would try to avoid the use of words such as `summarize` that have different spellings in English and American, but we already have `polygonize` and use of function names of the form `verb_noun` such as `generate_terrain`, so we should stick with the proposed name.

Function needs a docstring of course.

xrspatial/analytics.py
```diff
@@ -0,0 +1,88 @@
...
+
+    Returns
+    -------
+    summarized_terrain: xarray.Dataset
+        Dataset with slope, aspect, curvature variables with a naming convention of
+        `terrain.name-variable_name`
+
+    Examples
+    --------
+    .. sourcecode:: python
```
ianthomas23, line 26: Please can you put a line of whitespace after this `.. sourcecode:: python` line?

Otherwise when you generate the docs using sphinx it gives a warning. The large number of these warning got in the way of diagnosing the docs build failure earlier in the week, so we should adopt a policy of having the line of whitespace.

ianthomas23 (changes requested):
This all looks good, just a single character addition requested!

xrspatial/analytics.py
thuydotm: Thanks Ian, changed as requested!

ianthomas23:
Thanks @thuydotm!

PR #636, opened 2022-01-17
Tests improvements

### LINKED ISSUE #626: Improvements to tests
Recently there have been some big improvements to the tests, such as PRs #612 and #613. Here are a few other possible improvements.

  - [x] Tests using 2D arrays should always have `shape[0] != shape[1]`, just in case there are array transpose errors anywhere. e.g. `test_focal.py`s `convolve_2d_data` has shape `(6, 6)`.
  - [x] There are few more `np.isclose(...).all()` that should be replaced with `np.testing.allclose(...)`, e.g. `test_terrain.py` line 39.
  - [ ] There is a `from __future__ import absolute_import` in `test_tiles.py` that I don't think is needed. Such imports were usually related to Python 2/3 changes. But I have not fully researched this.
  - [x] Data for tests should probably use `pytest` fixtures (https://docs.pytest.org/en/6.2.x/fixture.html).

There are some other possible improvements that are really style issues. These affect more code than just the tests, and will probably need some discussion/agreement on policy.

  - [ ] I prefer to group together import of functions from the same file in a single line, rather than a line per import.
  - [x] I would happily increase our max line length from 79 to 100 or so.

thuydotm:
Fixes #626

### DIFF
--- a/xrspatial/tests/conftest.py
+++ b/xrspatial/tests/conftest.py
@@ -5,6 +5,7 @@ import numpy as np
 
 @pytest.fixture
 def random_data(size, dtype):
-    data = np.random.randint(low=-100, high=100, size=size)
+    rng = np.random.default_rng(2841)
+    data = rng.integers(-100, 100, size=size)
     data = data.astype(dtype)
     return data
--- a/xrspatial/tests/conftest.py
+++ b/xrspatial/tests/conftest.py
@@ -1,0 +1,10 @@
+import pytest
+
+import numpy as np
+
+
+@pytest.fixture
+def random_data(size, dtype):
+    data = np.random.randint(low=-100, high=100, size=size)
+    data = data.astype(dtype)
+    return data
--- a/xrspatial/tests/general_checks.py
+++ b/xrspatial/tests/general_checks.py
@@ -75,9 +75,7 @@ def assert_nan_edges_effect(result_agg):
         result_agg.data[:, -1],
     ]
     for edge in edges:
-        np.testing.assert_allclose(
-            edge, np.full(edge.shape, np.nan), equal_nan=True
-        )
+        np.testing.assert_array_equal(edge, np.nan)
 
 
 def assert_numpy_equals_dask_numpy(numpy_agg, dask_agg, func, nan_edges=True):
--- a/xrspatial/tests/general_checks.py
+++ b/xrspatial/tests/general_checks.py
@@ -88,14 +88,15 @@ def assert_numpy_equals_dask_numpy(numpy_agg, dask_agg, func, nan_edges=True):
     np.testing.assert_allclose(numpy_result.data, dask_result.data.compute(), equal_nan=True)
 
 
-def assert_numpy_equals_cupy(numpy_agg, cupy_agg, func, nan_edges=True):
+def assert_numpy_equals_cupy(numpy_agg, cupy_agg, func, nan_edges=True, atol=0, rtol=1e-7):
     numpy_result = func(numpy_agg)
     if nan_edges:
         assert_nan_edges_effect(numpy_result)
 
     cupy_result = func(cupy_agg)
     general_output_checks(cupy_agg, cupy_result)
-    np.testing.assert_allclose(numpy_result.data, cupy_result.data.get(), equal_nan=True)
+    np.testing.assert_allclose(
+        numpy_result.data, cupy_result.data.get(), equal_nan=True, atol=atol, rtol=rtol)
 
 
 def assert_numpy_equals_dask_cupy(numpy_agg, dask_cupy_agg, func, nan_edges=True):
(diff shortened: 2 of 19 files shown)


xrspatial/tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+import numpy as np
+
+
+@pytest.fixture
+def random_data(size, dtype):
+    data = np.random.randint(low=-100, high=100, size=size)
```
ianthomas23, line 8: Use of the original numpy random interface is deprecated, it is now recommended to use the `default_rng` instead.

Also, all tests should be deterministic so random number generators should have a `seed`, even though in this case we are only using the numbers to compare different algorithms, e.g. `numpy` vs `cupy`. Although the test should pass regardless of the random numbers generated, in practice we want to avoid any test failure occurring at some point in the future as the failure will occur in an unrelated PR and will be unrepeatable and hence very difficult to track down.

So here we could have something like:
```python
rng = np.random.default_rng(2841)
data = rng.integers(-100, 100, size=size, dtype=dtype)
```

xrspatial/tests/general_checks.py
```diff
@@ -29,23 +48,65 @@ def general_output_checks(input_agg: xr.DataArray,
...
-            output.get(), expected_results, equal_nan=True, rtol=1e-06
-        )
-        dask_cupy_func = lambda output, expected: np.testing.assert_allclose(  # noqa: E731, E501
-            output.compute().get(), expected_results,
-            equal_nan=True, rtol=1e-06
-        )
+        get_numpy_data = lambda output: output  # noqa: E731
+        get_dask_numpy_data = lambda output: output.compute()  # noqa: E731
+        get_cupy_data = lambda output: output.get()  # noqa: E731
+        get_dask_cupy_data = lambda output: output.compute().get()  # noqa: E731
```
ianthomas23, line 54: This is really nice!

xrspatial/tests/general_checks.py
```diff
@@ -29,23 +48,65 @@ def general_output_checks(input_agg: xr.DataArray,
...
+    # nan edge effect
+    edges = [
+        result_agg.data[0, :],
+        result_agg.data[-1, :],
+        result_agg.data[:, 0],
+        result_agg.data[:, -1],
+    ]
+    for edge in edges:
+        np.testing.assert_allclose(
+            edge, np.full(edge.shape, np.nan), equal_nan=True
```
ianthomas23, line 79: Because we are asserting that all values are equal to `np.nan` we can use the simpler
```python
np.testing.assert_array_equal(edge, np.nan)
```
Unless I have missed some subtlety when using `cupy` or `dask`.

xrspatial/tests/test_pathfinding.py
```diff
@@ -1,31 +1,67 @@
...
+def result_4_connectivity():
+    expected_result = np.array([[np.nan, 1, 0., np.nan],
+                                [np.nan, 2, np.nan, np.nan],
+                                [np.nan, 3, np.nan, np.nan],
+                                [np.nan, 4, np.nan, np.nan],
+                                [np.nan, np.nan, np.nan, np.nan]])
+    return expected_result
+
+
+def test_a_star_search_no_barries(input_data):
```
ianthomas23, line 56: Typo: `barries` should be `barrier` or `barriers` ?

xrspatial/tests/test_pathfinding.py
```diff
@@ -39,12 +75,15 @@ def test_a_star_search():
                         assert np.nanmax(path_agg) > 0
                         assert np.nanmin(path_agg) == 0
 
+
+def test_a_star_search_with_barries(input_data):
```
ianthomas23, line 79: `barries` typo again.

xrspatial/tests/test_pathfinding.py
```diff
@@ -78,38 +105,34 @@ def test_a_star_search():
     assert np.isnan(no_snap_path_agg).all()
```
ianthomas23, line 105: Could use this instead
```python
np.testing.assert_array_equal(no_snap_path_agg, np.nan)
```
and also in 2 places below.

ianthomas23 (changes requested):
Thanks @thuydotm, this is a huge amount of work and a significant improvement in readability.

There are a couple of places where we could use more fixtures and non-square arrays, but let's not make this PR any larger; we can come back to these details later.

I should probably run the `rtxpy` tests locally before we commit this in case there are any differences that the CI doesn't pick up. Some of my existing `rtxpy` tests were already dubious.

xrspatial/tests/conftest.py
thuydotm: Thanks Ian, this is a very good catch. As we use both int and float dtype for testing, I made a change that is slightly different. We'll convert to expected `dtype` after creating the array with `rng.integers()`

thuydotm:
Thanks @ianthomas23, I've made the changes as suggested. Can you please run the tests locally to see if it works with rtxpy?

ianthomas23 (approved):

Issue #609, opened 2021-12-22
Removed unwanted lines from .gitignore and MANIFEST.in


ianthomas23:
The latter controls which files are included in the sdist.

PR #642, opened 2022-01-26
Clean gitignore and manifest files

giancastro:
Fixes issue #609.


ianthomas23:
@giancastro Thanks for this. We'll take extra care with the next release to check that the right files are in the sdist and wheel.

ianthomas23 (approved):

PR #643, opened 2022-01-26
Benchmark results

jam-lock:
Add a .md file to aggregate benchmark results from different devices and OS


ianthomas23:
@jam-lock Thanks for this. Longer term we will have to find a way to easily include benchmark results in the documentation.

ianthomas23 (approved):

Issue #442, opened 2021-05-27
Handle error when examples folder already exists - CLI


giancastro:
`ValueError: Path xarray-spatial/xrspatial-examples already exists; please move it away, choose a different path, or use force.`

PR #644, opened 2022-01-26
handle CLI errors

giancastro:
Fixes issue #442.


ianthomas23 (approved):
I've confirmed that this gives an improved user experience.

PR #648, opened 2022-02-07
Add more benchmarks

thuydotm:
This PR refactor existing code in common.py so that we have a base class for benchmarking. It also adds following new benchmarks:
- curvature
- multispectral tools
- classify tools
- focal tools
- proximity tools
- pathfinding a star search
- perlin, terrain generation

### DIFF
--- a/benchmarks/benchmarks/aspect.py
+++ b/benchmarks/benchmarks/aspect.py
@@ -1,14 +1,10 @@
 from xrspatial import aspect
-from .common import get_xr_dataarray
+from .common import Benchmarking
 
 
-class Aspect:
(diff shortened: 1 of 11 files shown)


thuydotm:
@ianthomas23 Can you please take a look? The PR only adds new code, I'll update docs/comments where needed in a new PR. And the benchmarking results will come in another PR.

ianthomas23 (approved):
@thuydotm This is great! Many new benchmarks using some really compact code that is easy to understand and extend. I am looking forward to running the full benchmark suite, it looks like it will take quite some time now.

thuydotm:
Great, thanks Ian. Looking forward to seeing the benchmarking results on GPU.

PR #650, opened 2022-02-07
Run docs CI via workflow_dispatch

ianthomas23:
This alters the docs-publish github action to allow it to be run on demand via `workflow_dispatch`.

This is to make it easier to deal with doc build failures, such as happened with the v0.3.2 release last week.


ianthomas23:
I have to merge this so that the `workflow_dispatch` options appears in CI.

PR #657, opened 2022-02-16
More robust cuda and cupy identification

ianthomas23:
This PR makes the identification of cuda and cupy more robust. Previously it was possible to have cuda present but not have cupy installed, and some branches of code would view this as sufficient for running cupy code and hence fail.

The change is to add a new function `has_cuda_and_cupy` that combines the individual `has_cuda` and `has_cupy` functions, which are now made private with underscores. Existing code has been modified to use the new `has_cuda_and_cupy()`, and all new code should use this too.

Also added a `pytest.mark.skipif` decorator called `cuda_and_cupy_available` so we don't have to repeat the `reason` string throughout the code.

I have confirmed that tests run and pass on a cuda-enabled system in the following configurations:
  1. No cupy
  2. With cupy
  3. With cupy and rtxpy

### DIFF
--- a/benchmarks/benchmarks/common.py
+++ b/benchmarks/benchmarks/common.py
@@ -2,7 +2,7 @@ import numpy as np
 import xarray as xr
 
 from xrspatial.gpu_rtx import has_rtx
-from xrspatial.utils import has_cuda, has_cupy
+from xrspatial.utils import has_cuda_and_cupy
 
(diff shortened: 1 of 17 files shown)


brendancol:
@ianthomas23 nice this sounds good.  Ive added @thuydotm as reviewer

PR #665, opened 2022-02-21
User guide fixes

DahnJ:
I have been going through the [user guide](https://xarray-spatial.org/user_guide/index.html) and have fixed some smaller problems I found along the way.

I'm happy to make any requested further changes.

### Changes
- `1_Surface`: Link for avalanche risk was dead, used another that cites the same angles
- `2_Proximity`: typo
- `3_Zonal`: The path's coordinates appeared to be on a different scale, did not show up on the map and lead to strange results. I also used a better visible color map (not sure if sequential is okay here).
  - The original can be seen [in the docs](https://xarray-spatial.org/v0.3.2/user_guide/zonal.html). Below is my result:

image

image


thuydotm:
Thanks @DahnJ. The changes in Surface and Proximity notebook look good to me. 

In the Zonal notebook, can you change the range of x and y to: `x_range=(-20, 20)`, and `y_range=(-20, 20)` so that we can keep `zone_df` as is, and no need to multiply its xs and ys by `1e6`? 

I'm ok with using the Hot color map, let's strengthen the line by increase the threshold value in `dynspread()`, i.e: `zones_shaded = dynspread(shade(zones_agg, cmap=Set1), max_px=1, threshold=1)`

DahnJ:
> Thanks @DahnJ. The changes in Surface and Proximity notebook look good to me.
> 
> In the Zonal notebook, can you change the range of x and y to: `x_range=(-20, 20)`, and `y_range=(-20, 20)` so that we can keep `zone_df` as is, and no need to multiply its xs and ys by `1e6`?
> 
> I'm ok with using the Hot color map, let's strengthen the line by increase the threshold value in `dynspread()`, i.e: `zones_shaded = dynspread(shade(zones_agg, cmap=Set1), max_px=1, threshold=1)`

Thanks for the quick reply. I changed the range and went back to `Set1` with `threshold=1`, it's already nicely visible.

thuydotm:
Thanks for your contribution, this has been merged!

PR #667, opened 2022-02-22
move all tests to GitHub action

thuydotm:


thuydotm:
As we are moving all out tests to GitHub Actions, we can safely remove Travis, AppVeyor and tox. It seems that AppVeyor has been added to project settings so that it runs automatically even after deleting the `.appveyor.yml` config file. @brendancol Can you remove it from the project settings?

PR #670, opened 2022-02-23
Remove stale test files from project root

brendancol:
Looks like we can remove these test files.  @thuydotm can you confirm?

## Proposed Changes

  - Remove test_cli file from root
  - Remove test_pip_packaging from root


thuydotm:
LGTM. Thanks @brendancol

PR #677, opened 2022-03-04
Don't calculate angle when not needed

akupitz:
use less calculations in points with same y value

Fixes #

## Proposed Changes

  -
  -
  -


thuydotm:
Thanks for making this change! LGTM.

PR #711, opened 2022-05-24
Fix rtx viewshed rendering blank image

a7az0th:
Fix for wrong camera ray generation in viewshed. 

While debugging the above issue I noticed that the raster had some sort of scaling applied to the latitude and longitude, but not to the height. I took that scale factor into account while triangulating the terrain and now Hillshade and Viewshed
produce almost identical results compared to their numpy counterparts.

image
On the first row you have:
the original terrain | numpy viewshed | cupy rtx viewshed
On the second row you have:
the original terrain | numpy hillshade | cupy rtx viewshed with shadows


thuydotm:
Thanks @a7az0th, I'll merge this to the main branch and we can address the issue with discrepancy between CPU and GPU later.

PR #715, opened 2022-06-05
Improved viewshed rtx. Now result should match the CPU version

a7az0th:
Tried to mimic what the CPU does as best as I can


thuydotm:
LGTM

PR #725, opened 2022-06-22
WIP Housing price feature engineering notebook

thuydotm:
Add new notebook to discover how geo-related aspects affect housing price in the USA. We will consider a house based on the information provided in an existing dataset with some addtional spatial attributes extracted from its location using xarray-spatial (and probably some elevation dataset, and census-parquet as well?).

We'll first build a machine learning model and train it with all existing features. For each newly added feature, we'll retrain it and compare the results to find out which features help enrich the model.

To be determined:
- What dataset to work on as orignal housing price dataset?
- What are the new features? Slope using an elevation dataset? Population density, ... from Census if none available in the existing features? Proximity distance to nearest hospital / grocery store / university / pharmacy?

### DIFF
--- a/examples/housing_price_feature_engineering.ipynb
+++ b/examples/housing_price_feature_engineering.ipynb
@@ -1,0 +1,311 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "housing_price_feature_engineering",
+      "provenance": [],
+      "collapsed_sections": []
+    },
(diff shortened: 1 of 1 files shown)


brendancol:
@thuydotm awesome! I think it would be interesting to look at housing prices as related to viewshed.  For example, how does  have a view shed which intersects with a beach or a nuclear power plant influence price?  The national building footprint dataset could be helpful with this.

brendancol (approved):

PR #755, opened 2023-01-13
Multispectral tools: convert data to float32 dtype before doing calculations

### LINKED ISSUE #726: Overflow error with NDVI calculation
**Describe the bug**
ndvi is giving incorrect results when input arrays are unsigned integer due to overflow in subtraction. 

**Expected behavior**
Consider the following

```
import numpy as np
import xarray as xr
import xrspatial

a = xr.DataArray(np.array([[1,1,1],[1,1,1]], dtype='uint16'))
b = xr.DataArray(np.array([[0,1,2],[0,1,2]], dtype='uint16'))
xrspatial.ndvi(a,b)
```
Results:
```
<xarray.DataArray 'ndvi' (dim_0: 2, dim_1: 3)>
array([[1.000000e+00, 0.000000e+00, 6.148915e+18],
       [1.000000e+00, 0.000000e+00, 6.148915e+18]], dtype=float32)
Dimensions without coordinates: dim_0, dim_1
```
Values in third column are due to overflow of unsigned integer during subtraction, see https://github.com/numpy/numpy/issues/21237.
Using the above data, this can be illustrated by 
```
>>> a[0,2] - b[0,2]
<xarray.DataArray ()>
array(65535, dtype=uint16)
```

A solution could be to call np.subtract directly in e.g. `_normalized_ratio_cpu`, though that may not be robust to all cases. I think there are some alternatives in the numpy issues linked above
```
>>> np.subtract(a[0,2], b[0,2], dtype='int32')
<xarray.DataArray ()>
array(-1, dtype=int32)

```

thuydotm:
Fixes #726.

TODO:
- [x] add tests

### DIFF
--- a/xrspatial/multispectral.py
+++ b/xrspatial/multispectral.py
@@ -146,7 +146,9 @@ def arvi(nir_agg: xr.DataArray,
                                       cupy_func=_arvi_cupy,
                                       dask_cupy_func=_arvi_dask_cupy)
 
-    out = mapper(red_agg)(nir_agg.data, red_agg.data, blue_agg.data)
+    out = mapper(red_agg)(
+        nir_agg.data.astype('f4'), red_agg.data.astype('f4'), blue_agg.data.astype('f4')
+    )
 
     return DataArray(out,
                      name=name,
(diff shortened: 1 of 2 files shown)


brendancol (changes requested):
I know f32 is ideal for performance.  But this sounds a little dangerous if the user is assuming there data stays as the same type during calculation.  Could add a little more detail on why this is important to cast?

brendancol (approved):

brendancol:
@thuydotm Can you follow up with a PR with documentation of how these types are handled?

https://github.com/makepath/xarray-spatial/issues/766

PR #759, opened 2023-01-18
Zonal crosstab user guide notebook enhancement

thuydotm:
A few improvements to the notebook:
- use Fahrenheit scale instead of Kelvin as Fahrenheit is more familiar to users in general.
- add more plots to make the results more intuitive


brendancol (approved):
I think we can continue to evolve this example to fit some different use cases. But this is a good addition and we can open more in the future

PR #763, opened 2023-01-19
classify.binary: handle NaNs and infinite values

### LINKED ISSUE #757: classify.binary(): handle NaN values
classify.binary() currently classified NaNs as 0s. We should keep NaNs as NaNs though.

thuydotm:
Fixes #757

To be consistent with all other tools in classify module, this PR changes the way classify.binary() treats NaNs and infinite values. It now considers those values as outliers and classify them as NaNs instead of 0s as of previously.

- [x] numpy case
- [x] cupy case


brendancol (approved):
Hey @thuydotm this LGTM

PR #796, opened 2024-04-14
Bunch of Fixes

brendancol:
The following PR includes a bunch of fixes that I need to get CI/CD working again.

The most technically unstable is an update to `experimental/polygonize.py` which now has a [custom implementation](https://github.com/numba/numba/issues/8466) of `generated_jit`, but should in the future use `nb.extends.overload` style...

I also:
- got codecov working again
- added Python 3.12 support to CI/CD
- dropped 3.8 / 3.9 support from CI/CD


tcmetzger (approved):
Thanks!

PR #797, opened 2024-04-22
Move docs to readthedocs [WIP]

tcmetzger:
Fixes #794

PR #798, opened 2024-04-23
Update readthedocs configuration [WIP]

tcmetzger:
Follow-up to #797


brendancol (approved):

tcmetzger:
Turns out the extra_requirements config parameter (https://docs.readthedocs.io/en/stable/config-file/v2.html#packages) on readthedocs currently isn't working properly, so I added a `pre_build` step instead. This way, readthedocs are now able to install the docs dependencies defined in setup.cfg.

Issue #1215, opened 2026-04-17
Security: unbounded allocation via TIFF tile dimensions (DoS)


brendancol:
## Summary

The TIFF reader checks image dimensions but ignores tile dimensions. An adversarial TIFF can claim a 1x1 image (passes the existing guard) with a 2^30 x 2^30 tile (unchecked), and the decompressor will try to allocate terabytes.

## Reproducer

Build a TIFF with:
- `ImageWidth=1, ImageLength=1` (passes the default ~1B pixel guard)
- `TileWidth=2^30, TileLength=2^30` (unchecked)

`_decode_strip_or_tile` in `xrspatial/geotiff/_reader.py` computes `pixel_count = tw * th * samples ≈ 2^60` and passes it to `decompress()`. For LZW, `lzw_decompress` then does `np.empty(expected_size, dtype=np.uint8)` at terabyte scale. OOM.

## Affected paths

- `_read_tiles` (`_reader.py:436`) checks image dims at line 494 but not `tw`/`th`.
- `_read_cog_http` (`_reader.py:578`) has the same gap.
- `read_geotiff_gpu` (`__init__.py:1005`) checks image dims at line 1109, then `gpu_decode_tiles` does `cupy.zeros(n_tiles * tile_width * tile_height * bytes_per_pixel)` with no cap.

The strip path happens to be safe because `strip_rows = min(rps, height - strip_row)` pins it to the (checked) image height.

## Threat model

TIFF from an untrusted source (HTTP, user upload, third-party mosaic) triggers unbounded allocation. No code execution, just DoS -- but a one-file DoS.

## Proposed fix

After reading tile dims from the IFD, call `_check_dimensions(tile_width, tile_height, samples, max_pixels)` in `_read_tiles`, `_read_cog_http`, and `read_geotiff_gpu`. Legitimate COGs use `tile_width <= 512`, so capping per-tile pixels at `max_pixels` rejects the attack without breaking anything real.

## Tests

- A TIFF with small image dims + huge tile dims raises `ValueError`.
- Normal tile sizes (256, 512) still pass.

Found during a security audit of the geotiff subpackage.

PR #1221, opened 2026-04-19
Reject TIFFs whose declared tile grid exceeds TileOffsets length (#1219)

### LINKED ISSUE #1219: Security: OOB GPU read when TIFF tile grid exceeds TileOffsets length
**Describe the bug**

`_assemble_tiles_kernel` in `xrspatial/geotiff/_gpu_decode.py` reads `tile_out_offsets[tile_idx]` where `tile_idx` is computed from the output pixel's position (`tile_row * tiles_across + tile_col`). If a TIFF declares image dimensions implying more tiles than its TileOffsets tag supplies, `tile_idx` can exceed `len(tile_out_offsets)`. The kernel then reads past the end of device memory.

The CPU path (`_read_tiles`) silently skips tiles whose index exceeds the offsets array, so the caller gets a zero-padded output instead of an error.

**Reproduction**

Craft a TIFF where:
- `ImageWidth` / `ImageLength` and `TileWidth` / `TileLength` imply `tiles_across * tiles_down = N` tiles.
- `TileOffsets` tag has count `M` with `M < N`.

On the GPU path (`open_geotiff_gpu`), the kernel launches one thread per output pixel. Threads whose pixel maps to a tile index `>= M` read past the end of `d_decomp_offsets`.

**Expected behavior**

The reader should reject the file with a `ValueError` during header parsing, naming the mismatch between the declared tile grid and the TileOffsets length.

**Impact**

- GPU path: out-of-bounds device read. Produces garbage pixels, or trips a CUDA illegal memory access that kills the process.
- CPU path: silent wrong output (zero-filled tiles where offsets were missing).
- Adversarial input only. Files written by `xrspatial.geotiff._writer` or GDAL always have matching counts.

**Context**

Found during `/sweep-security` audit of the geotiff subpackage. Related to #1215 (tile dimension guard) and a separate predictor-kernel finding filed as a sibling issue.

**Desktop:**
- OS: Linux
- Python: 3.11
- xarray-spatial: current master

brendancol:
## Summary

A TIFF can declare `ImageWidth / ImageLength` and `TileWidth / TileLength` that imply `tiles_across * tiles_down = N` tiles while its `TileOffsets` tag has fewer than `N` entries. The GPU `_assemble_tiles_kernel` computes `tile_idx = tile_row * tiles_across + tile_col` from the output pixel position and reads `tile_out_offsets[tile_idx]` past the end of the device buffer. The CPU path silently skips the missing tiles (`if tile_idx >= len(offsets): continue` in `_read_tiles`), returning a zero-padded raster with no error.

## Fix

Fail fast in the header layer. A new `validate_tile_layout(ifd)` in `_header.py` checks that `len(tile_offsets) >= tiles_across * tiles_down` (times `samples_per_pixel` when planar config is 2) and raises `ValueError` with the mismatch. It's called from:

- `_read_tiles` (the windowed-read path)
- `_read_cog_http` (the HTTP COG path)
- `open_geotiff_gpu` in `geotiff/__init__.py` before the GPU dispatch

## Test plan

`TestTileLayoutValidation` in `xrspatial/geotiff/tests/test_security.py`:

- [x] `validate_tile_layout` raises when TileOffsets count is smaller than the declared grid
- [x] `validate_tile_layout` accepts a well-formed tiled TIFF
- [x] `validate_tile_layout` is a no-op for stripped TIFFs
- [x] `_read_tiles` raises `ValueError` instead of silently zero-padding
- [x] `read_to_array` end-to-end raises `ValueError` on the malformed input
- [x] Exact-count TIFFs still read correctly

The helper `_make_short_offsets_tiff` builds a normal tiled TIFF and rewrites the TileOffsets IFD entry to advertise a smaller count without truncating the data blob, so the resulting file passes every earlier check.

Full geotiff suite: 441 passed, 4 skipped. 3 pre-existing matplotlib-deepcopy failures in `TestPalette` reproduce on master.

Closes #1219. Related to #1215.

### DIFF
--- a/.claude/sweep-security-state.json
+++ b/.claude/sweep-security-state.json
@@ -10,7 +10,8 @@
       "last_inspected": "2026-04-17",
       "issue": 1215,
       "severity_max": "HIGH",
-      "categories_found": [1, 4]
+      "categories_found": [1, 4],
+      "followup_issues": [1219]
     }
   }
 }
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -1068,7 +1068,7 @@ def read_geotiff_gpu(source: str, *,
             "Install it with: pip install cupy-cuda12x")
 
     from ._reader import _FileSource, _check_dimensions, MAX_PIXELS_DEFAULT
-    from ._header import parse_header, parse_all_ifds
+    from ._header import parse_header, parse_all_ifds, validate_tile_layout
     from ._dtypes import tiff_dtype_to_numpy
     from ._geotags import extract_geo_info
     from ._gpu_decode import gpu_decode_tiles
@@ -1134,6 +1134,11 @@ def read_geotiff_gpu(source: str, *,
         _check_dimensions(width, height, samples, max_pixels)
         # A single tile's decoded bytes must also fit under the pixel budget.
         _check_dimensions(tw, th, samples, max_pixels)
+
+        # Reject malformed TIFFs whose declared tile grid exceeds the
+        # supplied TileOffsets length. The GPU tile-assembly kernel would
+        # read OOB otherwise. See issue #1219.
+        validate_tile_layout(ifd)
 
     finally:
         src.close()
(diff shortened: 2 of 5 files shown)


brendancol:

Issue #1206, opened 2026-04-15
bump: uint16 overflow, unbounded default count, and closure serialization OOM


brendancol:
**Describe the bug**

`bump()` has three problems that produce wrong results or OOM on large rasters.

**uint16 coordinate overflow** (`bump.py:295`): `locs` uses `dtype=np.uint16`, so any dimension over 65,535 wraps silently. A coordinate of 70,000 becomes 4,464. Bumps land in the wrong place with no warning.

**Unbounded default count** (`bump.py:289`): When `count` is omitted it defaults to `w * h // 10`. That's 419,430 bumps at 2048x2048. For a 30TB raster (~174k x 174k) it's roughly 3 billion -- about 30 GB just for `locs` and `heights` before any work starts.

**Full locs/heights arrays in every dask chunk closure** (`bump.py:48-99`): `_bump_dask_numpy` and `_bump_dask_cupy` close over the full `locs` and `heights`. Each chunk in the task graph gets its own serialized copy. At 2048x2048 default count with 128x128 chunks that's 5 MB per chunk across 256 chunks = 1.29 GB of graph payload.

**Expected behavior**

- Coordinates should work for rasters wider/taller than 65,535 px without silent truncation.
- Default count should cap at a reasonable ceiling for large inputs.
- Dask closures should only carry bumps that fall within each chunk's spatial extent (plus `spread` padding).

**Benchmark (512x512, count=100)**

| Backend | Wall time (ms) | Peak traced (MB) |
|---------|---------------|-------------------|
| numpy | 0.52 | 23.01 |
| dask+numpy | 22.0 | 17.96 |

**Closure serialization (2048x2048, default count=419,430)**

- Per-chunk payload: 5.03 MB
- 256 chunks: 1.29 GB total
- At 30TB scale the graph serialization alone blows past available memory

PR #1234, opened 2026-04-23
Reject oversize bump output rasters before allocation (#1231)

### LINKED ISSUE #1231: bump() output raster allocation lacks memory guard
**Describe the bug**

`bump()` has a memory guard on the `locs`/`heights` bump arrays (added in #1206), but the output raster itself is allocated with no check. `_finish_bump` at `xrspatial/bump.py:45` calls `np.zeros((height, width))` where `height` and `width` come from user parameters or `agg.shape`.

A caller can trigger a huge allocation:

```python
from xrspatial import bump
bump(width=1_000_000, height=1_000_000)
```

With the default `count = min(w*h//10, 10_000_000) = 10_000_000`, the bump-count guard sees only `10M * 16 = 160 MB` and passes. Then `_finish_bump` tries to allocate `np.zeros((1_000_000, 1_000_000))` which is ~8 TB of float64 and OOMs the host.

**Expected behavior**

The memory budget check should also account for the output raster bytes (`8 * height * width` for float64) and raise `MemoryError` before any allocation happens when the required memory is larger than roughly half of available memory.

**Additional context**

Only the numpy and cupy backends are affected. The dask paths (`_bump_dask_numpy`, `_bump_dask_cupy`) build the output lazily via `da.from_delayed` and `da.zeros` on per-chunk tiles, so they never materialize the full raster. See #1206 for the earlier bump-count guard.

brendancol:
## Summary

- `_finish_bump` allocated `np.zeros((height, width))` with no memory check. The count guard added in #1206 only covered the `locs`/`heights` arrays, so `bump(width=1_000_000, height=1_000_000)` slipped through (count capped at 10M ~ 160 MB) and then tried to allocate ~8 TB for the output raster.
- Extended the memory budget check to include `raster_bytes = w * h * 8` when the backend will materialize the full array. Dask paths build output lazily per-chunk, so they are excluded from the raster guard and keep handling large dimensions.
- Added regression tests for the pathological case and for the dask-bypass case. The existing count-guard test still passes.

Fixes #1231.

## Test plan

- [x] `pytest xrspatial/tests/test_bump.py`: all 18 tests pass locally (15 existing + 3 new)
- [x] `bump(width=1_000_000, height=1_000_000)` now raises `MemoryError` with a message naming the output raster size
- [x] `bump(agg=dask_array_100k_x_100k, count=10)` still works lazily without tripping the guard


brendancol:

PR #1238, opened 2026-04-23
Clamp bilateral kernel radius to raster extent (#1236)

### LINKED ISSUE #1236: bilateral: unbounded sigma_spatial enables memory and compute DoS
**Describe the bug**

`bilateral()` in `xrspatial/bilateral.py` validates `sigma_spatial` only as `> 0`, with no upper bound. The kernel radius is computed as `radius = ceil(2 * sigma_spatial)` and then used to pad the input (when `boundary != 'nan'`) and to size the neighborhood loop in both the numba CPU kernel and the CUDA kernel.

Padding allocates shape `(H + 2*radius, W + 2*radius)`, so the allocation scales quadratically in `sigma_spatial`. With no cap, one float from the caller can DoS the host:

- `sigma_spatial=1e9` on a 100x100 raster gives `radius=2_000_000_000`. The padded float64 array would need about 1.28 x 10^20 bytes (128 EB).
- `sigma_spatial=1e5` gives radius 200000, about 320 GB padded on any input.
- Even with `boundary='nan'` (no padding), the clamped inner loop runs `min(rows, 2r+1) * min(cols, 2r+1)` iterations per pixel. On a 10000x10000 raster with `sigma_spatial=10000`, that is roughly 10^16 ops.

**Expected behavior**

Reject `sigma_spatial` values whose derived radius exceeds `max(rows, cols)` with a `ValueError` that names the offending value and the cap. A radius that large already covers the whole raster.

**Reproduction**

```python
import numpy as np
import xarray as xr
from xrspatial import bilateral

agg = xr.DataArray(np.zeros((100, 100)))
bilateral(agg, sigma_spatial=1e9)  # tries to allocate exabytes or runs effectively forever
```

**Affected backends**

numpy, cupy, dask+numpy, dask+cupy. The check lives in the public entrypoint, so one cap covers every backend.

**Severity**

HIGH. A caller passing one float to the public API can exhaust server memory or CPU.

**Category**

Cat 1 (Unbounded allocation / DoS) per the security sweep.

**Additional context**

Found during the bilateral security sweep on 2026-04-23. One-fix-per-PR convention applies.

brendancol:
## Summary

Fixes #1236. `bilateral()` validated `sigma_spatial` only as `> 0`, with no upper bound. The derived `radius = ceil(2 * sigma_spatial)` fed `_pad_array` (allocating `H+2r` by `W+2r` when `boundary != 'nan'`) and the dask `map_overlap` depth, so one float from the caller could exhaust host memory. `sigma_spatial=1e9` on a 100x100 raster maps to `radius=2e9` and about 128 EB of float64 padding.

The numba and CUDA inner loops are already clamped to `rows`/`cols`, so the filter output is unchanged once the radius is at least `max(rows, cols)`. This PR clamps the radius to `max(rows, cols)` in `bilateral()` before dispatch. The existing single-cell and 3D tests still pass.

Also records the finding in `.claude/sweep-security-state.json`.

## Test plan

- [x] `pytest xrspatial/tests/test_bilateral.py` -- 19 passed (all backends including GPU and dask+GPU)
- [x] New `test_bilateral_clamps_oversize_sigma_spatial` covers `boundary='nan'`, `boundary='nearest'`, and 3D inputs with `sigma_spatial=1e9`
- [x] Manual timing check: `bilateral(100x100, sigma_spatial=1e9)` now returns in seconds instead of allocating exabytes

### DIFF
--- a/.claude/sweep-security-state.json
+++ b/.claude/sweep-security-state.json
@@ -75,6 +75,13 @@
       "severity_max": "HIGH",
       "categories_found": [1],
       "notes": "HIGH (fixed #1231): _finish_bump allocated np.zeros((height, width)) with no memory guard. The existing count guard (added in #1206) only protected the locs/heights arrays, so bump(width=1_000_000, height=1_000_000) passed the guard (count capped at 10M ~ 160 MB) and then tried to allocate an 8 TB float64 raster. Fixed by extending the memory budget check to include raster_bytes = w * h * 8 when the backend will materialize the full array; dask paths build per-chunk and are excluded. No other HIGH findings: _bump_dask_numpy/_bump_dask_cupy build output lazily via da.from_delayed, no CUDA kernels (cupy path wraps the numba CPU kernel), no file I/O, no int32 overflow in realistic scenarios. MEDIUM (unfixed, Cat 6): bump() does not call _validate_raster on agg (dtype is not checked; shape unpacking catches wrong-ndim, but a non-numeric DataArray would fail confusingly downstream)."
+    },
+    "bilateral": {
+      "last_inspected": "2026-04-23",
+      "issue": 1236,
+      "severity_max": "HIGH",
+      "categories_found": [1],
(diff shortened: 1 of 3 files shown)


brendancol:

Issue #1241, opened 2026-04-23
convolution: unbounded circle_kernel/annulus_kernel radius enables memory DoS


brendancol:
**Describe the bug**

`circle_kernel()` and `annulus_kernel()` in `xrspatial/convolution.py` take a `radius` (and `outer_radius`/`inner_radius`) with no upper bound. The kernel is built by `_ellipse_kernel(half_w, half_h)`, where `half_w = int(radius_meters / cellsize_x)` and `half_h = int(radius_meters / cellsize_y)`. The float64 array that comes out has shape `(2*half_h + 1, 2*half_w + 1)`, so memory grows quadratically with the radius.

With no cap, one number from the caller can OOM the host:

- `circle_kernel(1, 1, 100000)` gives `half_w = half_h = 100000`, so the float64 kernel is `(200001, 200001)` ~ 320 GB.
- `circle_kernel(1, 1, 1000000)` is ~32 TB.
- `annulus_kernel` calls `circle_kernel` twice, so the same hole applies to both its radii.

The kernel shape is also used as the convolution depth in `convolve_2d`, so the pad and dask `map_overlap` depth blow up with it.

**Expected behavior**

Reject a kernel whose required memory exceeds roughly half of available memory with a `MemoryError` that names the radius and the memory estimate.

**Reproduction**

```python
from xrspatial.convolution import circle_kernel
circle_kernel(1, 1, 100000)  # tries to allocate ~320 GB
```

**Affected functions**

- `circle_kernel` (direct)
- `annulus_kernel` (via two `circle_kernel` calls)

Both are public API. They are re-exported by `xrspatial.focal` and `xrspatial.convolution` and used across the docs and user-guide notebooks.

**Severity**

HIGH. One number passed to the public API can OOM the server.

**Category**

Cat 1 (unbounded allocation / DoS) per the security sweep.

**Additional context**

Found during the convolution security sweep on 2026-04-23. Same pattern as the bilateral fix (#1236) and the bump raster fix (#1231). One-fix-per-PR convention applies.

PR #1258, opened 2026-04-24
Guard morphology kernel allocations against oversized kernels (#1256)

### LINKED ISSUE #1256: Guard morphology kernel allocations against oversized kernels
## Summary

The public morphology API (`morph_erode`, `morph_dilate`, `morph_opening`, `morph_closing`, `morph_gradient`, `morph_white_tophat`, `morph_black_tophat`) takes a user-supplied `kernel` argument. The only checks are 2D shape, uint8 dtype, and odd dimensions. Kernel size drives the padded input allocation on every backend, with no cap.

## Problem

In `xrspatial/morphology.py`:

- `_morph_numpy` (around line 152): `np.pad(arr, ((hy, hy), (hx, hx)), ...)` builds a padded float64 array of shape `(rows + ky - 1, cols + kx - 1)`.
- `_morph_cupy` (around line 245): same pattern on the GPU via `cp.pad`.
- Dask paths call `map_overlap(..., depth=(hy, hx))`, which grows per-chunk memory by the same factor.

A kernel like `np.ones((99999, 99999), dtype=np.uint8)` on a 1000x1000 raster will ask the process for roughly 80 GB of padded float64 memory with no warning. `bilateral.py` (#1236), `convolution.py` (#1241), and `bump.py` (#1231) have already fixed this class of issue with an `_available_memory_bytes()` guard.

## Proposed fix

1. Add a local `_available_memory_bytes()` helper copied from `convolution.py`. It reads `/proc/meminfo`, falls back to `psutil`, then to a 2 GB default.
2. Add `_check_kernel_memory(rows, cols, ky, kx)` that raises `MemoryError` if the padded float64 array would exceed 50% of available RAM. The message should name the kernel shape, the raster shape, and suggest a smaller kernel.
3. Call the check inside `_dispatch()` after `_validate_kernel` so every public entry point is covered across all four backends.

## Impact

Severity: HIGH (Cat 1, Unbounded Allocation / DoS). Found during the morphology security sweep on 2026-04-24. Nobody passes a kernel larger than their raster in practice, so the guard should not affect real workflows.

## Tests

Add a test that patches `_available_memory_bytes` to a tiny value and checks that `morph_erode` raises `MemoryError` with an informative message when a large kernel is passed.

brendancol:
## Summary

- The morphology public API (`morph_erode`, `morph_dilate`, `morph_opening`, `morph_closing`, `morph_gradient`, `morph_white_tophat`, `morph_black_tophat`) only validated kernel shape, dtype, and odd size. Kernel dimensions drove `np.pad`/`cp.pad` on every backend and `map_overlap` depth on the dask paths, so a 99999x99999 kernel on a 1000x1000 raster would ask for ~80 GB of padded float64 memory without a warning.
- Add `_available_memory_bytes()` (copied from `convolution.py`) and `_check_kernel_memory(rows, cols, ky, kx, func_name)` to `xrspatial/morphology.py`. The guard raises `MemoryError` before any allocation when the padded float64 array would exceed 50% of available RAM. The message includes the kernel shape, the raster shape, and a hint to use a smaller kernel.
- Wire the check into `_dispatch()` after `_validate_kernel` so every public API call is covered across numpy, cupy, dask+numpy, and dask+cupy.
- Matches the fix shipped for bilateral (#1236), convolution (#1241), and bump (#1231).

Closes #1256.

## Test plan

- [x] `pytest xrspatial/tests/test_morphology.py` -- 35 passed
- [x] Three new tests:
  - `test_oversized_kernel_raises_memory_error` patches `_available_memory_bytes` to 1 MB and checks that a 101x101 kernel on a 100x100 raster raises `MemoryError` with the expected message.
  - `test_kernel_memory_guard_allows_normal_use` checks that a default 3x3 kernel still works with a realistic 2 GB budget.
  - `test_kernel_memory_guard_runs_for_all_public_apis` checks the guard fires for `morph_erode`, `morph_dilate`, `morph_opening`, `morph_closing`, `morph_gradient`, `morph_white_tophat`, and `morph_black_tophat`.
- [x] Existing tests still pass (backend dispatch, NaN propagation, dataset support, cross-backend parity).


brendancol:

Issue #1240, opened 2026-04-23
contours(): unbounded segment buffer allocation enables memory DoS


brendancol:
**Describe the bug**

`contours()` allocates segment buffers sized to the full raster with no memory check. In `_contours_numpy()` at `xrspatial/contour.py:303-308`:

```python
max_segs_per_level = (ny - 1) * (nx - 1) * 2
for level in levels:
    seg_rows = np.empty((max_segs_per_level, 2), dtype=np.float64)
    seg_cols = np.empty((max_segs_per_level, 2), dtype=np.float64)
    ...
```

Each buffer holds float64 pairs (16 bytes per segment), so peak per level is `32 * (ny-1) * (nx-1)` bytes. A 10000x10000 raster uses ~3.2 GB per level; 20000x20000 uses ~12.8 GB per level. `_stitch_segments` then builds a Python dict keyed by every rounded endpoint on top of that.

`contours()` does not call `_validate_raster` and does not bound `n_levels` or the input shape, so the allocation is driven straight by a DataArray parameter.

**To reproduce**

```python
import numpy as np
import xarray as xr
from xrspatial import contours

data = np.zeros((20000, 20000), dtype=np.float64)
data[::100, :] = 1.0  # many crossings
agg = xr.DataArray(data, dims=['y', 'x'])
contours(agg, n_levels=10)  # tries ~12.8 GB per level
```

**Expected behavior**

The numpy path should estimate peak allocation (`2 * 16 * (ny-1) * (nx-1)` bytes for the segment buffers) and raise `MemoryError` before `np.empty` when the estimate exceeds about half of available memory. The error should mention the dask backend or a smaller `n_levels`.

**Scope**

- numpy: affected
- cupy: affected (transfers to CPU and calls `_contours_numpy`)
- dask+numpy, dask+cupy: not affected (each chunk is bounded by chunk size)

**Additional context**

Matches the memory guards in `bump.py` (#1231), `viewshed.py` (#1229), `bilateral.py` (#1236), `pathfinding.py`, and `proximity.py`, all of which call `_available_memory_bytes()` before large allocations.

Issue #1229, opened 2026-04-22
Add memory guard to viewshed numpy path


brendancol:
## Describe the bug

`_viewshed_cpu` in `xrspatial/viewshed.py` allocates several arrays sized from the raster dimensions with no memory check. Peak working memory is around 500 bytes per pixel once you count the sort buffer, so a 20000x20000 raster (400M pixels) tries to allocate ~200 GB and either OOM-kills or thrashes swap.

The dask path already has a guard — see `_viewshed_dask` around line 2172: it checks `r2_bytes` against `_available_memory_bytes()` and raises `MemoryError` with a hint to use `max_distance`. The numpy path has nothing.

## To Reproduce

```python
import numpy as np
import xarray as xr
from xrspatial import viewshed

# Tries to allocate a multi-hundred-GB event_list before any check
arr = np.zeros((20000, 20000), dtype=np.float64)
xs = np.arange(20000, dtype=float)
ys = np.arange(20000, dtype=float)
r = xr.DataArray(arr, coords=dict(x=xs, y=ys), dims=["y", "x"])
viewshed(r, x=10000.0, y=10000.0, observer_elev=5)
```

## Expected behavior

`MemoryError` before allocation, with a message pointing users at `max_distance=`.

## Proposed fix

Guard at the top of `_viewshed_cpu`: estimate peak working memory (~500 * H * W bytes) via `_available_memory_bytes()` and raise if it would blow past a safe fraction of RAM.

## Allocations that add up

- `event_list`: `3 * H * W * 7 * 8` = 168 bytes/pixel
- `status_values` + `status_struct` + `idle`: ~104 bytes/pixel
- `visibility_grid`: 8 bytes/pixel
- lexsort temporary: ~2x event_list during the sort

PR #1263, opened 2026-04-25
Add memory guards to dasymetric public API (#1261)

### LINKED ISSUE #1261: Add memory guards to dasymetric.pycnophylactic and disaggregate(method='limiting_variable')
**Describe the bug**

`pycnophylactic()` and `disaggregate(method='limiting_variable')` in `xrspatial/dasymetric.py` allocate full-shape working arrays without checking available memory first. `pycnophylactic()` additionally stores one boolean mask per zone, so memory grows with `N_zones * H * W`. Most other modules already check `_available_memory_bytes()` before this kind of allocation; dasymetric does not.

**Working memory footprint**

`_pycnophylactic_numpy` (lines 464-545) keeps the following alive at peak:

- four float64 full-shape arrays: `surface`, `smoothed`, `neighbour_sum`, `neighbour_count`
- two bool full-shape arrays: `valid`, `has_neighbours`
- a `zone_masks` dict holding one bool mask per zone in `values`

For 1000 zones on a 10000x10000 raster, iteration buffers come to about 3.4 GB (34 bytes per pixel) and the per-zone masks alone come to about 100 GB.

`_disaggregate_limiting_numpy` (lines 251-316) holds three to five full-shape masks plus the result and a `pixel_class` int32 array, which is roughly 4-5x the input size.

**Expected behavior**

Both public functions should call `_available_memory_bytes()` and raise `MemoryError` before the first allocation when projected working memory exceeds 0.5 * available. The check needs to account for zone count in pycnophylactic since masks scale with `len(values)`.

**Reproducer**

```python
import numpy as np
import xarray as xr

zones = xr.DataArray(np.random.randint(1, 10001, (10000, 10000)), dims=['y','x'])
values = {i: 100.0 for i in range(1, 10001)}
pycnophylactic(zones, values)  # tries to allocate ~1 TB of zone masks
```

**Additional context**

Found during the dasymetric security sweep. Same pattern as the fixes in #1231, #1229, #1240, #1257, #1256, #1236, and #1241.

A separate MEDIUM finding on dtype validation is being deferred to its own PR per the security-sweep policy of one fix per PR.

brendancol:
## Summary

- Adds `_available_memory_bytes()` and two budget helpers to `xrspatial/dasymetric.py` and calls them from `disaggregate()` and `pycnophylactic()` so both raise `MemoryError` before allocating when projected working memory exceeds 50% of available RAM
- The pycnophylactic budget scales with `len(values_dict)` because every zone keeps a full-shape bool mask alive at once

## Why

Pre-fix, `pycnophylactic()` on a 10000x10000 raster with 1000 zones tried to allocate ~100 GB of zone masks on top of ~3.4 GB of iteration buffers. `disaggregate(method='limiting_variable')` had no guard at all. Same pattern as #1231, #1229, #1240, #1236, and #1241.

Found during the dasymetric security sweep (`/sweep-security`).

## Test plan

- [x] All 53 existing dasymetric tests pass
- [x] 8 new tests in `TestMemoryGuard` cover both public functions across all three methods, the dask skip path, and the error-message contents
- [x] `.claude/sweep-security-state.json` updated with the dasymetric entry recording the fix and the deferred MEDIUM dtype-validation finding

Closes #1261


brendancol:

PR #1276, opened 2026-04-25
Add memory guard to emerging_hotspots public API (#1274)

### LINKED ISSUE #1274: emerging_hotspots: missing memory guard on numpy and cupy backends
**Describe the bug**

`emerging_hotspots()` in `xrspatial/emerging_hotspots.py` allocates full `(T, H, W)` working buffers on the numpy and cupy backends without checking available memory first. The public API only validates `ndim` and `shape[0] >= 2`, so a large `(time, y, x)` input drives the process into OOM with no feedback.

**Working memory footprint**

`_emerging_hotspots_numpy` (lines 343-373):

- `data = raster.data.astype(np.float32)`: full T*H*W float32 copy, 4 bytes per cell
- `gi_zscore = np.empty((n_times, ny, nx), dtype=np.float32)`: T*H*W * 4 bytes
- `gi_bin = np.empty((n_times, ny, nx), dtype=np.int8)`: T*H*W * 1 byte
- per-step convolved scratch (H*W float32) plus category/trend output buffers (about 13 bytes per H*W cell)

`_emerging_hotspots_cupy` (lines 380-416) has the same pattern on GPU plus a CPU mirror via `cupy.asnumpy(gi_zscore_gpu)`.

About 12 bytes per cell of the (T, H, W) cube. A `(100, 20000, 20000)` input projects to roughly 480 GB.

The dask backends process per-chunk via `map_blocks` and are not affected.

**Expected behavior**

`emerging_hotspots()` should call `_available_memory_bytes()` and raise `MemoryError` before the first allocation when projected working memory exceeds 0.5 * available. The check should run for the numpy and cupy backends only. Dask paths handle out-of-core processing themselves.

**Reproducer**

```python
import numpy as np
import xarray as xr
from xrspatial.emerging_hotspots import emerging_hotspots

kernel = np.ones((3, 3), dtype=np.float32)
data = np.zeros((100, 20000, 20000), dtype=np.float32)  # ~160 GB before any allocations
raster = xr.DataArray(data, dims=['time', 'y', 'x'])
emerging_hotspots(raster, kernel)  # tries to allocate another ~480 GB of working buffers
```

**Additional context**

Found during the emerging_hotspots security sweep. Same pattern as #1231, #1229, #1240, #1236, #1241, #1263, and #1264.

A separate MEDIUM finding (the public API does not call `_validate_raster()`, only checks ndim and shape, so non-numeric dtypes fail later with a confusing error) is being deferred per the one-fix-per-PR policy.

brendancol:
## Summary

- Adds `_available_memory_bytes()` and `_check_memory(n_times, ny, nx)` helpers to `xrspatial/emerging_hotspots.py` and calls them from the public API so the numpy and cupy backends raise `MemoryError` before allocating when the projected `(T, H, W)` cube would exceed 50% of available RAM
- Skips the guard for dask-backed inputs because their `map_blocks` and `map_overlap` paths process per-chunk and do not materialise the full cube
- 12 bytes per cube cell budget covers the float32 input copy, gi_zscore float32, and gi_bin int8, plus the small H*W temporaries (convolved scratch, category, trend_z, trend_p)

## Why

Pre-fix, the public API only validated `ndim` and `shape[0] >= 2`. A `(100, 20000, 20000)` input ran straight into about 480 GB of allocations on the numpy path with no warning. Same pattern as #1231, #1229, #1240, #1236, #1241, #1263, and #1264.

Found during the emerging_hotspots security sweep (`/sweep-security`).

## Test plan

- [x] All 30 existing emerging_hotspots tests pass
- [x] 5 new tests in `TestMemoryGuard` cover: numpy guard trips, numpy guard passes when budget is ample, cupy guard trips, dask guard is skipped, error message surfaces shape and projected GB
- [x] `.claude/sweep-security-state.json` updated with the emerging_hotspots entry recording the fix and a deferred MEDIUM finding (public API does not call `_validate_raster()`)

Closes #1274

### DIFF
--- a/.claude/sweep-security-state.json
+++ b/.claude/sweep-security-state.json
@@ -153,6 +153,13 @@
       "severity_max": null,
       "categories_found": [],
       "notes": "Clean. Small (271 LOC) module computing 3x3 second-derivative stencil. Cat 1: only single output buffer matching input shape (np.empty at line 37, cupy.empty at line 101) -- bounded by caller, per audit guidance not a finding. Cat 2: _cpu numba kernel uses range(1, rows-1)/range(1, cols-1) with simple (y, x) indices; no flat indexing or queue arrays; numba range loops produce int64. Cat 3: division by cellsize*cellsize on line 44 -- cellsize comes from get_dataarray_resolution() (raster property, not user-direct); cellsize=0 is unrealistic and would produce inf consistently across backends. NaN inputs propagate correctly through float arithmetic. Cat 4: _run_gpu (line 79-86) has full bounds guard via 'i + di <= out.shape[0] - 1 and j + dj <= out.shape[1] - 1' which guarantees i < shape[0] and j < shape[1] before the out[i, j] write; no shared memory; out is pre-filled with NaN at line 102 so threads outside the guard correctly leave NaN. Cat 5: no file I/O. Cat 6: curvature() calls _validate_raster at line 253; all four backend paths explicitly cast to float32 (lines 51, 62, 97, 112) so dtype is normalized before any computation; tests cover int32/int64/uint32/uint64/float32/float64 across numpy/cupy/dask+numpy/dask+cupy."
+    },
+    "emerging_hotspots": {
+      "last_inspected": "2026-04-25",
+      "issue": 1274,
+      "severity_max": "HIGH",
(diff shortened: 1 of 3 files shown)


brendancol:

PR #1277, opened 2026-04-25
Cap erode() parameters and run memory guard on every backend (#1275)

### LINKED ISSUE #1275: Cap iterations / radius / max_lifetime in erode() and run memory guard on every backend
**Describe the bug**

`erode()` in `xrspatial/erosion.py` accepts three user-controlled parameters with no upper bound: `iterations`, `params['radius']`, and `params['max_lifetime']`. Each one can drive an unbounded allocation or an unbounded CPU loop.

1. `iterations` (default 50000) sizes `rng.random((iterations, 2))` on the host at 16 bytes per particle, then copies the buffer to the GPU via `cupy.asarray(random_pos_np)`. At `iterations=10**12` that is roughly 16 TB on each side.

2. `params['radius']` (default 3) drives `_build_brush(radius)`, which walks `(2r+1)**2` cells and stores three arrays of the same length. At `radius=10**6` the offset and weight arrays alone are about 12 TB.

3. `params['max_lifetime']` (default 30) is the inner per-particle step loop in `_erode_cpu` and `_erode_gpu_kernel`. At `max_lifetime=10**12` with the default `iterations=50000` the JIT loop runs 5e16 times before returning.

The helper `_check_erosion_memory` exists at line 335 but only fires inside `_erode_dask_numpy` and `_erode_dask_cupy`. The numpy and cupy paths skip it, and the dask guard ignores the random-position buffer and the brush working set anyway.

Same shape as the diffuse bug fixed in #1268: a public iterative entry point with no caps on iteration count or per-step allocations.

**Expected behavior**

`erode()` should reject pathological parameter values with a `ValueError` or `MemoryError` before any large allocation. The fix from #1268 ports over directly: cap `iterations` via `_validate_scalar(max_val=...)`, cap `params['radius']` and `params['max_lifetime']` with explicit range checks, and run the memory guard on every backend (not just dask) with the random-position buffer and brush arrays included in the budget.

**Reproduction**

```python
import numpy as np
import xarray as xr
from xrspatial.erosion import erode

agg = xr.DataArray(np.zeros((64, 64), dtype=np.float32))

# All three accept the input without complaint and either OOM
# or pin a CPU core for hours:
erode(agg, iterations=10**10)               # ~160 GB rng buffer
erode(agg, params={'radius': 100_000})      # ~3.2 TB brush arrays
erode(agg, params={'max_lifetime': 10**9})  # CPU DoS
```

**Deferred follow-ups** (separate PRs, per the security-sweep one-fix-per-PR policy):

- Cat 3 HIGH: NaN input is not guarded in `_erode_cpu` or `_erode_gpu_kernel`. A NaN cell propagates through bilinear interpolation into `dir_x`/`dir_y`, NaN bounds checks fall through, and particles can deposit NaN via `cuda.atomic.add` into arbitrary cells of the heightmap.
- Cat 6 MEDIUM: `erode()` does not call `_validate_raster()`. Non-numeric or wrong-ndim input fails inside numba or cupy with a confusing error instead of a clean `ValueError`.

**Additional context**

Found by the security sweep on 2026-04-25. Erosion is on the safety-critical list.

brendancol:
Closes #1275.

## Summary

- Caps `iterations`, `params['radius']`, and `params['max_lifetime']` at the public `erode()` entry via `_validate_scalar(max_val=...)`. New constants: `_MAX_ITERATIONS=1e8`, `_MAX_RADIUS=1024`, `_MAX_LIFETIME=1e5`. Out-of-range values raise `ValueError` before any allocation runs.
- Rewrites `_check_erosion_memory` to include the `random_pos` buffer and brush arrays in its budget (the old version only counted the raster). Wires it into `_erode_numpy` and `_erode_cupy` so the numpy and cupy paths get the guard too, not just dask.
- Mirrors the diffuse fix in #1268.

## Test plan

- [x] All 15 existing erosion tests still pass
- [x] New: `iterations=0`, `iterations>cap`, `iterations=10**12`, non-int `iterations` raise `ValueError`/`TypeError` before any allocation
- [x] New: `radius=0`, `radius>cap` raise `ValueError`
- [x] New: `max_lifetime=0`, `max_lifetime>cap` raise `ValueError`
- [x] New: monkeypatched `_available_memory_bytes` makes the numpy and cupy paths raise `MemoryError` (dask is no longer the only guarded backend)
- [x] New: huge `iterations` on a tiny raster raises `MemoryError` mentioning `random_pos` (the buffer dominates)
- [x] New: defaults still work end-to-end

## Deferred follow-ups (per the security-sweep one-fix-per-PR policy)

- Cat 3 HIGH: NaN input is not guarded in `_erode_cpu` or `_erode_gpu_kernel`. A NaN cell propagates through bilinear interpolation into `dir_x`/`dir_y`, NaN bounds checks fall through, and particles can deposit NaN via `cuda.atomic.add` into arbitrary cells of the heightmap.
- Cat 6 MEDIUM: `erode()` does not call `_validate_raster()`. Non-numeric or wrong-ndim input fails inside numba/cupy with a confusing error.


brendancol:

Issue #1287, opened 2026-04-27
kde: unbounded width/height allocation in numpy and cupy backends


brendancol:
**Describe the bug**

`kde()` and `line_density()` take `width` and `height` as user-controlled ints with no upper bound. The eager numpy and cupy backends allocate `np.zeros((height, width), dtype=float64)` (or `cupy.zeros`) before any check runs.

Concrete sites in `xrspatial/kde.py`:

- `_run_kde_numpy` line 308: `np.zeros(shape, dtype=np.float64)`
- `_run_kde_cupy` line 314: `cupy.zeros(shape, dtype=cupy.float64)`
- `line_density` line 706: `np.zeros(shape, dtype=np.float64)`

A caller passing `width=1_000_000, height=1_000_000` asks for ~8 TB of float64 memory (or VRAM on the GPU path). On Linux that either OOMs the process or takes the kernel with it. On the GPU it surfaces as a raw cupy allocator error that says nothing about which parameter is too large.

Same pattern as the fixes in #1223 (rasterize), #1236 (bilateral), #1256 (morphology), #1241 (convolution), #1240 (contour), #1231 (bump), #1229 (viewshed), #1257 (glcm), #1261 (dasymetric), and #1262 (cost_distance).

The dask backends (`_run_kde_dask_numpy`, `_run_kde_dask_cupy`) build per-tile allocations lazily via `da.from_delayed`, so they are bounded by chunk size and do not need a separate guard.

**Expected behavior**

`kde()` and `line_density()` raise `MemoryError` before allocating, with a message that names `width=`/`height=` so the user knows which parameter to shrink.

**Reproduce**

```python
from xrspatial.kde import kde
kde([0.0], [0.0], bandwidth=1.0, width=1_000_000, height=1_000_000)
```

**Fix sketch**

1. Add a local `_available_memory_bytes()` helper (same pattern as convolution, morphology, bump: /proc/meminfo first, psutil fallback, 2 GB last resort).
2. Add `_check_grid_memory(rows, cols)` that raises `MemoryError` when `rows * cols * 8` exceeds 50% of available RAM.
3. Call the guard from `kde()` and `line_density()` once grid dims are resolved, before backend dispatch. Dask paths can skip it since chunk size already bounds them.

Issue #1284, opened 2026-04-27
focal apply()/focal_stats()/hotspots() accept unbounded user kernels


brendancol:
## What is the bug?

`focal.apply()`, `focal.focal_stats()`, and `focal.hotspots()` accept any 2-D NumPy array as the `kernel` argument. The validator (`convolution.custom_kernel()`) only checks `ndarray`-ness and odd row/column counts. There is no cap on kernel shape.

The kernel-size memory guard added in #1241 (`_check_kernel_memory`) only runs inside `circle_kernel` / `annulus_kernel`. A caller who builds the kernel directly with NumPy bypasses it, so a tiny raster with a giant kernel allocates many gigabytes:

```python
import numpy as np
import xarray as xr
from xrspatial.focal import apply

raster = xr.DataArray(np.zeros((10, 10), dtype=np.float32))
kernel = np.ones((50001, 50001), dtype=np.float32)  # ~10 GB on its own
apply(raster, kernel)
```

Downstream paths driven by `kernel.shape`:

- `_apply_numpy` (focal.py:385) allocates `kernel_values = np.zeros_like(kernel, dtype=data.dtype)` once per call.
- `_apply_numpy_boundary` (focal.py:404-412) and `_convolve_2d_numpy_boundary` (convolution.py:374-382) call `_pad_array(data, (kernel.shape[0]//2, kernel.shape[1]//2))` for any non-NaN boundary, which allocates `(H + 2*pad_h, W + 2*pad_w)`.
- The dask paths (`_apply_dask_numpy`, `_apply_dask_cupy`, `_focal_stats_dask_cupy`) set the `map_overlap` `depth` to `kernel.shape // 2` per chunk.
- `hotspots()` calls `convolve_2d(data, kernel/kernel.sum(), boundary)`, which inherits the same padded allocation.

Same shape as the bilateral fix in #1236, where `sigma_spatial` drove an unbounded radius. That one was patched by clamping the derived radius to `max(rows, cols)` of the raster before dispatch.

## How to reproduce

1. Build a small DataArray, for example `xr.DataArray(np.zeros((10, 10), dtype=np.float32))`.
2. Build a kernel with a half-extent larger than the raster, for example `np.ones((50001, 50001), dtype=np.float32)` (~10 GB on its own; padded allocation is much bigger).
[...]

Issue #1283, opened 2026-04-27
Memory guard for geodesic slope/aspect


brendancol:
The geodesic method in `slope()` and `aspect()` allocates a `(3, H, W)` float64 stacked array (data, lat, lon) per call with no memory guard. A large enough lat/lon-tagged raster passed to `slope(method='geodesic')` or `aspect(method='geodesic')` will OOM the process.

A 50000x50000 float64 raster makes a ~30 GB stacked buffer before padding or output is allocated. The numpy, cupy, and dask paths in `xrspatial/slope.py` and `xrspatial/aspect.py` all do this.

Other modules in this repo already guard against the same shape of problem: dasymetric (#1263), cost_distance (#1264), morphology (#1258), glcm_texture (#1259).

Proposed fix: add a `_check_geodesic_memory(rows, cols, func_name)` helper in `xrspatial/geodesic.py`, modeled on `_check_kernel_memory` in `xrspatial/morphology.py`. Raise `MemoryError` when the stacked float64 buffer plus the float32 output plus a padding allowance exceeds 50% of available RAM. Call it from the geodesic branch of `slope()` and `aspect()` before dispatch.

Severity: HIGH. DoS vector, no existing guard.

Issue #1267, opened 2026-04-25
diffusion: unbounded allocations and unbounded steps loop


brendancol:
## Summary

`xrspatial.diffusion.diffuse` has no memory guard on its core allocations, and `steps` has no upper bound. Large rasters or large step counts can OOM the host or pin a CPU.

## What goes wrong

1. `diffuse()` always materializes a full float64 alpha raster via `np.full(agg.shape, alpha_scalar, dtype=np.float64)` (diffusion.py line 305), even when the input is a dask DataArray with scalar diffusivity. A 100000x100000 input forces an 80 GB numpy allocation in the public API before any backend dispatch.

2. `_diffuse_step_numpy` allocates `np.empty((rows, cols))` every step (line 57) with no memory check.

3. `_diffuse_cupy` allocates `cp.empty_like(u)` every step (line 123) with no GPU memory check.

4. `_validate_scalar(steps, ..., min_val=1)` only sets a lower bound. `steps=10**12` is accepted and would loop forever.

## Fix

- Add a peak-memory guard like the cost_distance / morphology one. Budget ~32 bytes per pixel (u, out, alpha, padded copy as float64) and raise MemoryError before the first allocation in the eager numpy / cupy paths.
- For the dask + scalar case, don't materialize a full numpy alpha array. Pass the scalar through.
- Cap `steps` so a single call can't pin a CPU forever.

## Acceptance criteria

- Oversize raster raises MemoryError before allocating.
- Scalar diffusivity on a dask raster doesn't materialize a full numpy alpha array.
- Out-of-range `steps` raises a clean ValueError.
- Existing tests still pass.

PR #1297, opened 2026-04-28
Guard resample() against unbounded output allocations (#1295)

### LINKED ISSUE #1295: Guard resample() against unbounded scale_factor / target_resolution
**Describe the bug**

`resample()` does not bound the output dimensions derived from user-supplied `scale_factor` or `target_resolution`. The output shape is computed as:

```python
out_h, out_w = max(1, round(in_h * scale_y)), max(1, round(in_w * scale_x))
```

A caller passing `scale_factor=1e6` against a small raster (or `target_resolution=1e-9` against a meter-scale raster) produces a huge output shape that is then allocated inside `_run_numpy`, `_run_cupy`, the `_AGG_FUNCS` numba kernels, and `_nan_aware_interp_np` via `np.empty` / `cupy.zeros` / `scipy.ndimage.map_coordinates` with no memory check. The process OOMs (or hangs) before any work actually begins.

**Expected behavior**

When the projected output buffer would exceed available memory, the function should raise `MemoryError` cleanly before any large allocation is attempted, mirroring the guards added in #1284 (focal), #1287 (kde / line_density), #1283 (geodesic slope/aspect), #1262 (cost_distance), and #1267 (diffuse).

**Reproducer**

```python
import numpy as np
import xarray as xr

agg = xr.DataArray(
    np.zeros((100, 100), dtype=np.float32),
    dims=('y', 'x'),
    coords={'y': np.arange(100), 'x': np.arange(100)},
)

# Triggers a 1e8 x 1e8 float64 allocation (~80 PB)
resample(agg, scale_factor=1e6, method='nearest')
```

**Fix outline**

Add `_available_memory_bytes()` and `_check_resample_memory(out_h, out_w)` in `xrspatial/resample.py`. Budget about 12 bytes per output cell (float64 working buffer plus float32 output) and raise `MemoryError` when the projected memory exceeds 50 percent of available RAM. Call the guard from `resample()` after the output shape is computed but before backend dispatch. Add a GPU-side equivalent for the cupy-eager branch using `cupy.cuda.runtime.memGetInfo`. Skip the guard for pure dask paths since per-chunk allocation is bounded by chunk size; that can be addressed separately if needed.

**Additional context**

Found during the resample security sweep audit (Cat 1: Unbounded Allocation / DoS).

brendancol:
## Summary

Closes #1295.

`resample()` did not bound the output dimensions derived from user-supplied `scale_factor` or `target_resolution`. The eager numpy and cupy backends allocated the full `(out_h, out_w)` working buffer up front, so `scale_factor=1e9` on a 4x4 raster requested ~190 EB before any guard ran.

This adds `_available_memory_bytes()` / `_available_gpu_memory_bytes()` and `_check_resample_memory` / `_check_resample_gpu_memory` helpers (12 B/cell budget covering the float64 working buffer, float32 output, and the scipy/cupyx `map_coordinates` temporary). The guard runs in `resample()` after the output shape is computed but before backend dispatch. Eager numpy and cupy paths run the guard; dask paths skip it because per-chunk allocations are already bounded by chunk size.

Same pattern as #1287 (kde / line_density), #1284 (focal), #1283 (geodesic), #1262 (cost_distance), #1267 (diffuse).

## Test plan

- [x] `pytest xrspatial/tests/test_resample.py` passes (62 tests, 6 new in `TestMemoryGuard`)
- [x] `scale_factor=1e9` raises `MemoryError`
- [x] `target_resolution=1e-9` raises `MemoryError`
- [x] aggregate-method `ValueError` still wins over `MemoryError`
- [x] normal inputs (`scale_factor=2.0`) still work
- [x] error message names `scale_factor` and `target_resolution`
- [x] dask path skips the guard


brendancol:

Issue #1299, opened 2026-04-28
sky_view_factor(): numpy and cupy backends have no memory guard


brendancol:
## Description

`sky_view_factor()` on the numpy and cupy backends has no memory guard. `_run_numpy` (line 146) calls `data.astype(np.float64)` and `_svf_cpu` (line 60) allocates `np.empty((rows, cols), dtype=np.float64)` sized directly from the input dimensions. `_run_cupy` (line 151) does the same with `data.astype(cupy.float64)` and `cupy.full(data.shape, cupy.nan, dtype=cupy.float64)` (line 152).

That is ~16 bytes/pixel of working memory before any check runs. A 50000x50000 numpy raster asks for ~40 GB of host memory before anything errors out.

The dask paths are bounded per-chunk by `map_overlap`, matching the resample / kde precedent.

Same asymmetric-guard pattern already fixed in sieve #1296, kde #1287, resample #1295, focal #1284, geodesic #1283, mahalanobis #1288, true_color #1291, diffuse #1267, erode #1275, emerging_hotspots #1274, and dasymetric #1261.

## Expected behavior

`sky_view_factor()` raises `MemoryError` with a clear message on the eager numpy and cupy backends when the projected working set exceeds available memory. Dask paths skip the guard.

## Proposed fix

Add `_available_memory_bytes()`, `_check_memory(rows, cols)`, and `_check_gpu_memory(rows, cols)` helpers (16 bytes/pixel, 50% threshold). Call them from `_run_numpy` and `_run_cupy` before the float64 cast and output allocation.

Issue #1262, opened 2026-04-25
cost_distance: missing memory guard on CuPy backend


brendancol:
**Describe the bug**

PR #1253 added `_check_memory(height, width)` to the numpy path of `cost_distance` so oversize rasters raise `MemoryError` before the Dijkstra kernel allocates ~37 bytes/pixel of working memory. The cupy backend runs the same allocation pattern with no such guard.

`_cost_distance_cupy` (xrspatial/cost_distance.py:349-409) allocates:

- `cp.full((height, width), cp.inf, dtype=cp.float64)` at line 367 (8 bytes/pixel)
- `source_mask` and `passable` boolean masks (~2 bytes/pixel)
- `cp.where(...).astype(cp.float32)` output and intermediates (~12 bytes/pixel)

A 100000x100000 cupy raster requests ~80 GB just for `dist`. That exceeds GPU memory on every consumer card and most workstation cards. What you actually see is a cupy allocator error that does not mention `max_cost=` or dask. The numpy path raises a clean `MemoryError` with both hints; cupy should do the same.

The dask+cupy `map_overlap` path calls `_cost_distance_cupy` per chunk, so a guard inside `_cost_distance_cupy` also covers per-chunk allocations there. The dask+cupy unbounded fallback converts to dask+numpy and picks up the existing guard.

**Expected behavior**

`cost_distance` on an oversize CuPy raster raises `MemoryError` with a message pointing to `max_cost=` and dask backends, like the numpy path already does. No raw CUDA OOM, no opaque cupy allocator error.

**Reproduce**

```python
import cupy as cp
import xarray as xr
import numpy as np
from xrspatial import cost_distance

H, W = 100000, 100000  # ~80 GB float64
src = cp.zeros((H, W), dtype=cp.float64)
src[0, 0] = 1.0
fric = cp.ones((H, W), dtype=cp.float64)

raster = xr.DataArray(src, dims=['y', 'x'])
friction = xr.DataArray(fric, dims=['y', 'x'])
raster['y'] = np.arange(H)
raster['x'] = np.arange(W)
friction['y'] = np.arange(H)
friction['x'] = np.arange(W)

cost_distance(raster, friction)  # GPU OOM, no clean error
```

**Additional context**

[...]

PR #1310, opened 2026-04-29
Guard gpu_rtx hillshade and viewshed against unbounded GPU allocations (#1308)

### LINKED ISSUE #1308: Guard gpu_rtx hillshade and viewshed against unbounded GPU allocations
## Description

`xrspatial.gpu_rtx` has two GPU entry points (`hillshade_rtx` and `viewshed_gpu`) that allocate cupy device buffers sized by the input raster with no memory check up front. On a large raster the cupy allocator eventually fails, but it does so with an internal error that doesn't name the input shape, so the user has no clean way to know what to turn down.

Both entry points call `create_triangulation` in `gpu_rtx/mesh_utils.py`, which allocates `verts` (12 B/pixel) and `triangles` (about 24 B/pixel). Each entry point then allocates more per-pixel buffers on top:

- `hillshade_rtx`: `d_rays` 32, `d_hits` 16, `d_aux` 12, `d_output` 4. Total with mesh: about 100 B/pixel.
- `viewshed_gpu`: `d_rays` 32, `d_hits` 16, `d_visgrid` 4, `d_vsrays` 32. Total with mesh: about 120 B/pixel.

A 30000x30000 raster asks for 90-108 GB of VRAM before any check runs.

## Reproducer

```python
import cupy
import numpy as np
import xarray as xr
from xrspatial import hillshade

H = W = 30000
data = cupy.zeros((H, W), dtype=cupy.float32)
agg = xr.DataArray(data, dims=['y', 'x'])
agg['x'] = np.arange(W)
agg['y'] = np.arange(H)[::-1]

hillshade(agg, shadows=True)  # OOMs on the GPU with an opaque allocator error
```

## Proposed Fix

Add `_available_gpu_memory_bytes()` and `_check_gpu_memory(rows, cols)` helpers in the `gpu_rtx` subpackage, mirroring the cost_distance (#1262) and sky_view_factor (#1299) pattern: `cupy.cuda.runtime.memGetInfo()` wrapped in try/except (return 0 when unavailable), raise `MemoryError` when the estimate exceeds 50% of free VRAM.

Wire the guard into `hillshade_rtx` and `viewshed_gpu` at function entry. Use 120 B/pixel as the budget so the worst-case `viewshed_gpu` path covers both.

## Severity

HIGH, Cat 1 (Unbounded Allocation / DoS) per the security sweep. Reachable from the public `hillshade()` and `viewshed()` APIs.

## Out of scope

Noting these so they don't get bundled into this PR (one security fix per PR):

- Cat 2 MEDIUM: int32 triangle indices in `mesh_utils` overflow above ~46341x46341. The new memory guard rejects rasters that large first, so this is more of a documentation item.
- Cat 3 MEDIUM: `mesh_utils.py:19` `scale = maxDim / maxH` divides by zero on an all-zero raster.
- Cat 5 LOW: `mesh_utils.write()` opens a user-supplied path without canonicalization, but the only call site sits behind `if False:` in `create_triangulation`.
- Cat 6 MEDIUM: `hillshade_rtx` and `viewshed_gpu` don't call `_validate_raster` directly. The parent `hillshade()`/`viewshed()` do, so input validation already runs.

brendancol:
## Summary

- `hillshade_rtx` and `viewshed_gpu` allocated cupy device buffers sized by raster shape with no memory check. A 30000x30000 raster asked for 90-108 GB of VRAM (100-120 B/pixel) before cupy surfaced an opaque allocator error.
- Adds `gpu_rtx/_memory.py` with `_available_gpu_memory_bytes()` and `_check_gpu_memory(func_name, h, w)`, same shape as #1262 / #1299. 120 B/pixel budget covers both entry points (worst case is `viewshed_gpu`). Wired in at the top of `hillshade_rtx` and `viewshed_gpu`.
- 9 new tests in `xrspatial/tests/test_gpu_rtx_memory.py`: 5 helper-unit tests that don't need RTX, 4 end-to-end through `hillshade()` / `viewshed()` gated on `has_rtx()`.

Closes #1308.

## Test plan

- [x] `pytest xrspatial/tests/test_gpu_rtx_memory.py -v` (9 passed)
- [x] `pytest xrspatial/tests/test_hillshade.py xrspatial/tests/test_viewshed.py` (81 passed, no regressions)
- [x] Helper unit-tested with patched `_available_gpu_memory_bytes`
- [x] `MemoryError` raised end-to-end from `hillshade(shadows=True)` and `viewshed()` when free VRAM is patched to a tiny value
- [x] Small raster passes the guard at real free-VRAM levels


brendancol:

brendancol:

PR #1312, opened 2026-04-29
Reject NaN/Inf weights in mcda combine and ahp_weights (#1311)

### LINKED ISSUE #1311: wlc/wpm/owa/ahp_weights silently propagate NaN weights to all-NaN raster
## Description

`_validate_weights` in `xrspatial.mcda.combine` (and the `owa()` order-weights check) only validates that weights sum to about 1.0:

```python
total = sum(weights.values())
if abs(total - 1.0) > 0.01:
    raise ValueError(...)
```

If any weight is NaN, the sum is NaN, `abs(NaN - 1.0) > 0.01` is False, and the check passes. `wlc()`, `wpm()`, and `owa()` then propagate the NaN to every pixel and return an all-NaN raster, no exception raised.

`ahp_weights` in `xrspatial.mcda.weights` has the same problem. Its guard is `if val <= 0:`, which NaN slips past. The matrix fills with NaN, the eigenvector solve produces NaN, and the function returns NaN weights.

Cat 3 (NaN as logic error) finding from the mcda security sweep. HIGH severity because the output looks structurally valid: same shape, no warning, no exception. Anything downstream (`constrain`, plotting, export) will treat it as real data.

## Reproduction

```python
import xarray as xr
import numpy as np
from xrspatial.mcda.combine import wlc

data = xr.Dataset({
    'a': xr.DataArray(np.array([[0.5, 0.5], [0.5, 0.5]])),
    'b': xr.DataArray(np.array([[0.5, 0.5], [0.5, 0.5]])),
})
result = wlc(data, {'a': 0.5, 'b': float('nan')})
print(result.values)
# Expected: ValueError naming the offending criterion
# Actual:   [[nan nan] [nan nan]]
```

## Expected behaviour

`_validate_weights` and the `owa()` order-weights check should raise `ValueError` and name the offending criteria (or positions) when a weight is NaN or infinite. `ahp_weights` should reject NaN/Inf comparison values the same way it rejects non-positive ones.

## Affected files

- `xrspatial/mcda/combine.py` -- `_validate_weights` (~lines 22-39), `owa` order-weights check (~lines 154-158)
- `xrspatial/mcda/weights.py` -- `ahp_weights` value guard (~line 94)

## Fix sketch

Sweep the weight dict once, raise `ValueError` on any NaN/Inf, then run the sum-to-1 check. Apply the same sweep to `order_weights` in `owa`. In `ahp_weights`, extend the `val <= 0` check to also reject `np.isnan(val)` and `np.isinf(val)`.

One fix per PR, per the existing security-sweep convention.

brendancol:
## Summary

Closes #1311.

`_validate_weights` and `owa()`'s order-weights check only verified that weights summed to about 1.0. Since `abs(NaN - 1.0) > 0.01` is False, a NaN weight slipped past the check, hit every pixel, and the function happily returned an all-NaN raster with no exception. `ahp_weights` has the same problem one level down: its `if val <= 0:` guard lets NaN through into the comparison matrix, and the eigenvector solve produces NaN weights from there.

## Changes

- `xrspatial/mcda/combine.py`: added `_check_finite_weights` and called it from `_validate_weights`. Same idea applied per-position to `order_weights` in `owa()`. Error messages name the offending criterion or list index.
- `xrspatial/mcda/weights.py`: `ahp_weights` rejects NaN/Inf comparison values before the existing `val <= 0` check.
- `xrspatial/tests/test_mcda.py`: 18 new tests for NaN/Inf in `wlc`, `wpm`, `owa` (criterion weights and order weights), and `ahp_weights`. A few positive-path tests too, to make sure the new check doesn't break valid inputs.

All 166 mcda tests pass.

## Test plan

- [x] `pytest xrspatial/tests/test_mcda.py::TestNonFiniteWeights1311 -v` -- 18/18 pass
- [x] `pytest xrspatial/tests/test_mcda.py` -- 166/166 pass
- [x] The reproduction from #1311 now raises `ValueError` naming the bad criterion

## Notes

One fix per PR, per the security-sweep convention. The mcda audit also turned up some MEDIUM items that aren't fixed here: `sensitivity._monte_carlo` eagerly computing the full dask Dataset, `combine.owa` stacking criteria via `xr.concat` without a size guard, `n_samples=0` divides by zero, and `wpm` will accept zero-base/negative-weight combinations. Those are recorded in `.claude/sweep-security-state.csv` for follow-up.


brendancol:

PR #1380, opened 2026-04-30
Guard _idw knearest against unbounded grid x k allocation (#1377)

### LINKED ISSUE #1377: _idw knearest: no memory guard on grid x k allocation
**Describe the bug**

The k-nearest path in `idw()` materialises two `(grid_pixels, k)` arrays via `scipy.spatial.cKDTree.query` with no memory guard. At a 50000 x 50000 template with `k=12`, that is 2.5e9 pixels x 12 neighbours x 16 bytes (one float64 distance + one int64 index per cell) which works out to about 480 GB, allocated eagerly before any IDW arithmetic runs.

Site in `xrspatial/interpolate/_idw.py`:

- `_idw_knearest_numpy` line 87: `dists, indices = tree.query(query_pts, k=k)`. Returns a `(grid_pixels, k)` float64 distance array and an int64 index array. The follow-up `weights`, `z_vals`, and `wz` arrays are the same shape. Peak usage is roughly `grid_pixels * k * 16` bytes for the cKDTree return alone.

A caller with a large template or large `k` hits this silently. The process either OOMs or scipy raises an opaque allocator error that does not name the template grid size or `k`.

Same pattern as the guards added in #1287 (kde), #1295 (resample), #1296 (sieve), #1309 (kriging), and #1374 (spline TPS).

The `_idw_dask_numpy` path calls `_idw_knearest_numpy` per chunk via `map_blocks`, so chunk size bounds the per-chunk allocation. The numpy entry point runs on the full template and needs a guard regardless of backend.

**Expected behaviour**

`idw(..., k=k)` raises `MemoryError` before the cKDTree query, with a message that names the template grid size and `k` so the user knows what to reduce.

**Reproduce**

```python
import numpy as np
import xarray as xr
from xrspatial.interpolate import idw

n = 1000
x = np.random.rand(n)
y = np.random.rand(n)
z = np.random.rand(n)
ny, nx = 50_000, 50_000
template = xr.DataArray(
    np.zeros((ny, nx)), dims=['y', 'x'],
    coords={'y': np.arange(ny, dtype=np.float64),
            'x': np.arange(nx, dtype=np.float64)},
)
idw(x, y, z, template, k=12)
```

**Fix sketch**

1. Estimate `grid_pixels * k * 16` bytes (float64 dists + int64 indices from cKDTree.query) at the top of `idw()` when `k` is set.
2. Compare against `0.8 * _available_memory_bytes()` using the helper from `xrspatial.zonal`, same pattern as `_check_kriging_memory`.
3. Raise `MemoryError` naming `grid_pixels` and `k` when the estimate exceeds the threshold.

brendancol:
Closes #1377.

## Summary

- Add `_check_idw_memory(grid_pixels, k)` in `xrspatial/interpolate/_idw.py`. Estimates `grid_pixels * k * 16` bytes (the float64 distance array plus the int64 index array returned by `scipy.spatial.cKDTree.query`) and raises `MemoryError` if it exceeds `0.8 * _available_memory_bytes()`.
- Wire the guard into the public `idw()` entrypoint when `k` is set, gated to non-dask templates. The dask+numpy backend dispatches `_idw_knearest_numpy` per chunk via `map_blocks`, so chunk size already bounds the per-chunk allocation. GPU backends reject `k` early.
- Three tests covering oversize-raises, normal-succeeds, and the error message naming `grid_pixels` and `k`.

## Why

Without the guard, a 50000 x 50000 template with `k=12` allocates about 480 GB inside scipy with no message that identifies the template grid or `k`. Same shape as the kriging guard from #1309 and the spline guard from #1374.

## Test plan

- [x] `pytest xrspatial/tests/test_interpolation.py` passes (37 tests, 3 new)
- [x] New tests exercise the oversize and normal paths with monkeypatched `_available_memory_bytes`
- [x] Existing IDW, kriging, spline tests still pass

### DIFF
--- a/xrspatial/interpolate/_idw.py
+++ b/xrspatial/interpolate/_idw.py
@@ -232,6 +232,37 @@ def _idw_dask_cupy(x_pts, y_pts, z_pts, x_grid, y_grid,
 # Public API
 # ---------------------------------------------------------------------------
 
+def _check_idw_memory(grid_pixels, k):
+    """Raise MemoryError if the k-nearest IDW path would exceed memory.
+
+    ``scipy.spatial.cKDTree.query(query_pts, k=k)`` returns two arrays of
+    shape ``(grid_pixels, k)``: a float64 distance array (8 B/cell) and an
+    int64 index array (8 B/cell), for a peak of ``grid_pixels * k * 16``
+    bytes before any IDW arithmetic runs.  Downstream ``weights``,
+    ``z_vals``, and ``wz`` arrays share the same shape, so this estimate
+    bounds the dominant allocation.
+    """
+    g = int(grid_pixels)
+    kk = int(k)
+    estimate = g * kk * 16
+
+    try:
+        from xrspatial.zonal import _available_memory_bytes
+        avail = _available_memory_bytes()
+    except ImportError:
+        avail = 2 * 1024 ** 3
+
+    if estimate > 0.8 * avail:
+        raise MemoryError(
+            f"idw() needs ~{estimate / 1e9:.1f} GB to allocate the "
+            f"(grid_pixels, k) distance and index arrays of shape "
+            f"({g}, {kk}) for the k-nearest cKDTree query, but only "
+            f"~{avail / 1e9:.1f} GB is available.  Reduce the template "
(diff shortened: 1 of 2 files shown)


brendancol:

PR #1414, opened 2026-05-02
Record accuracy sweep results for corridor

brendancol:
## Summary

- Audited `xrspatial/corridor.py` against the 5 accuracy categories (precision loss, NaN/Inf propagation, off-by-one, projection corrections, backend inconsistency).
- No CRIT/HIGH issues. One LOW: corridor inherits float32 from `cost_distance` so `normalized = corridor - corridor_min` can lose precision near the min when costs accumulate to large values. The arithmetic in corridor itself does not introduce extra loss.
- NaN handling is correct: `corridor.min()` uses `skipna=True`, the result is checked with `np.isfinite` before normalization, and unreachable corridors return all-NaN.
- All four backends route through pure xarray arithmetic. The threshold path dispatches to `dask.where`, `cupy.where`, or `np.where` via a try/except pattern.

## Test plan

- [x] CSV updated with module, date, severity, categories, notes


brendancol:

PR #1434, opened 2026-05-03
reproject: validate grid/bounds/precision params

### LINKED ISSUE #1433: reproject: grid/bounds/precision parameters not validated
## Description

`reproject()` accepts several user-controlled grid parameters that are not range-checked:

| Parameter | Bad input today |
| -- | -- |
| `resolution` | `0` -> raw `ZeroDivisionError` in `_grid.py:163-200`; negative -> silently produces backwards coords |
| `width` / `height` | `0` -> `ZeroDivisionError`; negative -> `max(1, int(round(neg)))` keeps a negative `res_x`, output is silently wrong |
| `bounds` | `(10, 10, 0, 0)` (right < left, top < bottom) -> negative `res_x`, `res_y` -> output coords descending and downstream resampler emits all-nodata silently |
| `transform_precision` | `-1` -> `np.linspace(0, end, 0)` empty -> `IndexError` from `_bilinear_interp_2ch` indexing `gr[-1]` |

## Expected behavior

Each parameter is range-checked at the top of `reproject()`. Bad input raises a clean `ValueError` with a message naming the offending parameter.

## Proposed fix

- `resolution`: must be a positive finite scalar or 2-tuple of same.
- `width` / `height`: must be a positive int.
- `bounds`: must be a 4-tuple of finite floats with `right > left` and `top > bottom`.
- `transform_precision`: must be a non-negative int.

`merge()` consumes the same `resolution` / `bounds` parameters so the same checks apply there.

brendancol:
Closes #1433.

## Summary

- New `_validate_grid_params()` helper in `xrspatial/reproject/_grid.py` that range-checks `resolution`, `bounds`, `width`, `height`, `transform_precision`.
- Wired into `reproject()` and `merge()` (the latter only validates the params that reach it).
- Bad inputs now raise `ValueError` from the public API instead of `ZeroDivisionError` / `IndexError` deep in the resampler or silently producing wrong-signed coords.

## Test plan

- [x] 24 new tests in `TestValidateGridParams` / `TestValidateMergeGridParams`.
- [x] `pytest xrspatial/tests/test_reproject.py` — 120 passed (96 existing + 24 new).
- [x] CI run on full suite.


brendancol:
@copilot can you fix the merge conflict?

PR #1446, opened 2026-05-04
Preserve input attrs through reproject() and merge()

### LINKED ISSUE #1445: reproject and merge silently drop input DataArray attrs
## Problem

Both `reproject()` and `merge()` in `xrspatial/reproject/__init__.py` build their output `attrs` from scratch instead of carrying the input attrs forward.

`reproject()` at lines ~672-697:

```python
out_attrs = {
    'crs': tgt_wkt,
    'nodata': nd,
}
```

`merge()` at lines ~1473-1477:

```python
attrs={
    'crs': tgt_wkt,
    'nodata': nd,
},
```

Anything attached to the input — `units`, `long_name`, `description`, `scale_factor`, `add_offset`, `_FillValue`, custom GeoTIFF tags, history, provenance — is dropped on the floor.

## Impact

- Downstream code that reads `units` (e.g. unit-aware plots, validators) sees nothing.
- Packed datasets relying on `scale_factor` / `add_offset` lose the unpacking metadata. Any subsequent math is wrong.
- `_FillValue` / `nodata` overrides set by users disappear.
- Provenance text (`history`, `description`, `long_name`) is gone.

This is the same bug class as #1407 (`sky_view_factor` cellsize loss). Operations that transform a raster shouldn't strip every attr the user attached.

## Fix

Merge input attrs first, then override the ones the operation legitimately changes:

```python
out_attrs = {**raster.attrs}
out_attrs['crs'] = tgt_wkt
out_attrs['nodata'] = nd
```

A few input attrs become stale after reprojection and should be popped explicitly:

- `transform`: the affine transform changed; the old one is wrong.
- `crs_wkt`: we re-emit canonical `crs`, drop the duplicate.
- `res`: refers to the old grid resolution.

For `merge()`, use `rasters[0].attrs` as the base (consistent with the default `strategy='first'`), and fall back `name` to `rasters[0].name` before `'merged'`.

## Tests

`xrspatial/tests/test_reproject.py` doesn't currently assert any attr survives the round-trip — only `'crs' in result.attrs`. Add coverage so the next regression in this area gets caught.

brendancol:
## Summary

- `reproject()` and `merge()` now merge input `attrs` forward instead of building output attrs from scratch, so `units`, `long_name`, `scale_factor`, `add_offset`, `_FillValue`, and similar metadata survive the transform.
- Stale grid attrs (`transform`, `res`, `crs_wkt`) are popped before the new `crs`/`nodata` are written, since they describe the old grid and would mislead downstream consumers.
- `merge()` falls back its output `name` to `rasters[0].name` before `'merged'`, matching the pattern `reproject()` already uses.

## Test plan

- [x] New `TestMetadataPreservation` covers `units`, `scale_factor`/`add_offset`, `long_name`, stale `transform`, stale `res`, stale `crs_wkt`, and the `crs` override for `reproject()`.
- [x] Same class covers first-raster attr preservation, stale-`transform` drop, and the `name` fallback for `merge()`.
- [x] `pytest xrspatial/tests/test_reproject.py -x` -> 146 passed.

Closes #1445

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -669,10 +669,16 @@ def reproject(
             )
 
     ydim, xdim = _find_spatial_dims(raster)
-    out_attrs = {
-        'crs': tgt_wkt,
-        'nodata': nd,
-    }
+    # Carry input attrs forward so units, long_name, scale_factor, etc.
+    # survive the transform. Pop attrs that are stale after reprojection:
+    # the affine `transform` and grid `res` describe the old grid, and
+    # `crs_wkt` would duplicate (or contradict) the canonical `crs` we re-emit.
+    out_attrs = {**raster.attrs}
+    out_attrs.pop('transform', None)
+    out_attrs.pop('crs_wkt', None)
+    out_attrs.pop('res', None)
+    out_attrs['crs'] = tgt_wkt
+    out_attrs['nodata'] = nd
     if tgt_vertical_crs is not None:
         out_attrs['vertical_crs'] = tgt_vertical_crs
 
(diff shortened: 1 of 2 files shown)


brendancol:

PR #1449, opened 2026-05-04
Honor per-raster nodata sentinels in merge()

### LINKED ISSUE #1448: merge() ignores per-raster nodata sentinels, leaks invalid pixels into mosaic
## Summary

`xrspatial.reproject.merge()` only detects the nodata sentinel of the first raster in the input list. That single sentinel is then used for every other raster during reprojection and merging. When inputs use different sentinels, the non-first rasters' invalid pixels are not recognized as nodata and silently flow into the mosaic as real data.

## Where

`xrspatial/reproject/__init__.py:1385`:

```python
nd = nodata if nodata is not None else _detect_nodata(rasters[0], nodata)
```

`nd` is then passed unchanged into `_reproject_chunk_numpy` for every raster in both `_merge_inmemory` and `_merge_block_adapter`.

## Concrete example

- Raster A has NaN nodata, valid data = 10.
- Raster B has -9999 nodata, valid data = 20, with -9999 sentinels in part of its footprint.

Calling `merge([a, b], strategy='mean')` returns -9999 values in B's nodata regions because `_reproject_chunk_numpy` was told the sentinel is NaN, so it never converted -9999 to NaN before resampling. Worse, with `strategy='mean'` and a 0-sentinel input, sentinel zeros are folded into averages, producing wrong numerical results with no warning.

The user-supplied `nodata` argument is the desired *output* sentinel and is unrelated to the per-raster *input* sentinels, so checking `nodata` is not None does not save us either.

## Fix

Detect each raster's own nodata sentinel independently of the user-supplied output `nodata`. Reproject each raster with its own input sentinel, canonicalize to NaN before the strategy merge, then convert NaN back to the user-requested output nodata at the end. The strategy merge already handles NaN canonical via its `use_nan` branch.

This affects both the in-memory and dask paths.

brendancol:
Closes #1448.

## Summary

`merge()` only detected the first raster's nodata sentinel and reused it for every other input, so rasters with a different sentinel had their invalid pixels passed through reprojection as real data. The mosaic looked fine but contained wrong numerical values.

## Fix

- Detect each raster's own input nodata sentinel in the `raster_infos` loop, independent of the user-supplied output `nodata`.
- Reproject each raster using its own input sentinel, then canonicalize to NaN before the strategy merge so all rasters share one missing-data marker.
- After the strategy merge, convert NaN back to the user-requested output `nodata` if that is not NaN.

Applied in both the in-memory path (`_merge_inmemory`) and the dask path (`_merge_block_adapter` / `_merge_dask`).

## Test plan

- [x] `pytest xrspatial/tests/test_reproject.py` (139 passed)
- [x] New `TestMergeMixedNodata` class covers:
  - NaN + -9999 sentinel mix, default output nodata
  - NaN + 0 sentinel mix with `mean` strategy (verifies zeros are not folded into averages)
  - Mixed sentinels with explicit user-requested `nodata=-9999` output

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -1381,7 +1381,7 @@ def merge(
             "Could not detect target CRS. Pass target_crs explicitly."
         )
 
-    # Detect nodata
+    # Detect output nodata (the sentinel the user asked for)
     nd = nodata if nodata is not None else _detect_nodata(rasters[0], nodata)
 
     # Gather source info for each raster
@@ -1396,6 +1396,10 @@ def merge(
         sb = _source_bounds(r)
         ss = (r.sizes[r.dims[-2]], r.sizes[r.dims[-1]])
         yd = _is_y_descending(r)
+        # Per-raster input nodata sentinel. Detected independently of the
+        # user-supplied output nodata so that mixed-sentinel inputs are
+        # canonicalized correctly during merge.
+        r_nd = _detect_nodata(r, None)
         raster_infos.append({
             'raster': r,
             'src_crs': src_crs,
@@ -1403,6 +1407,7 @@ def merge(
             'src_shape': ss,
             'y_desc': yd,
             'src_wkt': src_crs.to_wkt(),
+            'raster_nodata': r_nd,
         })
 
     # Compute unified output grid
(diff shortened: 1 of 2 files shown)


brendancol:

PR #1453, opened 2026-05-04
Guard _apply_vertical_shift against non-finite coords and add vertical CRS tests

### LINKED ISSUE #1451: _apply_vertical_shift can hang on non-finite coords; vertical CRS path is untested
Two related problems in `xrspatial/reproject` around the vertical-datum-shift code path.

## Hang on non-finite coords

`_apply_vertical_shift` in `xrspatial/reproject/__init__.py` (around lines 702-791) hands output-grid coordinates to `_interp_geoid_2d` directly. When the target CRS is projected, output coords are first inverse-projected to lon/lat at lines 773-776. Some projections (polar stereographic near the pole, the antimeridian on certain CRSes) emit `inf` or NaN from `transformer.transform`. Those values flow into `_interp_geoid_point` in `xrspatial/reproject/_vertical.py` (lines 127-130):

```python
while lon_w < -180.0:
    lon_w += 360.0
while lon_w >= 180.0:
    lon_w -= 360.0
```

On `inf`, neither loop terminates and the worker hangs.

The public `geoid_height` API guards against non-finite input at `_vertical.py:205-216`, but `_apply_vertical_shift` calls the JIT batch directly and skips that guard.

### Fix

After the optional inverse-projection step, replace any non-finite values in `xx_strip` and `yy_strip` with NaN before calling `_interp_geoid_2d`. The interpolator already returns `math.nan` for out-of-range input, so NaN-in NaN-out is fine. Update the per-strip `is_valid` mask to drop pixels with non-finite coordinates so NaN from the geoid does not contaminate the output.

The change is under ten lines, scoped to `_apply_vertical_shift`. `_vertical.py` is left alone.

## No coverage for the vertical CRS path

The `src_vertical_crs` and `tgt_vertical_crs` parameters of `reproject()` (defined at `__init__.py:507-516`) have no end-to-end tests in `xrspatial/tests/test_reproject.py`. `_apply_vertical_shift` is roughly 90 lines and is currently exercised only through unit tests of `geoid_height`.

### Tests to add

A `TestVerticalShift` class covering:

- EGM96 to ellipsoidal on a small lon/lat raster, verifying the shift matches the known undulation at a sample location.
- The reverse direction, confirming opposite sign.
- EGM96 to EGM2008, skipped if the EGM2008 grid is not available.
- Same-vertical-CRS no-op.
- Single-side vertical CRS no-op.
- A projected source/target (UTM 33N) to exercise the inverse-projection branch.
- A polar-stereographic case (EPSG:3413) on a 89-90 N raster to regression-test the hang. Assertion is that the call returns and produces finite output where the source is finite.

brendancol:
## Summary

- Mask non-finite lon/lat coords before handing them to `_interp_geoid_2d`. Inverse-projecting output grid coords for the geoid lookup can emit `inf` or NaN at projection singularities. Those values reach `_interp_geoid_point`, where the longitude wrap loop spins forever on `inf`. The fix replaces non-finite coords with NaN and drops the matching pixels from the valid-shift mask so NaN from the geoid does not contaminate the output.
- Add `TestVerticalShift` covering EGM96 to ellipsoidal and back, EGM96 to EGM2008 (skipped if the grid is not available), no-op when both vertical CRSes match or only one side is set, the projected-CRS inverse-projection branch, and a polar-stereographic regression test for the hang.

Closes #1451

## Test plan

- [x] `pytest xrspatial/tests/test_reproject.py::TestVerticalShift -v` passes (7/7)
- [x] `pytest xrspatial/tests/test_reproject.py -x` passes (143/143)
- [x] Polar-stereographic regression test returns finite output and no inf

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -775,12 +775,22 @@ def _apply_vertical_shift(data, y_coords, x_coords,
             xx_strip = np.asarray(lon_s, dtype=np.float64).reshape(n_rows, out_w)
             yy_strip = np.asarray(lat_s, dtype=np.float64).reshape(n_rows, out_w)
 
+        # Guard against non-finite output coords (projection singularities,
+        # antimeridian, polar regions). Hand NaN to the JIT batch so the
+        # longitude wrap loop in _interp_geoid_point does not see inf and
+        # spin forever.
+        finite_coord = np.isfinite(xx_strip) & np.isfinite(yy_strip)
+        if not finite_coord.all():
+            xx_strip = np.where(finite_coord, xx_strip, np.nan)
+            yy_strip = np.where(finite_coord, yy_strip, np.nan)
+
         # Apply each geoid shift
         strip_data = result[r0:r1]
         if is_nan_nodata:
             is_valid = np.isfinite(strip_data)
         else:
             is_valid = strip_data != nodata
+        is_valid = is_valid & finite_coord
 
         for (grid_data, g_left, g_top, g_rx, g_ry, g_h, g_w), sign in zip(geoids, signs):
             N_strip = np.empty((n_rows, out_w), dtype=np.float64)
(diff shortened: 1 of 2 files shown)


brendancol:

PR #1455, opened 2026-05-04
Preserve non-spatial coords through reproject() and merge()

### LINKED ISSUE #1454: reproject() and merge() drop non-spatial coords; y-axis convention undocumented
## M-6: non-spatial coords dropped by `reproject()` and `merge()`

In `xrspatial/reproject/__init__.py`, the output `DataArray` is built with only `{ydim: y_coords, xdim: x_coords}` (plus the band coord on the 3D path). Any other coord on the input is dropped:

- scalar coords (e.g. a single `time` timestamp)
- non-dimension coords (e.g. `spatial_ref` from rioxarray)
- any extra dim coord that does not apply to the rebuilt grid

This affects both `reproject()` (around the `result_data.ndim == 3` / 2D branches) and `merge()` (the final `xr.DataArray` construction).

Fix: after assigning y/x (and band) coords, copy any remaining coord whose dims do not include the spatial ydim or xdim. Coords aligned to ydim/xdim should still be dropped because their values are stale after the grid is rebuilt.

## M-8: y-axis output direction is undocumented

`_make_output_coords` in `xrspatial/reproject/_grid.py` always emits y from `top - res/2` down to `bottom + res/2`, so a y-ascending input produces a y-descending output. This is the standard raster convention (top-down, north-up) and is the right behaviour, but it is not stated in the `reproject()` or `merge()` docstrings, and the existing `test_south_up_matches_north_up` only checks `.values` equality.

Fix: add a sentence to the `Returns` section of both docstrings.

## Plan

- Update the 2D and 3D branches of `reproject()` to carry forward non-spatial coords.
- Update `merge()` to do the same from the first input raster.
- Document the y-descending output convention in both docstrings.
- Add `TestCoordsPreservation` covering scalar `time`, non-dim string coord, stale y-aligned coord drop, band coord round-trip, merge scalar coord, and y-descending output for numpy and dask backends.

brendancol:
Closes #1454.

## Summary

- `reproject()` and `merge()` rebuilt the output `DataArray` with only the y/x (and band) coords. Scalar coords like a single `time` timestamp, non-dimension coords like rioxarray's `spatial_ref`, and unrelated extra-dim coords were silently dropped.
- Both functions now copy any coord whose dims do not include the rebuilt spatial dims. Coords aligned to ydim or xdim are still dropped because their values are stale after the grid is rebuilt.
- Documents in the `Returns` section of both docstrings that the output y coordinate is always emitted in descending order (top-down, north-up) regardless of the input direction. This matches the standard raster convention.

## Test plan

- [x] `pytest xrspatial/tests/test_reproject.py -x` (155 passed)
- [x] New `TestCoordsPreservation` class covers:
  - scalar `time` coord round-trip through `reproject()`
  - non-dim string coord round-trip
  - stale y-aligned coord is dropped
  - band coord round-trip on 3D input
  - first-raster scalar coord round-trip through `merge()`
  - y-descending output for ascending input on numpy and dask backends

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -530,6 +530,17 @@ def reproject(
         The output ``attrs['crs']`` is in WKT format.
         If vertical transformation was applied, ``attrs['vertical_crs']``
         records the target vertical datum.
+
+        The output y coordinate is always emitted in descending order
+        (top-down, north-up) regardless of the input direction. This
+        matches the standard raster convention and the output of common
+        GIS libraries.
+
+        Non-spatial coords from the input (such as a scalar ``time``
+        coord or a non-dimension coord that is not aligned to the
+        spatial dims) are carried through to the output. Coords that
+        are aligned to the input y or x dims are dropped because their
+        values do not apply to the rebuilt grid.
     """
     _validate_raster(raster, func_name='reproject', name='raster',
                      ndim=(2, 3))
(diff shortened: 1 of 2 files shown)


brendancol:

PR #1456, opened 2026-05-04
Add transform_precision parameter to merge()

### LINKED ISSUE #1452: merge() lacks transform_precision parameter exposed by reproject()
## Summary

`xrspatial.reproject.merge()` is missing the `transform_precision` parameter that `reproject()` exposes. Both functions perform the same family of coordinate-transform work, so users currently have no way to control control-grid subdivisions when merging, or to request exact per-pixel pyproj transforms.

## Where the drift lives

In `xrspatial/reproject/__init__.py`:

- `reproject()` accepts `transform_precision=16` (around line 471) and documents it as the control-grid subdivision count, with `0` meaning exact per-pixel transforms.
- `merge()` does not accept it (around line 1333).
- `_merge_inmemory` hardcodes `16` when calling `_reproject_chunk_numpy` (around line 1594).
- `_merge_dask` binds `precision=16` into the per-block adapter via `functools.partial` (around line 1699).
- The existing `_validate_grid_params` call inside `merge()` already accepts a `transform_precision` argument; it is currently passed `None`.

## Fix scope

- Add `transform_precision=16` to the `merge()` signature (after `chunk_size`).
- Pass it to `_validate_grid_params` instead of `None`.
- Plumb it through `_merge_inmemory` and `_merge_dask` so the per-tile reproject calls use the user value rather than the hardcoded `16`.
- Update the `merge()` docstring with the parameter description, adapted from `reproject()`.

## What this issue intentionally does not change

`reproject()` also exposes `max_memory`, which is consumed only by the streaming reprojection path. `merge()` has no streaming path, so adding `max_memory` there would be misleading. The docstring will get a one-line note that the dask path should be used for very large outputs (which `merge()` already auto-promotes to).

brendancol:
## Summary

- Adds `transform_precision=16` to `merge()` and plumbs it through `_merge_inmemory` and `_merge_dask` (previously hardcoded to 16). Set to 0 for exact per-pixel transforms matching GDAL/rasterio.
- Passes the value through `_validate_grid_params` so the existing validation runs.
- Updates the `merge()` docstring with the parameter description and adds a Notes line saying the dask path should be used for very large outputs.
- `max_memory` is not added; `merge()` has no streaming path that would consume it.

Closes #1452

## Test plan

- [x] `pytest xrspatial/tests/test_reproject.py -x` (153 passed, including 5 new tests under `TestValidateMergeGridParams`)
- [x] `merge([raster], transform_precision=0)` runs and produces a result
- [x] Default `transform_precision=16` (no kwarg) still works
- [x] Negative and float values raise `ValueError` matching `transform_precision`
- [x] `transform_precision=0` and `transform_precision=16` produce numerically identical merged output (rtol=1e-10) on smooth EPSG:4326 inputs

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -1340,6 +1340,7 @@ def merge(
     nodata=None,
     strategy='first',
     chunk_size=None,
+    transform_precision=16,
     name=None,
 ):
     """Merge multiple rasters into a single mosaic.
@@ -1366,12 +1367,22 @@ def merge(
         Merge strategy: 'first', 'last', 'mean', 'max', 'min'.
     chunk_size : int or (int, int) or None
         Chunk size for dask output.
+    transform_precision : int
+        Control-grid subdivisions for the coordinate transform (default 16).
+        Higher values increase accuracy at the cost of more pyproj calls.
+        Set to 0 for exact per-pixel transforms matching GDAL/rasterio.
     name : str or None
         Name for the output DataArray.
 
     Returns
     -------
     xr.DataArray
+
+    Notes
+    -----
+    There is no streaming in-memory path; for very large output mosaics,
+    pass dask-backed inputs (or rely on the automatic promotion to the
+    dask path) so that each output chunk is computed independently.
     """
     if not rasters:
         raise ValueError("merge(): rasters list must not be empty")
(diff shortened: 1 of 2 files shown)


brendancol:

brendancol:

PR #1474, opened 2026-05-04
Support 3D rasters, expose nodata, document target_resolution tuple in resample

### LINKED ISSUE #1466: Support 3D rasters, expose nodata, document target_resolution tuple in resample
## Reason or Problem

Three small API gaps in `xrspatial.resample.resample()` showed up while auditing API consistency against `reproject()`:

1. 3D rasters are rejected. `_validate_raster` is called with the default `ndim=2`, so `(band, y, x)` inputs raise, even though `reproject()` accepts the same shape. Multi-band imagery has to be looped manually.
2. No `nodata` parameter. Integer rasters with sentinel values (e.g. `-9999`) cannot be resampled correctly without first converting to float and replacing the sentinel with NaN. `reproject()` already has a `nodata` keyword and falls back to `_FillValue`/`nodata` attrs.
3. The `target_resolution` docstring is wrong. The code at lines 801-803 already accepts a `(res_y, res_x)` tuple, but the docstring only documents `float`.

## Proposal

Update `resample()` to:

1. Accept 3D rasters (`band, y, x`). Loop over the leading dim and stack the per-band 2D results. Output dims and non-spatial coords match the input.
2. Add a `nodata=None` keyword. When set, replace input pixels equal to `nodata` with NaN before resampling. When None, fall back to `agg.attrs['_FillValue']`, then `agg.attrs['nodata']`. Output uses NaN as the sentinel regardless.
3. Fix the `target_resolution` docstring to `float or (float, float)`.

## Value

Brings `resample()` in line with `reproject()` for multi-band and integer-sentinel inputs. No behavior change for existing 2D float callers.

brendancol:
Closes #1466.

Three small API gaps in `xrspatial.resample.resample()`, all touching the same signature/dispatch surface, bundled into one PR.

## Summary

- 3D rasters (`band, y, x`) are now accepted. `_validate_raster` is called with `ndim=(2, 3)`. The 3D dispatch loops over the leading dim, recurses into the 2D path per band, and stacks the results with `xr.concat` followed by a `transpose` back to the original layout. Non-spatial coordinates (e.g. `band=[1, 2, 3]`) are preserved.
- New `nodata=None` keyword. Explicit value wins; otherwise it falls back to `agg.attrs['_FillValue']`, then `agg.attrs['nodata']`. Sentinel pixels are replaced with NaN before the resample using `agg.where(agg != nodata)`, which dispatches naturally for numpy / cupy / dask backings. The output advertises `_FillValue=nan` when masking was applied.
- `target_resolution` docstring updated to `float or (float, float)`. The tuple form was already accepted at lines 801-803; only the documentation was wrong.

## Test plan

- [x] All 75 tests in `test_resample.py` pass (62 existing + 13 new).
- [x] 3D raster shape (3, 8, 8) -> (3, 4, 4) at scale 0.5; `band` coord preserved.
- [x] 3D dask path (chunked over (1, 4, 4)) returns expected shape.
- [x] Per-band output of a 3D resample matches the 2D resample of each band.
- [x] `nodata=-9999` on an int raster produces NaN where the input was -9999, finite floats elsewhere.
- [x] `nodata` resolved from `_FillValue` and from `nodata` attrs.
- [x] Explicit param overrides the attr.
- [x] Without any nodata, `_FillValue` is not added to the output (no behavior change for existing 2D float callers).
- [x] `target_resolution=(2.0, 4.0)` on an 8x8 raster gives shape (4, 2).
- [ ] CuPy 3D test runs only when CUDA is available; verified locally to follow the same shape contract.

## Notes on the 3D dispatch

- The recursive call passes `nodata=float('nan')` so the per-band path doesn't try to re-mask attrs that were already masked at the 3D level. `_apply_nodata_mask` short-circuits on NaN sentinels.
- `xr.concat` was the cleanest way to stack the per-band outputs while keeping the per-band coordinate. A `transpose(*agg.dims)` afterwards keeps dim order stable.
- The identity fast-path (`scale_y == scale_x == 1.0`) sits above the 3D branch -- 3D identity copies pass through unchanged.

### DIFF
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -740,12 +740,53 @@ def _run_dask_cupy(data, scale_y, scale_x, method):
 
 # -- Public API --------------------------------------------------------------
 
+def _resolve_nodata(agg, nodata):
+    """Resolve the input-side nodata sentinel.
+
+    Explicit *nodata* wins. Otherwise fall back to ``_FillValue`` then
+    ``nodata`` in ``agg.attrs``. Returns ``None`` when no sentinel was
+    found (the caller skips the masking step).
+
+    NaN sentinels are returned as NaN so the caller can branch on
+    ``np.isnan`` rather than ``==`` (which never matches NaN).
+    """
+    if nodata is None:
+        for key in ('_FillValue', 'nodata'):
+            v = agg.attrs.get(key)
+            if v is not None:
+                nodata = v
+                break
+    if nodata is None:
+        return None
+    nd = float(nodata)
+    if np.isinf(nd):
+        raise ValueError(f"nodata must be finite or NaN, got {nodata!r}")
+    return nd
+
+
+def _apply_nodata_mask(agg, nodata):
+    """Return a float copy of *agg* with sentinel pixels replaced by NaN.
+
+    Works for numpy, cupy, dask+numpy, and dask+cupy backings via
+    xarray's ``.where`` (which dispatches per backend).
+    """
+    if nodata is None:
+        return agg
+    # Promote to float so NaN can be stored. xr.where keeps the backend.
+    if not np.issubdtype(agg.dtype, np.floating):
+        agg = agg.astype(np.float64)
+    if np.isnan(nodata):
+        return agg  # already-NaN sentinels need no replacement
+    return agg.where(agg != nodata)
+
+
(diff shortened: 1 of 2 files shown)


brendancol:

PR #1475, opened 2026-05-04
Resample polish: fix cubic depth comment, tighten NaN threshold, add edge tests

### LINKED ISSUE #1471: Resample polish: fix cubic depth comment, tighten NaN threshold, add edge tests
## Summary

Six small fixes from the resample audit:

- A-1: The comment in `xrspatial/resample.py` (lines 39-44) on the cubic overlap depth says depth 10 brings the IIR prefilter transient to "machine epsilon". The actual value is `0.268^10`, about `4e-7`, closer to single-precision epsilon (~1e-7) than float64 epsilon (~2e-16). The constant 10 is fine for float32 output, but the comment misrepresents it. Reword to "depth 10 brings the boundary transient to ~1e-7 (sufficient for float32 output and well under typical interpolation error)".

- A-6: The NaN-gating threshold in `_nan_aware_interp_np`, `_nan_aware_interp_cupy`, and the dask block variants (`_interp_block_np`, `_interp_block_cupy`) uses `z_wt > 0.01`. With 0.01, an output pixel can be marked valid even when only a tiny fraction of kernel weight came from valid input pixels (one cubic sidelobe can leak about 5% from a single neighbor). The standard choice for kernel-weighted nan-aware resampling is `> 0.5`, meaning more than half of the sampling kernel weight came from valid pixels. Switch all four sites and add a comment.

- T-3: Test that a raster with descending y coordinates (e.g. `y=[10, 9, ..., 3]`) resamples to descending y output with correct cell-centered values and no vertical flip.

- T-6: Test that `np.inf` and `-np.inf` either propagate or get treated as NaN. Lock in whichever the current code does.

- T-7: Test that the 3D-rejection error message mentions "must be 2D".

- T-9: Test that the same call produces bit-identical output across two invocations for bilinear, cubic, and average.

## Why

The doc fix and the tighter NaN gate are cheap to land. The tests cover behavior that current coverage misses: orientation, infs, and run-to-run determinism.

## Notes

No public API changes. The threshold bump is the only behavior change: pixels in the [0.01, 0.5] weight-fraction band that previously came back as a fractional value will now be NaN. That is the intended, more conservative behavior.

brendancol:
Closes #1471.

## Changes

- Reword the cubic `_INTERP_DEPTH` comment in `xrspatial/resample.py`. The previous comment claimed depth 10 brings the prefilter transient to "machine epsilon"; the actual value `0.268^10` is about `4e-7`, closer to float32 epsilon. The constant 10 is unchanged because it remains correct for float32 output.
- Raise the NaN-aware kernel-weight gate from `z_wt > 0.01` to `z_wt > 0.5` in `_nan_aware_interp_np`, `_nan_aware_interp_cupy`, `_interp_block_np`, and `_interp_block_cupy`. With the old threshold, an output pixel could pass through the gate when only a small fraction of the resampling kernel weight came from valid input pixels (one cubic sidelobe leaks about 5% from a single neighbour). The new threshold means more than half of the kernel weight must come from valid pixels. A short comment in `_nan_aware_interp_np` explains the choice; the other three sites point back to it.
- Add a test for descending y coordinates: an 8x8 grid with row 0 = 100 and row 7 = 0 must resample to a 4x4 grid that still has descending y, with cell-centered y values 9.5, 7.5, 5.5, 3.5 and the high-value row at the top of the output.
- Add tests that `+inf` and `-inf` propagate (not coerced to NaN) for bilinear, average, and nearest.
- Add a test that the 3D-rejection error message includes "must be 2D".
- Add a test that `resample(...)` produces bit-identical output across two consecutive calls for bilinear, cubic, and average.

## Behavior change

The NaN threshold bump is the only behavior change. Output pixels in the [0.01, 0.5] kernel-weight band, which previously came back as a fractional value, now come back as NaN. This is a more conservative gate.

## Test plan

- [x] `pytest xrspatial/tests/test_resample.py` (62 -> 70, all pass)
- [x] All eight new tests pass on the numpy backend

### DIFF
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -39,8 +39,9 @@ ALL_METHODS = set(INTERP_METHODS) | AGGREGATE_METHODS
 # Overlap depth (input pixels) each interpolation kernel needs from
 # neighbouring chunks when processing dask arrays.  Cubic requires extra
 # depth because the B-spline prefilter is a global IIR filter whose
-# boundary transient decays as ~0.268^n; depth 10 brings it to machine
-# epsilon.
+# boundary transient decays as ~0.268^n; depth 10 brings the boundary
+# transient to ~1e-7 (sufficient for float32 output and well under
+# typical interpolation error).
 _INTERP_DEPTH = {'nearest': 1, 'bilinear': 1, 'cubic': 10}
 
 # Approximate working-set size per output cell for the eager backends:
@@ -191,7 +192,11 @@ def _nan_aware_interp_np(data, out_h, out_w, order):
     z_data = _scipy_map_coords(filled, coords, order=order, mode='nearest')
     z_wt = _scipy_map_coords(weights, coords, order=order, mode='nearest')
 
-    result = np.where(z_wt > 0.01,
+    # Gate on majority weight: an output pixel is valid only when more
+    # than half of the resampling kernel weight came from valid input
+    # pixels.  This rejects pixels lit only by cubic-kernel sidelobes
+    # leaking small positive weight from a single neighbour.
+    result = np.where(z_wt > 0.5,
                       z_data / np.maximum(z_wt, 1e-10),
                       np.nan)
     return result.reshape(out_h, out_w)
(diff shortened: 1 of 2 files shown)


brendancol:

PR #1476, opened 2026-05-04
Preserve input float dtype through resample() (#1467)

### LINKED ISSUE #1467: Preserve input float dtype through resample()
## Describe the bug

`resample()` casts every input to `float64` internally and emits `float32` regardless of input dtype. That silently throws away precision for `float64` rasters and forces an unnecessary copy when input is already `float64`.

```python

data = np.linspace(0, 1, 64 * 64, dtype=np.float64).reshape(64, 64)
da = xr.DataArray(
    data, dims=('y', 'x'),
    coords={'y': np.arange(64), 'x': np.arange(64)},
    attrs={'res': (1.0, 1.0)},
)
out = resample(da, scale_factor=0.5)
print(out.dtype)  # float32 -- precision lost
```

## Expected behavior

- float64 input gives float64 output (no silent precision loss).
- float32 input gives float32 output (existing default).
- integer input gives float32 output, since NaN-sentinel resampling needs a float type.

## Proposal

Pick a working dtype inside each `_run_*` runner: float64 if input itemsize is at least 8, else float32. Skip the `astype` copy when input dtype already matches. Apply the same rule across all four backend runners (`_run_numpy`, `_run_cupy`, `_run_dask_numpy`, `_run_dask_cupy`).

Output dtype:
- float input: match input dtype.
- integer input: float32.

Update the docstring return-type description to match.

## Tests

- float64 input gives float64 output, values match a high-precision reference (atol=1e-12).
- float32 input gives float32 output.
- int32 input gives float32 output.
- Dask paths follow the same rules.

brendancol:
## Summary

- `resample()` picks its working dtype from the input instead of forcing float64 in / float32 out. float64 in keeps float64; float32 in keeps float32; integer / bool in returns float32 (NaN-sentinel resampling needs a float type).
- All four backend runners (numpy, cupy, dask+numpy, dask+cupy) and their block helpers thread the working/output dtypes through.
- Aggregation kernels (numba `@ngjit`) keep a hardcoded float64 accumulator internally; the runner casts the result to the output dtype at the end. That keeps existing float32 aggregation accuracy and avoids re-specialising the kernels.

Closes #1467.

## Test plan

- [x] float64 input gives float64 output, values match a float64 reference (atol=1e-12).
- [x] float32 input gives float32 output (existing default).
- [x] int16/int32/int64/uint8/uint16 input gives float32 output.
- [x] dask paths follow the same dtype rules (numpy backend).
- [x] int + nodata sentinel path is gated on the `nodata` parameter being present and skips otherwise (will activate once that parameter lands).
- [x] Existing 62 resample tests still pass.

### DIFF
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -1033,8 +1033,9 @@ def _apply_nodata_mask(agg, nodata):
     if nodata is None:
         return agg
     # Promote to float so NaN can be stored. xr.where keeps the backend.
+    # Integer / bool inputs become float32 (consistent with _output_dtype).
     if not np.issubdtype(agg.dtype, np.floating):
-        agg = agg.astype(np.float64)
+        agg = agg.astype(np.float32)
     if np.isnan(nodata):
         return agg  # already-NaN sentinels need no replacement
     return agg.where(agg != nodata)
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -44,11 +44,42 @@ ALL_METHODS = set(INTERP_METHODS) | AGGREGATE_METHODS
 _INTERP_DEPTH = {'nearest': 1, 'bilinear': 1, 'cubic': 10}
 
 # Approximate working-set size per output cell for the eager backends:
-# one float64 working buffer (8 B) plus a float32 output cell (4 B).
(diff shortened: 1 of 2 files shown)


brendancol:

brendancol:

PR #1478, opened 2026-05-04
Make cubic resample prefilter explicit so chunk seams stay sub-eps

### LINKED ISSUE #1464: Make cubic resample prefilter explicit so chunk seams stay sub-eps
**Describe the bug**

`xrspatial.resample.resample(..., method='cubic')` calls `scipy.ndimage.map_coordinates(data, coords, order=3, mode='nearest')` without `prefilter=False`. With `order=3`, scipy runs an internal `spline_filter` before sampling. The internal prefilter honors the `mode` argument since scipy 1.6, but the boundary transient of the IIR prefilter still depends on how the array is presented to it. A full unchunked array sees one global prefilter pass; each dask chunk sees its own prefilter pass on the (overlapped) chunk view. Even with overlap depth 10, the implicit per-block prefilter on the NaN-aware `filled` and `weights` arrays gives values that differ from the eager pass by ~1e-6 near chunk seams.

The current dask-vs-numpy parity test for cubic uses `atol=1e-5`, which hides the difference.

**Expected behavior**

A cubic resample of a smooth surface (e.g. a degree-2 polynomial) should match between the eager numpy path and the chunked dask+numpy path to within float64 round-off (~1e-10 atol).

**To reproduce**

```python
import numpy as np
import xarray as xr

y, x = np.mgrid[0:30, 0:30].astype(np.float64)
f = x*x + y*y - x*y
da_np = xr.DataArray(f, dims=('y', 'x'),
                     coords={'y': np.arange(30), 'x': np.arange(30)},
                     attrs={'res': (1.0, 1.0)})

import dask.array as da
da_dk = da_np.copy()
da_dk.data = da.from_array(da_np.data, chunks=(8, 8))

out_np = resample(da_np, scale_factor=0.5, method='cubic').values
out_dk = resample(da_dk, scale_factor=0.5, method='cubic').values

print(np.max(np.abs(out_np - out_dk)))  # ~1e-6, want ~1e-10
```

**Fix**

Call `scipy.ndimage.spline_filter(data, order=3, mode='nearest')` once per block (and per array - `filled` and `weights` get separate prefilter passes for the NaN-aware case), then call `map_coordinates(..., prefilter=False)` on the prefiltered coefficients. Mirror the change in the cupy path with `cupyx.scipy.ndimage.spline_filter`. Skip the explicit prefilter when `order < 2`; bilinear `order=1` prefilter is a no-op.

**Why it matters**

Chunk-seam parity for cubic resample is a correctness property users rely on when porting eager prototypes to chunked dask jobs. Tightening the parity bound from 1e-5 to 1e-10 in the test catches future regressions in prefilter handling.

brendancol:
Closes #1464.

## Summary

`scipy.ndimage.map_coordinates(order >= 2, prefilter=True)` does three things behind the scenes for `mode='nearest'`:

1. Edge-pad the input by 12 pixels.
2. Run `spline_filter` on the padded array.
3. Shift the sample coordinates by 12 to account for the pad.

The whole step happens inside each `map_coordinates` call. In the NaN-aware path it runs twice per block (once for `filled`, once for `weights`). In the dask path it runs once per chunk on the overlapped block, with the prefilter's IIR boundary transient anchored to the chunk's overlapped edge rather than the global edge. With `depth=10` overlap the transient leaked into chunk-interior samples at ~1e-6 near chunk seams.

This PR runs the prepad + `spline_filter` step ourselves, then calls `map_coordinates(prefilter=False)`. Boundary handling is now the same between numpy and dask paths, and the per-block prefilter on `filled` and `weights` runs once each instead of twice.

The IIR transient still depends on overlap depth. The previous `_INTERP_DEPTH['cubic'] = 10` left ~2e-6 residual, which is what surfaced this issue. Bumped to 16, which leaves ~7e-10 residual, below float32 epsilon, so the output rounds bit-identical between eager and chunked paths.

Same change in the cupy path with `cupyx.scipy.ndimage.spline_filter`.

## Test plan

- [x] Existing 62 resample tests still pass
- [x] New `TestCubicPrefilterParity` suite (6 tests) asserts `atol=1e-10` between eager numpy and dask+numpy for cubic resample of:
  - degree-2 polynomial (60x60, chunks=13)
  - random data (50x50, chunks=11)
  - random data at scale factors 0.5, 2.0, 0.7 (48x48, chunks=11)
  - random data with NaN (50x50, chunks=11)
- [x] Confirmed new tests fail without the fix: max diff was ~1e-6 on the random-data cases
- [x] Confirmed new tests pass with the fix at the 1e-10 bound

### DIFF
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -9,7 +9,10 @@ from functools import partial
 
 import numpy as np
 import xarray as xr
-from scipy.ndimage import map_coordinates as _scipy_map_coords
+from scipy.ndimage import (
+    map_coordinates as _scipy_map_coords,
+    spline_filter as _scipy_spline_filter,
+)
 
 try:
     import dask.array as da
@@ -39,9 +42,10 @@ ALL_METHODS = set(INTERP_METHODS) | AGGREGATE_METHODS
 # Overlap depth (input pixels) each interpolation kernel needs from
 # neighbouring chunks when processing dask arrays.  Cubic requires extra
 # depth because the B-spline prefilter is a global IIR filter whose
-# boundary transient decays as ~0.268^n; depth 10 brings it to machine
-# epsilon.
-_INTERP_DEPTH = {'nearest': 1, 'bilinear': 1, 'cubic': 10}
+# boundary transient decays as ~0.268^n.  Depth 16 puts the residual at
+# ~7e-10, comfortably below float32 epsilon so chunk-seam parity rounds
+# to zero in the float32 output.
+_INTERP_DEPTH = {'nearest': 1, 'bilinear': 1, 'cubic': 16}
 
 # Approximate working-set size per output cell for the eager backends:
 # one float64 working buffer (8 B) plus a float32 output cell (4 B).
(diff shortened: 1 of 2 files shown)


brendancol:

brendancol:

PR #1499, opened 2026-05-06
Decode TIFF predictor=2 sample-wise for multi-byte integer round-trip

brendancol:
## Summary
- TIFF predictor=2 decode and encode were byte-wise at a sample stride, which loses carries inside multi-byte samples. Files xrspatial wrote round-tripped because both halves were wrong in matching ways, but uint16, int16, uint32 and int32 files written by libtiff/GDAL with predictor=2+deflate/lzw silently decoded to wrong pixel values whenever an adjacent-sample difference crossed a byte boundary.
- Decode now operates on whole sample values in the file's byte order, per-channel for chunky multi-band data; the 8-bit fast path stays. Predictor=2 on floats stays byte-wise (the spec restricts predictor=2 to integers; floats are out of spec for this predictor) so old xrspatial-written float files still read back unchanged.
- The GPU encode dispatch was also fixed: it had been called with row_bytes multiplied by samples-per-pixel twice, which was inert only because the byte-wise math was wrong on both sides.

## Repro before the fix
```
import rasterio, numpy as np
arr = np.array([[1000, 2000, 3000, 4000]], dtype='uint16')
with rasterio.open('/tmp/r.tif', 'w', driver='GTiff',
                   height=1, width=4, count=1, dtype='uint16',
                   compress='deflate', predictor=2) as dst:
    dst.write(arr, 1)

from xrspatial.geotiff._reader import read_to_array
print(read_to_array('/tmp/r.tif')[0])  # before: [1000 1744 2488 3232]
```

## Test plan
- [x] New `test_predictor2_samplewise.py`: rasterio (libtiff) round-trips for LE uint16/int16/uint32/int32 with carries
- [x] tifffile round-trips for big-endian uint16/int16/uint32/int32
- [x] Multi-band chunky uint16 round-trip against libtiff
- [x] xrspatial-written uint16 readable by rasterio (interop the other direction)
- [x] GPU multi-byte decode parity for libtiff predictor=2 files
- [x] Existing predictor and writer test suites still pass (CPU and GPU paths)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 -----------
 
 
+### Unreleased
+
+#### Bug fixes and improvements
+- Decode TIFF predictor=2 sample-wise so multi-byte integer files written by libtiff/GDAL read back exactly
+
+
 ### Version 0.9.9 - 2026-05-05
 
 #### Bug fixes and improvements
--- a/xrspatial/geotiff/_compression.py
+++ b/xrspatial/geotiff/_compression.py
@@ -333,80 +333,164 @@ def lzw_compress(data: bytes) -> bytes:
 
 
 # -- Horizontal predictor (Numba) --------------------------------------------
+#
+# TIFF predictor=2 (horizontal differencing) operates on whole *samples*
+# (per-pixel-component values), not on individual bytes.  For multi-byte
+# integer samples the difference is computed across full sample values, so
+# carries between the low and high bytes are honoured.  The byte stream that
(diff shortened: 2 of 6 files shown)


brendancol:
Closing as duplicate of #1498 (already merged). The audit agent that opened this PR was working from a stale local main that didn't have #1498 yet, so it re-derived the same fix. Verified: with #1498 applied, the repro `[1000, 2000, 3000, 4000]` uint16 predictor=2 round-trips correctly through rasterio.

PR #1519, opened 2026-05-08
Reject mixed BitsPerSample per band in TIFF readers

### LINKED ISSUE #1505: Mixed BitsPerSample per band silently uses bps[0]
## Problem

TIFF allows a different bit depth per sample: `BitsPerSample` is a count-`SamplesPerPixel` array. The reader uses only `bps[0]` everywhere and assumes uniform bit width across bands.

Real-world example: a TIFF with `SamplesPerPixel=4`, `BitsPerSample=(16, 16, 16, 8)` (RGB at 16-bit, alpha at 8-bit). xrspatial decodes all four bands as 16-bit, producing garbage in the alpha band.

## Where

`xrspatial/geotiff/_dtypes.py` and `xrspatial/geotiff/_reader.py` — bit-depth dispatch.

## Fix sketch

Two reasonable options:

1. Reject mixed bit depths with a clear error message.
2. Decode each band at its own bit width into a separate array, then promote to a common dtype.

Option 1 is fine for now since this is rare in practice.

## Severity

Low — uncommon in practice, but silent data corruption when it happens.

brendancol:
Closes #1505

A TIFF whose `BitsPerSample` tag holds a different value per band (e.g. `(16, 16, 16, 8)` for RGB plus an 8-bit alpha) was silently decoded with the first band's width applied to every band, so the mismatched bands came back as garbage.

The fix adds a `_dtypes.resolve_bits_per_sample(bps)` helper and threads every reader site through it: stripped CPU, tiled CPU, GPU, VRT, the geo-info probe, and the palette colormap path. The helper accepts a scalar or a tuple/list whose entries all agree, and raises `ValueError` with the mixed values and a `gdal_translate` hint when they don't.

Decoding each band separately and promoting to a common dtype was the alternative; it's more code for a rare case, and silent promotion raises follow-up questions (whose nodata wins, what does the user actually want). An explicit error lets the user convert with GDAL or rasterio.

## Test plan
- [x] `pytest xrspatial/geotiff/tests/test_mixed_bps.py` -- 9 new tests: the helper plus an end-to-end RGB+8-bit-alpha file built from raw IFD bytes
- [x] `pytest xrspatial/geotiff/tests/` -- 699 passed; 3 pre-existing matplotlib palette failures are unrelated

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -194,7 +194,7 @@ def _read_geo_info(source, *, overview_level: int | None = None):
     overview_level : int or None
         Overview IFD index (0 = full resolution).
     """
-    from ._dtypes import tiff_dtype_to_numpy
+    from ._dtypes import resolve_bits_per_sample, tiff_dtype_to_numpy
     from ._geotags import extract_geo_info
     from ._header import parse_all_ifds, parse_header, select_overview_ifd
     from ._reader import _coerce_path, _is_file_like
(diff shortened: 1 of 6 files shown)


brendancol:

PR #1535, opened 2026-05-08
Apply LERC valid-mask in GPU decode path (depends on #1529)

brendancol:
## Summary

Follow-up to #1529. The CPU LERC reader applies the LERC valid-mask
and writes nodata into masked positions, but the GPU LERC tile-decode
path in `_gpu_decode.py` was still dropping the mask. A masked pixel
read back as 0 on GPU and as NaN or the sentinel on CPU.

This PR fixes the GPU side:

- The GPU LERC branch calls `lerc_decompress_with_mask` per tile and
  keeps any returned mask.
- After tile assembly, an invalid-mask sized to the output image is
  built on host, copied to the GPU once, and used to write the
  resolved fill value into masked positions.
- `gpu_decode_tiles` and `gpu_decode_tiles_from_file` get a
  `masked_fill=` kwarg that `read_geotiff_gpu` populates via
  `_resolve_masked_fill(ifd.nodata_str, file_dtype)` when compression
  is LERC.

Depends on #1529 (uses `lerc_decompress_with_mask` and
`_resolve_masked_fill` introduced there). Land #1529 first; this PR
should rebase cleanly off main once it is merged.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_lerc_valid_mask_gpu.py` passes (4 tests).
- [x] `pytest xrspatial/geotiff/tests/test_lerc.py test_lerc_max_z_error.py test_gpu_byteswap_1508.py test_lerc_valid_mask.py` passes (38 tests).
- [x] Manual repro: float32 LERC TIFF with a masked pixel read via `read_geotiff_gpu` returns NaN at the masked position (matches CPU); without the fix it returned 0.0.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -1617,7 +1617,9 @@ def read_geotiff_gpu(source: str, *,
 
     from ._reader import (
         _FileSource, _check_dimensions, MAX_PIXELS_DEFAULT, _coerce_path,
+        _resolve_masked_fill,
     )
+    from ._compression import COMPRESSION_LERC
     from ._header import (
         parse_header, parse_all_ifds, select_overview_ifd, validate_tile_layout,
     )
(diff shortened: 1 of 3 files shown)


brendancol:
@copilot review

PR #1569, opened 2026-05-11
Fix balanced_allocation dask+cupy TypeError on source_ids

### LINKED ISSUE #1568: balanced_allocation: TypeError on dask+cupy backend from cupy source_ids
**Describe the bug**

`balanced_allocation` raises `TypeError: Implicit conversion to a NumPy array is not allowed` when called on a dask+cupy raster.

```
File "xrspatial/balanced_allocation.py", line 359, in balanced_allocation
    float(np.sum(fric_weight[alloc == sid]))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
cupy/_core/core.pyx:1581: TypeError
```

Root cause: `_extract_sources` calls `da.unique(data).compute()`, which on a dask+cupy array returns a cupy array. `np.sort(uniq[mask])` does not convert, so `source_ids` stays cupy. Downstream, `source_ids[best_idx]` in `_allocate_biased` produces a cupy `alloc`, and the masked `fric_weight[alloc == sid]` tries an implicit numpy conversion.

**Expected behavior**

`balanced_allocation` should return a valid allocation raster for all four backends (numpy, cupy, dask+numpy, dask+cupy), matching the test `test_two_sources_uniform_friction`.

**Reproduction**

```
pytest xrspatial/tests/test_balanced_allocation.py::test_two_sources_uniform_friction
```

The `[dask+cupy]` parametrization fails; the other three pass.

**Fix**

Force `source_ids` to numpy at the end of `_extract_sources` so all downstream indexing produces numpy arrays.

brendancol:
## Summary
- Fixes #1568
- `_extract_sources` returned a cupy array on dask+cupy rasters because `da.unique().compute()` yields cupy and `np.sort` did not convert. That propagated into `source_ids[best_idx]`, making `alloc` cupy, and `fric_weight[alloc == sid]` tripped cupy's implicit-numpy-conversion guard.
- Convert the unique result to numpy via `.get()` before mask/sort so `source_ids` is always numpy.

## Test plan
- [x] `pytest xrspatial/tests/test_balanced_allocation.py` — 17/17 pass across numpy, cupy, dask+numpy, dask+cupy
- [x] Failing parametrization `test_two_sources_uniform_friction[dask+cupy]` now passes


brendancol:
Addressed Copilot's review note: now using the existing `_as_numpy()` helper in `_extract_sources` instead of inlining the `.get()` check, keeping backend-detection logic centralized.

Also folded in latest `main` to drop a few unrelated files that came along when I branched from a locally-ahead `main`. PR diff is now just `xrspatial/balanced_allocation.py`.

CI: 10/11 jobs green. The remaining failure on `run (windows-latest, 3.13)` is `test_perf_under_100ms_on_1024sq_uint8` in the geotiff module (mode resampling took 327ms vs a 100ms threshold on a slow Windows runner) — unrelated to this change.

Issue #1582, opened 2026-05-11
to_geotiff drops rioxarray nodatavals and CF _FillValue silently


brendancol:
## Summary

`to_geotiff` only reads `attrs['nodata']` when extracting the NoData
sentinel from a DataArray. DataArrays produced by rioxarray (the most
common GeoTIFF reader in the ecosystem) carry the sentinel under
`attrs['nodatavals']` (tuple, one per band), and CF-style xarray
pipelines use `attrs['_FillValue']`. Both are silently dropped on
write: the file lands without a GDAL_NODATA tag and the sentinel pixels
are stored as ordinary data values.

The user has to know to rename the attr by hand before the call. This
is exactly the failure mode the metadata-propagation sweep targets:
the call succeeds, the file looks fine, but a downstream reader sees a
raster with no nodata and treats the sentinel pixels as real data.

## Reproduction

```python
import numpy as np, xarray as xr
from xrspatial.geotiff import to_geotiff, open_geotiff

arr = np.array([[1., 2., -9999.], [3., -9999., 5.]], dtype=np.float32)
da = xr.DataArray(arr, dims=['y', 'x'],
                  coords={'y': [10., 20.], 'x': [100., 110., 120.]},
                  attrs={'crs': 4326, 'nodatavals': (-9999.,)})
to_geotiff(da, '/tmp/r.tif')
rd = open_geotiff('/tmp/r.tif')
print(rd.attrs.get('nodata'))   # None
print(rd.values)                # sentinel still present, no NaN
```

## Fix

In `to_geotiff` (CPU path) and `write_geotiff_gpu` (GPU path), when
`nodata is None` and `attrs['nodata']` is unset, fall back to
`attrs['nodatavals']` (rioxarray) and then `attrs['_FillValue']` (CF).
Use the first band's value if `nodatavals` is a sequence.

Found by the metadata-propagation sweep (Cat 1 + Cat 4).

brendancol:
Fixed in commit deef216 (PR #1585). The `_resolve_nodata_attr` helper now walks `attrs['nodata']` → `attrs['nodatavals']` → `attrs['_FillValue']` in both `to_geotiff` (CPU path) and `write_geotiff_gpu`. Coverage in `xrspatial/geotiff/tests/test_nodata_attr_aliases_1582.py`.

PR #1585, opened 2026-05-11
Fix geotiff metadata propagation gaps (#1580, #1582)

### LINKED ISSUE #1580: write_geotiff_gpu silently mis-orders (band, y, x) DataArrays
## Summary

`write_geotiff_gpu` does not handle band-first `(band, y, x)` 3D
DataArrays the way the CPU `to_geotiff` path does. The CPU writer
checks `data.dims[0] in ('band', 'bands', 'channel')` and moveaxis to
`(y, x, band)` (see `__init__.py:1037`). The GPU writer skips that
step entirely and treats `arr.shape[2]` as the band axis regardless of
dim names, so a rioxarray-style `(band, y, x)` CuPy DataArray is
written with the band axis stored as image width and the actual width
stored as samples-per-pixel.

`to_geotiff(...)` auto-dispatches to `write_geotiff_gpu` when the
input is CuPy-backed, so the same caller code that round-trips fine on
numpy silently produces a transposed, unreadable file on GPU.

## Reproduction

```python
from xrspatial.geotiff import open_geotiff, write_geotiff_gpu

arr = np.random.default_rng(0).integers(0, 255, (3, 16, 32), dtype=np.uint8)
da = xr.DataArray(
    cupy.asarray(arr),
    dims=['band', 'y', 'x'],
    coords={'band': np.arange(3),
            'y': np.arange(16, dtype=np.float64),
            'x': np.arange(32, dtype=np.float64)},
    attrs={'crs': 4326},
)
write_geotiff_gpu(da, '/tmp/bandfirst.tif', compression='none')
rd = open_geotiff('/tmp/bandfirst.tif')
print(rd.sizes)  # Frozen({'y': 3, 'x': 16, 'band': 32}) — wrong
```

Expected: `Frozen({'y': 16, 'x': 32, 'band': 3})`.

## Fix

Mirror the moveaxis branch from the CPU eager path before computing
`height, width = arr.shape[:2]`. Same condition: 3D and
`data.dims[0] in ('band', 'bands', 'channel')`.

Found by the metadata-propagation sweep (Cat 3 + Cat 5).

brendancol:
## Summary

Closes #1580 and #1582 -- two metadata-propagation defects flagged by
the geotiff sweep.

- `write_geotiff_gpu` treated `arr.shape[:2]` as `(height, width)`
  regardless of dim names, so a rioxarray-style `(band, y, x)` CuPy
  DataArray was written with the band axis stored as image width and
  the real width stored as samples-per-pixel. The CPU writer already
  remaps `(band, y, x) -> (y, x, band)` when `data.dims[0]` is
  band/bands/channel; the GPU writer now mirrors that step so both
  backends produce the same file for the same input.

- `to_geotiff` (and the GPU writer it dispatches to) only consulted
  `attrs['nodata']` when extracting the NoData sentinel, silently
  dropping rioxarray's `attrs['nodatavals']` and CF-style
  `attrs['_FillValue']`. A new `_resolve_nodata_attr` helper walks
  those aliases in priority order (`nodata` -> `nodatavals` ->
  `_FillValue`). xrspatial's own `nodata` key still wins when both
  are present, preserving the existing intra-library round-trip.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_gpu_writer_band_first_1580.py` -- 6 passed
- [x] `pytest xrspatial/geotiff/tests/test_nodata_attr_aliases_1582.py` -- 11 passed
- [x] `pytest xrspatial/geotiff/tests/test_gpu_writer_attrs_1563.py xrspatial/geotiff/tests/test_attrs_parity_1548.py` -- no regressions
- [x] Full `xrspatial/geotiff/tests/` -- only pre-existing matplotlib
      `TestPalette::test_xrs_plot_*` failures remain (unrelated to
      this branch; reproducible on `main`)

### DIFF
--- a/.claude/sweep-metadata-state.csv
+++ b/.claude/sweep-metadata-state.csv
@@ -1,2 +1,3 @@
 module,last_inspected,issue,severity_max,categories_found,notes
+geotiff,2026-05-11,1580;1582,HIGH,1;3;4;5,"write_geotiff_gpu silently mis-ordered (band, y, x) DataArrays (#1580). to_geotiff ignored rioxarray nodatavals and CF _FillValue attrs (#1582). Both fixed in same branch."
 reproject,2026-05-10,1572;1573,HIGH,1;3;4,geoid_height_raster dropped input attrs and used dims[-2:] for 3D inputs (#1572). reproject/merge ignored nodatavals (rasterio convention) when rioxarray absent (#1573). Fixed in same branch.
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -660,14 +660,18 @@ def _resolve_nodata_attr(attrs: dict):
     so that key is checked first for a clean intra-library round-trip.
     Falls back to two ecosystem conventions on miss:
 
-    * ``attrs['nodatavals']`` -- rioxarray's per-band tuple. Uses the
-      first band's value; bands with mixed sentinels are uncommon and
-      would need an explicit ``nodata=`` argument anyway.
+    * ``attrs['nodatavals']`` -- rioxarray's per-band tuple. Returns
+      the first entry that is not None, not non-numeric, and not NaN.
+      In practice this is band 0 for almost every real file; the skip
+      logic only matters when band 0 is missing a sentinel (NaN /
+      None) while a later band declares one. Bands with mixed concrete
+      sentinels are uncommon and would need an explicit ``nodata=``
+      argument anyway.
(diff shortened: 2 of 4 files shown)


brendancol:
Thanks @Copilot — valid catch. Pushed 2e87151 rewording the `nodatavals` bullet to describe the actual selection logic (first non-None, non-NaN, numeric entry), with a note that this is band 0 in practice and the skip only matters when band 0 itself lacks a sentinel.

PR #1593, opened 2026-05-11
Probe: capture copilot-review event payload

brendancol:
## Summary
- One-line comment touch to fire the `copilot-review` workflow so the unconditional `debug-event` job (landed in #1592) prints the event payload values for this PR.

## Test plan
- [ ] Inspect the `debug-event` job logs and record the printed `assoc` value
- [ ] Use that to fix the `if:` gate in a follow-up, then revert this PR and remove the debug job


brendancol:
Superseded by #1594 — debug output captured (`assoc=CONTRIBUTOR`), follow-up fix opened.

Issue #1597, opened 2026-05-11
read_geotiff_dask drops nodata mask when sentinel only appears in non-first chunks


brendancol:
## Summary

`read_geotiff_dask` declares the output dtype as `float64` for integer rasters paired with an in-range nodata sentinel, but per-chunk dtype handling in `_delayed_read_window` only promotes a chunk to `float64` when that chunk actually contains a sentinel pixel. When the sentinel falls in a non-first chunk, dask preallocates the concatenated output from the first chunk's dtype (`uint16`), then casts subsequent `float64` chunks back to `uint16`, replacing `NaN` with `0` and emitting `RuntimeWarning: invalid value encountered in cast`.

Net effect: the declared `float64` array silently becomes `uint16` at compute time, and nodata pixels become `0` instead of `NaN`. The numpy eager path is correct.

## Repro

```python
import numpy as np, tempfile, os
from xrspatial.geotiff import open_geotiff
from xrspatial.geotiff._writer import write

with tempfile.TemporaryDirectory() as d:
    path = os.path.join(d, 't.tif')
    arr = np.arange(64, dtype=np.uint16).reshape(8, 8) + 1
    arr[6:8, 6:8] = 65535  # sentinel only in bottom-right
    write(arr, path, nodata=65535, compression='none', tiled=False)

    eager = open_geotiff(path)
    dk = open_geotiff(path, chunks=4)
    r = dk.compute()
    print(eager.dtype, np.isnan(eager.values).sum())  # float64 4
    print(dk.dtype, r.dtype, (r.values[6:8, 6:8] == 0).all())  # float64 uint16 True
```

## Root cause

In `xrspatial/geotiff/__init__.py`:

* `read_geotiff_dask` computes `effective_dtype = float64` for masked-int rasters (line 1510 onward) and declares each dask block with `dtype=target_dtype` (line 1654).
* `_delayed_read_window` (line 1714) only calls `arr.astype(np.float64)` inside `if mask.any():`; if no sentinel pixel is in the chunk, `arr` stays at the file's integer dtype.
* The per-chunk `arr.astype(target_dtype)` cast only runs when the *user* passed an explicit `dtype=` kwarg (the caller threads `target_dtype=target_dtype if dtype is not None else None`, line 1650), so the float promotion is not enforced on chunks the mask missed.

## Proposed fix

[...]

PR #1626, opened 2026-05-11
Skip no-op astype in _delayed_read_window (#1624)

### LINKED ISSUE #1624: read_geotiff_dask per-chunk astype copies even when dtype already matches
**Describe the bug**

`_delayed_read_window` in `xrspatial/geotiff/__init__.py:1871-1872` calls `arr.astype(target_dtype)` on every chunk. When `target_dtype` already equals `arr.dtype` (the common case for float source rasters), `numpy.ndarray.astype` still allocates a new buffer and copies because its default is `copy=True`.

PR #1601 widened the call site to always pass `target_dtype` so dask declared dtype and per-chunk dtype agree. That fixed #1597 for integer rasters with an in-range nodata sentinel, where the dask graph declared float64 but only chunks that hit the sentinel actually promoted, so concatenation cast later chunks back to int and clobbered NaN with 0. Correct fix, but the always-on cast now allocates a same-dtype copy on every chunk of every read.

**Reproduction**

```python
from xrspatial.geotiff import to_geotiff, read_geotiff_dask
import xrspatial.geotiff as gt

H, W = 1024, 1024
data = np.random.rand(H, W).astype(np.float32)
arr_in = xr.DataArray(data, dims=['y', 'x'],
                      coords={'y': np.arange(H), 'x': np.arange(W)})
tmp = tempfile.mkdtemp()
path = os.path.join(tmp, 'probe.tif')
to_geotiff(arr_in, path, compression='none')

orig = gt._delayed_read_window
trace = []
def patched(*args, **kwargs):
    trace.append(kwargs.get('target_dtype'))
    return orig(*args, **kwargs)
gt._delayed_read_window = patched

read_geotiff_dask(path, chunks=256)
# Every delayed call carries target_dtype=float32 even though no cast is needed.
assert all(t == np.dtype('float32') for t in trace if t is not None)
```

**Expected behavior**

Skip the cast when `target_dtype == arr.dtype`. A 30 TB float32 read with 100 MB chunks shouldn't pay an extra 100 MB allocation and memcpy per task for a no-op cast.

**Fix**

```python
if target_dtype is not None and arr.dtype != target_dtype:
    arr = arr.astype(target_dtype)
```

The #1597 fix still holds. The integer-mask branch above promotes `arr` to float64 in place when sentinels hit, so when `target_dtype == float64` and `arr.dtype == float64`, the astype is a no-op; when `target_dtype` differs (caller-supplied dtype, or `effective_dtype=float64` over an unmasked int chunk), the astype runs as before.

**Context**

Found during the 2026-05-11 geotiff performance sweep.

brendancol:
Closes #1624.

## Summary

- Gate the per-chunk `arr.astype(target_dtype)` in `_delayed_read_window` on a real dtype mismatch.
- After #1601 widened the call site to always pass `target_dtype`, every dask chunk ran `astype` even when `arr.dtype == target_dtype`. `numpy.astype` defaults to `copy=True` and so allocated a chunk-sized same-dtype buffer and memcpy on every read.
- The #1597 mask path still promotes uint to float64 inline, so the contract that every chunk lands in the dask-declared dtype is unchanged.

## Test plan

- [x] New regression test `test_dask_no_op_astype_1624.py` wraps `read_to_array` so it returns an ndarray subclass that records `astype` calls, then asserts no same-dtype `astype` survives the per-chunk return path.
- [x] Test fails on `main` with the expected error message; passes after the fix.
- [x] Existing `test_dask_int_nodata_chunks_1597.py` (the regression suite from #1601) still passes -- the int-with-sentinel promotion is intact.
- [x] Full `xrspatial/geotiff/tests/` suite passes apart from three pre-existing `TestPalette` failures on `main` that are unrelated to this change.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -1868,7 +1868,14 @@ def _delayed_read_window(source, r0, c0, r1, c1, overview_level, nodata,
                     if mask.any():
                         arr = arr.astype(np.float64)
                         arr[mask] = np.nan
-        if target_dtype is not None:
+        if target_dtype is not None and arr.dtype != target_dtype:
+            # Skip the cast when dtype already matches. ``numpy.astype``
+            # defaults to ``copy=True`` and would otherwise allocate a
+            # full chunk-sized buffer and memcpy on every read just to
+            # land in the same dtype the array already has. The int->
+            # float64 promotion above (sentinel-hit branch) keeps the
+            # contract that every chunk lands in the dask-declared
+            # dtype; this guard only elides no-op casts. See #1624.
             arr = arr.astype(target_dtype)
         return arr
     return _read(http_meta_key)
--- a/xrspatial/geotiff/tests/test_dask_no_op_astype_1624.py
+++ b/xrspatial/geotiff/tests/test_dask_no_op_astype_1624.py
@@ -41,37 +41,6 @@ def uint16_with_sentinel_in_first_chunk(tmp_path):
     return path, arr
 
 
-def test_float32_chunks_avoid_redundant_copy(float32_no_nodata_tif,
-                                             monkeypatch):
-    """Plain float32 read should not call astype with the same dtype.
-
-    Patches ``ndarray.astype`` via a wrapper installed on the chunk
-    return path to count same-dtype casts. Without the fix every chunk
-    triggers one; with the fix none do.
-    """
-    import xrspatial.geotiff as gt
-
-    path, _ = float32_no_nodata_tif
-    same_dtype_casts: list[tuple] = []
-
-    orig = gt._delayed_read_window
-
-    def wrapped(*args, **kwargs):
-        delayed = orig(*args, **kwargs)
-        return delayed
-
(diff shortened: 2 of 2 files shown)


brendancol:
Addressed both Copilot review comments in 4c66bb1:

1. **`test_float32_chunks_avoid_redundant_copy`** — deleted. The wrapper around `_delayed_read_window` never recorded any astype calls, the `same_dtype_casts` list was unused, and the only assertion was on output dtype. `test_astype_skipped_when_dtypes_match` already covers the same case correctly via an ndarray subclass that records `astype` invocations. Confirmed it fails on the pre-fix code with `AssertionError: Same-dtype astype still runs per chunk (dtype=float32, calls=[dtype('float32')])` and passes after the fix.

2. **`test_uint16_mask_path_still_promotes`** — the previously-unused `arr` is now exercised: the test asserts that pixels equal to the 65535 sentinel become NaN and every other pixel matches `arr.astype(float64)` byte-for-byte. Catches mask-branch regressions instead of just dtype drift.

`pyflakes` clean. All three remaining tests pass.

PR #1629, opened 2026-05-11
Cap LERC and JPEG 2000 decompression output to block bomb attacks (#1625)

### LINKED ISSUE #1625: LERC and JPEG2000 codecs lack decompression-bomb cap
## Summary

`lerc_decompress_with_mask` (xrspatial/geotiff/_compression.py:1191) and `jpeg2000_decompress` (line 1114) call the underlying `lerc.decode(data)` / `glymur.Jp2k(tmp)[:]` with no pre-decode bound on the output size. The deflate / zstd / lz4 / packbits codecs all received a decompression-bomb cap in #1533, but LERC and JPEG2000 were missed.

## Impact

A crafted GeoTIFF tile compressed with LERC or JPEG2000 can declare arbitrarily large pixel dimensions inside the codestream. The reader's post-decode size validation in `_decode_strip_or_tile` (line 793) catches the bomb only **after** the underlying library has materialised the full decoded buffer in memory, so the host is OOM-killed before the check fires.

LERC compresses constant-value blocks at >700,000:1: a 94-byte blob can decode to 64 MiB. A ~1 KB on-disk LERC tile can request multiple GB of host memory before the size check rejects it. JPEG2000 has similar amplification potential through tile-size declarations in the codestream.

## Verification

```python
import lerc, numpy as np
arr = np.zeros((4096, 4096), dtype=np.float32)
encoded = lerc.encode(arr, 1, False, None, 0.0, 1)
blob = bytes(encoded[2])
# 94 bytes -> 64 MB. info[4]=4096, info[5]=4096, info[6]=1 bands
```

The existing test_decompression_caps.py covers deflate / zstd / lz4 / packbits but not lerc / jpeg2000 / jpeg.

## Fix

Pre-validate the codestream's declared output size against the caller-supplied `expected_size * 1.05 + 1` cap before invoking the underlying library:

- LERC: `lerc.getLercBlobInfo(data)` returns `(errCode, version, dataType, nDim, nCols, nRows, nBands, ...)` without decoding. Compute `nCols * nRows * nBands * dtype_size` and raise `ValueError` if it exceeds the margin cap.
- JPEG2000: read `glymur.Jp2k(tmp).shape` before calling `[:]`.

JPEG (Pillow) already has its own `Image.MAX_IMAGE_PIXELS` guard and emits `Image.DecompressionBombError`; it is protected at the library level and does not need a wrapper-level cap.

## Severity

MEDIUM — exploitable only on readers that have the optional `lerc` / `glymur` packages installed, but no special privileges or crafted-bytestream knowledge is required beyond producing a tiny LERC blob with large declared dimensions.

Found by `/sweep-security` on the geotiff module (2026-05-11).

brendancol:
## Summary

`lerc_decompress_with_mask` and `jpeg2000_decompress` did not enforce the same pre-decode output-size cap that deflate, zstd, lz4, and packbits got in #1533. A crafted GeoTIFF tile compressed with LERC or JPEG 2000 could declare arbitrarily large pixel dimensions in the codestream, and the underlying library (`lerc.decode` / `glymur.Jp2k[:]`) materialised the full buffer before the existing post-decode size check in `_decode_strip_or_tile` ever ran.

LERC compresses constant-value blocks at >700,000:1, so a 94-byte blob requests 64 MiB; a ~1 KB on-disk tile can ask for several GB. JPEG 2000 has similar amplification through codestream-declared dimensions.

## Fix

Pre-validate each codestream's declared output against the same `expected_size * 1.05 + 1` cap the other codecs use:

- LERC: `lerc.getLercBlobInfo(blob)` returns `nCols`, `nRows`, `nBands`, and `dataType` from the header without decoding. The wrapper computes `nCols * nRows * nBands * dtype_bytes` and raises `ValueError` when it exceeds the cap.
- JPEG 2000: `glymur.Jp2k(file).shape` reads only the SIZ marker (sub-millisecond on a 2000x2000 file), so checking it before `[:]` is cheap.

Pillow's JPEG decoder already has its own `Image.MAX_IMAGE_PIXELS` guard, so no wrapper-level cap is needed for the JPEG codec.

`expected_size` is plumbed through:
- `decompress()` -> `lerc_decompress` / `jpeg2000_decompress`
- `_decode_strip_or_tile`'s LERC branch (which bypasses `decompress()` to capture the valid mask)
- `_gpu_decode.py`'s CPU fallback paths for both codecs

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_decompression_caps.py` -- 24/24 pass (6 new tests covering LERC + JPEG 2000 bomb rejection, legitimate round-trip, and backward-compat with `expected_size=0`)
- [x] `pytest xrspatial/geotiff/tests/test_compression.py xrspatial/geotiff/tests/test_lerc.py xrspatial/geotiff/tests/test_jpeg2000.py xrspatial/geotiff/tests/test_security.py` -- 89/89 pass
- [x] Full geotiff suite: 1267 pass, 3 unrelated matplotlib `RecursionError` failures in `TestPalette` plot tests (no compression code touched)

Closes #1625.


brendancol:
Addressed all four Copilot review comments in bb30f0e:

1. **`np.prod` overflow (JPEG 2000)** — replaced `int(np.prod(shape))` with `math.prod(int(d) for d in shape)`. `np.prod` multiplies in fixed-width `intp` and can wrap on attacker-declared dimensions (SIZ uses uint32 fields, so the product can exceed int64), silently under-counting `declared` and letting bombs through. `math.prod` uses Python arbitrary-precision ints.

2. **Fail-closed on unreadable SIZ marker (JPEG 2000)** — previously, an exception from `jp2.shape`/`jp2.dtype` set both to `None`, skipped the cap, and fell through to `jp2[:]` (the bomb path). The dtype fallback `getattr(jp2, 'dtype', np.uint8)` also defaulted to the smallest dtype, undercounting declared bytes. Both failure modes now raise `ValueError` before any pixel decoding runs. Added `test_jpeg2000_unreadable_shape_fails_closed` (monkeypatches `Jp2k.shape` to raise and asserts `jp2[:]` is never invoked).

3. **LERC `errCode` not checked** — `lerc.getLercBlobInfo` returns `(errCode, ...)`; when `errCode != 0` the remaining fields may be garbage. `_check_lerc_bomb` now returns early in that case so `lerc.decode` produces the canonical error.

4. **Unused `import os` in `test_jpeg2000_bomb_raises`** — removed (flake8 F401).

`pyflakes` clean. 25/25 in `test_decompression_caps.py`, 89/89 in the broader compression+security suite.

Issue #1684, opened 2026-05-12
geotiff: write_vrt nodata type hint is float-only but to_geotiff/write_geotiff_gpu accept int


brendancol:
## Summary

`xrspatial.geotiff.write_vrt` annotates its `nodata` kwarg as `float | None`:

```python
def write_vrt(vrt_path: str, source_files: list[str], *,
              relative: bool = True,
              crs_wkt: str | None = None,
              nodata: float | None = None) -> str:
```

(`xrspatial/geotiff/__init__.py:3421`)

The sibling writers `to_geotiff` and `write_geotiff_gpu` leave `nodata` untyped and document it as accepting "float, int, or None". Integer sentinels (e.g. `nodata=65535` for a uint16 raster, `nodata=-9999` for an int32 raster) are common and pass through the rest of the I/O surface, but on `write_vrt` they violate the declared type and trip strict type checkers / IDEs.

Within `_vrt.py`, the underlying `write_vrt` (`xrspatial/geotiff/_vrt.py:595`) has the same float-only annotation and is rendered into a generic `<NoDataValue>{nodata}</NoDataValue>` XML element with `str()` formatting, so an int value is rendered without complaint at runtime.

## Proposed fix

Widen the `nodata` type annotation on both the public wrapper (`xrspatial/geotiff/__init__.py:3424`) and the internal implementation (`xrspatial/geotiff/_vrt.py:598`). Either match the `to_geotiff`/`write_geotiff_gpu` style (drop the annotation, document the type in the docstring) or annotate as `float | int | None`. Update the docstring in both places to read "float, int, or None" so the surface lines up with the other writers.

Found during the geotiff API consistency sweep (Cat 3, MEDIUM).

Issue #1654, opened 2026-05-12
Public geotiff API: type annotations missing on window / path / on_gpu_failure


brendancol:
The api-consistency sweep on 2026-05-12 found one MEDIUM drift across the public `xrspatial.geotiff` surface: the same parameter is annotated on some siblings but not on others. Same kwarg, same semantics, different annotation behaviour for type-checkers and IDEs.

## Missing annotations

`window` (4-tuple `(r0, c0, r1, c1)` or None):

- `open_geotiff(source, *, dtype=None, window=None, ...)` -- no annotation
- `read_vrt(source: str, *, dtype=None, window=None, ...)` -- no annotation
- `read_geotiff_dask(source: str, *, ..., window: tuple | None = None, ...)` -- has annotation
- `read_geotiff_gpu(source: str, *, ..., window: tuple | None = None, ...)` -- has annotation

`path` (str or binary file-like):

- `to_geotiff(data, path, *, ...)` -- no annotation
- `write_geotiff_gpu(data, path, *, ...)` -- no annotation
- `write_vrt(vrt_path: str, ...)` -- annotated (VRT is path-only, no file-like)

`on_gpu_failure` and the deprecated `gpu` alias:

- `open_geotiff(... on_gpu_failure=_ON_GPU_FAILURE_SENTINEL)` -- no annotation; docstring documents `{'auto', 'strict'}`
- `read_geotiff_gpu(... on_gpu_failure=_ON_GPU_FAILURE_SENTINEL, gpu=_GPU_DEPRECATED_SENTINEL)` -- both unannotated; the deprecated `gpu` alias accepts the same string values

The sentinel default does not block an annotation: the documented and accepted user values are `'auto'` and `'strict'`, both `str`.

## Severity

MEDIUM. These are public kwargs with documented value sets. Today `mypy --strict` and IDE autocomplete cannot validate user code against them, even though sibling functions in the same module do annotate them.

## Not a breaking change

Annotation-only -- no runtime behaviour changes, no default changes, no kwarg renames. Compatible with existing callers because annotations are advisory and runtime accepts the same values.

## Fix plan

Add annotations to the public signatures:

- `window: tuple | None = None` on `open_geotiff` and `read_vrt`
- `path: str | typing.BinaryIO = ...` (or equivalent) on `to_geotiff` and `write_geotiff_gpu`. `write_vrt.vrt_path: str` stays as-is because VRT writes are path-only.
[...]

PR #1707, opened 2026-05-12
Pin nodata + streaming_buffer_bytes annotations on writer trio (#1705)

### LINKED ISSUE #1705: geotiff: nodata + streaming_buffer_bytes type-hint drift across writer trio
## Summary

Follow-up to #1654 (annotate-parity sweep): two more drift sites surfaced by the api-consistency sweep on 2026-05-12. Both are type-hint gaps where the same parameter has an explicit annotation on one public function in the writer trio and bare `=None` / different breadth on its sibling. Type-checkers and IDE autocomplete therefore see different types for the same parameter depending on which entry point the caller picked.

### 1. `nodata`: annotated on `write_vrt`, unannotated on `to_geotiff` and `write_geotiff_gpu`

`xrspatial/geotiff/__init__.py`

```python
def to_geotiff(data, path, *, crs=..., nodata=None, ...):
    """... nodata : float, int, or None ..."""

def write_geotiff_gpu(data, path, *, crs=..., nodata=None, ...):
    """... nodata : float, int, or None ..."""

def write_vrt(vrt_path, source_files, *, relative=True, crs_wkt=None,
              nodata: float | int | None = None) -> str:
    ...
```

`write_vrt` got the `float | int | None` annotation in #1684. The two sibling writers still expose `nodata=None` with no annotation, even though every docstring in the trio describes the same accepted-type set ("float, int, or None"). `mypy --strict` therefore infers `Any` for `to_geotiff(..., nodata=...)` and `write_geotiff_gpu(..., nodata=...)` while inferring `float | int | None` for `write_vrt(..., nodata=...)`.

### 2. `streaming_buffer_bytes`: `int` on `to_geotiff`, `int | None` on `write_geotiff_gpu`

`xrspatial/geotiff/__init__.py`

```python
def to_geotiff(..., streaming_buffer_bytes: int = 256 * 1024 * 1024, ...):

def write_geotiff_gpu(..., streaming_buffer_bytes: int | None = None):
```

`write_geotiff_gpu`'s implementation immediately does `del streaming_buffer_bytes` (the kwarg is accepted only for API parity with `to_geotiff`; the GPU writer has no streaming concept). The default and type annotation differ from `to_geotiff` even though the documented purpose and accepted-value set are the same. Either both should be `int = <default>` (the CPU writer's contract) or both `int | None = None` (the GPU writer's no-op contract). Matching `to_geotiff`'s `int` annotation is the cheaper fix: users who pass through the same kwargs to either entry point get the same type signature, and the GPU writer's no-op semantics are unchanged by tightening the type alone.

## Severity

MEDIUM (Cat 3, type hint drift). Annotation-only; no runtime behaviour change. Same shape as the gap #1654 closed for `window`, `path`, and `on_gpu_failure`.

## Proposed fix

1. Add `nodata: float | int | None = None` to `to_geotiff` and `write_geotiff_gpu`.
2. Change `write_geotiff_gpu`'s `streaming_buffer_bytes` annotation from `int | None = None` to `int = 256 * 1024 * 1024`, matching `to_geotiff`. The existing `del streaming_buffer_bytes` no-op stays.
3. Extend `tests/test_signature_annotations_1654.py` with pinned-annotation assertions for both kwargs across the writer trio so future drift fails CI.

Non-breaking. `nodata=None` and `streaming_buffer_bytes=256*1024*1024` are the existing defaults and the existing call sites in the codebase already match those types.

## Discovered by

`/sweep-api-consistency` against the `geotiff` module on 2026-05-12.

brendancol:
## Summary

Follow-up to #1654. The api-consistency sweep on 2026-05-12 found two more annotation gaps across `xrspatial.geotiff`'s writer trio.

- `nodata`: `write_vrt` got `float | int | None` in #1684, but `to_geotiff` and `write_geotiff_gpu` still exposed bare `nodata=None`. Type-checkers inferred `Any` on the eager writers while inferring `float | int | None` on the VRT writer, even though all three docstrings describe the same accepted-type set.
- `streaming_buffer_bytes`: `int = 256*1024*1024` on `to_geotiff` vs `int | None = None` on `write_geotiff_gpu`. The GPU writer no-ops the kwarg (`del streaming_buffer_bytes` on entry) so the type signature was the only consistency dimension. Pin both to `int = 256*1024*1024` so a caller forwarding the same kwargs to either entry point sees the same hint and default.

Annotation-only change; no runtime behaviour, defaults (effective for `to_geotiff` and the GPU no-op), or kwarg renames. The GPU writer's `del`-on-entry remains, so the value continues to be ignored on the GPU path.

Closes #1705.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_signature_annotations_1705.py` covers the new pinned-annotation assertions plus a `to_geotiff(nodata=<int>)` round-trip smoke test and a no-op assertion on the GPU writer's `streaming_buffer_bytes` kwarg.
- [x] `pytest xrspatial/geotiff/tests/test_signature_annotations_1654.py` still passes (the #1654 annotations are unchanged).
- [x] `pytest xrspatial/geotiff/tests/test_signature_parity_1631.py` still passes (`__all__` and `inspect.signature` contracts unaffected).
- [x] `pytest xrspatial/geotiff/tests/test_features.py -k all` still passes.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -1000,7 +1000,7 @@ def _extract_rich_tags(attrs: dict) -> dict:
 def to_geotiff(data: xr.DataArray | np.ndarray,
                path: str | BinaryIO, *,
                crs: int | str | None = None,
-               nodata=None,
+               nodata: float | int | None = None,
                compression: str = 'zstd',
                compression_level: int | None = None,
                tiled: bool = True,
@@ -2854,7 +2854,7 @@ def read_geotiff_gpu(source: str, *,
 def write_geotiff_gpu(data: xr.DataArray | cupy.ndarray | np.ndarray,
                       path: str | BinaryIO, *,
                       crs: int | str | None = None,
-                      nodata=None,
+                      nodata: float | int | None = None,
                       compression: str = 'zstd',
                       compression_level: int | None = None,
                       tiled: bool = True,
@@ -2865,7 +2865,7 @@ def write_geotiff_gpu(data: xr.DataArray | cupy.ndarray | np.ndarray,
                       overview_resampling: str = 'mean',
                       bigtiff: bool | None = None,
                       max_z_error: float = 0.0,
-                      streaming_buffer_bytes: int | None = None) -> None:
+                      streaming_buffer_bytes: int = 256 * 1024 * 1024) -> None:
     """Write a CuPy-backed DataArray as a GeoTIFF with GPU compression.
 
     Tiles are extracted and compressed on the GPU via nvCOMP, then
@@ -2960,10 +2960,12 @@ def write_geotiff_gpu(data: xr.DataArray | cupy.ndarray | np.ndarray,
         does not implement LERC (nvCOMP has no LERC backend), so any
         non-zero value raises ``ValueError``. Accepted at the signature
         level for API parity with ``to_geotiff``.
-    streaming_buffer_bytes : int or None
+    streaming_buffer_bytes : int
         Accepted for API parity with ``to_geotiff``. The GPU writer
         materialises the entire array on device and has no streaming
-        concept, so this kwarg is a no-op.
+        concept, so this kwarg is a no-op. Default matches
+        ``to_geotiff`` (256 MB) so callers passing the same kwargs to
+        either entry point see the same default and the same type.
     """
     if not tiled:
         raise ValueError(
--- a/xrspatial/geotiff/tests/test_signature_annotations_1705.py
+++ b/xrspatial/geotiff/tests/test_signature_annotations_1705.py
@@ -1,0 +1,142 @@
+"""Regression test for #1705: writer-trio nodata / streaming_buffer_bytes annotations.
+
+Follow-up to #1654. The api-consistency sweep on 2026-05-12 found two
+remaining annotation gaps across the public writer trio (``to_geotiff``,
+``write_geotiff_gpu``, ``write_vrt``):
+
+* ``nodata`` -- annotated as ``float | int | None`` on ``write_vrt``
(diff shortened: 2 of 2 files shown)


brendancol:
Addressed the four Copilot review comments in f06720b:

1. **`test_signature_annotations_1705.py:110` (nodata int-vs-float)** -- relaxed to `float(r.attrs['nodata']) == -9999.0` with a comment pointing at `_geotags.extract_geo_info` which does `nodata = float(nodata_str)` on read.

2. **`test_signature_annotations_1705.py:142` (file-size assertion)** -- replaced `os.path.getsize(p1) == os.path.getsize(p2)` with `p1.read_bytes() == p2.read_bytes()`. The writer is deterministic for a fixed input so byte equality is the right contract and a regression producing different bytes at the same size still trips.

3. **`test_signature_annotations_1705.py:127` (GPU skip)** -- switched to the geotiff test-suite convention: `cupy = pytest.importorskip('cupy')` followed by `if not cupy.cuda.is_available(): pytest.skip('cupy + CUDA required')`. Matches `test_predictor3_big_endian.py`, `test_dask_cupy_combined.py`, and friends.

4. **`xrspatial/geotiff/__init__.py:2868` (streaming_buffer_bytes default change)** -- intentionally **not** reverted. The PR's stated purpose is annotation + default parity with `to_geotiff` so callers forwarding the same kwargs to either entry point see the same signature via `inspect.signature`. The kwarg is a runtime no-op (`del streaming_buffer_bytes` on entry) so the surface default change has no behavioural effect. The test at `test_write_geotiff_gpu_streaming_buffer_bytes_annotated` explicitly pins both the annotation (`'int'`) and the default (`256 * 1024 * 1024`).

All 7 tests in `test_signature_annotations_1705.py` pass locally, including the GPU runtime no-op on cupy + CUDA.

PR #1755, opened 2026-05-12
geotiff: support fsspec URIs in read_geotiff_dask (#1749)

### LINKED ISSUE #1749: geotiff: read_geotiff_dask does not support fsspec / cloud URIs
**Describe the bug**

`read_geotiff_dask()` falls through to `_read_geo_info()` for non-HTTP sources, and `_read_geo_info()` uses plain `open(source, 'rb')` at `xrspatial/geotiff/__init__.py:480`. That call cannot resolve fsspec URIs (`s3://`, `gs://`, `az://`, `memory://`), so the dask path raises `FileNotFoundError` before any chunk task is built.

The eager path (`open_geotiff` -> `_read_to_array` -> `_make_source` in `_reader.py`) does route through `_make_source` and handles fsspec URIs via `_CloudSource`. The dask metadata read does not.

**Reproduction**

```python
import fsspec, xrspatial.geotiff as xg

# Put a TIFF in fsspec's memory filesystem
fs = fsspec.filesystem('memory')
with open('local.tif', 'rb') as f:
    fs.pipe_file('/test.tif', f.read())

xg.open_geotiff('memory:///test.tif')               # works
xg.open_geotiff('memory:///test.tif', chunks=2)     # FileNotFoundError
```

**Expected behavior**

Dask reads should support the same source schemes as eager reads. `_read_geo_info` should route through `_make_source` / `_is_fsspec_uri` so cloud and memory URIs resolve identically.

**References**

- `xrspatial/geotiff/__init__.py:1955` (dask path call site)
- `xrspatial/geotiff/__init__.py:480` (offending `open()` call)
- `xrspatial/geotiff/_reader.py:1020` (`_make_source` with fsspec branch)

brendancol:
## Summary

- `read_geotiff_dask` failed on `s3://`, `gs://`, `az://`, `memory://` and other fsspec URIs with `FileNotFoundError`. The eager path already handled cloud URIs via `_read_to_array` + `_CloudSource`, but the dask path's metadata-only step (`_read_geo_info` in `xrspatial/geotiff/__init__.py`) used a plain `open(source, 'rb')` call.
- Per-chunk pixel reads in the dask graph were already cloud-aware (they go through `_read_to_array` which dispatches to `_CloudSource`), so only the upfront metadata read needed fixing.
- Route fsspec URIs in `_read_geo_info` through `_CloudSource.read_all()`. The local-path `mmap` fast path is unchanged. The HTTP path (which uses `_parse_cog_http_meta`) is unchanged.

Closes #1749.

## Test plan

- [x] New `test_dask_path_fsspec_uri_1749` in `TestCloudStorage` writes a small TIFF, copies it into the fsspec `memory://` filesystem, and verifies `open_geotiff('memory:///...', chunks=4)` returns a dask-backed DataArray whose values match both the eager read and the original array.
- [x] Full `TestCloudStorage` class still passes (6/6).
- [x] Wider `xrspatial/geotiff/tests/` suite shows the same pre-existing failures as `main` (GPU/matplotlib unrelated to this change).

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -460,7 +460,7 @@ def _read_geo_info(source, *, overview_level: int | None = None):
     from ._dtypes import resolve_bits_per_sample, tiff_dtype_to_numpy
     from ._geotags import extract_geo_info_with_overview_inheritance
     from ._header import parse_all_ifds, parse_header, select_overview_ifd
-    from ._reader import _coerce_path, _is_file_like
+    from ._reader import _CloudSource, _coerce_path, _is_file_like, _is_fsspec_uri
 
     source = _coerce_path(source)
     if _is_file_like(source):
@@ -476,6 +476,17 @@ def _read_geo_info(source, *, overview_level: int | None = None):
             source.seek(cur)
         except (OSError, AttributeError):
             pass
+        close_data = False
+    elif isinstance(source, str) and _is_fsspec_uri(source):
+        # fsspec URI (s3://, gs://, az://, memory://, ...): pull the
+        # whole file via _CloudSource for metadata parsing. Per-chunk
+        # pixel reads in the dask graph go through _read_to_array
+        # which opens its own _CloudSource, so this fetch is metadata-only.
+        _src = _CloudSource(source)
+        try:
+            data = _src.read_all()
+        finally:
+            _src.close()
         close_data = False
     elif isinstance(source, str):
         with open(source, 'rb') as f:
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -460,9 +460,33 @@ def _read_geo_info(source, *, overview_level: int | None = None):
     from ._dtypes import resolve_bits_per_sample, tiff_dtype_to_numpy
     from ._geotags import extract_geo_info_with_overview_inheritance
     from ._header import parse_all_ifds, parse_header, select_overview_ifd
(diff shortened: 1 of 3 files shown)


brendancol:
Fixed in 16b76ea. fsspec URIs now go through `_parse_cog_http_meta` (the bounded range-prefetch parser used for HTTP COGs) instead of `_CloudSource.read_all`. Three changes:

- `_read_geo_info`: fsspec branch switched from `read_all()` to `_parse_cog_http_meta(_CloudSource(source))`. Metadata reads stay capped at `MAX_HTTP_HEADER_BYTES`.
- `read_geotiff_dask`: the per-graph metadata prefetch now treats fsspec URIs the same as HTTP URIs (one parse, wrapped in a `dask.delayed` so all chunk tasks share it).
- `_delayed_read_window._read`: per-chunk reads for fsspec sources dispatch through `_fetch_decode_cog_http_tiles` with a `_CloudSource`, so each chunk issues range GETs (and coalesces neighbours) instead of pulling the whole file. Added `read_ranges` and `read_ranges_coalesced` to `_CloudSource` to match the `_HTTPSource` interface that the tile decoder expects.

Test `test_dask_path_fsspec_uri_1749` still passes, and a new `test_dask_path_fsspec_uri_no_full_download_1749` monkeypatches `_CloudSource.read_all` to raise and confirms `open_geotiff(..., chunks=4)` plus `.compute()` succeeds, proving the bounded-prefetch path is taken end-to-end.

PR #1757, opened 2026-05-12
geotiff: validate tile_size and chunks size args (#1752)

### LINKED ISSUE #1752: geotiff: validate tile_size and dask chunks at entry points
**Describe the bug**

Two public entry points accept size parameters without checking that they are positive integers:

- `to_geotiff(tiled=True, tile_size=0)` reaches `xrspatial/geotiff/_writer.py:648` where `math.ceil(width / tw)` divides by zero.
- `read_geotiff_dask(chunks=0)` or `chunks=(0, N)` unpacks at `xrspatial/geotiff/__init__.py:2051` and feeds the value to dask chunk math; negative or zero values produce confusing downstream errors.

This is not subtle data corruption, but the module aims to be watertight at the API surface.

**Expected behavior**

- `to_geotiff` should raise `ValueError` when `tile_size <= 0`.
- `read_geotiff_dask` should raise `ValueError` when any chunk dimension `<= 0`.

Both messages should name the offending parameter and the invalid value.

**References**

- `xrspatial/geotiff/_writer.py:648`
- `xrspatial/geotiff/__init__.py:2051`

brendancol:
## Summary

- `to_geotiff(..., tiled=True, tile_size=0)` reached the tiled writer where `math.ceil(width / tile_size)` raised `ZeroDivisionError` with no indication that `tile_size` was the bad input. Negative values produced a nonsensical tile grid.
- `read_geotiff_dask(chunks=0)` (or `chunks=(0, N)`) propagated zero into dask's chunk math and surfaced as a confusing `range()` / empty-chunks error.
- Add up-front validation at both entry points. Reject non-positive ints, non-int values, and malformed `chunks` tuples with a `ValueError` that names the parameter and the bad value. `tiled=False` skips the `tile_size` check since the writer ignores `tile_size` in strip mode.
- Style matches the existing `window=` validator in `read_geotiff_dask`.

Closes #1752.

## Test plan

- [x] New `test_size_param_validation_1752.py` (11 tests): `tile_size` 0 / negative / non-int rejected, `tile_size=1` still writes; `chunks` 0 / negative / `(0, N)` / `(N, -1)` / 3-tuple rejected; `chunks=256` and `chunks=(4, 8)` still read.
- [x] Verified tests fail on main (8 of the 11 raise wrong error / no error) and pass with the fix.
- [x] Full geotiff suite: 2132 passed; the 12 pre-existing failures (matplotlib `RecursionError` in `test_features.py`, GPU `test_predictor2_big_endian_gpu_1517.py`, GPU `test_no_georef_windowed_coords_1710.py`) reproduce on main without this change.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -1239,6 +1239,21 @@ def to_geotiff(data: xr.DataArray | np.ndarray,
 
     path = _coerce_path(path)
 
+    # Reject non-positive tile_size up front. The tiled writer computes
+    # the tile grid as ``math.ceil(width / tile_size)``; tile_size=0 hits
+    # ZeroDivisionError deep inside the writer, and negative values
+    # produce a nonsensical tile grid. tiled=False ignores tile_size, so
+    # only validate when tiled output is actually requested.
+    if tiled:
+        if not isinstance(tile_size, (int, np.integer)) or isinstance(
+                tile_size, bool):
+            raise ValueError(
+                f"tile_size must be a positive int, got "
+                f"{tile_size!r} (type {type(tile_size).__name__}).")
+        if tile_size <= 0:
+            raise ValueError(
+                f"tile_size must be a positive int, got tile_size={tile_size}.")
+
     # Up-front validation: catch bad compression names before they reach
     # any of the deeper write paths (streaming, GPU, VRT, COG) where the
     # error surfaces from _compression_tag with a less obvious traceback.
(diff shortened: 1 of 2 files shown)


brendancol:
Fixed in beb05ec. The scalar branch now accepts `np.integer` (matching the tuple-element check) and coerces to plain `int` so the downstream `isinstance(chunks, int)` sites keep working. Added two regression tests covering `chunks=np.int64(256)` and `chunks=(np.int64(256), 256)`.

PR #1760, opened 2026-05-12
geotiff: honour ModelTiepointTag when ModelPixelScaleTag is absent (#1750)

### LINKED ISSUE #1750: geotiff: tiepoint-only georef silently drops origin (reports has_georef=True at 0,0)
**Describe the bug**

`_extract_transform()` at `xrspatial/geotiff/_geotags.py:466-468` handles the case where a TIFF carries a `ModelTiepointTag` but no `ModelPixelScaleTag`. The current branch is:

```python
# Tiepoint without scale: still flag as georeferenced (origin known)
if tiepoint is not None:
    return GeoTransform(), True
```

That returns the default `GeoTransform()` (origin `(0, 0)`, unit pixels) with `has_georef=True`. The tiepoint's actual model X/Y is discarded. Downstream coordinate construction trusts `has_georef=True` and produces a real coordinate axis with the wrong origin, silently relocating the raster.

**Expected behavior**

The tiepoint encodes a real origin. Either:

- Honour the tiepoint origin: use `tp_x - tp_i * 1.0` and `tp_y + tp_j * 1.0` as origin with `pixel_width = 1.0`, `pixel_height = -1.0`, documenting the unit-pixel-scale assumption; or
- Set `has_georef=False` so downstream code falls back to the integer-pixel coord path used for non-georef files.

Either is acceptable. A silent (0, 0) origin is not.

**References**

- `xrspatial/geotiff/_geotags.py:466`

brendancol:
## Summary

- `_extract_transform()` returned a default `GeoTransform()` with origin (0, 0) and `has_georef=True` whenever a TIFF carried ModelTiepointTag but no ModelPixelScaleTag, silently dropping the tiepoint's real-world X / Y. Downstream coordinate construction then placed the raster at (0, 0) with no warning.
- Use the tiepoint X / Y to compute the origin (matching the tiepoint-with-scale branch) and fall back to the GeoTIFF spec convention of unit pixel scale (1.0, -1.0) when ModelPixelScaleTag is absent.
- Considered returning `has_georef=False` instead. Rejected because the tiepoint encodes a real model X / Y; flagging the raster as ungeoreferenced would still mislabel it.

Closes #1750.

## Test plan

- [x] New `TestTiepointWithoutScale_1750` in `xrspatial/geotiff/tests/test_geotags.py` builds a TIFF with ModelTiepointTag only, asserts the extracted origin equals tp_x / tp_y and that pixel scale falls back to (1.0, -1.0). Covers both tp_i=tp_j=0 and non-zero tp_i / tp_j cases.
- [x] Confirmed the new tests fail on main (origin_x = 0.0 vs expected tiepoint X) and pass with the fix.
- [x] Existing `xrspatial/geotiff/tests/` suite delta is identical to main; pre-existing GPU / matplotlib failures are unrelated to this change.

### DIFF
--- a/xrspatial/geotiff/_geotags.py
+++ b/xrspatial/geotiff/_geotags.py
@@ -463,9 +463,29 @@ def _extract_transform(ifd: IFD) -> tuple[GeoTransform, bool]:
 
         return GeoTransform(pixel_width=sx, pixel_height=-sy), True
 
-    # Tiepoint without scale: still flag as georeferenced (origin known)
+    # Tiepoint without scale: honour the tiepoint origin and fall back to
+    # unit pixel size.  Per the GeoTIFF spec a ModelTiepointTag encodes a
+    # real-world (X, Y) for pixel (I, J); dropping it would silently relocate
+    # the raster to (0, 0).  Unit scale (1.0, -1.0) is the documented fallback
+    # when ModelPixelScaleTag is absent.
     if tiepoint is not None:
-        return GeoTransform(), True
+        if not isinstance(tiepoint, tuple):
+            tiepoint = (tiepoint,)
+        tp_i = tiepoint[0] if len(tiepoint) > 0 else 0.0
+        tp_j = tiepoint[1] if len(tiepoint) > 1 else 0.0
+        tp_x = tiepoint[3] if len(tiepoint) > 3 else 0.0
+        tp_y = tiepoint[4] if len(tiepoint) > 4 else 0.0
+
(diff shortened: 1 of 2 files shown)


brendancol:
Addressed both review comments in 02fa09f. Reworded the fallback comment in _geotags.py and the helper docstring in test_geotags.py to distinguish the spec convention (sy as a positive magnitude) from the GeoTransform sign convention used here (pixel_height = -sy). Comment- and docstring-only; tests still pass (12/12).

PR #1761, opened 2026-05-12
geotiff: refuse VRT non-nearest resample algorithms (#1751)

### LINKED ISSUE #1751: geotiff: VRT ResampleAlg is parsed but ignored (silent nearest-neighbour fallback)
**Describe the bug**

`ComplexSource` in `read_vrt` parses `<ResampleAlg>` (`xrspatial/geotiff/_vrt.py:356-360`) and stores it on the source record. The resample step at `xrspatial/geotiff/_vrt.py:798-804` always calls `_resample_nearest()` when `SrcRect` and `DstRect` differ, regardless of the parsed algorithm.

A VRT requesting `Bilinear`, `Cubic`, `Average`, or `Mode` returns nearest-neighbour pixels with no warning. Downstream analytics see numerically different output from GDAL with no indication of why.

**Expected behavior**

At minimum, raise `NotImplementedError` (or warn) when `resample_alg` is set to anything other than `Nearest` / `NEAR` / `None`. The error message should name the requested algorithm and point at this issue.

If implementation scope allows, add bilinear and cubic resampling.

**References**

- `xrspatial/geotiff/_vrt.py:356` (parse site)
- `xrspatial/geotiff/_vrt.py:798` (resample call site)

brendancol:
## Summary

- Closes #1751. `ComplexSource` parses `<ResampleAlg>` from VRT XML (`_vrt.py:356-360`) but `read_vrt` always calls `_resample_nearest` at the placement step (`_vrt.py:798-804`), so a VRT declaring `Bilinear` / `Cubic` / `CubicSpline` / `Lanczos` / `Average` / `Mode` silently received nearest-sampled pixels mislabelled as the requested algorithm -- quietly wrong analytics.
- Raise `NotImplementedError` at the resample call site when `src.resample_alg` is anything other than nearest. The case-insensitive accept list is `''`, `None`, `Nearest`, `NearestNeighbour`, `NearestNeighbor`, `NEAR`.
- The check sits inside the `needs_resample` branch rather than at parse time so a `ComplexSource` declaring `Bilinear` with matching SrcRect/DstRect sizes is still accepted -- no resample kernel runs, so no mislabelled output.
- Implementing the higher-quality kernels is out of scope for this PR; the goal is to stop the silent wrong-numbers behaviour. Error message names the requested algorithm and points at issue #1751 so a user can re-export with `Nearest` or wait for follow-up work.

## Test plan

- [x] New `test_vrt_resample_alg_1751.py` (16 cases): each of Bilinear/Cubic/CubicSpline/Lanczos/Average/Mode raises with the algorithm name and `1751` in the message; case-insensitive `bilinear` also raises; Nearest / NearestNeighbour / NearestNeighbor / NEAR / nearest / NEAREST / empty / absent `<ResampleAlg>` all round-trip; `Bilinear` at matching SrcRect/DstRect sizes does NOT raise (pins down the resample-site placement).
- [x] Confirmed all 16 tests fail on `main` without the source change and pass with it.
- [x] `pytest xrspatial/geotiff/tests/ -q` (1992 pass, 3 skipped; the GPU `test_predictor2_big_endian_gpu_1517` and `test_no_georef_windowed_coords_1710` failures are pre-existing environment issues, identical on main).

### DIFF
--- a/xrspatial/geotiff/_vrt.py
+++ b/xrspatial/geotiff/_vrt.py
@@ -151,9 +151,11 @@ class _Source:
     # GDAL ``<ComplexSource><ResampleAlg>`` values are ``NearestNeighbour``
     # (default), ``Bilinear``, ``Cubic``, ``CubicSpline``, ``Lanczos``,
     # ``Average``, ``Mode``.  Only nearest-neighbour is implemented in
-    # the placement step (issue #1694); the resample site raises
-    # ``NotImplementedError`` for any other declared algorithm rather
-    # than silently substituting nearest (issue #1751).  Higher-quality
+    # the placement step (issue #1694).  When ``SrcRect`` and ``DstRect``
+    # sizes differ (resampling actually required), the read raises
+    # ``NotImplementedError`` for any non-nearest algorithm rather than
+    # silently substituting nearest (issue #1751); a non-nearest alg with
+    # matching rect sizes is a no-op and passes through.  Higher-quality
     # resamplers are tracked for follow-up.
     resample_alg: str | None = None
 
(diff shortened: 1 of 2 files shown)


brendancol:
Clarified both comments in 4bf143f. They now state that the NotImplementedError fires only when SrcRect and DstRect sizes differ (resampling actually required), and that a non-nearest alg with matching rect sizes is a no-op that passes through.

Issue #1708, opened 2026-05-12
geotiff: read_to_array leaks into public namespace but is not in __all__ or docs


brendancol:
## Summary

`xrspatial.geotiff.__init__.py` imports `read_to_array` from `._reader` at module level (line 46) so that the eager `open_geotiff` body can call it. This brings the name into the `xrspatial.geotiff` public namespace as a side-effect, but `read_to_array` is not in `__all__` and is not advertised in the module-level "Public API" docstring.

```python
# xrspatial/geotiff/__init__.py
from ._reader import UnsafeURLError, read_to_array  # line 46

__all__ = [
    'GeoTIFFFallbackWarning',
    'UnsafeURLError',
    'open_geotiff',
    'read_geotiff_gpu',
    'read_geotiff_dask',
    'read_vrt',
    'to_geotiff',
    'write_geotiff_gpu',
    'write_vrt',
]
```

`from xrspatial.geotiff import read_to_array` succeeds today. The library's own test in `xrspatial/geotiff/tests/test_band_validation_1673.py` relies on that import six times (lines 49, 58, 68, 77, 86, 95, 113) -- proof that the leaked name is being used as if public. External users who pattern-match against the test will also import it from the public namespace. If we later promote `read_to_array` to private by adding an underscore or by removing the top-level re-export, those users break with no deprecation path.

This is the orphan-API pattern flagged by the api-consistency sweep (Cat 5): a function name that walks like a public API (importable from the package's namespace) but does not appear in `__all__`, the docstring's Public API list, or the user docs.

## Severity

MEDIUM (Cat 5, public API surface drift). Not a correctness bug; it's a hidden surface that constrains future cleanup.

## Proposed fix

Hide the leak. The internal usage in `__init__.py` keeps working, but the name disappears from the public namespace:

```python
# xrspatial/geotiff/__init__.py
from ._reader import UnsafeURLError
from ._reader import read_to_array as _read_to_array
```

Then update each internal reference (lines around 686, 2533) to `_read_to_array`, and update the single test file `test_band_validation_1673.py` to import from `xrspatial.geotiff._reader` directly:

```python
from xrspatial.geotiff._reader import read_to_array
```

That makes the internal-only status explicit, removes the orphan-API surface, and keeps the test surface working through the canonical module path. The change is internal-rename only; no external user code can depend on the leaked top-level name because it was never in `__all__` or the docs in the first place.

[...]

PR #1762, opened 2026-05-12
perf(geotiff): drop bytes(bytearray) copy in TIFF layout assemble (#1756)

### LINKED ISSUE #1756: geotiff: bytes(bytearray) at end of _assemble_*_layout doubles peak memory
**What**

`_assemble_standard_layout` and `_assemble_cog_layout` in `xrspatial/geotiff/_writer.py` build up the output TIFF file in a `bytearray`, then return `bytes(output)` at the end. The `bytes(...)` call copies the entire bytearray contents, transiently doubling peak memory.

Lines:

- `xrspatial/geotiff/_writer.py:1035` (`_assemble_standard_layout`)
- `xrspatial/geotiff/_writer.py:1118` (`_assemble_cog_layout`)

**Why it matters**

The eager (non-streaming) writer materializes the full output buffer in memory. The bytearray-to-bytes conversion at the end of the assembly path doubles peak memory transiently. Measured:

```
output (bytearray) size: 95.37 MB
bytes(output) peak: 95.37 MB extra
```

So a 1 GB TIFF write adds 1 GB of peak memory beyond the bytearray itself. For users writing large COGs from memory, this can push a borderline-OK write into OOM territory.

**Fix**

Return the `bytearray` directly. `_write_bytes` calls `f.write(file_bytes)` which accepts any buffer-protocol object, and the post-write `parse_header(file_bytes[:16])` validation also accepts bytearray slicing. The `bytes(...)` call is a pure copy with no contract benefit.

The streaming writer (`write_streaming`) already writes pixel data directly to the file handle and does not have this issue.

**Reproduction**

```python
import io, numpy as np, tracemalloc
import xarray as xr

arr = xr.DataArray(np.random.randint(0, 255, (10000, 10000), dtype=np.uint8), dims=['y', 'x'])
out = io.BytesIO()
tracemalloc.start()
to_geotiff(arr, out, compression='none', tiled=False)
_, peak = tracemalloc.get_traced_memory()
tracemalloc.stop()
print(f"output size {len(out.getvalue())/(1024*1024):.1f} MB, peak {peak/(1024*1024):.1f} MB")
```

Observed: peak ~3x output size. After fix: peak ~2x output size.

**Severity**

MEDIUM. Real measurable memory overhead, easy fix, narrow blast radius.

brendancol:
## Summary

- `_assemble_standard_layout` and `_assemble_cog_layout` build the output TIFF in a `bytearray`, then ended with `return bytes(output)` which copies the entire buffer
- Returning the `bytearray` directly drops transient peak memory in half on eager writes; downstream consumers (`_write_bytes`, the `parse_header` validation slice, `BytesIO.write`, file handle `write`) all accept the buffer protocol so no caller changes are needed
- Type annotations on `_assemble_tiff`, both layout helpers, and `_write_bytes` are updated to reflect that the return type is now `bytearray`

Fixes #1756.

## Measurements

On a 10000x10000 uint8 raster (95 MB output):

```
Before: peak Python-allocated = 202 MB
After:  peak Python-allocated = 107 MB
```

The savings scale linearly with output size: a 1 GB write drops 1 GB of peak memory, a 10 GB write drops 10 GB.

## Test plan

- [x] 6 new regression tests in `test_assemble_layout_no_bytes_copy_1756.py` cover the `bytearray` return type at the layout, assembler, and end-to-end levels (`BytesIO` and `tmp_path` round-trips)
- [x] 2123 existing geotiff tests pass; 10 unrelated failures (`test_no_georef_windowed_coords_1710`, `test_predictor2_big_endian_gpu_1517`) reference the now-private `read_to_array` attribute (commit 8adb749 / issue #1708) and predate this change
- [x] CPU and GPU writer paths both exercise `_assemble_tiff`; the GPU path is covered by `test_gpu_writer_attrs_1563.py` and `test_kwarg_behaviour_2026_05_12.py` (all 36 pass)


brendancol:
Addressed in ef09c2b: realigned the docstring with what the test actually checks instead of implementing the monkey-patch.

The writer module looks ``bytes`` up via ``builtins`` (not as a module-global alias), so monkey-patching ``builtins.bytes`` would affect every call site - including struct packing and slicing the test relies on - and patching only the writer's namespace would not intercept calls without invasive rebinding. The type assertions in the test already catch the specific regression at issue (``return bytes(output)`` in the assembler would flip both the top-level type and the slice type from ``bytearray`` to ``bytes``), so I renamed the test to ``test_assemble_tiff_output_is_mutable_buffer_with_valid_header`` and rewrote the docstring to describe that contract.

Issue #1710, opened 2026-05-12
open_geotiff windowed reads of non-georef TIFFs produce float64 half-pixel-shifted coords


brendancol:
## Description

A TIFF with no GeoTIFF tags (no `ModelPixelScale`, no `ModelTiepoint`, no GeoKeys) gets different `y`/`x` coordinates depending on whether `window=` is passed.

Full read takes the `has_georef=False` branch in `_geo_to_coords` and returns integer pixel coords `[0, 1, 2, ...]` with `int64` dtype.

Windowed read of the same file skips `_geo_to_coords` entirely and synthesises coords from the default `GeoTransform(origin=0, pixel_width=1, pixel_height=-1)`. That produces float64 coords like `[-0.5, -1.5, -2.5, ...]` because the `PixelIsArea` half-pixel shift gets applied to a unit transform that was never real.

Every read path has the same split: `open_geotiff`, `read_geotiff_dask`, `read_geotiff_gpu`, and dask+cupy.

## Reproduction

```python
import os, tempfile
import numpy as np
from xrspatial.geotiff import open_geotiff
from xrspatial.geotiff._writer import write

with tempfile.TemporaryDirectory() as tmp:
    arr = np.arange(64, dtype=np.float32).reshape(8, 8)
    path = os.path.join(tmp, "no_georef.tif")
    write(arr, path, compression="none", tiled=False)

    full = open_geotiff(path)
    win = open_geotiff(path, window=(0, 0, 4, 4))

    print(full.y.dtype, full.y.values[:4])  # int64 [0 1 2 3]
    print(win.y.dtype,  win.y.values)       # float64 [-0.5 -1.5 -2.5 -3.5]
```

## Expected

Windowed reads of a non-georef TIFF should match the full-read convention: integer pixel indices `[r0, r0+1, ..., r1-1]` for y and `[c0, c0+1, ..., c1-1]` for x, dtype `int64`.

The DataArray also should not carry a synthetic `attrs['transform']` for a file with no GeoTIFF tags. A non-georef file should round-trip through `to_geotiff` without picking up a fabricated identity transform.

## Affected code

- `xrspatial/geotiff/__init__.py` `open_geotiff` lines 695-707 (eager numpy windowed coord)
- `xrspatial/geotiff/__init__.py` `read_geotiff_dask` lines 1839-1864 (dask windowed coord)
[...]

PR #1763, opened 2026-05-12
geotiff: GPU stripped windowed coords honour has_georef=False (#1753)

### LINKED ISSUE #1753: geotiff: GPU stripped windowed read of non-georef TIFF emits float64 coords (regression of #1710)
## Summary

`open_geotiff(path, gpu=True, window=...)` on a TIFF with no GeoTIFF tags (no `ModelPixelScale` / `ModelTiepoint` / `GeoKeyDirectory`) returns float64 coords like `[-0.5, -1.5, ...]` synthesised from the default unit `GeoTransform`, instead of integer pixel coords `[0, 1, 2, ...]` that the eager numpy and dask paths emit.

This was fixed in #1710 for the eager numpy path (`open_geotiff`), the dask path (`read_geotiff_dask`), and the tiled GPU path (`_gpu_apply_window_band`). The stripped-fallback branch in `read_geotiff_gpu` was missed.

## Reproduction

```python
import numpy as np
from xrspatial.geotiff._writer import write

arr = np.arange(64, dtype=np.float32).reshape(8, 8)
path = "/tmp/no_georef.tif"
write(arr, path, compression='none', tiled=False)  # stripped, no georef

eager = open_geotiff(path, window=(0, 0, 4, 4))
dask = open_geotiff(path, chunks=4, window=(0, 0, 4, 4))
gpu = open_geotiff(path, gpu=True, window=(0, 0, 4, 4))

print(eager.y.dtype, eager.y.values)   # int64 [0 1 2 3]
print(dask.y.dtype, dask.y.values)     # int64 [0 1 2 3]
print(gpu.y.dtype, gpu.y.values)       # float64 [-0.5 -1.5 -2.5 -3.5]  <- WRONG
```

The eager / dask / tiled-GPU paths all guard on `geo_info.has_georef`. The stripped-GPU fallback in `read_geotiff_gpu` only checks `t is None`, which is never true for non-georef files because `_extract_transform` returns a default `GeoTransform()` instance with `has_georef=False`. The default transform has `pixel_width=1.0`, `pixel_height=-1.0`, so the PixelIsArea branch produces `(c0 + 0.5) * 1.0 + 0` and `(r0 + 0.5) * -1.0 + 0` -> `[-0.5, -1.5, ...]`.

## Affected backends

- `gpu=True` stripped reads with `window=...`
- `gpu=True, chunks=...` stripped reads with `window=...`

The tiled-GPU and CPU paths are not affected.

## Severity

HIGH (Cat 2 - coords preservation). Coord dtype changes silently between the input full-read (int64) and the windowed-read on the GPU backend (float64), breaking backend parity. Downstream coord-based math (clipping, interp, xarray slicing) gets different answers depending on which backend is active.

## Test failures

`xrspatial/geotiff/tests/test_no_georef_windowed_coords_1710.py`:
- `TestGpuWindowedCoords::test_windowed_read_integer_coords`
- `TestGpuWindowedCoords::test_dask_cupy_windowed_integer_coords`
- `TestBackendParity::test_dtype_parity_windowed`

brendancol:
## Summary

Fix backend inconsistency on `open_geotiff(path, gpu=True, window=...)` for non-georef stripped TIFFs. PR #1738 (the GPU stripped fallback that forwards `max_pixels` / `window` / `band` to the CPU helper) regressed the #1710 fix for the GPU stripped fallback path: a non-georef file read with `gpu=True, window=...` returned float64 coords synthesised from the default unit `GeoTransform` (e.g. `y=[-0.5, -1.5, -2.5, -3.5]`), while the eager numpy, dask, and tiled GPU paths returned integer pixel coords (`np.arange(c0, c1, dtype=int64)`).

The fix routes the `has_georef=False` case through the integer-coord branch on the GPU stripped fallback, mirroring the eager and dask paths.

## Root cause

The windowed coord branch added in #1738 only checks `if t is None:`. That is unreachable because `_extract_transform` always returns a default `GeoTransform()` instance with `has_georef=False`. The default transform has `pixel_width=1.0`, `pixel_height=-1.0`, so the PixelIsArea sub-branch produced `(c0 + 0.5) * 1.0 + 0` and `(r0 + 0.5) * -1.0 + 0` -> `[-0.5, -1.5, ...]`.

`dask+cupy` shares the same code path, so the bug fired on every `gpu=True` windowed read of a non-georef stripped TIFF regardless of chunking.

## Fix

- Route `not getattr(geo_info, 'has_georef', True)` through the integer-coord branch alongside the existing `t is None` check at `__init__.py:2724-2747`.
- Mirror the eager numpy path's windowed coord shortcut at `open_geotiff` lines 834-836.
- Also fix a latent off-by-one in the integer-coord branch: was `arange(r1-r0)` instead of `arange(r0, r1)`, so an offset window like `(2, 3, 6, 7)` would have returned `[0, 1, 2, 3]` instead of `[2, 3, 4, 5]`. The branch was unreachable so the off-by-one was latent.

## Tests

14 new regression tests in `test_gpu_stripped_no_georef_window_1753.py`:
- `TestStrippedGpuWindowedNoGeoref`: origin window, offset window, dask+cupy variants, uint16 dtype, no transform attr
- `TestStrippedGpuWindowedBackendParity`: dtype-parity (3 windows) and value-parity (3 windows) across all four backends
- `TestStrippedGpuWindowedGeorefStillWorks`: PixelIsArea origin and offset windows still get float64 coords with the correct half-pixel shift

## Test plan

- [x] All 14 new regression tests pass.
- [x] All 15 existing tests in `test_no_georef_windowed_coords_1710.py` pass (3 previously failed in `TestGpuWindowedCoords` and `TestBackendParity`).
- [x] All 2005 non-stale geotiff tests still pass.
- [x] 10 pre-existing failures in `test_predictor2_big_endian_gpu_1517.py` are unrelated stale test-infrastructure from PR #1708 (renamed `read_to_array` to `_read_to_array`).

Closes #1753


brendancol:
Added matching x-axis dtype parity assertions in `test_dtype_parity` (commit ddc18a4): `eager.x.dtype == dask.x.dtype == gpu.x.dtype == gpu_dk.x.dtype` and `eager.x.dtype == np.int64`. All 14 tests in the file still pass locally.

brendancol:
Rebased onto main after #1759 landed the same code fix. The PR now contains only the test file (`test_gpu_stripped_no_georef_window_1753.py`) and a sweep state CSV bump — no `__init__.py` change. The test file is kept as additional defense-in-depth coverage for the stripped GPU fallback (parametrized parity, uint16 variant, dask+cupy offset window). All 14 tests pass locally.

If reviewers prefer to drop the duplicate test coverage instead, this PR can be closed without merging — the fix is already on main.

Issue #1632, opened 2026-05-11
open_geotiff drops user-defined CRS WKT on read (crs/crs_wkt not set, only crs_name)


brendancol:
## What happened

`open_geotiff` reads a GeoTIFF whose CRS is user-defined (WKT stored in the GeoTIFF citation, GeoKey `GEOKEY_PROJECTED_CS_TYPE == 32767`) and emits the WKT under `attrs['crs_name']` only. Neither `attrs['crs']` nor `attrs['crs_wkt']` is populated. `to_geotiff` ignores `attrs['crs_name']`, so a read -> write round-trip silently drops the CRS.

Users who only read files see the WKT survive (as `crs_name`). Pipelines that read, process, and write a user-defined CRS GeoTIFF lose the projection on the output. Files with a recognized EPSG (e.g. EPSG:4326) are unaffected because the EPSG resolves to WKT via pyproj and the read path emits both `attrs['crs']` and `attrs['crs_wkt']`.

## Repro

```python
import numpy as np
import xarray as xr
import tifffile
from xrspatial.geotiff import open_geotiff, to_geotiff

WKT = (
    'PROJCS["User defined LCC",GEOGCS["NAD83",DATUM["North American Datum 1983",'
    'SPHEROID["GRS 1980",6378137,298.257222101]],PRIMEM["Greenwich",0],'
    'UNIT["degree",0.0174532925199433]],PROJECTION["Lambert Conformal Conic 2SP"],'
    'PARAMETER["central_meridian",-100],PARAMETER["latitude_of_origin",40],'
    'PARAMETER["standard_parallel_1",30],PARAMETER["standard_parallel_2",50],'
    'UNIT["metre",1]]'
)

arr = np.ones((4, 4), dtype=np.float32)
da = xr.DataArray(arr, dims=['y', 'x'],
                  coords={'y': np.linspace(50, 47, 4),
                          'x': np.linspace(10, 13, 4)},
                  attrs={'crs_wkt': WKT})
to_geotiff(da, "src.tif", compression="none")

rd = open_geotiff("src.tif")
print("crs:",      rd.attrs.get("crs"))      # None
print("crs_wkt:",  rd.attrs.get("crs_wkt"))  # None
print("crs_name:", rd.attrs.get("crs_name"))  # WKT survives here

to_geotiff(rd, "dst.tif", compression="none")
with tifffile.TiffFile("dst.tif") as tif:
[...]

PR #1772, opened 2026-05-13
geotiff: warn on WKT-only CRS write (#1768)

### LINKED ISSUE #1768: geotiff: WKT-only CRS writes opaque citation, hurts non-libgeotiff interop
## What

`build_geo_tags` in `xrspatial/geotiff/_geotags.py` handles a WKT-only CRS (no EPSG provided) by writing `ProjectedCSType` / `GeographicType` = 32767 (user-defined) and stuffing the raw WKT into `GTCitationGeoKey -> GeoAsciiParams`. No projection parameter GeoKeys (`ProjLinearUnitsGeoKey`, `GeogAngularUnitsGeoKey`, ellipsoid parameters, projection method, etc.) are emitted.

## Why this hurts

The citation-only WKT trick relies on every reader being willing to parse a WKT string stuffed into a citation field. libgeotiff and GDAL accept that. Most other tools (QGIS via non-GDAL paths, tifffile, in-house GeoTIFF parsers, some web map ingesters) inspect the structured GeoKeys and treat citation as a free-form name -- so they see a user-defined CRS with no parameters and fall back to either an unknown projection or a blind read with no CRS. The result is a silent interop hazard: the file appears to round-trip through `xrspatial` itself (because the read path lifts the WKT back out of the citation, issue #1632) while looking projection-less to anything else.

## Affected users

Anyone passing `attrs['crs_wkt']` (with no `attrs['crs']`) into `to_geotiff`, or anyone whose source file already had a user-defined CRS and is rewriting it. Files with EPSG codes (`attrs['crs'] = <int>`) are unaffected.

## Fix options

1. Emit the richer GeoKey set when the WKT carries enough information (parse linear/angular units, ellipsoid params, projection method out of the WKT). This is the GDAL-style write path. It's invasive and easy to get wrong -- the GeoKey machinery in `_geotags.py` would need a structured WKT parser.
2. Document the limitation in the public docstrings of `to_geotiff` and `build_geo_tags`, and emit a `UserWarning` when the WKT-only path is taken pointing users at EPSG codes as the preferred input. Converts the silent hazard into a visible one without risking regressions in the write path.

Option 2 keeps the existing citation-WKT bytes intact (so existing round-trips still work) while making the limitation discoverable. Option 1 is a follow-up.

## Reproducer

```python

da = xr.DataArray(
    np.ones((4, 4), dtype=np.float32),
    dims=['y', 'x'],
    coords={'y': np.linspace(50, 47, 4), 'x': np.linspace(10, 13, 4)},
    attrs={'crs_wkt': 'PROJCS["User defined LCC", ...]'},
)
to_geotiff(da, "out.tif", compression='none')
# Result: GeoKeyDirectory carries ProjectedCSType=32767 plus a single
# GTCitationGeoKey pointing at the WKT in GeoAsciiParams. No
# ProjLinearUnits, no GeogAngularUnits, no Projection key.
```

Read this file with a non-libgeotiff parser and the CRS is effectively lost.

brendancol:
Fixes #1768.

## Path chosen

Doc + warning. The richer-WKT path (parsing WKT into structured GeoKeys -- ProjLinearUnits, GeogAngularUnits, ellipsoid params, projection method) would require a WKT parser inside `_geotags.py` and a meaningful redesign of `build_geo_tags`. That's a bigger change that risks regressing the EPSG path. The warning path converts a silent interop hazard into a visible one without touching the bytes on disk -- existing round-trips through `to_geotiff` keep working, and callers learn at write time that their output may not survive non-libgeotiff readers.

## Summary

- `build_geo_tags` now emits a `UserWarning` when it takes the WKT-only path (`crs_wkt` supplied, `crs_epsg` is `None`). The message points callers at EPSG codes as the preferred input.
- Docstrings on `build_geo_tags`, `to_geotiff`, and `write_geotiff_gpu` describe what's emitted on the WKT-only path and recommend EPSG for interop.
- The GeoKey + GeoAsciiParams bytes are unchanged, so existing files keep round-tripping through libgeotiff / GDAL / this package.

## Tests

`xrspatial/geotiff/tests/test_wkt_only_crs_warning_1768.py`:

- `build_geo_tags` warns on the WKT-only path and the citation tags are still emitted.
- `build_geo_tags` does not warn on the EPSG path or when no CRS is supplied.
- `to_geotiff` warns end-to-end on `attrs['crs_wkt']` only, and does not warn on `attrs['crs']`.
- The citation byte-level contract from #1632 is regression-guarded.

All 35 tests in the new file plus the existing #1632 suite pass. Broader geotiff smoke (writer, geotags, writer_matrix, write_vrt_crs, metadata_round_trip) is green.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_wkt_only_crs_warning_1768.py`
- [x] `pytest xrspatial/geotiff/tests/test_user_defined_crs_wkt_1632.py`
- [x] `pytest xrspatial/geotiff/tests/test_writer.py xrspatial/geotiff/tests/test_geotags.py xrspatial/geotiff/tests/test_writer_matrix.py xrspatial/geotiff/tests/test_write_vrt_crs_1715.py xrspatial/geotiff/tests/test_metadata_round_trip_1484.py`

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -1174,8 +1174,7 @@ def to_geotiff(data: xr.DataArray | np.ndarray,
                bigtiff: bool | None = None,
                gpu: bool | None = None,
                streaming_buffer_bytes: int = 256 * 1024 * 1024,
-               max_z_error: float = 0.0,
-               photometric='auto') -> None:
+               max_z_error: float = 0.0) -> None:
     """Write data as a GeoTIFF or Cloud Optimized GeoTIFF.
 
     Dask-backed DataArrays are written in streaming mode: one tile-row
@@ -1267,30 +1266,6 @@ def to_geotiff(data: xr.DataArray | np.ndarray,
         bounded by ``abs(decoded - original) <= max_z_error``. Only used
         when ``compression='lerc'``; passing a non-zero value with any
         other codec raises ``ValueError``.
-    photometric : str or int
-        Photometric interpretation for the TIFF Photometric tag (262).
-
-        * ``'auto'`` (default) -- MinIsBlack (1) for any band count.
-          ExtraSamples for every band beyond the first is tagged ``0``
-          (unspecified). Multispectral rasters (e.g. R, G, B, NIR)
-          round-trip through this default without being silently
-          labelled as RGB+alpha. Prior versions treated any 3+ band
-          array as RGB and the 4th band as unassociated alpha -- the
-          behaviour change is intentional (issue #1769).
-        * ``'rgb'`` -- RGB (Photometric=2). Three colour bands; any
-          additional bands are tagged ``0`` (unspecified).
-        * ``'rgba'`` -- RGB with the 4th band tagged as unassociated
-          alpha (TIFF ExtraSamples=2). Requires at least 4 bands.
-        * ``'minisblack'`` or ``'miniswhite'`` -- grayscale; multi-band
-          extras tagged ``0``.
-        * An ``int`` -- written verbatim into Photometric for advanced
-          callers (e.g. ``3`` for Palette, ``5`` for CMYK).
-
-        A user-supplied ``extra_tags`` entry of ``(TAG_PHOTOMETRIC,
-        ...)`` or ``(TAG_EXTRA_SAMPLES, ...)`` overrides the writer's
-        chosen value; only these two tag ids are overridable (other
-        auto-emitted tags such as ImageWidth or StripOffsets remain
-        protected).
     """
     from ._reader import _coerce_path
 
@@ -1436,8 +1411,7 @@ def to_geotiff(data: xr.DataArray | np.ndarray,
                               overview_levels=overview_levels,
                               overview_resampling=overview_resampling,
                               bigtiff=bigtiff,
-                              streaming_buffer_bytes=streaming_buffer_bytes,
-                              photometric=photometric)
+                              streaming_buffer_bytes=streaming_buffer_bytes)
             return
         except ImportError as e:
             # ``write_geotiff_gpu`` raises ImportError when cupy itself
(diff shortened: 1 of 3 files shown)


brendancol:
Pushed a follow-up commit addressing the failing CI and the Copilot review.

**Root cause of the 9-job failure:** the original PR also introduced a public `photometric` keyword to `to_geotiff` that forwarded into `write_geotiff_gpu`, `write_streaming`, and `_writer.write`, none of which accept that argument. Every non-streaming write raised `TypeError: write() got an unexpected keyword argument 'photometric'`. The kwarg is unrelated to the WKT-only CRS warning that #1768 is about, so I dropped it along with its docstring. If we want to expose photometric control later it should land as its own PR with the writer signatures updated to match.

**Review responses inline.** For the byte-level test, the previous version only checked that the GeoKeyDirectory had more than 4 entries and that `PROJCS[` was somewhere in the citation. It now pins the 32767 user-defined CRS marker by scanning the GeoKeyDirectory entries for `ProjectedCSType` / `GeographicType` and asserts the full WKT body round-trips through GeoAsciiParams.

For the warning message, the same branch fires for any non-EPSG CRS string (including PROJ strings that `_wkt_to_epsg` cannot resolve). The text now says "the supplied CRS string -- WKT or unresolved PROJ -- stored in GTCitationGeoKey" instead of asserting it is WKT.

For the test module's `pytest.importorskip("tifffile")`, the import-or-skip now lives in the one byte-level test that actually uses tifffile, so the build/end-to-end warning coverage still runs without it.

**On the `stacklevel=2` dedup concern**: keeping `stacklevel=2` is intentional. The warning is informational metadata about the bytes being written, not a caller mistake to be pinned to one user line; `default::UserWarning` filters dedup by `(module, lineno)` of the warning location, and the two `build_geo_tags` call sites in `_writer.py` (the georef'd and georef-less branches) each get their own record. That gives "warn once per session per branch" semantics, which is what we want for a metadata-quality nudge.

**On the docstring/test-count remark**: the test file defines six top-level functions and three of them exercise multiple paths through assertion sets; the "35 tests" framing in an earlier description was wrong, six is what the file contains and what runs. No change needed.

Issue #1581, opened 2026-05-11
Reading uint TIFF with negative GDAL_NODATA sentinel raises OverflowError


brendancol:
## Summary

A uint TIFF whose `GDAL_NODATA` tag is set to a negative sentinel (e.g. `-9999`) raises `OverflowError` on read instead of masking the nodata pixels to NaN. The `-9999` sentinel is common on unsigned-integer rasters because many GDAL pipelines write the tag as a string without checking dtype range. Both rasterio and GDAL promote the array to float and mask in this case; xarray-spatial fails the cast.

## Reproduction

```python
import numpy as np
import xarray as xr
from xrspatial.geotiff import to_geotiff, open_geotiff

arr = np.array([[1, 2, 3], [4, 5, 6]], dtype=np.uint16)
da = xr.DataArray(arr, dims=['y', 'x'])
to_geotiff(da, '/tmp/test.tif', crs=4326, nodata=-9999)

open_geotiff('/tmp/test.tif')
# OverflowError: Python integer -9999 out of bounds for uint16
```

## Affected backends

The same cast pattern lives in four places, so every backend hits the OverflowError:

- `xrspatial/geotiff/__init__.py` ~line 557: numpy eager read (`open_geotiff`)
- `xrspatial/geotiff/__init__.py` ~line 625: `_apply_nodata_mask_gpu` (cupy read)
- `xrspatial/geotiff/__init__.py` ~line 1528: `_delayed_read_window` (dask read)
- `xrspatial/geotiff/_reader.py`: `_resolve_masked_fill` and `_sparse_fill_value` (LERC masked fill + sparse tile fill)

Each does `arr.dtype.type(int(nodata))` or `dtype.type(int(v))` with no range check.

## Expected behavior

For any integer dtype with an out-of-range nodata sentinel, promote to float64 and skip the sentinel comparison (no value in the file can match anyway). Keep `attrs['nodata']` set to the user-declared value so write round-trips preserve it. This matches rasterio's behavior.

## Severity

HIGH (Cat 2 NaN propagation). This crashes the read of a class of TIFFs that are widespread in practice. Behavior matches across numpy / cupy / dask, so it is a shared accuracy bug rather than a backend divergence.

Issue #1616, opened 2026-05-11
read_vrt drops integer-source nodata when VRT declares a float dataType


brendancol:
**Describe the bug**

When a VRT XML declares `dataType="Float32"` or `"Float64"` but the underlying source GeoTIFFs are integer with an in-range `GDAL_NODATA` sentinel, `read_vrt` leaves the sentinel pixels as literal float values instead of NaN-masking them. The result is a float array whose `attrs['nodata']` advertises `65535.0`, but pixels at the sentinel positions still contain `65535.0` rather than `np.nan`. NaN-aware downstream code (`np.isnan`, `xr.where`, slope/aspect, statistics) sees those pixels as valid data and gets wrong answers.

**Repro**

```python
import numpy as np, tempfile, os
from xrspatial.geotiff import read_vrt
from xrspatial.geotiff._writer import write

band0 = np.array([[1, 2], [3, 65535]], dtype=np.uint16)
with tempfile.TemporaryDirectory() as td:
    p0 = os.path.join(td, 'b0.tif')
    write(band0, p0, nodata=65535, compression='none', tiled=False)

    vrt = os.path.join(td, 'mismatch.vrt')
    with open(vrt, 'w') as f:
        f.write(f'''<VRTDataset rasterXSize="2" rasterYSize="2">
  <GeoTransform>0.0, 1.0, 0.0, 0.0, 0.0, -1.0</GeoTransform>
  <VRTRasterBand dataType="Float32" band="1">
    <NoDataValue>65535</NoDataValue>
    <SimpleSource>
      <SourceFilename relativeToVRT="0">{p0}</SourceFilename>
      <SourceBand>1</SourceBand>
      <SrcRect xOff="0" yOff="0" xSize="2" ySize="2"/>
      <DstRect xOff="0" yOff="0" xSize="2" ySize="2"/>
    </SimpleSource>
  </VRTRasterBand>
</VRTDataset>''')

    r = read_vrt(vrt)
    print(r.dtype)                    # float32
    print(r.attrs['nodata'])          # 65535.0
    print(r.values[1, 1])             # 65535.0 -- should be nan
    print(np.isnan(r.values[1, 1]))   # False
```

**Expected behavior**

`r.values[1, 1]` should be `np.nan` after the read, matching the eager and dask paths for single-file int rasters with an in-range sentinel. `attrs['nodata']` should still surface the original sentinel for downstream code that needs to round-trip on write.

**Root cause**

[...]

Issue #1564, opened 2026-05-11
geotiff: read_vrt leaves integer nodata sentinel in pixels, divergent from open_geotiff


brendancol:
**Describe the bug**

`read_vrt` (also reachable via `open_geotiff` for `.vrt` paths) does not honour the nodata sentinel for integer source rasters. It sets `attrs['nodata']` on the returned DataArray but leaves the integer sentinel value (e.g. `65535`) intact in the pixel array. All other read paths (`open_geotiff` eager, `read_geotiff_dask`, `read_geotiff_gpu`, and the dask+cupy combination) promote integer rasters with a nodata sentinel to `float64` and replace the sentinel with NaN.

The result is a silent backend divergence: code that follows the convention "`attrs['nodata']` is set ⇒ values have been NaN-masked" (the convention codified in `_apply_nodata_mask_gpu` and used by all non-VRT reads) will treat the literal sentinel as a real value when the input route happens to be a VRT.

**Repro**

```python
import numpy as np, xarray as xr, tempfile, os
from xrspatial.geotiff import to_geotiff, open_geotiff, read_vrt, write_vrt

with tempfile.TemporaryDirectory() as d:
    arr = np.array([[1, 2, 3], [65535, 5, 6]], dtype=np.uint16)
    da = xr.DataArray(
        arr, dims=['y','x'],
        coords={'y': np.arange(2), 'x': np.arange(3)},
        attrs={'crs': 4326, 'nodata': 65535},
    )
    tif = os.path.join(d, 'src.tif')
    to_geotiff(da, tif, compression='none', nodata=65535)

    print(open_geotiff(tif).dtype, open_geotiff(tif).values)
    # float64
    # [[ 1.  2.  3.]
    #  [nan  5.  6.]]

    vrt = os.path.join(d, 'src.vrt')
    write_vrt(vrt, [tif])
    via_vrt = read_vrt(vrt)
    print(via_vrt.dtype, via_vrt.attrs.get('nodata'), via_vrt.values)
    # uint16 65535.0
    # [[    1     2     3]
    #  [65535     5     6]]
```

Float-with-nodata via VRT works correctly. Only the integer branch diverges.

**Expected behavior**

[...]

PR #1778, opened 2026-05-13
geotiff: guard int(nodata) on NaN/Inf GDAL_NODATA strings (#1774)

### LINKED ISSUE #1774: geotiff: int(nan) crash on integer TIFF with GDAL_NODATA="nan"
**Describe the bug**

Reading an integer GeoTIFF whose `GDAL_NODATA` tag holds a non-finite string ("NaN", "nan", "inf") crashes the read with `ValueError: cannot convert float NaN to integer`. The crash fires at three sites in `xrspatial/geotiff/__init__.py` that call `int(nodata)` on the integer dtype branch without first checking whether the float is finite:

- `open_geotiff` eager numpy path (line ~900)
- `_apply_nodata_mask_gpu` GPU path (line ~971)
- `_delayed_read_window` dask path (line ~2291)

The sibling helpers `_resolve_masked_fill` and `_sparse_fill_value` in `xrspatial/geotiff/_reader.py` already guard with `not math.isnan(v) and not math.isinf(v)` (issue #1581 / PR #1583), so the unguarded sites are an unfinished pass of the same fix.

**Reproduction**

A 2x2 uint16 TIFF whose `GDAL_NODATA` tag string is `"nan"`:

```python
import struct, tempfile, os
import numpy as np

bo = '<'
pixels = np.array([[10, 20], [30, 40]], dtype=np.uint16)

def short_tag(tag, val):
    return (tag, 3, 1, struct.pack(f'{bo}H', val))
def long_tag(tag, val):
    return (tag, 4, 1, struct.pack(f'{bo}I', val))
def ascii_tag(tag, data):
    return (tag, 2, len(data), data)

# ... assemble a minimal TIFF with GDAL_NODATA="nan\x00"
# (see test_nodata_nan_int_*.py for the full helper)

open_geotiff(path)  # ValueError: cannot convert float NaN to integer
```

**Expected behavior**

The read should treat a non-finite nodata sentinel on an integer file as a no-op (no integer pixel can equal NaN, so the mask would be all-False) and leave the file dtype unchanged. This mirrors the out-of-range sentinel handling already in place for `uint16 + GDAL_NODATA="-9999"`.

**Categories**

- Cat 2 (NaN propagation): a NaN nodata silently raises instead of being treated as missing
- Cat 5 (backend inconsistency): `_resolve_masked_fill` and `_sparse_fill_value` already guard; the three `__init__.py` sites do not

brendancol:
## Summary

Closes #1774.

`open_geotiff` / `read_geotiff_dask` / `_apply_nodata_mask_gpu` used to crash with `ValueError: cannot convert float NaN to integer` when reading an integer TIFF whose `GDAL_NODATA` tag was the string `"nan"` / `"inf"` / `"-inf"`. `_geotags.py:extract_geo_info` parses the tag through `float(nodata_str)`, so a `"nan"` tag surfaces as Python NaN; three sites in `xrspatial/geotiff/__init__.py` then called `int(nodata)` without checking finiteness.

The fix gates each `int(nodata)` cast on `np.isfinite(nodata)`, mirroring the `_resolve_masked_fill` / `_sparse_fill_value` helpers in `_reader.py` (the unfinished pass of #1581). A non-finite sentinel on an integer file cannot match any pixel value, so the mask is a no-op and the file dtype is preserved; `attrs['nodata']` still carries the raw NaN/Inf sentinel so a write round-trip keeps the original GDAL_NODATA tag.

- `xrspatial/geotiff/__init__.py`: gate `int(nodata)` on `np.isfinite(nodata)` at the eager numpy path (`open_geotiff`), the GPU helper (`_apply_nodata_mask_gpu`), and the dask delayed reader (`_delayed_read_window`). The `read_geotiff_dask` effective-dtype branch already used `try/except` but is tightened with the same gate for readability.
- `xrspatial/geotiff/tests/test_nodata_nan_int_1774.py`: 15 regression tests across all three backends and the finite-sentinel regression guard.

## Test plan

- [x] 15 new tests in `test_nodata_nan_int_1774.py` (3 NaN variants + 6 Inf variants on the eager numpy path, in-range finite sentinel still masks, dask NaN + Inf, GPU NaN + Inf + finite).
- [x] All 34 pre-existing nodata-related tests still pass (`test_nodata_out_of_range_1581.py`, `test_nodata_attr_aliases_1582.py`, `test_nodata_no_extra_copy_1553.py`, `test_gpu_nodata_1542.py`).
- [x] All 2023 other geotiff tests still pass. (7 pre-existing failures in `test_predictor2_big_endian_gpu_1517.py` are unrelated -- they reference `xrspatial.geotiff.read_to_array` which was hidden from the public namespace in #1708. 3 pre-existing matplotlib palette failures in `test_features.py` are also unrelated.)


brendancol:
Good catch. Pushed `1dd393f` adding `float(nodata).is_integer()` alongside the `np.isfinite` gate at all four sites. Without it, a fractional `GDAL_NODATA="3.5"` on a `uint16` file would truncate via `int(3.5) == 3` and silently mask real pixel value 3.

The new check matches the existing `_writer.py:280-282` and `_vrt.py:827-829` pattern that handles the same case on the VRT side (#1564 / #1616). Added 5 regression tests, including a "30.5 must not mask pixel 30" truncation-aliasing guard so the integerness gate cannot regress to a bare `np.isfinite`.

PR #1779, opened 2026-05-13
geotiff: extend #1752 size-arg validation to write_geotiff_gpu / read_geotiff_gpu / read_vrt (#1776)

### LINKED ISSUE #1776: geotiff: extend #1752 size-arg validation to write_geotiff_gpu, read_geotiff_gpu, read_vrt
## What's broken

Issue #1752 added `tile_size` and `chunks` validation to `to_geotiff` and `read_geotiff_dask`. The matching kwargs on three sibling entry points were skipped, so they still reach the deeper writer/dask paths with invalid values and surface confusing errors that do not name the bad parameter.

The gaps:

1. `write_geotiff_gpu(tile_size=)` is missing the validation `to_geotiff` does. The four cases #1752 covers for `to_geotiff`:
   - `tile_size=0` raises `ZeroDivisionError` from `gpu_compress_tiles`
   - `tile_size=-1` raises `struct.error` (`'H' format requires 0 <= number <= 65535`)
   - `tile_size=False` raises `ZeroDivisionError` (bool slipped through because `False == 0`)
   - `tile_size=256.0` raises `TypeError` (`'float' object cannot be interpreted as an integer`)

2. `read_geotiff_gpu(chunks=)` is missing the validation `read_geotiff_dask` does:
   - `chunks=0` raises `ZeroDivisionError`
   - `chunks=-1` silently produces output (no validation at all)
   - `chunks=(0, 256)` raises `ZeroDivisionError`
   - `chunks=(256, -1)` silently produces output

3. `read_vrt(chunks=)` is missing the same validation as 2.

`open_geotiff(chunks=)` is fine on the CPU path because it routes through `read_geotiff_dask`'s already-validated branch. The gap shows up for `open_geotiff(gpu=True, chunks=X)`, direct `read_geotiff_gpu(chunks=X)`, and `read_vrt(chunks=X)`.

## How to reproduce

```python
import numpy as np
import xarray as xr
import cupy
import os
import tempfile
from xrspatial.geotiff import (
    write_geotiff_gpu, read_geotiff_gpu, read_vrt,
    to_geotiff, write_vrt,
)

arr = np.arange(100, dtype=np.float32).reshape(10, 10)
da = xr.DataArray(arr, dims=['y', 'x'])

    tif = os.path.join(d, 'a.tif')
    to_geotiff(da, tif)
    vrt = os.path.join(d, 'a.vrt')
    write_vrt(vrt, [tif])

    write_geotiff_gpu(xr.DataArray(cupy.asarray(arr), dims=['y', 'x']),
                      os.path.join(d, 'out.tif'),
                      tile_size=0)        # ZeroDivisionError
    read_geotiff_gpu(tif, chunks=0)        # ZeroDivisionError
    read_geotiff_gpu(tif, chunks=-1)       # silently passes
    read_vrt(vrt, chunks=-1)               # silently passes
```

## Suggested fix

Mirror the up-front guard `#1752` added to `to_geotiff` and `read_geotiff_dask`:

- `write_geotiff_gpu`: reject non-positive int or non-int `tile_size` with a `ValueError` naming the parameter.
- `read_geotiff_gpu` and `read_vrt`: reject non-positive int or non-int `chunks` (and tuple-of-non-positive, and wrong-length tuple) with the same message format `read_geotiff_dask` uses.

Source touches are small; the guards already exist on the sibling entry points and can be reused.

## Tests

Tests that should be added in the same PR:

- `tile_size=0`, `tile_size=-1`, `tile_size=256.0`, `tile_size=True/False` on `write_geotiff_gpu`
- `chunks=0`, `chunks=-1`, `chunks=(0, N)`, `chunks=(N, -1)`, wrong-length tuple, non-int on `read_geotiff_gpu` and `read_vrt`
- Positive int and `np.int64` scalar still work (regression pin against the integer-scalar branch).

Surfaced by the test-coverage sweep on the geotiff module (pass 13, 2026-05-13).

brendancol:
## Summary

- Closes #1776.
- Adds `tile_size` validation to `write_geotiff_gpu` and `chunks` validation to `read_geotiff_gpu` / `read_vrt`, mirroring the guards #1752 added to `to_geotiff` / `read_geotiff_dask`.
- Factors the rules into two shared helpers (`_validate_tile_size_arg`, `_validate_chunks_arg`) so all six entry points emit identical error formats.

## What was broken

`write_geotiff_gpu(tile_size=0)` raised `ZeroDivisionError` from `gpu_compress_tiles`. `tile_size=-1` raised `struct.error`. `tile_size=256.0` raised `TypeError`. `read_geotiff_gpu(chunks=0)` and `read_vrt(chunks=0)` raised `ZeroDivisionError` from dask chunk math. `chunks=-1` was silently accepted on both and produced a corrupt chunk grid.

## Test plan

- [x] 34 new tests in `test_size_param_validation_gpu_vrt_1776.py` cover the full matrix on all three entry points + dispatcher thread-through tests.
- [x] Pre-existing 13 #1752 tests still pass after the refactor of `to_geotiff` / `read_geotiff_dask` onto the shared helpers.
- [x] Broader related-test sweep (`test_writer.py`, `test_vrt_*`, `test_kwarg_behaviour_*`, etc.) passes -- 110/110.

Surfaced by the test-coverage sweep on the geotiff module (pass 13, 2026-05-13).


brendancol:
Thanks @copilot-pull-request-reviewer — fixed in 03d141f. Took option (b): added an `allow_none` flag (default `False`) to `_validate_chunks_arg`. `read_geotiff_dask` now keeps its #1752 behaviour of rejecting `chunks=None` with a clear `ValueError` (it was emitting a confusing `TypeError: cannot unpack non-iterable NoneType object` from the chunk-unpack math). `read_geotiff_gpu` and `read_vrt` pass `allow_none=True` since their public defaults are `chunks=None` (eager read). Docstring updated and three new regression tests pin all three branches.

PR #1781, opened 2026-05-13
geotiff: require tile_size to be a multiple of 16 (#1767)

### LINKED ISSUE #1767: to_geotiff accepts tile_size values that are not multiples of 16
`to_geotiff` accepts `tile_size` values that violate the TIFF 6 spec.

The spec requires TileWidth and TileLength to be multiples of 16. The current validation in `xrspatial/geotiff/__init__.py` only checks that `tile_size` is a positive int, so `tile_size=17` slips through and writes a file that the in-repo reader round-trips fine but that strict TIFF readers (GDAL, libtiff strict mode) may reject.

Reproducer:

```python
import numpy as np
import xarray as xr

da = xr.DataArray(np.zeros((32, 32), dtype="float32"), dims=("y", "x"))
to_geotiff(da, "/tmp/bad.tif", tiled=True, tile_size=17)  # writes anyway
```

Fix: when `tiled=True`, require `tile_size` to be a positive multiple of 16. Raise `ValueError` naming the bad value and suggesting the nearest valid choices (`(tile_size // 16) * 16` and `((tile_size // 16) + 1) * 16`). Update the docstring.

brendancol:
Closes #1767.

## Summary

- `to_geotiff` and `write_geotiff_gpu` now refuse `tile_size` values that aren't a positive multiple of 16 when `tiled=True`. TIFF 6 spec requires TileWidth and TileLength to be multiples of 16, and other tools (GDAL, libtiff strict mode) may reject files that violate it.
- The error names the bad value and suggests the nearest valid choices (`(tile_size // 16) * 16` and the next multiple up).
- Docstrings on `to_geotiff` updated to call out the constraint.

## Test plan

- [x] New tests in `xrspatial/geotiff/tests/test_tile_size_multiple_of_16_1767.py` cover `tile_size=17`, `tile_size=1`, `tile_size=8` (sub-16 hint), `tile_size=24` (both-sides hint), `tile_size=16`, `tile_size=128`, `tile_size=256`, `tile_size=512`, and `tile_size=17` with `tiled=False` (must pass).
- [x] `test_size_param_validation_1752.py::test_to_geotiff_tile_size_one_still_writes` was inverted (it asserted `tile_size=1` worked; that's now disallowed) and replaced with a `tile_size=16` positive-path check.
- [x] Existing fixtures that wrote files with `tile_size=2/4/8/100` for multi-tile coverage were bumped to multiples of 16; raster sizes grew only where the test logic depended on a >1-tile grid.
- [x] Full geotiff test suite passes apart from three pre-existing Python 3.14 matplotlib palette deepcopy failures unrelated to this change.


brendancol:

Issue #1796, opened 2026-05-13
VRT source reads do not honor max_pixels safety limit


brendancol:
**Describe the bug**

`read_vrt(..., max_pixels=...)` caps the assembled VRT output and some resample intermediates, but it does not pass `max_pixels` into the source GeoTIFF reads. A VRT can request a small output or destination rect while forcing a much larger source `SrcRect` decode, bypassing the caller's safety budget.

**Expected behavior**

The `max_pixels` limit should apply to every materialized source window and intermediate created while reading a VRT, not just the final VRT output.

**Screenshots**

Not applicable.

**Desktop**

Not applicable.

**Smartphone**

Not applicable.

**Additional context**

The source read in `xrspatial/geotiff/_vrt.py` calls `read_to_array(...)` without forwarding `max_pixels`.

PR #1807, opened 2026-05-13
Make read_vrt chunks mode lazy

brendancol:
Closes #1798.\n\nAdds a lazy CPU dask VRT path that builds windowed tasks instead of assembling the full mosaic before chunking.\n\nTested: pytest xrspatial/geotiff/tests/test_read_vrt_lazy_chunks_1798.py

PR #1824, opened 2026-05-13
geotiff: per-tile dim check uses default cap, not caller budget

### LINKED ISSUE #1823: VRT source reads with small max_pixels reject normal-tile sources
PR #1803 added `max_pixels=max_pixels` to the source `read_to_array` call inside `read_vrt`'s source loop so that a tiny VRT output cannot force a huge source decode (#1796). The intent is correct, but `_read_tiles` and `_read_tiles_cog_http` use the same `max_pixels` for two separate things:

1. **Output window budget** (`_check_dimensions(out_w, out_h, samples, max_pixels)`) — enforced for the data the caller will receive. This is what catches the #1796 attack.
2. **Per-tile dimension sanity** (`_check_dimensions(tw, th, samples, max_pixels)` at `_reader.py:1565` and `:2022`) — meant to reject crafted TIFFs whose `TileWidth` / `TileLength` headers declare absurd per-tile sizes (e.g. `tw = 2^31`).

After #1803 a user who sets `max_pixels=10000` to bound their *output* now also fails the per-tile check on every normal source: a default 256×256 tile (65,536 pixels) trips the cap even when the requested window is much smaller.

Repro:

```python
import numpy as np, tempfile, os
from xrspatial.geotiff import to_geotiff, read_vrt

with tempfile.TemporaryDirectory() as td:
    to_geotiff(np.zeros((10, 10), dtype=np.uint8),
               os.path.join(td, 'src.tif'), compression='none')
    open(os.path.join(td, 'm.vrt'), 'w').write(
        '<VRTDataset rasterXSize="10" rasterYSize="10">'
        '<VRTRasterBand dataType="Byte" band="1"><SimpleSource>'
        '<SourceFilename relativeToVRT="1">src.tif</SourceFilename>'
        '<SourceBand>1</SourceBand>'
        '<SrcRect xOff="0" yOff="0" xSize="10" ySize="10"/>'
        '<DstRect xOff="0" yOff="0" xSize="100" ySize="100"/>'
        '</SimpleSource></VRTRasterBand></VRTDataset>')
    read_vrt(os.path.join(td, 'm.vrt'), max_pixels=10000)
# PixelSafetyLimitError: TIFF image dimensions (256 x 256 x 1 = 65,536 pixels)
# exceed the safety limit of 10,000 pixels.
```

The output is 100×100=10,000 pixels, exactly at the cap; the source-tile check rejects 256×256=65,536 first.

Fix: the per-tile dimension sanity check should use a separate, large absolute cap (e.g. `MAX_PIXELS_DEFAULT = 1e9`) rather than the caller's output budget. The output-window check at the same call sites continues to enforce the user-supplied `max_pixels`, preserving the #1796 protection.

Existing test `xrspatial/geotiff/tests/test_vrt_dstrect_resample_cap_1737.py::test_dstrect_at_cap_succeeds` is the regression: it was authored against the pre-#1803 contract and now fails. Six open PRs share this failure on Python 3.14 (the matrix's first-fail) and have the rest of the matrix cancelled fail-fast, so they all appear BLOCKED downstream.

brendancol:
Closes #1823.

## Summary
- `_check_dimensions(tw, th, samples, max_pixels)` at `_reader.py:1565` and `:2027` is a sanity check on `TileWidth` / `TileLength` header values (rejects crafted `2**31`-style tile dims). After #1803 it consumed the caller's `max_pixels`, so a small output budget like `read_vrt(..., max_pixels=10000)` failed on every normal source: `to_geotiff` writes a default 256x256 tile (= 65,536 pixels) which overshoots the caller's output budget at the per-tile check.
- Switch the per-tile call sites to `MAX_PIXELS_DEFAULT` (1e9). The output-window check on the same call paths still enforces the caller's `max_pixels`, so the #1796 protection is preserved.

## Why on main, separate PR
Six open PRs (`#1811 #1817 #1818 #1819 #1820 #1822`) all fail `test_vrt_dstrect_resample_cap_1737::test_dstrect_at_cap_succeeds` on Python 3.14 because of this regression. Fixing it on main unblocks all of them at once.

## Test plan
- [x] `pytest xrspatial/geotiff/tests/test_vrt_source_tile_check_1823.py` (new, 4 tests)
- [x] `pytest xrspatial/geotiff/tests/test_vrt_dstrect_resample_cap_1737.py` (regression fixed)
- [x] `pytest xrspatial/geotiff/tests/test_vrt_source_max_pixels_1796.py` (#1796 protection still enforced)
- [x] Filtered geotiff suite: 25 passed (pixel-safety / max_pixels / tile-check / dstrect)

### DIFF
--- a/xrspatial/geotiff/_reader.py
+++ b/xrspatial/geotiff/_reader.py
@@ -1560,9 +1560,14 @@ def _read_tiles(data: bytes, ifd: IFD, header: TIFFHeader,
         raise ValueError(
             f"Invalid tile dimensions: TileWidth={tw}, TileLength={th}")
 
-    # Reject crafted tile dims that would force huge per-tile allocations.
-    # A single tile's decoded bytes must also fit under the pixel budget.
-    _check_dimensions(tw, th, samples, max_pixels)
+    # Reject crafted tile dims (e.g. TileWidth = 2**31). This guards the
+    # TIFF header against malformed values; it is not the caller's output
+    # budget. The output-window check below uses ``max_pixels`` and is
+    # what enforces the user's per-call memory cap. The source-read path
+    # under ``read_vrt`` (#1796) relies on that output check to honour a
+    # small caller ``max_pixels`` against a normal-tile source; see
+    # #1823.
+    _check_dimensions(tw, th, samples, MAX_PIXELS_DEFAULT)
 
     # Per-tile compressed-byte cap (issue #1664). Same env var as the
     # HTTP path. mmap slicing is bounded by the file size, but the slice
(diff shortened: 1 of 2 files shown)


brendancol:
Closing as effectively merged. The per-tile dim check fix from this PR rode into main via cherry-picks that were included in the squash-merges of #1817, #1818, #1820, and #1807; main's `xrspatial/geotiff/_reader.py:1570` and `:2039` now read `_check_dimensions(tw, th, samples, MAX_PIXELS_DEFAULT)`. The accompanying tests were updated for #1821's minimal-read semantics in #1827.

Issue #1823 can be closed too.

PR #1834, opened 2026-05-14
geotiff: cover gil_friendly kwarg behaviour added in #1826 (#1830)

### LINKED ISSUE #1830: Test: cover gil_friendly kwarg behaviour in geotiff writer/compression
## Coverage gap

PR #1826 added a `gil_friendly: bool = False` parameter to:

- `xrspatial.geotiff._compression.deflate_compress`
- `xrspatial.geotiff._compression.compress`
- `xrspatial.geotiff._writer._prepare_strip`
- `xrspatial.geotiff._writer._prepare_tile`
- `xrspatial.geotiff._writer._compress_block`

The flag gates a documented optimisation: when `True`, deflate is forced through stdlib `zlib.compress` (which releases the GIL) even if the optional `deflate` (libdeflate) PyPI binding is installed. The libdeflate binding does not release the GIL, so the writer's parallel strip/tile paths pass `gil_friendly=True` to keep thread-pool scaling (measured 5x vs 1.2x with 8 threads in the PR description).

The PR also added a one-shot `UserWarning` in `deflate_compress` when libdeflate is missing.

### Behaviour with no direct test coverage

1. `deflate_compress(data, gil_friendly=True)` with libdeflate installed must bypass the libdeflate binding and call `zlib.compress` instead. A regression dropping the `and not gil_friendly` clause in `_compression.py:137` would silently re-route the parallel writer through the GIL-holding binding and ship the documented thread-pool scaling regression.
2. `deflate_compress(data, gil_friendly=False)` with libdeflate installed must call `deflate.zlib_compress` (output stays wire-compatible but the call path differs).
3. `compress(compression=DEFLATE, gil_friendly=True/False)` must forward the flag to `deflate_compress`. Other codec branches (LZW/PackBits/zstd/lz4) must ignore the flag since their bindings already release the GIL.
4. The one-shot fallback `UserWarning` must fire exactly once when libdeflate is missing (existing test in `test_parallel_writer_1800.py::test_deflate_compress_fallback_when_libdeflate_missing` suppresses the warning rather than asserting it; the latch behaviour is uncovered).
5. The parallel strip path in `_write_stripped` must invoke `_prepare_strip` with `gil_friendly=True`; the sequential strip path must invoke it with the default (`False`). Same matrix for `_write_tiled` -> `_prepare_tile` and the `write_streaming` parallel-tile branch through `_compress_block`.

These are all Cat 4 HIGH parameter-coverage gaps under the test-coverage sweep rubric: a documented bool flag with multiple meaningful branches, none of which is directly exercised. Existing tests (`test_parallel_writer_1800.py`, `test_compression_level.py`) cover round-trip correctness and the thread-pool dispatch but never observe which deflate backend ran.

### Fix scope

Tests only. No source changes.

brendancol:
## Summary

- Closes #1830 by adding 20 direct tests for the `gil_friendly` kwarg PR #1826 threaded through `deflate_compress` / `compress` / `_prepare_strip` / `_prepare_tile` / `_compress_block`.
- The flag forces deflate through stdlib `zlib.compress` so the parallel writer paths actually scale across threads; existing tests covered round-trip correctness and pool dispatch but never observed which deflate backend ran or what flag value the writer passed.
- New file: `xrspatial/geotiff/tests/test_gil_friendly_kwarg_1830.py`. No source changes.

### What the new tests pin

- `deflate_compress(gil_friendly=True)` bypasses libdeflate even when the optional `deflate` PyPI binding is installed; `gil_friendly=False` keeps using it; both round-trip through stdlib `zlib.decompress` at levels 1/6/9.
- The fallback `UserWarning` fires exactly once when libdeflate is missing, is suppressed when the module-global latch is set, and is not re-emitted across subsequent calls (no per-call spam).
- `compress(DEFLATE, gil_friendly=...)` forwards the flag; the flag is a no-op for LZW/PackBits/zstd/lz4/none.
- `_write_stripped` parallel path passes `gil_friendly=True` on every strip; the sequential path leaves it at `False`. Same matrix for `_write_tiled`.
- `_prepare_strip` and `_prepare_tile` forward the kwarg into `deflate_compress`. The parallel `_write_tiled` branch passes `True` positionally to `_prepare_tile`, pinned via the function signature.
- End-to-end round-trip across both strip/tile and parallel/sequential modes.

### Verification

- Local: `pytest xrspatial/geotiff/tests/test_gil_friendly_kwarg_1830.py` -> 20 passed.
- Adjacent suites still green: `test_parallel_writer_1800.py` + `test_compression.py` + `test_compression_level.py` + `test_streaming_write.py` + new file -> 99 passed.
- Mutation checks:
  - Drop the `and not gil_friendly` guard in `_compression.py:137` -> 2 tests red (`test_deflate_compress_gil_friendly_true_bypasses_libdeflate`, `test_compress_forwards_gil_friendly_to_deflate`).
  - Flip the writer's `gil_friendly=True` to `False` in `_write_stripped` -> `test_write_stripped_parallel_path_uses_gil_friendly` red.
  - Delete the fallback `warnings.warn` block -> `test_deflate_compress_fallback_warning_fires_when_libdeflate_missing` red.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_gil_friendly_kwarg_1830.py -q` -> 20 passed on a host with `deflate` installed.
- [ ] CI: same file runs on a host without `deflate` -- the libdeflate-installed tests skip cleanly, the fallback tests stay green.

### DIFF
--- a/xrspatial/geotiff/tests/test_gil_friendly_kwarg_1830.py
+++ b/xrspatial/geotiff/tests/test_gil_friendly_kwarg_1830.py
@@ -34,6 +34,7 @@ from xrspatial.geotiff._compression import (
     COMPRESSION_PACKBITS,
     COMPRESSION_ZSTD,
     _HAVE_LIBDEFLATE,
+    LZ4_AVAILABLE,
     compress,
     deflate_compress,
 )
@@ -270,8 +271,12 @@ def test_compress_gil_friendly_ignored_for_non_deflate_codecs():
         (COMPRESSION_PACKBITS, raw),
         (COMPRESSION_LZW, raw),
         (COMPRESSION_ZSTD, raw),
-        (COMPRESSION_LZ4, raw),
     ]
+    # ``lz4`` is an optional dependency. On CI runners that ship without it
+    # (some macOS images) the codec dispatch path raises ImportError; skip
+    # that row rather than fail the whole non-deflate-codec coverage test.
+    if LZ4_AVAILABLE:
+        matrix.append((COMPRESSION_LZ4, raw))
     for tag, payload in matrix:
         out_false = compress(payload, tag, gil_friendly=False)
         out_true = compress(payload, tag, gil_friendly=True)
--- a/xrspatial/geotiff/tests/test_gil_friendly_kwarg_1830.py
+++ b/xrspatial/geotiff/tests/test_gil_friendly_kwarg_1830.py
@@ -41,6 +41,7 @@ from xrspatial.geotiff._compression import (
 from xrspatial.geotiff._reader import read_to_array
 from xrspatial.geotiff._writer import (
     _PARALLEL_MIN_BYTES,
+    _compress_block,
     _prepare_strip,
     _prepare_tile,
     _write_stripped,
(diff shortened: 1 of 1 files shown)


brendancol:
Addressed the Copilot coverage gap in commit a880bc3. The section header claimed ``write_streaming`` was verified but the file only exercised ``_write_stripped`` / ``_write_tiled`` / ``_prepare_strip`` / ``_prepare_tile``. ``write_streaming`` routes per-tile compression through ``_compress_block``, which carries its own ``gil_friendly`` forwarding (``True`` positionally on the parallel tile-segment branch, default ``False`` on the serial branch).

Added three direct pins:
- ``_compress_block(gil_friendly=True)`` reaches ``deflate_compress`` with the flag set.
- ``_compress_block`` default keeps ``gil_friendly=False`` so serial segments stay on libdeflate.
- ``write_streaming`` on a multi-tile dask array drives at least one ``deflate_compress`` call with ``gil_friendly=True``, pinning the parallel branch end-to-end.

PR #1837, opened 2026-05-14
geotiff: writer pre-inverts MinIsWhite pixels so values round-trip (#1836)

### LINKED ISSUE #1836: photometric='miniswhite' silently inverts pixel values on round-trip
**Describe the bug**

\`to_geotiff(..., photometric='miniswhite')\` writes the TIFF Photometric tag (value 0) but does not invert pixel values. The reader unconditionally inverts \`photometric == 0\` single-band data at \`_reader.py:2334\` (\`_apply_photometric_miniswhite\`) and applies it during reads at \`_reader.py:2524\`. The two paths are asymmetric, so a round-trip silently corrupts pixel values.

**Reproduce**

\`\`\`python

a = np.array([[0, 255]], dtype=np.uint8)
da = xr.DataArray(a, dims=('y', 'x'),
                  coords={'y': [0.0], 'x': [0.0, 1.0]},
                  attrs={'res': (1.0, 1.0)})
to_geotiff(da, 'out.tif', photometric='miniswhite')
r = open_geotiff('out.tif')
print(r.values)  # [[255 0]] -- inverted
\`\`\`

For analytical rasters this is silent data corruption.

**Expected behavior**

Either:
- (a) The writer inverts pixels symmetrically with the reader so the round-trip is the identity, or
- (b) \`photometric='miniswhite'\` is rejected for analytical data with a clear error.

Preferred: (a). Mirror the reader inversion (\`_apply_photometric_miniswhite\`):
- unsigned: \`np.iinfo(dtype).max - arr\`
- float: \`-arr\`
- signed/other: pass through

**Additional context**

Tag set on write: \`_writer.py:94\` (\`_PHOTOMETRIC_NAME_MAP\`).
Reader inversion: \`_reader.py:2334\` (\`_apply_photometric_miniswhite\`), applied at \`_reader.py:2524\`.

brendancol:
Fixes #1836.

## Summary
- ``to_geotiff(..., photometric='miniswhite')`` set the TIFF Photometric tag to 0 without inverting pixel values. The reader unconditionally inverts single-band MinIsWhite data, so the round trip returned ``iinfo(dtype).max - original`` (uint) or ``-original`` (float). For analytical rasters this was silent data corruption.
- Pre-invert pixels and the nodata sentinel on the writer side so the reader's inversion restores the user's values. Signed integer data passes through unchanged on both sides (matches the reader).
- Refuse ``photometric='miniswhite'`` with ``cog=True`` or explicit ``overview_levels``: overview reducers (``min``, ``max``, ``mode``, ...) do not commute with the pixel inversion, so the on-disk summary statistics would not match the user-domain values.
- Update ``test_miniswhite_backend_parity_1797`` to assert the new round-trip identity. The previous expectation pinned the sign-flip bug; HTTP-served files keep their expectations because ``tifffile`` does not pre-invert.

## Test plan
- [x] New regression tests in ``test_miniswhite_writer_roundtrip_1836.py`` cover uint8 / uint16 / float32 round-trip, signed-int passthrough, nodata-via-NaN round-trip, and the ``cog`` / ``overview_levels`` refusals.
- [x] All MinIsWhite / writer / nodata tests still pass (522 passed).
- [x] Remaining 11 failures in the full ``xrspatial/geotiff/tests/`` run reproduce on ``main`` (pre-existing GPU and ``test_features`` issues, unrelated to this PR).

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -3715,6 +3715,26 @@ def write_geotiff_gpu(data: xr.DataArray | cupy.ndarray | np.ndarray,
             "compression is tile-based; the strip layout is not "
             "implemented on the GPU path. Use to_geotiff(..., gpu=False, "
             "tiled=False) for strip output on CPU.")
+    # MinIsWhite pre-inversion (issue #1836) runs in the eager CPU writer.
+    # The GPU writer assembles tile bytes directly on device; threading
+    # the pixel + nodata-sentinel transform through that pipeline is out
+    # of scope for the round-trip fix. Refuse the combination so callers
+    # do not silently get inverted on-disk values. Move the array to the
+    # CPU and call the eager ``write`` path for MinIsWhite output.
+    from ._writer import _resolve_photometric as _resolve_photo_gpu
+    _gpu_samples_hint = (data.shape[2] if hasattr(data, 'shape')
+                         and data.ndim == 3 else 1)
+    _gpu_resolved_photo, _ = _resolve_photo_gpu(
+        photometric, _gpu_samples_hint)
+    if _gpu_resolved_photo == 0 and _gpu_samples_hint == 1:
+        raise NotImplementedError(
+            "photometric='miniswhite' on the GPU writer is not "
+            "supported: the writer-side pixel inversion that mirrors "
+            "the reader's unconditional MinIsWhite inversion (issue "
+            "#1836) is only wired into the eager CPU ``write`` path. "
+            "Move the array to host memory and call to_geotiff with "
+            "gpu=False, or write with photometric='minisblack' / "
+            "'auto'.")
     # write_geotiff_gpu is always tiled, so validate tile_size here and
     # keep parity with the public to_geotiff entry point.
     _validate_tile_size_arg(tile_size)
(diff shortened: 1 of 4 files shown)


brendancol:
Addressed all six Copilot comments in commit fb9f8d2:

1. **extra_tags TAG_PHOTOMETRIC override** — the pre-inversion decision keyed off the kwarg, so an override could leave pixels and the on-disk tag in different photometric domains. Now refused with a ``ValueError`` for the single-band MinIsWhite collision case.
2. **Overview-rejection test isolation** — split into ``test_miniswhite_rejected_with_cog_no_overviews`` and ``test_miniswhite_rejected_with_explicit_overviews_no_cog`` so each branch of the ``cog or overview_levels is not None`` guard is exercised separately.
3. **``_invert_nodata_for_miniswhite`` docstring** — corrected to call out that ``+/-inf`` on a float sentinel is negated (matching the reader); only NaN passes through unchanged.
4. **Module docstring** — describes both the unsigned (``iinfo.max - original``) and float (``-original``) pre-fix failure modes.
5. **Dask / GPU paths** — these silently bypassed the writer-side inversion. Added explicit refusal with a clear error in ``write_streaming`` and ``write_geotiff_gpu`` directing callers to ``.compute()`` and the eager writer. New tests cover the dask refusal.
6. **uint nodata coverage** — added ``test_uint16_miniswhite_with_in_range_nodata_round_trip`` so the unsigned-sentinel inversion branch is exercised directly, not just via the float path.

All 525 related tests still pass locally.

Issue #1784, opened 2026-05-13
geotiff: malformed VRT SrcRect is swallowed by the lenient source-read fallback


brendancol:
**Describe the bug**

`read_vrt()` in `xrspatial/geotiff/_vrt.py` validates `DstRect` for negative sizes at lines 637-648 (issue #1737), but never validates `SrcRect`. A `SimpleSource` with `<SrcRect xOff="0" yOff="0" xSize="-100" ySize="-100"/>` reaches the source-read path and produces a `ValueError` inside `read_to_array()` when the window goes negative.

That `ValueError` is then caught by the broad source-read fallback at lines 738-746:

```python
try:
    src_arr, _ = read_to_array(src.filename, window=(read_r0, read_c0, read_r1, read_c1), band=src.band - 1)
except (OSError, ValueError, struct.error,) + _CODEC_DECODE_EXCEPTIONS as e:
    ...  # warns and zero-fills under default (non-strict) mode
```

The fallback's purpose is to keep going when a referenced tile is missing or unreadable. Malformed XML is a different failure -- the file is present, the rectangle is just nonsense -- but the lenient path can't tell the two apart and treats both as "skip and warn." The result is a hole in the mosaic plus a confusing warning that mentions a window the caller never asked for.

Negative `SrcRect` offsets and offsets that fall outside the source raster have the same problem.

**Expected behavior**

Reject negative `SrcRect` sizes and negative offsets up-front, mirroring the `DstRect` check at line 643. The malformed-XML failure should always reach the caller, regardless of strict mode, with a typed `ValueError` that names the bad field.

**Categories**

- Cat 3 (input validation): negative SrcRect dimensions not validated
- Cat 4 (error handling): malformed XML masquerades as a missing-tile warning

Issue #1737, opened 2026-05-12
geotiff: VRT SimpleSource DstRect xSize/ySize allows unbounded intermediate allocation in _resample_nearest


brendancol:
## Summary

A crafted VRT file with a `<SimpleSource>` whose `<DstRect xSize=...  ySize=...>` is much larger than the declared `<VRTDataset rasterXSize ... rasterYSize=...>` extent triggers an unbounded intermediate allocation inside `_resample_nearest`. The output buffer is bounded by `_check_dimensions(out_w, out_h, n_bands, max_pixels)` at `_vrt.py:517`, but the resampled-source intermediate is allocated at `dr.y_size x dr.x_size` before the clip is taken.

Reproduction (under 30 lines of code) builds a `<DstRect xSize='50000' ySize='50000'/>` against a 10x10 source under a 100x100 VRT extent. The output array is 100x100 (10 KB) but the call passes through `_resample_nearest(src_arr, 50000, 50000)` which allocates ~2.5 GB of uint8 inside `np.repeat(np.repeat(src_arr, ry, axis=0), rx, axis=1)`. Larger values (`xSize='200000' ySize='200000'`) push past 40 GB on the same input. The peak allocation is unbounded by `max_pixels` and not gated by any explicit check.

## Reproducer

```python
import os, tempfile, numpy as np
from xrspatial.geotiff import to_geotiff
from xrspatial.geotiff._vrt import read_vrt

with tempfile.TemporaryDirectory() as td:
    src = os.path.join(td, 'src.tif')
    to_geotiff(np.zeros((10, 10), dtype=np.uint8), src, compression='none')
    vrt = os.path.join(td, 'bomb.vrt')
    with open(vrt, 'w') as f:
        f.write('''<VRTDataset rasterXSize="100" rasterYSize="100">
  <VRTRasterBand dataType="Byte" band="1">
    <SimpleSource>
      <SourceFilename relativeToVRT="1">src.tif</SourceFilename>
      <SourceBand>1</SourceBand>
      <SrcRect xOff="0" yOff="0" xSize="10" ySize="10"/>
      <DstRect xOff="0" yOff="0" xSize="50000" ySize="50000"/>
    </SimpleSource>
  </VRTRasterBand>
</VRTDataset>''')
    # Reads a 100x100 array but allocates ~2.5 GB inside _resample_nearest
    arr, _ = read_vrt(vrt)
```

## Severity

[...]

Issue #1697, opened 2026-05-12
read_vrt: out-of-bounds window silently clamped instead of raising (drift from local/HTTP)


brendancol:
## Summary

`read_vrt(path, window=...)` silently clamps invalid window coordinates instead of raising `ValueError`. The local TIFF path (#1634) and HTTP COG path (#1669) both reject the same bad windows. VRT slipped through.

`xrspatial/geotiff/_vrt.py:407-414`:

```python
if window is not None:
    r0, c0, r1, c1 = window
    r0 = max(0, r0)
    c0 = max(0, c0)
    r1 = min(vrt.height, r1)
    c1 = min(vrt.width, c1)
else:
    r0, c0, r1, c1 = 0, 0, vrt.height, vrt.width
```

#1634 added a validator on the local path (`_reader.py:2013-2020`) that raises `ValueError` for any out-of-bounds, zero-size, or inverted window. The HTTP path got the same validator (`_reader.py:1620-1628`). VRT was skipped, probably by accident when #1634 landed.

## Reproducer

```python
from xrspatial.geotiff._vrt import read_vrt
out, _ = read_vrt(path, window=(-1, 0, 2, 2))
print(out.shape)  # (2, 2) -- r0 clamped to 0, no exception
```

On a 4x4 VRT:

| window | actual | expected |
| --- | --- | --- |
| `(-1, 0, 2, 2)` | `(2, 2)` array | `ValueError` |
| `(0, 0, 5, 5)` | `(4, 4)` array | `ValueError` |
| `(2, 2, 2, 2)` | `(0, 0)` array | `ValueError` |

Why it matters: `open_geotiff` builds the y/x coord arrays from the caller-supplied window indices. A clamped read returns a smaller array than the coords, so the user gets a `CoordinateValidationError` from xarray's internals instead of a clear xrspatial error. This is exactly the failure mode #1634 was filed for.

## Fix

Replace the clamp with the validator the other two paths already use:

```python
if window is not None:
    r0, c0, r1, c1 = window
    if (r0 < 0 or c0 < 0
            or r1 > vrt.height or c1 > vrt.width
            or r0 >= r1 or c0 >= c1):
        raise ValueError(
            f"window={window} is outside the VRT extent "
            f"({vrt.height}x{vrt.width}) or has non-positive size.")
else:
    r0, c0, r1, c1 = 0, 0, vrt.height, vrt.width
```

[...]

PR #1850, opened 2026-05-14
geotiff: default _vrt.read_vrt missing_sources to 'raise' (#1843)

### LINKED ISSUE #1843: VRT missing_sources default before stable release
## Current behaviour

`read_vrt` in `_vrt.py` defaults `missing_sources='warn'`. When a backing source can't be read, the call:

1. emits `GeoTIFFFallbackWarning`,
2. appends a record to `vrt.holes` (surfaced as `attrs['vrt_holes']` on the public DataArray), and
3. continues, leaving the destination region filled with the buffer's `fill` value (NaN for float, 0 for integer).

`missing_sources='raise'` and `XRSPATIAL_GEOTIFF_STRICT=1` both already raise.

## The concern

For an integer raster without a nodata sentinel, the zero-fill hole is indistinguishable from real zero-valued data unless the caller checks `attrs['vrt_holes']`. Before a stable release, that default is too lenient.

## Options

**(a) Switch default to `'raise'`.**
- Pro: safest contract. Matches the "reject up front" handling elsewhere in the module (#1697, #1737, #1784, band validation). Integer rasters stop silently looking correct on failure.
- Con: user-visible break. Callers wanting partial mosaics pass `missing_sources='warn'` (one kwarg).

**(b) Hybrid `'auto'` default.**
- Pro: only raises when the affected band lacks a nodata sentinel (the actually-ambiguous case). Float bands and integer-with-nodata bands keep the lenient path.
- Con: three-valued policy is harder to document and reason about. Doesn't fully solve the loss-of-provenance problem (the caller still has to inspect `vrt_holes` to know which source failed).

**(c) Keep `'warn'`.**
- Pro: no break. The strict env var and `vrt_holes` attr already give safety-conscious callers an opt-in.
- Con: defaults are what most callers use. Pre-stable is when defaults should tighten, not stay loose.

## Chosen direction: (a)

The module has been hardening toward stable with consistent up-front rejection of malformed input. A silent zero-fill on a `Byte` raster is the one place lenience still leaks through. The escape hatch already exists and is one kwarg long.

The env var `XRSPATIAL_GEOTIFF_STRICT=1` stays orthogonal (force-raise across the whole module).

## Scope

This issue covers the internal `read_vrt` in `xrspatial/geotiff/_vrt.py`. Aligning the public dispatcher default in `__init__.py` is a separate follow-up.

brendancol:
Closes #1843.

## Summary

- Flip the internal `_vrt.read_vrt` default `missing_sources` from `'warn'` to `'raise'` so an unreadable VRT source halts the call instead of silently leaving a zero-fill hole.
- Update docstring and one test (`test_missing_source_still_takes_lenient_warning_path` from #1784) that relied on the lenient default; it now passes `missing_sources='warn'` explicitly to exercise the opt-in path.
- Add `test_vrt_missing_sources_default_raise_1843.py` covering the new default, the explicit-warn escape hatch, and `XRSPATIAL_GEOTIFF_STRICT=1` precedence.

## Why

The lenient default produced a zero-fill hole on integer rasters without a nodata sentinel, indistinguishable from real zero pixels unless the caller checked `attrs['vrt_holes']`. The rest of the module already rejects malformed input up front (#1697, #1737, #1784, band validation). Before stable, the safer contract should be the default; lenience should be opt-in.

## Migration

Callers that want partial mosaics:

```python
read_vrt(path, missing_sources='warn')
```

The public `read_vrt` and `open_geotiff` wrappers in `__init__.py` still default to `'warn'` and forward the kwarg explicitly, so user-facing behaviour does not change in this PR. Aligning the public default is a separate follow-up.

`XRSPATIAL_GEOTIFF_STRICT=1` keeps its module-wide force-raise semantics regardless of this kwarg.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_vrt_missing_sources_default_raise_1843.py` (3 new tests)
- [x] `pytest xrspatial/geotiff/tests/test_geotiff_vrt_srcrect_validation_1784.py` (updated lenient-path test)
- [x] `pytest xrspatial/geotiff/tests/test_vrt_missing_sources_policy_1799.py` (existing explicit-kwarg tests still green)
- [x] `pytest xrspatial/geotiff/tests/test_open_geotiff_missing_sources_1810.py` (public dispatcher unchanged)
- [x] Full geotiff suite passes excluding three pre-existing unrelated failures (matplotlib deepcopy in `test_features.py`, GPU predictor tests, GPU tile size test)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Bug fixes and improvements
 - Refresh the geotiff mmap cache when a file is replaced under the same path so re-reads after an atomic-rename overwrite no longer return stale bytes
 - Decode TIFF predictor=3 un-transpose by file byte order so big-endian floating-point TIFFs read back exactly
+- Default internal `_vrt.read_vrt` `missing_sources` to `'raise'` so an unreadable VRT source no longer produces a silent zero-fill hole on integer rasters; pass `missing_sources='warn'` to opt back into the previous lenient behaviour (#1843)
 
 
 ### Version 0.9.9 - 2026-05-05
--- a/xrspatial/geotiff/_vrt.py
+++ b/xrspatial/geotiff/_vrt.py
@@ -852,7 +852,7 @@ def _apply_integer_sentinel_mask(arr, vrt, band):
 def read_vrt(vrt_path: str, *, window=None,
              band: int | None = None,
              max_pixels: int | None = None,
-             missing_sources: str = 'warn',
+             missing_sources: str = 'raise',
              parsed: VRTDataset | None = None,
              ) -> tuple[np.ndarray, VRTDataset]:
     """Read a VRT file by assembling pixel data from its source files.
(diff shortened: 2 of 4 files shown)


brendancol:
Updated the fallback comment block in _vrt.py to describe the new raise-by-default behavior and the explicit missing_sources='warn' opt-in. Thanks for catching the drift.

Issue #1664, opened 2026-05-12
Apply tile byte cap to local files + SSRF hardening on HTTP geotiff reads


brendancol:
**Describe the bug**

Hardening report — not an active CVE. Two paper-cuts in the COG reader become real risk when a service accepts user-supplied URLs or paths.

1. Tile byte cap is HTTP-only. In `xrspatial/geotiff/_reader.py:1408-1416` the `XRSPATIAL_COG_MAX_TILE_BYTES` cap is checked inside `_fetch_decode_cog_http_tiles` but not inside `_read_tiles` or `_read_strips`. A crafted local TIFF with a huge `TileByteCounts` or `StripByteCounts` still walks into `_decode_strip_or_tile` carrying a giant slice. Compressed payloads can be small and decompress (deflate/zstd/lzw) into hundreds of MB or GB, so the decompressor becomes the OOM trigger even though mmap slicing itself is bounded by file size. The current `test_local_path_unaffected_by_cap` even pins the broken behavior in place by asserting the cap does NOT apply locally — that needs flipping.

2. SSRF in `_HTTPSource`. `xrspatial/geotiff/_reader.py:406-503` accepts arbitrary URLs with no:
   - scheme allow-list (`file://`, `gopher://`, etc.)
   - host filtering — `localhost`, `127.0.0.1`, IPv6 loopback `::1`, link-local `169.254.0.0/16` (cloud IMDS), and RFC1918 private ranges all reachable
   - explicit connect/read timeouts (urllib defaults to infinite; urllib3 has no per-call timeout in our usage)
   - explicit redirect cap

For a service that takes `open_geotiff(url=user_input)` this is a textbook SSRF surface. `http://169.254.169.254/...` probes cloud metadata; `http://127.0.0.1:6379/` talks to internal Redis.

**Expected behavior**

- Local tile and strip readers apply the same `XRSPATIAL_COG_MAX_TILE_BYTES` cap the HTTP path uses. Keep the existing 256 MiB default so legitimate local reads still work; the env var stays the single tuning knob.
- `_HTTPSource.__init__` rejects non-http(s) schemes by default. `XRSPATIAL_GEOTIFF_ALLOWED_SCHEMES` opts back in.
- Hostname resolves through `socket.getaddrinfo` and any loopback, link-local, or RFC1918 private IP is rejected unless `XRSPATIAL_GEOTIFF_ALLOW_PRIVATE_HOSTS=1`.
[...]

PR #1851, opened 2026-05-14
geotiff: honour window and max_pixels for stripped HTTP reads (#1842)

### LINKED ISSUE #1842: Stripped HTTP TIFF reads ignore window and caller max_pixels
The stripped branch of `_fetch_decode_cog_http_tiles` in `xrspatial/geotiff/_reader.py` (around line 2003) calls `source.read_all()` and then slices the decoded array. Two issues:

1. A windowed HTTP read of a stripped multi-billion-pixel COG still downloads the whole raster. The tiled branch a few lines below explicitly promises otherwise: "A windowed HTTP read of a multi-billion-pixel COG only allocates the window." The stripped path silently violates that.

2. The caller's `max_pixels` is dropped. `_read_strips` runs with its default `MAX_PIXELS_DEFAULT` (1 billion), so a 50x50 window read with `max_pixels=2500` against a 100k x 100k stripped image raises `PixelSafetyLimitError` if the file declares over a billion pixels, and gets no safety check otherwise. Neither matches the tiled branch behaviour from #1664 / #1823.

Fix: for stripped HTTP files with a non-None window, fetch only the byte ranges of strips intersecting `[r0:r1]` via `read_ranges_coalesced` (same as the tiled path), decode them, and place the result into a windowed output. Apply `max_pixels` to the windowed `(r1-r0)*(c1-c0)*samples` allocation, not the full image. When window is None, keep the full read but pass `max_pixels` into `_read_strips` so the dimension check honours the caller's cap.

Related: #1664, #1823.

brendancol:
## Summary

Fixes #1842.

- `_fetch_decode_cog_http_tiles` had a `not ifd.is_tiled` branch that called `source.read_all()` and sliced the result. That dropped two safety contracts the tiled branch had from #1664 and #1823: (1) the comment a few lines below promises windowed HTTP reads "only allocate the window", but stripped COGs were downloading the whole raster anyway; (2) the caller's `max_pixels` was silently swapped for `MAX_PIXELS_DEFAULT` (1B), so a 50x50 window read with `max_pixels=2500` against a huge stripped file went through unchecked.
- Move the work into a new `_fetch_decode_cog_http_strips` helper. For windowed reads it computes which strips intersect `[r0:r1]`, fetches only those byte ranges via `read_ranges_coalesced` (same coalescing logic as the tiled path), decodes each via `_decode_strip_or_tile`, and places each strip's intersection with the window into the output. Per-strip byte cap (#1664), chunky vs planar layouts, and sparse strips all follow `_read_strips`. For full-image reads the path stays on `read_all` + `_read_strips`, but `max_pixels` is now threaded through.

## Test plan

- [x] new regression tests in `xrspatial/geotiff/tests/test_http_stripped_window_max_pixels_issue_A_1842.py`:
  - byte-range coverage: a window covering one strip issues exactly one strip GET at that strip's offset, and `read_all` is never called.
  - windowed `max_pixels` honoured: a 50x50 window with `max_pixels=2500` succeeds against a 65,536-pixel source; the same window with `max_pixels=2499` raises `PixelSafetyLimitError`.
  - full-image read still capped: `window=None` with `max_pixels=100` against a 65,536-pixel source raises `PixelSafetyLimitError`.
  - parametrised round-trip parity across six window shapes (strip-boundary spans, tiny straddling windows, last strip, full image).
- [x] full `xrspatial/geotiff/tests/` suite passes apart from 11 pre-existing failures on main (matplotlib + Python 3.14 deepcopy recursion, GPU tests without a GPU, an unrelated tile-size validation test); none of them touch the HTTP path.
- [x] HTTP-adjacent suites (`test_http_*`, `test_cog_http_*`, `test_reader.py`, `test_local_tile_byte_cap_1664.py`, `test_ssrf_hardening_1664.py`) all pass: 106/106.

### DIFF
--- a/xrspatial/geotiff/_reader.py
+++ b/xrspatial/geotiff/_reader.py
@@ -2029,27 +2029,21 @@ def _fetch_decode_cog_http_strips(
     if rps is None or rps <= 0:
         raise ValueError(f"Invalid RowsPerStrip: {rps!r}")
 
-    # Per-strip compressed-byte cap (#1664). Mirrors the local strip path
-    # and the tiled HTTP path: a crafted ``StripByteCounts`` entry can
-    # request an unbounded HTTP Range GET or decompress a few KiB into
-    # gigabytes. Apply the cap before any fetch list is built.
+    # Per-strip compressed-byte cap (#1664). A crafted ``StripByteCounts``
+    # entry can request an unbounded HTTP Range GET or decompress a few
+    # KiB into gigabytes. The cap applies to strips we actually fetch:
+    # - Full-image path: validated inside ``_read_strips`` over every
+    #   strip (full file is materialised regardless).
+    # - Windowed path: validated inside the fetch-range loop below so a
+    #   small window only fails on strips it intersects -- mirrors the
+    #   tiled HTTP path's per-tile check (#1851).
     max_tile_bytes = _max_tile_bytes_from_env()
-    for _strip_idx, _bc in enumerate(byte_counts):
-        if _bc > max_tile_bytes:
-            raise ValueError(
-                f"TIFF strip {_strip_idx} declares "
-                f"StripByteCount={_bc:,} bytes, which exceeds the "
-                f"per-strip safety cap of {max_tile_bytes:,} bytes. "
-                f"The file is malformed or attempting denial-of-service. "
-                f"Override via XRSPATIAL_COG_MAX_TILE_BYTES if this file "
-                f"is legitimate."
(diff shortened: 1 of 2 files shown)


brendancol:
Implemented both: per-strip byte cap now lives inside the windowed fetch-range loop (mirrors the tiled path), and decoded strip size goes through a MAX_PIXELS_DEFAULT guard before _decode_strip_or_tile. Regression tests added for both cases.

Issue #1833, opened 2026-05-14
VRT writer maps 64-bit integer rasters to wrong dtype


brendancol:
**Describe the bug**

\`write_vrt\` in \`xrspatial/geotiff/_vrt.py\` emits the wrong \`dataType\` for 64-bit integer source rasters. The dtype lookup at line 1366-1368 has no entry for \`bps=64\`:

\`\`\`python
elif sf == 2:
    vrt_dtype_name = {8: 'Int8', 16: 'Int16', 32: 'Int32'}.get(bps, 'Int32')
else:
    vrt_dtype_name = {8: 'Byte', 16: 'UInt16', 32: 'UInt32'}.get(bps, 'Byte')
\`\`\`

Signed 64-bit falls back to \`Int32\` and unsigned 64-bit falls back to \`Byte\`. The VRT reader has explicit \`Int64\` and \`UInt64\` support (\`_vrt.py:183-184\`), so the writer is the only side that drops the precision.

**Reproduce**

Write a uint64 raster via \`to_geotiff(da, "out.vrt")\`. The emitted VRT declares \`dataType=\"Byte\"\` and reading it back returns uint8 with values truncated to 255.

**Expected behavior**

64-bit integer rasters should round-trip through VRT without dtype loss. Signed 64-bit should emit \`Int64\` and unsigned 64-bit should emit \`UInt64\`.

**Fix**

Add \`64: 'Int64'\` to the signed branch and \`64: 'UInt64'\` to the unsigned branch of the lookup.

PR #1852, opened 2026-05-14
geotiff: parse GDAL_NODATA as int first to keep 64-bit sentinels exact (#1847)

### LINKED ISSUE #1847: geotiff: extract_geo_info float-only nodata parse loses uint64 max / int64 max sentinels
## Summary

The reader's GDAL_NODATA parsing in `_geotags.extract_geo_info` (and the related helpers `_reader._resolve_masked_fill` / `_reader._sparse_fill_value`) parses the nodata string via `float()` unconditionally. For 64-bit integer rasters whose declared sentinel is `uint64` max (`2**64 - 1`) or `int64` max (`2**63 - 1`), the float64 round-trip rounds the sentinel to a value that exceeds the dtype's representable range. The downstream integer-mask gate `info.min <= int(nodata) <= info.max` then rejects the cast and the sentinel pixel survives as a literal valid integer rather than being masked to NaN.

Repro:

```python
import numpy as np
import xarray as xr
import xrspatial.geotiff as g
import tempfile, os

    arr = np.full((16, 16), 100, dtype=np.uint64)
    arr[0, 0] = 2**64 - 1  # uint64 max as sentinel
    da = xr.DataArray(arr, dims=['y','x'])
    p = os.path.join(d, 't.tif')
    g.to_geotiff(da, p, nodata=2**64-1)
    out = g.open_geotiff(p)
    print(out.dtype, out.values[0, 0])  # uint64 18446744073709551615
    # expected: float64 NaN, after sentinel-to-NaN promotion
```

Same failure pattern for `np.int64` with `nodata = 2**63 - 1`. Both values are unrepresentable in float64; the nearest float rounds up by 1 ULP, which then fails the dtype-range check.

Affected sites:
- `xrspatial/geotiff/_geotags.py:619` (`extract_geo_info` — main GDAL_NODATA parse)
- `xrspatial/geotiff/_reader.py:1166` (`_resolve_masked_fill` — LERC fill)
- `xrspatial/geotiff/_reader.py:1299` (`_sparse_fill_value` — SPARSE_OK fill)

This is the same class of bug `_vrt._parse_band_nodata` (PR #1833) fixed for the VRT XML parse path: try `int()` first, fall back to `float()` only for NaN/Inf/scientific/fractional. The VRT-side fix did not extend to the TIFF source-of-truth parser, so:

* Direct `open_geotiff(uint64.tif)` with nodata=uint64 max → sentinel lost.
* `write_vrt([uint64.tif])` stringifies the float-parsed sentinel into XML as `1.8446744073709552e+19`, which the VRT reader subsequently rejects as out-of-range.

Categories: Cat 2 (NaN propagation: sentinel pixel survives unmasked) + Cat 5 (backend inconsistency: VRT XML path handles 64-bit sentinels via `_parse_band_nodata` but the TIFF path does not).

## Fix sketch

Lift the int-first parse from `_vrt._parse_band_nodata` into a shared helper and reuse it across the three TIFF-side sites. The helper takes the raw string and the target dtype; it tries `int(text)`, falls back to `float(text)` for NaN/Inf/fractional/scientific, and returns the parsed value at full Python int precision when possible.

## Test plan

- [ ] uint64 nodata=2**64-1: NaN-mask fires, attrs['nodata'] surfaces 2**64-1 (int)
- [ ] int64 nodata=2**63-1: NaN-mask fires, attrs['nodata'] surfaces 2**63-1
- [ ] int64 nodata=-2**63 (representable): no regression
- [ ] uint16 nodata=65535 (fits float64): no regression
- [ ] uint64 source → write_vrt → read_vrt: sentinel round-trips

brendancol:
## Summary

Closes #1847.

`extract_geo_info`, `_resolve_masked_fill`, and `_sparse_fill_value` parsed the GDAL_NODATA tag through `float()` unconditionally. uint64 max (`2**64 - 1`) and int64 max (`2**63 - 1`) are not exactly representable in float64; the nearest float sits one ULP above the dtype's max so the downstream `info.min <= int(nodata) <= info.max` gate rejected the cast and the sentinel pixel survived as a literal valid integer rather than being masked to NaN. `write_vrt` then stringified the float-parsed sentinel into the VRT XML as `1.8446744073709552e+19`, which the VRT reader subsequently rejected as out-of-range -- so the bug propagated through the whole 64-bit pipeline rather than just the eager TIFF read.

The VRT XML parse path was already int-aware via `_vrt._parse_band_nodata` (PR #1833). This change lifts the int-first parse into a shared helper `_parse_nodata_str` in `_geotags.py` and routes the three TIFF-side sites through it. The helper tries `int(text)` first to preserve full precision, falls back to `float(text)` for NaN / Inf / scientific notation / fractional values. Downstream gates already handle integer values transparently because `np.isfinite(int)` works and `int(int)` is a no-op, so no call sites change.

Categories: Cat 2 (NaN propagation: sentinel pixel survived as a literal valid number on all four backends) + Cat 5 (backend inconsistency: VRT XML parse handled 64-bit sentinels but the TIFF parse fed by `write_vrt` did not).

## Repro before fix

```python
import xrspatial.geotiff as g

    arr = np.full((16, 16), 100, dtype=np.uint64)
    arr[0, 0] = 2**64 - 1
    da = xr.DataArray(arr, dims=['y','x'])
    p = os.path.join(d, 't.tif')
    g.to_geotiff(da, p, nodata=2**64-1)
    out = g.open_geotiff(p)
    print(out.dtype, out.values[0, 0])
    # Before: uint64 18446744073709551615  (sentinel survives unmasked)
    # After:  float64 nan
```

## Test plan

- [x] Unit-level `_parse_nodata_str` matrix (int branch + float fallback + edge cases)
- [x] Eager `open_geotiff` masks uint64 max / int64 max to NaN
- [x] int64 min / uint16 max / int32 -9999 / float -9999.0 regression guards still pass
- [x] `read_geotiff_dask(chunks=)` masks 64-bit sentinels via the windowed reader
- [x] `write_vrt -> read_vrt` round-trip preserves the integer XML literal and masks the sentinel
- [x] GPU path masks 64-bit sentinels (fix is upstream of per-backend code)
- [x] All 2434 non-stale geotiff tests pass; 1 pre-existing failure unrelated (`test_size_param_validation_gpu_vrt_1776.test_tile_size_positive_works` asserts pre-#1767 tile_size=4 behaviour)


brendancol:
Addressed all three: unused tempfile import removed, GPU gating switched to the _gpu_only / _HAS_GPU helper used by test_gpu_nodata_1542.py, and _parse_nodata_str signature widened to str | None. I also widened GeoInfo.nodata from float | None to int | float | None since that's the one downstream type that now actually receives int values from the helper. The _resolve_masked_fill / _sparse_fill_value callers don't carry return annotations and their nodata_str parameter is already str | None, so no further churn needed there.

PR #1871, opened 2026-05-14
geotiff: reject invalid PlanarConfiguration values

### LINKED ISSUE #1870: geotiff: invalid PlanarConfiguration values accepted as chunky
## Describe the bug

`IFD.planar_config` in `xrspatial/geotiff/_header.py:237` returns the raw tag value with a default of 1. The TIFF 6.0 spec defines only two valid values: 1 (Chunky) and 2 (Planar). The readers special-case 2 and treat everything else as chunky:

- `_reader.py:1685`
- `_reader.py:1881`
- `_reader.py:2318`
- `_reader.py:2519`
- `__init__.py:3041`

A malformed file with `PlanarConfiguration=0`, `3`, `255`, etc. silently decodes under an assumed chunky layout instead of being rejected.

## Expected behavior

Validate `planar_config` at the header layer. Reject values outside {1, 2} with a typed `ValueError` so callers don't decode garbage under an assumed layout. Falling back to the default (1) when the tag is absent is fine; only an explicit non-{1,2} value should raise.

## Additional context

Severity: MEDIUM (no silent corruption when the layout happens to match, but no diagnostic when it doesn't).

brendancol:
Closes #1870.

## Summary

`IFD.planar_config` returned the raw tag value with a default of 1, so a malformed file with `PlanarConfiguration=0`, `3`, or `255` silently decoded under the default chunky layout. TIFF 6.0 defines only `1` (Chunky) and `2` (Planar).

Validate in the property: missing tag still defaults to `1` per spec; any explicit value outside `{1, 2}` raises `ValueError`. Also handle the length-1 tuple wrapping that the orientation property already accounts for.

## Test plan

- [x] New `TestPlanarConfigValidation` covers chunky, planar, zero, three, large value, missing tag, uniform tuple, invalid tuple.
- [x] `test_reader.py`, `test_mixed_bps.py`, `test_dask_planar_multiband.py`, `test_geotiff_planar_strip_truncation_1782.py`, `test_predictor_multisample.py`, `test_features.py` (192 tests) pass. The one pre-existing failure in `test_features.py::TestPalette::test_xrs_plot_with_palette` is the unrelated tight_layout recursion tracked separately.

### DIFF
--- a/xrspatial/geotiff/_header.py
+++ b/xrspatial/geotiff/_header.py
@@ -236,7 +236,19 @@ class IFD:
 
     @property
     def planar_config(self) -> int:
-        return self.get_value(TAG_PLANAR_CONFIG, 1)
+        # TIFF 6.0 defines only 1 (Chunky) and 2 (Planar). Treating any
+        # other value as chunky would silently decode a malformed file
+        # under an assumed layout. Missing tag defaults to 1 per spec.
+        v = self.get_value(TAG_PLANAR_CONFIG, 1)
+        if isinstance(v, tuple):
+            v = v[0] if v else 1
+        v = int(v)
+        if v not in (1, 2):
+            raise ValueError(
+                f"Invalid PlanarConfiguration: {v}. TIFF 6.0 allows only "
+                f"1 (Chunky) or 2 (Planar)."
+            )
+        return v
 
     @property
     def jpeg_tables(self) -> bytes | None:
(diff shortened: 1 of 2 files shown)


brendancol:
Thanks for catching the type coercion gap. You're right that a malformed TIFF declaring tag 284 as FLOAT/DOUBLE would resolve to a Python float, and `int(1.5) == 1` would slip through as Chunky.

Pushed 0623639: reject non-`int` values (and exclude `bool`, which is an `int` subclass but never legitimately a TIFF tag value) before the `(1, 2)` membership check. Added regression tests for the float and bool cases.

PR #1877, opened 2026-05-14
geotiff: real out-of-core read_geotiff_gpu(chunks=...) with optional GDS path

### LINKED ISSUE #1876: geotiff: read_geotiff_gpu(chunks=...) is not actually out-of-core
## Describe the bug

`read_geotiff_gpu(chunks=...)` and `open_geotiff(gpu=True, chunks=...)` both decode the entire raster into a single CuPy array first and call `.chunk()` on the resulting DataArray afterward (`xrspatial/geotiff/__init__.py:3328`). The docstring at `__init__.py:2740` advertises "out-of-core GPU pipelines" via `chunks=`, which implies lazy decoding.

For a raster that does not fit in GPU memory, the eager materialization runs out of memory before the user ever sees a Dask graph. The chunking is purely cosmetic past that point.

## Expected behavior

Either:

1. **Decode lazily** by building a Dask graph that calls per-chunk GPU decoders, so peak GPU memory is bounded by chunk size.
2. **Be honest about it** in the docstring and emit a runtime warning when a `chunks=` value is small relative to the raster, so users do not assume out-of-core semantics.

Option 2 is the pragmatic short-term fix. Option 1 is real lazy GPU I/O, a larger project.

## Additional context

Severity: HIGH for users with rasters near GPU memory size. The promise in the docstring is doing the load-bearing work; until that is either delivered on or retracted, `chunks=` is a foot-gun.

brendancol:
Closes #1876.

## Summary

The previous behaviour decoded the full raster into a single CuPy array and then called `.chunk()` on the resulting DataArray. Peak GPU memory was the whole-image size regardless of chunk size, so rasters that did not fit on the device OOM'd before the chunking ever took effect even though the docstring promised an out-of-core pipeline.

This PR replaces the post-decode chunk wrap with a real Dask+CuPy graph. Two paths produce the same shape of result:

1. **Direct disk->GPU** when KvikIO is available and the file qualifies (local + tiled + chunky + no sparse tiles + orientation=1 + photometric != 0). Each chunk task picks just the tiles overlapping its window, runs them through `gpu_decode_tiles_from_file` (which tries KvikIO GDS first, then falls back to mmap + `gpu_decode_tiles` per-window), and crops to the chunk window on device. When GDS hardware is present the file bytes never traverse host memory.

2. **CPU decode + `cupy.asarray`** for everything else: HTTP/fsspec sources, no KvikIO, planar=2, sparse tiles, MinIsWhite photometric, non-trivial orientation, stripped layouts. Reuses `read_geotiff_dask`'s well-tested per-chunk windowed delayed graph and maps each numpy block through `cupy.asarray` to upload it. Peak device memory is still one chunk; the trade-off is per-chunk CPU decode rather than GDS DMA.

Both paths bound peak device memory to one chunk. `chunks=None` keeps the eager whole-image GPU decode path for files that fit on device.

The misleading `RuntimeWarning` from the previous version is dropped and the docstring is rewritten.

## Test plan

- [x] `tests/test_gpu_chunks_out_of_core_1876.py` (10 tests): cupy-backed dask chunks, eager vs lazy value equality, attrs survive, `open_geotiff(gpu=True, chunks=...)` propagates, GDS spy fires once per chunk when KvikIO is available, no-KvikIO fallback path exercised (and direct path NOT called), window subset, multi-band, single-band `band=k` selection.
- [x] Full `xrspatial/geotiff/tests/` suite: 2633 passing, 7 skipped. The handful of pre-existing failures (`test_features.py::TestPalette`, `test_predictor2_big_endian_gpu_1517`, `test_size_param_validation_gpu_vrt_1776::test_tile_size_positive_works`) reproduce on `main` and are tracked separately.


brendancol:
Reworked this from a docstring-and-warning patch into a real out-of-core implementation per your direction.

New behaviour for ``read_geotiff_gpu(chunks=N)``:

- **KvikIO + qualifying file** (local + tiled + chunky + no sparse + orientation=1 + non-MinIsWhite): each chunk task picks the tiles overlapping its window, DMA's them disk->GPU via GDS, and crops on device. Bytes never traverse host memory.
- **Anything else**: per-chunk CPU windowed decode (via ``read_geotiff_dask``) then ``cupy.asarray`` per block. Still bounded to one chunk in device memory.

``chunks=None`` keeps the eager whole-image GPU decode for files that fit on the device.

Spy-based test verifies the direct path fires exactly once per chunk when KvikIO is available, and a separate test monkey-patches ``importlib.util.find_spec`` to confirm the fallback path takes over and the direct path is not touched.

brendancol:
The two Copilot review comments above are against the earlier docstring+warning version of this branch (commit 4e8c563) and no longer apply to the current state (commit e97a2b6, "real out-of-core read_geotiff_gpu(chunks=...)").

1. The `open_geotiff` docstring's "Dask+CuPy for out-of-core GPU pipelines" (now at `xrspatial/geotiff/__init__.py:575`) is accurate: this PR makes `read_geotiff_gpu(chunks=...)` actually out-of-core, so the line on `open_geotiff` no longer overpromises.
2. The redundant `import warnings` from the prior version is gone: the rework removed the warning entirely in favor of a real Dask+CuPy graph, so there is no local `import warnings` to drop.

PR #1878, opened 2026-05-14
fix: validate PlanarConfiguration tag against TIFF 6.0 spec (#1870)

### LINKED ISSUE #1870: geotiff: invalid PlanarConfiguration values accepted as chunky

rmotgi1227:
## Summary

Closes #1870.

`IFD.planar_config` in `xrspatial/geotiff/_header.py` returned the raw tag value without validation. TIFF 6.0 defines exactly two valid `PlanarConfiguration` values — **1** (Chunky/Interleaved) and **2** (Planar/Separated). Any other explicit value (0, 3, 255, …) was silently treated as chunky, potentially yielding corrupt output with no diagnostic.

### Changes

**`xrspatial/geotiff/_header.py`**
- `IFD.planar_config` now validates the tag value.
- An explicit value outside `{1, 2}` raises a typed `ValueError` that names the offending value.
- A missing tag still falls back to the spec default (`1`, Chunky) — no change in the happy path.

**`xrspatial/geotiff/tests/test_geotiff_planar_config_validation_1870.py`** (new)
- Parametrized test covering invalid values `0`, `3`, `10`, and `255` — each must raise `ValueError` and include the bad value in the message.
- Tests that `PlanarConfiguration=1` opens without error.
- Test that a file with **no** `PlanarConfiguration` tag opens without error (default is 1 per spec).

### Test results

```
6 passed in 0.03s
```

## Test plan

- [x] `python3 -m pytest xrspatial/geotiff/tests/test_geotiff_planar_config_validation_1870.py -v` — all 6 pass
- [x] Verified existing strip/tile read tests still pass
- [x] Manual check: `open_geotiff` on a real GeoTIFF with no PlanarConfiguration tag reads correctly

### DIFF
--- a/xrspatial/geotiff/_header.py
+++ b/xrspatial/geotiff/_header.py
@@ -236,7 +236,18 @@ class IFD:
 
     @property
     def planar_config(self) -> int:
-        return self.get_value(TAG_PLANAR_CONFIG, 1)
+        v = self.get_value(TAG_PLANAR_CONFIG, 1)
+        if isinstance(v, tuple):
+            v = v[0] if v else 1
+        v = int(v)
+        if v not in (1, 2):
+            raise ValueError(
+                f"Invalid PlanarConfiguration tag value {v!r}. "
+                "TIFF 6.0 only defines 1 (Chunky/Interleaved) and 2 "
+                "(Planar/Separated); a file with an explicit unsupported value "
+                "cannot be decoded safely."
+            )
+        return v
 
     @property
     def jpeg_tables(self) -> bytes | None:
--- a/xrspatial/geotiff/tests/test_geotiff_planar_config_validation_1870.py
+++ b/xrspatial/geotiff/tests/test_geotiff_planar_config_validation_1870.py
@@ -1,0 +1,149 @@
+"""Regression tests for issue #1870.
+
+The TIFF 6.0 spec defines exactly two valid PlanarConfiguration values:
+  1 - Chunky (interleaved, the default)
+  2 - Planar (separated, one plane per sample)
+
+Prior to this fix, ``IFD.planar_config`` returned the raw tag value
+without validation. A malformed file carrying ``PlanarConfiguration=0``,
+``3``, ``255``, etc. would silently decode under an assumed chunky layout
+instead of being rejected, potentially producing garbage output.
+
+These tests pin the contract:
+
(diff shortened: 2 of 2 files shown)


brendancol:
@rmotgi1227 Welcome and thanks for the PR.  Any chance I could get you to fill out a "New contributor" issue template which can access from https://github.com/xarray-contrib/xarray-spatial/issues  > New issue > New contributor introduction

brendancol:
@rmotgi1227 Thanks for the submission, but there was a duplicate PR that won out over this PR with the following reasoning:

• Accept PR #1871 and close PR #1878.

  They solve the same core issue, but they are not equivalent:

  - #1871 is stronger: it rejects non-integer malformed values too, including cases like 1.5 that #1878 would silently
    coerce with int(1.5) == 1.
  - #1871 has better focused coverage in test_header.py: valid 1, valid 2, missing tag default, invalid numeric values,
    tuple handling, float rejection, and bool rejection.
  - #1878 is good but weaker as-is: it still uses int(v), has a larger bespoke TIFF builder for this issue, and its
    summary says it tests PlanarConfiguration=2, but the diff I saw does not include that test.
  - CI is fully green for #1871. #1878 was mostly green, but two Windows jobs were still in progress when checked.

  Because they are not the same, I would not apply the “prefer #1878 because new user” tie-breaker here. Close #1878
  with a note thanking the contributor and explaining that #1871 covers the same issue plus the non-integer malformed-
  tag edge case.

PR #1927, opened 2026-05-15
geotiff: share parsed VRTDataset across chunk tasks via dask.delayed (#1923)

### LINKED ISSUE #1923: geotiff: VRT chunked path embeds full VRTDataset in every chunk task
## Summary

`_read_vrt_chunked` in `xrspatial/geotiff/_backends/vrt.py` passes `parsed_vrt=vrt` directly to each per-chunk `dask.delayed` call. Without wrapping the dataset in a single shared `dask.delayed`, dask embeds the entire VRTDataset (filenames, src/dst rects, per-source metadata) into every task graph entry.

## Impact

For a VRT with N sources and M chunks, the dask graph carries `N * M` copies of the source metadata. A 1000-source mosaic split into 1000 chunks builds a graph that holds ~57 MB of redundant per-task `_Source` records on the driver (measured: 60440 bytes per pickled `VRTDataset` x 1000 tasks). Under distributed/process schedulers the same payload also serialises N copies over the wire.

```
VRTDataset 1000 sources: 60440 bytes (59.0 KB)
Embedded in 1000 chunk tasks (no delayed wrap): 57.6 MB driver graph
With dask.delayed(vrt, pure=True): 59.0 KB total
```

## Existing pattern

The sibling `read_geotiff_dask` already solves the same problem for COG HTTP metadata at `_backends/dask.py:142-147`:

```python
# Wrap the parsed metadata in a single dask Delayed so every
# window task takes it as a graph input, not a Python closure.
# Without this, the (TIFFHeader, IFD) pair -- which can carry
# multi-million-entry TileOffsets/TileByteCounts tuples on
# large COGs -- would be embedded in each chunk task and
# serialised N times under distributed/process schedulers.
http_meta_key = dask.delayed(http_meta, pure=True)
```

The GDS chunked GPU path at `_backends/gpu.py:1035` does the same with `(offsets, byte_counts)`. The VRT chunked path is the outlier.

## Fix

Wrap the parsed `VRTDataset` in `dask.delayed(vrt, pure=True)` before the per-chunk loop in `_read_vrt_chunked`, and pass that delayed reference as `parsed_vrt=` to each task. `_vrt_chunk_read` already accepts the parsed value as a function argument, so no changes are required on the consumer side.

## Audit context

Found during performance sweep against the geotiff subpackage on 2026-05-15.

brendancol:
## Summary

- `_read_vrt_chunked` in `xrspatial/geotiff/_backends/vrt.py` passed the parsed `VRTDataset` as a plain kwarg to each per-chunk `dask.delayed` call, embedding the full source list (filenames, src/dst rects, per-source nodata) into every task's kwargs.
- For an N-source VRT split into M chunks the graph carried N*M copies of the source metadata; under distributed/process schedulers each task pickle shipped that payload independently.
- Wrap the dataset in `dask.delayed(vrt, pure=True)` once and thread that single shared reference through `parsed_vrt=`, matching the `http_meta_key` pattern in `_backends/dask.py:147` and the `meta_key` pattern in `_backends/gpu.py:1035`.

## Numbers

Structural pre-fix check on a 4x4 source VRT split into 64 chunks: 64 of 64 `_vrt_chunk_read` tasks held an inline `VRTDataset` in `kwargs['parsed_vrt']`.

At the documented `_MAX_VRT_DASK_CHUNKS` cap a 1000-source VRT split into 1000 chunks builds a graph that embeds ~57 MB of redundant source metadata (60 KB pickled `VRTDataset` x 1000 task copies). Post-fix the graph holds one shared copy.

Closes #1923.

## Test plan

- [x] New `test_vrt_chunked_shared_dataset_1923.py` adds three tests:
  - `test_vrt_chunked_dataset_is_shared_graph_input`: walks every `_vrt_chunk_read` task in the dask graph and asserts none of them embed an inline `VRTDataset` in `kwargs['parsed_vrt']`.
  - `test_vrt_chunked_decode_unchanged_after_shared_wrap`: confirms decoded pixels match the eager `read_vrt` baseline.
  - `test_vrt_chunked_band_kwarg_still_validates`: confirms the wrap does not change band validation error semantics.
- [x] 438 existing VRT tests under `xrspatial/geotiff/tests/` pass; the 1 unrelated `tile_size=4` validator failure (`test_size_param_validation_gpu_vrt_1776.py`) predates this change.


brendancol:

Issue #1933, opened 2026-05-15
geotiff: predictor=3 + integer dtype silently misdecoded on read


brendancol:
## Summary

The reader accepts a malformed IFD that claims `Predictor=3` (Floating-Point Predictor, TIFF Technical Note 3) paired with an integer `SampleFormat` (1=uint or 2=int) without complaint. `_apply_predictor` in `xrspatial/geotiff/_reader.py` routes to `fp_predictor_decode` solely on the predictor tag, with no check that the data is actually float. The byte-swizzle unshuffle then runs on integer bytes, producing garbage pixel values that look like valid integers.

The writer side already rejects this combination: `_writer.py` `_resolve_predictor` raises `ValueError(\"predictor=3 (floating-point) requires float data, got dtype=...\")` when called with a non-float dtype. xrspatial-written files cannot hit this asymmetry, but external or adversarial files can.

## Reproduction

A synthetic IFD with `BitsPerSample=32`, `SampleFormat=1` (UINT), `Predictor=3`, and uncompressed strip data round-trips through `open_geotiff` without raising, and the returned uint32 array does not match the bytes that were written. The bug is silent: no warning, no exception, and a downstream consumer has no signal that the decoded data is wrong.

## Fix

Reject `Predictor=3` when `SampleFormat` is not `3` (float) on read. Either:

1. Raise `ValueError` (mirror the writer message: \"predictor=3 (floating-point) requires float data, got SampleFormat=<n>\"), or
2. Warn and fall back to `predictor=1` (no predictor).

Option (1) is cleaner and matches the writer's contract. The check belongs in `_apply_predictor` (or one step earlier where the IFD is first validated) so every backend path (eager numpy, dask, GPU) picks it up via the shared routine.

## Regression test

A unit test in `xrspatial/geotiff/tests/` that builds a small in-memory TIFF with the bad combination and asserts the new exception.

PR #1951, opened 2026-05-15
geotiff: GPU + dask+GPU coverage for predictor=3 + integer reject (#1933)

brendancol:
## Summary

Closes a Cat 1 HIGH backend-parity gap for issue #1933 (`_validate_predictor_sample_format`). The eager numpy and dask paths are covered by `test_predictor3_int_dtype_1933.py`; the cupy and dask+cupy paths had no targeted tests.

The validator is invoked at two GPU sites:

- `_backends/gpu.py:443` -- tiled eager GPU read
- `_backends/gpu.py:999` -- GDS chunked GPU read

A regression dropping either call would let malformed `predictor=3 + integer SampleFormat` tiled files decode silently to garbage bytes on GPU.

9 new tests, all passing on a CUDA host:

- `read_geotiff_gpu` on stripped + tiled malformed files
- `open_geotiff(gpu=True)` and `open_geotiff(chunks=, gpu=True)` dispatchers
- `read_geotiff_gpu(chunks=)` on stripped + tiled
- Legitimate `predictor=3 + float32` tiled file still round-trips on GPU (eager + dask+GPU)
- Error-message parity between GPU and eager paths

Mutation against the tiled GPU validator at `gpu.py:443` flipped the tiled-raises test red. Mutation against the GDS chunked validator at `gpu.py:999` flipped the chunked-tiled + chunked-dispatcher tests red.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_predictor3_int_dtype_gpu_1933.py` (9 passed)
- [x] `pytest xrspatial/geotiff/tests/test_predictor3_int_dtype_1933.py xrspatial/geotiff/tests/test_predictor3_int_dtype_gpu_1933.py` (19 passed)
- [x] Mutation: drop `_validate_predictor_sample_format` at gpu.py:443 -> `test_gpu_eager_tiled_raises` fails
- [x] Mutation: drop `_validate_predictor_sample_format` at gpu.py:999 -> `test_read_geotiff_gpu_chunked_tiled_raises` + `test_open_geotiff_chunks_gpu_dispatcher_raises` fail

### DIFF
--- a/xrspatial/geotiff/tests/test_predictor3_int_dtype_gpu_1933.py
+++ b/xrspatial/geotiff/tests/test_predictor3_int_dtype_gpu_1933.py
@@ -1,0 +1,299 @@
+"""GPU + dask+GPU backend coverage for issue #1933.
+
+#1933 added ``_validate_predictor_sample_format`` and wired it into
+every IFD-read site (eager numpy, dask, GPU tiled, GPU stripped). The
+eager and dask paths are covered by ``test_predictor3_int_dtype_1933``;
+this module closes the GPU coverage gap.
+
+The validator is invoked at two GPU sites:
+
+* ``_backends/gpu.py:443`` -- the tiled eager GPU read path. Reached when
+  the file is tiled and ``bps == file_dtype.itemsize * 8`` (so the
+  bps_mismatch fallback at line 358 does not take over).
+* ``_backends/gpu.py:999`` -- the GDS chunked GPU path
+  (``_read_geotiff_gpu_chunked_gds``). Reached when the file qualifies
+  for direct disk->GPU decode.
+
+The stripped GPU path falls back to CPU via ``_read_to_array`` and the
+CPU-side validator there fires; the dask+GPU non-GDS path delegates to
+``read_geotiff_dask`` which has its own validator (covered by the
+existing dask test). The two NEW call sites have no targeted tests.
+
+A regression dropping either of those two validator calls would let
+malformed predictor=3 + integer tiled files decode silently to
+garbage bytes on GPU. The eager-test asserts the error path is wired
+on CPU; this module asserts the GPU dispatcher path is wired too.
+"""
+from __future__ import annotations
+
+import importlib.util
+
+import numpy as np
+import pytest
+
+from xrspatial.geotiff._compression import COMPRESSION_NONE
+from xrspatial.geotiff._dtypes import LONG, SHORT, numpy_to_tiff_dtype
+from xrspatial.geotiff._header import (
+    TAG_BITS_PER_SAMPLE,
+    TAG_COMPRESSION,
+    TAG_IMAGE_LENGTH,
+    TAG_IMAGE_WIDTH,
+    TAG_PHOTOMETRIC,
+    TAG_PREDICTOR,
+    TAG_SAMPLE_FORMAT,
+    TAG_SAMPLES_PER_PIXEL,
+    TAG_STRIP_BYTE_COUNTS,
+    TAG_STRIP_OFFSETS,
+    TAG_ROWS_PER_STRIP,
+    TAG_TILE_BYTE_COUNTS,
+    TAG_TILE_LENGTH,
+    TAG_TILE_OFFSETS,
+    TAG_TILE_WIDTH,
+)
+from xrspatial.geotiff._writer import (
+    _assemble_standard_layout,
+    _write_stripped,
+)
+
+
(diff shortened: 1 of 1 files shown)


brendancol:
## PR Review: geotiff: GPU + dask+GPU coverage for predictor=3 + integer reject (#1933)

- None

- [ ] `test_read_geotiff_gpu_chunked_tiled_raises` (test_predictor3_int_dtype_gpu_1933.py:213) accepts the rejection from either the GDS validator at gpu.py:999 or the CPU dask fallback. On a host without KvikIO the GDS validator at line 999 is never exercised, so the mutation-kill claim in the PR description only holds on KvikIO-enabled CI. Consider gating the test on `kvikio` availability (`pytest.importorskip("kvikio")`) or adding a sibling test that does so, otherwise the line-999 call site is not actually pinned by the suite in the general case.
- [ ] `test_predictor3_float32_gpu_round_trip` and `test_predictor3_float32_dask_gpu_round_trip` (lines 245, 258) use `tifffile.imwrite` for the legitimate-file case while the malformed cases use the in-tree `_assemble_standard_layout`. Building the valid file through the same in-tree writer would remove the `tifffile`/`imagecodecs` skip dependency and keep the round-trip contract honest against the writer the rest of the suite exercises.

- [ ] `_build_predictor3_uint32_stripped_tiff` (line 70) duplicates the helper in `test_predictor3_int_dtype_1933.py:42`. A shared helper in `tests/conftest.py` would keep the two modules from drifting if the malformed-file shape needs to change.
- [ ] Module docstring (lines 11-18) hard-codes `gpu.py:443` and `gpu.py:999`. These will rot when the file is edited. A symbolic reference (the function name) would survive line renumbering.
- [ ] `test_gpu_error_message_matches_eager` (line 285) uses `==` on the full message. If the eager site ever appends file path or band info, both messages will diverge and the test fails for a non-regression. A `startswith` or shared substring check is more robust.

### What looks good
- Tiled vs stripped split is well motivated: the stripped path falls through to the CPU validator, so the tiled fixture is what actually exercises gpu.py:443.
- Both dispatchers (`open_geotiff(gpu=True)` and `open_geotiff(chunks=, gpu=True)`) are covered alongside the direct `read_geotiff_gpu` entrypoints.
- Tmp filenames are unique per test, no collisions.
- Skip guard via `cupy.cuda.is_available()` is correct and keeps the file inert on non-GPU CI.
- Test-only PR, no source changes, matches the deep-sweep coverage charter.

### Checklist
- [x] Algorithm matches reference/paper (validator behaviour, no algorithm change)
- [x] All implemented backends produce consistent results (parity test asserts identical message)
- [ ] NaN handling is correct (n/a, integer rejection)
- [x] Edge cases are covered by tests (tiled + stripped + dispatcher + chunked)
- [x] Dask chunk boundaries handled correctly (chunked tests use `chunks=` smaller than the array)
- [ ] Benchmark exists or is not needed (not needed, error path)
- [ ] README feature matrix updated (if applicable) (n/a)
- [x] Docstrings present and accurate (modulo the line-number rot nit)

brendancol:
## PR Review: geotiff: GPU + dask+GPU coverage for predictor=3 + integer reject (#1933) (re-review after rebase)

Rebased cleanly onto `origin/main` at `1e34c92`, picking up the Windows fix from `9ce0e60`. Local run of `xrspatial/geotiff/tests/test_predictor3_int_dtype_gpu_1933.py` passes (8 passed, 1 skipped on CPU-only host where libkvikio is unavailable).

### Blockers
- None

### Suggestions
- None

### Nits
- The `pytest.importorskip("kvikio")` call raises a `PytestDeprecationWarning` on hosts where the `kvikio` module is importable but its shared library is missing. Passing `exc_type=ImportError` would silence the warning ahead of pytest 9.1, but this is cosmetic.

### What looks good
- Test coverage now spans cupy and dask+cupy paths for the predictor=3 + integer SampleFormat reject behavior added in #1933.
- The in-tree writer keeps the round-trip tests self-contained and avoids relying on any external writer behavior.
- kvikio gating via `importorskip` keeps the GPU-only tests honest on CPU CI runners.

### Checklist
- [x] Rebase clean on `origin/main`
- [x] Tests pass locally
- [x] No code changes beyond rebase
- [x] Pushed with `--force-with-lease` to fork

brendancol:
Rebased onto current main at `4dec5b6` to pick up the Python 3.14 CI matrix limit (`b1579f8`).

Issue #1934, opened 2026-05-15
geotiff: _apply_nodata_mask_gpu allocates instead of mutating in-place


brendancol:
## Summary

`_apply_nodata_mask_gpu` in `xrspatial/geotiff/_backends/_gpu_helpers.py` uses `cupy.where(mask, nan, arr_gpu)` to replace nodata sentinels with NaN. `cupy.where` allocates a fresh output buffer the same shape as the input array. Writing the NaN into the existing buffer with `cupy.putmask` (or boolean assignment) avoids the chunk-sized allocation and the corresponding device-to-device copy.

## Locations

- `_apply_nodata_mask_gpu` (line ~45 of `_gpu_helpers.py`):
  - Float path: `arr_gpu = cupy.where(arr_gpu == sentinel, nan, arr_gpu)`
  - Integer path: `arr_gpu = arr_gpu.astype(cupy.float64)` followed by `arr_gpu = cupy.where(mask, nan, arr_gpu)`

The integer path already allocates a new float64 buffer via `astype`; the trailing `cupy.where` then allocates a second buffer of the same size. Both can collapse into a single allocation followed by an in-place write.

## Call sites

All call sites pass a freshly decoded GPU buffer that is not shared with caller-visible state:

- `_backends/gpu.py:387` - `arr_gpu` returned from `read_to_array`'s GPU stripped path
- `_backends/gpu.py:684` - `arr_gpu` returned from the GPU tile decoder
- `_backends/gpu.py:1072` - per-chunk `arr` from `_decode_window_gpu_direct`

In-place mutation is safe at each.

## Proposed fix

Use `cupy.putmask(arr_gpu, mask, nan)` (or `arr_gpu[mask] = nan`) so the existing buffer is reused. Saves one chunk-sized device buffer per call on GPU-bound workloads with large chunks.

## Severity

LOW. Allocator pressure reduction rather than a correctness issue. Filed during the 2026-05-15 deep sweep on `geotiff` despite earlier classification as marginal.

PR #1952, opened 2026-05-15
geotiff: GPU writer overview loop uses cupy.putmask for in-place NaN rewrite (#1948)

### LINKED ISSUE #1948: GPU writer overview loop: redundant cupy.copy() before in-place NaN rewrite
### Where

`xrspatial/geotiff/_writers/gpu.py:498` in `write_geotiff_gpu`'s COG overview generation loop.

### What

After each call to `make_overview_gpu`, the loop rewrites NaN cells back to the nodata sentinel via:

```python
current = make_overview_gpu(current, method=overview_resampling, nodata=nodata)
cumulative_factor *= 2
if (nodata is not None
        and np.dtype(str(current.dtype)).kind == 'f'
        and not np.isnan(float(nodata))):
    nan_mask = cupy.isnan(current)
    if bool(nan_mask.any().item()):
        current = current.copy()
        current[nan_mask] = np.dtype(
            str(current.dtype)).type(nodata)
```

`make_overview_gpu` returns a freshly allocated cupy buffer at every call site:

- 2D path: `_block_reduce_2d_gpu` returns the result of `cupy.nanmean` / `cupy.nanmin` / `cupy.around(...).astype(...)` / `cropped[::2, ::2].copy()` / `cupy.asarray(cpu_result)` -- all fresh allocations.
- 3D path: `cupy.stack(bands, axis=2)` -- fresh allocation.

Nothing else aliases the buffer between the `make_overview_gpu` return and the in-place NaN rewrite, so the explicit `current = current.copy()` allocates a second chunk-sized GPU buffer just to mutate it.

### Why it matters

For an 8192x8192 float32 raster with 4 auto-generated overview levels, the extra allocations sum to roughly:

```
(2048x2048 + 1024x1024 + 512x512 + 256x256) * 4 bytes ~= 21 MB
```

per write. Modest, but the pattern is identical to the one fixed in #1934 for `_apply_nodata_mask_gpu`: replace the `copy()` + indexed write with `cupy.putmask(current, nan_mask, sentinel)` so the existing buffer is mutated in place and one chunk-sized device allocation per overview level is skipped.

### Suggested fix

```python
current = make_overview_gpu(current, method=overview_resampling, nodata=nodata)
cumulative_factor *= 2
if (nodata is not None
        and np.dtype(str(current.dtype)).kind == 'f'
        and not np.isnan(float(nodata))):
    nan_mask = cupy.isnan(current)
    if bool(nan_mask.any().item()):
        cupy.putmask(current, nan_mask,
                     np.dtype(str(current.dtype)).type(nodata))
```

The fix mirrors the in-place sentinel rewrite already used on the freshly-allocated GPU buffer at line 426 in the NaN->sentinel branch above, and the `_apply_nodata_mask_gpu` rewrite in #1934.

### Severity

LOW. Sentinel-poisoning on a multi-level pyramid is the only path that hits it, and the allocations are bounded by overview size (smaller than the full raster), but the fix is a one-line `cupy.putmask` substitution with the same correctness contract.

brendancol:
Closes #1948.

## Summary

The COG overview loop inside `write_geotiff_gpu` used to allocate a
fresh `current.copy()` before rewriting NaN cells back to the nodata
sentinel:

```python
current = make_overview_gpu(current, ...)
...
nan_mask = cupy.isnan(current)
if bool(nan_mask.any().item()):
    current = current.copy()
    current[nan_mask] = sentinel
```

`make_overview_gpu` returns a freshly allocated cupy buffer at every
call site (the 2-D path ends in `cupy.nan*` / `cupy.around(...).astype(...)`
/ `cropped[::2, ::2].copy()`; the 3-D path ends in `cupy.stack`), so
nothing aliased the buffer between the return and the in-place rewrite.
The `current.copy()` allocated a second chunk-sized GPU buffer per
overview level for no semantic gain.

The fix replaces the two-line rewrite with
`cupy.putmask(current, nan_mask, sentinel)` so the existing buffer is
mutated in place. Mirrors the in-place sentinel rewrite
`_apply_nodata_mask_gpu` adopted in #1934.

For an 8192x8192 float32 raster with 4 auto-generated overview levels,
the extra allocations sum to roughly 21 MB per write; the saving is
modest but the pattern aligns with #1934 and removes a redundant
device allocation per overview level.

## Test plan

- [x] `test_gpu_writer_overview_loop_uses_putmask_1948` -- structural
  guard against reintroducing the redundant copy.
- [x] `test_gpu_writer_cog_overview_sentinel_roundtrip_1948` -- COG
  write -> CPU read preserves the sentinel through every overview level
  (uses an all-sentinel quadrant so every 2x reduction still hits a
  fully-sentinel 2x2 block and triggers the rewrite branch).
- [x] `test_gpu_writer_overview_uses_make_overview_gpu_fresh_buffer_1948`
  -- contract: every overview method in `_block_reduce_2d_gpu` returns
  a cupy buffer whose `data.ptr` differs from the input.
- [x] Existing GPU writer + overview + COG suite (137 tests) passes
  unchanged.

Filed under the performance sweep run from `/deep-sweep performance`
on the geotiff module.

### DIFF
--- a/xrspatial/geotiff/_writers/gpu.py
+++ b/xrspatial/geotiff/_writers/gpu.py
@@ -479,6 +479,18 @@ def write_geotiff_gpu(data: xr.DataArray | cupy.ndarray | np.ndarray,
         # all-sentinel cell NaN back to the sentinel after each level
         # so the on-disk overview tiles still carry the sentinel value
         # external readers expect.
+        #
+        # The sentinel rewrite uses ``cupy.putmask`` rather than
+        # ``current.copy(); current[nan_mask] = sentinel`` because
+        # ``make_overview_gpu`` always returns a freshly allocated
+        # buffer (``_block_reduce_2d_gpu`` ends in ``cupy.nan*`` /
+        # ``cupy.around(...).astype(...)`` / ``cropped[::2, ::2].copy()``,
+        # and the 3-D path ends in ``cupy.stack``) so nothing else
+        # aliases the buffer between the return and the rewrite.
+        # Dropping the redundant ``copy()`` skips one chunk-sized GPU
+        # allocation per overview level. Mirrors the in-place sentinel
+        # rewrite ``_apply_nodata_mask_gpu`` adopted in #1934. See
+        # issue #1948.
         current = arr
         cumulative_factor = 1
         for target_factor in overview_levels:
@@ -495,9 +507,9 @@ def write_geotiff_gpu(data: xr.DataArray | cupy.ndarray | np.ndarray,
                         and not np.isnan(float(nodata))):
                     nan_mask = cupy.isnan(current)
                     if bool(nan_mask.any().item()):
-                        current = current.copy()
-                        current[nan_mask] = np.dtype(
-                            str(current.dtype)).type(nodata)
+                        cupy.putmask(
+                            current, nan_mask,
+                            np.dtype(str(current.dtype)).type(nodata))
             oh, ow = current.shape[:2]
             parts.append(_gpu_compress_to_part(current, ow, oh, samples))
 
--- a/xrspatial/geotiff/_writers/gpu.py
+++ b/xrspatial/geotiff/_writers/gpu.py
@@ -493,6 +493,17 @@ def write_geotiff_gpu(data: xr.DataArray | cupy.ndarray | np.ndarray,
         # issue #1948.
         current = arr
         cumulative_factor = 1
+        # ``make_overview_gpu`` preserves dtype, so the sentinel cast is
+        # loop-invariant. Hoist it (and the float/finite gate) out of the
+        # inner ``while`` to skip redundant per-level scalar work.
+        rewrite_nodata = (
+            nodata is not None
+            and np_dtype.kind == 'f'
+            and not np.isnan(float(nodata))
+        )
+        sentinel_scalar = (
+            np_dtype.type(nodata) if rewrite_nodata else None
+        )
         for target_factor in overview_levels:
             # Halve repeatedly until the cumulative decimation matches
             # the requested factor. Validation has already established
@@ -502,14 +513,10 @@ def write_geotiff_gpu(data: xr.DataArray | cupy.ndarray | np.ndarray,
                 current = make_overview_gpu(current, method=overview_resampling,
                                             nodata=nodata)
                 cumulative_factor *= 2
-                if (nodata is not None
-                        and np.dtype(str(current.dtype)).kind == 'f'
-                        and not np.isnan(float(nodata))):
+                if rewrite_nodata:
                     nan_mask = cupy.isnan(current)
                     if bool(nan_mask.any().item()):
-                        cupy.putmask(
-                            current, nan_mask,
-                            np.dtype(str(current.dtype)).type(nodata))
+                        cupy.putmask(current, nan_mask, sentinel_scalar)
             oh, ow = current.shape[:2]
             parts.append(_gpu_compress_to_part(current, ow, oh, samples))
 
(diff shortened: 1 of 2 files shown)


brendancol:
## PR Review: geotiff: GPU writer overview loop uses cupy.putmask for in-place NaN rewrite (#1948)

- None

- None

- [ ] `xrspatial/geotiff/_writers/gpu.py:499` -- the scalar cast `np.dtype(str(current.dtype)).type(nodata)` could be hoisted out of the inner `while` so it runs once per `target_factor` instead of once per halving step. The cost is negligible (host-side scalar construction) so this is purely cosmetic.
- [ ] `xrspatial/geotiff/tests/test_gpu_writer_overview_inplace_1948.py:79` -- the source-text guard scans a fixed 1500-char window after `make_overview_gpu(current`. If the surrounding code grows past that window the structural assertion will silently stop covering the right region. Anchoring on the next `parts.append(` line would be more robust.

### What looks good
- The aliasing argument holds: `make_overview_gpu` returns a fresh buffer on every reducer path, and the new `test_gpu_writer_overview_uses_make_overview_gpu_fresh_buffer_1948` codifies that contract across mean/nearest/min/max/median, so the in-place mutation is safe.
- The rewrite branch stays gated on `kind == 'f'`, so integer overviews (which can't carry NaN) never reach `putmask` and dtype semantics are preserved.
- The sentinel value is still cast to `current.dtype` before the write, matching the prior indexed-assignment semantics — no upcast surprise from `putmask`'s broadcasting.
- Correctness test uses an all-sentinel quadrant so every 2x reduction still hits a fully-sentinel 2x2 block and exercises the rewrite branch at every level, including a real COG round-trip through the CPU reader.

### Checklist
- [x] NaN handling is correct
- [x] README feature matrix updated (if applicable)

brendancol:
## PR Review: geotiff: GPU writer overview loop uses cupy.putmask for in-place NaN rewrite (#1948) (re-review after rebase)

Re-reviewing branch at `288ff83` after rebase onto current `main`. Two commits: the original `bbec25c` putmask swap and the follow-up `288ff83` that hoists the loop-invariant sentinel cast and anchors the test slice on `parts.append(`. CPU-side tests (`test_gpu_writer_overview_loop_uses_putmask_1948`) pass locally; the two GPU-gated tests skip without a device.

### Blockers
- None

### Suggestions
- None

### Nits
- None

### What looks good
- The hoist of `rewrite_nodata` and `sentinel_scalar` out of the inner `while` is safe. `make_overview_gpu` preserves dtype (the comment cites the four buffer-producing call sites inside `_block_reduce_2d_gpu` and the 3-D `cupy.stack` path), and `np_dtype` is captured once on line 399 from the source `arr`. The cast and the float/finite gate no longer repeat per level.
- The `cupy.putmask(current, nan_mask, sentinel_scalar)` call is correct against the same buffer `make_overview_gpu` just returned, with `bool(nan_mask.any().item())` still gating the rewrite so the kernel is skipped on overview levels that have no all-sentinel cells.
- The source-level test now anchors its window on the next `parts.append(` instead of a fixed character count, so the guard tracks the loop body even if surrounding code shifts.
- The third GPU test (`make_overview_gpu_fresh_buffer_1948`) covers all five reducer methods (`mean`, `nearest`, `min`, `max`, `median`) and asserts the returned device pointer differs from the input. That is the contract the in-place rewrite depends on, and it is now under test.
- The round-trip COG test stays green: the bottom-right quadrant stays all-sentinel so each 2x reduction still produces at least one NaN block that exercises the rewrite branch, and both overview levels read back with the sentinel preserved.

### Checklist
- [x] Rebased cleanly on `origin/main`
- [x] CPU test passes locally (`3 passed in 0.94s`; two GPU tests skip without a device)
- [x] No changes outside `_writers/gpu.py` and the new regression test
- [x] No public API or signature changes
- [x] Sentinel-rewrite semantics preserved (float dtype gate, finite nodata gate, dtype-cast sentinel)

brendancol:
Rebased onto current main at `19fd38f` to pick up the Python 3.14 CI matrix limit (`b1579f8`).

PR #1953, opened 2026-05-15
geotiff: preserve transform on 1xN / Nx1 georeferenced writes (#1945)

### LINKED ISSUE #1945: geotiff: 1xN / Nx1 georeferenced writes silently strip transform
## Summary

A DataArray that is georeferenced but degenerate along one spatial axis (1 row or 1 column, e.g. a single scanline or single profile) silently loses its transform when written with `to_geotiff`. `coords_to_transform` returns `None` whenever either coord array has length < 2, and the writer then falls through and produces a non-georeferenced TIFF. On read, the file comes back with integer pixel coords and no `transform` attribute, so the user has no signal that the spatial metadata was dropped.

## Reproduction

```python

da = xr.DataArray(
    np.arange(8, dtype='float32').reshape(1, 8),
    dims=('y', 'x'),
    coords={'x': np.linspace(-120.0, -113.0, 8), 'y': np.array([45.0])},
    attrs={'crs': 4326},
)
to_geotiff(da, '/tmp/strip.tif')

r = open_geotiff('/tmp/strip.tif')
print(r.coords['x'].values[:3], r.coords['x'].dtype)  # [0 1 2] int64
print(r.attrs.get('transform'))                       # None
```

The same thing happens for Nx1.

## Affected code

`xrspatial/geotiff/_coords.py:277`:

```python
if len(x) < 2 or len(y) < 2:
    return None
```

`xrspatial/geotiff/_writers/eager.py:516-519` calls `_transform_from_attr` first, then falls back to `_coords_to_transform`. If both return `None`, the writer proceeds without raising and produces a non-georeferenced TIFF. The GPU writer at `gpu.py:348-350` has the same fall-through.

## Fix direction

Two parts.

1. In `coords_to_transform`, don't bail when only one axis has length 1. The non-degenerate axis still pins a pixel size, and the file's row/column count fixes the other extent; the writer only needs an origin and a pixel size per axis. For the 1x1 case there is no way to recover pixel size from coords alone, so prefer `attrs['transform']` when the caller supplied one and otherwise raise a clear `ValueError`.

2. In the eager, dask, and GPU writer paths, when coord arrays are present but no transform can be derived and no `attrs['transform']` was supplied, raise rather than silently writing a non-georeferenced file. A degenerate-shape array with real spatial coords expresses intent to georeference, and the writer should fail closed instead of dropping metadata on the floor.

Round-trip invariant tests covering 1xN, Nx1, and 1x1 writes across numpy, dask+numpy, dask+cupy, and cupy backends.

brendancol:
## Summary

Closes #1945.

- `coords_to_transform` now recovers the per-axis pixel size from the non-degenerate axis on 1xN / Nx1 inputs and assumes square pixels for the degenerate one, instead of bailing out and letting the writer produce a non-georeferenced TIFF.
- For 1x1 inputs (no axis carries a step), `attrs['transform']` is honoured; if none is supplied, the writer fails closed with a `ValueError` rather than silently dropping spatial metadata.
- A new `require_transform_for_georeferenced` helper enforces the fail-closed rule from each writer entry point: `to_geotiff` eager, `to_geotiff` dask, `to_geotiff` with a `.vrt` path (VRT-tiled), and `write_geotiff_gpu`.

## Why this is a behaviour change

Before:

```python
da = xr.DataArray(
    np.arange(8, dtype='float32').reshape(1, 8),
    dims=('y', 'x'),
    coords={'x': np.linspace(-120.0, -113.0, 8), 'y': np.array([45.0])},
    attrs={'crs': 4326},
)
to_geotiff(da, '/tmp/strip.tif')
open_geotiff('/tmp/strip.tif').coords['x'].values[:3]  # array([0, 1, 2]) int64
```

After: the coords round-trip back as floats, and `attrs['transform']` is populated.

A 1x1 DataArray without an explicit `attrs['transform']` now raises `ValueError` instead of silently producing a non-georeferenced file.

## Test plan

- [x] New tests in `xrspatial/geotiff/tests/test_degenerate_georef_1945.py`: 1xN, Nx1, and 1x1 round-trips across numpy, dask+numpy, GPU (cupy), dask+cupy, and the VRT-tiled writer (15 tests).
- [x] Full geotiff test suite: 3109 passed locally. The 8 failures on this branch already fail on `main` (unrelated `read_to_array` monkeypatch and signature-parity tests).

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -51,6 +51,7 @@ from ._coords import (
     transform_tuple as _transform_tuple,
     transform_from_attr as _transform_from_attr,
     coords_to_transform as _coords_to_transform,
+    require_transform_for_georeferenced as _require_transform_for_georeferenced,
 )
 from ._geotags import GeoTransform, RASTER_PIXEL_IS_AREA, RASTER_PIXEL_IS_POINT
 from ._reader import UnsafeURLError
--- a/xrspatial/geotiff/_coords.py
+++ b/xrspatial/geotiff/_coords.py
@@ -265,11 +265,13 @@ def require_transform_for_georeferenced(
         raise ValueError(
             f"Cannot infer GeoTIFF transform from a "
             f"{tuple(da.sizes.values())} array with spatial coords on "
-            f"both axes: at least one axis has length 1, so the pixel "
-            f"size cannot be recovered from coord differences. Supply "
-            f"the affine transform explicitly via ``attrs['transform']`` "
-            f"(rasterio 6-tuple ``(px, 0, ox, 0, py, oy)``) or drop the "
-            f"coords if a non-georeferenced TIFF is desired."
+            f"both axes: both axes are degenerate (1x1), so neither "
+            f"coord array carries a pixel size step. 1xN and Nx1 inputs "
+            f"recover the pixel size from the non-degenerate axis, but "
+            f"a 1x1 cannot. Supply the affine transform explicitly via "
+            f"``attrs['transform']`` (rasterio 6-tuple "
+            f"``(px, 0, ox, 0, py, oy)``) or drop the coords if a "
+            f"non-georeferenced TIFF is desired."
         )
 
 
--- a/xrspatial/geotiff/_coords.py
+++ b/xrspatial/geotiff/_coords.py
@@ -232,6 +232,47 @@ def transform_from_attr(attr_val) -> 'GeoTransform | None':
     )
 
 
+def require_transform_for_georeferenced(
+    da: xr.DataArray, geo_transform
+) -> None:
+    """Raise if ``da`` carries spatial coords but no transform was derived.
+
+    Used by the writer entry points (#1945). A DataArray whose spatial
+    dim names appear in ``da.coords`` is an explicit caller request for a
+    georeferenced output. Silently falling through to a non-georeferenced
+    TIFF -- which is what the old code did for 1x1 inputs and inputs with
+    a degenerate axis -- corrupted round-trips. If the writer cannot
+    recover a transform from coords *and* the caller did not supply
+    ``attrs['transform']``, fail closed instead.
+
+    ``geo_transform`` is the value the writer has already resolved (from
+    ``attrs['transform']`` first, then from coord arrays). If it's not
+    ``None`` we have a transform and there's nothing to check.
+    """
+    if geo_transform is not None:
+        return
+    if da.ndim == 3:
+        spatial = tuple(d for d in da.dims if d not in _BAND_DIM_NAMES)
+        if len(spatial) == 2:
+            ydim, xdim = spatial[0], spatial[1]
+        else:
+            ydim = da.dims[-2]
+            xdim = da.dims[-1]
+    else:
+        ydim = da.dims[-2]
+        xdim = da.dims[-1]
+    if xdim in da.coords and ydim in da.coords:
+        raise ValueError(
+            f"Cannot infer GeoTIFF transform from a "
+            f"{tuple(da.sizes.values())} array with spatial coords on "
+            f"both axes: at least one axis has length 1, so the pixel "
+            f"size cannot be recovered from coord differences. Supply "
+            f"the affine transform explicitly via ``attrs['transform']`` "
+            f"(rasterio 6-tuple ``(px, 0, ox, 0, py, oy)``) or drop the "
+            f"coords if a non-georeferenced TIFF is desired."
+        )
+
+
 def coords_to_transform(da: xr.DataArray) -> 'GeoTransform | None':
     """Infer GeoTransform from DataArray coordinates.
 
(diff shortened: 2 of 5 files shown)


brendancol:
## PR Review: geotiff: preserve transform on 1xN / Nx1 georeferenced writes (#1945)

- None

- [ ] `xrspatial/geotiff/_coords.py:264` — `require_transform_for_georeferenced` error message says "at least one axis has length 1, so the pixel size cannot be recovered from coord differences", but the new `coords_to_transform` does recover from 1xN / Nx1 just fine; this helper only fires for the 1x1 case (both axes len < 2). Wording could read "both axes have length 1" to match the actual trigger condition and avoid misleading users who hit it.
- [ ] `xrspatial/geotiff/tests/test_degenerate_georef_1945.py:75` — only the default `raster_type='area'` is exercised. A `raster_type='point'` 1xN / Nx1 round-trip would lock in the origin offset path (no half-pixel shift) alongside the area path the tests already cover.
- [ ] `xrspatial/geotiff/_coords.py:367` — borrow comments say "x increases left-to-right by convention" and "y decreases top-to-bottom by convention" but the borrow uses `abs(...)`, which discards the sign carried by the non-degenerate axis. For a Nx1 raster stored bottom-up (positive `pixel_height`), the synthesised `pixel_width` is still positive, which happens to match the convention but is the wrong answer if the caller is mirroring an unusual source raster. A test pinning the sign behaviour on a y-ascending Nx1 input would document the choice.

- [ ] `xrspatial/geotiff/_coords.py:267` — `require_transform_for_georeferenced` calls `da.dims[-2]` for any non-3D input. A 1D DataArray would raise an IndexError before the spatial-coord check. The writers already reject 1D earlier, but a defensive `if da.ndim < 2: return` would make the helper self-contained.
- [ ] `xrspatial/geotiff/_coords.py:352` — the inline comment uses an em dash; project style elsewhere in the file uses hyphens or parentheses for asides.

### What looks good
- 1x1 fail-closed wired into all four writer entry points (eager, dask via `to_geotiff`, VRT, GPU) through a single helper, so the policy stays in sync.
- The flip from `or` to `and` on `len(x) < 2` keeps the 1x1 short-circuit while letting the 1xN / Nx1 fallback run.
- Round-trip tests cover numpy, dask+numpy, GPU, dask+cupy, and VRT paths plus the 1x1 raise.
- `attrs['transform']` precedence is preserved so the 1x1 case has a clean opt-in.

### Checklist
- [x] NaN handling is correct
- [ ] Benchmark exists or is not needed
- [ ] README feature matrix updated (if applicable)

brendancol:
## PR Review: geotiff: preserve transform on 1xN / Nx1 georeferenced writes (#1945) (re-review after rebase)

Re-reviewed at head `81d9704` after a clean rebase onto `origin/main` (which now carries the Windows fix `9ce0e60`). Local run of `test_degenerate_georef_1945.py` is 19 passed in 0.58s on the CPU paths; GPU classes are skipped here. Force-pushed with lease.

### Blockers
- None

### Suggestions
- None

### Nits
- The error message in `require_transform_for_georeferenced` calls out the 1x1 case in detail. For an unusual future shape (say a 0-length axis sneaking through) the wording would still read 1x1. Not worth changing for this PR, but a phrasing like "both spatial axes are degenerate (len < 2)" would future-proof it.
- `_strip_1xN`/`_strip_Nx1` accept a `raster_type` argument but mutate the attrs dict in place. Building the attrs dict per call (as the code does) is fine, just noting that callers reusing the literal would share state if that pattern ever changed.

### What looks good
- The split is right: `coords_to_transform` now returns a usable transform whenever at least one axis carries a step, and `require_transform_for_georeferenced` is the single fail-closed gate that catches the 1x1 case before a non-georeferenced TIFF gets written. The guard is wired into all three writer entry points (`to_geotiff`, `_write_vrt_tiled`, `write_geotiff_gpu`) with matching comments, so the behaviour cannot drift between CPU, dask, GPU, and VRT paths.
- The borrow convention (`pixel_width = abs(pixel_height)`, `pixel_height = -abs(pixel_width)`) is pinned by `TestCoordsToTransformBorrowSignPinning` against both an ascending-y Nx1 and a descending-x 1xN. That nails down the sign choice so a refactor cannot quietly flip it.
- Test matrix covers numpy, dask+numpy, cupy, dask+cupy, and VRT, plus a `raster_type='point'` parametrisation that pins the no-half-pixel-shift path on degenerate writes. The round-trip assertion checks both float coord dtype and that `attrs['transform']` came back, which is exactly the regression #1945 reported.
- Reader-side behaviour is untouched; this is a writer-only change with a clear error path for the genuinely ambiguous shape.

### Checklist
- [x] Rebase clean on `origin/main` (no conflicts)
- [x] `test_degenerate_georef_1945.py` passes locally on CPU (19/19, GPU skipped)
- [x] Fail-closed guard reaches CPU, dask, GPU, and VRT writers
- [x] Sign convention on borrowed axis covered by a dedicated test
- [x] `raster_type='point'` covered for both 1xN and Nx1
- [ ] CI green on Windows 3.12 after the rebase (depends on `9ce0e60`; verify on GH)

brendancol:
Rebased onto current main at `2b72bba` to pick up the Python 3.14 CI matrix limit (`b1579f8`).

PR #1954, opened 2026-05-15
geotiff: keep no-georef int coords through to_geotiff round-trip (#1949)

### LINKED ISSUE #1949: geotiff: to_geotiff turns no-georef int64 pixel coords into a fake transform on round-trip
A round-trip of a no-georef TIFF through `open_geotiff` -> `to_geotiff` -> `open_geotiff` silently injects a synthetic `attrs['transform']` and switches the y/x coord dtype from `int64` to `float64`.

The read side documents the no-georef contract: integer pixel coords `[0, 1, 2, ...]` with `int64` dtype, no `attrs['transform']` (#1710, #1753). The write side does not honour the same signal -- `_coords_to_transform` infers a unit `GeoTransform(origin_x=-0.5, origin_y=-0.5, pixel_width=1.0, pixel_height=1.0)` from `[0, 1, 2, ...]` coords and writes it as real `ModelPixelScale` / `ModelTiepoint` tags. On the next read the file looks like a georeferenced raster and the no-georef branch never fires.

## Reproduction

```python
import numpy as np
import tempfile, os
import tifffile

with tempfile.TemporaryDirectory() as td:
    src = os.path.join(td, "src.tif")
    arr = np.arange(20, dtype=np.float32).reshape(4, 5)
    tifffile.imwrite(src, arr, photometric='minisblack', planarconfig='contig')

    da = open_geotiff(src)
    print(da.coords['y'].dtype, da.coords['y'].values)
    # int64 [0 1 2 3]
    print('transform' in da.attrs)
    # False

    out = os.path.join(td, "out.tif")
    to_geotiff(da, out, compression='none')

    da2 = open_geotiff(out)
    print(da2.coords['y'].dtype, da2.coords['y'].values)
    # float64 [0. 1. 2. 3.]
    print(da2.attrs.get('transform'))
    # (1.0, 0.0, -0.5, 0.0, 1.0, -0.5)
```

## Impact

- Cat 2 (coords): coord dtype silently changes between input and output (HIGH per sweep criteria).
- Cat 1 (attrs): writer adds an inferred `transform` attr the input never had (MEDIUM).
- Downstream code that branches on `'transform' in da.attrs` to detect georef status will mis-classify the round-tripped file as georeferenced.

## Cause

`xrspatial/geotiff/_writers/eager.py:519`, `_writers/eager.py:839`, `_writers/gpu.py:350` all fall back to `_coords_to_transform(data)` when `attrs.get('transform')` is missing. `_coords_to_transform` (`_coords.py:235`) reads `da.coords[xdim].values`, computes `pixel_width = float(x[1] - x[0])`, and returns `GeoTransform(...)` -- with no check for the integer-dtype signal that the read side uses to mean "no georef".

Affects every writer path: `to_geotiff` eager, `to_geotiff` dask streaming (via the same fallback), `to_geotiff(vrt=...)` per-tile, `write_geotiff_gpu`. The GPU dask+cupy writer goes through the same helper.

## Proposed fix

`_coords_to_transform` returns `None` when either spatial coord array has an integer dtype (which only happens from the read-side no-georef fallback). Float coord arrays continue to produce a `GeoTransform`. Callers already handle `None` correctly: `write()` skips the geokey emission when `geo_transform is None`.

brendancol:
Closes #1949.

## Summary

- `_coords_to_transform` now returns `None` when either spatial coord array has an integer dtype, so the read-side no-georef contract (#1710, #1753) survives a round-trip through `to_geotiff`.
- Before the fix, integer pixel coords `[0, 1, 2, ...]` produced a unit `GeoTransform(origin_x=-0.5, origin_y=-0.5, pixel_width=1.0, pixel_height=1.0)` that the writer emitted as `ModelPixelScale` / `ModelTiepoint` tags. On the next read the file looked georeferenced, the coord dtype silently flipped to `float64`, and a synthetic `attrs['transform']` appeared.
- An explicit `attrs['transform']` still wins because the writers consult it first; the helper only feeds the implicit-from-coords fallback.

## Test plan

- [x] New unit tests on `_coords_to_transform` for int64 / int32 / float / 3D-yxband.
- [x] New round-trip tests through `to_geotiff` (eager numpy, dask streaming, GPU).
- [x] Double-round-trip stability test.
- [x] Explicit `attrs['transform']` override test.
- [x] Existing `test_coords_to_transform_3d_1643.py`, `test_metadata_round_trip_1484.py`, `test_attrs_parity_1548.py`, `test_coord_regularity_1720.py`, `test_coords_1813.py`, `test_georef_edges.py` still pass.
- [x] Full geotiff test suite passes (9 pre-existing failures on `main` are unrelated to this change: `test_predictor2_big_endian_gpu_1517.py`, `test_reader_kwarg_order_1935.py`, `test_size_param_validation_gpu_vrt_1776.py::TestWriteGeotiffGpuTileSize::test_tile_size_positive_works`).

### DIFF
--- a/xrspatial/geotiff/_coords.py
+++ b/xrspatial/geotiff/_coords.py
@@ -250,6 +250,11 @@ def coords_to_transform(da: xr.DataArray) -> 'GeoTransform | None':
     :func:`coords_to_transform` against the original DataArray, so the
     helper must handle both layouts to keep the geo-transform consistent
     with the file's coord arrays. See issue #1643.
+
+    Integer-dtype x/y coords are treated as a no-georef sentinel and
+    return ``None`` before the uniformity check runs; this is
+    intentional so the read-side ``np.arange`` placeholder round-trips
+    without inventing a fake unit transform (see issue #1949).
     """
     if da.ndim == 3:
         # Drop the band-like dim and keep the two spatial dims in their
--- a/xrspatial/geotiff/_coords.py
+++ b/xrspatial/geotiff/_coords.py
@@ -277,6 +277,23 @@ def coords_to_transform(da: xr.DataArray) -> 'GeoTransform | None':
     if len(x) < 2 or len(y) < 2:
         return None
 
+    # Integer coord dtype is the read-side no-georef signal: the
+    # ``has_georef=False`` branch in ``coords_from_pixel_geometry``
+    # emits ``np.arange(N, dtype=np.int64)`` for files without GeoTIFF
+    # transform tags (#1710, #1753). Synthesising a GeoTransform from
+    # those ``[0, 1, 2, ...]`` arrays would inject a fake unit transform
+    # (``origin=-0.5, pixel_width=1.0``) into the written file's
+    # ModelPixelScale / ModelTiepoint tags. The next read then takes
+    # the georef branch and the coord dtype silently flips to
+    # ``float64`` with ``attrs['transform']`` present, breaking the
+    # no-georef contract that downstream code branches on. See issue
+    # #1949. Float coord arrays still produce a GeoTransform (the
+    # canonical georef path), and an explicit ``attrs['transform']``
+    # bypasses this helper entirely so callers can still write a
+    # transform alongside int coords by setting the attr.
+    if x.dtype.kind in ('i', 'u') or y.dtype.kind in ('i', 'u'):
+        return None
+
     # GeoTIFF only supports an affine transform; non-uniform spacing
     # cannot be expressed faithfully. Validate up-front instead of
     # silently writing a transform that only matches the first step.
--- a/xrspatial/geotiff/tests/test_no_georef_writer_round_trip_1949.py
+++ b/xrspatial/geotiff/tests/test_no_georef_writer_round_trip_1949.py
@@ -1,0 +1,286 @@
+"""Round-trip tests for issue #1949.
+
+A TIFF with no GeoTIFF tags (no ``ModelPixelScale``, no
+``ModelTiepoint``, no GeoKeys) reads back with integer pixel coords
+``[0, 1, 2, ...]`` and ``int64`` dtype (the no-georef branch added in
+#1710 and #1753).
+
+Before the fix, writing that DataArray back through ``to_geotiff``
+silently injected a synthetic ``ModelPixelScale`` + ``ModelTiepoint``
+because ``_coords_to_transform`` happily inferred a unit transform from
+the integer pixel coords. The next read then took the georef branch
+and the y/x dtype flipped from ``int64`` to ``float64`` with a
+``transform`` attr present.
+
+The fix makes ``_coords_to_transform`` return ``None`` when either
+spatial coord array carries an integer dtype, so the no-georef contract
+survives an arbitrary number of read -> write -> read cycles across
+every writer path.
+"""
+from __future__ import annotations
+
+import importlib.util
+import os
+import tempfile
+
+import numpy as np
+import pytest
+import tifffile
+import xarray as xr
+
+from xrspatial.geotiff import _coords_to_transform, open_geotiff, to_geotiff
+
+
+def _gpu_available() -> bool:
+    if importlib.util.find_spec("cupy") is None:
+        return False
+    try:
+        import cupy
+        return bool(cupy.cuda.is_available())
+    except Exception:
+        return False
+
+
+_HAS_GPU = _gpu_available()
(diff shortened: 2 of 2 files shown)


brendancol:
## PR Review: geotiff: keep no-georef int coords through to_geotiff round-trip (#1949)

- None

- [ ] `xrspatial/geotiff/_coords.py:293` short-circuits before `_is_regular(x, ...)` runs. Integer arange coords are uniform so the regularity check would pass anyway, but if a caller ever passes hand-built int coords with non-uniform spacing (e.g. `[0, 1, 3, 7]`), the early return now hides that error rather than raising. A docstring note on `coords_to_transform` clarifying that integer dtype is treated as a no-georef sentinel (not just "uniform int coords") would head off confusion if someone later constructs int-indexed georeffed arrays.
- [ ] `xrspatial/geotiff/tests/test_no_georef_writer_round_trip_1949.py:96` iterates `(np.int32, np.int16, np.uint32)` inside one test function. Splitting into `pytest.mark.parametrize` would surface which dtype fails when one regresses, instead of stopping at the first.

- [ ] `xrspatial/geotiff/tests/test_no_georef_writer_round_trip_1949.py:128` builds a 3D `(y, x, band)` array with int y/x coords but does not call `to_geotiff` on it. Extending the helper test into a round-trip on the 3D layout would close the loop with the existing 3D coverage in `test_coords_to_transform_3d_1643.py`.
- [ ] `xrspatial/geotiff/tests/test_no_georef_writer_round_trip_1949.py:172` uses `random.default_rng(seed=1949)` for round-trip pixel data but never asserts pixel-value equality after round-trip. Adding `np.testing.assert_array_equal(da.data, da2.data)` would catch a regression where the writer truncates or rescales pixel bytes alongside the coord fix.
- [ ] `xrspatial/geotiff/_coords.py:280` comment block is long enough that placing it inside a small helper docstring (or moving the rationale to the function docstring) would keep the function body easier to scan.

### What looks good
- The fix lives in one place (`_coords.py:293`) and every writer path (eager, dask, GPU, VRT) flows through it, so the no-georef contract holds across backends without duplicate guards.
- `kind in ('i', 'u')` covers signed and unsigned integers at any width, matching the read-side `np.int64` arange and any user-built int8/int16/uint32 variant.
- The explicit `attrs['transform']` escape hatch is exercised by `test_explicit_transform_attr_still_writes`, confirming the helper only intercepts the implicit-from-coords fallback.
- Double round-trip stability test pins the fixed-point behaviour the issue was really after.
- GPU test is gated on cupy + CUDA availability and reuses the same assertions.

### Checklist
- [x] NaN handling is correct
- [ ] Benchmark exists or is not needed (not needed; metadata-only change)

brendancol:
## PR Review: geotiff: keep no-georef int coords through to_geotiff round-trip (#1949) (re-review after rebase)

Rebased onto `origin/main` at `9ce0e60` (Windows writer-returns fix). Conflict-free. Pushed `c996afc` which converts the new test module's unconditional `import tifffile` into `pytest.importorskip("tifffile")` so collection no longer aborts on CI shards that don't ship tifffile, matching the pattern in `test_miniswhite_nodata_1809.py` and `test_predictor3_big_endian.py`. Two stray unused stdlib imports (`os`, `tempfile`) were dropped at the same time. Local `pytest xrspatial/geotiff/tests/test_no_georef_writer_round_trip_1949.py` reports 13 passed.

### Blockers
- None

### Suggestions
- None

### Nits
- None

### What looks good
- The integer-dtype short-circuit in `coords_to_transform` is the right layer for this fix: it sits before the uniformity check and leaves the explicit `attrs['transform']` escape hatch intact, which the new `test_explicit_transform_attr_still_writes` pins.
- Coverage spans 2D, 3D, the dask streaming writer, a double round trip, and a GPU path gated on `cupy` availability, so all four backends are exercised.
- The expanded docstring on `coords_to_transform` calls out the no-georef sentinel and points to issues #1710, #1753, and #1949, which makes the rationale discoverable from the helper.
- `dtype.kind in ('i', 'u')` covers signed and unsigned widths uniformly; the parametrised `test_coords_to_transform_returns_none_for_int32_coords` locks that contract in.

### Checklist
- [x] Rebase clean on `origin/main`
- [x] Targeted test green locally (13/13)
- [x] CI collection regression on ubuntu-3.12 / macos / windows addressed via `importorskip`
- [x] Force-with-lease push to origin
- [ ] CI re-run on the new head

brendancol:
Rebased onto current main at `8da3ca7` to pick up the Python 3.14 CI matrix limit (`b1579f8`).

Issue #1860, opened 2026-05-14
geotiff: read_vrt and open_geotiff(.vrt) default missing_sources to lenient 'warn'


brendancol:
The public `read_vrt` in `xrspatial/geotiff/__init__.py:3798` defaults to `missing_sources='warn'`:

```python
def read_vrt(source: str, *,
             ...
             missing_sources: str = 'warn') -> xr.DataArray:
```

The internal `_vrt.read_vrt` it dispatches to defaults to `missing_sources='raise'` (`xrspatial/geotiff/_vrt.py:856`) and the module comment at line 256 explicitly tags `'warn'` as the lenient opt-in. The two defaults disagree and the public one is the lenient choice.

The same lenient default leaks into `open_geotiff(.vrt)` at `xrspatial/geotiff/__init__.py:714`: when the caller does not pass `missing_sources`, no kwarg is forwarded and the public `read_vrt` default takes over, so callers who reach VRT reads through `open_geotiff` get the lenient policy too.

**Repro**

```python
import xrspatial.geotiff as g
# vrt that references a missing tile
da = g.read_vrt("partial.vrt")          # warns, returns array with NaN holes
da = g.open_geotiff("partial.vrt")      # same: silently degraded output
```

**Fix**

Change the public `read_vrt` default to `missing_sources='raise'` to match the internal default, and update the docstring. Callers who want the older lenient behavior pass `missing_sources='warn'` explicitly.

Issue #1799, opened 2026-05-13
VRT missing-source handling returns partial mosaics by default


brendancol:
**Describe the bug**

VRT source read failures are skipped by default. The reader warns and records `attrs['vrt_holes']`, but still returns a partial mosaic. That behavior is dangerous for production pipelines because a downstream caller can easily miss warnings and treat zero-filled or partially missing areas as valid data.

**Expected behavior**

The VRT reader should provide a public missing-source policy, such as `missing_sources='raise'|'warn'`, and should strongly consider raising by default for unreadable sources.

**Screenshots**

Not applicable.

**Desktop**

Not applicable.

**Smartphone**

Not applicable.

**Additional context**

Strict mode already exists via `XRSPATIAL_GEOTIFF_STRICT=1`, but this behavior needs an explicit API-level contract so callers can choose safely without relying on environment variables.

Issue #1734, opened 2026-05-12
geotiff: read_vrt default lenient mode hides source read failures as zero-filled holes


brendancol:
**Describe the bug**

`xrspatial/geotiff/_vrt.py:647` wraps the per-source `read_to_array` call in a try/except that, by default, warns once and continues:

```python
except (OSError, ValueError, struct.error,) + _CODEC_DECODE_EXCEPTIONS as e:
    if _geotiff_strict_mode():
        raise
    warnings.warn(...)
    continue
```

A continuing read returns a mosaic with holes. For integer VRTs the hole is the dtype zero, which downstream functions cannot distinguish from real data unless the band has a nodata value set.

**Expected behavior**

Either invert the default to strict (lenient becomes opt-in via `XRSPATIAL_GEOTIFF_STRICT=0`), or surface a stronger signal than a warning, for example a `metadata['vrt_holes']` entry on the returned DataArray that records which sources failed and where.

**Additional context**

The current opt-in is `XRSPATIAL_GEOTIFF_STRICT=1`. Documented behavior should match what a new user will hit on a corrupt mosaic.

PR #1959, opened 2026-05-15
geotiff: chunked-VRT coverage for missing_sources policy (#1799)

brendancol:
## Summary

Closes a Cat 1 MEDIUM cross-backend equivalence gap for issue #1799 (VRT `missing_sources` policy). `test_vrt_missing_sources_policy_1799.py` covers the eager `read_vrt` path but not the chunked path.

The chunked path (`_backends/vrt.py:_read_vrt_chunked`) plumbs `missing_sources` through two separate code paths:

1. A parse-time static `os.path.exists` sweep at `vrt.py:595-621` that populates `attrs['vrt_holes']` at build time without forcing a compute (the contract documented in #1734 for partial-mosaic detection on the lazy graph).
2. Per-chunk forwarding of `missing_sources` to `_vrt_chunk_read` so each task applies the same warn/raise policy as the eager path.

A regression dropping the parse-time sweep would break partial-mosaic detection on the lazy build; a regression dropping the per-chunk forward would silently flip `missing_sources='raise'` to `'warn'` (or vice versa) on the chunked path.

8 new tests, all passing:

- `missing_sources='warn'` populates `vrt_holes` at build + emits warning on compute + present-source chunk decodes correctly
- `chunks=(h, w)` tuple form threads through identically
- `missing_sources='raise'` fails at compute on chunks touching missing sources but succeeds on windowed compute that misses them
- default `missing_sources` on the chunked path is `'raise'` (parity with eager default flipped in #1843 / #1860)
- Invalid policy raises `ValueError` at build, parity with eager path

Mutation against the `attrs['vrt_holes']` population at `vrt.py:620-621` flipped the two build-time tests red.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_vrt_chunked_missing_sources_1799.py` (8 passed)
- [x] `pytest xrspatial/geotiff/tests/test_vrt_chunked_missing_sources_1799.py xrspatial/geotiff/tests/test_vrt_missing_sources_policy_1799.py` (11 passed)
- [x] Mutation: drop `attrs['vrt_holes'] = chunked_holes` at vrt.py:620 -> 2 build-time tests fail

### DIFF
--- a/xrspatial/geotiff/tests/test_vrt_chunked_missing_sources_1799.py
+++ b/xrspatial/geotiff/tests/test_vrt_chunked_missing_sources_1799.py
@@ -97,7 +97,13 @@ class TestChunkedMissingSourcesWarn:
         )
         holes = result.attrs["vrt_holes"]
         assert len(holes) == 1
+        # Pin the full record schema (see ``_backends/vrt.py:608``) so a
+        # regression in either path that drops or renames a key is
+        # caught here.
+        assert set(holes[0].keys()) == {"source", "band", "dst_rect", "error"}
         assert holes[0]["source"].endswith("missing.tif")
+        assert holes[0]["band"] == 1
+        assert holes[0]["dst_rect"] == (4, 0, 4, 4)
 
     def test_compute_emits_per_task_warning(self, tmp_path):
         vrt_path, _ = _make_partial_vrt(str(tmp_path))
@@ -112,13 +118,14 @@ class TestChunkedMissingSourcesWarn:
             f"source after compute, got messages: {messages!r}"
         )
         # Present-source chunk decodes its 7.0 fill; missing-source
-        # chunk decodes to the source dtype's default fill (typically
-        # zero for float32). Pin the present-side decode value so a
-        # regression in the lenient path that wiped both halves would
+        # chunk decodes to NaN under the lenient policy on float32.
+        # Pin both halves so a regression in the lenient path that
+        # wiped the present side or changed the missing-side fill would
         # surface.
         np.testing.assert_array_equal(
             np.asarray(computed)[:, :4], np.full((4, 4), 7.0, dtype=np.float32),
         )
+        assert np.all(np.isnan(np.asarray(computed)[:, 4:]))
 
     def test_chunks_tuple_form(self, tmp_path):
         """Tuple ``chunks=(h, w)`` threads through identically."""
@@ -150,7 +157,7 @@ class TestChunkedMissingSourcesRaise:
         result = read_vrt(vrt_path, chunks=4, missing_sources="raise")
         # Build does not raise (the graph is lazy).
         # Computing a chunk that intersects the missing source raises.
-        with pytest.raises((OSError, FileNotFoundError, ValueError)):
+        with pytest.raises((OSError, ValueError)):
             result.compute()
 
     def test_compute_present_only_chunk_succeeds(self, tmp_path):
@@ -187,7 +194,7 @@ class TestChunkedMissingSourcesDefault:
     def test_chunked_default_raises_on_compute(self, tmp_path):
         vrt_path, _ = _make_partial_vrt(str(tmp_path))
         result = read_vrt(vrt_path, chunks=4)
-        with pytest.raises((OSError, FileNotFoundError, ValueError)):
+        with pytest.raises((OSError, ValueError)):
             result.compute()
 
 
--- a/xrspatial/geotiff/tests/test_vrt_chunked_missing_sources_1799.py
+++ b/xrspatial/geotiff/tests/test_vrt_chunked_missing_sources_1799.py
@@ -1,0 +1,208 @@
+"""Chunked-VRT coverage for ``missing_sources`` (issue #1799).
+
+``test_vrt_missing_sources_policy_1799`` covers the eager (non-chunked)
+``read_vrt`` path. The chunked path (``read_vrt(chunks=N)``, dispatching
+through ``_read_vrt_chunked``) plumbs ``missing_sources`` separately:
+
+* Parse-time approximation: a static ``os.path.exists`` sweep over every
+  source populates ``attrs['vrt_holes']`` on the returned DataArray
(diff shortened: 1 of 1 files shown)


brendancol:
## PR Review: geotiff: chunked-VRT coverage for missing_sources policy (#1799)

- None

- [ ] `test_vrt_chunked_missing_sources_1799.py:96-99` only checks that `holes[0]['source']` ends with `missing.tif`. The full record schema (`band`, `dst_rect`, `error`) is documented at `_backends/vrt.py:608` and a regression dropping any of those fields would not be caught. Asserting on the dict keys (or the `band` and `dst_rect` values) would pin the schema the eager path also returns.
- [ ] `test_vrt_chunked_missing_sources_1799.py:120-123` (test_compute_emits_per_task_warning) pins the present-side decode to 7.0 but does not assert anything about the missing-side fill. The PR description says it decodes to "typically zero for float32" but the test silently allows any value. If the lenient path stops zero-filling and returns garbage on the missing half, this test would still pass. Either pin the missing-side fill or drop the comment.
- [ ] `test_compute_intersecting_missing_raises` and `test_chunked_default_raises_on_compute` catch `(OSError, FileNotFoundError, ValueError)` which is loose. The eager-path equivalent at `test_vrt_missing_sources_policy_1799.py:41` catches `(OSError, ValueError)`. If you want parity with the eager test, drop `FileNotFoundError` (it is already a subclass of `OSError`).

- [ ] `test_vrt_chunked_missing_sources_1799.py:55-75` builds the VRT XML by string concatenation. The sibling helper in `test_vrt_chunked_shared_dataset_1923.py` does the same, so this is consistent, but an f-string template or `textwrap.dedent` would be easier to read.
- [ ] `test_vrt_chunked_missing_sources_1799.py:141` comment says "2 chunks vertically * 2 chunks horizontally = 4 tasks" but the assertion is only on `len(vrt_holes) == 1`. The 4-task claim is unverified; either assert on `result.data.numblocks` or drop the comment.
- [ ] `_make_partial_vrt` returns `(vrt_path, present_src_path)` but every call site discards the second element with `_`. Returning just `vrt_path` would simplify the signature.

### What looks good
- Test rationale is documented in module docstring and per-class docstrings, including the parse-time vs decode-time split and the #1734 / #1843 / #1860 cross-references.
- Mutation evidence in the PR body (dropping `attrs['vrt_holes'] = chunked_holes` at vrt.py:620 flips 2 tests red) gives concrete bite for the parse-time sweep.
- Uses unique tmp filenames via `tmp_path`, matching the project workflow note.
- Covers both `chunks=int` and `chunks=(h, w)` tuple forms, plus the windowed case where the missing source is excluded from the graph.
- Test-only PR with no source modifications; risk surface is bounded to the test suite.

### Checklist
- [x] Algorithm matches reference/paper (n/a, test-only)
- [x] All implemented backends produce consistent results (test pins cross-mode parity between eager and chunked)
- [x] NaN handling is correct (n/a for missing-source policy)
- [x] Edge cases are covered by tests (warn, raise, default, invalid policy, tuple chunks, windowed)
- [x] Dask chunk boundaries handled correctly (test exercises chunks that do and do not intersect missing sources)
- [x] No premature materialization or unnecessary copies (test explicitly checks `compute()` is deferred)
- [ ] Benchmark exists or is not needed (n/a, test-only)
- [ ] README feature matrix updated (if applicable) (n/a, test-only)

brendancol:
## PR Review: geotiff: chunked-VRT coverage for missing_sources policy (#1799) (re-review after rebase)

- None

- `test_vrt_chunked_missing_sources_1799.py:160` and `:197` catch `(OSError, ValueError)`. The eager path's contract on `missing_sources='raise'` is more specific than that. If the reader raises a `FileNotFoundError` (subclass of `OSError`) on missing files today, pinning that subclass would catch a regression where the raise switches to a generic `RuntimeError` or `Exception`. Consider narrowing once the exact exception type is confirmed against the eager test in `test_vrt_missing_sources_policy_1799.py`.
- `test_vrt_chunked_missing_sources_1799.py:130-141` (`test_chunks_tuple_form`): the assertion only checks `len(result.attrs["vrt_holes"]) == 1`. The chunks tuple form is meant to thread through identically, so asserting the chunk shape on `result.data` (e.g. `result.data.chunks == ((2, 2), (4, 4))`) would pin the actual tuple-form behavior the docstring claims.

- `test_vrt_chunked_missing_sources_1799.py:39` (`_make_partial_vrt`): could be a `pytest` fixture rather than a helper function. Minor; the helper is clear enough as is.
- `test_vrt_chunked_missing_sources_1799.py:57-76`: the VRT XML is built by string concatenation. A small helper or `textwrap.dedent` with a triple-quoted template would read more cleanly, but the current form is fine.

### What looks good
- Each test class targets one invariant (warn/raise/default/validation) with a focused docstring that explains why the contract matters.
- Both sides of the mosaic are asserted in `test_compute_emits_per_task_warning` (`:125-128`): present-source fill value pinned and missing-source NaN pinned.
- `test_compute_present_only_chunk_succeeds` (`:163`) covers the window scoping rule, which is a real edge case.
- `vrt_holes` record schema is pinned at `:103` (`{"source", "band", "dst_rect", "error"}`), which will catch silent key renames in the parse-time sweep.
- Parity test at `:209` confirms the eager and chunked paths reject the same invalid policy value.

### Checklist
- [x] Algorithm matches reference/paper (test-only PR; pins existing behavior)
- [x] All implemented backends produce consistent results (eager vs chunked parity pinned)
- [x] NaN handling is correct (lenient missing source -> NaN asserted at `:128`)
- [x] Edge cases are covered by tests (window scoping, tuple chunks, invalid policy)
- [x] Dask chunk boundaries handled correctly (chunks=4 splits at the missing-source boundary)
- [x] No premature materialization or unnecessary copies (`vrt_holes` checked before `.compute()`)
- [ ] Benchmark exists or is not needed (N/A - test only)

brendancol:
Rebased onto current main at `7085ee8` to pick up the Python 3.14 CI matrix limit (`b1579f8`).

Issue #1715, opened 2026-05-12
geotiff: write_vrt's crs_wkt kwarg drifts from to_geotiff/write_geotiff_gpu's crs


brendancol:
## Summary

The three public writer entry points in `xrspatial.geotiff` disagree on the CRS kwarg name and on the accepted input forms:

```python
def to_geotiff(data, path, *, crs: int | str | None = None, ...):
    """crs : int, str, or None
        EPSG code (int), WKT string, or PROJ string. ...
    """

def write_geotiff_gpu(data, path, *, crs: int | str | None = None, ...):
    """crs : int, str, or None
        EPSG code or WKT string.
    """

def write_vrt(vrt_path, source_files, *, relative=True,
              crs_wkt: str | None = None, ...):
    """crs_wkt : str or None
        CRS as a WKT string. If None, the CRS is taken from the first
        source GeoTIFF.
    """
```

The reader entry points (`open_geotiff`, `read_geotiff_dask`, `read_geotiff_gpu`, `read_vrt`) surface the file's CRS as `attrs['crs']` (an int EPSG when available) with the WKT fallback under `attrs['crs_wkt']`. So a round-trip pattern like

```python
da = open_geotiff('in.tif')
to_geotiff(da, 'out.tif', crs=4326)             # works
write_geotiff_gpu(da, 'out.tif', crs=4326)      # works
write_vrt('out.vrt', [...], crs=4326)           # TypeError: unexpected keyword 'crs'
```

forces the caller to remember which writer uses which kwarg and to convert EPSG int -> WKT manually for the VRT writer only. This is a real Cat 1 surface-drift finding: same concept (the output CRS), three writers, two distinct kwarg names.

The asymmetry also makes `write_vrt` impossible to call uniformly from generic write-wrapper code: code that forwards `crs=<value>` to whichever writer matches the output extension has to special-case VRT.

## Severity

MEDIUM (Cat 1, parameter naming drift). Not a correctness bug; the workaround is one line of pyproj. But it surprises users and makes IDE autocomplete inconsistent across the writer trio.

## Proposed fix

Add a new `crs` kwarg on `write_vrt` mirroring `to_geotiff`/`write_geotiff_gpu`. Keep `crs_wkt` for backward compatibility with a `DeprecationWarning`. Reject passing both.

```python
def write_vrt(vrt_path, source_files, *, relative=True,
              crs: int | str | None = None,
              crs_wkt: str | None = None,  # deprecated alias
[...]

PR #1962, opened 2026-05-15
geotiff: rename write_vrt's vrt_path kwarg to path (#1946)

### LINKED ISSUE #1946: geotiff: write_vrt's first positional kwarg vrt_path drifts from to_geotiff/write_geotiff_gpu path
## Summary

`xrspatial.geotiff` exposes three writers whose docstrings advertise themselves as parity siblings: `to_geotiff`, `write_geotiff_gpu`, and `write_vrt`. The first two use `path` as the destination kwarg; `write_vrt` uses `vrt_path`. A caller using `write_vrt(path=...)` hits `TypeError: write_vrt() got an unexpected keyword argument 'path'` even though the rest of the writer trio accepts `path=`.

```python
from xrspatial.geotiff import to_geotiff, write_geotiff_gpu, write_vrt
import inspect

print(list(inspect.signature(to_geotiff).parameters)[:2])         # ['data', 'path']
print(list(inspect.signature(write_geotiff_gpu).parameters)[:2])  # ['data', 'path']
print(list(inspect.signature(write_vrt).parameters)[:2])          # ['vrt_path', 'source_files']
```

## Severity

MEDIUM. Cat 1 (parameter naming drift) per the API consistency sweep. Both names exist in the public API, but most callers use the positional form (sweep grep through `xrspatial/geotiff/tests/` finds zero `vrt_path=` kwargs), so the real-world impact is small. Renaming is a breaking change, so the fix is a deprecation shim.

## Proposed fix

1. Add a `path` keyword-only alias on `write_vrt`. Accept either `path` or `vrt_path`; emit `DeprecationWarning` on `vrt_path` calls; raise `TypeError` if both are supplied.
2. Update the docstring to lead with `path`, retain `vrt_path` under a deprecated note for the historical alias.
3. Regression test that pins both names work and that the deprecation warning fires.

Detected by the deep-sweep-api-consistency sweep on 2026-05-15.

brendancol:
Closes #1946.

## Summary
- `write_vrt` now accepts `path` as its destination kwarg, matching `to_geotiff` and `write_geotiff_gpu`.
- `vrt_path` is kept as a deprecated alias that emits `DeprecationWarning`; passing both `path` and `vrt_path` raises `TypeError`.
- Mirrors the existing `crs` / `crs_wkt` shim in the same writer (#1715) and uses the same sentinel pattern (`_VRT_PATH_DEPRECATED_SENTINEL` in `_runtime.py`).

## Why
`xrspatial.geotiff` exposes three writers that the docstring advertises as parity siblings. `to_geotiff` and `write_geotiff_gpu` use `path`; `write_vrt` was using `vrt_path`, so `write_vrt(path=...)` failed with `TypeError: unexpected keyword argument` despite the writer trio claiming a shared surface. Detected by the deep-sweep-api-consistency sweep.

## Test plan
- [x] `pytest xrspatial/geotiff/tests/test_write_vrt_path_kwarg_1946.py` — 8 new tests pass locally
  - positional `write_vrt(path, sources)` works without warning
  - keyword `write_vrt(path=..., source_files=...)` works without warning
  - keyword `write_vrt(vrt_path=...)` works and emits `DeprecationWarning`
  - passing both raises `TypeError`
  - omitting both raises `TypeError`
  - `inspect.signature` pins `path` as the first positional arg
  - cross-writer signature parity (`to_geotiff`/`write_geotiff_gpu`/`write_vrt` all use `path`)
  - round-trip equivalence between the two kwarg names
- [ ] CI green

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -26,8 +26,9 @@ to_geotiff(data, path, ...)
 write_geotiff_gpu(data, path, ...)
     GPU-only writer using nvCOMP. ``to_geotiff(..., gpu=True)`` calls this
     internally.
-write_vrt(vrt_path, source_files, ...)
-    Generate a VRT mosaic XML from a list of GeoTIFF files.
+write_vrt(path, source_files, ...)
+    Generate a VRT mosaic XML from a list of GeoTIFF files. ``vrt_path``
+    is kept as a deprecated alias for ``path`` (#1946).
 """
 from __future__ import annotations
 
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -28,7 +28,8 @@ write_geotiff_gpu(data, path, ...)
     internally.
 write_vrt(path, source_files, ...)
     Generate a VRT mosaic XML from a list of GeoTIFF files. ``vrt_path``
-    is kept as a deprecated alias for ``path`` (#1946).
+    is kept as a deprecated alias for ``path``; passing both ``path`` and
+    ``vrt_path`` raises ``TypeError`` (#1946).
 """
 from __future__ import annotations
 
--- a/xrspatial/geotiff/_runtime.py
+++ b/xrspatial/geotiff/_runtime.py
@@ -47,6 +47,13 @@ _MISSING_SOURCES_SENTINEL = object()
 # rationale ``_CRS_WKT_DEPRECATED_SENTINEL`` documents above. See
 # issue #1946.
 _VRT_PATH_DEPRECATED_SENTINEL = object()
+# ``write_vrt`` also needs to distinguish "user passed path= explicitly"
+# (including an explicit ``path=None``, which is an error) from "user
+# passed nothing" (fall through to the ``vrt_path`` shim). Without this
+# sentinel, ``write_vrt(None, sources)`` silently fell through to the
+# ``path is None`` branch and raised a "missing required argument"
+# TypeError for the wrong reason. See PR #1962 review.
+_VRT_PATH_MISSING_SENTINEL = object()
 
 
 # Spatial dim names recognised on 3D writer inputs. ``y``/``x`` are the
--- a/xrspatial/geotiff/_runtime.py
+++ b/xrspatial/geotiff/_runtime.py
@@ -39,6 +39,14 @@ _CRS_WKT_DEPRECATED_SENTINEL = object()
 # (forward verbatim to read_vrt). Mirrors the on_gpu_failure pattern. See
 # issue #1810.
 _MISSING_SOURCES_SENTINEL = object()
+# ``write_vrt`` historically named its first positional kwarg ``vrt_path``
+# while ``to_geotiff`` / ``write_geotiff_gpu`` use ``path``. The deprecation
+# shim adds ``path`` as the new name and accepts ``vrt_path`` with a
+# DeprecationWarning. The sentinel pattern distinguishes "user passed
+# vrt_path= explicitly" from "user passed nothing", which is the same
+# rationale ``_CRS_WKT_DEPRECATED_SENTINEL`` documents above. See
+# issue #1946.
+_VRT_PATH_DEPRECATED_SENTINEL = object()
 
 
 # Spatial dim names recognised on 3D writer inputs. ``y``/``x`` are the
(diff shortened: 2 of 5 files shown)


brendancol:
## PR Review: geotiff: rename write_vrt's vrt_path kwarg to path (#1946)

- None

- [ ] `xrspatial/geotiff/_writers/vrt.py:19-25` — the `path` kwarg signature is `path: str | None = None` with a real `None` default, while every other kwarg in the writer trio that uses the sentinel pattern uses a private sentinel object so the shim can tell "user passed `None`" from "user passed nothing". With the current shape, `write_vrt(None, sources)` (positional `None`) silently falls through to the `path is None` branch and raises `TypeError: missing required argument 'path'`, which is the right behaviour but for the wrong reason. Consider using a `_PATH_SENTINEL` and treating an explicit `path=None` as the ambiguous case the way `crs_wkt` does, or document the asymmetry. Not load-bearing but worth a comment for the next reader.
- [ ] `xrspatial/geotiff/__init__.py:29-31` — the Public API docstring entry for `write_vrt` mentions `vrt_path` is "kept as a deprecated alias for `path`" but does not call out that the shim raises `TypeError` when both are passed. The `crs_wkt` entry (further up the same docstring family in `_writers/vrt.py:51-58`) does spell that rule out; mirror the wording here for consistency.

- [ ] `xrspatial/geotiff/_writers/vrt.py:80-85` — the long comment block explaining the shim is good but the phrase "preserve back-compat" reads slightly informal next to the rest of the file. Minor.
- [ ] `xrspatial/geotiff/tests/test_write_vrt_path_kwarg_1946.py:151-170` — `test_write_vrt_first_arg_name_matches_writer_trio` duplicates the assertion in `test_write_vrt_signature_first_arg_is_path` for `write_vrt`. The trio-parity check could just call the per-function tests, but keeping it standalone is fine if the intent is to catch any of the three writers drifting in isolation.

### What looks good
- The deprecation shim mirrors the existing `crs_wkt` (#1715) shape in the same file, which keeps the writer self-consistent.
- Both-supplied rejection is tested with `pytest.raises(TypeError, match="path.*vrt_path")`, and the deprecation warning is tested with `pytest.warns(DeprecationWarning, match='vrt_path')`. Round-trip equivalence between old and new kwarg is also pinned.
- Signature parity is asserted against `to_geotiff` and `write_geotiff_gpu` so a future re-rename trips a test.
- The sentinel lives in `_runtime.py` alongside the other writer-trio sentinels, preserving the module-cache identity rule the file documents.

### CI status
- macOS 3.13 (and the rest of the matrix) failed on `xrspatial/geotiff/tests/test_signature_annotations_1654.py::test_write_vrt_vrt_path_annotated`: PR-induced. The pre-existing test pinned `vrt_path`'s annotation to `'str'`, but the deprecation default of a sentinel forces the annotation to widen to `'str | None'`. Fixed in 6e46c58 by updating that test and adding a sibling `test_write_vrt_path_annotated` for the new canonical name.
- `request-copilot-review`: passed (no review content posted).
- ReadTheDocs: passed.

### Backward compatibility
- The old `vrt_path=` kwarg is still accepted and emits `DeprecationWarning`. Positional `write_vrt(some_path, sources)` keeps working unchanged because `path` sits where `vrt_path` used to. Passing both raises `TypeError`. This is the right call for an API rename in this project — the same shape `crs_wkt`/`crs` shipped under (#1715).
- A repo-wide grep for `write_vrt(vrt_path=` finds no call sites in the codebase that still use the deprecated alias. Other matches (`xrspatial/geotiff/_writers/eager.py`, several test files) are local variable names that happen to be called `vrt_path` and are passed positionally — those keep working.

### Checklist
- [x] Algorithm matches reference/paper (N/A — kwarg rename only)
- [x] All implemented backends produce consistent results (no backend code touched)
- [x] NaN handling is correct (no data-path changes)
- [x] Edge cases are covered by tests (both-supplied, neither-supplied, positional, keyword, deprecation, round-trip)
- [x] Benchmark exists or is not needed (not needed for a signature change)
- [ ] README feature matrix updated (if applicable) — N/A

brendancol:
Addressed in 9f83e12.

- `_writers/vrt.py`: `path` now defaults to a new `_VRT_PATH_MISSING_SENTINEL` (alongside the existing `_VRT_PATH_DEPRECATED_SENTINEL` for `vrt_path`). Explicit `path=None` or positional `write_vrt(None, sources)` raises `TypeError` with a message that names `'path'`. Caller-omitted falls through to the `vrt_path` shim as before.
- `__init__.py` Public API docstring: the `write_vrt` entry now spells out that passing both `path` and `vrt_path` raises `TypeError`, mirroring the `crs_wkt` rule.
- Tests: signature annotation for `path` is now plain `str` (test updated). Added two cases for explicit-`None` (keyword and positional).

brendancol:
Rebased onto main at `7b83a55` to pick up the int-coord hotfix from #1969.

PR #1963, opened 2026-05-15
geotiff: cross-library accuracy test (rasterio vs xrspatial vs zarr) (#1961)

### LINKED ISSUE #1961: geotiff: cross-library accuracy test (rasterio vs xarray-spatial vs zarr round-trip)
## Test idea

Read the same GeoTIFF three ways and assert the results agree:

1. `rasterio.open(path)` → numpy array, transform, CRS, nodata.
2. `xrspatial.geotiff.open_geotiff(path)` → `xr.DataArray`.
3. Write the xarray-spatial DataArray with `.to_zarr(...)`, reopen with `xr.open_zarr(...)`.

Pairwise checks across the three readers:

- pixel values (`array_equal` for integer dtypes, `allclose` with a documented tolerance for floats)
- shape, dtype, nodata sentinel
- `x` and `y` coordinate arrays within float tolerance
- CRS WKT after normalising through `_resolve_crs_to_wkt`
- affine transform (origin, pixel size) within `1e-9` relative

## Why

The project already has backend-parity tests inside its own implementation (numpy / cupy / dask / dask+cupy). What's missing is a test that pins the GeoTIFF reader against an external reference (rasterio) and against a round-trip through a different on-disk format (Zarr). A regression in header parsing, georef extraction, coord generation, nodata handling, or Zarr metadata propagation can pass every existing test and only surface when a user files a bug.

## Suggested inputs

Cover the cases that have drifted before:

- single-band float32 with a non-NaN nodata sentinel
- multi-band uint16 with per-band nodata
- north-up and south-up rasters (negative vs positive `pixel_height`)
- 1×N / N×1 stripe (#1945)
- tiled COG, no overviews
- no-georef raster with integer coords (#1949)

## Suggested location

`xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_<issue#>.py`

Gate the imports with `pytest.importorskip("rasterio")` and `pytest.importorskip("zarr")` so the CPU-only matrix without those deps stays green.

## Out of scope

- performance comparison (handled by the bench skill)
- GPU read paths (covered by the existing backend parity tests)
- cloud / HTTP sources (local-file only)

brendancol:
Closes #1961.

Adds xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py.
Reads the same TIFF with rasterio, open_geotiff, and a to_zarr/open_zarr
round trip, and asserts pixel values, shape, dtype, nodata sentinel, x/y
coords, CRS (normalised through _resolve_crs_to_wkt), and the affine
transform agree.

Cases:
- float32 with non-NaN nodata sentinel
- multi-band uint16 with per-band nodata
- north-up and south-up rasters
- 1xN / Nx1 stripes (#1945)
- tiled no-overview
- no-georef integer coords (#1949)

rasterio and zarr go through pytest.importorskip so the CPU-only matrix
without those deps stays green.

## Test plan
- [x] pytest xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py - 8 passed

### DIFF
--- a/xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py
+++ b/xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py
@@ -16,7 +16,7 @@ existing test and only surface when a user files a bug.
 Each input file covers a case that has drifted before:
 
 - single-band float32 with a non-NaN nodata sentinel
-- multi-band uint16 with per-band nodata
+- multi-band uint16 with a single dataset-level nodata sentinel
 - north-up and south-up rasters (negative vs positive ``pixel_height``)
 - 1xN / Nx1 stripe (#1945)
 - tiled COG, no overviews
@@ -30,8 +30,8 @@ import numpy as np
 import pytest
 import xarray as xr
 
-rasterio = pytest.importorskip('rasterio')
-zarr = pytest.importorskip('zarr')
+rasterio = pytest.importorskip('rasterio', exc_type=ImportError)
+zarr = pytest.importorskip('zarr', exc_type=ImportError)
 
 from rasterio.transform import Affine, from_origin  # noqa: E402
 
@@ -289,14 +289,28 @@ class TestSingleBandFloat32NodataSentinel:
         )
 
 
-class TestMultibandUint16PerBandNodata:
+class TestMultibandUint16SharedNodata:
+    """Multi-band uint16 with one dataset-level nodata sentinel.
+
+    The TIFF / GDAL convention stores a single ``GDAL_NODATA`` tag per
+    file, so a multi-band raster shares one sentinel across every band.
+    rasterio exposes a ``nodatavals`` tuple for compatibility, but writing
+    distinct per-band values through the GTiff driver collapses to the
+    last value written (rasterio's ``_set_nodatavals`` private API
+    overwrites the single GDAL_NODATA tag). xrspatial mirrors that model
+    on the read side and surfaces a scalar on ``attrs['nodata']``. A test
+    that pretends to write distinct values per band would silently round
+    trip the wrong sentinel, so this case fixes one value and exercises
+    masking across both bands.
+    """
+
     def test_round_trip(self, tmp_path):
         path = tmp_path / 'tmp_1961_uint16_multiband.tif'
         store = tmp_path / 'tmp_1961_uint16_multiband.zarr'
         band1 = np.arange(12, dtype=np.uint16).reshape(3, 4)
         band1[0, 0] = 0  # nodata cell on band 1
         band2 = np.full((3, 4), 7, dtype=np.uint16)
-        band2[2, 3] = 0
+        band2[2, 3] = 0  # nodata cell on band 2 (same sentinel value)
         data = np.stack([band1, band2])  # (2, 3, 4) for rasterio
 
         with rasterio.open(
--- a/xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py
+++ b/xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py
@@ -1,0 +1,445 @@
+"""Cross-library accuracy test (issue #1961).
+
+Read the same GeoTIFF three ways and assert the results agree:
+
+1. ``rasterio.open(path)`` -> numpy array, transform, CRS, nodata.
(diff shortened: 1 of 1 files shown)


brendancol:
## PR Review: geotiff: cross-library accuracy test (rasterio vs xrspatial vs zarr) (#1961)

- None

- [ ] `xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py:292` — `TestMultibandUint16PerBandNodata` advertises per-band nodata but writes a single shared `nodata=0` via rasterio's dataset-level parameter. Either rename the class to `TestMultibandUint16SharedNodata`, or set per-band nodata via `dst.nodatavals` / `dst.update_tags` and assert the reader picks up the distinct values.
- [ ] `xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py:33-34` — `pytest.importorskip('rasterio')` and `pytest.importorskip('zarr')` work today but pytest 9.1 deprecates implicit `ImportError` matching; passing `exc_type=ImportError` makes the skip stable across pytest upgrades. Worth aligning all geotiff importorskip calls in a separate sweep.

- [ ] `xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py:214-217` — the nodata equality check relies on `np.isnan(xrs.attrs.get('nodata', 0.0))` to dodge `np.isnan(None)`. Pulling the value into a local first and gating on `is not None` reads more clearly than relying on the default.
- [ ] `xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py:340-368` — the south-up case uses positive `pixel_height=+1.0`, which rasterio reports but most GeoTIFF writers never produce. Worth a one-line comment noting this is a synthetic case, since real south-up data tends to come from non-standard sources.
- [ ] `xrspatial/geotiff/tests/test_round_trip_parity_rasterio_zarr_1961.py:46` — `ATOL_FLOAT32 = 0.0` is fine for the round-trip path but the constant name implies a dtype-specific tolerance. Rename to `ATOL_PIXEL = 0.0` or inline it where used.

### What looks good
- The three-way structure (rasterio raw read, xrspatial reader, xrspatial -> zarr -> xarray reopen) is a real cross-validation, not self-comparison.
- Coord comparisons rebuild pixel-centre coords from the rasterio transform rather than reusing xrspatial's coords as the reference.
- CRS comparison goes through `pyproj.CRS` equality, which dodges WKT string drift across PROJ versions.
- Integer vs float nodata semantics are handled separately and the dtype promotion expectations match the reader's documented behaviour.
- Temp filenames use the `tmp_1961_*` prefix, matching the project convention for collision-resistant temp paths.
- Each test class targets a case that has regressed before (stripes #1945, no-georef #1949, south-up, tiled COG).

### Cross-library parity notes
The test reads each fixture with `rasterio.open(...).read()` to get a reference array, transform, CRS, and nodata that come from a different code path than xrspatial's reader. Pixel arrays go through `_assert_pixels_equal` with NaN-position equality, transform tuples go through `_assert_transforms_match` against the rasterio `Affine`, and CRS goes through pyproj equality. The Zarr leg writes the xrspatial DataArray, reopens it, and re-checks pixels, coords, and the `crs` / `crs_wkt` / `nodata` / `transform` attrs survive the trip. Coverage gaps worth noting for follow-up: no compressed inputs (LZW, Deflate, predictor=2/3), no overviews, no signed integer dtypes, no datetime or extra-dim coords on the zarr round-trip, and the multi-band case does not actually exercise per-band nodata despite the class name. The parity guarantees are meaningful for the cases listed.

### Checklist
- [ ] All implemented backends produce consistent results (numpy only; no cupy / dask leg)
- [x] NaN handling is correct
- [ ] Dask chunk boundaries handled correctly (not exercised)
- [ ] Benchmark exists or is not needed (not needed for a parity test)

brendancol:
Addressed the review in 0da084c:

- Renamed `TestMultibandUint16PerBandNodata` to `TestMultibandUint16SharedNodata`. The GTiff format stores one `GDAL_NODATA` tag per file, and rasterio's `_set_nodatavals` overwrites that single tag on disk, so a real per-band sentinel cannot round trip through this format. Docstring explains the limitation.
- Passed `exc_type=ImportError` to both `pytest.importorskip` calls.

All 8 tests pass locally.

brendancol:
Rebased onto main at `9305773` to pick up the int-coord hotfix from #1969.

Issue #1974, opened 2026-05-15
geotiff: open_geotiff silently drops max_cloud_bytes on gpu / chunks / VRT paths


brendancol:
The test-coverage sweep on 2026-05-15 (pass 16) flagged a silent-kwarg-drop in `open_geotiff`'s dispatcher matching the pattern fixed in #1561 / #1605 / #1685 / #1810.

## Repro

```python
import numpy as np, xarray as xr, tempfile, os
from xrspatial.geotiff import open_geotiff, to_geotiff, write_vrt

with tempfile.TemporaryDirectory() as d:
    p = os.path.join(d, 't.tif')
    da = xr.DataArray(
        np.zeros((4, 4), dtype=np.float32),
        dims=['y', 'x'],
        coords={'y': np.arange(4., 0, -1), 'x': np.arange(4.)},
        attrs={'crs': 4326, 'transform': (1.0, 0, 0.0, 0, -1.0, 4.0)},
    )
    to_geotiff(da, p)

    # All three silently accept the kwarg today; the budget is never applied.
    open_geotiff(p, max_cloud_bytes=8, gpu=True)        # silent drop
    open_geotiff(p, max_cloud_bytes=8, chunks=2)        # silent drop
    vrt = os.path.join(d, 't.vrt'); write_vrt(vrt, [p])
    open_geotiff(vrt, max_cloud_bytes=8)                # silent drop
```

## Root cause

`open_geotiff` accepts `max_cloud_bytes` in its signature but only forwards it to `_read_to_array` on the eager non-VRT branch (`xrspatial/geotiff/__init__.py:431`). The GPU branch at line 410, the dask branch at line 422, and the VRT branch at line 362 never reference the kwarg. Callers have no way to learn the budget is being ignored.

The two sibling backend-only kwargs `on_gpu_failure` (line 339) and `missing_sources` (line 355) already raise `ValueError` when used on a path where they do not apply. `max_cloud_bytes` defaults to the same `_MAX_CLOUD_BYTES_SENTINEL` style sentinel and slots into the same pattern, but the rejection guard is missing.

## Fix options

1. Add `ValueError` guards mirroring the existing `on_gpu_failure` / `missing_sources` pattern (refuses the kwarg on gpu / chunks / VRT paths). Narrower change.
2. Thread `max_cloud_bytes` through every backend so it has effect everywhere. Broader change since the GPU and dask paths would need to plumb the budget into their respective fsspec entry points.

[...]

Issue #1957, opened 2026-05-15
geotiff: CI failing on main — max_cloud_bytes missing from _CANONICAL_ORDER


brendancol:
## Summary

CI on `main` fails on every Python version because `test_open_geotiff_defines_canonical_order` rejects the current `open_geotiff` signature. The new `max_cloud_bytes` kwarg (added by #1932 / commit `8e61c14`) lives between `max_pixels` and `on_gpu_failure` in the function signature, but the test's `_CANONICAL_ORDER` constant in `xrspatial/geotiff/tests/test_reader_kwarg_order_1935.py:33` was not updated to match.

## Failing assertion (from CI run 25926000277, ubuntu-3.14)

```
AssertionError: open_geotiff kw-only params
  ['dtype', 'window', 'overview_level', 'band', 'name', 'chunks', 'gpu',
   'max_pixels', 'max_cloud_bytes', 'on_gpu_failure', 'missing_sources']
no longer match the canonical order
  ['dtype', 'window', 'overview_level', 'band', 'name', 'chunks', 'gpu',
   'max_pixels', 'on_gpu_failure', 'missing_sources'].
Update both the function and the _CANONICAL_ORDER constant together.
```

The test is doing its job: it fires when the signature drifts from the canonical list. The fix is to add `max_cloud_bytes` to the constant.

This is also blocking every other open PR (#1947, #1953, etc.) because their CI runs hit the same pre-existing failure on rebased `main`.

## Fix

`xrspatial/geotiff/tests/test_reader_kwarg_order_1935.py:33-44`: insert `"max_cloud_bytes"` between `"max_pixels"` and `"on_gpu_failure"` in `_CANONICAL_ORDER`.

Issue #1928, opened 2026-05-15
geotiff: eager fsspec cloud read pulls full object before dim guard


brendancol:
**Describe the bug**
Eager reads from cloud storage via fsspec pull the entire object into memory before any TIFF header parse or `max_pixels` guard runs.

`read_to_array()` at [`xrspatial/geotiff/_reader.py:2925`](../blob/main/xrspatial/geotiff/_reader.py#L2925) constructs a `_CloudSource` for any non-HTTP `://` source and immediately calls `src.read_all()`. `_CloudSource.read_all()` at [`xrspatial/geotiff/_reader.py:1339`](../blob/main/xrspatial/geotiff/_reader.py#L1339) does an unbounded `f.read()` with no size check, so a large or hostile remote TIFF on `s3://`, `gs://`, `az://`, or `abfs://` can exhaust memory and bandwidth before the dimensions are checked.

The HTTP path already reads only what it needs via `_parse_cog_http_meta` and range fetches against `_HTTPSource`. The dask backend also bounds its metadata reads. Eager cloud is the gap.

**Expected behavior**
Eager cloud reads should either:
- Refuse objects larger than a caller-configurable byte budget, raising before any data is downloaded. `_CloudSource.__init__` already knows the object size from `fsspec.size()`, so the check is cheap.
- Or (deeper fix) reuse the range-reader path, since `_CloudSource` already exposes `read_range` / `read_ranges` mirroring `_HTTPSource`.

The byte-budget fix is the smaller change and closes the safety hole. The range-based refactor can land separately as backend parity work.

**Proposed fix**
1. Add a `MAX_CLOUD_BYTES_DEFAULT` constant (256 MiB) in `_reader.py`, with `XRSPATIAL_GEOTIFF_MAX_CLOUD_BYTES` env override.
2. Plumb a `max_cloud_bytes` kwarg through `read_to_array` and `open_geotiff`. `None` opts out of the size check entirely.
3. Before `src.read_all()` in the fsspec branch, compare `src.size` against the budget. Refuse oversized objects with a clear error.

**Additional context**
Companion to PRs #1873 (HTTP / loopback test gating) and the bounded-metadata path on `_HTTPSource`. Surfaced by an external review of the geotiff module.

Issue #1810, opened 2026-05-13
geotiff: open_geotiff silently drops missing_sources when routing to read_vrt


brendancol:
## Summary

`open_geotiff` does not accept `missing_sources` and does not forward it to `read_vrt` when dispatching to a `.vrt` source. Callers who want strict failure on broken VRT sources must call `read_vrt` directly, defeating the documented dispatcher.

## Repro

```python
from xrspatial.geotiff import open_geotiff
open_geotiff("mosaic.vrt", missing_sources="raise")
# TypeError: open_geotiff() got an unexpected keyword argument 'missing_sources'
```

## Why this matters

Same class of bug as #1561 / #1605 / #1685 / #1795: a backend kwarg is reachable on the routed-to function but not through the documented dispatcher. #1806 added `missing_sources` to `read_vrt` but did not propagate it to `open_geotiff`. The VRT dispatch branch in `xrspatial/geotiff/__init__.py:813-815` builds the forwarded call without it:

```python
return read_vrt(source, dtype=dtype, window=window, band=band,
                name=name, chunks=chunks, gpu=gpu,
                max_pixels=max_pixels)
```

## Proposed fix

Add `missing_sources='warn'` to `open_geotiff`, forward it to `read_vrt`, and reject it when the source is not a VRT (TIF or file-like) with a clear `ValueError`, matching the pattern `open_geotiff` already uses for `on_gpu_failure` and `overview_level`. The default `'warn'` preserves historic behaviour.

Found during the geotiff API consistency sweep (Cat 5, MEDIUM).

PR #1977, opened 2026-05-15
geotiff: pin open_geotiff(max_cloud_bytes=...) dispatcher silent drop (#1974)

brendancol:
## Summary

Test-coverage sweep pass 16 on the geotiff module. Closes a Cat 4 HIGH parameter-coverage gap on the `open_geotiff` dispatcher's `max_cloud_bytes` kwarg.

`max_cloud_bytes` was added in #1928 as an eager fsspec cloud-budget guard and re-ordered into the canonical reader signature by #1957. `open_geotiff` accepts the kwarg in its signature but only forwards it to `_read_to_array` on the eager non-VRT branch (`xrspatial/geotiff/__init__.py:431`). The GPU branch at line 410, the dask branch at line 422, and the VRT branch at line 362 never reference the kwarg, so:

```python
open_geotiff(p, max_cloud_bytes=8, gpu=True)   # silently drops
open_geotiff(p, max_cloud_bytes=8, chunks=4)   # silently drops
open_geotiff(vrt, max_cloud_bytes=8)           # silently drops
```

Same class of dispatcher-silently-drops-backend-kwarg bug fixed by #1561, #1605, #1685, and #1810 for other backend-only kwargs. The two sibling kwargs `on_gpu_failure` (line 339) and `missing_sources` (line 355) already raise `ValueError` when used on a path where they do not apply.

Filed issue #1974 for the dispatcher fix. This sweep is test-only.

## Tests

`xrspatial/geotiff/tests/test_max_cloud_bytes_dispatcher_silent_drop_2026_05_15.py` (11 tests, all behaving as designed locally):

- 4 `xfail(strict=True)` pinning the fix surface (gpu, dask, vrt, dask+gpu). Flip to xpass (= test failure under `strict=True`) when the dispatcher fix lands.
- 3 passing pins on the current silent-drop behaviour so the fix is visible as a diff.
- 4 positive pins that the eager local + file-like paths accept the kwarg (the docstring no-op contract).

## Test plan

- [x] All 11 tests pass / xfail as designed on the local host (`pytest xrspatial/geotiff/tests/test_max_cloud_bytes_dispatcher_silent_drop_2026_05_15.py -v` -> 7 passed, 4 xfailed).
- [ ] CI to confirm parity (the dask+gpu xfail skips when cupy / CUDA is unavailable).


brendancol:
## PR Review

### What looks good

- The xfail-strict + counter-pin pattern is the right shape for a test-only sweep PR. The four `xfail(strict=True)` cases pin the fix target (they flip from xpass → fail when the dispatcher gains the ValueError), and the `TestCurrentSilentDropPins` siblings make the current buggy behaviour visible as a test diff at fix time. Reads cleanly and matches the precedent set by #1561 / #1605 / #1685 / #1810.
- The eager positive pins (`TestEagerLocalPathAcceptsMaxCloudBytes`, `TestEagerFileLikeAcceptsMaxCloudBytes`) lock in the documented no-op contract for local and file-like sources separately from the dispatch-rejection contract. Useful regression net regardless of how #1974 is resolved.
- Module docstring explicitly enumerates the two acceptable fix surfaces (ValueError guard vs. thread the budget through every backend), which keeps the test's "after" expectations precise without over-prescribing.
- Tests run as designed locally: 7 passed, 4 xfailed.

### Suggestions

- [ ] The cupy availability check is duplicated verbatim in `test_dispatcher_dask_gpu_path_rejects_max_cloud_bytes` (lines 209–217) and `TestCurrentSilentDropPins.test_gpu_path_silently_accepts_today` (lines 241–249). Worth a shared module-level fixture (or `pytest.importorskip` plus a `cupy.cuda.is_available()` skip) — twelve lines of boilerplate becomes one.
- [ ] `pytest.raises(ValueError, match=r"max_cloud_bytes")` requires the fix author's error message to mention `max_cloud_bytes` literally. The sibling `on_gpu_failure` / `missing_sources` errors do follow that convention, so this is consistent — but worth flagging in #1974 so the fix author doesn't ship a message like "this kwarg only applies on the eager path" and trip the xpass.
- [ ] `_build_local_tif` writes via the eager path, which after rebase routes through the writer's CRS validator (#1978). `attrs={'crs': 4326}` is fine, but if a future commit makes that attrs entry stricter (e.g. requires `crs_wkt` instead), this fixture quietly stops producing test inputs. Low risk, but the helper depends on a chunk of the write API that has been churning.

### Nits

- [ ] `TestCurrentSilentDropPins` will need wholesale removal when #1974 lands; the docstring already says so. Worth a one-line FIXME or `# remove with #1974` comment at class level so a future cleanup author can grep for it.
- [ ] `test_local_file_max_cloud_bytes_small_is_noop` passes `max_cloud_bytes=8`. The intent (8 bytes — far below the file size — proves local files skip the budget) reads from context but is not stated. A one-line comment like `# 8 bytes is far below the file size; local files skip the budget` would help a reader who lands here without #1928 context.

### Checklist

- [x] Algorithm matches reference/paper — not applicable (test-coverage sweep)
- [x] All implemented backends covered — the four backend dispatch arms each have a xfail pin
- [x] Edge cases covered — local vs. file-like vs. dask vs. gpu vs. vrt vs. dask+gpu
- [x] No premature materialisation in test fixtures
- [x] No benchmark required (test-only PR)
- [x] Docstrings present and accurate — module and class docstrings carry the fix-visibility contract

PR #1978, opened 2026-05-15
geotiff: reject bool and unresolvable EPSG in writer crs= (#1971)

### LINKED ISSUE #1971: geotiff: to_geotiff(crs=True) writes EPSG=1 silently; bool slips through int check
**Describe the bug**

`xrspatial.geotiff.to_geotiff(da, path, crs=True)` writes a file with the projected/geographic CRS key set to `1` and produces a DataArray on read-back with `attrs['crs'] == 1`. Two gaps in the validation chain cause this:

- `xrspatial/geotiff/_writers/eager.py:501` resolves the `crs` argument with `isinstance(crs, int)`. `bool` is a subclass of `int` in Python, so `crs=True` is accepted and `epsg=1`. `crs=False` would similarly map to `epsg=0`.
- `xrspatial/geotiff/_geotags.py:1077` writes the EPSG code without checking that it resolves with pyproj.
- On read, `xrspatial/geotiff/_geotags.py:557` accepts any numeric projected/geographic CRS key as `attrs['crs']`, even when `_epsg_to_wkt()` later emits a `GeoTIFFFallbackWarning` because pyproj cannot resolve it.

**Reproduction**

```python

da = xr.DataArray(np.zeros((4,4), dtype=np.float32),
                  coords={'y': np.arange(4), 'x': np.arange(4)},
                  dims=('y','x'))
    to_geotiff(da, f.name, crs=True)
    print(open_geotiff(f.name).attrs.get('crs'))  # prints 1
```

**Expected behavior**

`to_geotiff` should reject `bool` for `crs` and any integer that does not resolve with pyproj, raising a clear `ValueError` at the entry point. Bad CRS metadata cannot reach the file and cannot silently poison reprojection or merge code paths.

**Additional context**

Caught in a code review of the geotiff module. Sibling validators (e.g. `_validate_chunks_arg`, `_validate_tile_size`) already reject bools explicitly because `True`/`False` sneak through int checks; the CRS path needs the same treatment.

brendancol:
Closes #1971.

## Summary

- `crs=True` / `crs=False` are now rejected at the writer entry point with a clear `ValueError`. `bool` is an `int` subclass, so the previous `isinstance(crs, int)` check accepted them and wrote `EPSG=1` / `EPSG=0` to the file.
- Integer `crs=` values are now validated via `pyproj.CRS.from_epsg` before they reach the GeoTIFF tags. An int that does not resolve as a CRS raises `ValueError` instead of writing garbage and emitting a `GeoTIFFFallbackWarning` after the fact.
- The validator lives in `_crs.py` as `_validate_crs_arg` and is called from `to_geotiff` (eager), `_write_vrt_tiled`, `write_geotiff_gpu`, and `_resolve_crs_to_wkt` (used by `write_vrt`), so every writer entry point shares one rejection path.
- Pyproj is optional; the EPSG-resolves check is skipped when pyproj is not installed, matching the rest of the module.

## Test plan

- [x] New `xrspatial/geotiff/tests/test_crs_arg_validation_1971.py` covers `_validate_crs_arg` directly and the three writer entry points (eager, GPU import-only-checked, `.vrt` path).
- [x] `pytest xrspatial/geotiff/tests/ -k "crs or epsg"` — 151 passed, no regressions.
- [x] Full `pytest xrspatial/geotiff/tests/` — same 8 pre-existing failures as main (GPU predictor tests + a tile_size unit test), none related to this change.

### DIFF
--- a/xrspatial/geotiff/_crs.py
+++ b/xrspatial/geotiff/_crs.py
@@ -44,6 +44,62 @@ def _looks_like_wkt(s: str) -> bool:
     return s.lstrip().upper().startswith(_WKT_ROOT_KEYWORDS)
 
 
+def _validate_crs_arg(crs) -> None:
+    """Reject malformed ``crs=`` arguments before they reach the writer.
+
+    Closes two gaps in the writer entry points (issue #1971):
+
+    * ``bool`` is an ``int`` subclass, so ``crs=True`` and ``crs=False``
+      would otherwise slip through ``isinstance(crs, int)`` and write
+      ``EPSG=1`` / ``EPSG=0`` to the file. No CRS database resolves
+      those, so the result is silent metadata corruption.
+    * An ``int`` EPSG code that pyproj cannot resolve gets written
+      verbatim into ``ProjectedCSType`` / ``GeographicType``. The
+      file then round-trips with ``attrs['crs']`` set to the bad
+      value and only a ``GeoTIFFFallbackWarning`` to tell the caller
+      something is wrong.
+
+    Validates ``crs`` is one of ``None`` (no-op), ``int`` (a valid
+    EPSG code), or ``str`` (WKT/PROJ -- left for ``_wkt_to_epsg``
+    downstream). Pyproj is optional; the EPSG-resolves check is
+    skipped when pyproj is not installed, matching the rest of the
+    module's pyproj-optional posture. Under
+    ``XRSPATIAL_GEOTIFF_STRICT=1`` the pyproj error is re-raised
+    instead of being wrapped.
+    """
+    if crs is None:
+        return
+    if isinstance(crs, bool):
+        raise ValueError(
+            f"crs must be an int (EPSG code), str (WKT/PROJ), or None; "
+            f"got bool ({crs!r}). bool is an int subclass in Python, so "
+            f"passing True/False would otherwise be written as EPSG=1 / "
+            f"EPSG=0 -- neither resolves with any CRS database."
+        )
+    if isinstance(crs, int):
+        try:
+            from pyproj import CRS
+        except ImportError:
+            return
+        try:
+            CRS.from_epsg(crs)
+        except Exception as e:
+            if _geotiff_strict_mode():
+                raise
+            raise ValueError(
+                f"crs={crs!r} is not a valid EPSG code "
+                f"(pyproj: {type(e).__name__}: {e}). Pass a valid "
+                f"EPSG integer, a WKT string, or None."
+            ) from e
+        return
+    if isinstance(crs, str):
+        return
+    raise TypeError(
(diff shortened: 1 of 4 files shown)


brendancol:
## PR Review (self-review)

### Blockers

- [ ] `data.attrs['crs']` path bypasses the validator. Lines `_writers/eager.py:534`, `_writers/gpu.py:370`, and `_writers/eager.py:829` (the `_write_vrt_tiled` branch) do `elif crs_attr is not None: epsg = int(crs_attr)`. Passing `attrs={'crs': True}` and calling `to_geotiff(da, path)` with no explicit `crs=` kwarg still writes `EPSG=1`. Verified locally: round-trips with `attrs['crs'] == 1`. The validator needs to run on `crs_attr` after the `isinstance(crs_attr, str)` branch, not only on the kwarg.

### Suggestions

- [ ] `np.int64(4326)` / `np.integer` scalars now raise `TypeError` from `_validate_crs_arg` because `isinstance(np.int64(...), int)` is `False`. Before this PR they silently fell through to "no EPSG written" — different behaviour, possibly a regression for users with numpy-typed CRS values. Either accept `np.integer` (and coerce to `int`) or document the change. Same applies inside `_resolve_crs_to_wkt`.
- [ ] Reader-side `_geotags.py:557` was called out in the original report and is not touched here. Files written by other tools can still produce `attrs['crs'] = <bad int>` on read. Acceptable to leave for a follow-up, but worth noting in the PR body so reviewers don't expect closure of that gap.

### Nits

- [ ] `_resolve_crs_to_wkt` now runs `_validate_crs_arg` and then redoes the EPSG-resolves check via `CRS.from_epsg(crs).to_wkt()`. Two pyproj calls per int; cheap but slightly redundant.
- [ ] `_validate_crs_arg` raises `ValueError` for bool and `TypeError` for other non-int/non-str. Mixed types are intentional (semantic vs. type), but a reader scanning the test file might miss why. Could add a one-line comment to that effect.

### What looks good

- Bool rejection ordering: `isinstance(crs, bool)` check runs before `isinstance(crs, int)`, which is the correct order since bool is an int subclass.
- Pyproj is treated as optional — the EPSG-resolves check is skipped when pyproj is missing, matching the rest of the module.
- `XRSPATIAL_GEOTIFF_STRICT=1` re-raise path is preserved through the new validator.
- Test file covers the validator directly plus integration through `to_geotiff` and the `.vrt` dispatch.

### Checklist

- [x] Algorithm matches reference/paper — not applicable (validation logic)
- [x] All implemented backends produce consistent results — same validator runs in eager / GPU / VRT writers
- [ ] Edge cases covered by tests — numpy-int case is not covered
- [x] No new benchmark needed
- [ ] Docstrings present and accurate
- [ ] `data.attrs['crs']` path covered — **see blocker above**

Issue #1480, opened 2026-05-05
Fetch COG tiles concurrently in HTTP path to mask RTT


brendancol:
## Reason or Problem

`_read_cog_http` in `xrspatial/geotiff/_reader.py` walks tiles sequentially. Each tile is fetched with its own HTTP range request, and each request blocks on the previous one before the next is sent.

For a COG with N tiles read over a high-latency link, total wall time is bounded by N × RTT. A 100-tile COG over a 50 ms RTT link spends 5 seconds on round trips alone before any data flows.

## Proposal

Fetch tile ranges concurrently with a small thread pool. Each tile still issues its own HTTP range request, but the requests overlap on the wire so RTT is hidden behind concurrency.

**Design:**

Add a helper on `_HTTPSource` (or alongside `_read_cog_http`) that takes a list of `(offset, byte_count)` ranges and returns the bytes for each, in input order:

```python
def read_ranges(self, ranges: list[tuple[int, int]], max_workers: int = 8) -> list[bytes]:
    ...
```

Implementation uses `concurrent.futures.ThreadPoolExecutor`. `urllib3.PoolManager` is already used by `_HTTPSource` and is thread-safe, so the existing pool handles connection reuse across worker threads.

In `_read_cog_http`, replace the sequential loop:

```python
for tr in range(tiles_down):
    for tc in range(tiles_across):
        ...
        tile_data = source.read_range(off, bc)
        # decode + place tile
```

with two passes: collect all `(offset, byte_count, placement)` entries first, fetch them with `read_ranges`, then iterate the returned bytes and decode/place each tile.

`_HTTPSource.read_range` stays unchanged, so no other call sites are affected.

**Usage:**

No API change. Existing `read_to_array(http_url)` calls benefit automatically.

**Value:**

For 100 tiles × 50 ms RTT with 8 workers, total fetch drops from ~5 s to ~600 ms. Local file reads are unaffected; this only touches `_read_cog_http`.

## Stakeholders and Impacts

Touches `_HTTPSource` and `_read_cog_http`. No public API change.

## Drawbacks

- Extra worker threads (small constant).
- Servers without HTTP/1.1 keep-alive may see more connection setup. The urllib3 PoolManager already in use mitigates that.

## Alternatives

- Multipart Range requests (`bytes=0-N1,N2-N3,...`). Cuts request count further but server support is uneven; some CDNs return 200 with the full body, others reject with 400/416. Could be layered on top of the threadpool fallback later.
[...]

PR #1981, opened 2026-05-15
geotiff: parallelise tile decode in _fetch_decode_cog_http_tiles (#1980)

### LINKED ISSUE #1980: perf(geotiff): parallelise tile decode in _fetch_decode_cog_http_tiles
The HTTP COG reader at ``xrspatial/geotiff/_reader.py`` already fetches tile bytes concurrently via ``read_ranges_coalesced`` (issues #1480 and #1487). The Pass 3 decode loop, however, runs sequentially:

```python
for (band_idx, tr, tc), tile_data in zip(placements, tile_bytes_list):
    tile_pixels = _decode_strip_or_tile(...)
    ...
```

The local-file path in the same module (``_read_tiles`` around line 2017) parallelises decode whenever ``tile_pixels >= 64 * 1024`` and ``n_tiles > 1``:

```python
use_parallel = (n_tiles > 1 and tile_pixels >= 64 * 1024)
if use_parallel:
    with ThreadPoolExecutor(max_workers=n_workers) as pool:
        decoded = list(pool.map(_decode_one, tile_jobs))
```

Codec extensions (deflate, zstd, LZW) release the GIL, so the thread pool actually overlaps decode work across cores. For wide windowed reads of multi-tile COGs the HTTP path was leaving that parallelism on the table after a concurrent fetch.

Pass 10 of the geotiff performance sweep flagged this asymmetry. Fix mirrors the local-path threshold and pool inside ``_fetch_decode_cog_http_tiles``; the placement loop that writes pixels into ``result`` stays serial to avoid contended writes.

Found by /sweep-performance.

brendancol:
## Summary

- The HTTP COG reader fetched tiles concurrently (issues #1480, #1487) but decoded them sequentially in a Python ``for`` loop. The sibling local-file path in the same module (``_read_tiles``) already parallelises decode whenever ``tile_pixels >= 64 * 1024`` and ``n_tiles > 1``; the HTTP path was structurally similar but never adopted the same gate.
- Mirror the local-path threshold and ``ThreadPoolExecutor`` in ``_fetch_decode_cog_http_tiles``. Codec extensions (deflate, zstd, LZW) release the GIL inside their C implementations, so the pool overlaps decode work across cores. The placement loop that writes pixels into ``result`` stays serial to avoid contended writes.
- 5 new tests in ``test_cog_http_parallel_decode_2026_05_15.py``: parallel + serial round-trip correctness, structural pool-instantiation check above the threshold, single-tile path skips the decode pool, ``_decode_strip_or_tile`` call count == ``n_tiles``.

Pass 10 of /sweep-performance on the geotiff module. Closes #1980.

## Test plan

- [x] New tests pass: ``python -m pytest xrspatial/geotiff/tests/test_cog_http_parallel_decode_2026_05_15.py``.
- [x] Existing COG/HTTP suite passes: ``python -m pytest xrspatial/geotiff/tests/ -k 'cog or http'`` -> 262 passed.
- [x] Full geotiff suite passes apart from two pre-existing failures (``test_predictor2_big_endian_gpu_1517``, ``test_size_param_validation_gpu_vrt_1776::test_tile_size_positive_works``) flagged in prior sweep notes.


brendancol:
## PR Review: geotiff: parallelise tile decode in `_fetch_decode_cog_http_tiles`

- [ ] `_reader.py:2747`: the 64K-pixel threshold now appears in two places, `_read_tiles` (line 2017 pre-PR) and `_fetch_decode_cog_http_tiles`. Pull it up to a module-level constant so the comment about why the bound is inclusive lives in one spot.
- [ ] `_reader.py:2755`: `_decode_one` is only defined inside the `if decode_in_parallel:` branch, and the `else` branch inlines the same `_decode_strip_or_tile` call. `_read_tiles` defines `_decode_one` once above the branch and reuses it for both paths.

### Nits
- [ ] `_reader.py:2752`: `from concurrent.futures import ThreadPoolExecutor as _Pool`. `_read_tiles` imports it under its real name; small consistency thing.

### What looks good
- The gate matches `_read_tiles` exactly (`n_tiles > 1 and tile_pixels >= 64*1024`), so HTTP and local paths agree on when to spin up a pool.
- Pool is bounded by `min(n_decode_tiles, cpu_count())` and cleaned up via `with`, so no leaked threads on decode error.
- Placement loop stays serial, so there's no race on `result`.
- Tests cover the parallel branch, the serial branch, pool instantiation count, and a monkey-patched call-count check on `_decode_strip_or_tile` that would catch a dropped or duplicated tile.
- Test fixtures use a real range-aware HTTP server rather than mocks.

### Checklist
- [x] No algorithmic change; output bytes match the serial path.
- [x] Backend parity not applicable (HTTP COG path is numpy-only).
- [x] NaN handling not applicable (decode is byte-exact).
- [x] Edge cases covered (single-tile path skipped, threshold boundary tested).
- [ ] Dask chunk boundaries not exercised by this PR.
- [x] No new materialisation or copies.
- [ ] No benchmark added; sweep notes cover the perf budget.
- [x] README feature matrix N/A.
- [x] Docstrings present.

I pushed a follow-up commit (164c524) applying the two suggestions and the nit.

PR #1982, opened 2026-05-15
geotiff: reject (y, x, time) 3D writer inputs (#1972)

### LINKED ISSUE #1972: geotiff: (y, x, time) DataArrays silently treated as multiband rasters
**Describe the bug**

`xrspatial/geotiff/_validation.py:46` accepts any 3D writer input whose first two dims are spatial, regardless of the third dim name. The CPU writer (`_writers/eager.py:627`) and GPU writer (`_writers/gpu.py:335`) both treat that trailing axis as bands. Temporal stacks with `dims=('y','x','time')` are written as a 3-band TIFF without warning.

`('time','y','x')` is rejected (the existing `_validate_3d_writer_dims` catches it), so the failure mode is asymmetric.

**Reproduction**

```python

da = xr.DataArray(np.zeros((4,4,3), dtype=np.float32),
                  coords={'y': np.arange(4), 'x': np.arange(4), 'time': np.arange(3)},
                  dims=('y','x','time'))
    to_geotiff(da, f.name)  # writes 3-band TIFF, no error
```

**Expected behavior**

When the trailing dim of a 3D input is a known temporal name (e.g. `time`, `t`, `date`, `datetime`), `to_geotiff` should refuse with a clear error. The caller can reshape, rename the dim, or pick a single time slice explicitly. The fallback that permits `(y, x, <unknown>)` for raw numpy callers should remain in place for genuinely unknown dim names.

**Additional context**

This is the same silent-corruption class as issue #1812 (which added the original `_validate_3d_writer_dims`), extended to the case where the third dim is a known non-band name rather than an unknown one. The fix lives in `_validate_3d_writer_dims` so all writer entry points pick it up.

brendancol:
Closes #1972.

## Summary

- `_validate_3d_writer_dims` now refuses 3D writer inputs whose trailing dim is a known temporal name (`time`, `t`, `date`, `datetime`, `times`, `dates`). The mirror case `(time, y, x)` was already rejected by the existing layout check; this closes the asymmetry.
- The error message points the caller at `isel`, a reduction, or a rename to `band` if the temporal axis is genuinely the band axis.
- The bare `(y, x, *)` fallback for raw numpy callers building band-last arrays stays in place for genuinely unknown trailing dim names.

## Test plan

- [x] New `xrspatial/geotiff/tests/test_temporal_3d_writer_rejection_1972.py` covers each temporal alias, each y/x alias, the band-leading and band-trailing accept cases, the regression on `(time, y, x)`, and the writer entry point round trip.
- [x] Full `pytest xrspatial/geotiff/tests/` — same 8 pre-existing failures as main, none related.

### DIFF
--- a/xrspatial/geotiff/_runtime.py
+++ b/xrspatial/geotiff/_runtime.py
@@ -62,6 +62,12 @@ _VRT_PATH_MISSING_SENTINEL = object()
 _Y_DIM_NAMES = ('y', 'lat', 'latitude', 'row')
 _X_DIM_NAMES = ('x', 'lon', 'longitude', 'col')
 
+# Temporal dim names. Used by the 3D writer validator (#1972) to refuse
+# ``(y, x, <temporal>)`` inputs that would otherwise be silently treated
+# as multiband rasters. CF / xarray conventions cover ``time`` and ``t``;
+# the rest match common upstream-pipeline aliases.
+_TIME_DIM_NAMES = ('time', 't', 'date', 'datetime', 'times', 'dates')
+
 
 class GeoTIFFFallbackWarning(UserWarning):
     """Warning emitted when a geotiff helper falls back to a slower path.
--- a/xrspatial/geotiff/_validation.py
+++ b/xrspatial/geotiff/_validation.py
@@ -17,6 +17,17 @@ from ._coords import _BAND_DIM_NAMES
 from ._runtime import _TIME_DIM_NAMES, _X_DIM_NAMES, _Y_DIM_NAMES
 
 
+def _is_temporal_dim_name(name) -> bool:
+    """Return True if ``name`` is a known temporal dim alias.
+
+    Compared case-insensitively against ``_TIME_DIM_NAMES`` so that
+    CF-style ``'TIME'`` / ``'Time'`` reach the friendly temporal error
+    in the 3D writer validator instead of slipping through the
+    ``(y, x, *)`` band-position fallback (#1972).
+    """
+    return isinstance(name, str) and name.lower() in _TIME_DIM_NAMES
+
+
 def _validate_3d_writer_dims(dims) -> None:
     """Reject ambiguous 3D writer inputs (issue #1812).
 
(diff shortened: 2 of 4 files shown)


brendancol:
## PR Review (self-review)

### Suggestions

- [ ] Case-sensitivity gap. `_TIME_DIM_NAMES = ('time', 't', 'date', 'datetime', 'times', 'dates')` is matched with `in` (case-sensitive). Verified locally: `('y', 'x', 'TIME')` still slips through and writes a 3-band TIFF, because `'TIME'` is not in the tuple and the `(y, x, *)` fallback fires. Many real datasets use `'TIME'` or `'Time'` (CF allows either). Either store lowercase variants in the tuple, or compare with `d2.lower() in _TIME_DIM_NAMES`. The existing `_Y_DIM_NAMES` / `_X_DIM_NAMES` are also case-sensitive, but the failure mode there is loud rejection rather than silent corruption — temporal trailing dim is the bigger risk and warrants case-insensitive matching.

### Nits

- [ ] The error message references `data.isel({d2}=0)` literally. For `d2='time'` this renders as `data.isel(time=0)`, which is correct. For aliases like `d2='dates'` it renders as `data.isel(dates=0)`, which is also correct because that's the actual dim name. Fine as-is, but worth a passing read in case an alias was added that wasn't a real xarray dim name.
- [ ] `(time, y, x)` still falls through to the generic `"ambiguous dims"` error from the existing validator. The new dedicated temporal message is only attached to the trailing-dim case. Could symmetrise by checking `d0` against `_TIME_DIM_NAMES` too and emitting the same friendly message, though the asymmetry is functional (the leading case was already rejected).

### What looks good

- The fix lands in `_validate_3d_writer_dims`, so the eager, GPU, and `.vrt`-dispatch writer entry points all pick it up without further changes.
- The `(y, x, *)` fallback for genuinely unknown dim names (raw numpy callers building band-last arrays) stays intact.
- Error message points the caller at concrete fixes: `isel`, a reduction, or rename to `band`.
- Tests cover each temporal alias, each y/x alias paired with `time`, both accept cases (band-leading / band-trailing), and the regression on `(time, y, x)`.

### Checklist

- [x] Algorithm matches reference/paper — not applicable
- [x] All implemented backends covered — validator is shared across writer entry points
- [ ] Edge cases covered — uppercase / mixed-case temporal names slip through (see suggestion)
- [x] No new benchmark needed

PR #1983, opened 2026-05-15
geotiff: refuse non-numeric nodata / _FillValue early (#1973)

### LINKED ISSUE #1973: geotiff: non-numeric _FillValue crashes with opaque TypeError from np.isnan
**Describe the bug**

`xrspatial/geotiff/_attrs.py:228` returns a non-numeric `_FillValue` verbatim with a comment that this is intentional pass-through. Later, `xrspatial/geotiff/_writers/eager.py:654` calls `np.isnan(nodata)`, which raises a `TypeError` with no reference to the offending attribute.

**Reproduction**

```python

da = xr.DataArray(np.full((4,4), np.nan, dtype=np.float32),
                  coords={'y': np.arange(4), 'x': np.arange(4)},
                  dims=('y','x'),
                  attrs={'_FillValue': 'missing'})
    to_geotiff(da, f.name)
# TypeError: ufunc 'isnan' not supported for the input types,
# and the inputs could not be safely coerced to any supported
# types according to the casting rule ''safe''
```

**Expected behavior**

`_resolve_nodata` should reject a non-numeric `_FillValue` (or `nodata`) at the validation boundary with a clear `ValueError` naming the attribute, the value, and that nodata must be numeric. The `nodatavals` branch already filters non-numeric entries; the `_FillValue` branch should do the same instead of relying on a downstream NumPy ufunc to crash.

**Additional context**

Found in a code review of the geotiff module alongside #1971 and #1972.

brendancol:
Closes #1973.

## Summary

- `_validate_nodata_arg` rejects non-numeric `nodata=` kwargs at the writer entry point. `to_geotiff`, `_write_vrt_tiled`, and `write_geotiff_gpu` all call it before any downstream `np.isnan` usage.
- `_resolve_nodata_attr` now raises `ValueError` naming the offending attribute when `attrs['nodata']` or `attrs['_FillValue']` is non-numeric, instead of returning the value verbatim and letting NumPy crash later with a generic ufunc TypeError.
- `attrs['nodatavals']` (rioxarray's per-band tuple) keeps its existing skip-on-non-numeric behaviour, since those values often come from arbitrary upstream pipelines and a single bad entry should not block writing.

## Test plan

- [x] New `xrspatial/geotiff/tests/test_nodata_validation_1973.py` covers `_validate_nodata_arg`, `_resolve_nodata_attr` (both attrs), the writer kwarg path, the `_FillValue` attr path, and the `.vrt` dispatch path.
- [x] Full `pytest xrspatial/geotiff/tests/` — same 8 pre-existing failures as main, none related.

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -206,6 +206,18 @@ def _resolve_nodata_attr(attrs: dict):
     """
     nodata = attrs.get('nodata')
     if nodata is not None:
+        try:
+            float(nodata)
+        except (TypeError, ValueError) as e:
+            raise ValueError(
+                f"attrs['nodata']={nodata!r} is not numeric "
+                f"({type(nodata).__name__}). The writer needs a numeric "
+                f"sentinel to compare against pixel values; passing a "
+                f"non-numeric value would otherwise crash inside "
+                f"``np.isnan`` with an opaque ufunc error. Drop the "
+                f"attr, replace it with a numeric sentinel, or pass "
+                f"``nodata=`` explicitly (issue #1973)."
+            ) from e
         return nodata
 
     vals = attrs.get('nodatavals')
@@ -229,8 +241,16 @@ def _resolve_nodata_attr(attrs: dict):
     if fill is not None:
         try:
             ffv = float(fill)
-        except (TypeError, ValueError):
-            return fill  # non-numeric -- pass through verbatim
+        except (TypeError, ValueError) as e:
+            raise ValueError(
+                f"attrs['_FillValue']={fill!r} is not numeric "
+                f"({type(fill).__name__}). The writer needs a numeric "
+                f"sentinel to compare against pixel values; passing a "
+                f"non-numeric value would otherwise crash inside "
+                f"``np.isnan`` with an opaque ufunc error. Drop the "
+                f"attr, replace it with a numeric sentinel, or pass "
+                f"``nodata=`` explicitly (issue #1973)."
+            ) from e
         if np.isnan(ffv):
             return None
         return fill
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -17,6 +17,8 @@ Extracted in step 5 of issue #1813.
 """
 from __future__ import annotations
 
+import warnings
+
 import numpy as np
 
 from ._coords import (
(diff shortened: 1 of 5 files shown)


brendancol:
## PR Review (self-review)

### Suggestions

- [ ] Error-type drift versus the pre-existing bool check. `_writers/eager.py:261` already rejects `nodata=True` with `TypeError`, and `_geotags.py:1165` does the same. The new `_validate_nodata_arg` raises `ValueError`. In practice the bool check fires first, so `nodata=True` still raises `TypeError` from the existing site — but a future caller (`write_geotiff_gpu`, `_write_vrt_tiled`) hits my validator before the bool check exists in their path, and a bool would slip through (`float(True) == 1.0`). Worth adding a `isinstance(nodata, bool)` branch to `_validate_nodata_arg` so all three writer entry points reject bool the same way (consistent with `_validate_crs_arg` in #1978).
- [ ] `attrs['nodatavals'] = ('foo', 'bar')` (all non-numeric) silently resolves to `nodata=None` — the per-band skip loop never produces a numeric and the function returns `None` from the final fall-through. That matches the docstring ("skips non-numeric entries"), but a tuple where *every* entry is non-numeric is probably user error rather than a legitimate "no sentinel" case. Worth flagging when at least one non-numeric entry was seen and none were usable. Not blocking.

### Nits

- [ ] The error messages on the two `_resolve_nodata_attr` branches (`attrs['nodata']` and `attrs['_FillValue']`) are near-duplicates. Could factor into a helper or share a string. Minor.
- [ ] `_FillValue='NaN'` (the string) is accepted because `float('NaN')` succeeds, then treated as NaN and returns `None`. That's the documented behaviour for NaN values, just worth being aware that a string `'NaN'` is now indistinguishable from a real NaN. Probably fine.

### What looks good

- Validator runs at the boundary (`to_geotiff`, `_write_vrt_tiled`, `write_geotiff_gpu`), so the kwarg path is closed at each writer.
- `_resolve_nodata_attr` raises with the attr name in the message (`attrs['_FillValue']=...`), so the caller can tell which attr is wrong.
- `attrs['nodatavals']` keeps its per-entry skip behaviour — correct, since rioxarray's per-band tuples often carry partial sentinels.
- Tests cover the validator, both attr branches, the kwarg path, and the `.vrt` dispatch.

### Checklist

- [x] Algorithm matches reference/paper — not applicable
- [x] All implemented backends produce consistent results — validator runs in eager / GPU / VRT writers
- [x] NaN handling — `_FillValue=NaN` correctly returns `None`
- [x] Edge cases covered by tests — `None`, numeric, NaN-string, non-numeric, sequence
- [x] No new benchmark needed
- [ ] `nodata=True` reaches the new validator unguarded in the GPU and VRT writers — see suggestion

Issue #1984, opened 2026-05-15
geotiff: define a public contract for DataArray attrs (canonical / alias / pass-through)


brendancol:
## Reason or Problem

`xrspatial.geotiff` populates `DataArray.attrs` with a large surface area: `crs`, `crs_wkt`, `transform`, `nodata`, `nodatavals`, `_FillValue`, `raster_type`, `extra_tags`, `gdal_metadata`, `gdal_metadata_xml`, `image_description`, `extra_samples`, `colormap`, `colormap_rgba`, `cmap`, `x_resolution`, `y_resolution`, `resolution_unit`, `crs_name`, plus a long list of GeoKey-derived fields (`geog_citation`, `datum_code`, `angular_units`, `linear_units`, `semi_major_axis`, `inv_flattening`, `projection_code`, `vertical_crs`, `vertical_citation`, `vertical_units`).

Today there is no documented contract telling callers which keys are canonical, which are compatibility aliases, and which are pass-through. Downstream code is left guessing. For example, `attrs['nodata']`'s presence currently doubles as a flag that NaN-masking has already run (see comment at `_attrs.py:91-92`), which the reviewer flagged as overloaded.

## Proposal

Define and publish a three-tier classification for every attr emitted by the read path and consumed by the write path:

1. **Canonical** — owned by xrspatial. Round-trip stable. Examples: `crs`, `crs_wkt`, `transform`, `raster_type`, `extra_tags`, `gdal_metadata`/`gdal_metadata_xml`, the resolution group.
2. **Compatibility alias** — read for interop with other ecosystems (rioxarray `nodatavals`, CF `_FillValue`). Never emitted by our writers when a canonical key is present.
3. **Best-effort pass-through** — preserved through round-trip when possible but not guaranteed semantically. Examples: GeoKey-derived metadata that the writer cannot fully reconstruct.

**Design:**
- New module docstring in `xrspatial/geotiff/_attrs.py` listing every attr key by tier with a one-line semantic definition.
- A `attrs_contract.rst` page under `docs/source/` with the same table, plus the round-trip invariant per tier.
- Tests that lock the contract: a single fixture exercising every attr, with explicit assertions per tier.
- Pin `attrs['nodata']` semantics specifically. This is the subject of issue #5 (split declared vs masked nodata) and that issue is the prerequisite for nailing this one down.

**Usage:**
Library users read the contract page to know which keys they can rely on. Internal code stops adding new attrs without first picking a tier.

**Value:**
[...]

Issue #1930, opened 2026-05-15
geotiff: first-class backend parity contract with a golden test corpus


brendancol:
## Reason or Problem
The geotiff module spans many roles: TIFF parser, COG range client, VRT engine, CRS normaliser, CPU writer, streaming writer, GPU reader and writer, and a security boundary. The test count is high, which is good, but the number of backend-specific parity patches in recent PRs (and the per-finding comments referencing them) is a signal that behaviour drift between backends is a persistent risk.

The current safety net is implicit. Cross-backend parity is asserted only where a specific bug surfaced. There is no single contract that says "eager, dask, GPU, VRT, HTTP, and fsspec must all agree on pixels, transform, CRS, nodata, dtype, attrs."

## Proposal
Make backend parity a first-class contract with a small golden corpus.

**Design:**
Curate a corpus of small TIFFs that cover the dimensions where drift has shown up: tiled vs stripped, planar vs chunky, byte order, dtype range (int/uint/float, native and big-endian, predictor=2/3), compression (none, deflate, lerc, jpeg, lzw), nodata sentinels (int + NaN + miniswhite), overviews, GDAL_METADATA, and CRS variants (EPSG, WKT, citation-only).

For each file, run it through every backend that supports it (eager, dask, GPU, VRT, HTTP, fsspec where applicable) and assert exact match against a `rasterio` / GDAL baseline on pixels, transform, CRS, nodata, dtype, and attrs.

**Usage:**
The corpus runs as part of the normal test suite. New backends or kwargs that change pixel-level output bake in a new baseline only when the change is intentional.

**Value:**
Today each parity bug surfaces from a different angle and is fixed in isolation. A golden corpus catches drift in a single place and makes accepting a new pixel-level change an explicit baseline update rather than an implicit one.

## Drawbacks
- Maintenance cost: baseline updates need review.
- Corpus growth: keep it small (under 50 files) and put large fixtures behind a marker.
- Requires rasterio / GDAL in CI for the comparison run.

## Alternatives
- Continue with per-bug parity tests. Lower up-front cost, higher long-tail cost.
- Property-based / hypothesis-style parity. Useful but harder to debug when it fails.

## Stakeholders and Impacts
Affects every backend. Owner is whoever lands the corpus. Impact is on CI run time (modest), and on the workflow for any future change that produces different pixels (must explicitly bless the baseline).

## Additional Notes or Context
[...]

PR #1991, opened 2026-05-16
geotiff: golden corpus oracle harness (#1930, phase 1.2)

brendancol:
## Summary

Phase 1 PR 2 of #1930 (golden corpus for geotiff parity).

From the plan:

> **PR 2 -- Oracle harness.** `tests/golden_corpus/_oracle.py` with
> `compare_to_oracle(fixture, candidate_da)` that opens the fixture with
> rasterio and asserts pixel / transform / CRS / nodata / dtype /
> canonical-attrs match the candidate. NaN-aware float comparison.
> No backends wired yet.

One public function:

```python
def compare_to_oracle(fixture_path, candidate_da, *, lossy=False) -> None
```

Reads the fixture with rasterio and asserts, in order:

1. dtype (exact)
2. transform (rasterio `Affine` vs xrspatial `attrs['transform']`, 1e-9 tol)
3. CRS (falls back to EPSG-code equality when two WKTs describe the same
   coordinate system)
4. nodata (NaN-aware)
5. pixel values (bit-exact for ints, `array_equal(..., equal_nan=True)` for floats)

`lossy=True` skips step 5 and checks shape only. That path is for the
JPEG cells in Phase 2 PR 5.

## Canonical-attrs hook (#1984)

The plan asks the oracle to assert "canonical-attrs match the candidate".
The canonical-attrs contract is tracked in #1984 and is not settled, so
this PR:

- asserts only the obvious subset (crs / transform / nodata / dtype)
  via the helpers listed above,
- adds a `_assert_canonical_attrs(ref_attrs, candidate_da)` hook so a
  later PR can extend the contract without touching call sites,
- leaves a `TODO(#1984)` in that function.

## Out of scope

- Manifest schema / generator (Phase 1 PR 1, sibling PR).
- Fixture `.tif` files (Phase 2 PRs 3-9).
- Wiring the oracle into per-backend test cells (Phase 3 PRs 10-15).
- Pass-through GeoTIFF metadata (`gdal_metadata`, `extra_tags`, etc.) --
  those wait on #1984.

The oracle takes a raw filesystem path, not a manifest entry, so it
does not import anything from the sibling Phase 1 PR 1 work.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/golden_corpus/test_oracle.py` -- 16 tests, all green
- [x] Success: int, float-with-NaN, CRS-via-WKT, lossy
- [x] Failure: mismatched dtype / transform / CRS / nodata-NaN-vs-zero /
      missing nodata / pixel / pixel-NaN-vs-zero / lossy-dtype /
      lossy-transform / lossy-shape / missing-file
- [x] flake8 clean on the new files

### DIFF
--- a/xrspatial/geotiff/tests/golden_corpus/__init__.py
+++ b/xrspatial/geotiff/tests/golden_corpus/__init__.py
@@ -1,0 +1,12 @@
+"""Golden corpus for xrspatial geotiff parity tests (issue #1930).
+
+This package hosts the corpus fixtures, manifest, and oracle harness used
+by Phase 3 backend cells. The pieces are split across PRs so they can
+land independently:
+
+* Phase 1 PR 1 -- manifest + deterministic generator. The contract for
+  what a fixture is.
+* Phase 1 PR 2 -- the oracle harness (``_oracle``).
+* Phase 2     -- the seed fixtures themselves.
+* Phase 3     -- the per-backend test cells that call the oracle.
+"""
--- a/xrspatial/geotiff/tests/golden_corpus/_oracle.py
+++ b/xrspatial/geotiff/tests/golden_corpus/_oracle.py
@@ -1,0 +1,399 @@
+"""Oracle harness for the geotiff golden corpus (issue #1930, Phase 1.2).
+
+``compare_to_oracle(fixture_path, candidate_da)`` reads ``fixture_path`` with
+rasterio (the reference implementation) and asserts that an xrspatial-produced
+``xarray.DataArray`` agrees on every observable property: pixel values,
+dtype, transform, CRS, nodata sentinel, and a small set of canonical attrs.
+
+The Phase 3 backend cells (numpy, dask+numpy, cupy, dask+cupy, HTTP, VRT)
+each call this single function with the DataArray they produced; the oracle
+encapsulates "what parity means" so every backend agrees on it.
+
+Scope notes:
+
+* The full canonical-attrs contract is tracked in issue #1984 and is not
+  settled yet. Until it lands, the oracle asserts only the obvious subset
+  (``crs``, ``transform``, ``nodata``, ``dtype``) and leaves a hook
+  (``_assert_canonical_attrs``) that later PRs can fill in. See the TODO
+  in that function.
+* ``lossy=True`` skips bit-exact pixel comparison (for JPEG cells in
+  Phase 2) and instead verifies only shape, dtype, transform, and CRS.
+* The oracle does not import the corpus manifest from Phase 1 PR 1; it
+  takes a raw filesystem path so the two PRs stay decoupled.
+"""
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+import xarray as xr
+
+
+# ---------------------------------------------------------------------------
+# Rasterio lazy import
+# ---------------------------------------------------------------------------
+
+def _require_rasterio():
+    """Import rasterio lazily so the module is importable in environments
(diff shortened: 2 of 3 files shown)


brendancol:
## Self-review

### Blockers

None.

### Suggestions (worth addressing before this lands)

- [ ] `_oracle.py:268` -- `_assert_transform` short-circuits when the
      ref transform matches `Affine.identity()`. That's wrong: a real
      raster written at origin (0, 0) with 1.0 pixel size *is* an
      identity transform, and the oracle would silently skip the
      transform check for it. Several of my own tests use exactly that
      shape (`from_origin(0.0, H, 1.0, 1.0)` is close to identity when
      H==1, and not identical otherwise), and they pass only because
      both sides happen to agree. Better heuristic: treat "no georef"
      as `src.crs is None and src.transform.is_identity`, not transform
      alone. Latent bug, not blocking the demo, but worth fixing now
      before Phase 3 starts relying on it.

- [ ] `_oracle.py:268` -- `ref_transform is None` is dead code.
      `rasterio.open(...).transform` returns `Affine.identity()` when
      the file has no georef, not `None`. Drop the check or replace it
      with an `assert ref_transform is not None`.

### Nits

- [ ] `_oracle.py:236-247` -- `_assert_canonical_attrs` is a stub. The
      `_ = ...  # silence unused warning` lines work but read awkwardly.
      Rename the parameters to `_ref_attrs` / `_candidate_da` and the
      whole silencing block goes away.

- [ ] `test_oracle.py` -- the `test_lossy_rejects_shape_mismatch`
      docstring could mention that this is the only test that exercises
      `_assert_shape_only`. Helps a future reader looking for coverage
      of that helper.

### What I like about it

- The per-property helpers stay small and additive. When #1984 lands,
  `_assert_canonical_attrs` is the only function that grows.
- Falling back to EPSG-code equality for CRS is the safer choice;
  structural WKT comparison breaks across libgeotiff / PROJ versions
  and we'd spend time chasing whitespace-in-WKT failures otherwise.
- Tests don't depend on the Phase 1 PR 1 manifest or generator. That
  keeps this PR mergeable independently of the sibling.

### Checklist

- [x] Algorithm matches reference (rasterio is the oracle by design)
- [x] NaN handling correct
- [x] Edge cases covered (NaN-vs-zero, EPSG-equivalent WKT, missing file)
- [n/a] Dask chunk boundaries -- no dask code
- [x] No premature materialization beyond the unavoidable `.compute()` / `.get()` in `_candidate_pixels`
- [n/a] Benchmark -- test infrastructure
- [n/a] README feature matrix -- test infrastructure

brendancol:
Fixes pushed in 6cdbb71:

- **Suggestion 1 (`_is_default_transform` short-circuit):** fixed. New
  `_ref_has_georef(src)` helper requires `src.crs is not None` *or* a
  non-identity `src.transform` before declaring the file georeferenced.
  Added `test_identity_transform_with_crs_still_compared` which fails on
  the old behaviour and passes on the new.
- **Suggestion 2 (`ref_transform is None` is dead code):** dropped. The
  `_assert_transform` signature now takes an explicit `ref_has_georef`
  kwarg from the caller.
- **Nit 1 (stub parameter names):** `_assert_canonical_attrs` parameters
  renamed to `_ref_attrs` / `_candidate_da`; `_ = ...` silencing block is
  gone.
- **Nit 2 (test docstring):** `test_lossy_rejects_shape_mismatch`
  docstring now points out it's the only test that exercises
  `_assert_shape_only`.

Also added `test_no_georef_fixture_tolerates_missing_candidate_transform`
to lock in the legitimate xrspatial #1710 behaviour. 18 tests, all green,
flake8 clean.

PR #1992, opened 2026-05-16
geotiff: golden corpus layout + manifest schema (#1930, phase 1.1)

brendancol:
## Summary

First PR of the Phase 1 split for #1930. Lands the corpus layout and
the manifest schema, nothing else.

From the implementation plan in #1930:

> **Corpus layout + manifest schema.** Add `xrspatial/geotiff/tests/golden_corpus/`
> with a YAML manifest describing every fixture (dtype, compression,
> predictor, nodata, crs, overviews, extra_tags) and a deterministic
> generator script that rebuilds the TIFFs from the manifest. No test
> cells yet, no fixtures yet. The generator is the contract for what a
> fixture is.

## What lands

- `manifest.yaml` -- schema documented in the header comment, one
  canonical example fixture covering every field, version: 1.
- `generate.py` -- deterministic generator. CLI flags: `--dry-run`,
  `--only <id>`, `--output-dir`. `generate.validate()` is the
  authoritative schema check.
- `README.md` -- how to regenerate, what each phase will add.
- `test_golden_corpus_manifest_1930.py` -- smoke test. Parses the
  manifest, validates every entry, runs the generator in dry-run mode,
  and checks the error messages a contributor is most likely to hit.

## Out of scope

- The oracle harness (`_oracle.py`). Sibling PR for Phase 1 PR 2 (#1991).
- Any real fixture `.tif` files. Phase 2 PRs add them in batches.
- Backend wiring (eager / dask / gpu / vrt / http). Phase 3.
- Adding `pyyaml` / `rasterio` to `setup.cfg`'s `tests` extra. The
  smoke test uses `importorskip` until Phase 2 needs real writes in CI.

## Test plan

- `pytest xrspatial/geotiff/tests/test_golden_corpus_manifest_1930.py`
  -- 12 tests, all pass locally.
- `python -m xrspatial.geotiff.tests.golden_corpus.generate --dry-run`
  -- exits 0, prints planned outputs.
- `python -m xrspatial.geotiff.tests.golden_corpus.generate --output-dir /tmp/x`
  -- writes the example fixture end-to-end with normalised mtime;
  byte-stable across reruns.

### DIFF
--- a/xrspatial/geotiff/tests/golden_corpus/README.md
+++ b/xrspatial/geotiff/tests/golden_corpus/README.md
@@ -11,7 +11,7 @@ are wired (Phase 3).
 golden_corpus/
   manifest.yaml   # the contract: every dimension a fixture can have
   generate.py     # deterministic generator that rebuilds .tif files
-  fixtures/       # output dir (populated by Phase 2 PRs, gitignored)
+  fixtures/       # output dir; populated and committed by Phase 2 PRs
   README.md       # this file
 ```
 
--- a/xrspatial/geotiff/tests/golden_corpus/README.md
+++ b/xrspatial/geotiff/tests/golden_corpus/README.md
@@ -1,0 +1,64 @@
+# Geotiff golden corpus
+
+Tracking issue: [#1930](https://github.com/xarray-contrib/xarray-spatial/issues/1930).
+This is Phase 1 PR 1 of that plan: corpus layout + manifest schema only.
+No actual `.tif` files live in this directory yet (Phase 2). No backends
+are wired (Phase 3).
+
+## Layout
+
+```
+golden_corpus/
+  manifest.yaml   # the contract: every dimension a fixture can have
+  generate.py     # deterministic generator that rebuilds .tif files
+  fixtures/       # output dir (populated by Phase 2 PRs, gitignored)
+  README.md       # this file
+```
+
+## Regenerating the corpus
+
+From the repository root, with the `tests` extras and `rasterio` +
+`pyyaml` installed:
+
+```
+python -m xrspatial.geotiff.tests.golden_corpus.generate
+```
+
+Useful flags:
+
+* `--dry-run` validates the manifest and prints the planned outputs.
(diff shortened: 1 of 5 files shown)


brendancol:
## Self-review (PR #1992)

Read through the PR after opening and pushed a fix-up (`7dd04d7`) for
every item below.

### Suggestions (fixed)

- [x] `manifest.yaml` documented `compression_level` and `max_z_error`,
  but `_rasterio_kwargs` ignored both. Now threaded through to
  `zlevel` / `zstd_level` / `jpeg_quality` / `max_z_error` depending
  on the codec.
- [x] Validator did not enforce `predictor: 3` with float dtypes or
  `predictor: 2` with integer dtypes. Wrong pairings now raise a clear
  `ManifestError` instead of a confusing rasterio failure at write
  time.
- [x] `external_overview` was free-form; now type-checked as bool.
- [x] README claimed `fixtures/` was gitignored. Phase 2 commits the
  real files there, so the wording was wrong; replaced.

### Nits (fixed)

- [x] `typing.Iterable` -> `collections.abc.Iterable`.
- [x] `load_manifest` took a `path` argument but `generate()` never
  passed one. Added a `manifest_path` kwarg so it is reachable.

### Considered, not changed

- `_make_pixels` noise path on integer dtypes does its arithmetic in
  float64 across `info.max - info.min`. Fine for the geotiff-realistic
  dtypes (uint8/16/32, int8/16/32). int64 would lose precision but no
  realistic fixture uses it.
- Validator does not check `compression_level` ranges per codec
  (deflate 0-9, etc.). Easy to add but duplicates GDAL's own checks.
  Deferring until a Phase 2 fixture actually trips one.

### Test plan

- `pytest xrspatial/geotiff/tests/test_golden_corpus_manifest_1930.py`
  -- 15 pass (was 12 before the fix-up).
- `python -m xrspatial.geotiff.tests.golden_corpus.generate --output-dir /tmp/x`
  -- still writes a byte-stable .tif with the normalised mtime.

PR #1993, opened 2026-05-16
geotiff: golden corpus phase 2.3 - tiled/stripped + byte order (#1930)

brendancol:
## Summary

Phase 2 PR 3 of the implementation plan in #1930. Adds four fixtures
that cover the layout x byte_order cross product on a single small
uint16 raster, plus the on-disk byte-order fix the matrix depends on.

Quoting the plan:

> PR 3 - Tiled vs stripped + byte order variants (4 fixtures)
>
> Add four fixtures to the manifest that exercise the (layout x byte_order) cross product:
> - tiled + little-endian
> - stripped + little-endian
> - tiled + big-endian
> - stripped + big-endian

## Fixtures

All four are 32x32 uint16, `pixel_pattern: noise`, `pixel_seed: 1930`,
`nodata: 0`, no compression, EPSG:4326 default georef. The only axes
that vary are the two the matrix is meant to exercise. 32x32 is the
smallest size that lets GDAL keep the tiled flag (it drops the flag
when the image fits in a single tile). Files are ~2.5 KB each.

| id                    | layout    | byte_order | tile_size / blocksize |
|-----------------------|-----------|------------|-----------------------|
| `stripped_le_uint16`  | stripped  | little     | blocksize 16          |
| `tiled_le_uint16`     | tiled     | little     | tile_size 16          |
| `stripped_be_uint16`  | stripped  | big        | blocksize 16          |
| `tiled_be_uint16`     | tiled     | big        | tile_size 16          |

## Generator extension

The manifest schema already had `byte_order` in `REQUIRED_FIELDS` and
the validator accepted `big`, but `_rasterio_kwargs` was forwarding the
choice as `endian=big`. rasterio intercepts that lowercase kwarg and
drops it before it reaches GDAL, so big-endian fixtures came out
little-endian on disk. First commit changes that one line to use the
GDAL-side `ENDIANNESS=BIG` creation option, which rasterio forwards
verbatim. Isolated to one site in `_rasterio_kwargs`; no other
generator behaviour changes.

## Smoke test

`xrspatial/geotiff/tests/test_golden_corpus_layout_endian_1930.py`:

- asserts the four files exist and have the promised byte-order magic
  (`II` / `MM`) and tiled flag,
- builds an `xarray.DataArray` from the rasterio read of each fixture
  and passes it through `_oracle.compare_to_oracle` (identity round
  trip) so the oracle is confirmed to accept these fixtures before
  Phase 3 backend wiring lands.

## Test plan

- [x] `python -m pytest xrspatial/geotiff/tests/test_golden_corpus_layout_endian_1930.py -v`
- [x] `python -m pytest xrspatial/geotiff/tests/test_golden_corpus_manifest_1930.py xrspatial/geotiff/tests/golden_corpus/test_oracle.py`
- [x] `python -m xrspatial.geotiff.tests.golden_corpus.generate --dry-run`


brendancol:
## PR Review: golden corpus phase 2.3 - tiled/stripped + byte order

This is a fixtures-and-generator-fix PR, so the algorithm / backend
parity / benchmark sections of the standard review don't apply. Review
focuses on the generator fix, fixture determinism, and the smoke test.

None.

### Suggestions (addressed in follow-up commit)
- [x] `test_golden_corpus_layout_endian_1930.py`: `src.is_tiled` is
  pending-deprecation in rasterio 1.4+. Switched to
  `src.profile.get("tiled", False)` which carries the same flag without
  the warning.
- [x] Added `test_le_and_be_pixels_match` to assert the LE and BE
  siblings decode to the same logical pixels. Locks in the matrix
  invariant that on-disk byte order is the only axis that varies, so a
  future generator change can't silently desync the pairs ahead of
  Phase 3 backend comparisons.

### Nits (left as-is, optional)
- `test_fixture_round_trips_through_oracle` unpacks `_tiled, _magic`
  from the shared parametrize list but only uses `fid`. Could use a
  separate `parametrize("fid", [...])` for that test. Low value;
  re-using one list keeps the four ids in one place.
- `src.shape == (32, 32)` is hardcoded. If the manifest dimensions
  change later, this disagrees silently. Could be read from
  `generate.load_manifest()`. Not worth the indirection for a smoke
  test.

### What looks good
- Two-commit split keeps the one-line generator fix
  (`ENDIANNESS=BIG`) isolated from the fixture additions; the fix
  commit explains why `endian=big` was being dropped by rasterio.
- Fixtures are deterministic (`pixel_seed: 1930`, fixed mtimes via
  `DETERMINISTIC_EPOCH`, sorted output order); regeneration
  produces byte-identical files.
- All four fixtures share width / height / dtype / nodata /
  compression so the only axes that vary are the two the matrix is
  meant to exercise.
- 32x32 is the smallest size that keeps GDAL's tiled flag on disk
  (GDAL drops it when the image fits in one tile); file size stays
  under 2.5 KB each, well below the 4 KB cap.
- The oracle identity round-trip confirms `compare_to_oracle` accepts
  the four variants before Phase 3 backend wiring lands.

### Checklist
- [x] Generator fix isolated to one site, no behaviour drift
- [x] Fixtures deterministic and byte-stable across re-runs
- [x] All four files exist, parse as TIFFs, match declared layout and byte order
- [x] Oracle accepts the new fixtures (identity round-trip)
- [x] LE / BE siblings carry identical logical pixels
- [x] Files under 4 KB each (~2.5 KB)
- [x] Schema unchanged; the manifest extension was a generator-side bug fix
- [x] No `--no-verify`; commits go through hooks

Issue #1988, opened 2026-05-15
geotiff: split declared nodata from masked-data state in DataArray attrs


brendancol:
## Reason or Problem

`attrs['nodata']` is overloaded in the geotiff module. From `_attrs.py:91-92`:

> The `nodata` attr is intentionally NOT set here because each caller sets it next to its own nodata-masking step (the value's presence in attrs signals "this array has been NaN-masked").

So `attrs['nodata']` presence today means two things at once:

1. The file declared this sentinel value, and
2. The reader already replaced sentinel pixels with NaN in the array.

Downstream code cannot tell which. A consumer that wants to write the array back out using the same sentinel has no clean signal for whether the array is still in "raw int with sentinels" form or "float with NaNs masked." This is the root cause the reviewer pointed at: "downstream functions over-interpret `attrs['nodata']`."

## Proposal

Split the two meanings into two attrs:

- `attrs['nodata']` — the declared file sentinel value, as a scalar of the file's source dtype. Set whenever the source declared one. Independent of whether masking happened.
- `attrs['masked_nodata']` — boolean, true iff the reader replaced sentinel pixels with NaN in the in-memory array. Always set on read so its absence is meaningful too.

**Design:**

Read path:

- All four read backends (numpy / dask / gpu / dask+gpu) and the http/cog and vrt paths set `attrs['nodata']` from the source whenever the source declares one, regardless of whether the array is float-with-NaN or int-with-sentinels.
- Each backend sets `attrs['masked_nodata']` to True iff it performed the float-and-mask step. For int reads that keep sentinels as-is, set it to False.

Write path:

- `_resolve_nodata_attr` keeps reading `attrs['nodata']` as the canonical key. Falls back to `nodatavals` / `_FillValue` for foreign DataArrays.
- `to_geotiff` and `write_geotiff_gpu` consult `attrs['masked_nodata']` to decide whether to unmask (sentinel-restore) before writing, or write the array as-is.

Tests:

- New `test_nodata_semantics_split.py` exercising the matrix: source-has-sentinel × backend × {float-NaN-masked, int-sentinel-preserved}. Asserts both attrs are set correctly on every cell.
- Existing nodata tests (`test_nodata_attr_aliases_1582.py`, `test_nodata_int64_precision_1847.py`, `test_nodata_nan_int_1774.py`, etc.) stay green; many will need a small tweak to also check `masked_nodata`.

**Usage:**
[...]

PR #1995, opened 2026-05-16
geotiff: golden corpus phase 2.5 -- compression (#1930)

brendancol:
## Summary
Phase 2 PR 5 of #1930. Compression slice of the golden corpus.

Fixtures (six entries across the five major codecs):
- compression_none_uint8: baseline uncompressed
- compression_deflate_predictor2_uint16: deflate + predictor 2 (int)
- compression_deflate_predictor3_float32: deflate + predictor 3 (float)
- compression_lzw_predictor2_int16: lzw + predictor 2 on signed int
- compression_lerc_float32: lerc with max_z_error 0 (lossless setting)
- compression_jpeg_uint8_ycbcr: jpeg YCbCr, tolerance.lossy: true

Plan quote from #1930:
> PR 5 -- Compression -- none, deflate, lerc, lzw (with predictor 2/3 where relevant) + jpeg as a lossy cell that skips bit-eq (5 fixtures)

Deflate splits into two entries because predictor 3 is float-only and predictor 2 is integer-only (the validator enforces this), so six entries cover the five codecs.

## What is added
- 6 entries in `manifest.yaml`
- 6 generated `.tif` files under `golden_corpus/fixtures/`, all under 8 KB (largest is 4.4 KB)
- `test_golden_corpus_compression_1930.py` with 15 parametrised cases

## Lossy semantics
The jpeg fixture sets `tolerance.lossy: true` in the manifest. The smoke test calls `compare_to_oracle(..., lossy=True)` for that fixture and pins both halves of the contract added in #1991: strict mode rejects a perturbed candidate, lossy mode accepts it.

## Codec availability
LERC and JPEG depend on the GDAL build. If a `.tif` is missing from disk the test skips that case rather than failing. Both codecs were available on the local rasterio 1.4.4 build used to generate the fixtures.

## Not in this PR
- Backend wiring (numpy/dask/cupy/dask+cupy reads). That is Phase 3.
- Issue #1988 (split declared nodata). Per the plan this PR can land without it and be amended later.

## Test plan
- [x] ``pytest xrspatial/geotiff/tests/test_golden_corpus_compression_1930.py`` passes (15/15)
- [x] ``pytest xrspatial/geotiff/tests/test_golden_corpus_manifest_1930.py xrspatial/geotiff/tests/golden_corpus/test_oracle.py`` passes (33/33)
- [x] all ``.tif`` files under 8 KB


brendancol:
## Self-review

Walked through every changed file. Notes below; addressed in cccdb9b -> 23de1a4 where appropriate.

### Addressed
- `_fixture_path` docstring claimed the generator silently skips writes for unavailable codecs. It does not, it raises. Reworded to describe the actual reason a file might be absent (a contributor rebuilt the corpus locally without the codec).
- `test_jpeg_lossy_mode_required` comment claimed YCbCr-decoded checker values were 0 or 255 with wrap-around. Decoded values are well below 255 and `.clip` prevents wrap. Reworded.

### Verified
- All six fixtures open with rasterio at the expected dtype, shape, count, and CRS.
- LERC with `max_z_error: 0.0` round-trips bit-exact against the source `_make_pixels` array. The oracle accepts it in strict mode.
- JPEG round-trip is not bit-exact (as intended). The oracle accepts it with `lossy=True` and rejects a perturbed candidate with `lossy=False`. Both halves of the contract from #1991 are covered.
- File sizes: largest is 4462 bytes (none), smallest is 465 (deflate p2). All under the 8 KB cap.
- Existing golden-corpus tests (manifest + oracle) still pass.

### Codec availability
LERC and JPEG codec availability depends on the GDAL build. Both were present on the local rasterio 1.4.4 build used to generate the fixtures. If CI runs on a GDAL build without one of them, the smoke test skips the missing case rather than failing.

### Out of scope
- Backend wiring is Phase 3.
- #1988 (split declared nodata) can land later per the plan.

PR #1996, opened 2026-05-16
geotiff: golden corpus phase 2.4 - dtype variants (#1930)

brendancol:
## Summary

Phase 2 PR 4 of #1930 (golden corpus for geotiff parity).

From the plan:

> PR 4 -- Dtype variants -- int8/uint8/int16/uint16/int32/uint32/float32/float64 single-band (8 fixtures)

## What lands

Eight new manifest entries and eight `.tif` files, ids:

- `dtype_int8`
- `dtype_uint8`
- `dtype_int16`
- `dtype_uint16`
- `dtype_int32`
- `dtype_uint32`
- `dtype_float32`
- `dtype_float64`

All eight share the manifest defaults (stripped layout, blocksize 64, compression none, photometric minisblack, EPSG:4326, fixed transform) and override only `dtype`, `pixel_pattern`, `pixel_seed`, and pixel dimensions (20 x 20).

## Generator extension

`generate.py` grows a new `pixel_pattern` value, `noise_with_corners`. It generates the same seeded noise as `noise` and then forces the four corner pixels of every band to the dtype's min and max sentinels. The corner stamping runs for integer / unsigned dtypes only; float fixtures keep plain `noise`. A NaN-in-pixels variant for floats is deferred to Phase 2 PR 6 (nodata), since the schema currently has no slot for "plant a NaN at pixel (i, j)" and the plan flagged this.

## Size

All eight files stay under the 4 kB ceiling stated in the plan. The 20 x 20 pixel size is the smallest that keeps the float64 file (3566 bytes) under 4 kB without compression. Smallest is int8 at 766 bytes.

## Smoke test

`xrspatial/geotiff/tests/test_golden_corpus_dtype_variants_1930.py` pins, per fixture:

- the `.tif` exists on disk and is under 4 kB,
- `rasterio.open` reports the dtype the manifest declared,
- `_oracle.compare_to_oracle(path, rasterio-read DataArray)` accepts the file,
- corner pixels carry the dtype's min/max sentinels (integer fixtures only),
- the eight ids are in the manifest with the right dtypes.

23 tests, all green.

## Out of scope

- Backend wiring. Phase 3.
- Compression / predictor variants. PR 5.
- NaN-in-pixels for the float fixtures. PR 6 (nodata).

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_golden_corpus_dtype_variants_1930.py` (23 passed)
- [x] `pytest xrspatial/geotiff/tests/test_golden_corpus_manifest_1930.py xrspatial/geotiff/tests/golden_corpus/test_oracle.py` (33 passed, no regressions)
- [x] `python -m xrspatial.geotiff.tests.golden_corpus.generate` is byte-stable across reruns (sha256 verified)
- [x] flake8 clean on changed Python files


brendancol:
## Self-review

### Blockers
None.

### Suggestions
None blocking.

### Nits (fixed in follow-up commit a11494d)
- `pixel_seed` schema doc only mentioned `noise`; now lists `noise_with_corners` too. (`manifest.yaml`)
- The validator silently accepted 1x1 rasters with `noise_with_corners`. With width or height of 1, all four corner stamps land on the same pixel and the pattern collapses to a constant. Added a >=2x2 guard and a unit test. (`generate.py`, `test_golden_corpus_dtype_variants_1930.py`)

### What looks good
- Eight fixtures, one per dtype. They share defaults and override only `dtype`, `pixel_pattern`, `pixel_seed`, and 20x20 dimensions, so the only axis that varies is dtype.
- All files under the 4 kB ceiling: float64 at 3566 bytes, int8 at 766 bytes.
- Generator extension is small: one new enum value plus one corner-stamp block. Integer-only, gated on the new pattern name. The float path is untouched.
- The `noise_with_corners` pattern guarantees the dtype's min and max sentinels land in the four corner pixels, so backend readers exercise the edge values.
- NaN-in-pixels for float fixtures is deferred to PR 6 (nodata) as the plan suggested.
- Smoke test pins existence, size cap, rasterio dtype, oracle acceptance, corner sentinels (integer fixtures), and the manifest mapping.
- Byte-stable regeneration verified by sha256.

### Checklist
- [x] Generator changes are additive (existing patterns untouched)
- [x] Manifest validator accepts every new entry
- [x] All eight `.tif` files committed and under 4 kB
- [x] Oracle accepts a rasterio-built candidate DataArray for every dtype
- [x] Integer corner sentinels verified per fixture
- [x] flake8 clean on changed Python files
- [x] No backend wiring (Phase 3)

PR #1997, opened 2026-05-16
geotiff: golden corpus phase 2.7, overview / COG variants (#1930)

brendancol:
## Summary

Phase 2 PR 7 of the golden corpus plan in #1930. Adds three fixtures
on the overview-storage axis:

- `overview_internal_uint16`: internal overviews at [2, 4]
- `overview_external_ovr_uint16`: sidecar `.tif.ovr`
- `cog_internal_overview_uint16`: COG layout (tiled, IFD-ordered)

The generator now honours `external_overview: true` (schema-only in
Phase 1) and gains a `cog: true` flag that routes through the GDAL
COG driver. Each fixture stays under the 12 KB budget.

## Test plan

- [x] `python -m xrspatial.geotiff.tests.golden_corpus.generate` rebuilds the three fixtures deterministically (md5 stable across runs)
- [x] `pytest xrspatial/geotiff/tests/test_golden_corpus_overview_cog_1930.py` passes
- [x] `pytest xrspatial/geotiff/tests/test_golden_corpus_manifest_1930.py` still passes
- [x] `pytest xrspatial/geotiff/tests/golden_corpus/test_oracle.py` still passes


brendancol:
## Self-review

Walked through the diff against the Phase 2 PR 7 brief in #1930 and against the Phase 1 oracle. Findings and how each was handled (all in commit 1eb2a6b on this branch):

### Fixed in this PR

- `test_cog_fixture_ifd_order_base_first` was misnamed. The body checks GDAL's `LAYOUT=COG` marker, not literal IFD offsets. Renamed to `test_cog_fixture_carries_cog_layout_marker` and updated the docstring to be honest about what it actually probes.
- Two sidecar-absence checks used `Path.exists() is False`. Replaced with `not path.exists()` so the test does not depend on `exists()` returning a literal bool.
- Added a paragraph at the top of the test module spelling out the oracle gap: `_oracle.compare_to_oracle` reads only the base IFD, so overview-level comparison is out of scope until a follow-up PR extends the oracle.

### Verified, no change needed

- COG round-trip: CRS (`EPSG:4326`), transform, and overviews `[2, 4]` all survive the `rasterio.shutil.copy(driver="COG")` path. Tile size of 16 is a multiple of 16 (COG driver requirement).
- Determinism: regenerating the three fixtures twice produces identical md5s.
- Size budget: 9976 / 7443 / 8876 bytes for the three `.tif` files, plus 2540 bytes for the `.ovr` sidecar. All under the 12 KB cap.
- Generator dispatch: `cog: true` and `external_overview: true` are mutually exclusive (validator rejects the combination), matching the COG spec.
- Phase 1 oracle accepts all three fixtures at level 0 when the candidate is built from rasterio's own read.

### Out of scope

- Backend wiring is Phase 3.
- Overview-level oracle comparison is a future oracle extension and not part of this PR.
- Other Phase 2 PRs land their own fixtures via the same `manifest.yaml`; merge conflicts on the `fixtures:` list are expected.

PR #1998, opened 2026-05-16
geotiff: golden corpus phase 2.9 -- GDAL_METADATA + extra_tags (#1930)

brendancol:
## Summary

Phase 2 PR 9 of the plan in #1930:

> PR 9 -- GDAL_METADATA + extra_tags variants (2 fixtures)

Two new fixtures land in the geotiff golden corpus:

- `gdal_metadata_namespaced_uint16`: GDAL_METADATA in the default domain,
  IMAGE_STRUCTURE, SUBDATASETS, and a custom user domain. Covers the
  reader's cross-domain key handling.
- `extra_tags_uint16`: ImageDescription, Software, Artist, and a private
  numeric TIFF tag (65000) written as real IFD entries. Attribute drift
  on pass-through tags is a real cross-backend pain point and these
  exercise it.

## Generator changes

The manifest already had `gdal_metadata` and `extra_tags` schema keys, but
the wiring for `extra_tags` routed every entry through
`dst.update_tags(...)`. That lands the value in GDAL_METADATA XML rather
than the real TIFF tag code, and rasterio cannot emit private numeric
tags at all. The generator now writes the raster with rasterio first
(for the GeoTIFF tags) then rewrites in place with tifffile, keeping
the GeoTIFF tags and appending the requested extras as real IFD
entries. Validation for both fields was added so typos fail at manifest
load.

The canonical-attrs contract (#1984) is out of scope here; the oracle's
existing checks accept these fixtures as-is.

## Test plan

- [x] pytest xrspatial/geotiff/tests/test_golden_corpus_metadata_tags_1930.py
- [x] pytest xrspatial/geotiff/tests/test_golden_corpus_manifest_1930.py
- [x] pytest xrspatial/geotiff/tests/golden_corpus/test_oracle.py
- [x] Both .tif fixtures under 4 KB


brendancol:
## Self-review (resolved in 19a62ce)

Test-corpus tooling only; no production reader/writer code touched.

### Suggestions (addressed)

- [x] `generate.py` -- the tifffile post-pass kept only the six GeoTIFF tag codes, so a fixture combining `extra_tags` with `gdal_metadata` would have silently dropped GDAL_METADATA. The canonical `example_tiled_uint16_deflate_pred2` fixture in `main` does exactly that, so the original change would have broken it. Added 42112 GDAL_METADATA to `_GEOTIFF_TAG_CODES`. SampleFormat / ExtraSamples are intentionally left out because tifffile re-derives them from the pixel dtype.
- [x] `generate.py` -- `bool` subclasses `int`, so the `(str, int)` check on `extra_tags` keys let YAML `true`/`false` through. Reject `bool` explicitly first.

### Nits (addressed)

- [x] `generate.py` -- `_normalize_extra_tag_key` ran twice per item (sort key + loop body). Resolve once into `[(code, value)]` and sort by code.
- [x] `manifest.yaml` -- noted that IMAGE_STRUCTURE is a GDAL-managed domain and `COMPRESSION` / `INTERLEAVE` are placeholders that GDAL overwrites on write.
- [x] `test_golden_corpus_metadata_tags_1930.py` -- swapped `height, width = ...` to `width, height = ...` so it lines up with `np.arange(width)` / `np.arange(height)` below.

### Coverage added

- `test_validator_rejects_bool_extra_tags_key` -- pins the bool guard.
- `test_validator_rejects_non_string_gdal_metadata_domain` -- pins the domain-key type check.

### Verification

- `pytest xrspatial/geotiff/tests/golden_corpus xrspatial/geotiff/tests/test_golden_corpus_manifest_1930.py xrspatial/geotiff/tests/test_golden_corpus_metadata_tags_1930.py` -- 45 passed.
- Both fixtures are bit-identical to the previous commit (md5 hashes match).

PR #1999, opened 2026-05-16
geotiff: golden corpus phase 2.8, CRS variants (#1930)

brendancol:
## Summary
Phase 2 PR 8 of #1930. Three golden-corpus fixtures, one per CRS encoding the manifest allows. The citation-only fixture exposed a gap in `_crs_equal`; this PR closes it with a minimal PROJ-dict fallback.

## Fixtures
* `crs_epsg_3857`: `crs: { epsg: 3857 }`. Web Mercator. Straight EPSG path.
* `crs_wkt_utm10n`: `crs: { wkt: "..." }`. WKT for EPSG:32610 with AUTHORITY blocks stripped, so the bytes on disk are not byte-identical to `from_epsg(32610).to_wkt()`. PROJ still resolves it to EPSG:32610, which is what the oracle's EPSG-code fallback (PR #1991) was built for.
* `crs_citation_only`: `crs: { citation: "User-provided projection, no formal EPSG/WKT" }`. A name-only WKT: no AUTHORITY, no EPSG. libgeotiff mutates the WKT on round-trip (axis order, UNIT AUTHORITY) and neither side has an EPSG code, so structural `CRS.__eq__` and the EPSG fallback both fail.

## Oracle change
`_crs_equal` gets one extra branch: when both sides return `None` from `to_epsg()` and structural equality fails, compare `crs.to_dict()` (PROJ form). That dict is stable across the libgeotiff round-trip, so the citation-only fixture has a defined, documented answer instead of an unconditional false. This is the minimum change needed; bigger CRS semantics belong in #1984.

## Citation-only generator path
The manifest's `citation` form goes through the same rasterio writer as `epsg` / `wkt`. `_resolve_crs` builds a name-keyed WKT (`GEOGCS["<citation>",...]` with `DATUM["unknown",SPHEROID["unknown",6378137,0]]`) and hands it to rasterio. No `tifftools` / `gdal_edit` post-processing was needed; rasterio accepts the fake-WKT directly and libgeotiff serialises it as a citation-style GeoKey block.

## Tests
- `test_crs_epsg_3857_fixture_reports_epsg`: rasterio reads back `CRS.from_epsg(3857)`.
- `test_crs_wkt_utm10n_fixture_resolves_to_epsg_via_fallback`: rasterio resolves to EPSG:32610; oracle accepts a candidate with the bare EPSG int.
- `test_crs_citation_only_fixture_oracle_accepts_via_proj_dict`: oracle accepts the round-tripped WKT via PROJ-dict comparison.
- `test_crs_citation_only_fixture_rejects_unrelated_crs`: negative pin. EPSG:4326 (same proj family, different ellipsoid) is still rejected.

## Sizes
8x8 uint8 fixtures, all three `.tif` files under 600 bytes.

## Test plan
- [x] `pytest xrspatial/geotiff/tests/golden_corpus/test_oracle.py` passes (22 tests, 4 new).
- [x] Generator dry-run validates the full manifest.
- [x] Re-running the generator is byte-stable (md5sums identical across runs).


brendancol:
## Self-review of #1999

Ran a review pass on this PR. Two real bugs in the PROJ-dict fallback I added in the first commit, plus one missing-coverage gap. All addressed in f4be47b.

### Blockers (fixed)
- [x] `_oracle.py:179`: `CRS.to_dict()` returns `{}` for LOCAL_CS-style WKTs that PROJ has no canonical form for. The fallback as written would have treated any two such CRSes as equal, a silent false-positive in the oracle. Now short-circuits on empty dicts. `test_crs_equal_rejects_empty_proj_dict` pins it. The test uses two LOCAL_CS WKTs with different UNIT blocks so rasterio's own `CRS.__eq__` reports them unequal and the test actually reaches the fallback (otherwise the early-return on structural equality short-circuits first).

### Suggestions (fixed)
- [x] `test_oracle.py`: the WKT fixture was only exercised via the EPSG-int branch of `_candidate_crs`. Added `test_crs_wkt_utm10n_fixture_accepts_wkt_attr` to exercise the `attrs['crs_wkt']` branch on the same fixture so both paths share a verdict.

### Nits (documented, not fixed)
- [x] `_oracle.py` docstring: `CRS.to_dict()` drops the GEOGCS / PROJCS name, so two citation-only CRSes with the same shape but different names would compare equal under the fallback. The corpus has one citation fixture today so this is theoretical. Documented as a known limit with a pointer to `to_dict(projjson=True)` (preserves the name, needs an axis-order normaliser) for when it bites.

### What looks good
- Three fixtures are deterministic and tiny (under 600 bytes each).
- `_resolve_crs` did not need any post-processing for citation-only. rasterio accepts the fake-WKT directly and libgeotiff serialises it as a citation-style GeoKey block.
- 18 pre-existing oracle tests still pass.

### Checklist
- [x] Manifest schema unchanged
- [x] Generator deterministic across runs (md5sum verified)
- [x] All 24 oracle tests pass
- [x] No new failures in the wider \`xrspatial/geotiff/tests/\` suite (two pre-existing GPU failures are present on \`main\` and unrelated)

PR #2000, opened 2026-05-17
geotiff: document attrs tier classification in _attrs.py (#1984)

brendancol:
## Summary

PR 1 of 7 from the implementation plan on issue #1984.

Documents the three-tier attrs contract in the module docstring of `xrspatial/geotiff/_attrs.py`. Docstring-only; no behaviour change.

The three tiers are:

- **Canonical**: keys xrspatial owns and round-trips without loss (`crs`, `crs_wkt`, `transform`, `nodata`, `raster_type`, `extra_tags`, `gdal_metadata`, `gdal_metadata_xml`, `x_resolution`, `y_resolution`, `resolution_unit`, `_xrspatial_geotiff_contract`).
- **Compatibility alias**: read for ecosystem interop, never emitted by writers when the canonical key is present (`nodatavals` from rioxarray, `_FillValue` from CF).
- **Best-effort pass-through**: preserved when the writer can reconstruct from canonical state, otherwise dropped (GeoKey-derived fields, `image_description`, `extra_samples`, colormap variants).

The contract version key `_xrspatial_geotiff_contract` is referenced in the docstring; population of that key lands in a later PR in the series.

## Note on nodata

Issue #1988 (declared-vs-masked nodata split) is a prerequisite for tightening the `nodata` canonical semantics. The docstring records the *intended* semantics post-#1988: `nodata` is the declared file sentinel as stored in the GDAL_NODATA tag.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_attrs_parity_1548.py -x -q` (5 passed)

PR #2001, opened 2026-05-17
geotiff: add attrs contract documentation page (#1984)

brendancol:
PR 2 of 7 implementing issue #1984.

## Summary

Adds a user-guide page documenting the three-tier attrs contract emitted by the geotiff readers and consumed by the writers.

- Canonical keys (xrspatial-owned, byte-stable round-trip): ``crs``, ``crs_wkt``, ``transform``, ``nodata``, ``raster_type``, ``extra_tags``, ``gdal_metadata``, ``gdal_metadata_xml``, ``x_resolution``, ``y_resolution``, ``resolution_unit``, ``_xrspatial_geotiff_contract``.
- Compatibility aliases (read-only, never re-emitted): ``nodatavals`` (rioxarray), ``_FillValue`` (CF).
- Pass-through (GeoKey-derived; writer reconstructs what it can from CRS, drops the rest).

The page states per-tier round-trip invariants and explains the ``_xrspatial_geotiff_contract = 1`` version key for future revisions.

## Test plan

- [x] reST parses cleanly under docutils (Sphinx-specific ``:func:`` roles flagged as unknown, same pattern as the existing ``docs/source/reference/geotiff.rst``).
- [ ] Sphinx build in CI confirms the new page renders and shows up in the user-guide toctree (local ``sphinx-build`` fails on an unrelated ``conf.py`` import error from a cupy/dask/python 3.14 mismatch; CI runs in a clean env).
- [x] ``docs/source/user_guide/index.rst`` lists ``attrs_contract`` in the toctree so the page is reachable.

### DIFF
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -1,0 +1,230 @@
+..  _user_guide.attrs_contract:
+
+***********************
+GeoTIFF attrs contract
+***********************
+
+When :func:`xrspatial.geotiff.open_geotiff` returns a ``DataArray``, the
+``attrs`` mapping carries metadata recovered from the file's GeoTIFF
+tags and GeoKeys. xrspatial classifies those keys into three tiers.
+Each tier offers a different round-trip guarantee when the array is
+written back out with :func:`xrspatial.geotiff.to_geotiff`. Canonical
+keys are owned by xrspatial and survive a round-trip byte-for-byte.
+Compatibility aliases are recognised on read for interoperability with
+rioxarray and CF-style pipelines but are never re-emitted on write.
+Pass-through keys are surfaced verbatim from the file's GeoKey
+directory; whether they survive a round-trip depends on what the
+writer can reconstruct from the canonical CRS.
+
+.. contents:: On this page
+   :local:
+   :depth: 1
+
+
+Canonical keys
+==============
+
+xrspatial owns these keys. Every read path emits them when the source
+file carries the corresponding information, and every writer consumes
+them when serialising back to TIFF. A write followed by a read produces
+a byte-equivalent value for every canonical key that was set before the
+write.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 15 60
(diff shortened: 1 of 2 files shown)


brendancol:
## PR Review: geotiff: add attrs contract documentation page

- [ ] `docs/source/user_guide/attrs_contract.rst:54-58` — the `transform` description is wrong. The doc says `(origin_x, pixel_width, 0, origin_y, 0, pixel_height) affine transform tuple matching the GDAL ordering`. The emitted tuple is rasterio-style: `(pixel_width, 0.0, origin_x, 0.0, pixel_height, origin_y)` (see `xrspatial/geotiff/_coords.py:91-115`). Users who follow the doc to construct or interpret `attrs['transform']` will compute wrong georeferenced coordinates. Fix the tuple layout and replace "GDAL ordering" with "rasterio Affine ordering".

- [ ] `docs/source/user_guide/attrs_contract.rst:51` — `crs_wkt` is described as a "WKT2 string". Confirm the reader emits WKT2 (not WKT1) for every CRS source; if it stores whatever the file carries, drop the specific dialect.
- [ ] `docs/source/user_guide/attrs_contract.rst:185` — `cmap` is keyed off matplotlib availability and the doc says "Present only when matplotlib is importable." `colormap_rgba` is also conditional on the file's photometric interpretation (`Photometric == 3`); worth noting in the same row.

- [ ] `docs/source/user_guide/attrs_contract.rst:60-64` — the `nodata` row says "consumed by writers as the primary nodata source." The precedence chain on read is `nodata`, then `nodatavals`, then `_FillValue`, codified in `_resolve_nodata_attr`. Worth a one-liner.

### What looks good
- Three-tier table layout reads as a reference.
- Round-trip invariants section spells out each tier's guarantee clearly.
- Toctree entry added so the page is reachable from the user-guide index.

### Checklist
- [x] Three tiers documented
- [ ] `transform` tuple layout is wrong (see Blocker)
- [x] Toctree updated
- [x] Coordinated with PR #2000 module docstring

PR #2002, opened 2026-05-17
geotiff: add attrs compatibility-alias locking test (#1984)

brendancol:
PR 5 of 7 for issue #1984.

## Summary

Adds `xrspatial/geotiff/tests/test_attrs_contract_aliases_1984.py`, a single locking test that pins the *compatibility alias* tier of the attrs contract so future drift surfaces in one place. No production code changes; tests only.

The contract being pinned:

- Read paths must accept `attrs['nodatavals']` (rioxarray) and `attrs['_FillValue']` (CF) on a foreign DataArray, falling back to them when `attrs['nodata']` is absent.
- Write paths must not emit either alias when canonical `attrs['nodata']` is present; after a round-trip only `nodata` should appear.

Finer-grained cases continue to live in `test_nodata_attr_aliases_1582.py`. This file is intentionally contract-shaped and does not duplicate those cases.

## Tests added

- `test_read_uses_nodatavals_when_nodata_absent`: pins that a foreign DataArray carrying only the rioxarray tuple survives the writer.
- `test_read_uses_fill_value_when_nodata_absent`: same for CF `_FillValue`.
- `test_canonical_nodata_wins_over_aliases`: when all three keys are present and disagree, `attrs['nodata']` wins. Asserted at both the `_resolve_nodata_attr` chokepoint and the round-trip layer.
- `test_write_does_not_emit_aliases_when_canonical_present`: after writing a DataArray that has only `nodata` set, the reader returns attrs containing `nodata` and neither alias.
- `test_nan_in_nodatavals_is_skipped`: NaN entries in a rioxarray tuple are skipped, not treated as a concrete sentinel. Pure-NaN tuples resolve to `None`; mixed tuples return the first concrete numeric value.

## Bugs spotted

None. Reader paths in `_attrs._populate_attrs_from_geo_info` only set `attrs['nodata']`; neither `nodatavals` nor `_FillValue` is written into the output attrs after a round-trip, so all five assertions pass on the current code. No xfails needed.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_attrs_contract_aliases_1984.py -x -q` -- 5 passed.
- [x] `pytest xrspatial/geotiff/tests -x -q` sanity check. Two pre-existing failures observed on `main` (`test_predictor2_big_endian_gpu_1517.py` -- missing `read_to_array` attribute; `test_size_param_validation_gpu_vrt_1776.py::test_tile_size_positive_works` -- unrelated `_validation.py` ValueError). Both reproduce without this PR's changes, so unrelated.

Closes-step: 5 of 7 toward #1984.

### DIFF
--- a/xrspatial/geotiff/tests/test_attrs_contract_aliases_1984.py
+++ b/xrspatial/geotiff/tests/test_attrs_contract_aliases_1984.py
@@ -1,0 +1,159 @@
+"""Locking test for the compatibility-alias tier of the attrs contract
+(issue #1984, PR 5 of 7).
+
+xrspatial's canonical NoData attr is ``attrs['nodata']``. Two
+ecosystem aliases are tolerated on *read* paths so foreign DataArrays
+(rioxarray-style ``attrs['nodatavals']`` tuples, CF-style
+``attrs['_FillValue']``) survive a round-trip through ``to_geotiff``.
+
+This module pins both halves of that contract in one place so future
+drift surfaces here:
+
+* Read paths must accept the two aliases when ``attrs['nodata']`` is
+  absent, and must prefer ``nodata`` when present.
+* Write paths must not emit either alias into the resulting attrs
+  after a round-trip; only ``nodata`` should appear.
+
+A finer-grained set of regression tests for the same aliases lives in
+``test_nodata_attr_aliases_1582.py``. This file is intentionally
+contract-shaped: it does not duplicate those cases but locks the
+high-level rules.
+"""
+from __future__ import annotations
+
+import numpy as np
+import pytest
+import xarray as xr
+
+from xrspatial.geotiff import open_geotiff, to_geotiff
+from xrspatial.geotiff._attrs import _resolve_nodata_attr
+
+
+_SENTINEL = -9999.0
+
+
+def _da_float(arr, **attrs):
+    return xr.DataArray(
+        arr, dims=["y", "x"],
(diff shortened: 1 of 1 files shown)


brendancol:
## PR Review: geotiff: add attrs compatibility-alias locking test

None.

None.

- [ ] `xrspatial/geotiff/tests/test_attrs_contract_aliases_1984.py:38` — `_SENTINEL = -9999.0` is hardcoded at module scope. A one-line comment noting the value just needs to be float-castable and distinct from data would save the next reader a moment.
- [ ] `xrspatial/geotiff/tests/test_attrs_contract_aliases_1984.py:71-72` — fixed filename `alias_nodatavals.tif` per `tmp_path`. `tmp_path` already isolates per-test, so this is fine; just flagging that the project convention is unique tmp-file names.

### What looks good
- Five cases lock the contract: read-side fallback for both aliases, canonical-wins precedence, write-side suppression, and the NaN/None skip semantics in `nodatavals`.
- The precedence test asserts at both the `_resolve_nodata_attr` chokepoint and the round-trip layer, which makes a regression easier to localise.
- "Pure-NaN tuple resolves to None" pins a real corner of the resolver.

### Checklist
- [x] Tests for both aliases
- [x] Precedence (canonical wins) asserted
- [x] NaN/None skip semantics covered
- [x] Asserts on both resolver and round-trip
- [x] No production code touched

PR #2003, opened 2026-05-17
geotiff: stamp _xrspatial_geotiff_contract=1 on every read (#1984)

brendancol:
PR 3 of 7 on issue #1984.

Stamps ``attrs['_xrspatial_geotiff_contract'] = 1`` on every DataArray
returned by an xrspatial geotiff read path. Downstream code reads the
marker to learn which attrs-contract revision produced the array;
later PRs in the series will bump the value when the contract changes.

## Summary

- Added module-level constant ``_ATTRS_CONTRACT_VERSION = 1`` in
  ``xrspatial/geotiff/_attrs.py``.
- Stamped the value in ``_populate_attrs_from_geo_info``, which is the
  single chokepoint for the eager numpy, dask+numpy, GPU, dask+GPU,
  and COG/HTTP read paths.
- Stamped the value inline in ``_backends/vrt.py`` for the two VRT
  paths (eager and chunked) that build their attrs dict directly
  rather than going through the helper.

## Read paths covered

- ``open_geotiff`` (eager numpy)
- ``open_geotiff(chunks=...)`` / ``read_geotiff_dask`` (dask + numpy)
- ``open_geotiff(gpu=True)`` / ``read_geotiff_gpu`` (cupy / GPU)
- ``open_geotiff(gpu=True, chunks=...)`` (dask + cupy)
- COG / HTTP eager read (funnels through ``_read_to_array`` and the
  shared helper)
- ``read_vrt`` (VRT eager)
- ``read_vrt(chunks=...)`` (VRT chunked / dask)

## Test plan

- [x] ``pytest xrspatial/geotiff/tests/test_attrs_contract_version_1984.py -x -q``
- [x] ``pytest xrspatial/geotiff/tests -q`` (pre-existing GPU and fuzz
  failures on ``origin/main`` are unaffected; no test was previously
  asserting on the exact attrs dict so no other test needed updating)

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -52,6 +52,13 @@ _TIFF_ASCII = 2
 _TIFF_SHORT = 3
 
 
+# Contract version emitted on every read; bumped when the attrs contract
+# changes. Downstream code reads ``attrs['_xrspatial_geotiff_contract']``
+# to learn which attrs-contract revision produced the array. See issue
+# #1984 and ``docs/source/user_guide/attrs_contract.rst``.
+_ATTRS_CONTRACT_VERSION = 1
+
+
 # String identifiers (used in xrspatial attrs) -> TIFF ResolutionUnit tag ids.
 _RESOLUTION_UNIT_IDS = {'none': 1, 'inch': 2, 'centimeter': 3}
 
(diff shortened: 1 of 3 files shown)


brendancol:
## PR Review: geotiff: stamp _xrspatial_geotiff_contract=1 on every read

None.

- [ ] `xrspatial/geotiff/_attrs.py:113` — the stamp is the first line of `_populate_attrs_from_geo_info`. If a caller pre-populates `attrs['_xrspatial_geotiff_contract']`, it will be silently overwritten. No caller does today, but the unconditional overwrite is worth a one-line note in the helper's docstring.
- [ ] `xrspatial/geotiff/_backends/vrt.py:196, 566` — `_ATTRS_CONTRACT_VERSION` is referenced inline in two places. A short comment pointing at `_populate_attrs_from_geo_info` as the canonical site would help future maintainers know the helper is bypassed because VRT builds its attrs inline.

- [ ] `xrspatial/geotiff/tests/test_attrs_contract_version_1984.py:106-113` — `_gpu_available` swallows any cupy import error as "no GPU". If cupy is installed but broken (driver mismatch, etc.) the GPU tests skip silently rather than surfacing the cause. A `pytest.importorskip("cupy")` plus `cupy.cuda.is_available()` two-step would be more explicit.

### What looks good
- One stamp site in `_populate_attrs_from_geo_info` covers four backends (eager, dask, GPU, dask+GPU) plus the COG/HTTP path.
- VRT eager and chunked paths import the constant rather than hard-coding `1`, so a future bump stays in lockstep.
- Tests cover every read path the PR claims to cover, plus a pin on the literal value `1`.

### Checklist
- [x] All four primary backends covered
- [x] VRT eager + chunked covered
- [x] Constant imported (no duplicated literal)
- [x] Test pins the version value

PR #2004, opened 2026-05-17
geotiff: add attrs pass-through locking test (#1984)

brendancol:
PR 6 of 7 on issue #1984 (attrs-contract roadmap).

Adds `xrspatial/geotiff/tests/test_attrs_contract_passthrough_1984.py` to pin the current round-trip behaviour of every key in the pass-through tier of the attrs contract. Tests only; no production code changes.

## Background

The attrs contract has three tiers:

1. Canonical: the writer consumes the attr, so round-trip is guaranteed.
2. Pass-through: the writer does not consume the attr. The reader rebuilds the value from the GeoKey directory (or another TIFF tag) on read. Round-trip works only when the writer happens to emit a tag the reader can rebuild the attr from.
3. Ignored: the writer never touches the attr; it is dropped silently.

This PR locks the current state of every pass-through key so the next PR has a list of promotion candidates and a regression net around whatever it touches.

## Current state of every pass-through key

| Key | Source GeoKey / TIFF tag | Round-trip |
|---|---|---|
| `image_description` | tag 270 (ImageDescription) | reconstructible |
| `extra_samples` | tag 338 (ExtraSamples) | reconstructible |
| `colormap` | tag 320 (ColorMap, raw uint16) | reconstructible |
| `crs_name` | GTCitationGeoKey / ProjCitationGeoKey | dropped |
| `geog_citation` | GeogCitationGeoKey | dropped |
| `datum_code` | GeogGeodeticDatumGeoKey | dropped |
| `angular_units` | GeogAngularUnitsGeoKey | dropped |
| `linear_units` | ProjLinearUnitsGeoKey | dropped |
| `semi_major_axis` | GeogSemiMajorAxisGeoKey | dropped |
| `inv_flattening` | GeogInvFlatteningGeoKey | dropped |
| `projection_code` | ProjectionGeoKey | dropped |
| `vertical_crs` | VerticalCSTypeGeoKey | dropped |
| `vertical_citation` | VerticalCitationGeoKey | dropped |
| `vertical_units` | VerticalUnitsGeoKey | dropped |
| `colormap_rgba` | derived (reader gates on Photometric == 3) | dropped |
| `cmap` | derived (reader gates on Photometric == 3) | dropped |

The dropped rows reflect today's writer: `build_geo_tags` in `xrspatial/geotiff/_geotags.py` writes only the EPSG code into `GeographicType` / `ProjectedCSType` (plus `GTCitationGeoKey` on the user-defined WKT path). It does not emit the secondary GeoKeys the reader pulls those attrs from, so an EPSG round-trip cannot recover them. Per row, the call for review is whether this is intended behaviour or a gap to close in the next PR.

## Additional invariants pinned

* `test_passthrough_dropped_when_no_crs`: a file written without any CRS surfaces no pass-through attrs on read.
* `test_passthrough_does_not_promote_to_canonical`: setting a pass-through key on a DataArray without `crs` / `crs_wkt` does not cause the writer to synthesise a CRS.

## Test plan

```
pytest xrspatial/geotiff/tests/test_attrs_contract_passthrough_1984.py -x -q
pytest xrspatial/geotiff/tests -x -q
```

The new file's 18 cases pass locally. The broader geotiff suite keeps its pre-existing pass rate; two failures (`test_predictor2_big_endian_gpu_1517` and `test_size_param_validation_gpu_vrt_1776`) reproduce on clean `main` without this PR's diff.

### DIFF
--- a/xrspatial/geotiff/tests/test_attrs_contract_passthrough_1984.py
+++ b/xrspatial/geotiff/tests/test_attrs_contract_passthrough_1984.py
@@ -133,11 +133,29 @@ _PASSTHROUGH_CASES = [
     # derived ``colormap_rgba`` / ``cmap`` attrs are only emitted when
     # Photometric == 3 on read, which the writer does not set just
     # because attrs carries a colormap.
+    # The TIFF ColorMap tag (320) stores RGB triples as uint16 values in
+    # the 0-65535 range. Values below are written as-is and compared
+    # by-equality after the round-trip; if the writer ever rescales 8-bit
+    # input to 16-bit (or vice versa), update this fixture rather than
+    # the contract.
     ('colormap',          4326,  tuple([0] * 256 + [128] * 256 + [255] * 256),
                                                     'reconstructible'),
     ('colormap_rgba',     4326,  None,               'dropped'),
     ('cmap',              4326,  None,               'dropped'),
 ]
+
+
+def test_passthrough_cases_cover_all_keys():
+    """``_PASSTHROUGH_CASES`` and ``_ALL_PASSTHROUGH_KEYS`` carry the
+    same set in two forms. Pin them so a key added to one list and
+    forgotten on the other fails here rather than silently skipping
+    coverage in ``test_passthrough_dropped_when_no_crs``."""
+    case_keys = {c[0] for c in _PASSTHROUGH_CASES}
+    assert case_keys == set(_ALL_PASSTHROUGH_KEYS), (
+        f"_PASSTHROUGH_CASES and _ALL_PASSTHROUGH_KEYS diverge.\n"
+        f"  only in cases: {sorted(case_keys - set(_ALL_PASSTHROUGH_KEYS))}\n"
+        f"  only in keys : {sorted(set(_ALL_PASSTHROUGH_KEYS) - case_keys)}"
+    )
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


brendancol:
## PR Review: geotiff: add attrs pass-through locking test

None.

- [ ] `xrspatial/geotiff/tests/test_attrs_contract_passthrough_1984.py:142` — the `colormap` case writes a 256-entry colormap with `[0, 128, 255]`-style values. The TIFF ColorMap spec stores RGB triples as scaled uint16 (0-65535). If `to_geotiff` rescales internally, pin or document the convention; otherwise a future change to that conversion rule could break this test for non-substantive reasons.
- [ ] `xrspatial/geotiff/tests/test_attrs_contract_passthrough_1984.py:233-237` — `warnings.simplefilter('ignore')` hides every warning from the writer. If the writer ever raises a warning that signals data loss (e.g. dropping a key), this test would miss it. Narrow the filter to the specific `UserWarning` from the CRS WKT path, or assert the expected warning explicitly.

- [ ] `xrspatial/geotiff/tests/test_attrs_contract_passthrough_1984.py:62-79, 118-145` — `_PASSTHROUGH_CASES` and `_ALL_PASSTHROUGH_KEYS` carry the same set in two forms. A small `assert {c[0] for c in _PASSTHROUGH_CASES} == set(_ALL_PASSTHROUGH_KEYS)` at module load (or as its own test) would keep the two from drifting.

### What looks good
- Parametrisation captures all 16 pass-through keys with one of two outcomes, which makes the promotion-to-canonical decisions in the next PR mechanical.
- `test_passthrough_dropped_when_no_crs` and `test_passthrough_does_not_promote_to_canonical` cover the two cross-cutting invariants the per-key matrix does not.
- Failure messages quote the writer reconstruction logic, so a regression points at what changed.

### Checklist
- [x] All 16 pass-through keys parametrised
- [x] Cross-cutting invariants covered
- [x] No-CRS case covered
- [x] No production code touched

Issue #1986, opened 2026-05-15
geotiff: canonical round-trip invariants (byte- vs semantic-equivalent)


brendancol:
## Reason or Problem

The suite has round-trip tests scattered by incident: `test_metadata_round_trip_1484.py`, `test_no_georef_writer_round_trip_1949.py`, `test_int_coords_round_trip_hotfix_1962.py`, `test_miniswhite_writer_roundtrip_1836.py`. None state the round-trip invariant explicitly. After a write→read, is the array supposed to be byte-equivalent to the source, or only semantically equivalent? Where is metadata allowed to drift?

## Proposal

A canonical round-trip test module that enumerates every shape of input and states the invariant per case.

**Design:**

New file `xrspatial/geotiff/tests/test_round_trip_invariants.py` parametrized over the case list below. Two invariants:

- **byte-equivalent** — pixels and listed metadata fields are exactly equal after `read → write → read`.
- **semantic-equivalent** — pixels match exactly; metadata matches up to documented normalizations (e.g. EPSG int versus equivalent WKT, `PixelIsArea` default omitted on rewrite).

Cases to cover:

| Case | Invariant | Notes |
|------|-----------|-------|
| int single-band | byte | declared canonical |
| float single-band | byte for pixels, semantic for nodata | NaN-as-sentinel does not get a GDAL_NODATA tag |
| with declared nodata (int) | byte | sentinel preserved |
| with declared nodata (float, non-NaN) | byte | sentinel preserved |
| multiband planar | byte | sample-per-pixel pinned |
| multiband chunky | byte | |
| no-georef | semantic | transform attr omitted, GeoKeys absent on rewrite |
| PixelIsPoint | byte | raster_type round-trips |
| with overviews (COG) | byte for full-res, semantic for overview decimation when regenerated | |
| VRT over int sources | semantic | XML normalization expected |

Each case also gets a second invariant: write → read → write → read must converge to a fixed point.

**Usage:**
Single file maintainers grep when extending the writer. Each new case states what we promise.

**Value:**
The reviewer asked for "either byte-equivalent metadata where promised, or semantically equivalent metadata where byte stability is not promised." Today neither is documented, so silent metadata drift goes unnoticed until a downstream user complains.

## Stakeholders and Impacts

- Writers (`_writer.py`, `_writers/`, `_vrt.py`) — must hit the byte-equivalent target for the canonical cases. May surface drift that needs fixing in a follow-up.
[...]

PR #2007, opened 2026-05-17
geotiff: canonical round-trip invariants module (#1986)

brendancol:
## Summary

First slice of issue #1986. Adds `xrspatial/geotiff/tests/test_round_trip_invariants.py` -- one canonical module that enumerates the supported round-trip cases and pins the invariant per case.

The existing per-incident round-trip files (`test_metadata_round_trip_1484.py`, `test_int_coords_round_trip_hotfix_1962.py`, `test_no_georef_writer_round_trip_1949.py`, `test_miniswhite_writer_roundtrip_1836.py`) stay as regression markers for their bug numbers. This module is the canonical contract going forward.

## Cases covered

| Case | Invariant | Notes |
|------|-----------|-------|
| int single-band, no nodata | byte-equivalent | int16 / int32 / uint16 |
| float single-band, no nodata | byte-equivalent | float32 / float64; NaN-as-data preserved without synthetic GDAL_NODATA |
| int with declared nodata | documented drift | reader promotes int -> float64 + NaN; valid pixels byte-equal, sentinel survives in attrs |
| multiband chunky | byte-equivalent | 2 / 3 / 4 bands |
| no-georef | semantic | transform / crs attrs absent on read and through round-trip |
| PixelIsPoint | byte-equivalent | raster_type attr round-trips; default PixelIsArea leaves it absent |

Each case also asserts fixed-point convergence: a second `write -> read` cycle reproduces the first read exactly (pixels, dtype, dims, attrs key set).

## Deferred to follow-up PRs

Called out in the module docstring:

* Float with non-NaN declared nodata -- depends on the masked / declared nodata split from issue #1988.
* Planar multiband -- the writer emits chunky only.
* Overviews / COG.
* VRT round-trip with XML normalisation.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_round_trip_invariants.py -v` -- 14 passed
- [ ] CI green on the existing geotiff suite (the per-incident round-trip files are untouched)

Closes part of #1986.

### DIFF
--- a/xrspatial/geotiff/tests/test_round_trip_invariants.py
+++ b/xrspatial/geotiff/tests/test_round_trip_invariants.py
@@ -59,10 +59,17 @@ def _read_write_read(da: xr.DataArray, tmp_path, tag: str) -> xr.DataArray:
     return open_geotiff(path)
 
 
+# Canonical attrs whose values must lock across a write -> read cycle
+# whenever both reads have the key.
+_LOCKED_ATTRS = ('crs', 'transform', 'nodata', 'raster_type')
+
+
 def _assert_fixed_point(da1: xr.DataArray, da2: xr.DataArray) -> None:
     """``da1`` and ``da2`` come from two consecutive write -> read cycles.
 
-    They must agree exactly on pixels, dtype, dims, and attrs keys.
+    They must agree on pixels, dtype, dims, the attrs key set, and the
+    values of the canonical attrs listed in ``_LOCKED_ATTRS``. Other
+    attrs (best-effort pass-through) are only checked for presence.
     """
     assert da1.dtype == da2.dtype, (
         f"dtype drift: {da1.dtype} -> {da2.dtype}")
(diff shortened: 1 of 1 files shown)


brendancol:
## PR Review: geotiff: canonical round-trip invariants module (#1986)

Test-only PR. Steps for algorithm correctness, backend dispatch, CUDA register pressure, and README feature matrix do not apply -- no production code changed, no new public API. Review focuses on test quality, the asserted invariants, and coverage gaps the module itself flags.

### Blockers
None.

### Suggestions
- [x] `_assert_fixed_point` compares the attrs key set but not the values of canonical attrs. A writer regression that flips `crs` from `32610` to a WKT string, or perturbs `transform`, would pass the current check. Lock value parity for `crs`, `transform`, `nodata`, `raster_type` when present on both sides. Fixed in df2273a.
- [x] `TestIntSingleBandNoNodata` and `TestMultibandChunky` do not assert `'nodata' not in attrs`. The float-no-nodata test does. Add the symmetric assertion so the reader cannot quietly start emitting a synthetic sentinel. Fixed in df2273a.

### Nits
- [x] `arr[arr != -9999].ravel()` in `TestIntWithDeclaredNodata::test_int32_sentinel_promotes_and_masks` -- a boolean-masked array is already 1D, `.ravel()` is a no-op. Fixed in df2273a.
- [x] `TestNoGeorefSemantic::test_no_transform_survives_round_trip` does not assert dtype on the read DataArray. Added a `da1.dtype == np.float32` check so the array equality cannot accidentally pass via broadcasting. Fixed in df2273a.

### What looks good
- Module docstring states the two invariants explicitly and the deferred cases so a reader knows what is and is not promised.
- Fixed-point convergence (`_assert_fixed_point`) is checked per case, which is the missing piece in the per-incident files.
- The int-with-declared-nodata case is documented as drift rather than asserted-clean, which keeps the test honest while #1988's masked / declared split is pending.
- Filenames carry the `_1986` suffix and live under `tmp_path`, so parallel test runs and other worktrees will not collide.

### Checklist
- [x] Algorithm matches reference/paper -- N/A, test-only.
- [x] All implemented backends produce consistent results -- N/A, no backend dispatch changed; the reader/writer paths under test are numpy-only.
- [x] NaN handling is correct -- `np.isnan` mask, no `== np.nan`.
- [x] Edge cases are covered by tests -- multiple dtypes, NaN-as-data, no-georef, PixelIsPoint, default PixelIsArea absent.
- [x] Dask chunk boundaries handled correctly -- N/A, no dask in this slice.
- [x] Benchmark exists or is not needed -- not needed; contract tests.
- [x] Docstrings present and accurate -- module and class docstrings explain the invariant per case.

PR #2008, opened 2026-05-17
geotiff: add attrs['masked_nodata'] alongside attrs['nodata'] (#1988)

brendancol:
## Summary

Closes the foundational chunk of #1988: split the declared file sentinel from the masked-state signal in DataArray attrs on every geotiff read path.

- ``attrs['nodata']`` -- declared file sentinel as a scalar of the source dtype. Set whenever the source declared one. Independent of whether masking ran.
- ``attrs['masked_nodata']`` -- new boolean. ``True`` iff the reader replaced sentinel pixels with NaN in the in-memory array (final dtype is float). ``False`` iff the array still carries the literal integer sentinel.

The two signals were previously fused: ``attrs['nodata']`` doubled as "the file declared this sentinel" and "the reader already replaced sentinel pixels with NaN." Downstream code branching on the second meaning had to use the first as a proxy and lost the ability to tell a float-NaN array from an int-with-sentinel array.

## Changes

- New ``_set_nodata_attrs(attrs, nodata, *, array_dtype)`` in ``_attrs.py``. Sets both keys atomically from the final array dtype.
- Every read backend calls it from one site, after any masking and any user-requested dtype cast:
  - ``open_geotiff`` eager numpy (``__init__.py``)
  - ``read_geotiff_dask`` (``_backends/dask.py``) -- uses the declared graph dtype
  - ``read_geotiff_gpu`` -- three sites (eager tiled, eager stripped, chunked dask+gpu) in ``_backends/gpu.py``
  - ``read_vrt`` eager + chunked in ``_backends/vrt.py``
- Test matrix in ``test_nodata_semantics_split_1988.py``: source-has-sentinel x backend x {float-NaN-masked, int-sentinel-preserved}, plus direct unit coverage of ``_set_nodata_attrs``.

## What this does not change

This PR is additive. The writer still reads ``attrs['nodata']`` through ``_resolve_nodata_attr`` and ignores ``masked_nodata``; no existing nodata test changes its expectation. Migrating writers and downstream xrspatial code to branch on ``masked_nodata`` is the follow-up scope on #1988.

## Test plan

- [x] ``pytest xrspatial/geotiff/tests/test_nodata_semantics_split_1988.py`` (12 passed)
- [x] All existing nodata regression suites pass: ``test_nodata_nan_int_1774``, ``test_nodata_int64_precision_1847``, ``test_nodata_attr_aliases_1582``, ``test_miniswhite_nodata_1809``, ``test_vrt_band_nodata_1598``, ``test_overview_nodata_inheritance_1739``, ``test_vrt_int_source_float_dtype_1616``, ``test_vrt_multiband_int_nodata_1611`` (88 passed)
- [x] Full ``xrspatial/geotiff/`` suite (excluding the pre-existing fuzz failure on ``test_fuzz_hypothesis_1661.py``, which reproduces on ``main``): 952 passed

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -70,6 +70,7 @@ from ._attrs import (
     _extract_rich_tags,
     _populate_attrs_from_geo_info,
     _resolve_nodata_attr,
+    _set_nodata_attrs,
 )
 from ._backends._gpu_helpers import _is_gpu_data
 # Re-export only; called by xrspatial/geotiff/tests/test_nodata_*.py.
@@ -507,7 +508,6 @@ def open_geotiff(source: str | BinaryIO, *,
     # memory for a multi-MB raster.
     nodata = geo_info.nodata
     if nodata is not None:
-        attrs['nodata'] = nodata
         # When the reader applied MinIsWhite, the sentinel-equality mask
         # must compare against the inverted sentinel value (issue #1809).
         # ``read_to_array`` / ``_read_cog_http`` stash that value on
@@ -550,6 +550,11 @@ def open_geotiff(source: str | BinaryIO, *,
         target = np.dtype(dtype)
         _validate_dtype_cast(arr.dtype, target)
         arr = arr.astype(target)
+
+    # Set ``attrs['nodata']`` + ``attrs['masked_nodata']`` after the
+    # mask + optional dtype cast so ``masked_nodata`` reflects the
+    # final array dtype (issue #1988).
+    _set_nodata_attrs(attrs, nodata, array_dtype=arr.dtype)
 
     if arr.ndim == 3:
         dims = ['y', 'x', 'band']
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -152,15 +152,48 @@ def _extent_to_window(transform, file_height, file_width,
     return (row_start, col_start, row_stop, col_stop)
 
 
+def _set_nodata_attrs(attrs: dict, nodata, *, array_dtype) -> None:
+    """Set ``attrs['nodata']`` and ``attrs['masked_nodata']`` on a read.
+
+    Splits the two meanings previously fused into ``attrs['nodata']``
+    (issue #1988):
+
+    * ``attrs['nodata']`` -- declared file sentinel, as a scalar of the
+      source dtype. Set whenever the source declared one, regardless of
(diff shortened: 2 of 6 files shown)


brendancol:
## Self-review

### Blockers
None.

### Suggestions addressed in follow-up commit f7e6cb6

- `xrspatial/geotiff/tests/test_nodata_semantics_split_1988.py:29` had an unused `import xarray as xr`. Dropped.
- The original test file had no GPU or VRT coverage despite the PR touching five sites across `_backends/gpu.py` (three) and `_backends/vrt.py` (two). Added:
  - `TestVRTEager` (3 tests): float-typed VRT over int source with hit, int-typed VRT with no hit, VRT band with no `<NoDataValue>`.
  - `TestVRTChunked` (1 test): declared-float64 graph for in-range int sentinel.
  - `TestGPU` (3 tests, guarded by `_gpu_only`): `read_geotiff_gpu` on int source with hit, int source no-hit (stays int), and the dask + GPU path with in-range sentinel.

### Nits (left as-is)

- `xrspatial/geotiff/tests/test_nodata_semantics_split_1988.py:189` -- the hand-crafted uint16 TIFF builder duplicates `_build_uint16_tiff` from `test_nodata_nan_int_1774.py:54`. Local self-documenting duplication, leave for a future sweep.
- `xrspatial/geotiff/_attrs.py:108` -- `bool(np.dtype(array_dtype).kind == 'f')` -- the `bool(...)` wrapper is technically redundant (`==` on strings already yields a Python `bool`) but harmless and self-documenting.

### What looks good

- Single helper `_set_nodata_attrs` in `_attrs.py` keeps the contract in one place; every read site delegates to it.
- The masked_nodata signal is derived from the final array dtype, which is the right invariant to anchor on.
- Cross-backend divergence is intentional and documented: eager numpy stays int when the sentinel does not hit (`masked_nodata=False`), while dask declares float64 up front for any in-range sentinel (`masked_nodata=True`). Both reflect their actual array dtype, so the contract `masked_nodata == (dtype is float)` holds end-to-end.
- The change is additive. No existing nodata test had to be modified. The writer side is untouched and the next #1988 follow-up can migrate `_resolve_nodata_attr` / `to_geotiff` to consult `masked_nodata`.

### Final test counts

- New file: 19 tests, all passing locally (incl. 3 GPU tests on a cupy+CUDA env).
- Existing nodata regression suites: 88 passing.
- Full `xrspatial/geotiff/` suite (excluding pre-existing `test_fuzz_hypothesis_1661.py` failure that reproduces on main): 952 passing.

PR #2010, opened 2026-05-18
geotiff: deprecate vertical-CRS GeoKey attrs in DataArray.attrs (#1984)

brendancol:
## Summary

Part of the attrs-contract work in #1984. PR6 (#2009) locked which pass-through attrs round-trip; the vertical-CRS GeoKey attrs were pinned as "dropped on round-trip" because the writer in ``xrspatial.geotiff._geotags.build_geo_tags`` never emits the vertical GeoKey block (VerticalCSTypeGeoKey, VerticalCitationGeoKey, VerticalUnitsGeoKey).

This PR deprecates the three read-side emissions for one release cycle:

- ``attrs['vertical_crs']`` (VerticalCSTypeGeoKey)
- ``attrs['vertical_citation']`` (VerticalCitationGeoKey)
- ``attrs['vertical_units']`` (VerticalUnitsGeoKey)

The attrs still emit; only a ``DeprecationWarning`` is new. Removal is planned for a future release.

## Root cause

``build_geo_tags`` builds the GeoKey directory from canonical attrs (``crs`` / ``crs_wkt``) and does not look at the vertical attrs at all. A read that surfaces ``vertical_crs`` cannot be followed by a write that puts it back on disk, so the attr is effectively read-only and the contract treats it as broken.

## Conflict risk

Two sibling PRs run in parallel against the same docstring block in ``xrspatial/geotiff/_attrs.py`` and the same "Unreleased" CHANGELOG section:

- ``1984-pr7-deprecate-geographic-geokeys``
- ``1984-pr7-deprecate-projected-geokeys``

Whichever of the three merges first creates the "Deprecated (will be removed in a future release; see issue #1984)" section in the module docstring. The remaining two need a trivial rebase to append rather than recreate it.

## Test plan

- [x] ``pytest xrspatial/geotiff/tests/test_attrs_pr7_deprecate_vertical_1984.py`` (4 new tests)
- [x] ``pytest xrspatial/geotiff/tests/test_attrs_contract_*.py`` (47 existing contract tests)
- [x] ``pytest xrspatial/geotiff/tests/test_metadata_round_trip_1484.py xrspatial/geotiff/tests/test_round_trip_invariants.py``

Refs #1984. Builds on #2004 / #2009.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Unreleased
 
 #### Bug fixes and improvements
+- Deprecate read-side emission of vertical-CRS GeoKey attrs (vertical_crs, vertical_citation, vertical_units); the writer does not emit vertical GeoKeys so they do not round-trip. These attrs still emit for now but trigger a DeprecationWarning. Removal planned for a future release. (#1984)
 - Refresh the geotiff mmap cache when a file is replaced under the same path so re-reads after an atomic-rename overwrite no longer return stale bytes
 - Decode TIFF predictor=3 un-transpose by file byte order so big-endian floating-point TIFFs read back exactly
 - Default internal `_vrt.read_vrt` `missing_sources` to `'raise'` so an unreadable VRT source no longer produces a silent zero-fill hole on integer rasters; pass `missing_sources='warn'` to opt back into the previous lenient behaviour (#1843)
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -69,13 +69,20 @@ from canonical state, otherwise dropped on round-trip):
 - ``semi_major_axis``: GeogSemiMajorAxisGeoKey value.
 - ``inv_flattening``: GeogInvFlatteningGeoKey value.
 - ``projection_code``: ProjectedCSTypeGeoKey value.
-- ``vertical_crs``: VerticalCSTypeGeoKey value.
-- ``vertical_citation``: VerticalCitationGeoKey value.
-- ``vertical_units``: VerticalUnitsGeoKey value.
 - ``image_description``: TIFF ImageDescription tag.
 - ``extra_samples``: TIFF ExtraSamples tag.
 - ``colormap``, ``colormap_rgba``, ``cmap``: palette data attached to
   single-band paletted images.
+
+Deprecated (will be removed in a future release; see issue #1984):
+
+- ``vertical_crs``: VerticalCSTypeGeoKey value. The writer never emits
+  the vertical GeoKey block, so this attr cannot round-trip. It still
+  appears on read but triggers a ``DeprecationWarning``.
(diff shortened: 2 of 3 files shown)


brendancol:
## PR Review: geotiff: deprecate vertical-CRS GeoKey attrs

None.

- [ ] `xrspatial/geotiff/_attrs.py:266-289` — the three deprecation sites copy-paste the same nine-line `warnings.warn(...)` block with only the attr name changing. The sibling PR #2011 factors this into `_emit_deprecated_geographic_geokey` plus a `_DEPRECATED_GEOGRAPHIC_GEOKEY_ATTRS` tuple. If #2011 lands first, this PR should reuse the same helper (generalised, or a per-slice mirror). If this PR lands first, it should introduce the helper so #2011 / #2013 can adopt it on rebase.
- [ ] `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_vertical_1984.py:118-283` — `_build_tiff_with_vertical_geokeys` ships ~165 lines of ad-hoc TIFF/GeoKey serialisation. PR #2011 takes a smaller approach: drive `_populate_attrs_from_geo_info` directly with a synthetic `GeoInfo`, which exercises exactly the read-side surface PR 7 changes and runs much faster. Consider switching unless the full-stack `open_geotiff` path is materially different for the vertical attrs (it does not appear to be).
- [ ] `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_vertical_1984.py:136-141, 280-282` — `base = make_minimal_tiff(...)` is computed and never used; the `_ = base` line at the end papers over the unused-variable lint rather than removing the dead call. Drop the `make_minimal_tiff` call entirely, or take #2013's approach and patch the baseline in place.

- [ ] `xrspatial/geotiff/_attrs.py:266, 274, 282` — `stacklevel=2` from inside `_populate_attrs_from_geo_info` points at the backend caller, not at user code. Python silences `DeprecationWarning` for library callers by default, so this rarely surfaces to end users — but if the deprecation is later promoted to `FutureWarning` (visible to non-developers), the stacklevel will need to grow to walk past the backend frames. Worth a TODO at minimum.
- [ ] `xrspatial/geotiff/_attrs.py:71-83` — the three docstring entries each end with "Same deprecation reason as `vertical_crs`." A single shared sentence in the section header would read more cleanly than a repeated tail on every bullet.

### What looks good
- The deprecation rationale (writer never emits the vertical GeoKey block) is correct and traces back to `build_geo_tags` exactly as claimed.
- The new docstring section lands in the right place — under the pass-through tier, not the canonical tier.
- Tests assert both halves: warning fires AND attr value is still populated during the deprecation window.
- CHANGELOG entry is concrete and references #1984.

### Checklist
- [x] Warning fires on every emission path
- [x] Attr values still populated during deprecation window
- [x] Tests cover all three deprecated attrs
- [ ] Code structure aligned with sibling PRs (#2011 / #2013) — see Suggestion

PR #2011, opened 2026-05-18
geotiff: deprecate geographic-CRS GeoKey attrs in DataArray.attrs (#1984)

brendancol:
## Summary

PR 7 of the issue #1984 attrs-contract plan. The PR 6 locking test (merged as #2004) proved that six geographic-CRS GeoKey-derived attrs never survive a write -> read round-trip: the writer's `xrspatial/geotiff/_geotags.build_geo_tags` only emits the primary `GEOKEY_GEOGRAPHIC_TYPE` plus citation, never the secondary geographic GeoKeys these attrs are derived from.

Since the contract documented the attrs as "preserved when the writer can reconstruct from canonical state" but the writer demonstrably cannot, this PR deprecates them. The deprecated attrs are:

- `crs_name` (GTCitation / ProjCitation GeoKey; writer always overwrites from `crs_wkt`-derived data)
- `geog_citation` (GeogCitationGeoKey)
- `datum_code` (GeogGeodeticDatumGeoKey)
- `angular_units` (GeogAngularUnitsGeoKey)
- `semi_major_axis` (GeogSemiMajorAxisGeoKey)
- `inv_flattening` (GeogInvFlatteningGeoKey)

This PR is the warning-only phase: the attrs still land in `DataArray.attrs` on read, but each emission fires a `DeprecationWarning`. Removal is scheduled for a later release. The contract-version bump and actual removal will come in a follow-up.

## Changes

- `xrspatial/geotiff/_attrs.py`: route the six attrs through a new `_emit_deprecated_geographic_geokey` helper in `_populate_attrs_from_geo_info`; move the attrs out of the best-effort pass-through docstring section and into a new "Deprecated" section.
- `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_geographic_1984.py`: 15 tests covering per-attr warning emission, per-attr value still landing in attrs, no-warning-when-absent guard, and warning-text shape.
- `CHANGELOG.md`: deprecation note under Unreleased.

## Warning text

```
xrspatial.geotiff: attrs[<name>] is deprecated; the writer cannot reconstruct it from the canonical CRS so it will not round-trip. It will be removed in a future release. See issue #1984.
```

Fires as `DeprecationWarning` with `stacklevel=2`.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_attrs_pr7_deprecate_geographic_1984.py -v` (15 passed)
- [x] `pytest xrspatial/geotiff/tests/test_attrs_contract_passthrough_1984.py xrspatial/geotiff/tests/test_attrs_contract_canonical_1984.py -v` (35 passed, no regressions)
- [x] Full `pytest xrspatial/geotiff/tests/` sweep: 3541 passed, 8 pre-existing GPU failures unrelated to this change.

## Refs

- Issue #1984 (attrs contract plan)
- PR #2004 (PR 6 locking test that established these attrs do not round-trip)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Unreleased
 
 #### Bug fixes and improvements
+- Deprecate read-side emission of geographic-CRS GeoKey attrs (crs_name, geog_citation, datum_code, angular_units, semi_major_axis, inv_flattening); the writer cannot reconstruct them so they do not round-trip. These attrs still emit for now but trigger a DeprecationWarning. Removal planned for a future release. (#1984)
 - Deprecate read-side emission of projected-CRS GeoKey attrs (linear_units, projection_code); the writer cannot reconstruct them so they do not round-trip. These attrs still emit for now but trigger a DeprecationWarning. Removal planned for a future release. (#1984)
 - Deprecate read-side emission of vertical-CRS GeoKey attrs (vertical_crs, vertical_citation, vertical_units); the writer does not emit vertical GeoKeys so they do not round-trip. These attrs still emit for now but trigger a DeprecationWarning. Removal planned for a future release. (#1984)
 - Refresh the geotiff mmap cache when a file is replaced under the same path so re-reads after an atomic-rename overwrite no longer return stale bytes
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -223,6 +223,50 @@ present on the original file may be absent after write→read if the
 canonical CRS does not carry enough information to rebuild it.
 
 
+Deprecated GeoKey attrs (issue #1984)
+=====================================
+
+The following attrs are still populated on read for one release
+cycle but each emission fires a ``DeprecationWarning``. The writer's
+``build_geo_tags`` only emits the primary CRS GeoKey and citation for
+each axis (geographic, projected, vertical), so the secondary GeoKeys
+these attrs derive from are never written and the values do not
+survive a write→read round-trip. Migrate to ``crs`` / ``crs_wkt`` and
+derive any needed value with :mod:`pyproj`.
+
+Geographic-CRS GeoKey attrs: ``crs_name``, ``geog_citation``,
+``datum_code``, ``angular_units``, ``semi_major_axis``,
+``inv_flattening``.
+
+Projected-CRS GeoKey attrs: ``linear_units``, ``projection_code``.
+
+Vertical-CRS GeoKey attrs: ``vertical_crs``, ``vertical_citation``,
+``vertical_units``.
+
+Migration recipe::
+
+    from pyproj import CRS
+    crs = CRS.from_wkt(attrs['crs_wkt'])  # or CRS.from_epsg(attrs['crs'])
+
+    # Geographic
+    crs.name                                 # crs_name
+    crs.datum.to_epsg()                      # datum_code
+    crs.ellipsoid.semi_major_metre           # semi_major_axis
+    crs.ellipsoid.inverse_flattening         # inv_flattening
+    # geog_citation / angular_units: best-effort derive from
+    # ``crs`` / ``crs.axis_info``; the original GeoKey citation text
+    # is not generally recoverable.
+
+    # Projected
+    crs.coordinate_system.axis_list[0].unit_name   # linear_units
+    crs.to_epsg()                                  # projection_code
+
+    # Vertical
+    crs.sub_crs_list[-1].to_epsg()                 # vertical_crs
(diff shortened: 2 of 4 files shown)


brendancol:
## PR Review: geotiff: deprecate geographic-CRS GeoKey attrs

None.

- [ ] `xrspatial/geotiff/_attrs.py:75-85` — the new "Deprecated" docstring section says callers should "use `crs` / `crs_wkt` instead." For some of these attrs the migration is non-trivial: `inv_flattening` / `semi_major_axis` need `pyproj.CRS(...).ellipsoid.inverse_flattening` / `.semi_major_metre`, and `datum_code` needs `pyproj.CRS(...).datum.to_epsg()`. A short migration recipe in the docstring (or in `docs/source/user_guide/attrs_contract.rst`) would close the loop for users who get the warning and don't know where to look next.
- [ ] `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_geographic_1984.py:51-58` — `_DEPRECATED_CASES` has `(attr_name, field_name, value)` but `attr_name == field_name` in every row. Either drop the duplicate column or add a comment explaining when they would differ; today it just looks like a typo waiting to happen.
- [ ] `xrspatial/geotiff/_attrs.py:96` — `stacklevel=2` from inside `_emit_deprecated_geographic_geokey` points at `_populate_attrs_from_geo_info`, not at the user's `open_geotiff` call. Since `DeprecationWarning` is silenced by default for library code this only matters for tests today, but if the category later changes to `FutureWarning` (visible to end users), the stacklevel needs to walk past the backend frames to reach user code.

- [ ] `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_geographic_1984.py:74-84` — `test_deprecated_cases_cover_all_attrs` checks that the case list and the module-level tuple stay in sync. A `len(_DEPRECATED_CASES) == len(_DEPRECATED_GEOGRAPHIC_GEOKEY_ATTRS)` extra line would catch a duplicate-entry mistake that the set comparison silently absorbs.
- [ ] The helper `_emit_deprecated_geographic_geokey` is naming-scoped to "geographic", but sibling PRs #2010 / #2013 deprecate non-geographic attrs with identical mechanics. If this PR lands first, consider promoting the helper to a generic `_emit_deprecated_geokey_attr(attrs, name, value, *, category)` so the other two slices can reuse it without renaming.

### What looks good
- Helper + constant + warning-text factory is exactly the right shape. The other two PR 7 slices should adopt the same pattern.
- Driving `_populate_attrs_from_geo_info` directly with a synthetic `GeoInfo` is shorter, faster, and tighter than rebuilding TIFFs on disk. Sibling PRs that go full-stack should consider this approach.
- `test_warning_message_format` pins the canonical wording so a stylistic edit elsewhere does not silently break consumers that grep on the message.
- `test_no_warning_when_field_absent` is the right guard against an unconditional warning that would spam every read of every file.

### Checklist
- [x] Helper + constant kept in sync via test
- [x] Warning fires only when the deprecated field is populated
- [x] Attr values still emitted during deprecation window
- [x] CHANGELOG entry references #1984

brendancol:
Follow-up review fixes pushed. Rebased onto current `main` first so the docstring "Deprecated" section and the CHANGELOG "Unreleased" bullets merge cleanly with sibling PRs #2010 (vertical) and #2013 (projected); both are already on `main`.

Commits on top of the rebased PR commit (0d79dbf):

- 05b0cff S2 + N1: drop the redundant `field` column from `_DEPRECATED_CASES` (every row had `attr == field`), and add `test_deprecated_cases_has_no_duplicates` so a duplicate row can no longer be absorbed by the set-based coverage check.
- 0b5fbef S3: replace the fixed `stacklevel=2` with `_stacklevel_to_external_caller()`, which walks past every `xrspatial.geotiff*` frame (excluding the test subpackage) and returns the matching level. Different backends produce different chain depths (numpy: 3 frames, dask: 5 frames), so any single integer would only be right for one path. `test_warning_stacklevel_points_at_caller_file` pins the attribution to the user's file.
- caf0967 S1 + N2: promote the helper to `_emit_deprecated_geokey_attr(attrs, name, value, *, reason)`. The geographic, projected, and vertical tiers now all go through the generic helper and inherit the corrected stacklevel for free. Per-category reason clauses live in two module-level constants. `_emit_deprecated_geographic_geokey` is kept as a thin shim. The migration recipe added to the module docstring during the rebase is also mirrored into `docs/source/user_guide/attrs_contract.rst`.

Sibling PRs #2010 / #2013: both already on `main`, so no rebase needed there. This PR is the consolidation point for the helper.

Test counts: deprecation suite goes 15 -> 17 (added 1 length-assert, 1 stacklevel pin); projected + vertical + canonical + pass-through contract suites unchanged (42 tests, all green).

Nothing skipped.

PR #2013, opened 2026-05-18
geotiff: deprecate projected-CRS GeoKey attrs in DataArray.attrs (#1984)

brendancol:
## Summary

Part of the issue #1984 attrs contract rollout. PR 6 (#2004) locked which pass-through attrs survive a write + read round-trip. Two projected-CRS GeoKey attrs are dropped on round-trip because `xrspatial.geotiff._geotags.build_geo_tags` only emits the primary `GEOKEY_PROJECTED_CS_TYPE` (3072) and never the secondary projected GeoKeys:

- `linear_units` (ProjLinearUnitsGeoKey, 3076)
- `projection_code` (ProjectionGeoKey, 3074)

The reader rebuilds the attrs from those secondary GeoKeys on read, but the writer never emits them, so a write + read cycle silently drops both values.

This PR starts a one-release-cycle deprecation:

- `_populate_attrs_from_geo_info` still sets both attrs when the source file carries the underlying GeoKey, but each emission now fires a `DeprecationWarning` referencing issue #1984.
- The `_attrs.py` module docstring moves the two keys from the "Best-effort pass-through" list to a new "Deprecated" section.
- `CHANGELOG.md` has an Unreleased entry.

Removal of the emission is a separate follow-up after the deprecation cycle.

## Coordination / rebase risk

A sibling PR (`1984-pr7-deprecate-geographic-geokeys`) is adding the same "Deprecated" section to the same docstring block for the geographic-CRS attrs. Whichever branch lands second will hit a docstring conflict and need to merge its bullet list into the existing "Deprecated" section. The conflict is mechanical; behaviour on the two PRs is independent.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_attrs_pr7_deprecate_projected_1984.py` -- 3 passed
- [x] `pytest xrspatial/geotiff/tests/test_attrs_contract_*1984.py` -- 47 passed (canonical / aliases / version / pass-through tiers still hold)
- [x] `pytest xrspatial/geotiff/tests/` -- pre-existing GPU failures only; the 3 new tests pass and no new failures appear on CPU paths

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Unreleased
 
 #### Bug fixes and improvements
+- Deprecate read-side emission of projected-CRS GeoKey attrs (linear_units, projection_code); the writer cannot reconstruct them so they do not round-trip. These attrs still emit for now but trigger a DeprecationWarning. Removal planned for a future release. (#1984)
 - Refresh the geotiff mmap cache when a file is replaced under the same path so re-reads after an atomic-rename overwrite no longer return stale bytes
 - Decode TIFF predictor=3 un-transpose by file byte order so big-endian floating-point TIFFs read back exactly
 - Default internal `_vrt.read_vrt` `missing_sources` to `'raise'` so an unreadable VRT source no longer produces a silent zero-fill hole on integer rasters; pass `missing_sources='warn'` to opt back into the previous lenient behaviour (#1843)
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -65,10 +65,8 @@ from canonical state, otherwise dropped on round-trip):
 - ``geog_citation``: GeographicTypeGeoKey citation string.
 - ``datum_code``: GeogGeodeticDatumGeoKey value.
 - ``angular_units``: GeogAngularUnitsGeoKey value.
-- ``linear_units``: ProjLinearUnitsGeoKey value.
 - ``semi_major_axis``: GeogSemiMajorAxisGeoKey value.
 - ``inv_flattening``: GeogInvFlatteningGeoKey value.
-- ``projection_code``: ProjectedCSTypeGeoKey value.
 - ``vertical_crs``: VerticalCSTypeGeoKey value.
 - ``vertical_citation``: VerticalCitationGeoKey value.
 - ``vertical_units``: VerticalUnitsGeoKey value.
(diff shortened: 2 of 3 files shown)


brendancol:
## PR Review: geotiff: deprecate projected-CRS GeoKey attrs

None.

- [ ] `xrspatial/geotiff/_attrs.py:260-281` — the two deprecation sites copy-paste the same eight-line `warnings.warn(...)` block. Sibling PR #2011 factors the same pattern into `_emit_deprecated_geographic_geokey` plus a `_DEPRECATED_GEOGRAPHIC_GEOKEY_ATTRS` tuple. If #2011 lands first, this PR should reuse the helper (generalised or per-slice). If this PR lands first, introduce the helper here so the other slices can adopt it on rebase.
- [ ] `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_projected_1984.py:35-122` — `_make_projected_tiff_with_secondary_geokeys` rewrites the GeoKeyDirectory by appending fresh bytes at end-of-file and repointing the IFD entry. The old GeoKeyDirectory bytes are left in the file as dead data. Works, but if PR #2011's approach (drive `_populate_attrs_from_geo_info` directly via a synthetic `GeoInfo`) is acceptable for the deprecation surface, it is much smaller and side-steps the byte-rewriting tax entirely.
- [ ] `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_projected_1984.py:210-244` — the `test_linear_units_emits_deprecation_warning` and `test_projection_code_emits_deprecation_warning` tests duplicate ~30 lines apiece. The case-table-plus-parametrise style in PR #2011 covers the same surface in fewer lines and would scale better if a third projected attr is ever added.

- [ ] `xrspatial/geotiff/_attrs.py:260, 269` — `stacklevel=2` lands the warning at the backend caller, not user code. Same caveat as in #2010 / #2011: today `DeprecationWarning` is silenced by default so this rarely surfaces; if the category is later promoted to `FutureWarning`, the stacklevel will need to grow.
- [ ] `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_projected_1984.py:265-285` — the "still emits" test asserts on both `linear_units` and `projection_code` but only one TIFF carries both. If `_make_projected_tiff_with_secondary_geokeys(include_linear_units=True, include_projection_code=False)` ever produced different attrs than the combined fixture, this test would not catch it. Parametrising over each attr would be safer.

### What looks good
- Root-cause section in the PR body is precise (primary `GEOKEY_PROJECTED_CS_TYPE` emitted, secondary keys 3074 / 3076 not), which matches what `build_geo_tags` actually does.
- The "still emits" test guards against premature removal.
- CHANGELOG entry references #1984 and lists both deprecated attrs.

### Checklist
- [x] Warning fires on every emission path
- [x] Attr values still populated during deprecation window
- [ ] Code structure aligned with sibling PRs (#2010 / #2011) — see Suggestion

PR #2014, opened 2026-05-18
geotiff: deprecate matplotlib colormap variants in DataArray.attrs (#1984)

brendancol:
## Summary

Issue #1984 set up a three-tier attrs contract for the geotiff reader/writer pair. PR 6 (#2004) locked which pass-through attrs round-trip through ``open_geotiff -> to_geotiff -> open_geotiff``. Two keys in that tier do NOT round-trip: ``attrs['cmap']`` (matplotlib ``ListedColormap``) and ``attrs['colormap_rgba']`` (RGBA palette array). The reader only emits both when the source file declares ``Photometric=3`` (palette images), and the writer never selects ``Photometric=3`` from attrs alone, so the two keys disappear on any write-then-read cycle.

This PR deprecates read-side emission of both keys for one release cycle. Callers that want a ``ListedColormap`` should build one from the canonical ``attrs['colormap']`` (raw uint16 RGB triples from TIFF tag 320) instead; that key stays in the pass-through tier because the writer folds it into ``extra_tags`` via ``_merge_friendly_extra_tags`` and the reader rebuilds the attr from tag 320 on the next read.

Scope of the change:

- ``xrspatial/geotiff/_attrs.py``: wrap each of the two assignments in ``_populate_attrs_from_geo_info`` with ``warnings.warn(..., DeprecationWarning, stacklevel=2)``. Both attrs continue to be emitted on the returned DataArray during the deprecation window; only the warning is new. The matplotlib import is conditional, so the ``colormap_rgba`` warning fires on both branches (matplotlib present and ``ImportError`` fallback).
- Module docstring: split the ``colormap, colormap_rgba, cmap`` line. ``colormap`` stays in "Best-effort pass-through". ``colormap_rgba`` and ``cmap`` move to a new "Deprecated (will be removed in a future release; see issue #1984)" section.
- ``CHANGELOG.md``: one line under Unreleased.
- New test file ``xrspatial/geotiff/tests/test_attrs_pr7_deprecate_colormap_variants_1984.py`` (4 tests) pinning: warning fires for ``colormap_rgba`` (always) and ``cmap`` (matplotlib-only); both attrs still emit during the window; plain ``attrs['colormap']`` is unaffected.
- ``test_features.py::TestPalette`` and ``test_metadata_round_trip_1484.py::test_colormap_round_trip`` ignore the new ``DeprecationWarning`` locally to keep their reports clean.

Warning text (both keys; ``<name>`` is ``cmap`` or ``colormap_rgba``):

> xrspatial.geotiff: attrs['<name>'] is deprecated; the writer cannot set Photometric=3 so it does not round-trip. Construct a ListedColormap from attrs['colormap'] in caller code if needed. It will be removed in a future release. See issue #1984.

Related: #1984, #2004.

## Test plan

- [x] ``pytest xrspatial/geotiff/tests/test_attrs_pr7_deprecate_colormap_variants_1984.py`` -- 4 passed.
- [x] ``pytest xrspatial/geotiff/tests/test_attrs_contract_{passthrough,canonical,aliases,version}_1984.py`` -- 47 passed.
- [x] ``pytest xrspatial/geotiff/tests/test_metadata_round_trip_1484.py`` -- 10 passed, no new warning leaks.
- [x] ``pytest xrspatial/geotiff/tests/test_features.py::TestPalette`` -- 7 passed, no new warning leaks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Refresh the geotiff mmap cache when a file is replaced under the same path so re-reads after an atomic-rename overwrite no longer return stale bytes
 - Decode TIFF predictor=3 un-transpose by file byte order so big-endian floating-point TIFFs read back exactly
 - Default internal `_vrt.read_vrt` `missing_sources` to `'raise'` so an unreadable VRT source no longer produces a silent zero-fill hole on integer rasters; pass `missing_sources='warn'` to opt back into the previous lenient behaviour (#1843)
+- Deprecate read-side emission of matplotlib colormap-derived attrs (cmap, colormap_rgba) on palette TIFFs; the writer cannot set Photometric=3 so they do not round-trip. Construct ListedColormap from attrs['colormap'] in caller code. These attrs still emit for now but trigger a DeprecationWarning. Removal planned for a future release. (#1984)
 
 
 ### Version 0.9.9 - 2026-05-05
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -174,13 +174,38 @@ must not assume a specific pass-through key survives a round-trip.
    * - ``colormap``
      - tuple
      - Raw ``ColorMap`` TIFF tag (tag id 320) values.
+
+
+Deprecated keys
+===============
+
+These keys are still emitted on read for one release cycle, but each
+emission triggers a ``DeprecationWarning``. The writer cannot
+reconstruct them from canonical attrs, so they do not round-trip.
+Callers should migrate to the canonical alternative listed below
+before the warning-only window closes. See issue #1984.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 22 18 60
+
+   * - Key
+     - Type
+     - Definition and migration
    * - ``colormap_rgba``
      - array
-     - Decoded RGBA colormap, when one is present.
+     - Decoded RGBA colormap. Emitted on read when the file's
+       photometric interpretation is ``Photometric == 3`` (palette).
+       The writer cannot set ``Photometric == 3`` so the attr does
+       not round-trip. Reshape ``attrs['colormap']`` to
+       ``(n_colors, 3)`` and append an alpha channel in caller code
+       if needed.
    * - ``cmap``
      - ``matplotlib.colors.ListedColormap``
-     - Matplotlib colormap built from ``colormap_rgba``. Present only
-       when matplotlib is importable.
+     - Matplotlib colormap built from the palette. Same
+       ``Photometric == 3`` gate, same round-trip gap. Construct a
+       ``ListedColormap`` from ``attrs['colormap']`` in caller code
+       if needed.
 
 
 Round-trip invariants
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -74,8 +74,19 @@ from canonical state, otherwise dropped on round-trip):
 - ``vertical_units``: VerticalUnitsGeoKey value.
 - ``image_description``: TIFF ImageDescription tag.
 - ``extra_samples``: TIFF ExtraSamples tag.
-- ``colormap``, ``colormap_rgba``, ``cmap``: palette data attached to
-  single-band paletted images.
+- ``colormap``: raw uint16 RGB triples from the TIFF ColorMap tag (320),
+  attached to single-band paletted images.
+
+Deprecated (will be removed in a future release; see issue #1984):
+
+- ``colormap_rgba``: RGBA palette array, only emitted on read when the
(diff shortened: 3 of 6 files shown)


brendancol:
## PR Review: geotiff: deprecate matplotlib colormap variants in DataArray.attrs

Review covers both commits on the branch: the original colormap deprecation (`c526d47`) and the shared-helper refactor (`0b5797a`).

- [ ] `xrspatial/geotiff/_attrs.py:179` — the helper uses `warnings.warn(..., stacklevel=2)`, but the helper is itself called from `_populate_attrs_from_geo_info`. `stacklevel=2` from inside the helper aims the warning at the helper's caller (= `_populate_attrs_from_geo_info` itself). The original inline `warnings.warn(..., stacklevel=2)` calls in `c526d47` pointed one frame higher, at `_populate_attrs_from_geo_info`'s caller (the backend). Empirically: after the refactor the warning surfaces at `xrspatial/geotiff/_attrs.py:350` (the helper call site) rather than at the backend that called `_populate_attrs_from_geo_info`. Bump `stacklevel=2` → `stacklevel=3` in `_emit_deprecated_geokey_attr` to preserve the previous behaviour. The helper docstring's "aims the warning at the caller of `_populate_attrs_from_geo_info`" line is only true after that bump.

- [ ] `xrspatial/geotiff/_attrs.py:139` — helper name `_emit_deprecated_geokey_attr` doesn't quite fit `cmap` / `colormap_rgba`. Those attrs come from the TIFF ColorMap tag (320), not from a GeoKey. The helper docstring claims it's for "the issue #1984 PR 7 deprecation slices (vertical, geographic, projected, colormap variants)", and three of those four are GeoKey-derived. A more accurate name like `_emit_deprecated_attr` would cover the colormap case too without overpromising scope.
- [ ] `docs/source/user_guide/attrs_contract.rst:177-189` — the user-guide page still lists `colormap_rgba` and `cmap` under the pass-through tier with no deprecation note. Users reading the docs will not see the change until the page mirrors the new "Deprecated" section in the module docstring.
- [ ] `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_colormap_variants_1984.py` — there's no direct unit test for `_emit_deprecated_geokey_attr`. The colormap tests exercise it indirectly through the warning text match. A small test pinning (a) the warning text shape for a synthetic `(name, reason, migration)`, and (b) the `migration=None` branch, would close the loop now that the helper is shared across the PR 7 series.

- [ ] `xrspatial/geotiff/_attrs.py:340-344` — `_colormap_reason` and `_colormap_migration` are defined inside the `if geo_info.colormap is not None:` block. If sibling PRs (#2010 / #2013) adopt the helper, lifting these to module-level constants (parallel to PR #2011's `_DEPRECATED_GEOGRAPHIC_GEOKEY_ATTRS`) would let each slice keep its wording in a fixed location instead of re-deriving it at each call site.
- [ ] `xrspatial/geotiff/_attrs.py:88-89` — the migration hint for `colormap_rgba` says "Construct a `ListedColormap` from `attrs['colormap']`". Users who want an RGBA array (the actual type of `colormap_rgba`) will need to reshape `attrs['colormap']` to `(n_colors, 3)` and add an alpha channel. The current hint steers them at `ListedColormap` which they'd then have to convert further. A `colormap_rgba`-specific hint would be slightly more direct, though the current shared hint isn't wrong.
- [ ] `xrspatial/geotiff/tests/test_attrs_pr7_deprecate_colormap_variants_1984.py:124-227` — the in-test TIFF builder is the third copy of nearly-identical low-level TIFF/GeoKey serialisation across the PR 7 series (#2010 has one, #2013 has one). Worth promoting one of them to `conftest.py` once the series settles.

### What looks good
- The helper collapses 27 lines of inline `warnings.warn` copy-paste into one call per attr while keeping the emitted text byte-identical (verified by the existing `pytest.warns` matches still passing).
- Warning-filter regex in `test_features.py` and the `warnings.catch_warnings()` block in `test_metadata_round_trip_1484.py` cleanly scope the new deprecation noise to the tests that intentionally exercise the path.
- `test_plain_colormap_attr_not_deprecated` distinguishes "attrs['colormap']" from "attrs['colormap_rgba']" by matching on the closing bracket; that's a small detail that matters because the migration hint mentions `attrs['colormap']` inside the deprecated warning.
- CHANGELOG entry and module docstring "Deprecated" section both reference #1984 and explain the round-trip gap.

### Checklist
- [x] Deprecation rationale matches the writer's behaviour (no Photometric=3 path)
- [x] Plain `attrs['colormap']` (canonical) still emits without a warning
- [x] Both attrs still emit during the deprecation window
- [ ] `stacklevel` preserves pre-refactor warning location (see Blocker)
- [x] Existing palette tests are filter-scoped, not silently broken
- [ ] User-guide doc reflects the new "Deprecated" tier

Issue #1407, opened 2026-05-02
sky_view_factor: horizon angle ignores cell size, wrong by factor of cell_size


brendancol:
**Describe the bug**

`sky_view_factor` computes horizon angles using ray distance in cells rather than ground units. The kernels (`_svf_cpu` and `_svf_gpu` in `xrspatial/sky_view_factor.py`) compute:

```python
dz = elev - center
dist = float(r)            # r is the ray step index, in cells
elev_angle = atan2(dz, dist)
```

`dz` is in elevation units (typically meters); `dist` is in cells. For any DEM whose cell size is not 1.0 in the same unit as elevation, the ratio is wrong by a factor of `cell_size`.

**Concrete example**

A 30 m SRTM tile with cell size 30 m and a neighbor 30 m higher should give a horizon angle of `atan(30/30) = 45°`. The current code computes `atan(30/1) ≈ 88°`, which collapses SVF toward 0 even on gentle terrain.

For an aerial photogrammetry DEM at 0.5 m cell size with a neighbor 1 m above, the true angle is `atan(1/0.5) ≈ 63.4°`. The current code returns `atan(1/1) = 45°`.

**Expected behavior**

Read cell size from the DataArray (via `get_dataarray_resolution` from `xrspatial.utils`) and compute true ground distance per ray step:

```python
ground_dx = r * dx * cellsize_x
ground_dy = r * dy * cellsize_y
dist = sqrt(ground_dx**2 + ground_dy**2)
```

Allow callers to pass explicit `cellsize_x` / `cellsize_y` overrides for cases where the DataArray lacks usable coordinates.

**Why existing tests miss this**

`test_known_svf_simple_ramp` uses cell size 0.5 with a `z = x` ramp but only asserts `0.5 < svf < 1.0`. Both the buggy 45° angle and the correct 63.4° angle satisfy that band, so the bug slips through.

**Scope**

- `_svf_cpu` (CPU kernel)
- `_svf_gpu` (CUDA kernel)
- `_run_numpy`, `_run_cupy`, `_run_dask_numpy`, `_run_dask_cupy`
- Public `sky_view_factor()` API
- Tests covering non-unit cell size and anisotropic cell size

**Categories**: Cat 4 (missing projection / cell-size correction)
**Severity**: HIGH

PR #2024, opened 2026-05-18
rasterize: propagate like.attrs/coords and emit _FillValue (#2018)

### LINKED ISSUE #2018: rasterize: drop attrs, recompute coords, and don't emit _FillValue when like= is given
`rasterize()` in `xrspatial/rasterize.py` returns a DataArray with empty `attrs` and freshly-built coords on every code path, including when the caller passes a `like=` template raster.

The function signature already accepts `like: xr.DataArray` and pulls width, height, bounds, and dtype from it, but the final return at the bottom of `rasterize()` (around line 2227) builds:

```python
return xr.DataArray(
    out,
    name=name,
    dims=['y', 'x'],
    coords={'y': y_coords, 'x': x_coords},
)
```

Three categories of metadata get dropped that any chained spatial pipeline expects:

## 1. `attrs` are silently dropped (HIGH)

When the caller passes `like=elevation`, the resulting raster has empty `attrs`. So `slope(rasterize(gdf, like=elevation))` sees no `res`, no `crs`, no `transform`, no `nodatavals`. Downstream calls then recompute cell size from coords (which themselves may be slightly off, see point 2), and chained pipelines silently give different numbers than running them off `elevation` directly. Same class of bug as the sky_view_factor cellsize problem in #1407.

Fix: copy `like.attrs` onto the output when `like` is provided.

## 2. `coords` are recomputed by `linspace` instead of reused from `like` (MEDIUM)

When `like=` is given, `_extract_grid_from_like` reads `like.coords['x']` and `like.coords['y']`, then derives `(xmin, ymin, xmax, ymax)`. The output coords are then rebuilt from those bounds via `np.linspace(...)`. For pixel-exact alignment with `like`, the output should reuse `like.coords['x']` and `like.coords['y']` directly. Float64 round-trip is usually safe, but `linspace` from re-derived bounds is not bit-identical to the original coords and breaks `xr.align` between the rasterized output and `like`.

## 3. `_FillValue` / `nodatavals` not emitted (MEDIUM)

`rasterize(..., fill=-9999, dtype=np.int32)` produces an integer raster where `-9999` means nodata, but the result has no `_FillValue` attr and no `nodatavals` attr, so a downstream tool cannot identify nodata pixels.

Fix: emit `attrs['_FillValue'] = fill` (and `attrs['nodatavals'] = [fill]` for backwards compatibility with the rest of xrspatial) on every return path.

## Backends affected

All four backends (numpy, cupy, dask+numpy, dask+cupy) route through the same final `xr.DataArray(...)` constructor at line 2227, so the bug is uniform across backends. One fix site.

## Tests

Add to `xrspatial/tests/test_rasterize.py`:

- `test_like_propagates_attrs` -- `like` with `attrs={'crs': 'EPSG:32610', 'transform': (...)}` -> output has same attrs
- `test_like_preserves_coords` -- `like.coords['x'].values` is identical (bit-equal) to `result.coords['x'].values` when `like=` is used
- `test_fill_value_recorded` -- `rasterize(..., fill=-9999, dtype=np.int32)` produces `attrs['_FillValue'] == -9999`

The tests should be backend-parametrized over numpy / cupy / dask+numpy / dask+cupy so all four paths verify the same metadata behavior.

brendancol:
## Summary

Closes #2018.

`rasterize()` was returning a DataArray with empty `attrs` and freshly-built `linspace` coords on every code path, even when the caller passed `like=template`. The result silently broke chained spatial pipelines.

- `slope(rasterize(gdf, like=elevation))` saw no `crs`, no `res`, no `transform`, and recomputed cell size from coords that were not bit-identical to `elevation`. Same class of bug as the #1407 sky_view_factor cellsize issue.
- `rasterize(..., fill=-9999, dtype=np.int32)` produced an integer raster with no `_FillValue` and no `nodatavals` attr, so downstream tools could not identify nodata pixels.

All four backends (numpy, cupy, dask+numpy, dask+cupy) route through the same final `xr.DataArray(...)` constructor, so the fix is at one site.

## Changes

- `_extract_grid_from_like` now also returns the template's x/y coords and attrs.
- The final `xr.DataArray(...)` constructor reuses `like.coords` directly when the resolved output grid matches `like`, so the output is bit-identical to the template and `xr.align` keeps working.
- `like.attrs` are copied onto the output (defensive `dict(...)` so mutating the output doesn't leak into the template).
- `fill` is emitted as `attrs['_FillValue']` and `attrs['nodatavals'] = (fill,)` when `fill` is not NaN. The default NaN fill leaves attrs empty (no pollution).

## Test plan

- [x] Added `TestMetadataPropagation` to `xrspatial/tests/test_rasterize.py` with 9 cases:
  - `test_like_propagates_attrs` -- `crs`/`transform`/`res` from `like` land on output
  - `test_like_preserves_coords_bit_identical` -- output x/y coords are `np.array_equal` to `like.coords`
  - `test_like_attrs_isolated_from_template` -- mutating output attrs does not mutate `like.attrs`
  - `test_fill_value_recorded_when_not_nan` -- `fill=-9999` -> `_FillValue == -9999`, `nodatavals == (-9999,)`
  - `test_fill_value_omitted_for_nan` -- default fill leaves `_FillValue` and `nodatavals` absent
  - `test_no_like_no_attrs_pollution` -- no `like`, NaN fill -> `attrs == {}`
  - `test_like_attrs_propagated_dask` -- dask+numpy backend behaves identically
  - `test_like_attrs_propagated_cupy` -- cupy backend behaves identically (skipped if no CUDA)
  - `test_like_attrs_propagated_dask_cupy` -- dask+cupy backend behaves identically (skipped if no CUDA)
- [x] Full rasterize suite passes: 193 passed, 2 skipped.
- [x] All 9 new tests pass, including across the four backends (numpy, cupy, dask+numpy, dask+cupy).


brendancol:
## PR Review: rasterize: propagate like.attrs/coords and emit _FillValue (#2018)

Nice fix — the bug is real and the one-site fix at the final `xr.DataArray` constructor is the right call. I found one blocker around conflicting nodata attrs carried over from `like`, a few suggestions around NaN detection and coord coverage, plus minor nits.

- [ ] `xrspatial/rasterize.py:2255-2259` — when `fill` defaults to NaN and `like.attrs` already carries `_FillValue`, `nodatavals`, or `nodata` from a previous round-trip, those stale sentinels are copied straight through and then NOT overwritten (because `fill_is_nan` skips the emit branch). Reproduces with:
  ```python
  like = xr.DataArray(..., attrs={'_FillValue': -9999, 'nodatavals': (-9999,), 'crs': 'EPSG:4326'})
  out = rasterize([(box(2,2,8,8), 1.0)], like=like)  # fill defaults to NaN
  # out.attrs['_FillValue'] == -9999, but actual unwritten pixels are NaN.
  ```
  The geotiff writer's `_resolve_nodata_attr` (xrspatial/geotiff/_attrs.py:640) checks `nodata` first, then `nodatavals`, then `_FillValue`, so a `to_geotiff(out)` round-trip would tag pixels with the wrong sentinel. Strip any inherited `_FillValue` / `nodatavals` / `nodata` from `out_attrs` before deciding whether to emit a fresh pair based on `fill`. Same logic should apply on the non-NaN branch — overwrite `nodata` too, not just `_FillValue` and `nodatavals`, otherwise an inherited `nodata=-9999` outranks the freshly-set `_FillValue=0`.

- [ ] `xrspatial/rasterize.py:2256` — `isinstance(fill, float) and np.isnan(fill)` misses numpy scalar floats. `rasterize(..., fill=np.float32(np.nan))` currently emits `_FillValue=nan` even though semantically NaN. Use a try/except wrapper or `np.isnan(np.asarray(fill, dtype=float))` so any numeric NaN is treated the same.

- [ ] `xrspatial/rasterize.py:1948-1949,1965-1967` — `_extract_grid_from_like` only returns `like.coords['x']` and `like.coords['y']`. A `like` produced by rioxarray typically also carries non-dimension coords like `spatial_ref` (the CRS coord) and any band/time coords. These get dropped on the output, so `rio.write_crs()` followed by `rasterize` loses the CRS coord even though `attrs['crs']` survives. Propagate all of `like.coords` (or at minimum any scalar non-dim coords that don't conflict with the new y/x sizes).

- [ ] `xrspatial/rasterize.py:2236-2240` — the bit-identical reuse predicate checks `final_bounds == like_bounds` with float equality. After `_extract_grid_from_like` derives `like_bounds` via `float(np.min(x)) - px/2`, repeated float arithmetic can produce bounds that differ by 1 ULP from a caller-supplied identical-looking tuple. Either compare with `np.allclose(...)` or just key the reuse on `width == like_width and height == like_height and bounds is None` (caller didn't override bounds → we know `final_bounds is like_bounds`).

- [ ] `xrspatial/tests/test_rasterize.py:1681-1683` — `test_fill_value_recorded_when_not_nan` passes `fill=-9999` as a Python int and asserts `_FillValue == -9999`. Add an assertion that the stored sentinel's type is compatible with `final_dtype` (or at least round-trips correctly through `to_geotiff`). The geotiff path coerces to float, so this is mostly latent, but the dtype mismatch is worth pinning.

- [ ] `xrspatial/tests/test_rasterize.py` — no test for the round-trip described in the PR body: `rasterize → to_geotiff → read back → nodata matches`. That's the use case #2018 calls out. A single test that writes to a `tmp_path`, reads back, and asserts the file's nodata sentinel matches `fill` would lock in the user-visible fix.

- [ ] `xrspatial/tests/test_rasterize.py` — no test covering the inherited-nodata blocker above. Add one for `like.attrs={'_FillValue': -9999}` with `fill=np.nan` and assert `_FillValue` is NOT in `result.attrs` (or is NaN).

- [ ] `xrspatial/rasterize.py:1965-1967` — `like.coords['x'].copy()` and `like.coords['y'].copy()` defensively copy, but `xr.DataArray(..., coords={'x': ...})` already wraps the input in a new `IndexVariable`. The copy is redundant for the reuse branch (and it defeats the bit-identical claim slightly, since `.copy()` allocates new buffers — still equal under `array_equal`, but a wasted allocation on every `like=` call).

- [ ] `xrspatial/rasterize.py:2255` — `dict(like_attrs) if like_attrs is not None else {}` is fine, but `_extract_grid_from_like` already does `dict(like.attrs)` on line 1967. The second `dict(...)` is redundant since `like_attrs` is already a fresh dict. Pick one site for the defensive copy.

- [ ] `xrspatial/tests/test_rasterize.py:1695-1697` — the comment "When the caller passed `like`, the output should reuse its coords exactly (not rebuild via linspace)" reads slightly long. Trim to one line.

- [ ] `xrspatial/rasterize.py:2252-2253` — comment says "for xrspatial internal compatibility" but `nodatavals` is rioxarray's convention, not xrspatial's. The actual xrspatial-native key is `nodata`. Either emit `nodata` (matching `_resolve_nodata_attr`'s primary key) or reword the comment to say "for rioxarray compatibility".

### What looks good

- Single fix site for all four backends — they really do all route through the final `xr.DataArray(...)` constructor, so the parity claim holds.
- Defensive `dict(like.attrs)` so output mutation doesn't leak into the template — caught a real footgun.
- The `like.coords['x'].sizes['x'] == final_width` size gate handles the case where the caller overrides `width`/`height`/`bounds` while still passing `like` (you fall back to fresh linspace).
- 9 tests is good coverage for what's there; the cross-backend trio (numpy/dask/cupy/dask+cupy) catches the silent-drop case the metadata sweep flagged.

### Checklist

- [x] Algorithm matches reference/paper (N/A — metadata propagation, not an algorithm change)
- [x] NaN handling is correct (modulo the float32 scalar suggestion)
- [ ] Edge cases are covered by tests (inherited `_FillValue` from `like` missing)
- [x] Dask chunk boundaries handled correctly (same final constructor)
- [x] No premature materialization or unnecessary copies (modulo the redundant `.copy()` nit)
- [x] Benchmark exists or is not needed (metadata-only change)
- [x] README feature matrix updated (N/A — no new function)

PR #2028, opened 2026-05-18
reproject: validate lon/lat shape parity in geoid_height and itrf_transform (#2026)

### LINKED ISSUE #2026: reproject: geoid_height and itrf_transform silently accept mismatched lon/lat array shapes
`geoid_height` and `itrf_transform` in `xrspatial.reproject` don't check that `lon` and `lat` arrays are the same shape. When they aren't, the numba parallel kernel reads past the end of the shorter array and the function returns wrong numbers without raising.

## Reproducer

```python
import numpy as np
from xrspatial.reproject import geoid_height, itrf_transform

# lon has 3 elements, lat has 2. Should raise; instead returns 3 values,
# the third using out-of-bounds lat memory.
print(geoid_height(np.array([0.0, 90.0, 45.0]),
                   np.array([0.0, 45.0])))
# [17.14536524 -59.66452408 -43.61396885]

print(itrf_transform(np.array([-74.0, 0.0, 45.0]),
                     np.array([40.7, 0.0]),
                     src='ITRF2014', tgt='ITRF2020', epoch=2024.0))
# 3-element result; the third entry is garbage
```

## Why it matters

Both functions are public API. The numba kernels run under `@njit(parallel=True)`, so OOB reads are undefined behaviour rather than an IndexError. The user gets plausible-looking numbers and no warning. You only catch it if you happen to write a test for it.

## Proposed fix

Check `lon.shape == lat.shape` (and broadcastable `h.shape`) at the public-API boundary in `geoid_height` and `itrf_transform`, and raise `ValueError` with the mismatched shapes. The existing finite-value checks are the right place to add it.

Scope: `_vertical.py` (`geoid_height`) and `_itrf.py` (`itrf_transform`).

Tests: scalar+scalar still works, matching arrays don't regress, 1-D and 2-D mismatches raise, 0-D `h` still broadcasts against 1-D `lon`/`lat`.

Spotted during the reproject security sweep on 2026-05-17.

brendancol:
Closes #2026.

`geoid_height` and `itrf_transform` ravel `lon` and `lat` separately, then index them in a numba `@njit(parallel=True)` kernel using `lon.shape[0]`. If `lat` was shorter than `lon`, the kernel read past the end of `lat` and returned wrong numbers without raising. Both are public API, so the bug was invisible unless someone happened to test for it.

This adds an early shape check in both functions that rejects mismatched `lon` and `lat` with a `ValueError` naming both shapes. `itrf_transform` also pre-empts non-broadcastable `h` shapes so the error points at the public API instead of the broadcast call.

Spotted during the reproject security sweep on 2026-05-17 (Cat 4 OOB read plus Cat 6 missing input validation).

## Test plan

- [x] `pytest xrspatial/tests/test_reproject.py::TestGeoidShapeMismatch -v`, 6 new tests, all pass
- [x] `pytest xrspatial/tests/test_reproject.py::TestItrfShapeMismatch -v`, 6 new tests, all pass
- [x] `pytest xrspatial/tests/test_reproject.py -k "geoid or itrf"`, 57 tests, no regressions
- [ ] CI green


brendancol:
## PR Review: reproject: validate lon/lat shape parity in geoid_height and itrf_transform (#2026)

None.

- [ ] `xrspatial/reproject/_itrf.py:330-340`: the `np.broadcast_shapes(h_in.shape, lon_in.shape)` pre-check doesn't catch every shape that the downstream `np.broadcast_to(np.atleast_1d(h_in), lon_arr.shape)` later rejects, because `lon_arr` has been raveled to 1-D by then. Repro: `lon=np.array([-74,0,45]), lat=np.array([40.7,0,10]), h=np.array([[5.,6.,7.]])`. `broadcast_shapes((1,3),(3,)) == (1,3)` so the pre-check passes, then `broadcast_to((1,3) -> (3,))` raises numpy's `input operand has more dimensions than allowed by the axis remapping`. That's exactly the unhelpful error the PR is trying to pre-empt. Pre-existing behaviour, not introduced here, but the comment on lines 324-328 implies this case is handled. Either tighten the check (compare against `lon_arr.shape` after `atleast_1d` + ravel) or shorten the comment so it doesn't overpromise.
- [ ] `xrspatial/tests/test_reproject.py:1960-2023`: `TestItrfShapeMismatch` doesn't cover the case above (multi-dim `h` vs 1-D `lon` that broadcasts to a different ndim). One test there would pin down whichever behaviour you go with.

- [ ] `xrspatial/reproject/_itrf.py:334-340`: the `try / except ValueError: raise ValueError(...)` block could use `raise ValueError(...) from e` so the original numpy traceback is preserved. The replacement message is clearer than the numpy one, so this is purely a debugging aid.
- [ ] `xrspatial/reproject/_itrf.py:324-328`: the comment explaining the `h` check is longer than the check itself, and the "numpy's broadcast_to raises for true incompat shapes" line is a bit hand-wavy given the issue above.
- [ ] Follow-up from the sweep (LOW): `geoid_height_raster` in `xrspatial/reproject/_vertical.py:225-278` still doesn't check that `raster.coords[ydim]` and `raster.coords[xdim]` are finite. With `+inf` coords the `while lon_w < -180: lon_w += 360` loop at `_vertical.py:127` is an infinite loop. The sweep-state CSV calls this LOW and the PR scopes itself to the MEDIUM, so not a blocker for this PR. Reasonable as a separate issue.

### What looks good

- Shape check sits before `ravel()`, so multi-dim mismatches and the 0-d vs 1-D case are both rejected with the same wording.
- Error messages name both shapes, which makes the failure easy to act on.
- Scalar fast-path (`np.ndim(lon) == 0 and np.ndim(lat) == 0`) is preserved; `geoid_height(0.0, 0.0)` still returns a Python float.
- Empty arrays still work: `geoid_height(np.array([]), np.array([]))` returns shape `(0,)`.
- Cost is O(1) (one shape compare plus one `broadcast_shapes` call, both before any allocation).
- The exact reproducer from #2026 lands in `test_geoid_rejects_1d_mismatch` verbatim, so the prior wrong-numbers behaviour is locked out.
- 12 new tests pass locally; running the broader geoid/itrf slice shows no regressions.

### Checklist
- [x] Algorithm matches reference/issue: OOB read scenario is closed.
- [x] All implemented backends produce consistent results: N/A, NumPy-only public API.
- [x] NaN handling is correct: pre-existing `np.isfinite` checks still run after the new shape check.
- [x] Edge cases are covered by tests: scalar/scalar, matching 1-D, matching 2-D, mismatched 1-D, mismatched 2-D, scalar vs 1-D, `h` broadcast variations.
- [x] Dask chunk boundaries handled correctly: N/A.
- [x] No premature materialization or unnecessary copies: validation is O(1).
- [x] Benchmark exists or is not needed: not needed, validation-only.
- [x] README feature matrix updated: N/A, no new public function.
- [x] Docstrings present and accurate: no docstring change needed; the error wording is self-documenting.

Issue #2025, opened 2026-05-18
reproject vertical_crs path crashes on dask, cupy, and 3D multi-band inputs


brendancol:
The vertical datum shift inside `reproject(..., src_vertical_crs=..., tgt_vertical_crs=...)` only works for the 2-D in-memory numpy backend. Three valid input shapes raise on call:

## 1. Dask backend
`_apply_vertical_shift` calls `.copy()`, slices `result[r0:r1]`, then assigns via boolean fancy indexing (`strip_data[is_valid] += ...`). Dask arrays do not support `__setitem__` with a boolean mask, so the call raises before any output is produced.

Minimal repro:
```python
import numpy as np, xarray as xr, dask.array as da
from xrspatial.reproject import reproject

data = da.from_array(np.random.rand(64, 64) * 100, chunks=(32, 32))
raster = xr.DataArray(
    data, dims=['y','x'],
    coords={'y': np.linspace(40, 30, 64), 'x': np.linspace(-80, -70, 64)},
    attrs={'crs':'EPSG:4326'},
)
reproject(raster, 'EPSG:4326',
          src_vertical_crs='EGM96', tgt_vertical_crs='ellipsoidal')
# ValueError: operands could not be broadcast together with shapes (nan,) (4096,)
```

## 2. CuPy backend
`_apply_vertical_shift` calls Numba JIT functions (`_interp_geoid_2d`) that only accept numpy arrays. With a cupy-backed raster, `np.empty((n_rows, out_w))` plus the JIT call sees a numpy strip output but the result array is cupy, and the boolean-index update raises.

Repro yields `TypeError: Unsupported type <class 'numpy.ndarray'>` when `result[is_valid] += ...` mixes cupy result with numpy update.

## 3. 3D multi-band
For a `(y, x, band)` source, the reprojected result is `(H, W, B)` but `is_valid` has shape `(n_rows, W)` and `N_strip` has shape `(n_rows, W)`. Broadcasting fails:
```
ValueError: operands could not be broadcast together with shapes (64,64,3) (64,64)
```

## Why it matters

The public `reproject()` signature documents `src_vertical_crs` and `tgt_vertical_crs` as supported alongside dask, cupy, and multi-band inputs. There are no warnings or guards; users get a hard crash on three of the four backends.

## Suggested fix

Rework `_apply_vertical_shift` to:
1. Bail out and apply the shift via `dask.array.map_blocks` for dask inputs.
[...]

PR #2029, opened 2026-05-18
reproject: fix vertical_crs shift on dask, cupy, and 3-D inputs (#2025)

brendancol:
## Summary
- Fixes `reproject(..., src_vertical_crs=..., tgt_vertical_crs=...)` on dask, cupy, and 3-D multi-band inputs. Each used to crash before producing output.
- Routes dask through `map_blocks` so the output stays lazy and the per-chunk shift is computed from each block's row/column slab.
- Round-trips cupy arrays via host (the CPU JIT geoid lookup cannot accept cupy buffers) and returns a cupy result.
- Loops over band slices for 3-D `(y, x, band)` results so the same N(y, x) correction applies to every band.
- Recorded on the deep-sweep accuracy state CSV (issue #2025).

## Test plan
- [x] `python -m pytest xrspatial/tests/test_reproject.py` — 245 passed locally.
- [x] New `TestVerticalShift::test_dask_backend_matches_numpy` confirms dask path matches numpy bit-for-bit and keeps the result lazy.
- [x] New `TestVerticalShift::test_multiband_3d_applies_shift_per_band` confirms each band sees the same geoid undulation correction and band 0 equals the 2-D reference.
- [x] New `TestVerticalShift::test_cupy_backend_matches_numpy` confirms the vertical-shift increment matches numpy on cupy backend (gated on `cupy` being importable).


brendancol:
## PR Review #2029: reproject vertical_crs shift on dask, cupy, 3-D

### Blockers
None. The three HIGH crashes are fixed, the numpy 2-D path is algebraically unchanged, and the dask path produces bit-identical output to numpy on the cases I ran (EGM96->EGM2008 round-trip, and geographic->UTM with `need_inverse`).

### Suggestions
- `xrspatial/reproject/__init__.py:874-886`: the cupy branch is gated on `isinstance(data, cp.ndarray)` and is unreachable when `data` is a dask array whose chunks are cupy. The dask branch at line 865 fires first, so `_apply_vertical_shift_dask` would hand cupy chunks to `_apply_vertical_shift_numpy`, where the strip pipeline mixes `np.zeros`, `np.tile`, etc. with cupy. Today this is dead code because `_reproject_dask(is_cupy=True)` collapses to numpy upstream. Either detect cupy inside `_block` and host-bounce per chunk, or leave a comment saying dask-of-cupy is intentionally unreached.
- `xrspatial/reproject/__init__.py:1005`: `da.map_blocks(... meta=np.array((), dtype=data.dtype))` hardcodes numpy meta. Same caveat: if the input dask array ever carries cupy meta, the output meta should match. Worth a comment or an assert on entry.
- `xrspatial/tests/test_reproject.py:2226-2247`: the cupy test compares the geoid increment (`shifted - base`) rather than full output. That works around the identity-CRS cupy zero-output bug, but if `_apply_vertical_shift` later regresses to a no-op for cupy, both `base_cu` and `shifted_cu` would be zero, `delta_cu` would be zero, and the assertion would still pass at `delta_cu == 0` for any row where `delta_np` is also zero. Add `assert np.any(np.abs(delta_cu) > 0)` to pin that the shift actually fired.
- `xrspatial/reproject/__init__.py:990`: `_block` falls back to passing the full `y_arr`/`x_arr` when `block_info is None`. That branch is only hit by synthetic calls; in practice `map_blocks` always provides `block_info`. Either drop the `None` branch or note in the docstring that it exists for unit-testing the closure directly.

### Nits
- `xrspatial/reproject/__init__.py:856-863` and `:874-878`: the dask check uses alias `_da`, the cupy uses `cp`. The leading-underscore alias for `dask.array` is unusual here; compare line 988 in `_apply_vertical_shift_dask`.
- `xrspatial/reproject/__init__.py:921`: `data.shape[:2]` assumes ndim >= 2. For 1-D or 0-D input it would silently return a wrong shape. Not reachable through `reproject()`, but `_apply_vertical_shift_numpy` is now shared by both backends, so an `assert data.ndim in (2, 3)` would be cheap.
- `xrspatial/reproject/__init__.py:927`: the "Process in row strips to bound memory" comment is stale for the dask path. Per chunk the strip loop runs once because chunks are typically smaller than 128 rows.
- `_projections.py:2337` (`np.mean(flat_lon[:100])` datum-grid selection) is correctly deferred. Not in scope here.

### What looks good
- The numpy leaf is shared by both wrappers, so the dask `_block` can't diverge from the numpy path.
- The new `N_total` accumulator replaces the two-pass `+=` without changing the result. Verified bit-for-bit on EGM96->EGM2008 (`max diff: 0.0`).
- 3-D loop applies the same N(y, x) to every band, since N depends only on horizontal position. The test at `test_reproject.py:2168-2188` pins inter-band differences and band-0 vs the 2-D reference.
- Dask output stays lazy. `isinstance(out.data, da.Array)` holds after `reproject(..., src_vertical_crs=..., tgt_vertical_crs=...)`.
- Each new test exercises a path that crashed before.
- Three expected files touched, no stray edits.
- Full reproject suite: 254 tests passed locally.

### Checklist
- [x] Signs/combinations preserved from the original
- [x] Dask matches numpy bit-for-bit on geographic and projected target
- [x] NaN/nodata handling preserved per band, finite-coord guard still applied
- [x] 2-geoid round-trip exercised; multi-band reference vs 2-D pinned
- [x] Per-block y/x slabs use global `array-location`
- [x] No `.compute()` or `.values` in the dask path
- [n/a] Benchmark: geoid shift is a small fraction of reproject cost
- [n/a] README feature matrix: no new public function
- [x] Docstrings updated with the backend-handling note

PR #2034, opened 2026-05-18
reproject: support 3-D rasters on dask, cupy, and dask+cupy (#2027)

### LINKED ISSUE #2027: reproject() and merge() drop band dim or crash on 3D dask/cupy inputs
`reproject()` and `merge()` accept 3D rasters via `_validate_raster(ndim=(2, 3))`, but only the eager-numpy path actually handles them. The dask, cupy, and dask+cupy paths either drop the band dim from the advertised output metadata or crash with confusing errors. This is separate from #2025, which only covers the `_apply_vertical_shift` branch.

## 1. `reproject()` 3D dask+numpy advertises wrong dims

```python

arr = xr.DataArray(
    da.from_array(np.random.rand(32, 32, 3).astype(np.float32), chunks=(16, 16, 3)),
    dims=['y', 'x', 'band'],
    coords={'y': np.linspace(55, 45, 32), 'x': np.linspace(-5, 5, 32)},
)
r = reproject(arr, 'EPSG:32633')
print(r.shape, r.dims)   # (39, 28) ('y', 'x')   <-- band silently dropped
r.compute()              # ValueError: replacement data has shape (39, 28, 3); Variable has shape (39, 28)
```

The lazy `xr.DataArray` claims to be 2D `(y, x)` but the underlying dask graph computes a 3D result. Any caller that inspects `result.dims` or `result.shape` sees consistent-looking but wrong information, then crashes on compute.

Root cause: `_reproject_dask` builds a 2D template from `out_shape` (which is always 2D, coming from `_compute_output_grid`), but `_reproject_chunk_numpy` returns 3D `(H, W, n_bands)` for 3D source data.

## 2. `reproject()` 3D cupy raises a cryptic CUDA TypeError

```python

data = cp.asarray(np.random.rand(32, 32, 3).astype(np.float32))
arr = xr.DataArray(
    data, dims=['y', 'x', 'band'],
    coords={'y': np.linspace(55, 45, 32), 'x': np.linspace(-5, 5, 32)},
)
reproject(arr, 'EPSG:32633')
# TypeError: Signature mismatch: 2 argument types given, but function takes 3 arguments
```

Root cause: `_reproject_chunk_cupy` has no `ndim == 3` branch (unlike `_reproject_chunk_numpy`). It hands the 3D source window straight to `_resample_cupy_native`, whose CUDA kernels are written for 2D arrays.

The dask+cupy path fails the same way.

## 3. `merge()` 3D inputs fail at DataArray construction

```python
from xrspatial.reproject import merge

a = xr.DataArray(np.random.rand(8, 8, 3), dims=['y','x','band'],
                 coords={'y': np.linspace(5,-5,8), 'x': np.linspace(-5,0,8), 'band': [1,2,3]},
                 attrs={'crs':'EPSG:4326'})
b = xr.DataArray(np.random.rand(8, 8, 3), dims=['y','x','band'],
                 coords={'y': np.linspace(5,-5,8), 'x': np.linspace(0,5,8), 'band': [1,2,3]},
                 attrs={'crs':'EPSG:4326'})
merge([a, b], resolution=1.0)
# ValueError: different number of dimensions on data and dims: 3 vs 2
```

`merge()` accepts 3D via `_validate_raster(ndim=(2, 3))` but the final `xr.DataArray(..., dims=[ydim, xdim])` is hardcoded 2D.

## Why it matters

The docstring promises `reproject()` works on numpy, cupy, dask+numpy, and dask+cupy. For 3D rasters, only the eager-numpy backend works. The dask path is worse than a clean failure because it advertises 2D metadata for what is actually 3D data. The input dim count `(y, x, band)` does not survive to the output: that is a Cat 3 (dim count drift) bug for the dask path and a Cat 5 (backend-inconsistent metadata) bug for the cupy paths.

## Suggested fix sketch

- `_reproject_dask`: detect 3D sources and build a 3D template that keeps the band axis in the source position.
- `_reproject_chunk_cupy`: add a 3D branch that loops over bands, matching what `_reproject_chunk_numpy` already does.
- `merge()`: either honour the band dim or tighten the validator to `ndim=(2,)` so the API surface stops promising what the implementation can't deliver.
- Add cross-backend parity tests in `xrspatial/tests/test_reproject.py` covering 3D inputs on each of the four backends.

brendancol:
Closes #2027.

## Summary

`reproject()` accepted 3-D `(y, x, band)` inputs via `_validate_raster(ndim=(2, 3))`, but only the eager numpy path actually handled them. The dask path advertised a 2-D template for chunks that returned 3-D arrays (lazy shape disagreed with `.compute()` output), the cupy path tripped a CUDA signature mismatch, and `merge()` crashed at output DataArray construction.

- `_reproject_chunk_cupy`: per-band loop matching `_reproject_chunk_numpy`
- `_reproject_dask` / `_reproject_block_adapter`: 3-D template with `(out_h, out_w, n_bands)` shape so lazy metadata matches actual chunk output
- `_reproject_dask_cupy`: route 3-D sources through `map_blocks` (fast path is 2-D-only)
- Empty-chunk early returns in both chunk functions now carry the band axis
- `merge()`: tighten validator to `ndim=(2,)` so callers see a clean error instead of a `ValueError: different number of dimensions on data and dims` crash

## Backend coverage

- [x] numpy (was already working; kept under regression tests)
- [x] dask + numpy
- [x] cupy
- [x] dask + cupy

## Test plan

- [x] `TestReproject3DBackends`: lazy shape, computed shape, eager-vs-dask parity, integer dtype round-trip, cupy parity, dask+cupy compute
- [x] `TestMerge3DRejection`: clean `ValueError` instead of dim mismatch
- [x] Full `test_reproject.py` suite still green (267 tests)

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -251,10 +251,19 @@ def _reproject_chunk_numpy(
     c_min = np.nanmin(src_col_px)
     c_max = np.nanmax(src_col_px)
 
+    # 3-D source: empty-chunk returns must carry the band axis or the
+    # dask map_blocks template (which is 3-D for 3-D sources) sees a
+    # shape mismatch (#2027).
+    _src_ndim = getattr(source_data, 'ndim', 2)
+    if _src_ndim == 3:
+        _empty_shape = (*chunk_shape, source_data.shape[2])
+    else:
+        _empty_shape = chunk_shape
+
     if not np.isfinite(r_min) or not np.isfinite(r_max):
-        return np.full(chunk_shape, nodata, dtype=np.float64)
+        return np.full(_empty_shape, nodata, dtype=np.float64)
     if not np.isfinite(c_min) or not np.isfinite(c_max):
-        return np.full(chunk_shape, nodata, dtype=np.float64)
+        return np.full(_empty_shape, nodata, dtype=np.float64)
 
     r_min = int(np.floor(r_min)) - 2
     r_max = int(np.ceil(r_max)) + 3
@@ -263,7 +272,7 @@ def _reproject_chunk_numpy(
 
     # Check overlap
     if r_min >= src_h or r_max <= 0 or c_min >= src_w or c_max <= 0:
-        return np.full(chunk_shape, nodata, dtype=np.float64)
+        return np.full(_empty_shape, nodata, dtype=np.float64)
 
     # Clip to source bounds
     r_min_clip = max(0, r_min)
@@ -276,7 +285,7 @@ def _reproject_chunk_numpy(
     _MAX_WINDOW_PIXELS = 64 * 1024 * 1024  # 64 Mpix (~512 MB for float64)
     win_pixels = (r_max_clip - r_min_clip) * (c_max_clip - c_min_clip)
     if win_pixels > _MAX_WINDOW_PIXELS:
-        return np.full(chunk_shape, nodata, dtype=np.float64)
+        return np.full(_empty_shape, nodata, dtype=np.float64)
 
     # Extract source window
     window = source_data[r_min_clip:r_max_clip, c_min_clip:c_max_clip]
@@ -337,6 +346,14 @@ def _reproject_chunk_cupy(
     src_crs = _crs_from_wkt(src_wkt)
     tgt_crs = _crs_from_wkt(tgt_wkt)
 
+    # 3-D source: empty-chunk returns must carry the band axis to match
+    # the dask+cupy map_blocks template (#2027).
+    _src_ndim = getattr(source_data, 'ndim', 2)
+    if _src_ndim == 3:
+        _empty_shape = (*chunk_shape, source_data.shape[2])
+    else:
+        _empty_shape = chunk_shape
+
     # Try CUDA transform first (keeps coordinates on-device)
     cuda_result = None
     if src_crs is not None and tgt_crs is not None:
@@ -372,7 +389,7 @@ def _reproject_chunk_cupy(
         )
         if not (np.isfinite(r_min_val) and np.isfinite(r_max_val)
                 and np.isfinite(c_min_val) and np.isfinite(c_max_val)):
-            return cp.full(chunk_shape, nodata, dtype=cp.float64)
+            return cp.full(_empty_shape, nodata, dtype=cp.float64)
         r_min = int(np.floor(r_min_val)) - 2
         r_max = int(np.ceil(r_max_val)) + 3
         c_min = int(np.floor(c_min_val)) - 2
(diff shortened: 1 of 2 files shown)


brendancol:
## PR Review (self-review after follow-ups in 2b8c725)

### Blockers
- None.

### Suggestions (addressed in 2b8c725)
- `_reproject_chunk_cupy` 3-D branch was running `cp.where(band_data == nodata, cp.nan, band_data)` even on the `_use_native_cuda=True` path. The 2-D path skips that conversion because the native CUDA kernels handle the sentinel internally. The 3-D branch now mirrors that contract: pre-convert only on the CPU-coords fallback.
- 3-D cupy coverage with `nodata=0` (uint8 sentinel) was missing — only the `nodata=np.nan` path was tested. Added `test_reproject_3d_cupy_uint8_sentinel_nodata`.

### Nits (addressed in 2b8c725)
- Dropped `getattr(source_data, 'ndim', 2)` defensives in both chunk functions; numpy/cupy/dask all expose `ndim` directly.
- Added a comment on the 3-D dask template explaining that the band axis is intentionally collapsed to one output chunk regardless of the source's band chunking.

### What looks good
- Output rank now matches input rank on every backend; lazy `result.shape` agrees with `result.compute().shape`.
- Empty-chunk early returns in both chunk functions carry the band axis, so partial-overlap projections on 3-D sources no longer trip the `map_blocks` template assembly.
- `_reproject_dask_cupy` delegates 3-D to `map_blocks` instead of retrofitting the inline fast path — keeps the diff focused.
- `merge()` validator tightened to `ndim=(2,)` — `_place_same_crs`, `_merge_arrays_numpy`, and the final `xr.DataArray(..., dims=[ydim, xdim])` are all 2-D-only, so the previous `ndim=(2, 3)` was a promise the implementation never kept.
- Tests cover numpy / dask / cupy / dask+cupy with eager-vs-dask `assert_allclose` parity, integer dtype round-trips on CPU and GPU, and the issue's exact reproducers.
- 268 tests pass; no regressions in the existing 2-D paths.

### Checklist
- [x] Algorithm matches the existing 2-D per-band loop pattern
- [x] All implemented backends produce consistent results (`assert_allclose` vs numpy)
- [x] NaN handling is correct
- [x] Edge cases covered: uint8 round-trip on CPU and GPU, no-overlap empty chunks
- [ ] Benchmark — 3-D bench would be nice-to-have but the existing `bench_reproject_vs_rioxarray.py` is 2-D; not blocking for a correctness fix
- [x] README feature matrix — reproject already listed; backend coverage went from advertised-but-broken to actually-supported for 3-D
- [x] Docstrings updated on `_reproject_dask` and `_reproject_block_adapter`

PR #2056, opened 2026-05-18
geotiff: reject zero/negative dimensions in stripped TIFF reads (#2053)

### LINKED ISSUE #2053: [Bug] Stripped TIFFs with zero/negative dimensions return empty arrays instead of failing closed
## Summary

`_check_dimensions` in `xrspatial/geotiff/_reader.py` (line 102) only enforces the upper bound (`total > max_pixels`). It does not reject `width <= 0`, `height <= 0`, or `samples <= 0`. A malformed stripped TIFF that declares `ImageWidth=0` or `ImageLength=0` produces an empty array instead of raising at the source.

## Validation gap (stripped vs tiled)

The tiled path is covered. `_read_tiles` calls `validate_tile_layout(ifd)` from `_header.py`, which raises `ValueError` for `width <= 0`, `height <= 0`, or `tile_width/tile_height <= 0`. The stripped path has no equivalent. `_read_strips` reads `width = ifd.width; height = ifd.height` at line 1707 and then only validates the clamped output window at line 1766 via `_check_dimensions(out_w, out_h, ...)`. A source IFD with `ImageWidth=0` flows through, produces `out_w=0`, and the upper-bound check passes.

## Affected call sites

- `_read_strips` line 1707 (reads source w/h, then jumps to post-window check)
- `_fetch_decode_cog_http_strips` line 2380 (HTTP equivalent, same gap)

## Downstream risk

An empty `(0, N)` or `(N, 0)` array threads through `read_to_array` / `open_geotiff`. Spatial functions downstream handle empty arrays inconsistently. Some raise on empty input, some silently produce empty output, some return NaN-filled output. Failing closed at the parser is the simpler contract.

## Proposed fix

Two-layer defense:
1. Add `_check_source_dimensions(width, height, samples)` next to `_check_dimensions`. Raises `ValueError` on any value <= 0.
2. Call it in `_read_strips` and `_fetch_decode_cog_http_strips` right after reading `width = ifd.width; height = ifd.height; samples = ifd.samples_per_pixel`, before any window clamping. Leave `_check_dimensions` unchanged so the post-window path still returns an empty array when the caller passes a window outside the image bounds.

brendancol:
Closes #2053.

## Summary

The stripped TIFF read paths in `xrspatial/geotiff/_reader.py` previously trusted `ImageWidth`, `ImageLength`, and `SamplesPerPixel` straight off the IFD. `_check_dimensions` (line 102) only enforces the upper bound, and the post-window clamp collapses a malformed `ImageWidth=0` source into a silent empty array instead of raising.

The tiled paths are already covered: `_read_tiles` calls `validate_tile_layout(ifd)` in `_header.py`, which rejects `<= 0` on width, height, tile_width, and tile_height. The stripped paths had no equivalent.

## Fix

Two-layer defense:

1. New `_check_source_dimensions(width, height, samples)` helper next to `_check_dimensions`. Raises `ValueError("Invalid TIFF dimensions: ...")` on any value `<= 0`.
2. Both stripped read paths call it right after reading the IFD, before window clamping:
   - `_read_strips` (local files, line ~1707)
   - `_fetch_decode_cog_http_strips` (HTTP COGs, line ~2380)

## What stays the same

The post-window `_check_dimensions` calls are unchanged. A windowed read whose window starts at the image edge still legitimately returns an empty array. The strict check only runs on source IFD dims, not on the clamped output dims (option A from the issue thread).

The tiled paths are unchanged; `validate_tile_layout` already provides equivalent coverage. The new test file pins the parity so a future refactor of either side cannot silently break the other.

## Test plan

- [x] `python -m pytest xrspatial/geotiff/tests/test_strip_zero_dims_2053.py` (17 tests, all passing)
- [x] `python -m pytest xrspatial/geotiff/tests/ -k "strip or dim or validation"` (507 passed, 1 pre-existing failure on main unrelated to this change in `test_size_param_validation_gpu_vrt_1776.py`)

Test coverage:

- Unit tests on `_check_source_dimensions` for zero / negative / mixed inputs
- End-to-end tests: patch `ImageWidth`, `ImageLength`, `SamplesPerPixel` to 0 in a valid stripped TIFF on disk, assert `open_geotiff` raises `ValueError`
- Negative-cast test: 0xFFFFFFFF as ImageWidth, assert it raises (either via the strict check or via the upper-bound `_check_dimensions`)
- Windowed-empty-still-allowed test: a window starting at the image's bottom edge still produces an empty array, confirming the strict check sits before the clamp
- Parity pin: tiled TIFFs with `ImageWidth=0` and `ImageLength=0` still reject through `validate_tile_layout`
- HTTP path: the same patched-bytes test via a static `_HTTPSource` stand-in covers `_fetch_decode_cog_http_strips`

### DIFF
--- a/xrspatial/geotiff/_reader.py
+++ b/xrspatial/geotiff/_reader.py
@@ -111,6 +111,27 @@ def _check_dimensions(width, height, samples, max_pixels):
         )
 
 
+def _check_source_dimensions(width, height, samples):
+    """Validate the source IFD dimensions of a TIFF before any windowing.
+
+    Companion to :func:`_check_dimensions`, which only enforces the
+    upper bound. The stripped read paths read ``width``,  ``height``,
+    and ``samples_per_pixel`` straight off the IFD and then clamp the
+    output window to those values, so a malformed file with
+    ``ImageWidth = 0`` (or a negative value, which would parse as a
+    huge unsigned int but can also surface via signed-cast errors)
+    would produce an empty array silently. The tiled paths are already
+    protected by :func:`validate_tile_layout` in ``_header.py``; this
+    helper closes the same gap for the stripped path. Issue #2053.
+    """
+    if width <= 0 or height <= 0 or samples <= 0:
+        raise ValueError(
+            f"Invalid TIFF dimensions: ImageWidth={width}, "
+            f"ImageLength={height}, SamplesPerPixel={samples} "
+            f"(all must be > 0)"
+        )
+
+
 #: Default per-tile (or per-strip) compressed-byte cap. A crafted
 #: ``TileByteCounts`` / ``StripByteCounts`` entry can declare arbitrarily
 #: many bytes. On HTTP, the reader would issue a Range GET sized by the
@@ -1707,6 +1728,12 @@ def _read_strips(data: bytes, ifd: IFD, header: TIFFHeader,
     width = ifd.width
     height = ifd.height
     samples = ifd.samples_per_pixel
+    # Source-IFD dim check (issue #2053). The tiled path is already
+    # covered by ``validate_tile_layout``; this is its stripped-path
+    # parity. Run before any window clamping so a malformed
+    # ``ImageWidth=0`` IFD fails at the source rather than collapsing
+    # to an empty post-clamp window.
+    _check_source_dimensions(width, height, samples)
     compression = ifd.compression
     rps = ifd.rows_per_strip
     offsets = ifd.strip_offsets
@@ -2380,6 +2407,12 @@ def _fetch_decode_cog_http_strips(
     width = ifd.width
     height = ifd.height
     samples = ifd.samples_per_pixel
+    # Source-IFD dim check (issue #2053). Mirror of the local-path
+    # check in ``_read_strips`` so HTTP COG reads of a malformed
+    # stripped file fail at the source rather than collapsing to an
+    # empty post-clamp window. Tiled paths already get the equivalent
+    # check from ``validate_tile_layout``.
+    _check_source_dimensions(width, height, samples)
     compression = ifd.compression
     rps = ifd.rows_per_strip
     offsets = ifd.strip_offsets
--- a/xrspatial/geotiff/tests/test_strip_zero_dims_2053.py
+++ b/xrspatial/geotiff/tests/test_strip_zero_dims_2053.py
@@ -22,7 +22,6 @@ tiled path can't silently regress the stripped path.
 """
 from __future__ import annotations
 
-import io
 import struct
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


brendancol:
## PR Review: geotiff: reject zero/negative dimensions in stripped TIFF reads (#2053)

None.

None.

- [ ] `xrspatial/geotiff/tests/test_strip_zero_dims_2053.py:25` — `import io` is unused. Drop it.

### What looks good

- Fix is small. One helper, two call sites, plus tests. The diff is easy to audit.
- Two-layer defense rationale matches "option A" from the issue thread: strict source-IFD check, unchanged post-window upper-bound check. The call-site comments name `validate_tile_layout` for parity, so the design intent is visible at the call sites, not buried in the helper docstring.
- Test coverage is thorough:
  - Unit tests on `_check_source_dimensions` for zero and negative inputs on width, height, and samples
  - End-to-end tests that patch a real on-disk TIFF and assert `open_geotiff` raises
  - Parity-pin tests for the tiled branch (catches a future regression in `validate_tile_layout`)
  - HTTP-path tests via a static `_HTTPSource` stand-in (covers `_fetch_decode_cog_http_strips`)
  - Negative-cast test (`0xFFFFFFFF`) confirms the file is rejected by some check, even if the strict one were bypassed
  - The windowed-empty test pins option A: a window at the image edge still returns an empty array
- The IFD-entry-patching helpers (`_find_ifd_entry_offset`, `_patch_inline_long`) are self-contained and well-commented. They make malformed-TIFF fixtures cheap to write without rolling a full hand-crafted IFD by hand.
- The check sits in the right place: right after `ifd.width / height / samples_per_pixel` are read, before window clamping or other validation. A regression that moves it past the clamp would surface as the windowed-empty test passing where the source-zero tests fail.

### Checklist

- [x] Algorithm matches reference / paper — n/a, validation check
- [x] All implemented backends produce consistent results — local and HTTP stripped paths both tightened; tiled was already covered
- [x] NaN handling is correct — n/a
- [x] Edge cases are covered by tests — zero and negative on all three dims, both branches
- [x] Dask chunk boundaries handled correctly — n/a, dask path doesn't reach `_read_strips` / `_fetch_decode_cog_http_strips` directly
- [x] No premature materialization or unnecessary copies — helper is O(1)
- [x] Benchmark exists or is not needed — not needed, one-line check on the header-parse path
- [x] README feature matrix updated — n/a, no public API change
- [x] Docstrings present and accurate — helper has a docstring; call-site comments reference the issue

### Out-of-scope follow-up

`validate_tile_layout` rejects `<= 0` on width, height, tile_width, and tile_height, but does not check `samples_per_pixel`. A separate PR could close that gap for parity with the new stripped check.

PR #2084, opened 2026-05-18
geotiff: converge the two deprecated-attr emit helpers in _attrs.py (#2017)

### LINKED ISSUE #2017: geotiff: converge the two deprecated-attr emit helpers in _attrs.py
## Background

PR 7 of issue #1984 deprecated 13 attrs across four slices:

- #2010 vertical-CRS GeoKey attrs
- #2011 geographic-CRS GeoKey attrs
- #2013 projected-CRS GeoKey attrs
- #2014 matplotlib colormap variants

`xrspatial/geotiff/_attrs.py` now exposes two emit helpers that do almost the same job:

```python
def _emit_deprecated_geokey_attr(attrs, name, value, *, reason): ...

def _emit_deprecated_attr(attrs, name, value, *, reason, migration=None): ...
```

`_emit_deprecated_geokey_attr` covers the three GeoKey slices and bakes the sentence "...{reason} so it will not round-trip..." into the warning text. `_emit_deprecated_attr` covers the colormap variants and supports an optional `migration=` clause (e.g., "Reshape `attrs['colormap']` to `(n_colors, 3)` and append an alpha channel").

The split exists because the warning-text contracts are pinned by separate test suites:

- `test_attrs_pr7_deprecate_geographic_1984.py` asserts on the GeoKey-tier wording verbatim via `_deprecated_geographic_geokey_warning`.
- `test_attrs_pr7_deprecate_vertical_1984.py` and `test_attrs_pr7_deprecate_projected_1984.py` assert on substring matches.
- `test_attrs_pr7_deprecate_colormap_variants_1984.py` asserts on the byte-identical pre-refactor wording for `cmap` and `colormap_rgba`.

Unifying them was out of scope for the colormap slice but is mechanical.

## Proposal

Collapse the two helpers into one:

```python
def _emit_deprecated_attr(
    attrs, name, value, *,
    reason: str,
    migration: str | None = None,
    suffix: str | None = None,
) -> None: ...
```

Where:

- `reason` is the per-tier short clause (e.g., "the writer cannot reconstruct it from the canonical CRS").
- `suffix` defaults to `"so it will not round-trip."` to match the GeoKey-tier baked wording, and is overridden by the colormap slice with its own phrasing.
- `migration` is the optional per-attr migration recipe used today only by the colormap variants.

Alternative: drop `suffix` entirely and require every caller to bake its own "round-trip" sentence into `reason`. That makes the helper simpler but forces a coordinated update of the three GeoKey test suites that pin the current wording.

## Acceptance criteria

- One emit helper in `_attrs.py`. The text builder `_deprecated_geokey_warning` either stays as a thin alias for the GeoKey-tier callers or gets folded in.
- All four PR 7 test suites (`test_attrs_pr7_deprecate_*_1984.py`) pass without their existing assertions weakening. If the wording must shift to consolidate, update the assertions in the same PR.
- `_stacklevel_to_external_caller()` keeps doing the heavy lifting for stacklevel resolution; no behavior change there.
- `_DEPRECATED_GEOGRAPHIC_GEOKEY_ATTRS` stays as the tier-membership constant; mirror constants for the projected, vertical, and colormap-variants tiers if it makes the tier tests cleaner.

## Why this is bounded

The two helpers were left side-by-side because the colormap slice landed last and unifying them required touching three sibling test suites. With all four slices now in main, the work is a single PR that touches one file plus the four test files. No production-behavior change visible to users.

Refs #1984.

brendancol:
## Summary

- Collapse the two PR 7 emit helpers in `_attrs.py` into one. `_emit_deprecated_attr` now takes an optional `suffix` so the GeoKey tiers and the colormap-variants tier go through the same code path.
- `_emit_deprecated_geokey_attr` becomes a thin shim that pins `suffix=_GEOKEY_DEPRECATION_SUFFIX` ("so it will not round-trip.") so the three GeoKey-axis callers don't repeat the suffix.
- Rendered warning text is byte-identical for every existing call site. The four PR 7 test suites pass without weakening their assertions. `_stacklevel_to_external_caller` already walks past every `xrspatial.geotiff*` frame, so the extra shim layer doesn't shift where warnings are attributed.

Closes #2017. Refs #1984.

## Backend coverage

Refactor of the attrs-emission helpers. No backend-specific code touched.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/ -k deprecat` (37 passed)
- [x] `pytest xrspatial/geotiff/tests/test_attrs_pr7_*.py` (30 passed)
- [x] Byte-identical wording check against the pre-refactor strings for the GeoKey and colormap-variants forms.

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -212,11 +212,12 @@ _DEPRECATED_GEOGRAPHIC_GEOKEY_ATTRS = (
 )
 
 
-# Per-category reason clauses spliced into the deprecation warning by
-# :func:`_emit_deprecated_geokey_attr`. Kept here so the wording stays
-# in lockstep across the three GeoKey-axis tiers (geographic, projected,
-# vertical) and so the test suite can match the canonical strings
-# verbatim.
+# Per-category reason clauses spliced into the GeoKey-tier deprecation
+# warning by :func:`_emit_deprecated_geokey_attr` (which delegates to
+# the unified :func:`_emit_deprecated_attr`). Kept here so the wording
+# stays in lockstep across the three GeoKey-axis tiers (geographic,
+# projected, vertical) and so the test suite can match the canonical
+# strings verbatim.
 _GEOKEY_DEPRECATION_REASON_HORIZONTAL_CRS = (
     "the writer cannot reconstruct it from the canonical CRS"
 )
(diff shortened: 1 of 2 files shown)


brendancol:
Closing as superseded by #2089, which removed the deprecated-attr emission entirely and bumped the attrs contract to v2. The two helpers this PR set out to converge (`_emit_deprecated_geokey_attr` and `_emit_deprecated_attr`) no longer exist on main, along with the four PR 7 test suites that pinned their warning text. The convergence work is no longer needed.

Issue #2075, opened 2026-05-18
to_geotiff writes invalid TIFF for zero-height/zero-width arrays instead of failing closed


brendancol:
**Describe the bug**

`to_geotiff` accepts arrays whose spatial height or width is zero. It writes a TIFF file, and the reader then refuses to open it with "Invalid image dimensions". Clip and window pipelines can produce empty rasters, so this turns a recoverable input-validation error into a stale-file-on-disk failure that only surfaces at read time.

The eager validation at `xrspatial/geotiff/_writers/eager.py:670` checks `arr.ndim` but never that height and width are positive. The streaming path at `xrspatial/geotiff/_writer.py:1939` then computes zero tile/strip counts and writes a file with no pixel data.

**Reproducer**

```python
import numpy as np
import xarray as xr
from xrspatial.geotiff import to_geotiff, open_geotiff

for shape in [(0, 5), (5, 0), (0, 0)]:
    da = xr.DataArray(np.zeros(shape, dtype=np.float32), dims=('y', 'x'))
    to_geotiff(da, f'/tmp/empty_{shape[0]}x{shape[1]}.tif')  # succeeds
    open_geotiff(f'/tmp/empty_{shape[0]}x{shape[1]}.tif')    # raises "Invalid image dimensions"
```

**Expected behavior**

`to_geotiff` raises `ValueError` at the write entry point with a message that names the offending dimension (e.g. "got shape (0, 5) with height=0"). The failure happens before any bytes hit disk.

**Actual behavior**

The writer produces a file. The error surfaces later at read time with a generic "Invalid image dimensions" message that does not name the writer as the source.

**Source locations**

- Eager entry: `xrspatial/geotiff/_writers/eager.py:670` (the `arr.ndim not in (2, 3)` check).
- Streaming entry: `xrspatial/geotiff/_writer.py:1939` (tile/strip count math runs on `height`/`width`).

**Additional context**

Affects all four backends (numpy, cupy, dask+numpy, dask+cupy). The validation gap is upstream of backend dispatch.

Issue #1812, opened 2026-05-13
to_geotiff: silent data corruption on 3D input with non-whitelisted leading dim


brendancol:
**Describe the bug**

`to_geotiff` silently corrupts data when the input is a 3D `xr.DataArray`
whose leading dim is not in `('band', 'bands', 'channel')`. The writer
relies on a `moveaxis` at `xrspatial/geotiff/__init__.py:1686` (eager) and
`:1636` (dask) that only fires when `data.dims[0]` is in the
`_BAND_DIM_NAMES` whitelist. Anything else (e.g. `('time', 'y', 'x')`)
skips the moveaxis, so the array passes through untouched and the writer
treats the leading axis as the spatial `y` axis.

The read backends honour the on-disk axis order and emit `(y, x, band)`,
so the round-trip produces an array with the wrong shape and wrong
per-slice contents.

**Repro**

```python
import numpy as np
import xarray as xr
import tempfile
from xrspatial.geotiff import open_geotiff, to_geotiff

arr = np.zeros((2, 4, 5), dtype=np.uint8)
arr[0] = 1
arr[1] = 2

da = xr.DataArray(arr, dims=('time', 'y', 'x'),
                  attrs={'crs': 'EPSG:4326'})

tmp = tempfile.NamedTemporaryFile(suffix='.tif', delete=False)
tmp.close()
to_geotiff(da, tmp.name, crs=4326)

out = open_geotiff(tmp.name)
print(out.shape)                  # (2, 4, 5)  wrong; should be (4, 5, 2)
print(out.values[:, :, 0].sum())  # 12         wrong; should be arr[0].sum() == 20
print(out.values[:, :, 1].sum())  # 12         wrong; should be arr[1].sum() == 40
```

**Expected behaviour**

The writer should refuse an ambiguous 3D layout and raise `ValueError`,
telling the caller to either rename the leading dim to one of
`_BAND_DIM_NAMES` or transpose to `(y, x, band)`. Silent pass-through
gives the user no signal that the layout was misinterpreted.

The same gate is used in three writer paths, all of which need the fix:

- `to_geotiff` eager branch (`xrspatial/geotiff/__init__.py:1686`)
- `to_geotiff` dask-streaming branch (`xrspatial/geotiff/__init__.py:1636`)
- `write_geotiff_gpu` (`xrspatial/geotiff/__init__.py:3636+`)

**Additional context**

Surfaced by the 2026-05-13 metadata propagation sweep. Existing tests
[...]

PR #2096, opened 2026-05-19
geotiff: reject zero-band 3D writes at the entry point (#2095)

### LINKED ISSUE #2095: geotiff: zero-band 3D writes silently produce a bogus single-band TIFF
**Describe the bug**

`xrspatial.geotiff.to_geotiff` (and the underlying `write` / `write_streaming` / GPU writers) silently writes a corrupt single-band TIFF when handed a 3D DataArray whose band axis has length 0.

`_validate_writer_spatial_shape` in `xrspatial/geotiff/_validation.py` validates the spatial height and width for 3D inputs, but does not check the band/sample dimension. A DataArray with shape `(0, y, x)` and dims `('band', 'y', 'x')`, or shape `(y, x, 0)` with dims `('y', 'x', 'band')`, passes validation. In `_writer.py` around line 1145, `samples_per_pixel = first_arr.shape[2] if first_arr.ndim == 3 else 1` resolves to 0, and the file is assembled anyway around line 1198.

**Reproduction**

```python
import numpy as np
import xarray as xr

# Band-first, zero bands
da1 = xr.DataArray(np.zeros((0, 5, 5), dtype="uint8"), dims=("band", "y", "x"))
to_geotiff(da1, "zero_bands_first.tif")
print(open_geotiff("zero_bands_first.tif").shape)  # -> (5, 5) uint8

# Band-last, zero bands
da2 = xr.DataArray(np.zeros((5, 5, 0), dtype="uint8"), dims=("y", "x", "band"))
to_geotiff(da2, "zero_bands_last.tif")
print(open_geotiff("zero_bands_last.tif").shape)  # -> (5, 5) uint8
```

Both calls succeed and the file round-trips as a (5, 5) single-band uint8 raster.

**Expected behavior**

The writer should fail closed at the public entry point with a clear message like "band/sample dimension must be positive; got shape (0, 5, 5) with 0 bands." No file should be written.

**Why it matters**

A clip, window, or selection that produces zero bands is a programmer error upstream. The writer treating that as "write a 2D raster of the same spatial size" is silent data fabrication. The on-disk file looks legitimate and reads back without an error from `open_geotiff`. Downstream consumers have no signal that the original band axis collapsed.

**Scope of fix**

- Extend `_validate_writer_spatial_shape` (or add a peer validator) to reject `bands == 0` on both layouts.
- Wire the new check into every public writer entry point that already calls the spatial validator: `to_geotiff` / `write` / `write_streaming` and the GPU writer.
- Tests for both layouts on numpy and dask inputs.

brendancol:
## Summary
- The 3D spatial-shape validator checked height and width but skipped the band axis. Zero-band inputs (`(0, y, x)` band-first or `(y, x, 0)` band-last) reached IFD assembly with `samples_per_pixel == 0` and produced a TIFF that round-tripped as a 2D single-band raster.
- Extended `_validate_writer_spatial_shape` in `xrspatial/geotiff/_validation.py` to also reject `bands == 0` on both layouts. The check runs before any bytes are written.
- Every public writer entry point (`to_geotiff`, `write`, `write_streaming`, `write_geotiff_gpu`) already calls this helper, so no caller-side changes were needed.

## Backend coverage
- numpy: tested
- dask + numpy: tested (`write_streaming` direct plus `to_geotiff` dask path)
- cupy: GPU test included, skipped when CUDA is unavailable
- dask + cupy: same entry point as cupy; the validator runs before any GPU work

## Test plan
- [x] `pytest xrspatial/geotiff/tests/test_to_geotiff_zero_bands_2095.py`
- [x] `pytest xrspatial/geotiff/tests/test_to_geotiff_empty_shape_2075.py` (regression)
- [x] `pytest xrspatial/geotiff/tests/test_to_geotiff_3d_dim_validation_1812.py` (regression)

Closes #2095


brendancol (commented):
## PR Review: geotiff: reject zero-band 3D writes at the entry point (#2095)

### Blockers
None.

### Suggestions
- [ ] `xrspatial/geotiff/tests/test_to_geotiff_zero_bands_2095.py:75,91,123`: the assertions `assert "write" in msg` / `assert "write_streaming" in msg` / `assert "write_geotiff_gpu" in msg` are weak. Every entry-point message contains the substring "write" (the error template says "write a raster with no bands"), so the bare-`write` direct-call test cannot tell `entry_point="write"` apart from `"write_streaming"` or `"write_geotiff_gpu"`. Anchor to the message prefix (e.g. `msg.startswith("write ")` for the bare-`write` test) or assert the exact entry_point name followed by " cannot write" to guarantee the right path actually fired.

### Nits
- [ ] `xrspatial/geotiff/_validation.py:166-174`: in `to_geotiff` this spatial validator runs before `_validate_3d_writer_dims` (#1812 / #1972). For an array of shape `(5, 5, 0)` with dims `('y', 'x', 'time')`, this validator now infers `bands = 0` (since `'time'` is not in `_BAND_DIM_NAMES`, it falls to the band-last branch) and raises the "no bands" error rather than the friendlier ambiguous-dim / temporal-dim message from `_validate_3d_writer_dims`. The rejection is still correct, but the message points at the wrong root cause. Not blocking, since both messages name the right call to fix, but worth a comment or a reorder if you want the ambiguous-dim message to win.

### What looks good
- One fail-closed gate, called from every public writer entry point that already calls the spatial-shape validator. No caller-side changes needed.
- Test coverage hits both layouts on numpy + dask, plus direct `write`, `write_streaming`, and `write_geotiff_gpu`. Tests use `tmp_path` fixtures with issue-numbered filenames so they cannot collide.
- Mirrors the #2075 fail-closed pattern, including the `entry_point` message threading.
- Validator runs before any IFD math or device dispatch, so no bytes hit disk and no GPU work starts.

### Checklist
- [x] Algorithm matches reference/paper — N/A (validator)
- [x] Benchmark exists or is not needed — not needed (input rejection)
- [x] README feature matrix updated (if applicable) — N/A

brendancol (commented):
## Follow-up: review findings disposition

- **Suggestion** (test assertions): **Fixed.** Replaced `assert "write" in msg` / `assert "write_streaming" in msg` / `assert "write_geotiff_gpu" in msg` with `msg.startswith("<entry_point> cannot write")`. The error template starts with `"<entry_point> cannot write a raster ..."`, so anchoring to the prefix means a wrong path firing now fails the test.
- **Nit** (validator ordering): **Documented.** Added a docstring note to `_validate_writer_spatial_shape` explaining that this validator runs before `_validate_3d_writer_dims`, so an ambiguous-dim array with a zero-length axis (e.g. `(5, 5, 0)` with `('y', 'x', 'time')`) now triggers the "no bands" message rather than the ambiguous-dim diagnostic. Both messages name the right call to fix, so I kept the ordering and explained the trade-off rather than reordering.

Tests re-run: `test_to_geotiff_zero_bands_2095.py` (7/7 pass), `test_to_geotiff_empty_shape_2075.py` (5/5), `test_to_geotiff_3d_dim_validation_1812.py` (19/19).

PR #2098, opened 2026-05-19
geotiff: reject zero-band 3D writer inputs (#2094)

### LINKED ISSUE #2094: GeoTIFF writers need shared post-layout shape validator
## Problem

The GeoTIFF writer entry points validate empty spatial dimensions, but they do not validate the resolved band/sample dimension after normalizing layout. A 3D DataArray with zero bands can pass validation and still write a TIFF.

Observed locally:

```python
import numpy as np
import xarray as xr

da = xr.DataArray(
    np.empty((0, 5, 5), dtype=np.uint8),
    dims=("band", "y", "x"),
)
to_geotiff(da, "zero_band.tif", compression="none")
print(open_geotiff("zero_band.tif").shape)  # (5, 5)

da2 = xr.DataArray(
    np.empty((5, 5, 0), dtype=np.uint8),
    dims=("y", "x", "band"),
)
to_geotiff(da2, "zero_band_last.tif", compression="none")
print(open_geotiff("zero_band_last.tif").shape)  # (5, 5)
```

That is silent data fabrication: an empty band axis writes a file that reads back as a non-empty single-band raster.

## Why this matters

GeoTIFF is the I/O boundary for downstream raster functions. If this module accepts and normalizes an invalid layout into plausible raster bytes, later analysis code has no way to distinguish a real single-band raster from an invalid zero-band write.

## Proposed fix

Add one shared post-layout writer validator that runs after each writer has normalized layout to the internal convention, and before any bytes are emitted. It should validate the actual write layout, not just the original public input shape:

- `height > 0`
- `width > 0`
- `samples_per_pixel > 0`
- optional: reject ranks other than the writer-supported 2D / band-last 3D layout at this same boundary

This validator should be used by:

- eager `to_geotiff` path after band-first to band-last remap
- direct `_writer.write` path
- streaming dask writer
- VRT tiled writer / per-tile path where applicable
- GPU writer after band-axis normalization

## Acceptance tests

Add regression tests that assert `ValueError` for:

- `dims=("band", "y", "x")`, shape `(0, 5, 5)`
- `dims=("y", "x", "band")`, shape `(5, 5, 0)`
- direct `write(np.empty((5, 5, 0), dtype=np.uint8), ...)`
- dask-backed equivalents for streaming where feasible
- VRT output path if it can receive a zero-band 3D array
- GPU entry point if GPU tests are available / can be skipped when CuPy is absent

Related code areas:

- `xrspatial/geotiff/_validation.py::_validate_writer_spatial_shape`
- `xrspatial/geotiff/_writers/eager.py::to_geotiff`
- `xrspatial/geotiff/_writer.py::write` and `write_streaming`
- `xrspatial/geotiff/_writers/gpu.py::write_geotiff_gpu`

brendancol:
Closes #2094.

## Summary

- New shared validator `_validate_writer_post_layout_shape` runs after each writer normalises layout to its internal band-last convention; refuses height, width, or samples-per-pixel of zero with a message that names the entry point.
- Wired into `to_geotiff` (eager numpy / cupy and dask streaming paths after the band-first `moveaxis`), `_writer.write`, `_writer.write_streaming`, and `write_geotiff_gpu` after the GPU `moveaxis`.
- The pre-layout `_validate_writer_spatial_shape` stays in place; it covers the bare 2D and dims-aware spatial check. The new post-layout helper covers the band axis the pre-layout helper left unchecked.

## Backend coverage

- numpy: regression tests for band-first `(0, 5, 5)` and band-last `(5, 5, 0)` through `to_geotiff` and direct `_writer.write`.
- dask+numpy: regression test through `to_geotiff` streaming for both layouts; direct `write_streaming` test on band-last.
- cupy / dask+cupy: regression test through `write_geotiff_gpu` for both layouts (skipped when cupy is absent).

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_to_geotiff_zero_band_2094.py -v` -- 9 passed.
- [x] `pytest xrspatial/geotiff/tests/test_to_geotiff_empty_shape_2075.py xrspatial/geotiff/tests/test_degenerate_shapes_backends_2026_05_11.py xrspatial/geotiff/tests/test_strip_zero_dims_2053.py` -- 45 passed.
- [x] Broader writer sweep `pytest xrspatial/geotiff/tests/ -k "writer or roundtrip or to_geotiff or write"` -- 928 passed, 1 pre-existing failure unrelated to this change.

### DIFF
--- a/xrspatial/geotiff/_validation.py
+++ b/xrspatial/geotiff/_validation.py
@@ -173,6 +173,63 @@ def _validate_writer_spatial_shape(shape, dims=None,
         )
 
 
+def _validate_writer_post_layout_shape(
+        shape, entry_point: str = "to_geotiff") -> None:
+    """Reject empty / unsupported post-layout writer inputs (issue #2094).
+
+    Runs AFTER the writer has normalised its input to the internal
+    convention: 2D ``(h, w)`` or band-last 3D ``(h, w, samples)``.
+    Validates every dimension that becomes a TIFF IFD field once the
+    array is on disk: height, width, and (for 3D) samples-per-pixel.
+
+    The pre-layout sibling ``_validate_writer_spatial_shape`` only
+    inspects the two spatial dims and leaves the band axis unchecked.
+    A 3D input with a zero-length band axis (``(0, 5, 5)`` band-first
+    or ``(5, 5, 0)`` band-last) used to pass that helper, hit the
+    band-first ``moveaxis``, and then be written as a TIFF whose IFD
+    advertised ``SamplesPerPixel=1`` and ``height=5, width=5`` -- the
+    file decoded back as a non-empty single-band raster from an
+    empty input. That is silent data fabrication; the writer must
+    refuse the input instead.
+
+    ``shape`` is the array shape AFTER any band-first -> band-last
+    ``moveaxis`` has run, so the helper does not need a ``dims`` arg.
+    ``entry_point`` is the function name used in the error message
+    so direct callers of ``write`` / ``write_streaming`` /
+    ``write_geotiff_gpu`` see the function they actually invoked.
+    """
+    if shape is None:
+        return
+    ndim = len(shape)
+    if ndim not in (2, 3):
+        # Other rank errors are handled by the existing ndim check in
+        # each writer; do not shadow that message.
+        return
+    h = int(shape[0])
+    w = int(shape[1])
+    if h <= 0 or w <= 0:
+        raise ValueError(
+            f"{entry_point} cannot write an empty raster: got shape "
+            f"{tuple(int(s) for s in shape)} with height={h}, width={w}. "
+            f"Both spatial dims must be positive. A common cause is a "
+            f"clip or window that produced an empty selection; check "
+            f"the upstream operation before writing."
+        )
+    if ndim == 3:
+        samples = int(shape[2])
+        if samples <= 0:
+            raise ValueError(
+                f"{entry_point} cannot write a zero-band raster: got "
+                f"shape {tuple(int(s) for s in shape)} with "
+                f"samples_per_pixel={samples}. A 3D writer input must "
+                f"have at least one band along the band axis; an empty "
+                f"band axis used to round-trip as a single-band TIFF, "
+                f"silently fabricating one band of pixels from an empty "
+                f"input (issue #2094). Check the upstream selection "
+                f"(e.g. ``data.isel(band=slice(0, 0))``) before writing."
+            )
+
+
 def _validate_dtype_cast(source_dtype, target_dtype):
     """Validate that casting source_dtype to target_dtype is allowed.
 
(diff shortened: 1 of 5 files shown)


brendancol (commented):
## PR Review

### Blockers
None.

### Suggestions
- `_writer.py:1633-1634` and `_writer.py:1963-1964`: at these two call sites the pre-layout `_validate_writer_spatial_shape(shape, entry_point=...)` is now fully subsumed by the new `_validate_writer_post_layout_shape`. Both helpers default to band-last when `dims=None`, so both pick `h, w = shape[0], shape[1]`, raise the same empty-raster error for the spatial-zero case, and only the new helper catches the zero-band-axis case. The pre-layout call here is redundant. Consider dropping it from `write` and `write_streaming` to leave a single validator. Not a regression; the duplication is cosmetic.

### Nits
- `_validation.py:227`: the hint `data.isel(band=slice(0, 0))` reads as a fix suggestion, but that selection is exactly what produces a zero-band input. A clearer reading would be something like "an upstream `isel(band=slice(0, 0))` would produce this shape; verify the selection."

### What looks good
- Validator runs at the right boundary in every entry point: after the band-first `moveaxis` in the eager and dask `to_geotiff` paths and in `write_geotiff_gpu`, and at the top of `write` / `write_streaming` where input is already band-last by contract.
- Error messages name the actual entry point the user called, matching the pattern set by `_validate_writer_spatial_shape`.
- Regression tests cover both layouts (`(0, 5, 5)` band-first and `(5, 5, 0)` band-last) across numpy, dask streaming, direct `write` / `write_streaming`, and `write_geotiff_gpu`, plus a positive control on a valid 3D write.
- Public surface unchanged. The new helper is underscore-prefixed and the docs only list public entry points, so no docs update is needed.
- VRT path correctly excluded from the validator wiring: `_write_vrt_tiled` already rejects all 3D inputs, so zero-band data cannot reach the per-tile writer.

### Checklist
- [x] Refusal matches the issue spec (`height > 0`, `width > 0`, `samples_per_pixel > 0` after layout normalization).
- [x] Hooked into all listed touchpoints except VRT (intentional; VRT path is 2D-only upstream).
- [x] Error message names the offending dim and the entry point.
- [x] Tests assert `ValueError` and that no file was written.
- [x] CuPy regression test included for `write_geotiff_gpu` (skipped without CUDA).
- [x] No backend matrix change.
- [x] No docstring drift on public functions.

brendancol (commented):
## Follow-up Review

Addressed the previous review:

- **Suggestion (dismissed)**: The redundant pre-layout `_validate_writer_spatial_shape` call in `_writer.write` / `_writer.write_streaming` is kept on purpose. The spatial-shape validator is called at every writer entry point; dropping it at only two of them would create an asymmetric pattern that is more confusing than the cosmetic redundancy. The two function calls execute negligible work and provide defense-in-depth if the post-layout helper is ever refactored.
- **Nit (fixed in 93e71df)**: Error hint reworded so the `data.isel(band=slice(0, 0))` example reads as the likely cause to inspect upstream, not as a fix suggestion.

No new findings on this pass. Tests still pass (`pytest xrspatial/geotiff/tests/test_to_geotiff_zero_band_2094.py -v` -- 9 passed).

PR #2099, opened 2026-05-19
geotiff: fix GPU MinIsWhite guard axis for band-first DataArrays (#2097)

### LINKED ISSUE #2097: geotiff: GPU MinIsWhite guard uses wrong axis for band-first DataArrays
**Describe the bug**

`write_geotiff_gpu` rejects `photometric='miniswhite'` only for single-band inputs (the CPU writer pre-inverts pixels and the nodata sentinel for round-trip safety per #1836; the GPU writer does not implement that inversion). The samples count the guard uses is computed from the wrong axis for band-first DataArrays, so the guard misfires.

`xrspatial/geotiff/_writers/gpu.py:255` runs:

```python
_gpu_samples_hint = (data.shape[2] if hasattr(data, 'shape')
                     and data.ndim == 3 else 1)
```

This runs **before** the band-first to band-last remap at `xrspatial/geotiff/_writers/gpu.py:393`. For a single-band DataArray with dims `('band', 'y', 'x')` and shape `(1, H, W)`, `data.shape[2]` is `W`, not `1`. The guard at line 259 (`if _gpu_resolved_photo == 0 and _gpu_samples_hint == 1`) only fires when `W` happens to equal 1, so a real single-band MinIsWhite request can sneak through.

**Effect**

The GPU writer assembles the TIFF without applying the writer-side pixel + nodata-sentinel inversion that the reader expects for MinIsWhite single-band data (#1836). The on-disk bytes do not round-trip through `open_geotiff`.

**Expected behavior**

The samples_hint should use the same band-axis logic the rest of the writer uses:

- if `data.ndim == 3` and `data.dims[0] in _BAND_DIM_NAMES`, bands are at `data.shape[0]`
- otherwise (band-last 3D), bands are at `data.shape[2]`
- 2D inputs have 1 band

Then re-check the MinIsWhite guard. A band-first `('band', 'y', 'x')` shape `(1, H, W)` DataArray with `photometric='miniswhite'` should raise `NotImplementedError` on the GPU writer regardless of `W`.

**Scope of fix**

- Replace the `data.shape[2]`-based sample count with band-aware logic at the GPU writer entry point in `xrspatial/geotiff/_writers/gpu.py`.
- Add a regression test: band-first single-band DataArray + GPU writer + `photometric='miniswhite'` raises `NotImplementedError`. Skip without cupy.

brendancol:
## Summary
- The GPU writer's MinIsWhite single-band rejection (#1836) read samples-per-pixel from `data.shape[2]` before the band-first to band-last remap. For a band-first DataArray of shape `(1, H, W)` with dims `('band', 'y', 'x')`, the guard saw `samples_hint == W`, so the `== 1` check only fired when `W` happened to be 1. Real single-band MinIsWhite requests on band-first inputs slipped past and the on-disk bytes were not the writer-side-inverted values the reader expects.
- Compute the samples_hint using the same band-axis logic the remap itself uses: `shape[0]` for band-first 3D, `shape[2]` for band-last 3D, `1` for 2D. The MinIsWhite guard then fires on every single-band layout.
- Test the corrected guard on three layouts (band-first 3D, band-last 3D, 2D) plus a no-CUDA-required unit test that pins the samples-hint logic directly so a future refactor cannot silently regress it.

## Backend coverage
- cupy: GPU tests included; they exercise the guard before any CUDA dispatch, so they run on CI even without a GPU runner present
- numpy / dask: not affected (CPU writer already pre-inverts MinIsWhite per #1836)
- dask + cupy: same entry point as cupy; the guard runs before any GPU work

## Test plan
- [x] `pytest xrspatial/geotiff/tests/test_gpu_miniswhite_band_first_2097.py`
- [x] `pytest xrspatial/geotiff/tests/test_miniswhite_writer_roundtrip_1836.py` (regression)
- [x] `pytest xrspatial/geotiff/tests/test_miniswhite_nodata_1809.py` (regression)

Closes #2097


brendancol (commented):
## PR Review: geotiff: fix GPU MinIsWhite guard axis for band-first DataArrays (#2097)

### Blockers
None.

### Suggestions
- [ ] `xrspatial/geotiff/tests/test_gpu_miniswhite_band_first_2097.py:88-117` (`test_samples_hint_band_first_without_gpu`): the test redefines the samples-hint logic inline rather than importing the production code. If someone changes the production code in `_writers/gpu.py` and forgets to update this test, the test still passes because it is testing its own copy of the logic, not the writer. Either extract the samples-hint computation into a small named helper in `gpu.py` (e.g. `_gpu_samples_hint(data)`) and call it from this test, or have this test patch `_resolve_photometric` and call `write_geotiff_gpu` so the production path actually runs. Without that link the no-GPU unit test is reassurance theater.

### Nits
- [ ] `xrspatial/geotiff/_writers/gpu.py:263-272`: the `getattr(data, 'ndim', None) == 3` and `getattr(data, 'dims', None)` guards are defensive. The rest of `write_geotiff_gpu` already assumes `data` is a DataArray with `.dims` and `.shape`, so these guards are dead code in practice. Matches the surrounding block's style though, and dropping them would be its own diff.

### What looks good
- Samples-hint computation now mirrors the band-axis logic the `moveaxis` at `gpu.py:393` uses, so the guard and the actual on-device band placement stay in sync.
- Coverage hits all three layouts: band-first 3D, band-last 3D, and 2D. The standalone logic test is meant to run without a GPU so CI catches future regressions even on no-GPU runners (modulo the suggestion above).
- No regression on `test_miniswhite_writer_roundtrip_1836.py` or `test_miniswhite_nodata_1809.py`.
- Validator runs before any GPU dispatch, so a rejected request consumes no device memory.

### Checklist
- [x] Algorithm matches reference/paper — N/A (input rejection)
- [x] Dask chunk boundaries handled correctly — N/A (no dask path changed)
- [x] Benchmark exists or is not needed — not needed (input rejection)
- [x] README feature matrix updated (if applicable) — N/A

brendancol (commented):
## Follow-up: review findings disposition

- **Suggestion** (test redefined production logic): **Fixed.** Extracted the samples-hint logic into a module-level `_compute_gpu_samples_hint` helper in `gpu.py`. `write_geotiff_gpu` now calls it, and `test_samples_hint_band_first_without_gpu` calls the same helper directly via a small `_FakeArray` stand-in. A regression that moves the lookup back to a blind `data.shape[2]` will now fail on CI even without a CUDA device.
- **Nit** (defensive `getattr` guards): **Deferred.** Dropping the guards would be its own diff and matches the surrounding block's style; not worth bundling into this PR.

Tests re-run: `test_gpu_miniswhite_band_first_2097.py` (4/4), `test_miniswhite_writer_roundtrip_1836.py` (10/10), `test_miniswhite_nodata_1809.py` (10/10).

Issue #1885, opened 2026-05-14
geotiff: extract read_geotiff_gpu into _backends/gpu.py (#1813 step 7)


brendancol:
## Reason

`read_geotiff_gpu` (`xrspatial/geotiff/__init__.py` lines 2723–3338) is the largest single function in the file at about 615 lines. With GPU helpers (#1813 step 6), validators (step 4), attrs helpers (step 5), and sentinels (step 2) already extracted, the body is now a near-mechanical lift into its own backend module.

## Proposal

Move `read_geotiff_gpu` to `xrspatial/geotiff/_backends/gpu.py`. Re-export from `__init__.py` so `from xrspatial.geotiff import read_geotiff_gpu` is unchanged. `open_geotiff`'s GPU-branch dispatch continues to work via the same re-import path.

## Drawbacks

Largest extraction in the sequence; reviewers should verify the body is unchanged except for import statements. The parity matrix in #1813 step 1 backstops behavioural regressions.

Refs: #1813.

PR #2103, opened 2026-05-19
geotiff tests: fix stale patch target and pre-multiple-of-16 tile_size

### LINKED ISSUE #2101: geotiff tests: 8 failures from stale patch target and pre-multiple-of-16 tile_size value
**Describe the bug**

Running `pytest xrspatial/geotiff/tests` produces 8 failures, all in two files:

`test_predictor2_big_endian_gpu_1517.py` (7 failures):

```
AttributeError: <module 'xrspatial.geotiff' ...> has no attribute 'read_to_array'
```

`_block_cpu_fallback` does `monkeypatch.setattr(geotiff_pkg, 'read_to_array', _no_fallback, raising=True)`. After PR #1885 moved `read_geotiff_gpu` into `xrspatial/geotiff/_backends/gpu.py`, the package no longer exposes `read_to_array` at that name — `__init__.py` imports it as `_read_to_array`, and the actual call site is the `_read_to_array` symbol inside `_backends/gpu.py`. The monkeypatch with `raising=True` therefore aborts the test before the GPU path runs.

`test_size_param_validation_gpu_vrt_1776.py::test_tile_size_positive_works`:

```
ValueError: tile_size must be a positive multiple of 16 (TIFF 6 spec
requirement for TileWidth/TileLength), got tile_size=4; try tile_size=16.
```

The test passes `tile_size=4`. The validator now requires multiples of 16 (a later change), so the original value is no longer valid. The test's intent (small but valid tile size still works) is fine; only the literal needs updating.

**Expected behavior**

`pytest xrspatial/geotiff/tests` should run clean.

**Fix**

- Patch the call site `xrspatial.geotiff._backends.gpu._read_to_array` instead of the no-longer-existing `xrspatial.geotiff.read_to_array`.
- Change the `tile_size=4` assertion to `tile_size=16`, the smallest currently valid value, and update the docstring accordingly.

**Additional context**

CPU and dask tests are unaffected. The failures appear only on GPU-enabled environments (these tests are gated on cupy + CUDA + tifffile).

brendancol:
Closes #2101.

## Summary

Two unrelated stale-test bugs caused 8 failures in `pytest xrspatial/geotiff/tests` on GPU-enabled environments:

- `test_predictor2_big_endian_gpu_1517.py::_block_cpu_fallback` patched `xrspatial.geotiff.read_to_array`, a name that disappeared when `read_geotiff_gpu` was moved into `xrspatial/geotiff/_backends/gpu.py` (PR #1885). The package imports the symbol as `_read_to_array`, and the actual call site is the bound `_read_to_array` inside `_backends.gpu`. With `raising=True`, the monkeypatch aborted before the GPU path ran. Patch the call site instead. (7 failures.)
- `test_size_param_validation_gpu_vrt_1776.py::test_tile_size_positive_works` passed `tile_size=4`. The validator later grew a multiple-of-16 requirement (TIFF 6 spec), so 4 is no longer accepted. Use `tile_size=16`, the smallest valid value, preserving the test's intent. (1 failure.)

Backend coverage: tests-only change. No production code touched. The patched assertion still exercises numpy / cupy / dask-cupy paths exactly as before.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_predictor2_big_endian_gpu_1517.py xrspatial/geotiff/tests/test_size_param_validation_gpu_vrt_1776.py` -- 56 passed.
- [x] `pytest xrspatial/geotiff/tests` -- no failures.

### DIFF
--- a/xrspatial/geotiff/tests/test_predictor2_big_endian_gpu_1517.py
+++ b/xrspatial/geotiff/tests/test_predictor2_big_endian_gpu_1517.py
@@ -46,14 +46,16 @@ def _block_cpu_fallback(monkeypatch):
     ``read_geotiff_gpu`` returns a cupy-backed array even when its silent CPU
     fallback fires (the fallback wraps the CPU result with ``cupy.asarray``),
     so ``isinstance(gpu_da.data, cupy.ndarray)`` cannot distinguish the two
-    paths. Patching the module-bound ``read_to_array`` to raise turns any
-    silent fallback into a hard test failure, which is what we want when the
-    point of a test is to exercise the actual GPU decode kernels.
+    paths. ``read_geotiff_gpu`` lives in ``xrspatial.geotiff._backends.gpu``
+    and calls the locally bound ``_read_to_array`` symbol there; patching
+    that binding to raise turns any silent fallback into a hard test failure,
+    which is what we want when the point of a test is to exercise the actual
+    GPU decode kernels.
 
     Tests that legitimately rely on the CPU fallback (e.g. stripped
     layouts) must not call this helper.
     """
-    from xrspatial import geotiff as geotiff_pkg
+    from xrspatial.geotiff._backends import gpu as gpu_backend
 
     def _no_fallback(*args, **kwargs):
         raise AssertionError(
@@ -62,7 +64,7 @@ def _block_cpu_fallback(monkeypatch):
         )
 
     monkeypatch.setattr(
-        geotiff_pkg, 'read_to_array', _no_fallback, raising=True,
+        gpu_backend, '_read_to_array', _no_fallback, raising=True,
     )
 
 
--- a/xrspatial/geotiff/tests/test_size_param_validation_gpu_vrt_1776.py
+++ b/xrspatial/geotiff/tests/test_size_param_validation_gpu_vrt_1776.py
@@ -113,9 +113,10 @@ class TestWriteGeotiffGpuTileSize:
             write_geotiff_gpu(gpu_da, out, tile_size=False)
 
     def test_tile_size_positive_works(self, gpu_da, tmp_path):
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: geotiff tests: fix stale patch target and pre-multiple-of-16 tile_size

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good

- `_block_cpu_fallback` now patches `xrspatial.geotiff._backends.gpu._read_to_array`, which is the symbol the four CPU-fallback call sites in `_backends/gpu.py` (lines 391, 588, 601, 658) actually reach for. That makes the helper effective at catching any silent fallback inside `read_geotiff_gpu`. The previous target (`xrspatial.geotiff.read_to_array`) never existed at that name after PR #1885, so the helper was a no-op that aborted with `AttributeError` before the GPU path ran.
- The CPU baseline calls in the affected tests (`from xrspatial.geotiff._reader import read_to_array`) keep working: they go through `_reader.read_to_array`, which the new patch leaves alone, and they run before `_block_cpu_fallback` anyway.
- `tile_size=16` is the smallest value that passes the current validator (`tile_size % 16 == 0`, positive). The updated docstring matches.

### Checklist

- [x] Algorithm matches reference/paper — N/A (tests-only).
- [x] All implemented backends produce consistent results — N/A (test fix, no production code path changed).
- [x] NaN handling is correct — N/A.
- [x] Edge cases are covered by tests — preserved (`tile_size=0`, negative, float, bool True/False, np.int64 still tested).
- [x] Dask chunk boundaries handled correctly — N/A.
- [x] No premature materialization or unnecessary copies — N/A.
- [x] Benchmark exists or is not needed — not needed.
- [x] Docstrings present and accurate — `_block_cpu_fallback` and `test_tile_size_positive_works` docstrings updated to match the new behaviour.

PR #2108, opened 2026-05-19
geotiff: annotate open_geotiff(max_cloud_bytes=...) (#2106)

### LINKED ISSUE #2106: geotiff: open_geotiff(max_cloud_bytes=...) missing type annotation
Found by deep-sweep api-consistency on 2026-05-18.

## Issue

``xrspatial.geotiff.open_geotiff`` is the only kwarg on the public signature without a Python type annotation. Every other kwarg on the function (and every kwarg on every other public reader/writer in the module) has one.

```python
>>> import inspect, xrspatial.geotiff as g
>>> sig = inspect.signature(g.open_geotiff)
>>> for n, p in sig.parameters.items():
...     ann = "NO-HINT" if p.annotation is inspect.Parameter.empty else str(p.annotation)
...     print(f"{n}: {ann}")
source: str | BinaryIO
dtype: str | np.dtype | None
window: tuple | None
overview_level: int | None
band: int | None
name: str | None
chunks: int | tuple | None
gpu: bool
max_pixels: int | None
max_cloud_bytes: NO-HINT      <-- only kwarg without an annotation
on_gpu_failure: str
missing_sources: str
allow_rotated: bool
allow_unparseable_crs: bool
band_nodata: str | None
mask_nodata: bool
```

The docstring already declares the type:

> max_cloud_bytes : int or None, optional

so the surface and the docs disagree. ``inspect.signature``, IDE autocomplete, Sphinx output, and ``mypy --strict`` all see a bare parameter for the only fsspec-related kwarg on the public read entry point.

The default is the module-internal ``_MAX_CLOUD_BYTES_SENTINEL`` (used so the function can distinguish "caller passed nothing" from "caller passed ``None``"); the annotation should still be ``int | None`` to match the docstring and what callers actually pass.

## Cat / Severity

API Consistency Cat 3 (type hints and docstrings), severity MEDIUM. Type annotation is missing on a public function while sibling parameters in the same function -- and every other public read/write entry point in the module -- have them.

## Fix

Add ``int | None`` annotation to ``max_cloud_bytes`` on ``open_geotiff``. No behaviour change; the sentinel default is preserved. Regression test imports ``open_geotiff`` and asserts ``inspect.signature(...)``'s annotation on the kwarg is non-empty so the gap cannot reopen.

## Out of scope

- Renaming or default change. Default stays the sentinel; semantics unchanged.
- Other geotiff entry points: every other kwarg on every other public reader/writer already has an annotation, so no follow-up rows to address.

brendancol:
Closes #2106.

## Summary
- ``open_geotiff(max_cloud_bytes=...)`` was the only kwarg on the public read/write surface without a Python type annotation; the docstring already declared ``int or None`` so the surface and the docs disagreed
- Adds ``int | None`` to the annotation. Default stays the module-internal ``_MAX_CLOUD_BYTES_SENTINEL`` so behaviour is unchanged
- Regression test pins the immediate gap and parametrises over every public reader/writer entry point (``open_geotiff``, ``read_geotiff_gpu``, ``read_geotiff_dask``, ``read_vrt``, ``to_geotiff``, ``write_geotiff_gpu``, ``write_vrt``) so a future addition without an annotation surfaces in CI

## Why this matters
- ``inspect.signature`` is the source of truth for IDE autocomplete, Sphinx output, and ``mypy --strict``. A bare parameter blocks the static-typing path for the only fsspec-related kwarg on the public read entry point
- Cat 3 (Type hints and docstrings), severity MEDIUM under the api-consistency sweep template. Not a behaviour change

## Out of scope
- No rename. Sentinel default and validation rules unchanged
- Other geotiff entry points already had annotations on every kwarg, so no follow-up rows to address

## Test plan
- [x] ``pytest xrspatial/geotiff/tests/test_open_geotiff_max_cloud_bytes_annot_2106.py -v`` (8 passed)
- [x] ``pytest xrspatial/geotiff/tests/test_reader_kwarg_order_1935.py xrspatial/geotiff/tests/test_writer_kwarg_order_1922.py -v`` (existing parity tests still pass)
- [x] Local smoke test: ``open_geotiff`` round-trip with ``max_cloud_bytes=...``, ``max_cloud_bytes=None``, and the existing ``gpu=True + max_cloud_bytes=...`` rejection guard all still work
- [x] CUDA smoke test on a GPU host (``read_geotiff_gpu`` returns CuPy-backed DataArray, gpu+max_cloud_bytes rejection still fires)

Filed by deep-sweep api-consistency on 2026-05-18.


brendancol:
## PR Review: geotiff: annotate open_geotiff(max_cloud_bytes=...) (#2106)

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/geotiff/__init__.py:252`: the annotation is `int | None`, but the default `_MAX_CLOUD_BYTES_SENTINEL` is an `object()`. `mypy --strict` flags `Incompatible default for argument` because of that mismatch. The same pattern already exists a few lines down for `on_gpu_failure: str = _ON_GPU_FAILURE_SENTINEL` and `missing_sources: str = _MISSING_SOURCES_SENTINEL`, so the new line is consistent with what's already there. A typed sentinel class would clear mypy fully, but that's a bigger refactor and not in scope.
- `test_open_geotiff_max_cloud_bytes_annot_2106.py:46`: the assertion checks that the annotation string contains both `"int"` and `"None"` as substrings. Something like `Mapping[int, None]` would pass while not being the intended type. The parametrized test below catches the more interesting failure mode (missing annotation), so this is fine as a pin.

### What looks good
- Scope is tight: one annotation, one regression test file.
- The test does two useful things: pins the named kwarg, and parametrizes over every public reader/writer so the next missing annotation surfaces in CI rather than IDE autocomplete.
- Lazy-eval annotations are handled correctly. The file uses `from __future__ import annotations`, so the runtime annotation is the string `'int | None'`, and the test's `str(...)` wrapping is a no-op for that case while still working if the import is ever removed.
- Docstring at `__init__.py:299` already declared `int or None`; the signature now matches.
- Verified locally: the 8 new tests pass, and the existing kwarg-order parity tests in `test_reader_kwarg_order_1935.py` and `test_writer_kwarg_order_1922.py` still pass. Every other public entry point already had all kwargs annotated, so the new parametrized assertion is green on day one and won't fire false positives.

### Checklist
- [x] Algorithm matches reference/paper: N/A, signature-only change
- [x] All implemented backends produce consistent results: N/A
- [x] NaN handling is correct: N/A
- [x] Benchmark exists or is not needed: not needed
- [x] README feature matrix updated: N/A

PR #2111, opened 2026-05-19
geotiff: use ceil semantics for COG overview block reduction

### LINKED ISSUE #2105: COG overviews drop edge pixels for odd-sized rasters
**Describe the bug**

`_block_reduce_2d` in `xrspatial/geotiff/_writer.py:460` floors the input dimensions to an even multiple and crops before reducing:

```python
h, w = arr2d.shape
h2 = (h // 2) * 2
w2 = (w // 2) * 2
cropped = arr2d[:h2, :w2]
```

The GPU mirror at `xrspatial/geotiff/_gpu_decode.py:3102` does the same thing. For a 5x5 input the crop produces a 4x4 buffer, which then halves to a 2x2 overview. GDAL's overview generator uses ceil semantics (5x5 -> 3x3) so that every base pixel still contributes to the overview.

The current floor-and-crop approach silently drops the last row and last column on any odd-sized input, which corrupts overview pixels along the trailing edges and means downstream reads at `overview_level >= 1` see partial coverage that does not match the source extent.

**Expected behavior**

Block reduction should use ceil semantics. A 5x5 input should produce a 3x3 overview where the trailing row/col reduce a 1x2, 2x1, or 1x1 residual block rather than being dropped. CPU and GPU paths should agree, both for COG writing and for any read-time overview generation that uses the same helper.

**Additional context**

- Affects all resampling methods (`nearest`, `mean`, `median`, `min`, `max`, `mode`, `cubic`).
- Found during a code review pass over the geotiff module.
- Matches the reference behaviour of GDAL `gdaladdo` and rasterio overview builders, which both round up.

brendancol:
Closes #2105.

## Summary

- `_block_reduce_2d` (CPU) and `_block_reduce_2d_gpu` (GPU) used to floor the input dims to even and crop, dropping the trailing row/col on odd-sized rasters. A 5x5 became a 2x2 overview rather than the expected 3x3 ceil shape.
- Switched both helpers to GDAL ceil semantics. Odd-sized inputs are padded on the trailing edge with NaN (float-promoted for integer dtypes) so the 2x2 block reshape works at every input size and the nan-aware aggregations naturally exclude the padded cells from the residual block.
- Cubic uses edge-replicate padding so the spline kernel sees a valid neighbourhood at the trailing edge. Mode uses NaN-padding with the count tie-break preserved (lowest valid wins).
- Even-sized inputs keep the existing fast path with no padding overhead.

## Backend coverage

- numpy: covered.
- cupy: covered.
- dask+numpy: writer overview path is single-tile, so no dask-specific change needed.
- dask+cupy: same.

## Test plan

- [x] New `test_cog_overview_ceil_2105.py` pins output shape per method and dtype across 5x5, 5x4, 4x5, 7x3, 1x1, etc., and verifies the trailing pixels actually reach the overview.
- [x] Sentinel masking still works on odd-sized inputs (mean / min / max with `nodata`).
- [x] GPU mirror produces the same values as CPU on odd-sized inputs.
- [x] Existing nodata sweep (`test_cog_overview_nodata_1613`, `test_cog_cubic_*`, `test_cog_int_overview_nodata_*`) still passes.
- [x] `test_mode_overview_perf` reference was floor-and-crop; updated the reference to ceil and removed the dead skip path for 1x1.
- [x] Full `xrspatial/geotiff/tests/` suite: 4129 passed, 25 skipped.

### DIFF
--- a/xrspatial/geotiff/_gpu_decode.py
+++ b/xrspatial/geotiff/_gpu_decode.py
@@ -3113,7 +3113,9 @@ def _block_reduce_2d_gpu(arr2d, method, nodata=None):
 
     if method == 'nearest':
         # Top-left pixel of each 2x2 block; direct stride keeps the
-        # trailing row/col for odd-sized inputs (issue #2105).
+        # trailing row/col for odd-sized inputs (issue #2105). The
+        # ``.copy()`` materialises a contiguous device buffer so
+        # downstream nodata-mask rewrites don't touch ``arr2d``.
         return arr2d[::2, ::2].copy()
 
     if method == 'mode':
@@ -3138,10 +3140,13 @@ def _block_reduce_2d_gpu(arr2d, method, nodata=None):
 
     # Block reshape for mean/min/max/median. Odd-sized inputs get a
     # NaN-padded trailing edge so the residual block reduces correctly;
-    # the nan-aware reductions exclude the padded cells.
+    # the nan-aware reductions exclude the padded cells. The padded
+    # buffer keeps the source float dtype (float32 stays float32) so an
+    # odd-shape GPU read does not pay a 2x device-memory cost for a
+    # silent promote to float64.
     if arr2d.dtype.kind == 'f':
         if need_pad:
-            padded = cupy.full((h2, w2), cupy.float64('nan'), dtype=cupy.float64)
+            padded = cupy.full((h2, w2), np.nan, dtype=arr2d.dtype)
             padded[:h, :w] = arr2d
             blocks = padded.reshape(oh, 2, ow, 2)
         else:
--- a/xrspatial/geotiff/_gpu_decode.py
+++ b/xrspatial/geotiff/_gpu_decode.py
@@ -3181,11 +3181,14 @@ def _block_reduce_2d_gpu(arr2d, method, nodata=None):
             if info.min <= nodata_int <= info.max:
                 sentinel = np.dtype(str(arr2d.dtype)).type(nodata_int)
                 if need_pad:
-                    # The float-padded view holds the original int values
-                    # exactly representable in float64; compare against
-                    # ``float(sentinel)`` to find sentinel cells in the
-                    # valid region (the padded edge is already NaN).
-                    mask = blocks == float(sentinel)
+                    # Compute the integer-width sentinel mask before
+                    # padding, so a 64-bit sentinel near INT64_MAX (not
+                    # exactly representable in float64) still matches.
+                    # The padded edge already holds NaN in ``blocks`` and
+                    # stays False here so the where below leaves it alone.
+                    int_mask = cupy.zeros((h2, w2), dtype=bool)
+                    int_mask[:h, :w] = arr2d == sentinel
+                    mask = int_mask.reshape(oh, 2, ow, 2)
                 else:
                     int_blocks = arr2d.reshape(oh, 2, ow, 2)
                     mask = int_blocks == sentinel
--- a/xrspatial/geotiff/_gpu_decode.py
+++ b/xrspatial/geotiff/_gpu_decode.py
@@ -3090,26 +3090,35 @@ GPU_OVERVIEW_METHODS = ('mean', 'nearest', 'min', 'max', 'median', 'mode',
 def _block_reduce_2d_gpu(arr2d, method, nodata=None):
     """2x block-reduce a single 2D CuPy plane using *method*.
 
+    Output dimensions follow GDAL's ceil semantics (5x5 -> 3x3) so the
+    overview pyramid covers the full source extent for odd-sized rasters.
+    Odd inputs are NaN-padded along the trailing edge (float-promoted for
+    integer dtypes) so the 2x2 block reshape works and the residual block
+    is reduced via the same nan-aware aggregations (issue #2105). Mirrors
+    the CPU helper :func:`xrspatial.geotiff._writer._block_reduce_2d` so
+    the two backends produce identical overviews.
+
     When ``nodata`` is supplied and ``arr2d`` is a float dtype, cells that
     equal the sentinel are masked back to NaN before the reduction so the
-    ``cupy.nan*`` aggregation routines correctly skip them. Mirrors the
-    CPU helper :func:`xrspatial.geotiff._writer._block_reduce_2d` so the
-    two backends produce identical overviews when ``nodata`` is set.
+    ``cupy.nan*`` aggregation routines correctly skip them.
     """
     import cupy
     import numpy as np
 
     h, w = arr2d.shape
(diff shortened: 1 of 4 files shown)


brendancol (commented):
Self-review of the four fixes spawned by the rockout pass.

**Coverage**
- CPU and GPU `_block_reduce_2d` both swap to ceil dims and pad the trailing edge with NaN; nearest strides directly; cubic uses edge-replicate so the spline kernel sees valid edge values.
- Even-sized inputs keep the existing fast path with no padding overhead.
- All resampling methods (nearest, mean, min, max, median, mode, cubic) are covered for both float and integer dtypes, with sentinel masking still applied on odd-sized inputs.

- The integer sentinel comparison in the padded branch happens against `float(sentinel)`. Safe for the integer ranges `_resolve_int_nodata` accepts, but if a future change widens that range the equality could lose precision.
- Mode tie-break: NaN cells get count 0 because `NaN == NaN` is False. argmax then picks the first valid cell, and numpy sort puts NaN at the end so "lowest valid wins" still holds. Worth a second eye.
- `test_mode_overview_perf` had a reference that hard-coded the floor-and-crop behaviour; updated the reference to ceil semantics rather than skipping the diverging shapes.

**Not in scope here**
- Sidecar `.ovr` discovery (covered in #2114).
- Multi-tiepoint validation (covered in #2118).
- Rotated GeoTIFF read opt-out (covered in #2116).

brendancol (commented):
Follow-up to my self-review.

**Int64 sentinel concern -> fixed.**
Verified the precision-loss case is real: `float(INT64_MAX)` rounds up to `2**63`, so the padded branch's `blocks == float(sentinel)` would miss the sentinel for int64 / uint64 rasters near the dtype max. Switched both CPU and GPU to compute the equality mask on the original integer array at native width and pad with `False` so the broadcast onto the float-promoted blocks view still works. Added a regression test using `np.iinfo(np.int64).max` as the sentinel on an odd-sized raster. New commit: 761f75d.

**Mode tie-break with NaN -> verified, no change needed.**
Re-confirmed by working the cases:
- Block `[5, NaN, NaN, NaN]` after sort = `[5, NaN, NaN, NaN]`. Counts: `[1, 0, 0, 0]`. argmax picks index 0 -> 5.
- Block `[5, 5, 7, NaN]` after sort = `[5, 5, 7, NaN]`. Counts: `[2, 2, 1, 0]`. argmax picks index 0 -> 5 (lowest-valid wins).
- Block `[5, NaN, NaN, NaN]` cannot become all-NaN because every output block overlaps at least one valid source cell by construction (residual blocks at the trailing edge are 1x1, 1x2, or 2x1).

`numpy.argmax` returns the first maximum, so the NaN cells (count 0) lose to any valid cell (count >= 1) and the sorted-ascending order delivers the smallest valid value on ties.

**`test_mode_overview_perf` reference**
Updated to ceil semantics rather than skipping odd shapes, as called out. No further action.

brendancol (commented):
Re-review with suggestions and nits.

- **Float32 input is silently promoted to float64 in the padded branch.** `_writer.py:621` and `_gpu_decode.py:3144` use `np.full((h2, w2), nan, dtype=np.float64)` even when `arr2d.dtype` is float32. Odd-shape float32 inputs then pay 2x intermediate memory for the padded buffer before casting back at the end. The even-shape path keeps the source dtype. Suggest `dtype=arr2d.dtype if arr2d.dtype.kind == 'f' else np.float64` so float32 stays float32 across the padded reduction.

- **uint64 sentinel near `UINT64_MAX` is not in the regression test.** `test_int64_sentinel_near_max_masks_in_padded_branch` covers int64 but `_resolve_int_nodata` also accepts uint64 in range, and `float(2**64 - 1)` rounds to `2**64`. Worth a parallel test using `np.iinfo(np.uint64).max` so both 64-bit dtypes are pinned.

- **`_replicate_pad` is redefined on every cubic call.** `_writer.py:522` puts the helper inside `_block_reduce_2d`, which closes over `h, w, h2, w2, need_pad`. Hoisting it to module level and passing those as args would let any future scalar-pad path share the same code; minor on a non-hot path.

### Nits (optional)

- **`return arr2d[::2, ::2].copy()`** at `_writer.py:484`. The `.copy()` preserves the previous contract (caller may write to the result) but is now defensive against a view that won't be mutated. Either keep with a short note or drop the copy.

- **Comment drift in the int branch.** `_writer.py:651-685` still narrates the pre-fix `cropped.astype(np.float64).reshape(...)` flow even though the padded branch now uses an explicit int-width mask. The behaviour is right; the comment block reads as if the float-cast still happens before the mask.

- **`cubic` mask-empty fallthrough is implicit.** `_writer.py:544` short-circuits with `if mask.any():` and falls through to the no-nodata cubic path at line 575. Correct but the control flow is easier to follow if the early return is explicit, e.g. `else: pass` or a `# no sentinel pixels present; fall through to default cubic` marker.

### What looks good

- Ceil semantics applied uniformly across CPU + GPU + all six reductions; even-shape fast path preserved (no padding overhead).
- INT64 sentinel fix (commit 761f75d) verified via the new regression test.
- Mode tie-break walkthrough confirmed: NaN cells count 0 lose to any valid cell, sort puts NaN last so "lowest valid wins" holds.
- `test_mode_overview_perf` reference updated rather than skipped on odd shapes; that keeps the per-pixel comparator honest.

brendancol (commented):
Addressed the suggestions and nits from the re-review (commit 293dc83).

| Item | Disposition |
|---|---|
| Float32 promotion in padded branch | Fixed -- `_writer.py` and `_gpu_decode.py` now use `arr2d.dtype` for the NaN-padded buffer; float32 stays float32 across the pad. |
| uint64 sentinel near `UINT64_MAX` not in tests | Added `test_uint64_sentinel_near_max_masks_in_padded_branch` (parallel of the int64 regression). |
| `_replicate_pad` redefined per cubic call | Hoisted to module-level `_replicate_pad_2d(src, h2, w2)`. The cubic path now reuses the helper across all branches. |
| `.copy()` on the nearest branch | Kept with an inline note: downstream nodata-mask rewrites mutate the result, so the strided view needs to be materialised first. Mirror note on the GPU path. |
| Int branch comment drift | Reworded to describe the native-width mask the code currently uses (prior text still narrated the pre-fix float-cast flow). |
| Cubic mask-empty fall-through | Two explicit fall-through comments (float and integer nodata branches) so the control flow into the default cubic path is no longer implicit. |

Added one extra test (`test_float32_padded_branch_keeps_source_dtype`) to pin the dtype-preservation contract end-to-end, since the float32 path's correctness now depends on it.

Full `xrspatial/geotiff/tests/` suite: 4132 passed, 25 skipped.

PR #2113, opened 2026-05-19
geotiff: apply per-tile byte cap on GPU read path (Cat 1 MEDIUM)

brendancol:
## Summary

- Security sweep pass 18 identified an asymmetry between the CPU and GPU read paths: the CPU paths (`_read_tiles`, `_fetch_decode_cog_http_tiles`) enforce the `_max_tile_bytes_from_env()` per-tile cap (default 256 MiB) but `read_geotiff_gpu` did not. `validate_tile_layout` only bounds the offsets array length, leaving `TileByteCounts` entries unchecked.
- A crafted local TIFF with multi-hundred-MB `TileByteCount` values could pass through the eager GPU decode path; `_check_gpu_memory` would only catch the aggregate sum at ~90% of free VRAM, missing the per-tile budget the CPU paths refuse at 256 MiB.
- Add the per-tile cap loop to `read_geotiff_gpu` (after `validate_tile_layout`) so a single oversized tile is rejected before any decode work runs. The error wording matches the CPU guards.
- Update `.claude/sweep-security-state.csv` to record the fix.

## Test plan

- [x] Add `xrspatial/geotiff/tests/test_gpu_tile_byte_cap_2026_05_18.py` covering the rejection, the error message wording, the `XRSPATIAL_COG_MAX_TILE_BYTES` env override, and the legit-read pass-through under the default cap. Tests reuse the forged-TIFF helpers from `test_local_tile_byte_cap_1664.py`.
- [x] Run the new GPU test suite plus the CPU `test_local_tile_byte_cap_1664.py` and `test_security.py` on a GPU host: 59 passing, including 4 new GPU coverage tests.
- [x] Smoke test the broader GPU read suite (`test_gpu_nodata_1542.py`, `test_gpu_window_band_1605.py`, `test_gpu_byteswap_1508.py`): 24 passing, no regressions.


brendancol (commented):
## PR Review: per-tile byte cap on the GPU read path

- **Cap is missing on the dask+GPU chunked path.** `_read_geotiff_gpu_chunked_gds` at `xrspatial/geotiff/_backends/gpu.py:1078` calls `validate_tile_layout(ifd)` and proceeds to `_decode_window_gpu_direct` per chunk without checking each `byte_counts` entry against `_max_tile_bytes_from_env()`. The same forged TIFF this PR rejects on the eager path slips past when the caller uses `read_geotiff_gpu(path, chunks=N)`. Land the cap at metadata-parse time inside `_read_geotiff_gpu_chunked_gds` so the dask graph never builds for a hostile file.

- **Sparse tile pass-through is implicit.** The loop at `_backends/gpu.py:503-513` iterates every entry in `byte_counts` including the zeros marking sparse tiles. Sparse tiles pass under any positive cap and the CPU mirror at `_reader.py:2085-2094` behaves the same way; both call sites would benefit from a one-line comment so a future reader does not add a `bc > 0` guard and accidentally start rejecting sparse files.

- **`test_env_override_lifts_cap` swallows every non-ValueError with `except Exception: pass`.** That makes the test pass on any failure mode that is not the cap error, which is broader than needed. Tighten to the specific decoder exceptions the truncated payload would raise (e.g. the `_compression` module's errors), or assert that the cap message is *not* present in whatever exception fires.

- **Test surgery helper `_patch_byte_counts` is a near-copy of the CPU companion** (`test_local_tile_byte_cap_1664.py`). Hoist to `tests/conftest.py` or a shared helper module so the CPU and GPU caps stay byte-pinned to the same TIFF surgery. As-is, fixing a bug in one helper requires remembering to mirror it.

### Nits (optional)

- **`for _tile_idx, _bc in enumerate(byte_counts):`** at `_backends/gpu.py:503`. Both names are used inside the loop body, so the underscore prefix (idiomatic for unused locals) misleads. `tile_idx, bc` would match the loop's actual semantics.

- **`.claude/sweep-security-state.csv` row 21** drops the PR/issue number column to an empty string. If the column is consumed by sweep tooling that links findings to commits, the new entry looks unresolved.

- **`test_normal_gpu_read_under_default_cap`** uses a 64x64 raster split into 32x32 tiles -- well under any cap, which validates the negative case but does not exercise anything near the default 256 MiB ceiling. A second positive case with a larger fixture (or a parametrized cap close to the actual tile size) would strengthen the pass-through guarantee.

### What looks good

- Cap matches the CPU mirror exactly in wording, env var, and error format, so the operator-facing failure mode is consistent across backends.
- `validate_tile_layout` runs first, so length-bounds errors keep firing before the per-tile loop -- no spurious raise on a structurally malformed table.
- Test fixture builder rewrites a *real* COG that survives `parse_header` rather than synthesising raw bytes, which keeps the cap test exercising the same parse path production hits.

### Checklist

- [x] Algorithm matches reference: cap mirrors CPU strip/tile paths.
- [x] Backend coverage: eager GPU added; dask+GPU chunked path is the gap above.
- [x] Edge cases: sparse tiles pass through (implicit); huge tile fails as intended.
- [x] No premature materialization or copies.
- [x] Test cap-rejection + env-override behaviour.
- [ ] Same cap on `_read_geotiff_gpu_chunked_gds` -- still missing.

brendancol (commented):
Addressed the review (commit 6d68acd).

| Item | Disposition |
|---|---|
| Dask + GPU chunked path missing the cap | Fixed -- cap now lands at metadata-parse time inside `_read_geotiff_gpu_dask_path` (before the GDS qualification gate) and again inside `_read_geotiff_gpu_chunked_gds` when that function is reached directly. Hostile files are rejected at graph-build time, not at first `.compute()`. |
| Sparse tile pass-through implicit | Both cap-loop comment blocks (eager + chunked) now spell out that `byte_count == 0` passes by design. |
| `test_env_override_lifts_cap` swallowed every exception | Replaced the dual `except ValueError` / `except Exception` arms with a single `except Exception` that asserts the cap message is *not* present. A regression that re-fires the cap through a wrapped exception now produces a focused failure. |
| TIFF surgery helper duplicated | Hoisted to `xrspatial/geotiff/tests/_tiff_surgery.py`. Both the CPU and the GPU cap tests now import the same `patch_byte_counts`. The `struct` import drops from each test file. |
| Loop variables `_tile_idx, _bc` | Renamed to `tile_idx, bc` (both are read inside the body, so the underscore prefix was misleading). |
| Chunked-GPU regression test | Added `TestGpuChunkedTileByteCap::test_chunked_huge_tile_byte_count_rejected` covering the new cap on `chunks=...` reads. |

Sweep CSV column and the small-fixture nit on the legit-read test are stylistic; the CSV column is the same shape sweep tooling already accepts (other rows with no linked PR also leave the field empty), and the legit-read test stays at the small-fixture size because the goal is regression-on-the-no-op-path, not coverage near the cap.

Full `xrspatial/geotiff/tests/` suite: 4023 passed, 25 skipped.

PR #2124, opened 2026-05-19
geotiff: move no-georef signal off coord shape onto attrs marker

### LINKED ISSUE #2120: geotiff: to_geotiff silently strips georef on int64 step-1 user coords
**Describe the bug**

`to_geotiff` silently strips georef from any DataArray whose `x` and `y` coords are contiguous int64 step-1 arrays. `_is_no_georef_sentinel` (`xrspatial/geotiff/_coords.py:39-77`) accepts every ascending step-1 int64 coord array as the read-side no-georef placeholder, but real user grids match that exact shape.

**Reproducer**

```python

x = np.array([500, 501, 502], dtype=np.int64)
y = np.array([1000, 1001], dtype=np.int64)
da = xr.DataArray(np.zeros((2, 3), dtype=np.float32),
                  coords={'y': y, 'x': x}, dims=('y', 'x'),
                  attrs={'crs': 4326})

to_geotiff(da, '/tmp/repro.tif')
out = open_geotiff('/tmp/repro.tif')
print(out.x.values)  # -> [0, 1, 2]  georef silently lost
```

**Expected behaviour**

The output should round-trip with the original coords, or `to_geotiff` should raise so the user can supply `attrs['transform']` explicitly. Silent metadata loss in a foundational IO path is worse than a noisy error.

**Root cause**

`_is_no_georef_sentinel` was tightened in #2087 from "any integer dtype" to "exact arange pattern" to fix an earlier silent-strip bug. The new check is still purely shape-based. The docstring at lines 63-68 admits the trade-off: a user-authored int64 step-1 grid still matches.

**Fix sketch**

Require an explicit no-georef marker instead of inferring from coord shape. Two options:

- (a) `attrs['_xrspatial_no_georef'] = True`. The reader stamps it when emitting placeholder coords; the writer checks it before treating coords as no-georef.
- (b) Stamp a private flag on the coord variable (`coord.attrs`) that the reader sets and the writer checks. Survives ops that preserve coord attrs.

With either fix, int64 coords without the marker should fall through to the normal `coords_to_transform` path: synthesise a real transform, or raise `NonUniformCoordsError` on irregular spacing.

**Environment**

- xrspatial `main` (commit 22aa43c)
- python 3.11, numpy 2.x

brendancol:
Closes #2120

## Summary

- The writer used to detect the read-side no-georef placeholder by coord shape (int64 ascending step-1 on both axes). User-authored grids matching that pattern lost their georef on round-trip with no warning. The reproducer `x=[500,501,502], y=[1000,1001]` came back as `[0,1,2]` with no transform.
- Move the signal off shape onto `attrs['_xrspatial_no_georef']`. The reader stamps the marker together with the placeholder coords; the writer checks the marker and ignores coord shape.
- A user-authored int64 step-1 grid without the marker now falls through to `coords_to_transform`, which synthesises a unit transform and preserves CRS attrs.

## Backend coverage

- numpy, dask+numpy, GPU (cupy + dask+cupy): all go through `_populate_attrs_from_geo_info`, which now stamps the marker.
- VRT: emits empty coords on the no-georef path so the marker is not needed there.

## Test plan

- [x] New regression tests in `test_no_georef_marker_2120.py` covering: user grid keeps CRS and gains a transform; explicit-marker writes skip transform synthesis; round-trip via a real no-georef file preserves the marker.
- [x] Updated `test_int_coord_sentinel_2087.py` and `test_no_georef_writer_round_trip_1949.py` to pin the new contract (shape alone is no longer the signal).
- [x] Full `xrspatial/geotiff/tests` suite: 4222 passed, 25 skipped.
- [x] Full non-geotiff suite: 3672 passed, 15 skipped.

### DIFF
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -102,6 +102,16 @@ write.
    * - ``_xrspatial_geotiff_contract``
      - int
      - Contract version. Currently ``2``. See `Versioning`_.
+   * - ``_xrspatial_no_georef``
+     - bool
+     - Stamped ``True`` on reads of files with no GeoTIFF transform
+       tags. The reader emits int64 placeholder y/x coords for these
+       files; the marker tells the writer to reproduce that
+       no-georef shape on round-trip rather than synthesising a
+       fake unit transform. Absence of the marker means the array
+       has spatial coords the writer can interpret as georef. A
+       caller can opt into no-georef writes on a hand-built array
+       by setting this attr explicitly. See issue #2120.
 
 
 Compatibility aliases
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -129,7 +129,11 @@ import numpy as np
 from ._coords import (
     transform_tuple_from_pixel_geometry as _transform_tuple_from_pixel_geometry,
 )
-from ._geotags import RASTER_PIXEL_IS_AREA, RASTER_PIXEL_IS_POINT
+from ._geotags import (
+    RASTER_PIXEL_IS_AREA,
+    RASTER_PIXEL_IS_POINT,
+    _NO_GEOREF_KEY,
+)
 
 
 # Per-codec valid compression-level ranges, used by ``to_geotiff`` for
@@ -167,13 +171,6 @@ _TIFF_SHORT = 3
 # ``DeprecationWarning``. Downstream code that read those keys via
 # ``attrs[key]`` now sees ``KeyError`` rather than the deprecated value.
 _ATTRS_CONTRACT_VERSION = 2
-
-
-# The no-georef marker key lives in ``_coords`` to avoid a circular
-# import (``_attrs`` already imports from ``_coords``). Re-export here
-# so callers using ``_attrs`` for the contract constants can find it
-# alongside.
-from ._coords import _NO_GEOREF_KEY  # noqa: E402,F401
 
 
 # String identifiers (used in xrspatial attrs) -> TIFF ResolutionUnit tag ids.
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -169,6 +169,13 @@ _TIFF_SHORT = 3
 _ATTRS_CONTRACT_VERSION = 2
 
 
+# The no-georef marker key lives in ``_coords`` to avoid a circular
+# import (``_attrs`` already imports from ``_coords``). Re-export here
+# so callers using ``_attrs`` for the contract constants can find it
+# alongside.
+from ._coords import _NO_GEOREF_KEY  # noqa: E402,F401
+
+
 # String identifiers (used in xrspatial attrs) -> TIFF ResolutionUnit tag ids.
 _RESOLUTION_UNIT_IDS = {'none': 1, 'inch': 2, 'centimeter': 3}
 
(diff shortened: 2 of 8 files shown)


brendancol (commented):
## PR Review: geotiff: move no-georef signal off coord shape onto attrs marker

None.

- [ ] `xrspatial/geotiff/_attrs.py:173` — the mid-file `from ._coords import _NO_GEOREF_KEY` with `# noqa: E402,F401` works but reads oddly. `_geotags.py` is imported by both `_attrs` and `_coords`, so parking the constant there would let both files pick it up at the top with no noqa.
- [ ] `xrspatial/geotiff/_backends/vrt.py:268` — VRT's no-transform branch leaves `coords = {}` and skips the marker. Safe today (no x/y coords for the writer to misinterpret) but if a future change adds placeholder coords here without the marker, the silent-strip is back. Setting `attrs[_NO_GEOREF_KEY] = True` defensively in that `else` would prevent that.

- [ ] `xrspatial/geotiff/_coords.py:64` — `_is_no_georef_sentinel` is now only called from `test_int_coord_sentinel_2087.py`. The docstring already says so, but a one-line comment near the definition ("kept for diagnostic / test pinning only") would help anyone greping for callers.
- [ ] `xrspatial/geotiff/_coords.py:61` — `_has_no_georef_marker` uses `is True` (identity, not truthiness). That's deliberate -- a stray `attrs['_xrspatial_no_georef'] = 'yes'` shouldn't flip the writer -- but an inline note would save the next reader a moment.

### What looks good
- The marker resolves a genuine ambiguity. The old shape-based check was forced to guess between two cases that share a shape; the new check just asks.
- All four read paths funnel through `_populate_attrs_from_geo_info`, so the marker is stamped once and propagates uniformly.
- `isel` and the other in-memory ops preserve `da.attrs`, so the marker survives the usual transformations between read and write.
- New tests in `test_no_georef_marker_2120.py` cover both directions: user grid keeps georef, explicit marker writes no-georef.
- The two existing test files that pinned the old behaviour were updated to pin the new contract rather than deleted.

### Checklist
- [x] Algorithm matches reference/paper -- N/A, contract change.
- [x] All implemented backends produce consistent results -- eager / dask / GPU share `_populate_attrs_from_geo_info`.
- [x] NaN handling is correct -- unchanged by this PR.
- [x] Edge cases covered by tests -- user grid w/o marker, real no-georef read, slicing, explicit opt-in.
- [x] Dask chunk boundaries handled correctly -- N/A, attrs-only change.
- [x] Docstrings present and accurate -- `_has_no_georef_marker`, `_is_no_georef_sentinel`, `coords_to_transform`, and `require_transform_for_georeferenced` all rewritten for the new role.

brendancol (commented):
## Follow-up: all review items addressed

### Suggestions
- **Fixed** — `_NO_GEOREF_KEY` moved to `_geotags.py`. Both `_attrs` and `_coords` now import it at the top of the file; the mid-file noqa import is gone.
- **Fixed** — VRT no-transform branches (eager + chunked) now stamp `attrs[_NO_GEOREF_KEY] = True` defensively, matching the contract from `_populate_attrs_from_geo_info`.

### Nits
- **Fixed** — Added a heading comment on `_is_no_georef_sentinel` flagging it as kept for tests only.
- **Fixed** — Added an inline note on `_has_no_georef_marker` documenting that the `is True` identity check is deliberate (stray non-boolean stamps should not flip the writer).

Full geotiff suite still green: 4222 passed / 25 skipped.

PR #2128, opened 2026-05-19
geotiff: drop crs/crs_wkt when allow_rotated drops the transform

### LINKED ISSUE #2126: geotiff: allow_rotated=True leaks CRS attrs while dropping the transform
**Describe the bug**

With ``allow_rotated=True``, ``open_geotiff`` drops ``attrs['transform']`` (the rotated affine cannot be expressed as the axis-aligned rasterio 6-tuple, so the reader sets ``has_georef=False``) but still emits ``attrs['crs']`` / ``attrs['crs_wkt']``. The docstring at ``xrspatial/geotiff/__init__.py:385`` promises CRS is dropped on the rotated opt-in path. Downstream code that branches on ``"crs" in da.attrs`` therefore treats the array as spatially meaningful even though the transform is absent.

**Where**

``_populate_attrs_from_geo_info`` writes ``attrs['crs']`` and ``attrs['crs_wkt']`` at ``xrspatial/geotiff/_attrs.py:358-361`` before checking ``has_georef`` at line 374. The transform write below the check correctly skips on ``has_georef=False``; the CRS writes do not.

**Reproducer**

A rotated test fixture exists in the repo (see ``test_allow_rotated_2116.py``). Opening it with ``allow_rotated=True`` returns a DataArray with no transform attr but with crs / crs_wkt still populated.

**Fix**

Move the CRS writes inside the ``has_georef`` branch (option a from the audit), or expose an explicit ``rotated_affine`` public attr and make downstream handling explicit (option b). Option a matches the documented contract and is the smaller change.

**Environment**

- xrspatial main (commit 22aa43c)

brendancol:
Closes #2126

## Summary

- ``_populate_attrs_from_geo_info`` used to write ``crs`` / ``crs_wkt`` before the ``has_georef`` check, so the ``allow_rotated=True`` opt-in path dropped the transform but kept the CRS attrs. Public docs already promised CRS was dropped on rotation -- the implementation lagged.
- The fix gates the CRS writes on rotated-opt-in detection: skip them when ``transform.rotated_affine`` is set and ``has_georef`` is False. Plain no-georef files (no transform tags at all) still keep their CRS attrs.
- Public docstring in ``open_geotiff`` updated to spell out that both ``crs`` and ``crs_wkt`` are dropped; the old reference to ``geo_info.transform.rotated_affine`` (which is not accessible from the returned DataArray) was corrected.

## Backend coverage

- numpy / dask / GPU / dask+GPU all share ``_populate_attrs_from_geo_info``.
- VRT does not deal with rotated transforms (GDAL geo_transform is axis-aligned), so no VRT change needed.

## Test plan

- [x] New regression tests in ``test_allow_rotated_crs_drop_2126.py`` covering: rotated opt-in drops ``crs`` and ``crs_wkt``; plain no-georef keeps CRS; axis-aligned georef keeps both CRS and transform; end-to-end via ``open_geotiff`` (eager and dask).
- [x] Existing rotated tests (``test_allow_rotated_geotiff_2115.py``, ``test_rotated_transform_attr_1764.py``) still pass.
- [x] Full geotiff suite: 4223 passed, 25 skipped.


brendancol (commented):
## PR Review: drop crs/crs_wkt when allow_rotated drops the transform

None.

- [ ] No new public attr is exposed for the rotated 6-tuple. The original docstring (now corrected) referenced ``geo_info.transform.rotated_affine`` but ``geo_info`` is not on the returned DataArray, so the rotated matrix is unreachable from the public API. The audit (#2126) listed exposing a ``rotated_affine`` attr as an alternative to dropping CRS. We took the drop-CRS path, which closes the documented contract gap, but a follow-up to surface the rotated 6-tuple as ``attrs['rotated_affine']`` would let callers handle the rotation explicitly instead of treating it as silently absent. Out of scope for this PR -- worth a follow-up issue.

- [ ] ``xrspatial/geotiff/_attrs.py:367`` -- ``getattr(src_t, 'rotated_affine', None)`` is defensive against older GeoTransform pickles. The current dataclass always has the field, so a plain ``src_t.rotated_affine`` would also work. Defensive ``getattr`` is fine; just noting.
- [ ] ``xrspatial/geotiff/tests/test_allow_rotated_crs_drop_2126.py:18-19`` -- the ``import io`` and ``import struct`` at the top are unused. Drop them.

### What looks good
- The detection (``transform.rotated_affine is not None`` and ``has_georef=False``) cleanly separates the rotated opt-in path from plain no-georef files. CRS attrs survive on plain no-georef reads, which preserves the existing useful behaviour.
- Tests cover both unit-level (``_populate_attrs_from_geo_info`` directly) and end-to-end (``open_geotiff`` with a synthetic rotated TIFF carrying a GeoKey-encoded CRS).
- The eager and dask paths both run through ``_populate_attrs_from_geo_info``, so the fix lands in one place.
- Updated the ``open_geotiff`` docstring to drop the broken reference to ``geo_info.transform.rotated_affine`` and spell out that both ``crs`` and ``crs_wkt`` are dropped.

### Checklist
- [x] Algorithm matches reference/paper -- N/A, contract change.
- [x] All implemented backends produce consistent results -- eager / dask / GPU share ``_populate_attrs_from_geo_info``; VRT does not handle rotated transforms.
- [x] NaN handling is correct -- unchanged.
- [x] Edge cases covered by tests -- rotated + CRS, plain no-georef + CRS, axis-aligned + CRS.
- [x] Dask chunk boundaries handled correctly -- attrs-only change.
- [x] Docstrings present and accurate -- ``_populate_attrs_from_geo_info`` and the public ``open_geotiff`` docstring both updated.

brendancol (commented):
## Follow-up: review items addressed

### Suggestions
- **Deferred to follow-up** -- ``attrs['rotated_affine']`` exposure filed as #2129. The new public attr is API surface expansion that deserves its own PR with explicit consideration of the writer round-trip contract.

### Nits
- **Fixed** -- removed unused ``import io`` and ``import struct`` from ``test_allow_rotated_crs_drop_2126.py``.
- **Dismissed** -- defensive ``getattr(src_t, 'rotated_affine', None)`` kept as-is. Plain attribute access works on the current dataclass; the defensive form survives any future GeoTransform pickle that predates the field and costs nothing.

Tests still green.

brendancol:
## PR Review: geotiff: drop crs/crs_wkt when allow_rotated drops the transform

_None_

- [ ] `xrspatial/geotiff/tests/test_allow_rotated_crs_drop_2126.py:128` — the assertion `da.attrs.get('transform') is None or da.attrs.get('transform') == 'rotated'` accepts the literal string `'rotated'`, but no code path in `_populate_attrs_from_geo_info` ever assigns that string. The same defensive form appears in `test_allow_rotated_geotiff_2115.py:181`, so it is a copy of existing test style, but it weakens the assertion. Consider tightening to `assert 'transform' not in da.attrs` to actually pin the documented contract.
- [ ] `xrspatial/geotiff/_attrs.py:375` — `rotated_optin` detection is local to this function. If a second site ever needs the same gate (e.g. a future writer that learns to emit `rotated_affine`), the predicate will drift. Consider lifting to a small helper on `GeoInfo` or `_geotags`, e.g. `geo_info.is_rotated_optin()`. Not blocking for the current fix.
- [ ] `xrspatial/geotiff/tests/test_allow_rotated_crs_drop_2126.py:39-51` — `test_rotated_optin_drops_crs_epsg` and `test_rotated_optin_drops_crs_wkt` both build the same `GeoInfo` and call the same function; one parametrized test covering both keys would be tighter. Cosmetic.

- [ ] `xrspatial/geotiff/_attrs.py:364-365` — `src_t` and `has_georef` are now read once at the top of the function and reused twice. Good move, just note that the old comment block at lines 389-397 still reads as if `has_georef` is being looked up locally there. The comment is accurate but the reader has to scroll up to find the binding.
- [ ] `xrspatial/geotiff/tests/test_allow_rotated_crs_drop_2126.py:86-113` — `_write_rotated_tiff_with_geokeys` duplicates the GeoKeyDirectory + ModelTransformationTag construction that `test_allow_rotated_geotiff_2115.py` already does. Pulling into a shared `tests/_helpers.py` would let future rotated-CRS regressions reuse it.
- [ ] PR scope: the description says the PR fixes #2126 only, but commit `b0edaa5` pulls in the `pytest.importorskip("tifffile")` change for `test_masked_nodata_attr_2092.py` from PR #2140. The commit message calls this out, and project memory notes that security-sweep PRs should be one fix per PR. This is a CI unblock rather than a security fix, so it is defensible, just worth noting in the merge log so reviewers of #2140 do not re-apply.

### What looks good
- The gate is precise: `transform.rotated_affine is not None and not has_georef`. Both conditions are produced together in `_extract_transform` (geotags.py:691-693), so the predicate matches the only path that produces rotated `GeoInfo`.
- `getattr` with `True` default on `has_georef` keeps `_populate_attrs_from_geo_info` safe for callers that pass legacy `GeoInfo` shapes; the `not has_georef` term then evaluates False, so the legacy behaviour of emitting CRS attrs is preserved.
- Plain no-georef (no transform tags) keeps CRS attrs, covered by `test_plain_no_georef_keeps_crs`. Axis-aligned georef keeps both, covered by `test_axis_aligned_georef_keeps_crs_and_transform`. The end-to-end tests exercise both eager and dask paths through `open_geotiff`.
- Docstring on `open_geotiff` is updated to match the new behaviour, including the correction that `rotated_affine` is not surfaced on the returned DataArray (follow-up #2129 filed for that API surface).
- VRT path is correctly out of scope: `_gdal_geotransform_to_affine_tuple` produces axis-aligned transforms only, so `rotated_affine` is never set on the VRT branch.
- All four read backends (eager numpy, dask, GPU, dask+GPU) route through `_populate_attrs_from_geo_info`, so the single-point fix covers all of them.

### Checklist
- [x] Algorithm correctness vs cited reference (gate matches the only producer of rotated `GeoInfo`)
- [x] NaN / float handling — not applicable, attrs-only change
- [x] Edge cases: rotated-with-crs, plain no-georef, axis-aligned all covered
- [x] Off-by-one indexing — not applicable
- [x] Security / metadata guarantee enforced: docstring promise of CRS drop now matches behaviour
- [x] Backend completeness: numpy / dask / cupy / dask+cupy share `_populate_attrs_from_geo_info`; VRT correctly omitted
- [x] Dask `map_overlap` — not applicable
- [x] CuPy thread bounds / register pressure — not applicable
- [x] No `.values` / `.compute()` regressions — not applicable
- [x] Tests exercise all backends (eager + dask via `chunks=4`)
- [x] Reference / contract values asserted (not just "doesn't crash")
- [x] Temp files unique (use `tmp_path` with `tmp_2126_*` prefixes)
- [x] Regression test covers the threat model (CRS leakage on rotated opt-in)
- [x] Public docstring updated with Parameters / contract details
- [ ] README feature matrix — not applicable, internal attrs contract change

brendancol:
### Review follow-up

- Tightened the dead `or == 'rotated'` assertion in `test_open_geotiff_rotated_with_crs_drops_crs` to `'transform' not in da.attrs`, and added the same check to the dask variant (commit `8c976c7`).
- Refactor suggestion (lift `rotated_optin` detection to a `GeoInfo` helper): leaving for a follow-up. The inline check at `_attrs.py:381-385` is two conditions and only one call site today; adding the helper before a second writer learns rotated transforms would be premature.
- Bundled tifffile cherry-pick: PR #2140 merged separately, so the duplicate commit in this branch is harmless. Keeping it on the branch so CI on this PR doesn't depend on a separate merge landing first.

PR #2131, opened 2026-05-19
geotiff: forward allow_rotated to the HTTP dask metadata path

### LINKED ISSUE #2130: geotiff: HTTP dask read path ignores allow_rotated
**Describe the bug**

``read_geotiff_dask`` forwards ``allow_rotated`` to the local metadata path but not to the HTTP/fsspec path. ``open_geotiff(url, chunks=..., allow_rotated=True)`` therefore behaves differently from the eager remote read and the local chunked read for the same file: the HTTP dask path raises ``NotImplementedError`` on rotated transforms even when the caller opted in.

**Where**

``xrspatial/geotiff/_backends/dask.py:164`` calls ``_parse_cog_http_meta(_src, overview_level=overview_level)`` without ``allow_rotated``. The local path at line 205-206 forwards the kwarg correctly.

**Reproducer**

Serve a rotated GeoTIFF over a local HTTP server (mirroring ``test_open_geotiff_http_rotated_allow_rotated_reads_pixels`` in ``test_allow_rotated_geotiff_2115.py``) and open with ``chunks=`` + ``allow_rotated=True``. The local-chunks variant works; the HTTP variant should work the same way but does not.

**Fix**

Thread ``allow_rotated`` through ``_parse_cog_http_meta`` (and any helpers it calls that gate on rotation) so all four paths (eager local, eager HTTP, dask local, dask HTTP) honour the kwarg uniformly.

**Environment**

- xrspatial main (commit 22aa43c)

brendancol:
Closes #2130

## Summary

- ``read_geotiff_dask`` forwarded ``allow_rotated`` to the local chunked metadata path but not to the HTTP/fsspec path. Opening a rotated remote GeoTIFF with ``chunks=...`` raised ``NotImplementedError`` from the parser even when the caller had opted in.
- The parser has accepted ``allow_rotated`` since #2115; the dask backend just never threaded it through.
- One-line wiring fix: pass ``allow_rotated=allow_rotated`` into the ``_parse_cog_http_meta`` call.

## Backend coverage

- HTTP dask path -- this PR.
- Local dask, eager local, eager HTTP already supported.

## Test plan

- [x] New regression tests in ``test_http_dask_allow_rotated_2130.py`` using a range-aware HTTP test server: rotated default raises, rotated opt-in reads pixels through the dask path.
- [x] Existing rotated tests (``test_allow_rotated_geotiff_2115.py``) still pass.
- [x] Full geotiff suite: 4236 passed, 25 skipped.


brendancol (commented):
## PR Review: forward allow_rotated to the HTTP dask metadata path

None.

None.

None.

### What looks good
- Single-call wiring fix matching the documented contract (``_parse_cog_http_meta`` already accepted ``allow_rotated`` since #2115; just never received it from the dask backend).
- Test coverage exercises both directions (default raises, opt-in reads) via a range-aware HTTP test server, mirroring the pattern in ``test_http_dask_orientation_1794.py``.
- The downstream tile / strip fetch paths operate on raw bytes and do not care about rotation, so the fix is correctly scoped to the metadata parse call alone.

### Checklist
- [x] Algorithm matches reference/paper -- N/A.
- [x] All implemented backends produce consistent results -- HTTP dask now matches HTTP eager and local dask.
- [x] NaN handling is correct -- unchanged.
- [x] Edge cases covered by tests -- default-raises, opt-in-reads.
- [x] Dask chunk boundaries handled correctly -- unchanged; only the metadata parser call was touched.
- [x] Docstrings present and accurate -- inline comment explains the kwarg.

Issue #2087, opened 2026-05-18
Bug: integer spatial coords silently strip georef on write


brendancol:
**Describe the bug**

A DataArray with integer-spaced x/y coords (e.g. `x=[100,101,102]`, `y=[200,199]`) writes through `to_geotiff` with no transform tags and reads back as pixel coords `x=[0,1,2]`, `y=[0,1]` with no georef. The projection metadata is silently lost.

The cause is `coords_to_transform` at `xrspatial/geotiff/_coords.py:353`, which returns `None` for any integer x or y dtype. The fail-closed validator at `_coords.py:272` mirrors that, so the writer doesn't raise either. Both branches were intended as a no-georef sentinel for files round-tripped through `open_geotiff` (which emits `np.arange(N, dtype=int64)` for x/y when the source file has no GeoTIFF transform tags, #1710 / #1753 / #1949). The sentinel is too broad: integer dtype on its own catches both the read-side `arange` placeholder and any user-authored projected grid that happens to use integer-spaced coords.

**Repro**

```python
import numpy as np
import xarray as xr
from xrspatial.geotiff import to_geotiff, open_geotiff

da = xr.DataArray(
    np.zeros((2, 3), dtype=np.float32),
    coords={'y': np.array([200, 199]), 'x': np.array([100, 101, 102])},
    dims=('y', 'x'),
)
to_geotiff(da, 'out.tif')

out = open_geotiff('out.tif')
print(out.coords['x'].values)  # [0 1 2] -- georef lost
print(out.coords['y'].values)  # [0 1]
print(out.attrs.get('transform'))  # None
```

**Expected behavior**

Either the projected grid round-trips with its coords intact, or the write raises so the caller knows the georef is being dropped. Silent loss of the transform is the worst outcome.

**Fix**

Tighten the sentinel to match exactly what the reader emits: dtype is `int64` AND `np.array_equal(coord, np.arange(len(coord), dtype=int64))`. Only that pattern is treated as no-georef. Anything else (user-authored `[100,101,102]`, subset `[2,3,4]`, subsampled `[0,2,4]`, non-uniform `[1,2,5]`) falls through to the existing float-coord path, which either synthesizes a real transform from the spacing or raises `NonUniformCoordsError`.

Two `_coords.py` sites need the change: the validator at `:272` and `coords_to_transform` at `:353`.

**Trade-off**

[...]

PR #2141, opened 2026-05-19
geotiff: add Hypothesis property tests for round-trip metadata

### LINKED ISSUE #2134: geotiff: add Hypothesis property tests for metadata round trips
## Summary

Add Hypothesis-driven property tests for the GeoTIFF write/read round trip.

Round-trip coverage today is a long list of incident-specific files (`test_metadata_round_trip_1484.py`, `test_descending_coords_1716.py`, `test_no_georef_writer_round_trip_1949.py`, `test_int_coords_round_trip_hotfix_1962.py`, `test_round_trip_invariants.py`) plus one fuzz file (`test_fuzz_hypothesis_1661.py`) that covers dtype/codec/predictor but skips the metadata axes. The metadata axes (coord dtype, axis direction, degenerate shape, CRS/transform presence, nodata encoding, band axis position) are where most of the recent round-trip bugs have come from.

## Why property tests instead of more example tests

- Combinatorial blowup. Six axes with 3-4 values each is roughly 500-1500 combinations. Not feasible by hand.
- Bugs in the existing 200+ geotiff test files cluster around interactions between two axes (integer coords with degenerate shape, descending y with rotated transform, float nodata with int dtype). Single-axis example tests miss those.
- Hypothesis shrinking gives a minimal failing case for free when a future writer change breaks one of the corners.

## Property space

One strategy per axis, then a composite strategy that draws one value from each.

| Axis | Strategy |
|---|---|
| coord dtype | `sampled_from(['int32', 'int64', 'float32', 'float64'])` |
| axis direction | `sampled_from(['asc_asc', 'asc_desc', 'desc_asc', 'desc_desc'])` |
| shape | `sampled_from([(1, 1), (1, 8), (8, 1), (4, 5), (16, 16)])` |
| georef presence | `sampled_from(['crs_only', 'transform_only', 'both', 'neither'])` |
| nodata mode | `sampled_from(['in_range', 'out_of_range', 'fractional', 'nan', 'none'])` |
| band layout | `sampled_from(['band_first', 'band_last', 'no_band'])` |
| pixel dtype | `sampled_from(['uint8', 'int16', 'int32', 'float32', 'float64'])` |
| CRS | `sampled_from([4326, 3857, 32633, 26910, None])` (None paired with `crs_only=False`) |

Per-draw filtering for illegal combinations (e.g. `nodata=nan` with int dtype must promote to float on read, which is documented behaviour, not a failure).

## Round-trip invariant

After two cycles:

```
da0  -> write(p1) -> read(p1) = da1
da1  -> write(p2) -> read(p2) = da2
```

assert `da1 == da2` under semantic equality:

- `data`: `np.array_equal` with NaN-aware compare
- `dtype`: identical
- `dims`: identical (including band axis position)
- `coords`: per-axis `np.allclose` for float, `np.array_equal` for int; direction preserved
- `attrs['crs']`: same int EPSG, or both absent
- `attrs['transform']`: same `GeoTransform` tuple to 1e-9 relative tolerance, or both flagged no-georef via the marker from #2120
- `attrs['nodata']`: same value, or both absent; NaN compares equal to NaN

Fixed-point (`da1 == da2`) works because the writer is deterministic given the same input attrs. One full cycle is enough to surface drift.

## File location

`xrspatial/geotiff/tests/test_roundtrip_properties.py`

Skip the module if `hypothesis` is not installed, same pattern as `test_fuzz_hypothesis_1661.py`.

## Hypothesis profile

- Local: `settings(max_examples=200, deadline=None, suppress_health_check=[HealthCheck.too_slow])`
- CI profile registered as `ci` with `max_examples=50` and `derandomize=True` for reproducibility
- Seed printed on failure (Hypothesis default)

## Backend scope

numpy and dask+numpy for the first pass. cupy and dask+cupy share the same writer/reader code but need a CUDA runner, so leave those for a follow-up once the numpy invariants are pinned.

## Out of scope

- Byte-for-byte file equality. The writer is allowed to reorder IFD tags, change strip layout, etc. The `test_golden_corpus_*.py` files cover byte stability where it matters.
- Performance or timing assertions.
- VRT, COG, overviews. Those have their own round-trip suites.
- GPU code paths.

## Tie-in

This file is meant to back-stop several adjacent contracts. Cross-reference them in the docstring:

- backend parity matrix (`xrspatial/geotiff/tests/test_backend_parity_matrix.py`)
- no-georef marker from #2120
- nodata semantics split (declared vs masked, see `test_masked_nodata_attr_2092.py` and #2092)
- int coords sentinel from #2087

## Dev dependency

`hypothesis` is already imported by `test_fuzz_hypothesis_1661.py` via `pytest.importorskip`. It is not declared in `pyproject.toml` or `setup.py`. Adding it to a `test` extras group is a small follow-up but not blocking. The new file uses the same `importorskip` guard.

## Acceptance

- [ ] New file `xrspatial/geotiff/tests/test_roundtrip_properties.py` exists
- [ ] Composite strategy covers all 8 axes above
- [ ] Both `ci` and default profiles registered
- [ ] Green for 200 examples locally on numpy backend
- [ ] Green for 50 examples on dask+numpy backend
- [ ] Docstring cross-references #2087, #2092, #2120 and `test_backend_parity_matrix.py`

brendancol:
Closes #2134.

## Summary

- Add `xrspatial/geotiff/tests/test_roundtrip_properties.py`. A composite Hypothesis strategy over eight metadata axes (coord dtype, axis direction, shape, CRS/transform presence, nodata mode, band layout, pixel dtype, CRS EPSG) feeds two test cases that assert a fixed-point invariant across two write -> read cycles.
- One case runs on the numpy backend with the registered `rockout_default` profile (200 examples). A second case wraps the input in dask chunks and runs on the registered `ci` profile (50 examples) to exercise the streaming write path. GPU paths and byte equality stay out of scope per the issue.
- Register `hypothesis` in the `tests` extras in `setup.cfg`. The new module guards its imports with `pytest.importorskip("hypothesis")` so the test still skips cleanly in environments that don't install the extras.

## Notes

- Illegal combinations the writer already refuses (NaN / fractional / out-of-range nodata paired with an int pixel dtype, 1x1 float-coord rasters that can't derive a pixel size, etc.) are filtered with `assume(False)` rather than treated as failures. Hypothesis stats show ~30-50% invalid draws on the strategy filter and ~5% from the in-test writer refusal, which is expected for the documented refusal surface.
- The docstring cross-references `test_backend_parity_matrix.py`, the no-georef marker from #2120, the nodata semantics split from #2092, and the int coords sentinel from #2087, so future readers can trace the contract.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_roundtrip_properties.py::test_round_trip_fixed_point_numpy` -- 200 passing examples
- [x] `pytest xrspatial/geotiff/tests/test_roundtrip_properties.py::test_round_trip_fixed_point_dask` -- 50 passing examples
- [x] `pytest xrspatial/geotiff/tests/test_fuzz_hypothesis_1661.py` still green

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,6 +72,7 @@ gpu =
 tests =
     flake8
     geopandas
+    hypothesis
     isort
     noise >= 1.2.2
     dask
--- a/xrspatial/geotiff/tests/test_fuzz_hypothesis_1661.py
+++ b/xrspatial/geotiff/tests/test_fuzz_hypothesis_1661.py
@@ -47,7 +47,18 @@ ALLOWED_PARSE_EXCEPTIONS = (ValueError, TypeError)
 # explicitly rejected on write (see _VALID_COMPRESSIONS docstring); 'lerc' and
 # 'jpeg2000' are lossy or dtype-restricted and would need their own narrower
 # strategies, so they're omitted here.
-LOSSLESS_CODECS = ['none', 'deflate', 'lzw', 'packbits', 'zstd', 'lz4']
+#
+# ``zstd`` and ``lz4`` depend on optional third-party packages that are
+# not in the [tests] extras. Drop them from the strategy when those
+# packages are missing so the fuzz run does not flake the moment
+# Hypothesis happens to draw a codec the runner can't actually write.
+from xrspatial.geotiff._compression import LZ4_AVAILABLE, ZSTD_AVAILABLE
+
+LOSSLESS_CODECS = ['none', 'deflate', 'lzw', 'packbits']
+if ZSTD_AVAILABLE:
+    LOSSLESS_CODECS.append('zstd')
+if LZ4_AVAILABLE:
+    LOSSLESS_CODECS.append('lz4')
 
 # Dtype set kept small to keep CI fast. Float and int, signed and unsigned.
 ROUND_TRIP_DTYPES = ['uint8', 'uint16', 'int16', 'int32', 'float32', 'float64']
--- a/xrspatial/geotiff/tests/test_roundtrip_properties.py
+++ b/xrspatial/geotiff/tests/test_roundtrip_properties.py
@@ -49,8 +49,16 @@ Out of scope (deferred):
 
 Hypothesis profiles:
 
-* default: ``max_examples=200``, ``deadline=None``
-* ``ci``: ``max_examples=50``, ``derandomize=True``
+The two profiles below are registered as inheritance sources for the
+``@settings(parent=...)`` on the two test functions. They are not
+loaded as the active Hypothesis profile in CI -- the test functions
+inherit the settings directly via ``parent=``. Renaming or wiring up
+``settings.load_profile('reduced')`` from a CI conftest is fine; the
+contract is the example-count budget on each test, not the profile
+name.
+
+* ``local`` (numpy test): ``max_examples=200``, ``deadline=None``
+* ``reduced`` (dask test): ``max_examples=50``, ``derandomize=True``
 
 The module is skipped if ``hypothesis`` is not installed (same pattern
 as ``test_fuzz_hypothesis_1661.py``).
@@ -58,6 +66,7 @@ as ``test_fuzz_hypothesis_1661.py``).
 from __future__ import annotations
 
 import math
+import os
 
 import numpy as np
 import pytest
@@ -68,6 +77,7 @@ hypothesis = pytest.importorskip("hypothesis")
 from hypothesis import (  # noqa: E402
     HealthCheck,
     assume,
+    event,
     given,
     settings,
 )
@@ -91,14 +101,14 @@ _COMMON_SUPPRESS = [
 ]
 
 settings.register_profile(
-    "ci",
+    "reduced",
     max_examples=50,
     deadline=None,
     derandomize=True,
     suppress_health_check=_COMMON_SUPPRESS,
 )
 settings.register_profile(
-    "rockout_default",
+    "local",
     max_examples=200,
     deadline=None,
     suppress_health_check=_COMMON_SUPPRESS,
@@ -175,10 +185,11 @@ def _pick_nodata(mode: str, dtype_name: str, rng: np.random.Generator):
     pixel sample range (so no real pixel happens to equal it).
     ``out_of_range`` -- only valid for floats (would raise for ints).
     ``fractional`` -- only valid for floats.
-    ``nan`` -- only valid for floats.
+    ``nan`` -- only valid for floats; returned as ``float('nan')``.
     ``none`` -- no sentinel.
 
-    Returns ``None`` for the ``none`` case, or a Python scalar otherwise.
+    Returns ``None`` for the ``none`` case, ``float('nan')`` for the
+    ``nan`` case, or a Python ``int`` / ``float`` scalar otherwise.
     """
     dtype = np.dtype(dtype_name)
     if mode == 'none':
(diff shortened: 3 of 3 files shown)


brendancol (commented):
## PR Review: geotiff: add Hypothesis property tests for round-trip metadata

### Blockers

None.

### Suggestions

- [ ] `xrspatial/geotiff/tests/test_roundtrip_properties.py:388-396` and `:454-458`: the `except (ValueError, TypeError): assume(False)` block silently swallows ~8.5% of executions (17/200 on the numpy property per Hypothesis stats). If a previously-supported combo starts raising `ValueError`, the test will just skip the draw instead of failing -- a real regression becomes invisible. Two ways out: (a) narrow the strategy upfront so the writer-reject rate is near zero, or (b) record which combos get skipped via `hypothesis.event(...)` so a regression that bumps the skip rate from 8% to 30% shows up in CI stats. The minimal version of (b) is one line: `hypothesis.event(f"writer_rejected_{spec['nodata_mode']}_{spec['pixel_dtype']}")` before the `assume(False)`.

- [ ] `xrspatial/geotiff/tests/test_roundtrip_properties.py:84-105`: the profiles are registered at import time but no pytest hook calls `settings.load_profile('ci')` in CI. The `@settings(parent=settings.get_profile('ci'))` decorator does inherit those settings, so the dask test does run at 50 examples. But the name `'ci'` looks like it's meant to be the active profile in CI runs, which it isn't. Rename to `'reduced'` or `'fast'`, or add a note in the docstring spelling out that `'ci'` here is just an inheritance source.

- [ ] `xrspatial/geotiff/tests/test_roundtrip_properties.py:390`: the bound `e` in `except (ValueError, TypeError) as e:` is unused. Drop the `as e`, or log it -- `hypothesis.event(f"writer_rejected: {type(e).__name__}")` would solve the previous suggestion too.

### Nits

- [ ] `xrspatial/geotiff/tests/test_roundtrip_properties.py:101`: the profile name `rockout_default` reads as project-specific jargon. The issue's acceptance criterion says "default and ci". `'local'` would do, or just drop the registration and apply `@settings(max_examples=200, deadline=None, ...)` inline since only one test uses it.

- [ ] `xrspatial/geotiff/tests/test_roundtrip_properties.py:237` and `:238`: `crs_epsg` and `n_bands` are always drawn but only consumed for some draws (half and one third respectively). Strategy slots wasted, not a bug.

- [ ] `xrspatial/geotiff/tests/test_roundtrip_properties.py:452`: 200 + 50 = 250 tmpdirs per session, each with two small `.tif` files. pytest cleans them up at teardown, but a single tmpdir plus `os.unlink` after each round would keep the file count to two at any moment.

- [ ] `xrspatial/geotiff/tests/test_roundtrip_properties.py:175-179`: docstring says "Returns None for the none case, or a Python scalar otherwise"; the `nan` branch returns `float('nan')`, which is a scalar but worth flagging explicitly.

### What looks good

- All eight strategy axes from the issue are present.
- `assume(_is_legal_combo(spec))` is inside the composite strategy, so filtered draws don't burn the example budget.
- `_assert_fixed_point` compares NaN-as-data via mask-and-equal, and handles the transform tuple with `math.isclose`.
- `_LOCKED_ATTRS` includes `_NO_GEOREF_KEY`, so the #2120 marker is in the contract.
- Docstring cross-references #2087, #2092, #2120, and `test_backend_parity_matrix.py`.
- `setup.cfg` picks up the `hypothesis` dev-dep that the issue flagged as a follow-up.

### Checklist

- [x] Algorithm matches reference/paper -- N/A, test-only change.
- [x] All implemented backends produce consistent results -- numpy and dask+numpy covered; GPU out of scope per issue.
- [x] NaN handling is correct.
- [x] Edge cases covered by tests -- 1x1, 1xN, Nx1 shapes included.
- [x] Benchmark exists or is not needed -- not needed for test-only PR.

brendancol (commented):
## Follow-up review after 1478b36

All seven findings from the first pass are addressed in this push. Disposition:

### Suggestions (3 of 3 fixed)

- [x] Skip-rate visibility -- `hypothesis.event(f"writer_rejected:{type(exc).__name__}")` now tags each writer-reject `assume(False)`. Hypothesis stats output shows the rate (`~4.29% writer_rejected:ValueError` on the latest local run). A regression that bumps the rate will surface in CI.
- [x] `'ci'` profile name -- renamed to `'reduced'`. Docstring now explains that both profiles are inheritance sources for `@settings(parent=...)` rather than the active CI profile.
- [x] Unused `e` in `except` -- now bound as `exc` and consumed in the `event(...)` call.

### Nits (4 of 4 fixed)

- [x] `rockout_default` profile name -- renamed to `'local'`.
- [x] `crs_epsg` / `n_bands` always drawn -- now conditional inside the composite strategy. `crs_epsg` is drawn only when `georef in ('crs_only', 'both')`; `n_bands` only when `band_layout != 'no_band'`.
- [x] Tmpdir churn -- per-example `.tif` files are now `os.unlink`-ed in a `finally` block. Disk file count stays at 2 (numpy test) or 2 (dask test) at any moment instead of growing to 400 + 100 by session teardown.
- [x] `_pick_nodata` docstring -- now explicitly notes the `nan` branch returns `float('nan')`.

Tests pass post-merge with origin/main: 200 + 50 examples, 0 failures.

PR #2142, opened 2026-05-19
geotiff: extend backend parity matrix to 8 backends and 7 fixtures

### LINKED ISSUE #2132: geotiff: add formal backend parity test matrix
## Summary

`xrspatial.geotiff` exposes one logical read API (`open_geotiff`) that fans out to four backend modules (eager numpy in `__init__.py`, `_backends/dask.py`, `_backends/gpu.py`, `_backends/vrt.py`) and several source types (local path, HTTP, fsspec URI, BytesIO). Parity between those paths is asserted today, but the assertions are scattered across roughly 15+ files keyed to past bug numbers. Add a single matrix that names every supported (backend, source) pair and asserts the same set of fields on every cell.

## Where parity is checked today

The existing parity files each pin a slice of the contract for one historical bug:

- `xrspatial/geotiff/tests/test_backend_parity_matrix.py`: already a table-driven harness, but currently parametrized over only `numpy` and `dask+numpy`. Has the `assert_parity` helper, `_FixtureSpec` dataclass, and `_materialise` / `_coord_view` helpers ready to extend. Its docstring names it as "single source of truth" for fixture-vs-backend parity going forward.
- `xrspatial/geotiff/tests/test_backend_pixel_parity_matrix_1813.py`: 4-backend pixel-byte parity across dtype/compression/layout combinations, with its own `_BACKENDS` list and `_materialise` helper.
- `xrspatial/geotiff/tests/test_attrs_parity_1548.py`: 4-backend attrs key/value parity for one tifffile-written fixture.
- `xrspatial/geotiff/tests/test_backend_kwarg_parity_1561.py`: kwargs reach each backend (dispatcher does not silently drop them).
- `xrspatial/geotiff/tests/test_signature_parity_1631.py`: signature parity between `open_geotiff` and the explicit `read_geotiff_*` entry points.
- `xrspatial/geotiff/tests/test_miniswhite_backend_parity_1797.py`: MinIsWhite photometric handling parity.
- `xrspatial/geotiff/tests/test_vrt_backend_coverage_2026_05_11.py`: VRT GPU + dask+GPU coverage.
- `xrspatial/geotiff/tests/test_bytesio_source.py`: BytesIO round-trip.
- `xrspatial/geotiff/tests/test_cog_http_*.py`, `test_http_*.py`: HTTP COG behaviour, mostly read-side effects.

A backend change (for example #2127's `_set_nodata_attrs` rewire) has to be cross-checked against ~10 separate files to know whether it kept the contract. No single file says "this fixture is read by these 8 paths and they all return the same thing."

## What the matrix should cover

The matrix is `(fixture, backend) -> read DataArray`, plus a small set of error fixtures.

### Backends (8)

| id              | dispatch                                        | source type |
| --------------- | ----------------------------------------------- | ----------- |
| `numpy`         | `open_geotiff(path)`                            | local path  |
| `dask+numpy`    | `open_geotiff(path, chunks=N)`                  | local path  |
| `gpu`           | `open_geotiff(path, gpu=True)`                  | local path  |
| `dask+gpu`      | `open_geotiff(path, gpu=True, chunks=N)`        | local path  |
| `vrt-eager`     | `open_geotiff(vrt_path)`                        | local `.vrt`|
| `vrt-dask`      | `open_geotiff(vrt_path, chunks=N)`              | local `.vrt`|
| `http-cog`      | `open_geotiff('http://...')`                    | HTTP URL    |
| `fsspec-memory` | `open_geotiff('memory://...')` or BytesIO       | fsspec / IO |

GPU rows skip via the existing `_gpu_available()` predicate from `test_backend_pixel_parity_matrix_1813.py`. HTTP rows use `pytest-httpserver` (already a dev dep, see `test_cog_http_concurrent.py`). fsspec rows use `memory://` URIs and `io.BytesIO`. The BytesIO row also asserts the file-like rejection path for `gpu=True` / `chunks=N` separately, since those combinations are documented `ValueError` cases in `open_geotiff`.

### Fixtures (initial set)

Reuse the `_FixtureSpec` dataclass in `xrspatial/geotiff/tests/test_backend_parity_matrix.py`. Start with:

- int16 single-band stripped (already present)
- uint16 multiband tiled
- float32 single-band with nodata sentinel
- int8 single-band with `mask_nodata=False` (locks the #2092 / #2127 masked-flag contract)
- COG with internal overview
- VRT mosaic over two tiles
- MinIsWhite single-band (gated on `tifffile`)
- error fixture: rotated `ModelTransformationTag` without `allow_rotated`

### Assertions

The existing `assert_parity` in `test_backend_parity_matrix.py` already covers most fields. Concrete checks per cell:

1. `_assert_pixels_equal(ref, actual)`: byte-equal for integer dtype, `np.array_equal(equal_nan=True)` for float.
2. `da.dims == spec.expected_dims` and `da.shape == ref.shape`.
3. Per-axis coord values + dtype: `_coord_view(ref, axis).tobytes() == _coord_view(da, axis).tobytes()` and `ref_c.dtype == actual_c.dtype`.
4. `da.attrs.get(k) == ref.attrs.get(k)` for a fixed canonical subset (`raster_type`, `transform`, `crs`, `crs_wkt`, `nodata`, `masked_nodata`).
5. `da.dtype == spec.dtype` (catches silent upcast that the reference read would also exhibit).
6. `attrs['nodata']` semantics: present iff the fixture declares one; sentinel value matches; `masked_nodata` flag reflects whether the mask was applied (#2092).
7. Error fixtures: `pytest.raises(ExpectedExc, match=expected_msg)` runs on every backend that supports the source type.

### Helper location

Extend `xrspatial/geotiff/tests/test_backend_parity_matrix.py` instead of creating a new `_parity_matrix.py`. The dataclass, materialise helper, and `assert_parity` already live there with a docstring naming the file as single source of truth. Adding the remaining 6 backend entries to `_BACKENDS` and the fixtures above to `_FIXTURES` is mechanical.

The cross-backend helpers in `xrspatial/tests/general_checks.py` (`general_output_checks`, `assert_numpy_equals_dask_numpy`, `assert_numpy_equals_cupy`, `assert_numpy_equals_dask_cupy`) are designed for pure-function-of-DataArray operators and assume the input and output share attrs, dims, and coords. They do not fit here because the matrix compares different reads of the same file rather than an op applied across backends. The matrix stays in the geotiff test tree.

## Migration

Keep the bug-numbered files as named regression markers. Do not delete them. They should keep their narrow per-bug assertions and stop accumulating new general parity cases. New parity assertions land in the matrix.

Concretely:

- `test_backend_parity_matrix.py`: extend `_BACKENDS` and `_FIXTURES`, add the error-fixture sub-matrix.
- `test_backend_pixel_parity_matrix_1813.py`: stays. Locks the dtype/compression/layout sweep specific to the #1813 refactor.
- `test_attrs_parity_1548.py`, `test_backend_kwarg_parity_1561.py`, `test_miniswhite_backend_parity_1797.py`, `test_vrt_backend_coverage_2026_05_11.py`: stay as named regression markers.

## Out of scope

- Writer parity (`to_geotiff` / `write_geotiff_gpu` / `write_vrt`). Writer matrix is a separate follow-up; `test_writer_matrix.py` is the current single point.
- Performance / throughput assertions.
- Golden-corpus byte-equality (covered by `test_golden_corpus_*_1930.py`).
- A new top-level helper module. The existing file is the home.
- Property-based / fuzz coverage (`test_fuzz_hypothesis_1661.py` already does that for one slice).

## Acceptance

- `xrspatial/geotiff/tests/test_backend_parity_matrix.py` parametrizes over all 8 backends and at least the 7 fixtures listed above.
- `pytest xrspatial/geotiff/tests/test_backend_parity_matrix.py -v` reports one cell per (fixture, backend) pair, with GPU / HTTP / fsspec cells skipping cleanly when their deps are absent.
- Adding a new backend or fixture is one row appended to `_BACKENDS` or `_FIXTURES`; no other file edits.
- The existing per-bug parity files remain green and unchanged.

brendancol:
Closes #2132.

## Summary

- Extends `test_backend_parity_matrix.py` from a 2-backend / 1-fixture scaffold to the full 8-backend / 7-fixture matrix described in #2132.
- Adds an error-fixture sub-matrix (rotated `ModelTransformationTag` without `allow_rotated`) so the contract for raised exceptions has one home.
- Backend / source-type compatibility lives on the `_BackendSpec.compat` set, so incompatible cells skip cleanly without taking the matrix down.

## Backends covered

| id | dispatch | source |
| --- | --- | --- |
| `numpy` | `open_geotiff(path)` | local TIFF / VRT / HTTP / fsspec |
| `dask+numpy` | `open_geotiff(path, chunks=16)` | local TIFF / VRT / fsspec |
| `gpu` | `open_geotiff(path, gpu=True)` | local TIFF |
| `dask+gpu` | `open_geotiff(path, gpu=True, chunks=16)` | local TIFF |
| `vrt-eager` | `open_geotiff(path.vrt)` | local VRT |
| `vrt-dask` | `open_geotiff(path.vrt, chunks=16)` | local VRT |
| `http-cog` | `open_geotiff('http://...')` | local TIFF served via loopback HTTP |
| `fsspec-memory` | `open_geotiff('memory://...')` | local TIFF piped through fsspec |

GPU rows skip when cupy or CUDA are missing. fsspec rows skip when fsspec is absent. HTTP rows reuse the loopback `http.server` pattern from `test_cog_http_concurrent.py`.

## Fixtures covered

- int16 single-band stripped (pre-existing)
- uint16 multiband tiled deflate
- float32 single-band with `-9999.0` nodata sentinel (locks the #2092 `masked_nodata=True` contract)
- int8 single-band with `mask_nodata=False` (locks the #2092 / #2127 `masked_nodata=False` contract)
- COG with one internal overview
- VRT mosaic over two tiles
- MinIsWhite single-band uint8 (gated on tifffile)
- Error: rotated `ModelTransformationTag` without `allow_rotated`

## Out of scope

- Writer parity (separate matrix in `test_writer_matrix.py`).
- Pre-existing per-bug parity files (kept as named regression markers).
- BytesIO source-type row: the fsspec `memory://` row already covers the same code path for the matrix, and the existing `test_bytesio_source.py` keeps the explicit BytesIO behaviour pinned.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_backend_parity_matrix.py -v` reports 46 passed and 18 skipped (incompatible cells).
- [x] Existing per-bug parity files (`test_backend_pixel_parity_matrix_1813.py`, `test_attrs_parity_1548.py`, `test_backend_kwarg_parity_1561.py`, `test_miniswhite_backend_parity_1797.py`) stay green.
- [x] `ruff` and `flake8` clean on the changed file.

### DIFF
--- a/xrspatial/geotiff/tests/test_backend_parity_matrix.py
+++ b/xrspatial/geotiff/tests/test_backend_parity_matrix.py
@@ -148,16 +148,17 @@ _BACKENDS: list[_BackendSpec] = [
     _BackendSpec(
         backend_id="numpy",
         kwargs={},
-        compat=frozenset({_SRC_LOCAL_TIFF, _SRC_LOCAL_VRT,
-                          _SRC_HTTP, _SRC_FSSPEC}),
+        # VRT fixtures are owned by the ``vrt-eager`` / ``vrt-dask``
+        # rows below; routing them through ``numpy`` too would
+        # duplicate identical cells.
+        compat=frozenset({_SRC_LOCAL_TIFF, _SRC_HTTP, _SRC_FSSPEC}),
     ),
     _BackendSpec(
         backend_id="dask+numpy",
         kwargs={"chunks": 16},
         # Dask path supports fsspec URIs (#1749) but does not accept
-        # raw BytesIO. .vrt files are handled by the VRT reader which
-        # itself supports chunks via read_vrt's matrix.
-        compat=frozenset({_SRC_LOCAL_TIFF, _SRC_LOCAL_VRT, _SRC_FSSPEC}),
+        # raw BytesIO. VRT lives on the ``vrt-dask`` row.
+        compat=frozenset({_SRC_LOCAL_TIFF, _SRC_FSSPEC}),
     ),
     _BackendSpec(
         backend_id="gpu",
@@ -581,12 +582,13 @@ class _MatrixRangeHandler(http.server.BaseHTTPRequestHandler):
 
 
 @pytest.fixture(scope="session")
-def _matrix_http_server():
+def _matrix_http_server_session():
     """Shared loopback HTTP server for the http-cog backend row.
 
-    Started once per pytest session; the per-test fixture below uploads
-    its payload into the handler's ``payload_by_path`` map and tears it
-    down on teardown. The server runs on a free port chosen by the OS.
+    Started once per pytest session and torn down on session exit. The
+    payload dict on the handler is cleared between tests by the
+    function-scoped ``_matrix_http_server`` wrapper below; this fixture
+    only owns the socket and the thread.
     """
     handler_cls = type(
         "MatrixRangeHandler", (_MatrixRangeHandler,),
@@ -603,10 +605,32 @@ def _matrix_http_server():
         httpd.server_close()
 
 
-def _deliver_via_http(spec: _FixtureSpec, on_disk: Path,
+@pytest.fixture
+def _matrix_http_server(_matrix_http_server_session):
+    """Function-scoped HTTP server view: clears stale payloads after each test.
+
+    Without this, the session-scoped ``payload_by_path`` dict accumulates
+    one entry per cell and never releases the bytes. Keeping it
+    function-scoped means a test only sees the URL paths it uploaded.
+    """
+    base_url, handler_cls = _matrix_http_server_session
+    handler_cls.payload_by_path.clear()
+    try:
+        yield base_url, handler_cls
+    finally:
+        handler_cls.payload_by_path.clear()
+
+
+def _deliver_via_http(spec: "_FixtureSpec | _ErrorFixtureSpec", on_disk: Path,
                       base_url: str, handler_cls,
                       monkeypatch) -> str:
-    """Upload an on-disk fixture into the shared HTTP server and return URL."""
+    """Upload an on-disk fixture into the shared HTTP server and return URL.
+
+    The success matrix passes a :class:`_FixtureSpec`; the error
+    sub-matrix passes an :class:`_ErrorFixtureSpec`. Both expose
+    ``fix_id`` so the function consumes either.
+    """
+    del spec  # the spec is unused; signature kept for symmetry with fsspec
     monkeypatch.setenv("XRSPATIAL_GEOTIFF_ALLOW_PRIVATE_HOSTS", "1")
     with open(on_disk, "rb") as f:
         payload = f.read()
@@ -615,7 +639,8 @@ def _deliver_via_http(spec: _FixtureSpec, on_disk: Path,
     return f"{base_url}{url_path}"
 
 
-def _deliver_via_fsspec(spec: _FixtureSpec, on_disk: Path) -> str:
+def _deliver_via_fsspec(spec: "_FixtureSpec | _ErrorFixtureSpec",
+                        on_disk: Path) -> str:
     """Pipe an on-disk fixture into fsspec's memory:// filesystem.
 
     Returns the ``memory://`` URI the read path should consume. The
@@ -1000,15 +1025,6 @@ def test_backend_parity_matrix_errors(
         http_state=_matrix_http_server, monkeypatch=monkeypatch,
     )
 
-    label = (
-        f"error_fixture={error_spec.fix_id} backend={backend.backend_id} "
-        f"kwargs={backend.kwargs}"
-    )
-    # ``label`` is referenced via ``__tracebackhide__``-friendly assert
-    # messages on failure (pytest already prints the cell id in the
-    # parametrize header, but keeping the local mirrors the success
-    # matrix's assertion style).
-    del label
     with pytest.raises(error_spec.exc, match=error_spec.match):
         # ``open_geotiff`` may return lazily for chunked reads, so
         # force a materialisation inside the ``pytest.raises`` block
--- a/xrspatial/geotiff/tests/test_backend_parity_matrix.py
+++ b/xrspatial/geotiff/tests/test_backend_parity_matrix.py
@@ -1,4 +1,4 @@
-"""Required backend parity matrix per high-risk fixture (issue #1985).
+"""Required backend parity matrix per high-risk fixture (issue #1985, #2132).
 
 Single source of truth for "does backend X still match the reference on
 fixture Z." Existing scattered parity files (``test_attrs_parity_1548.py``,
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: geotiff: extend backend parity matrix to 8 backends and 7 fixtures

None.

- [ ] **Dead `label` variable in the error test** (`xrspatial/geotiff/tests/test_backend_parity_matrix.py:1003-1011`): the error matrix builds a `label` string and immediately `del`s it. The pytest parametrize id already carries that info. Drop the assignment, drop the `del`, drop the comment about tracebackhide-friendly messages. There is no assertion that uses `label` in this block; `pytest.raises` is doing the work.

- [ ] **Two of the VRT cells are duplicates** (`test_backend_parity_matrix.py:154-186`): `dask+numpy` has `_SRC_LOCAL_VRT` in its compat set, so `dask+numpy-vrt-mosaic` runs with `kwargs={"chunks": 16}` against the VRT path. `vrt-dask` runs with the same kwargs against the same path. Same story for `numpy-vrt-mosaic` vs `vrt-eager-vrt-mosaic` (both `kwargs={}`). These cells exercise the same code. The issue's plan names `vrt-eager` and `vrt-dask` as the dedicated VRT rows, so the cleanest fix is to trim `_SRC_LOCAL_VRT` out of the `numpy` and `dask+numpy` compat sets and let the VRT-named backends own VRT.

- [ ] **Type hints on `_deliver_via_http` / `_deliver_via_fsspec`** (`test_backend_parity_matrix.py:606-632`): both functions are annotated as taking `_FixtureSpec`, but the error matrix passes `_ErrorFixtureSpec` instances. Duck typing on `fix_id` covers it. Widen the annotation to `_FixtureSpec | _ErrorFixtureSpec` or drop the constraint.

- [ ] **Session-scoped HTTP server keeps payloads forever** (`test_backend_parity_matrix.py:583-603`): the `payload_by_path` dict on the shared handler is appended to per cell and never trimmed. Memory cost is tiny (each fixture is a few KB), but a per-test clear would be tidier. Skip if it never shows up as overhead.

### What looks good

- Backend and fixture descriptors are separated. Adding a new backend or fixture is a single row plus an optional builder, which is what the issue asked for.
- The per-backend `compat` set is a good way to express (backend, source) incompatibility. Skip messages name the axis that cut the cell.
- Pixel comparison is NaN-aware for float and byte-equal for int. dtype is checked against the spec rather than just the reference, so a uniform silent upcast across every backend still fails the cell.
- `masked_nodata` is asserted on both the `True` (float32 with sentinel) and `False` (int8 with `mask_nodata=False`) sides. That locks both branches of the #2092 / #2127 contract.
- The error sub-matrix mirrors the success matrix, so a new error fixture is one entry.
- 46 cells pass and 18 skip cleanly (incompatible pairs). Existing per-bug parity files stay green.

### Checklist
- [x] Algorithm matches the issue spec (#2132)
- [x] NaN handling is correct (NaN-aware float comparison)
- [x] Edge cases covered (nodata sentinel, multiband, COG, MinIsWhite, rotated error)
- [ ] N/A: no dask chunk-boundary logic in this PR (test harness only)
- [ ] N/A: no benchmark needed for a test harness
- [ ] N/A: README feature matrix unchanged (no new public function)

brendancol (commented):
## Follow-up: review findings addressed

All four items from the previous review pass have been applied in df3b8eb:

- **Dead `label` variable**: removed. The block now contains only `pytest.raises` and the inner read.
- **Duplicate VRT cells**: `_SRC_LOCAL_VRT` is no longer in the compat sets for `numpy` and `dask+numpy`. The VRT fixture is now routed only through `vrt-eager` and `vrt-dask`, matching the issue's named-row layout. Cell count moves from 46 passed / 18 skipped to 44 passed / 20 skipped.
- **Type hints**: `_deliver_via_http` and `_deliver_via_fsspec` are annotated as `_FixtureSpec | _ErrorFixtureSpec` so the error matrix call sites are covered by the signature.
- **HTTP payload accumulation**: split the session-scoped server into `_matrix_http_server_session` (owns the socket + thread) and a function-scoped `_matrix_http_server` wrapper that clears `payload_by_path` before and after each test.

Local test run: 44 passed, 20 skipped on the matrix file; the per-bug regression markers (`test_backend_pixel_parity_matrix_1813.py`, `test_attrs_parity_1548.py`, `test_backend_kwarg_parity_1561.py`, `test_miniswhite_backend_parity_1797.py`) stay at 212 passed.

PR #2143, opened 2026-05-19
geotiff: gate non-uniform-coord check on no-georef marker, drop dtype exemption (#2133)

### LINKED ISSUE #2133: geotiff: finish migration from integer coord sentinel to explicit no-georef attr
## Background

The geotiff round-trip used to detect "this file has no GeoTIFF transform tags" by looking at coord shape: `int64` ascending step-1 on both y/x axes meant "no georef" and the writer skipped transform synthesis. That heuristic was tightened in #2087 and then replaced in #2120 (commit 34cc415) with an explicit attr `attrs['_xrspatial_no_georef'] = True`.

This issue tracks the rest of the migration: one validator still infers no-georef from coord dtype, the shape helper is now test-only, and the round-trip test matrix could pin a couple more edge cases.

## What 34cc415 already did

- Added `_NO_GEOREF_KEY = '_xrspatial_no_georef'` in `xrspatial/geotiff/_geotags.py`.
- Stamped the marker on every read path that emits placeholder coords:
  - CPU reader: `_populate_attrs_from_geo_info` in `xrspatial/geotiff/_attrs.py` (else branch when `geo_info.transform is None`).
  - VRT reader: both eager and chunked no-transform branches in `xrspatial/geotiff/_backends/vrt.py`.
  - GPU / dask readers route through `_populate_attrs_from_geo_info` so they pick the marker up for free.
- Switched the writer-side detection to `_has_no_georef_marker(da)` in two places in `xrspatial/geotiff/_coords.py`:
  - `coords_to_transform` (return `None` instead of synthesising a fake unit transform).
  - `require_transform_for_georeferenced` (skip the fail-closed guard).
- Kept `_is_no_georef_sentinel` in the module as a diagnostic but wired it out of the writer path.
- Added `test_no_georef_marker_2120.py` and updated `test_int_coord_sentinel_2087.py` and `test_no_georef_writer_round_trip_1949.py` to pin the new contract.

### `_check_write_non_uniform_coords` in `_validation.py`

The validator at line 755 skips its uniformity check whenever a coord has integer dtype, with the comment "those coords are the 0..N-1 fallback the no-georef reader emits". After #2120 that exemption is wrong for the same reason the writer's old shape check was wrong: a user-authored int-coord grid (the exact case #2087 fixed) now bypasses uniformity validation. A user with int coords on a non-uniform grid will silently write a misrepresented transform.

The validator should gate the exemption on `_has_no_georef_marker(da)`, not on `dtype.kind`. Float coords with the marker set should also be exempt for symmetry with the writer.

### `_is_no_georef_sentinel` is dead on the production path

It only survives because `test_int_coord_sentinel_2087.py` still references it as a predicate. Two options:

1. Drop the helper and rewrite the tests to check `_has_no_georef_marker` instead.
2. Move the helper into the test file as a local fixture.

I lean toward option 1. The helper's docstring already says it is test-only, and keeping public-module dead code makes it easier for someone to plug it back into a write path by accident.

## Attribute name and the `georef_status` proposal

A separate issue is being filed to propose a richer `attrs['georef_status']` enum (`'georef' | 'no_georef' | 'rotated' | ...`) that would replace ad-hoc booleans across the geotiff module. If that lands first, the implementation here should consume `georef_status == 'no_georef'` rather than the dedicated `_xrspatial_no_georef` flag, and `_has_no_georef_marker` should be the single point of indirection.

If this issue lands first, keep `_xrspatial_no_georef` as an underscore-prefixed internal contract and migrate it under the `georef_status` umbrella later. Either way, callers should never read `_xrspatial_no_georef` directly outside the geotiff module.

## Read paths to audit

Each path that emits placeholder y/x coords must stamp the marker. 34cc415 covered the three known paths; double-check that no new reader has appeared since:

- `xrspatial/geotiff/_attrs.py`: `_populate_attrs_from_geo_info` else branch.
- `xrspatial/geotiff/_backends/vrt.py`: `read_vrt` and `_read_vrt_chunked` no-transform branches.

GPU (`_backends/gpu.py`) and dask (`_backends/dask.py`) readers both call `_populate_attrs_from_geo_info`, so they inherit the marker. A unit test that opens a no-georef fixture through each backend and asserts the marker is present would catch any future regression.

## Write paths to audit

Each writer must consult the marker only, never coord shape:

- `xrspatial/geotiff/_writers/eager.py`: `to_geotiff` (line 565-568) and the VRT per-tile write (line 946-948). Both call `_coords_to_transform(data)`, which already checks the marker.
- `xrspatial/geotiff/_writers/gpu.py`: `write_geotiff_gpu` (line 425-427). Same path.
- `xrspatial/geotiff/_writer.py` low-level `write` accepts a pre-computed `geo_transform`, so the marker decision happens upstream.

The validator fix above is the missing piece.

## Round-trip tests to add

- A user grid with int coords on a non-uniform spacing should raise `NonUniformCoordsError` even though the dtype is int. Today it silently writes a misrepresented transform via the validator exemption.
- A no-georef file opened through each of the four backends (CPU eager, CPU dask, GPU eager, GPU dask) carries the marker.
- Marker survives a `da.copy()` / `da.assign_attrs(...)` chain. xarray attr-propagation defaults sometimes drop underscore-prefixed keys, so this is worth pinning explicitly.
- Marker survives a `to_geotiff(da, path)` followed by `open_geotiff(path)` on a hand-built no-georef array (already covered by `test_marker_on_user_grid_skips_transform_synthesis` in #2120; add a corresponding 3D test).

## Parallel with PR #2127 / commit 478dbd5

PR #2127 threaded the `masked_nodata` decision through `_set_nodata_attrs` so the writer reads an explicit attr instead of inferring intent from dtype. Same pattern applies here: the no-georef decision lives on `attrs[_NO_GEOREF_KEY]`, the writer reads that attr only, and the validator should follow suit.

Both attrs share an invariant worth documenting in `docs/source/user_guide/attrs_contract.rst`: they are read-side-stamped, internal, and `is True` identity-checked so a stray third-party stamp cannot flip behaviour.

## Acceptance criteria

- [ ] `_check_write_non_uniform_coords` checks the marker, not dtype.
- [ ] `_is_no_georef_sentinel` removed from `_coords.py`; tests rewritten to use `_has_no_georef_marker` or a local predicate.
- [ ] New test: user int-coord grid with non-uniform spacing raises `NonUniformCoordsError`.
- [ ] New test: marker is present after reading a no-georef fixture through each backend.
- [ ] Decision recorded on the `georef_status` issue about which attr is canonical long-term.

Follow-up to #2120.

brendancol:
## Summary

- Threads `no_georef_marker` through the write-metadata validation
  context from the three writer entry points (`to_geotiff` eager CPU,
  GPU `write_geotiff_gpu`, dask-VRT streaming).
- `_check_write_non_uniform_coords` now gates the exemption on the
  marker instead of `dtype.kind`. A user-authored int-coord grid with
  non-uniform spacing raises `NonUniformCoordsError`; a marker-stamped
  placeholder grid still skips the check.
- Drops `_is_no_georef_sentinel` from `_coords.py`. The predicate was
  test-only after #2120; the shape check moves into
  `test_int_coord_sentinel_2087.py` as a local helper.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy. All four write paths pass the
marker into the validator.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/` (CPU paths) passes locally
- [x] New `test_no_georef_attr_migration_2133.py` covers the validator
      gap, marker presence on every read backend, marker survival
      across `da.copy()` / `da.assign_attrs(...)`, and a 3D round-trip
- [x] `test_int_coord_sentinel_2087.py::test_non_uniform_int_coords_raise`
      still passes (regex relaxed because the validator now catches the
      case before `coords_to_transform`)
- [ ] GPU read-marker tests skip on this box (no CUDA); CI runs them

Closes #2133.

### DIFF
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -113,7 +113,12 @@ write.
        fake unit transform. Absence of the marker means the array
        has spatial coords the writer can interpret as georef. A
        caller can opt into no-georef writes on a hand-built array
-       by setting this attr explicitly. See issue #2120.
+       by setting this attr explicitly. The writer uses an identity
+       check (``attrs[_xrspatial_no_georef] is True``), so only the
+       exact boolean ``True`` flips the no-georef path; truthy
+       strings like ``'yes'`` or ``1`` are ignored and the writer
+       proceeds with normal transform synthesis. See issues #2120
+       and #2133.
 
 
 Compatibility aliases
--- a/xrspatial/geotiff/_coords.py
+++ b/xrspatial/geotiff/_coords.py
@@ -24,6 +24,7 @@ windowed vs full reads.
 from __future__ import annotations
 
 from collections.abc import Mapping
+from typing import Any
 
 import numpy as np
 import xarray as xr
@@ -38,7 +39,7 @@ from ._geotags import _NO_GEOREF_KEY, GeoTransform, RASTER_PIXEL_IS_POINT
 _BAND_DIM_NAMES = ('band', 'bands', 'channel')
 
 
-def _has_no_georef_marker(da) -> bool:
+def _has_no_georef_marker(da: Any) -> bool:
     """True iff ``da`` was stamped by the reader as carrying no georef.
 
     The reader sets ``attrs[_NO_GEOREF_KEY] = True`` whenever it emits
--- a/xrspatial/geotiff/_coords.py
+++ b/xrspatial/geotiff/_coords.py
@@ -23,6 +23,8 @@ windowed vs full reads.
 """
 from __future__ import annotations
 
+from collections.abc import Mapping
+
 import numpy as np
 import xarray as xr
 
@@ -36,7 +38,7 @@ from ._geotags import _NO_GEOREF_KEY, GeoTransform, RASTER_PIXEL_IS_POINT
 _BAND_DIM_NAMES = ('band', 'bands', 'channel')
 
 
-def _has_no_georef_marker(da: xr.DataArray) -> bool:
+def _has_no_georef_marker(da) -> bool:
     """True iff ``da`` was stamped by the reader as carrying no georef.
 
     The reader sets ``attrs[_NO_GEOREF_KEY] = True`` whenever it emits
@@ -50,40 +52,16 @@ def _has_no_georef_marker(da: xr.DataArray) -> bool:
     boolean ``True`` flips the writer into no-georef mode. A stray
     third-party stamp like ``attrs['_xrspatial_no_georef'] = 'yes'``
     should not be treated as truthy and silently drop a transform.
+
+    Inputs that are not xarray DataArrays (e.g. a raw ``numpy.ndarray``
+    or ``cupy.ndarray`` passed directly to ``write_geotiff_gpu``) carry
+    no attrs and therefore no marker; return ``False`` rather than
+    raise so callers can use this as a plain predicate.
     """
-    return da.attrs.get(_NO_GEOREF_KEY) is True
-
-
-# Kept for diagnostic / test pinning only. The writer no longer
-# calls this helper -- it checks ``attrs[_NO_GEOREF_KEY]`` instead
-# (see :func:`_has_no_georef_marker` / issue #2120). Only the tests
-# in ``test_int_coord_sentinel_2087.py`` reference it now.
-def _is_no_georef_sentinel(coord: np.ndarray) -> bool:
-    """True iff ``coord`` matches the read-side no-georef placeholder shape.
-
-    ``coords_from_pixel_geometry`` emits ``np.arange(start, stop,
-    dtype=np.int64)`` for the y/x coords whenever the source file
-    carries no GeoTIFF transform tags -- both for full reads
-    (``start=0``) and windowed reads (``start=window_offset``). See
-    issues #1710, #1753, #1949.
-
-    The writer no longer treats coord shape alone as the no-georef
-    signal: it checks ``attrs[_NO_GEOREF_KEY]`` instead. See
-    :func:`_has_no_georef_marker` and issue #2120. This helper remains
-    available as a diagnostic for the placeholder shape, and the
-    existing tests in ``test_int_coord_sentinel_2087.py`` still pin
-    the predicate. It is no longer called from the writer path, so
-    user-authored int64 step-1 grids that match this pattern but lack
-    the marker keep their georef on round-trip.
-    """
-    if coord.dtype != np.int64:
+    attrs = getattr(da, 'attrs', None)
+    if not isinstance(attrs, Mapping):
         return False
-    n = len(coord)
-    if n < 1:
-        return False
-    return bool(np.array_equal(
-        coord, np.arange(coord[0], coord[0] + n, dtype=np.int64)
-    ))
+    return attrs.get(_NO_GEOREF_KEY) is True
 
 
 def coords_from_pixel_geometry(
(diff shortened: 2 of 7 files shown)


brendancol (commented):
## PR Review: geotiff: gate non-uniform-coord check on no-georef marker, drop dtype exemption (#2133)

### Blockers

None.

### Suggestions

- `docs/source/user_guide/attrs_contract.rst:107-116`: the
  `_xrspatial_no_georef` entry covers round-trip semantics but skips
  the identity-check invariant (`is True` only). The issue body
  flagged this as worth pinning so a third-party stamp like
  `'_xrspatial_no_georef': 'yes'` can't flip writer behaviour by
  accident. One line would close that gap.

### Nits

- `xrspatial/geotiff/_coords.py:41`: `_has_no_georef_marker(da)` lost
  its `xr.DataArray` annotation when it became tolerant of raw
  numpy / cupy arrays. The docstring covers the new contract, but
  an `Any` annotation would keep the signature self-documenting.
- `xrspatial/geotiff/_validation.py:765-768`: the dtype guard now
  accepts both `np.floating` and `np.integer`. `np.bool_` is a
  subdtype of integer in numpy, so a bool coord array would flow
  through the diff math. Bool coords would be very unusual; mention
  rather than fix.

### What looks good

- Marker is threaded through all three writer entry points
  (`to_geotiff`, `_write_vrt_tiled`, `write_geotiff_gpu`) the same way.
- Validator exemption now matches the writer's `_has_no_georef_marker`
  call, so the two stay in sync.
- `_has_no_georef_marker` handles raw numpy / cupy arrays passed
  straight to `write_geotiff_gpu`. The GPU test suite caught the
  earlier `AttributeError` and the fallback was added before push.
- Sentinel helper moved into the test file as a local predicate.
  Production code no longer carries dead diagnostic code.
- New test file covers the four acceptance criteria from the issue:
  validator gates on marker, marker present on every read backend,
  marker survives `copy()` / `assign_attrs()`, 3D round-trip.
- Existing `test_non_uniform_int_coords_raise` regex relaxed
  correctly. The validator now catches the case before
  `coords_to_transform` does, so the new message is the authoritative
  one.

### Checklist

- [x] Algorithm matches the #2133 acceptance criteria
- [x] All write backends thread the marker consistently
- [x] NaN handling: not applicable for this change
- [x] Edge cases covered by tests (3D, copy, assign_attrs, marker exemption)
- [x] Dask boundaries: not applicable (metadata-only change)
- [x] Benchmark not needed for a validator change
- [x] README feature matrix: not applicable (no new public API)

brendancol (commented):
## Follow-up: review dispositions

### Suggestion (fixed)

- `docs/source/user_guide/attrs_contract.rst`: added the identity-check invariant note. The entry now spells out that only the exact boolean ``True`` flips the writer into no-georef mode, with examples of stamps the writer ignores. Closes the gap from the original review.

### Nit (fixed)

- `xrspatial/geotiff/_coords.py:_has_no_georef_marker`: restored a type annotation. `da: Any` makes the predicate's tolerance of non-DataArray inputs explicit in the signature, not just the docstring.

### Nit (dismissed)

- `xrspatial/geotiff/_validation.py:765-768` bool-coord concern: false alarm. `np.issubdtype(np.bool_, np.integer)` returns ``False`` in numpy, so the existing dtype guard already rejects bool coord arrays. No code change needed. The follow-up commit message records the reason.

CPU-side test suites (`test_no_georef_attr_migration_2133.py`, `test_int_coord_sentinel_2087.py`, `test_no_georef_marker_2120.py`, `test_ambiguous_metadata_hooks_1987.py`) re-run green after the follow-up commit.

Issue #2136, opened 2026-05-19
geotiff: canonical georef_status attribute to disambiguate CRS vs transform presence


brendancol:
## Problem

The reader currently carries three independent metadata signals and downstream code has to reconcile them by hand:

- `attrs['crs']` / `attrs['crs_wkt']` from CRS-related GeoKeys
- `attrs['transform']` from `ModelPixelScale` / `ModelTiepoint` / `ModelTransformationTag`
- `attrs[_NO_GEOREF_KEY]` (= `'_xrspatial_no_georef'`), stamped by `_populate_attrs_from_geo_info` in `xrspatial/geotiff/_attrs.py` when no transform tags are present

A raster can have CRS but no transform, transform but no CRS, both, or neither. Consumers today have to infer the state from absence-or-presence of three keys plus a sentinel attr, and the mapping is not 1:1 with what the reader actually saw:

1. `_populate_attrs_from_geo_info` (`_attrs.py:381-394`) drops `attrs['transform']` and stamps `attrs[_NO_GEOREF_KEY] = True` when `has_georef=False`. But `has_georef=False` is also what `_extract_transform(..., allow_rotated=True)` returns for a rotated `ModelTransformationTag` (`_geotags.py:73-101`, per #2116), so the rotated-but-dropped case and the truly-no-transform case look identical downstream.
2. `attrs['crs_wkt']` can still be present with neither `attrs['crs']` (no EPSG resolve) nor `attrs['transform']` (no transform tags). The VRT path handles this explicitly in `_backends/vrt.py:284-289`: stamp `_NO_GEOREF_KEY`, then set `crs_wkt`. Same shape on disk, different downstream meaning.
3. `_resolve_crs_to_wkt` (`_crs.py:181-269`) accepts a transform-less write path but does not surface that the source had no transform; the writer ends up reconstructing georef from coords (`_coords.py:296-331`), which is exactly the case `_NO_GEOREF_KEY` was added to suppress (#2120, #2124).
4. Round-trip: the no-georef marker is preserved on write (per tests in `tests/test_int_coord_sentinel_2087.py` and `tests/test_no_georef_writer_round_trip_1949.py`), but the rotated-dropped case has no equivalent marker. A reader that opens a file written by `to_geotiff` from a rotated-dropped DataArray gets an axis-aligned identity transform back with no signal of the original rotation.

## Proposal

One canonical attr that encodes the five distinct states the reader can land in:

```python
attrs['georef_status'] = (
[...]

Issue #2135, opened 2026-05-19
geotiff: split overloaded masked_nodata into separate nodata lifecycle signals


brendancol:
## Problem

`attrs['masked_nodata']` is a single boolean trying to describe a multi-stage process. The fix in 478dbd5 (PR #2127, closing #2092) tightened it so the attr now means "the reader replaced sentinel pixels with NaN," which is better than the pre-fix "final dtype is float." Callers still cannot answer four distinct questions from one bit:

1. Did the source file declare a sentinel?
2. Did the read path run the sentinel-to-NaN masking step?
3. Were sentinel pixels actually found in the read window?
4. Did a user-requested dtype cast happen after the masking decision (which can promote int to float without masking)?

Today (1) is encoded by the presence of `attrs['nodata']`, (2) is `attrs['masked_nodata']`, and (3) and (4) are not surfaced. Where each one matters in practice:

**(1) declared sentinel exists.** A round-trip writer needs the file-declared sentinel so it can preserve it on write even if the in-memory array has no missing pixels. `attrs['nodata']` already covers this.

**(2) masking step ran.** A consumer deciding between `np.isnan(arr)` and `arr == attrs['nodata']` needs this. That is what `masked_nodata` means after #2092.

**(3) sentinel pixels present.** A QA function asking "does this tile contain any nodata" currently has to scan the buffer. With a bool set at read time, the answer is free. Same problem on the writer: if no sentinel pixels were touched, the unmask-before-write scan can be skipped. Today the writer pays for it every time.

**(4) dtype cast after masking.** Caller passes `dtype=float64` on an integer source with `mask_nodata=False`. Post-mask the buffer is integer with literal sentinels; post-cast it is float64 with literal sentinels (no NaNs). The #2092 follow-up reads pre-cast dtype so `masked_nodata` reports False here, but the cast itself is invisible downstream. A consumer wanting to know "float because promoted, or float because masked" cannot tell.

## Proposed attribute scheme

Split the lifecycle into separate signals. Keep `nodata` and `masked_nodata` for backward compatibility, add two new attrs:

- `nodata` (unchanged): declared file sentinel as a scalar of the source dtype. `None`/absent means no declared sentinel.
- `masked_nodata` (unchanged after #2092): bool. True iff the reader replaced sentinel pixels with NaN. Only emitted when `nodata is not None`.
[...]

PR #2144, opened 2026-05-19
geotiff: add internal GeoTIFFMetadata dataclass and migrate readers (#2139)

### LINKED ISSUE #2139: geotiff: introduce internal GeoTIFFMetadata dataclass; convert to/from attrs at API boundary
## Problem

Every read path in the geotiff module funnels through `_populate_attrs_from_geo_info`, and every writer through `_resolve_nodata_attr` / `_extract_rich_tags`. The shared surface is still a plain `dict[str, Any]`. Each backend builds and reads that dict by hand, and we have a tail of drift bugs plus a tail of helpers that exist only to paper over the dict shape. A small internal dataclass would let readers, writers, validators, and the VRT path pass a typed record around, and only marshal to/from `attrs` at the public API boundary.

Concrete drift cases from the current tree:

1. Two read paths bypass `_populate_attrs_from_geo_info` and build the attrs dict inline. `xrspatial/geotiff/_backends/vrt.py:275` and `xrspatial/geotiff/_backends/vrt.py:697` open `attrs = {'_xrspatial_geotiff_contract': _ATTRS_CONTRACT_VERSION}` and then set `attrs['crs']`, `attrs['crs_wkt']`, `attrs['transform']`, `attrs['raster_type']`, `attrs[_NO_GEOREF_KEY]`, and `attrs['vrt_holes']` directly. The eager, dask, and GPU paths funnel through the helper instead (`__init__.py:633`, `_backends/dask.py:342`, `_backends/gpu.py:437`, `:816`, `:1369`). Two of the four read paths (and any future fifth) must remember to stamp the contract version, set `crs` only when an EPSG resolves, set `crs_wkt` whenever any CRS is known, set `raster_type='point'` only on point rasters, and apply the no-georef marker on missing transform tags. Nothing in the type system enforces that list.

2. The reader's `_populate_attrs_from_geo_info` emits `attrs['crs']` only when `geo_info.crs_epsg is not None` and `attrs['crs_wkt']` only when `geo_info.crs_wkt is not None` (`_attrs.py:364-367`). The two VRT read paths run `epsg = _wkt_to_epsg(vrt.crs_wkt)` inline and only emit `attrs['crs']` when that returns non-None (`_backends/vrt.py:285-288`, `:715-718`). Three writers re-implement the resolve-from-attrs step from scratch (`_writers/eager.py:574-592`, `_writers/eager.py:913-931`, `_writers/gpu.py:438-456`), each duplicating the "attrs['crs'] may be int EPSG or WKT string, attrs['crs_wkt'] only carries WKT" branch tree. The duplicated logic has already drifted: `_writers/eager.py:585` validates `crs_attr` via `_validate_crs_arg` before coercing to int, while `_backends/vrt.py:287` does not validate at all (the VRT path constructs the int from `_wkt_to_epsg` and never sees a user-supplied `attrs['crs']`). Any future tightening of that validator has to land in three places.

3. The `nodata` / `masked_nodata` pair is split across `_set_nodata_attrs` and a per-backend `masked=` computation. The eager path sets `masked=(mask_nodata and arr.dtype.kind == 'f')` against the final array (`__init__.py:697-700`). The dask path sets `masked=(mask_nodata and target_dtype.kind == 'f')` against the declared graph dtype (`_backends/dask.py:350-353`). The VRT eager path captures a `pre_cast_dtype` before the optional user cast and uses `masked=(pre_cast_dtype.kind == 'f')` (`_backends/vrt.py:341`, `:366-369`). The VRT chunked path uses `masked=(declared_dtype.kind == 'f')` for the same reason (`_backends/vrt.py:743-746`). All four were introduced independently under issue #2092 because the pre-fix code inferred `masked` from the final dtype, which lied when `mask_nodata=False` left literal sentinel values in a float buffer. A typed record would put the "masked is the actual mask-decision the read path made" invariant in one place rather than four.

4. Pass-through attrs ride on the same dict. `attrs['vrt_holes']` exists only on the VRT path (`_backends/vrt.py:300`, `:774`). `attrs['extra_tags']` and `attrs['colormap']` exist on every path except VRT (`_attrs.py:412-413`, `:438`). `attrs['gdal_metadata_xml']` and `attrs['gdal_metadata']` exist on every path except VRT (`_attrs.py:407-410`). A typed record makes the per-backend coverage explicit and turns silent omissions into type-checker errors.

5. Writers resolve `nodata` from attrs via `_resolve_nodata_attr` (`_attrs.py:442-513`), which accepts three aliases (`nodata`, `nodatavals`, `_FillValue`). The companion validator `_validate_attrs_nodata_consistency` (`_validation.py:636`) refuses writes where `attrs['nodata']` and `attrs['nodatavals']` disagree. The same logic is duplicated by hand in three writers via `if nodata is None: nodata = _resolve_nodata_attr(data.attrs)` (`_writers/eager.py:593-594`, `_writers/eager.py:932-941`, `_writers/gpu.py:457-458`). A typed record built once at the API boundary lets those three writers take a `GeoTIFFMetadata` parameter and drop the resolve step.

## The dataclass shape

```python
from dataclasses import dataclass, field
from typing import Any

@dataclass(frozen=True)
class GeoTIFFMetadata:
    # Spatial reference
    transform: tuple[float, ...] | None = None   # rasterio 6-tuple
    crs_epsg: int | None = None                  # surfaces as attrs['crs']
    crs_wkt: str | None = None                   # surfaces as attrs['crs_wkt']
    raster_type: str = 'area'                    # 'area' | 'point'
    has_georef: bool = True                      # False -> stamps _NO_GEOREF_KEY

    # NoData semantics (issue #1988 / #2092)
    nodata: int | float | None = None            # declared file sentinel
    masked_nodata: bool | None = None            # True iff sentinel-to-NaN ran

    # Pass-through TIFF tags
    extra_tags: list[tuple] | None = None
    image_description: str | None = None
    extra_samples: tuple[int, ...] | None = None
    colormap: tuple[int, ...] | None = None

    # GDAL_METADATA
    gdal_metadata: dict | None = None
    gdal_metadata_xml: str | None = None

    # Resolution tags
    x_resolution: float | None = None
    y_resolution: float | None = None
    resolution_unit: str | None = None           # 'none' | 'inch' | 'centimeter'

    # VRT-only
    vrt_holes: list[dict] | None = None

    # Contract version
    contract_version: int = 2
```

The keys map 1:1 onto the existing attrs contract in `_attrs.py:18-122`, so external behaviour is unchanged. Field notes:

- `transform`, `crs_epsg`, `crs_wkt`, `raster_type`, and `has_georef` are the spatial reference fields. `has_georef=False` is the in-memory form of `attrs[_NO_GEOREF_KEY] = True` (`_attrs.py:394`); making it a bool means the writer's "did this come from a real georef" check is `md.has_georef` instead of `_is_no_georef_array(da)`.
- `nodata` and `masked_nodata` follow the issue #1988 split. `nodata` is the declared file sentinel. `masked_nodata` is `True` iff the in-memory array went through the sentinel-to-NaN step. `masked_nodata` is `None` (not `False`) when no sentinel was declared, mirroring the current "absence of `attrs['masked_nodata']` means no declared sentinel" rule (`_attrs.py:248-278`).
- `extra_tags`, `image_description`, `extra_samples`, and `colormap` are the pass-through tags that `_extract_rich_tags` folds back into the writer kwargs (`_attrs.py:573-621`). Keeping them on the dataclass means `_merge_friendly_extra_tags` can take a `GeoTIFFMetadata` and return one.
- `gdal_metadata` and `gdal_metadata_xml` stay as separate fields because `_extract_rich_tags` already prefers the XML form over the dict form (`_attrs.py:599-604`); merging them would lose that precedence.
- `vrt_holes` is VRT-only today and stays VRT-only on the dataclass.
- `contract_version` lives on the record so a future v3 reader can produce a v3 record and the conversion layer can decide what to stamp.

## Conversion functions

Two functions at the API boundary, both in `_attrs.py`:

```python
def metadata_to_attrs(md: GeoTIFFMetadata) -> dict[str, Any]:
    """Build the public attrs dict from a metadata record.
    Mirrors the current _populate_attrs_from_geo_info field-by-field."""

def attrs_to_metadata(attrs: dict[str, Any]) -> GeoTIFFMetadata:
    """Parse a (possibly user-supplied) attrs dict into a metadata record.
    Honours the alias resolution _resolve_nodata_attr already implements."""
```

`metadata_to_attrs` replaces the seven manual `attrs[...] = ...` sites in `_populate_attrs_from_geo_info` (`_attrs.py:362-439`) and the two inline blocks at `_backends/vrt.py:275-300` and `:697-720`.

`attrs_to_metadata` replaces the three duplicated "read crs / crs_wkt / nodata off attrs" blocks in the writers (`_writers/eager.py:574-611`, `:913-941`, `_writers/gpu.py:438-460`).

## Internal call sites

Functions that stop touching `attrs` directly and take/return `GeoTIFFMetadata`:

Read side:
- `_populate_attrs_from_geo_info(attrs, geo_info, *, window=None)` becomes `geo_info_to_metadata(geo_info, *, window=None) -> GeoTIFFMetadata`. Callers in `__init__.py:633`, `_backends/dask.py:342`, `_backends/gpu.py:437`, `:816`, and `:1369` switch to the new function and call `metadata_to_attrs(md)` immediately before constructing the DataArray.
- `_set_nodata_attrs(attrs, nodata, *, masked)` becomes a builder method on `GeoTIFFMetadata`, e.g. `md = md.with_nodata(nodata, masked=...)`. Callers in `__init__.py:697`, `_backends/dask.py:350`, `_backends/gpu.py:453`, `:817`, `_backends/vrt.py:366`, and `:743` switch to the builder.
- The two VRT inline blocks (`_backends/vrt.py:275-300`, `:697-720`) build a `GeoTIFFMetadata` directly from `vrt.crs_wkt` / `vrt.geo_transform` / `vrt.bands` / `vrt.holes` and call `metadata_to_attrs` at the DataArray-construction step.

Write side:
- `_resolve_nodata_attr(attrs)` becomes a no-op when the writer takes a `GeoTIFFMetadata`: the field is already resolved.
- `_extract_rich_tags(attrs)` becomes `_extract_rich_tags(md)`. The three writers (`_writers/eager.py:605`, `:932-941` via `_resolve_nodata_attr`, `_writers/gpu.py:458`) call `md = attrs_to_metadata(data.attrs)` once at entry and read fields off `md` instead of re-resolving from `data.attrs`.
- `_should_restore_nan_sentinel(attrs)` becomes a one-liner on the record (`md.masked_nodata is not False`).

Validators:
- `validate_write_metadata` (`_validation.py`) already takes a dict literal of `crs_kwarg`, `attrs_crs`, `attrs_crs_wkt`, `nodata_kwarg`, `attrs_nodata`, `attrs_nodatavals`, `coord_y`, `coord_x`. That dict is built in three call sites (`_writers/eager.py:298-307`, `:856-865`, `_writers/gpu.py:318-327`); they all collapse to `validate_write_metadata(md, crs_kwarg=crs, nodata_kwarg=nodata, coord_y=..., coord_x=...)`.

## Migration plan

Internal refactor; the public attrs surface does not change. Each step is independently testable:

1. Land `GeoTIFFMetadata` and the two conversion functions in `_attrs.py`. Add round-trip tests asserting `metadata_to_attrs(attrs_to_metadata(x)) == x` for representative attrs dicts (eager numpy, dask, GPU, VRT, no-georef, point raster, user-defined CRS WKT).

2. Rewrite `_populate_attrs_from_geo_info` as a thin wrapper: `attrs.update(metadata_to_attrs(geo_info_to_metadata(geo_info, window=window)))`. The existing helper signature is preserved so every reader call site keeps working. Tests in `test_round_trip_invariants.py` and `test_nodata_semantics_split_1988.py` cover the read paths; they should pass unchanged.

3. Replace the inline attrs blocks in `_backends/vrt.py:275-300` and `:697-720` with the new builder. That is the highest-drift surface today, so landing it second clears the bulk of the consistency gap.

4. Switch the four read-side `_set_nodata_attrs` callers to the builder method. The four `masked=` computations stay where they are; the dataclass only changes the assignment shape.

5. Land `attrs_to_metadata` on the write side. The three writers gain a `md = attrs_to_metadata(data.attrs)` call at entry, and the duplicated CRS / nodata / rich-tag resolve blocks collapse to field lookups on `md`. `validate_write_metadata` switches to taking `md` directly.

6. Once all read and write paths use `GeoTIFFMetadata` internally, deprecate the `attrs` parameter on `_resolve_nodata_attr`, `_extract_rich_tags`, `_should_restore_nan_sentinel`, and `_set_nodata_attrs`. The compatibility shims stay until one minor release after the dataclass lands.

Each step is a separate PR so the diff stays reviewable.

## Relation to sibling issues

The dataclass is where three companion proposals on the tracker live:

- `georef_status` (the proposal to replace `attrs[_NO_GEOREF_KEY] = True` with an enum-valued attr) becomes `md.has_georef`, or `md.georef_status: Literal['present', 'absent', 'placeholder']` if the enum form wins. Either way the field is named once, set once by the readers, and consumed once by the writers and `_is_no_georef_array`.
- `_xrspatial_no_georef`: the underscore-prefix attr stays on the public surface for back-compat but is no longer the in-memory signal. Read paths construct `md.has_georef = False`; the conversion function emits the legacy key for round-trip compatibility.
- Tightened nodata semantics (item 5 of the parent attrs-contract review): `md.nodata` and `md.masked_nodata` are the canonical pair. The alias attrs (`nodatavals`, `_FillValue`) are resolved inside `attrs_to_metadata` and never re-surface on a write.

Landing the dataclass first turns the sibling issues into short follow-ups (one field rename or one builder method each).

## Risk

Internal refactor:

- Public attrs surface does not change. The attrs contract documented in `_attrs.py:18-122` continues to hold key-for-key.
- The round-trip tests in `xrspatial/geotiff/tests/test_round_trip_invariants.py` and the parity tests in `test_round_trip_parity_rasterio_zarr_1961.py` enforce the contract on every read/write path; they should pass unmodified at each step of the migration.
- The four `masked_nodata` computations stay at their current call sites; the dataclass only changes the assignment target. Issue #2092's tests in `test_masked_nodata_attr_2092.py` and `test_nodata_semantics_split_1988.py` continue to gate the masking semantics.
- The two VRT inline attrs blocks (`_backends/vrt.py:275-300`, `:697-720`) are the only place where behaviour could observably change, because they currently emit a slightly different field set than `_populate_attrs_from_geo_info` (no `extra_tags`, no `gdal_metadata`, no resolution tags). Routing them through `metadata_to_attrs` brings them into line. The alignment is what we want, but it is a behaviour change in the sense that a VRT read will start carrying tags it did not before. Gated behind a follow-up so this PR can land as a pure refactor first.

## Out of scope

- Public renames of `attrs['crs']`, `attrs['crs_wkt']`, `attrs['nodata']`, or `attrs['masked_nodata']`.
- Bumping `_ATTRS_CONTRACT_VERSION` to 3.
- Adding new fields to the attrs contract.

brendancol:
## Summary

Closes #2139.

Adds an internal `GeoTIFFMetadata` frozen dataclass plus two boundary functions (`metadata_to_attrs`, `attrs_to_metadata`) and a `geo_info_to_metadata` builder. The four read paths and three writers all built/parsed the same `attrs` dict by hand; this PR consolidates the read side onto the typed record. Writers stay on the dict for now to keep the diff small and avoid overlap with the open attrs-contract work on the same files (#2133 / #2135 / #2136).

Reader-side migration:
- `_populate_attrs_from_geo_info` becomes a thin wrapper that builds a `GeoTIFFMetadata` and folds `metadata_to_attrs` back into the caller's dict. Field-for-field equivalent to the prior implementation, so the three non-VRT read paths see no change.
- The two VRT inline attrs blocks (eager + chunked) now build a `GeoTIFFMetadata` and run it through `metadata_to_attrs`, removing the duplicated stamp/set/mark sequences at `_backends/vrt.py:275` and `:697`. The VRT path emits the same key set it did before; aligning the VRT field set with the non-VRT readers is the issue's separate follow-up.

Public attrs surface is unchanged.

## Test plan

- [x] 25 new tests in `test_geotiff_metadata_2139.py` covering `geo_info_to_metadata`, `attrs_to_metadata`, round trips, and the `with_nodata` builder.
- [x] Existing tests pass: `test_attrs_contract_*_1984.py`, `test_round_trip_invariants.py`, `test_nodata_semantics_split_1988.py`, `test_masked_nodata_attr_2092.py`, `test_vrt_holes_attr_1734.py`, `test_nodata_attr_aliases_1582.py`.
- [x] Full geotiff sweep: 4066 passed, 25 skipped.

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -182,7 +182,9 @@ _RESOLUTION_UNIT_IDS = {'none': 1, 'inch': 2, 'centimeter': 3}
 # Reverse map of ``_RESOLUTION_UNIT_IDS``: TIFF ResolutionUnit tag ids back to
 # the xrspatial string identifiers. Used by ``geo_info_to_metadata`` so the
 # read side stores the same string label the writer expects on the way out.
-_RESOLUTION_UNIT_NAMES = {1: 'none', 2: 'inch', 3: 'centimeter'}
+# Derived from ``_RESOLUTION_UNIT_IDS`` so the forward and reverse maps cannot
+# drift if a new unit is added.
+_RESOLUTION_UNIT_NAMES = {v: k for k, v in _RESOLUTION_UNIT_IDS.items()}
 
 
 @dataclass(frozen=True)
@@ -316,6 +318,19 @@ def metadata_to_attrs(md: GeoTIFFMetadata) -> dict:
     if md.raster_type == 'point':
         attrs['raster_type'] = 'point'
 
+    # Three states on the (transform, has_georef) pair:
+    #
+    # * transform set + has_georef -> emit ``attrs['transform']``.
+    # * not has_georef             -> emit ``attrs[_NO_GEOREF_KEY] = True``.
+    # * transform None + has_georef -> emit neither.
+    #
+    # The third state is the eager VRT path's contract: the VRT builder
+    # constructs the record with ``has_georef=True`` but leaves
+    # ``transform=None`` because the inline VRT code stamps the
+    # rasterio-ordered transform tuple onto the dict a few lines later
+    # (after the GPU transfer / dtype cast / nodata mask steps). Removing
+    # this branch would re-introduce a duplicate transform write or, worse,
+    # would emit ``_NO_GEOREF_KEY`` on a georef'd VRT array. See #2139.
     if md.transform is not None and md.has_georef:
         attrs['transform'] = md.transform
     elif not md.has_georef:
@@ -359,6 +374,19 @@ def attrs_to_metadata(attrs) -> GeoTIFFMetadata:
     Honours the alias resolution :func:`_resolve_nodata_attr` already
     implements. ``attrs`` may be a plain dict, an
     ``xarray.core.utils.Frozen``, or ``None``.
+
+    Boundary contract -- this function parses leniently and lets the
+    writer enforce strict validation against the parsed record:
+
+    * ``attrs['crs']=True`` lands as ``crs_epsg=None`` rather than
+      raising. ``_validate_crs_arg`` (called from the writers) is the
+      validator that should reject bool values; the boundary parser
+      only needs to keep the bad value out of the record so the writer
+      sees ``crs_epsg=None`` and falls through to ``crs_wkt``. See
+      ``test_crs_arg_validation_1971.py``.
+    * ``transform`` is coerced via ``tuple(...)`` with no length or
+      numeric-type check. ``_transform_from_attr`` is the canonical
+      validator and runs inside the writer.
     """
     if attrs is None:
         attrs = {}
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -123,6 +123,8 @@ migration notes.
 from __future__ import annotations
 
 import warnings
+from dataclasses import dataclass, replace
+from typing import Any
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: geotiff: add internal GeoTIFFMetadata dataclass and migrate readers (#2139)

### Blockers

None.

### Suggestions

- [ ] `attrs_to_metadata` silently drops `attrs['crs']=True` to `crs_epsg=None` (`_attrs.py:391-398`). The writers today raise via `_validate_crs_arg` on a bool. Nothing in this PR calls `attrs_to_metadata` from a writer, so the difference is invisible right now, but `test_crs_arg_validation_1971.py` expects the writer to reject. When the writer-side migration lands the validator needs to either move onto `attrs_to_metadata` or stay in the writer with the parsed record still flagged enough to be rejected. A one-line comment recording the boundary contract ("parse leniently here; validate at the writer") would save the next person five minutes.

- [ ] `attrs_to_metadata` coerces `transform` via `tuple(transform)` (`_attrs.py:404`) without checking length or numeric type. `_transform_from_attr` does that check today. Same shape of problem as the CRS gate: harmless until a writer calls `attrs_to_metadata`. A note in the docstring about the deferred check is enough.

### Nits

- [ ] `metadata_to_attrs` (`_attrs.py:319-322`) has a three-state branch on `(transform, has_georef)`:
  - `transform is not None and has_georef` -> emit `transform`
  - `not has_georef` -> emit `_NO_GEOREF_KEY`
  - `transform is None and has_georef` -> emit nothing
  The third state is what the VRT eager path relies on (record has no transform yet; the existing block stamps it later). It's correct, but at first read the third case looks like a missed branch rather than the intended path. A one-line comment ("transform-less but georef'd record: caller stamps transform separately, eager VRT path") would make the intent obvious.

- [ ] The VRT eager builder uses `vrt.crs_wkt or None` (`_backends/vrt.py:286`) to convert an empty string to None. The original `if vrt.crs_wkt:` gate did the same thing. Reading the dataclass site without the original nearby, the empty-string normalisation looks accidental. Inline comment or a tiny `_nullify_empty` helper would help.

- [ ] `test_geotiff_metadata_2139.py:288-294` asserts every input key is present in the round-tripped output but doesn't assert the symmetric direction. Full equality would fail for the no-georef case because `metadata_to_attrs` always stamps `_xrspatial_geotiff_contract`, but a sibling test asserting "no unexpected keys other than the contract version" would catch future field additions slipping in unannounced.

- [ ] `_RESOLUTION_UNIT_NAMES` (`_attrs.py:185`) duplicates the inline `{1: 'none', 2: 'inch', 3: 'centimeter'}` literal that used to live in `_populate_attrs_from_geo_info`. Worth deriving it from `_RESOLUTION_UNIT_IDS` (`{v: k for k, v in _RESOLUTION_UNIT_IDS.items()}`) so the two maps can't drift.

### What looks good

- Reader-side migration matches the issue's PR-1 plan: dataclass + two boundary functions + thin-wrapper `_populate_attrs_from_geo_info` + VRT inline-block replacement. Writers stay on the dict, which both follows the issue's per-step PR guidance and avoids the open #2133 / #2135 / #2136 work on the same files.
- The two VRT inline blocks (`_backends/vrt.py:275`, `:697`) now build a `GeoTIFFMetadata` and run through `metadata_to_attrs`, removing the duplicated contract-version stamp and the duplicated CRS / raster_type / no-georef setters.
- `with_nodata` returns a new frozen instance and short-circuits on `nodata is None`, mirroring `_set_nodata_attrs`'s no-op rule so absence still signals "no declared sentinel".
- 25 new tests cover the builder, the parser, alias resolution, the no-georef path, the bool-EPSG guard, and round trips across five representative attrs dicts.
- Full geotiff test sweep: 4268 passed, 25 skipped. No backend-parity regression.

### Checklist

- [x] Algorithm matches reference/paper -- N/A, pure refactor.
- [x] All implemented backends produce consistent results -- 4268 geotiff tests pass.
- [x] NaN handling is correct -- unchanged; the `_set_nodata_attrs` call sites carry the same `masked=` computation as before.
- [x] Edge cases are covered by tests -- no-georef, point raster, user-defined WKT, VRT holes, bool-EPSG.
- [x] Dask chunk boundaries handled correctly -- N/A, attrs construction is per-DataArray, not per-chunk.
- [x] No premature materialization or unnecessary copies -- the dataclass is frozen and built once per read.
- [x] Benchmark exists or is not needed -- not needed; refactor is hot-path neutral.
- [x] README feature matrix updated -- N/A, no new public function.
- [x] Docstrings present and accurate -- all new functions carry docstrings tying back to issue #2139.

brendancol (commented):
## Review follow-up

All review items addressed in 5161385:

**Suggestions (both fixed):**

- `attrs_to_metadata` now documents the lenient-parse / writer-validates boundary contract in its docstring, calling out the dropped-bool-EPSG and the unchecked-tuple-transform cases explicitly.

**Nits (all fixed):**

- Three-state branch in `metadata_to_attrs` has a block comment explaining the `transform=None, has_georef=True` case is the eager VRT path's contract, not a missed branch.
- VRT eager builder comments why `vrt.crs_wkt or None` is the right normalisation (empty-string SRS).
- `_RESOLUTION_UNIT_NAMES` derives from `_RESOLUTION_UNIT_IDS` via dict comprehension so the two maps can't drift.
- Added `test_round_trip_emits_no_unexpected_keys` to lock the marshalling step. A future field added to `metadata_to_attrs` without an attrs-contract update will now fail this test rather than appearing silently on every read.

Final test count: 4071 geotiff tests pass (5 new; was 4066).

PR #2145, opened 2026-05-19
geotiff: add canonical attrs['georef_status'] (#2136)

### LINKED ISSUE #2136: geotiff: canonical georef_status attribute to disambiguate CRS vs transform presence

brendancol:
## Summary

One canonical `attrs['georef_status']` so downstream code can branch on a single value instead of reconciling `crs`, `crs_wkt`, `transform`, and `_xrspatial_no_georef` by hand. Five values cover what the reader can actually see:

- `full` -- CRS resolved plus axis-aligned transform.
- `transform_only` -- transform present, no CRS.
- `crs_only` -- CRS present, no transform tags.
- `none` -- neither.
- `rotated_dropped` -- rotated `ModelTransformationTag` dropped under `allow_rotated=True`. Previously indistinguishable from `none` via public attrs.

Changes:

- Two helpers in `_attrs.py`: `_compute_georef_status` (takes a `GeoInfo`) and `_compute_georef_status_from_parts` (for the VRT branches that don't build a `GeoInfo`). All read sites -- eager, dask, GPU, VRT-eager, VRT-chunked -- funnel through one decision rule.
- `_ATTRS_CONTRACT_VERSION` goes 2 to 3. The contract docstring at the top of `_attrs.py` documents the new key.
- Additive only. `crs` / `crs_wkt` / `transform` / `_xrspatial_no_georef` keep their pre-v3 semantics so existing consumers don't change.

## Backend coverage

- numpy eager: yes (via `_populate_attrs_from_geo_info`)
- cupy / GPU: yes (same helper)
- dask + numpy: yes (same helper)
- dask + cupy: yes (same helper)
- VRT eager and VRT chunked: yes (inline via `_compute_georef_status_from_parts`)

## Test plan

- [x] `test_georef_status_2136.py`: per-state reads across numpy, dask, GPU (skipped without CUDA), and both VRT branches, plus a parametrised round-trip test for the four writable states. `rotated_dropped` stays read-only because `to_geotiff` doesn't emit rotated `ModelTransformationTag`; the issue calls this out under Out of Scope.
- [x] `test_attrs_contract_canonical_1984.py`: `georef_status` added to the canonical key list, pinned to `full` for the canonical fixture.
- [x] `test_attrs_contract_version_1984.py` and `test_attrs_contract_passthrough_1984.py`: switched to tracking `_ATTRS_CONTRACT_VERSION` rather than the literal `2`, so the next bump only touches the constant.
- [x] Full geotiff suite locally: 4276 passed, 25 skipped.

Closes #2136.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -74,6 +74,12 @@ from ._reader import UnsafeURLError
 # and internal callers that genuinely need it can import directly from
 # ``xrspatial.geotiff._reader``. See issue #1708.
 from ._attrs import (
+    GEOREF_STATUS_CRS_ONLY,
+    GEOREF_STATUS_FULL,
+    GEOREF_STATUS_NONE,
+    GEOREF_STATUS_ROTATED_DROPPED,
+    GEOREF_STATUS_TRANSFORM_ONLY,
+    GEOREF_STATUS_VALUES,
     _LEVEL_RANGES,
     _VALID_COMPRESSIONS,
     _extent_to_window,
@@ -124,6 +130,12 @@ __all__ = [
     'ConflictingNodataError',
     'GeoTIFFAmbiguousMetadataError',
     'GeoTIFFFallbackWarning',
+    'GEOREF_STATUS_CRS_ONLY',
+    'GEOREF_STATUS_FULL',
+    'GEOREF_STATUS_NONE',
+    'GEOREF_STATUS_ROTATED_DROPPED',
+    'GEOREF_STATUS_TRANSFORM_ONLY',
+    'GEOREF_STATUS_VALUES',
     'InvalidCRSCodeError',
     'MixedBandMetadataError',
     'NonUniformCoordsError',
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -198,7 +198,10 @@ GEOREF_STATUS_CRS_ONLY = 'crs_only'
 GEOREF_STATUS_NONE = 'none'
 GEOREF_STATUS_ROTATED_DROPPED = 'rotated_dropped'
 
-_GEOREF_STATUS_VALUES = frozenset({
+# Public frozenset of every valid ``georef_status`` value. Exposed so
+# downstream code can validate user-set values without hard-coding the
+# five-string list (e.g. ``status in GEOREF_STATUS_VALUES``).
+GEOREF_STATUS_VALUES = frozenset({
     GEOREF_STATUS_FULL,
     GEOREF_STATUS_TRANSFORM_ONLY,
     GEOREF_STATUS_CRS_ONLY,
@@ -388,11 +391,14 @@ def _compute_georef_status(geo_info) -> str:
     reader change cannot accidentally re-route a real "no transform"
     file into the rotated bucket.
 
-    The four backends (eager, dask, GPU) call this through
-    :func:`_populate_attrs_from_geo_info`; the VRT inline paths import
-    this helper directly because they build their attrs dicts inline
-    (see ``_backends/vrt.py``). Keep all five callers in lockstep by
-    routing every decision through this one function.
+    The eager numpy, dask, and three GPU read sites (chunked / eager /
+    tile in ``_backends/gpu.py``) all call this through
+    :func:`_populate_attrs_from_geo_info`. The two VRT inline branches
+    (eager + chunked in ``_backends/vrt.py``) call
+    :func:`_compute_georef_status_from_parts` directly because they
+    build their attrs dict from a different dataclass and would have to
+    synthesise a fake ``GeoInfo`` to reuse this helper. Keep all the
+    call sites in lockstep through one of the two helpers.
     """
     transform = getattr(geo_info, 'transform', None)
     rotated_affine = (
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -48,6 +48,13 @@ Canonical (xrspatial owns these; round-trip stable):
   "no declared sentinel" signal. See ``_set_nodata_attrs``.
 - ``raster_type``: ``'area'`` (implicit / RasterPixelIsArea) or ``'point'``
   (explicit / RasterPixelIsPoint).
+- ``georef_status``: one of ``'full'``, ``'transform_only'``, ``'crs_only'``,
+  ``'none'``, ``'rotated_dropped'``. Single attr that encodes the five
+  distinct states the reader can land in when CRS / transform tags are
+  combined. See :func:`_compute_georef_status` for the decision table and
+  issue #2136 for the rationale. The attr is additive: ``crs`` / ``crs_wkt``
+  / ``transform`` / ``_xrspatial_no_georef`` remain present with unchanged
+  semantics so existing consumers keep working.
 - ``extra_tags``: list of ``(tag_id, type_id, count, value)`` tuples for
   TIFF tags outside the structured set. Omitted when no out-of-band
   tags are present.
@@ -170,7 +177,34 @@ _TIFF_SHORT = 3
 # matplotlib-colormap attrs that v1 still emitted under a
 # ``DeprecationWarning``. Downstream code that read those keys via
 # ``attrs[key]`` now sees ``KeyError`` rather than the deprecated value.
-_ATTRS_CONTRACT_VERSION = 2
+#
+# Version 3 (issue #2136) adds ``attrs['georef_status']`` to the canonical
+# tier. Existing keys (``crs``, ``crs_wkt``, ``transform``, the
+# ``_xrspatial_no_georef`` marker) keep their pre-v3 shape so downstream
+# code that branches on them still works; the new attr is additive and
+# disambiguates ``crs_only`` from ``none`` and ``rotated_dropped`` from
+# the truly-no-transform case.
+_ATTRS_CONTRACT_VERSION = 3
+
+
+# Canonical ``attrs['georef_status']`` values (issue #2136). One attr
+# encodes the five distinct states the reader can land in when CRS and
+# transform tags are combined; downstream code can branch on this rather
+# than reconstructing the state from the union of ``crs``, ``crs_wkt``,
+# ``transform``, and ``_xrspatial_no_georef``.
+GEOREF_STATUS_FULL = 'full'
+GEOREF_STATUS_TRANSFORM_ONLY = 'transform_only'
+GEOREF_STATUS_CRS_ONLY = 'crs_only'
+GEOREF_STATUS_NONE = 'none'
+GEOREF_STATUS_ROTATED_DROPPED = 'rotated_dropped'
+
+_GEOREF_STATUS_VALUES = frozenset({
+    GEOREF_STATUS_FULL,
+    GEOREF_STATUS_TRANSFORM_ONLY,
+    GEOREF_STATUS_CRS_ONLY,
+    GEOREF_STATUS_NONE,
+    GEOREF_STATUS_ROTATED_DROPPED,
+})
 
 
 # String identifiers (used in xrspatial attrs) -> TIFF ResolutionUnit tag ids.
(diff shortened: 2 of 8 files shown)


brendancol (commented):
## PR Review: canonical attrs['georef_status'] (#2136)

No blockers. A few small things worth addressing in this PR rather than a follow-up.

### Suggestions

- [ ] **xrspatial/geotiff/tests/test_georef_status_2136.py:50** -- `_ROTATED_M` is imported but never referenced. Drop it from the import (or actually use it to pin the matrix the test exercises).
- [ ] **GEOREF_STATUS_\* constants in `_attrs.py`** -- The issue frames `georef_status` as a downstream-consumer signal, but the five constants live in private `_attrs`. If you want callers to branch on them rather than the string literals, re-export from `xrspatial.geotiff`. Otherwise document that comparing against the literal strings is the expected API.

### Nits

- [ ] **xrspatial/geotiff/_attrs.py:391** -- Docstring says "The four backends (eager, dask, GPU)" but there are three distinct GPU call sites in `gpu.py` (chunked, eager, tile) on top of eager numpy + dask. Either drop the count or list the five paths.
- [ ] **`_GEOREF_STATUS_VALUES` frozenset** -- Declared but unused. If it stays, expose it so downstream code can validate; otherwise drop it.
- [ ] **`_backends/vrt.py:286, 717`** -- `bool(vrt.crs_wkt)` is currently safe because `_text` returns `None` (never `""`) for missing/empty SRS, but `_compute_georef_status` uses `crs_wkt is not None`. Align the two checks so a future change to `_text` cannot drift the decision.

### What looks good

- One decision function plus a booleans-only twin for the VRT branches that don't have a `GeoInfo`. Clean separation.
- Rotated-dropped detection keys off `transform.rotated_affine`, not `has_georef`, so a future change to how `has_georef` is set cannot leak a true-no-transform file into the rotated bucket.
- Pre-branch stamping means changes to which optional attrs get emitted can't shift the status value.
- Tests cover the unit decision, the five reader states across numpy / dask / GPU / VRT-eager / VRT-chunked, and round-trip stability for the four writable states. The `rotated_dropped` write-side asymmetry is documented inline.
- Contract version assertions now track `_ATTRS_CONTRACT_VERSION` instead of literals; the next bump only touches the constant.

### Checklist

- [x] Decision table matches the issue
- [x] All read paths funnel through one of the two helpers
- [x] Edge cases: empty `crs_wkt` handled (via `_text` returning None)
- [x] No materialization concerns (attrs work only)
- [x] Docstrings accurate modulo the count typo

brendancol (commented):
## Follow-up: review findings addressed

All suggestions and nits from the previous review pass are addressed in 89283ce.

### Fixed

- **Public constants exposed.** `GEOREF_STATUS_FULL`, `GEOREF_STATUS_TRANSFORM_ONLY`, `GEOREF_STATUS_CRS_ONLY`, `GEOREF_STATUS_NONE`, `GEOREF_STATUS_ROTATED_DROPPED`, and the new `GEOREF_STATUS_VALUES` frozenset are now re-exported from `xrspatial.geotiff` and added to `__all__`. Downstream code can `from xrspatial.geotiff import GEOREF_STATUS_FULL` instead of reaching into the private `_attrs` module. The `test_features.py` public-API frozen list grew by six names to match.
- **VRT `bool(vrt.crs_wkt)` -> `vrt.crs_wkt is not None`** in both VRT branches. Matches the `_compute_georef_status` GeoInfo helper exactly; the comment notes why the alignment matters even though `_text` returns None today.
- **Unused `_ROTATED_M` import** dropped from the test module.
- **`_compute_georef_status` docstring** rewritten to enumerate the actual call sites (eager numpy, dask, three GPU paths, two VRT helpers) instead of the stale "four backends" count.
- **`_GEOREF_STATUS_VALUES` renamed to `GEOREF_STATUS_VALUES`** since it is now part of the public surface, with a docstring comment.

### Added

- `test_public_constants_reexported`: pins the public re-exports so a future refactor cannot quietly drop one of the six names.

### Verified

Full geotiff test suite: 4277 passed, 25 skipped.

PR #2146, opened 2026-05-19
geotiff: split overloaded masked_nodata into lifecycle signals

### LINKED ISSUE #2135: geotiff: split overloaded masked_nodata into separate nodata lifecycle signals

brendancol:
Closes #2135.

## Summary

Splits the overloaded ``attrs['masked_nodata']`` boolean into separate
lifecycle signals so consumers can answer four distinct questions
without one bit that lies on the cast-after-mask path.

The existing ``nodata`` / ``masked_nodata`` keys keep their post-#2092
semantics. Two additive attrs land alongside them:

- ``nodata_pixels_present`` (bool): True iff the read window contained
  any pixel matching the declared sentinel before masking. Lets QA and
  writer code answer "any nodata in this tile" without rescanning. The
  eager numpy, GPU, and VRT eager paths compute it; the dask backends
  leave it unset because a strict per-chunk reduction would force eager
  ``.compute()``.
- ``nodata_dtype_cast`` (str): target dtype name (e.g. ``"float64"``)
  when the caller passed an explicit ``dtype=`` kwarg. Distinguishes
  float-because-masked from float-because-promoted, which the existing
  ``masked_nodata`` lookup alone cannot do.

## Backend coverage

- numpy eager: scan integrated into the existing mask block; also
  scans in the ``mask_nodata=False`` branch so the attr stays useful.
- dask numpy: ``pixels_present`` deliberately unset (lazy contract);
  ``dtype_cast`` recorded.
- GPU eager (stripped and tiled): new
  ``_apply_nodata_mask_gpu_with_presence`` helper returns the bool
  alongside the masked buffer.
- dask GPU: ``pixels_present`` unset; ``dtype_cast`` recorded.
- VRT eager: ``_apply_integer_sentinel_mask_with_presence`` plus a
  float-path NaN-presence proxy plus a no-mask-scan helper for the
  ``mask_nodata=False`` branch.
- VRT chunked: ``pixels_present`` unset; ``dtype_cast`` recorded.

## Overlap with sibling PRs

This change brushes ``_attrs.py`` and the four read backends. It does
not pre-empt the scope of #2133, #2136, or #2139; each of those owns a
different attrs lifecycle problem on the same module. The only edit to
``_set_nodata_attrs`` here adds two kwargs with ``None`` defaults so
the existing call shape stays valid.

## Test plan

- [x] ``xrspatial/geotiff/tests/test_nodata_lifecycle_attrs_2135.py``:
  18 tests across eager / dask / VRT (+ skip-marked GPU).
- [x] Existing ``test_masked_nodata_attr_2092.py`` (12) and
  ``test_nodata_semantics_split_1988.py`` (39) still green.
- [x] Full ``xrspatial/geotiff/tests/`` suite (excluding GPU-only
  paths that need cupy + CUDA): 4200 passed.

### DIFF
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -75,6 +75,23 @@ write.
        because the caller passed ``mask_nodata=False`` together with
        ``dtype=float...``. Only set when ``nodata`` is set; absence
        means no declared sentinel. See issue #2092.
+   * - ``nodata_pixels_present``
+     - bool
+     - Paired with ``nodata``. ``True`` iff the read window contained
+       at least one pixel matching the declared sentinel before
+       masking. Lets QA and writer code answer "any nodata in this
+       tile" without rescanning the buffer. Only emitted by the
+       eager-numpy, GPU, and VRT paths; the dask path leaves the attr
+       unset because a strict per-chunk reduction would force eager
+       ``.compute()``. See issue #2135.
+   * - ``nodata_dtype_cast``
+     - str
+     - Paired with ``nodata``. Set to the resolved target dtype name
+       (e.g. ``"float64"``) when the caller passed an explicit
+       ``dtype=`` kwarg, otherwise absent. Distinguishes
+       float-because-masked from float-because-promoted, which a
+       ``masked_nodata`` lookup alone cannot disambiguate. See issue
+       #2135.
    * - ``raster_type``
      - str
      - ``'point'`` when the file declares ``RasterPixelIsPoint``;
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -640,6 +640,12 @@ def open_geotiff(source: str | BinaryIO, *,
     # write is safe; an extra ``arr.copy()`` would just double peak
     # memory for a multi-MB raster.
     nodata = geo_info.nodata
+    # Track whether any sentinel pixel was present in the read window.
+    # Only meaningful when a declared sentinel exists; reported via
+    # ``attrs['nodata_pixels_present']`` (issue #2135) so consumers can
+    # answer "any nodata in this tile" without rescanning. ``None``
+    # keeps the attr out when ``mask_nodata=False`` and no scan happened.
+    nodata_pixels_present: bool | None = None
     if nodata is not None and mask_nodata:
         # When the reader applied MinIsWhite, the sentinel-equality mask
         # must compare against the inverted sentinel value (issue #1809).
@@ -651,7 +657,20 @@ def open_geotiff(source: str | BinaryIO, *,
         nodata_sentinel = getattr(geo_info, '_mask_nodata', nodata)
         if arr.dtype.kind == 'f':
             if nodata_sentinel is not None and not np.isnan(nodata_sentinel):
-                arr[arr == arr.dtype.type(nodata_sentinel)] = np.nan
+                mask_f = arr == arr.dtype.type(nodata_sentinel)
+                nodata_pixels_present = bool(mask_f.any())
+                if nodata_pixels_present:
+                    arr[mask_f] = np.nan
+            else:
+                # NaN-only sentinel on a float buffer: ``mask_nodata`` is
+                # a no-op, but downstream may want to know if any NaN
+                # pixels already exist in the source so the attr stays
+                # informative.
+                nodata_pixels_present = (
+                    bool(np.isnan(arr).any())
+                    if nodata_sentinel is not None
+                    else False
+                )
         elif arr.dtype.kind in ('u', 'i'):
             # Integer arrays: convert to float to represent NaN.
             # An out-of-range sentinel (e.g. uint16 file with
@@ -677,14 +696,55 @@ def open_geotiff(source: str | BinaryIO, *,
                 info = np.iinfo(arr.dtype)
                 if info.min <= nodata_int <= info.max:
                     mask = arr == arr.dtype.type(nodata_int)
-                    if mask.any():
+                    nodata_pixels_present = bool(mask.any())
+                    if nodata_pixels_present:
                         arr = arr.astype(np.float64)
                         arr[mask] = np.nan
+                else:
+                    # Sentinel cannot match any pixel in this dtype's
+                    # range. No mask was run; no pixels were present.
+                    nodata_pixels_present = False
+            else:
+                nodata_pixels_present = False
+    elif nodata is not None and not mask_nodata:
+        # ``mask_nodata=False``: the masking branch above is skipped,
+        # but ``attrs['nodata_pixels_present']`` should still surface so
+        # callers know whether literal sentinel pixels survive in the
+        # buffer (issue #2135). Mirror the same dtype / range / integer
+        # gates as the masking branch so an out-of-range or non-integer
+        # sentinel cleanly resolves to ``False`` rather than crashing in
+        # the equality check.
+        nodata_sentinel = getattr(geo_info, '_mask_nodata', nodata)
+        if nodata_sentinel is not None:
+            if arr.dtype.kind == 'f':
+                if np.isnan(nodata_sentinel):
+                    nodata_pixels_present = bool(np.isnan(arr).any())
+                else:
+                    nodata_pixels_present = bool(
+                        (arr == arr.dtype.type(nodata_sentinel)).any()
+                    )
+            elif arr.dtype.kind in ('u', 'i'):
+                if (np.isfinite(nodata_sentinel)
+                        and float(nodata_sentinel).is_integer()):
+                    nodata_int = int(nodata_sentinel)
+                    info = np.iinfo(arr.dtype)
+                    if info.min <= nodata_int <= info.max:
+                        nodata_pixels_present = bool(
+                            (arr == arr.dtype.type(nodata_int)).any()
+                        )
+                    else:
+                        nodata_pixels_present = False
+                else:
+                    nodata_pixels_present = False
 
+    dtype_cast_attr: str | None = None
     if dtype is not None:
         target = np.dtype(dtype)
         _validate_dtype_cast(arr.dtype, target)
         arr = arr.astype(target)
+        # Record the post-mask cast so downstream can tell
+        # float-because-masked from float-because-promoted (issue #2135).
+        dtype_cast_attr = target.name
 
     # ``attrs['masked_nodata']`` reflects whether the function
     # actually replaced sentinel pixels with NaN (#2092). The pre-fix
@@ -697,6 +757,8 @@ def open_geotiff(source: str | BinaryIO, *,
     _set_nodata_attrs(
         attrs, nodata,
         masked=(mask_nodata and arr.dtype.kind == 'f'),
+        pixels_present=nodata_pixels_present,
+        dtype_cast=dtype_cast_attr,
     )
 
     if arr.ndim == 3:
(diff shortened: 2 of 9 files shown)


brendancol (commented):
## PR Review: geotiff: split overloaded masked_nodata into lifecycle signals

### Blockers
None.

### Suggestions
- `xrspatial/geotiff/_backends/vrt.py:364`: when the integer-sentinel helper already returned `pixels_present=True`, the follow-up `np.isnan(arr).any()` scan still runs (since the `not nodata_pixels_present` guard only short-circuits the True case before re-checking). Today the helper sets True on int-with-sentinel hits, so the float-NaN proxy is gated correctly. If a future change starts using this code path with a float-source-already-NaN-masked-by-internal-reader buffer that also went through the helper, we will scan twice. Worth a comment pinning the assumption that exactly one branch sets the bool.
- `xrspatial/geotiff/_vrt.py:866` (`_apply_integer_sentinel_mask_with_presence`): duplicates the body of `_apply_integer_sentinel_mask`. Tolerable for clarity right now; if a third variant ever lands, factor the shared loop into a single helper that takes a "report presence" flag.

### Nits
- `xrspatial/geotiff/_backends/_gpu_helpers.py:45`: `_apply_nodata_mask_gpu_with_presence` repeats most of `_apply_nodata_mask_gpu`. The presence-reporting variant adds one `mask.any().item()` plus a tuple return; the rest is identical. A single internal helper plus thin wrappers would cut the duplication.
- `xrspatial/geotiff/__init__.py:711`: `nodata_pixels_present` initialised to `None`, then potentially set to `False` in the no-mask-scan branch. The `mask_nodata=True` branch never explicitly sets `False` when the dtype is neither float nor int (e.g. complex source). Leaving it `None` for that case is fine, but worth a one-line comment so a future maintainer does not "fix" the asymmetry.

### What looks good
- Additive contract: every existing caller of `_set_nodata_attrs` was updated by-name in this PR; the two new kwargs default to `None` so out-of-tree consumers stay valid.
- Dask backends honour the issue's recommendation to leave `nodata_pixels_present` unset rather than forcing eager compute.
- VRT path threads the presence bit through both the integer-sentinel helper (explicit return) and the inline float-NaN-masking case (post-mask proxy scan).
- `nodata_dtype_cast` only emitted when the caller actually passed a `dtype=` kwarg, so the absence-means-no-cast contract holds.
- `attrs_contract.rst` and the `_attrs.py` module docstring both updated; the test matrix from the issue is covered row-by-row.

### Checklist
- [x] No off-by-one in window scans (uses existing mask kernels).
- [x] All implemented backends produce consistent attrs.
- [x] NaN handling routes through the existing helpers; no `== np.nan` slips.
- [x] Edge cases covered: out-of-range sentinel, no-sentinel file, mask off + cast on, sentinel absent in window.
- [x] Dask chunk boundaries unaffected; per-chunk path unchanged.
- [x] No premature materialisation; dask path explicitly skips the reduction.
- [n/a] No benchmark needed for an attrs-only contract change.
- [n/a] README feature matrix unchanged (no new public function).
- [x] Docstrings and the contract doc updated.

brendancol (commented):
## Follow-up review pass

Walked back through the four findings from the first review. Pushed `17c23b9`.

### Dispositions

- **vrt.py:364 (Suggestion)**: fixed. Added an explicit invariant comment so the single-branch contract that lets the float-NaN proxy use a `not present` guard is no longer implicit.
- **__init__.py:711 (Nit)**: fixed. Documented the asymmetry between the float / int gates and the exotic-dtype fall-through so a future maintainer does not "fix" it by forcing a False default.
- **_gpu_helpers.py:45 (Nit)**: deferred. The duplication is real, but `_apply_nodata_mask_gpu` is on the GPU writer's hot path; introducing an opt-in flag would force every call site to pay either the Python-side branch cost or the extra bool computation. Added a docstring note pointing at the consolidation opportunity once a third variant lands.
- **_vrt.py:866 (Suggestion)**: deferred for the same reason. The base helper is shared with the per-task chunked-VRT path; folding the variants now adds branch cost to every chunk. Documented the same conditional consolidation plan.

Re-ran the 70 tests in `test_nodata_lifecycle_attrs_2135.py`, `test_masked_nodata_attr_2092.py`, and `test_nodata_semantics_split_1988.py`: all green.

Issue #1845, opened 2026-05-14
write_geotiff_gpu emits JPEG TIFFs that other readers reject


brendancol:
`xrspatial/geotiff/__init__.py` has two writer entry points that disagree about `compression='jpeg'`:

- `to_geotiff` (around line 1313-1334) rejects `compression='jpeg'` with a `ValueError`. The CPU encoder writes self-contained JFIF streams per tile and skips the JPEGTables tag (347), so the resulting files are unreadable by libtiff, GDAL, and rasterio.
- `write_geotiff_gpu` (around line 3314 onward) accepts the same value and writes the same broken format. The docstring at line 3374-3384 already calls the path "experimental and internal-reader-only until the JPEGTables fix lands".

A user who happens to land on the GPU writer (directly or via `to_geotiff(gpu=True)` once the auto-dispatch path is wider) gets a `.tif` on disk that decodes through xrspatial's own reader and fails everywhere else. There is no warning at write time. The asymmetry between the two writers in the same module is the surprising part.

### Proposed fix

Reject `compression='jpeg'` on `write_geotiff_gpu` by default with the same error text `to_geotiff` raises. Add an opt-in `allow_internal_only_jpeg: bool = False` kwarg for users who want the existing internal-only path. When the flag is set, proceed with the current encode path and emit a `GeoTIFFFallbackWarning` so it cannot be missed.

An env-var gate was considered and rejected: it is harder to discover, harder to audit at the call site, and forces test code to mutate process state.

### Existing tests to update

Tests in `xrspatial/geotiff/tests/test_gpu_writer_compression_modes_2026_05_11.py` that call `write_geotiff_gpu(..., compression='jpeg')` will need to pass the new flag to keep exercising the internal path.

PR #2147, opened 2026-05-19
geotiff: push down byte-affecting validation into array-level writers (#2138)

### LINKED ISSUE #2138: geotiff: harden or privatize lower-level write/read entry points that bypass top-level validation
## Problem

`xrspatial.geotiff` exposes a hardened public surface (`open_geotiff`, `to_geotiff`, `write_geotiff_gpu`, `write_vrt`, `read_geotiff_gpu`, `read_geotiff_dask`, `read_vrt`) but several lower-level functions remain importable and callable directly. These lower-level entry points skip validation that the public wrappers perform, so direct callers can produce subtly different files (or read subtly different DataArrays) without any warning.

The lower-level entry points are not prefixed with `_` and are not marked private in the module docstring, so static analysis treats them as importable API.

## Callable lower-level entry points

Confirmed via `grep -n 'def write\|def _write\|def to_geotiff\|def read\|def open'` across `_writer.py`, `_writers/*.py`, and `_reader.py`:

- `xrspatial.geotiff._writer.write` (`_writer.py:1638`) — array-level CPU writer that `to_geotiff` calls after coercion.
- `xrspatial.geotiff._writer.write_streaming` (`_writer.py:1999`) — dask streaming writer that `to_geotiff` calls on dask-backed DataArrays.
- `xrspatial.geotiff._reader.read_to_array` (`_reader.py:3188`) — array-level reader that `open_geotiff` calls and then post-processes. `__init__.py` imports it as `_read_to_array` to discourage external use, but the public name on `_reader` remains `read_to_array`.

Two helpers in `_writers/eager.py` already use the underscore convention and are correctly private:
- `_write_single_tile` (`_writers/eager.py:763`)
- `_write_vrt_tiled` (`_writers/eager.py:832`)

## Validation gaps between `to_geotiff` and `write` / `write_streaming`

Concrete checks that `to_geotiff` runs (`_writers/eager.py:56`) but `write` / `write_streaming` skip:

1. **Compression name validation against the canonical list.** `to_geotiff` checks `compression.lower() in _VALID_COMPRESSIONS` (`_writers/eager.py:312`) and surfaces the full list on the error. `write` calls `_compression_tag(compression)` (`_writer.py:1744`), which raises but without the canonical list, so a typo like `compression='zstandard'` produces a less actionable error.
2. **JPEG opt-in gate.** `to_geotiff` rejects `compression='jpeg'` unless `allow_internal_only_jpeg=True` (`_writers/eager.py:330`, issue #1845). `write` and `write_streaming` accept JPEG without the gate, so a direct caller can silently produce a file that does not round-trip through libtiff / GDAL / rasterio.
3. **`max_z_error` codec pairing.** `to_geotiff` rejects negative values and rejects non-zero values with non-LERC codecs (`_writers/eager.py:348`). `write` accepts both and passes them to `_write_tiled` / `_write_stripped` where they are silently ignored on non-LERC codecs.
4. **`crs` argument validation.** `to_geotiff` calls `_validate_crs_arg(crs)` (`_writers/eager.py:549`) to reject `bool` and other non-int types. `write` takes a raw `crs_epsg: int | None` and writes whatever it receives — passing `crs_epsg=True` round-trips as EPSG=1.
5. **NaN-to-sentinel rewrite with defensive copy.** `to_geotiff` rewrites NaN pixels to the nodata sentinel via `arr.copy()` to avoid mutating the caller's buffer (`_writers/eager.py:714`). `write` requires the caller to have done this already; a direct caller that passes a NaN-containing float array writes NaN bytes to disk, and the resulting file decodes with NaN pixels even where the sentinel tag is set.
6. **3D dim-order validation and band-first reorder.** `to_geotiff` runs `_validate_3d_writer_dims(data.dims)` (`_writers/eager.py:625`, issue #1812) and moves a leading `band` dim to the trailing axis (`_writers/eager.py:627`, `:690`). `write` assumes the array is already band-last; a direct caller passing `(band, y, x)` writes a file whose `y` axis carries band data.
7. **`float16` / `bool_` auto-promotion.** `to_geotiff` promotes `float16` -> `float32` and `bool_` -> `uint8` (`_writers/eager.py:702`). `write` calls `numpy_to_tiff_dtype` on the raw dtype; a `float16` direct call raises from inside the dtype mapper instead of being handled.
8. **Georeferenced-transform fail-closed.** `to_geotiff` calls `_require_transform_for_georeferenced(data, geo_transform)` (`_writers/eager.py:573`, issue #1945) to refuse silently writing a non-georeferenced TIFF when coords look georeferenced but a transform could not be derived. `write` accepts `geo_transform=None` and writes an identity-transform file.
9. **CRS fallback opt-in.** `to_geotiff` calls `_validate_crs_fallback(wkt_fallback, allow_unparseable_crs)` (`_writers/eager.py:734`, issue #1929) to refuse writing an unparseable CRS string to `GTCitationGeoKey`. `write` writes whatever `crs_wkt` it is handed.

## Validation gaps between `open_geotiff` and `read_to_array`

`read_to_array` returns `(np.ndarray, GeoInfo)`. `open_geotiff` then does work the direct caller has to replicate by hand:

1. **Ambiguous-metadata fail-closed.** `open_geotiff` calls `_validate_read_geo_info` (`__init__.py:626`, issue #1987) before populating attrs. Direct callers of `read_to_array` skip this and may consume a `GeoInfo` for a file that the public surface would reject.
2. **Nodata-to-NaN promotion.** `open_geotiff` masks the nodata sentinel to NaN, promoting integer arrays to `float64` when the sentinel matches a real pixel (`__init__.py:642`). Direct callers receive the raw sentinel bytes and have to write the same logic to avoid the same #1774 / #1809 traps.
3. **`mask_nodata` attr.** `open_geotiff` writes `attrs['masked_nodata']` (`__init__.py:697`, issue #2092) so the writer can decide whether to reverse the NaN promotion. Direct callers building a DataArray themselves can desync this attr from the buffer state.
4. **`attrs['transform']` and `attrs['crs']` population.** `open_geotiff` calls `_populate_attrs_from_geo_info` (`__init__.py:633`). Without it, the writer-side bit-stable round-trip path that depends on `attrs['transform']` does not work.

### Option A: push validation down

Move the missing checks into `write` / `write_streaming` / `read_to_array` so direct callers cannot skip them.

Pros: hardens every entry point in one pass. Direct callers benefit. The public wrappers become thinner.

Cons: some validations need state that only the public layer has (e.g. `data.attrs` for `_should_restore_nan_sentinel`, `data.dims` for `_validate_3d_writer_dims`). Pushing those down forces re-shaping the array-level API to accept that state explicitly. Duplicate work when the public wrapper already validated.

### Option B: formalize private-only

Rename `write` -> `_write`, `write_streaming` -> `_write_streaming`, `read_to_array` -> `_read_to_array` on the source module. Update the module docstring on `_writer.py` and `_reader.py` to spell out the private contract. Optionally add a runtime check that walks the call stack and warns on calls from outside `xrspatial.geotiff`.

Pros: cheap, no API reshape, makes the intent explicit to static analysers and code reviewers.

Cons: does not actually fix the footgun for callers who ignore the warning. Breaks any existing external imports (release-note item).

### Recommended: hybrid

1. **Push down the checks whose absence produces subtly different files**, because these are the silent-divergence cases:
   - Compression-name validation against `_VALID_COMPRESSIONS`.
   - JPEG opt-in gate.
   - `max_z_error` codec pairing.
   - `crs_epsg` / `crs_wkt` type validation (reject `bool`, reject unparseable WKT without the opt-in).
   - NaN-to-sentinel rewrite with defensive copy (gated on `restore_sentinel`, which `write` already accepts).
   - `float16` / `bool_` auto-promotion.
   - Empty-spatial-shape guard (already pushed down in `_writer.py:1740` and `:2058` for issue #2075 — use that as the template).

2. **Rename the rest to underscore-prefix** because they need DataArray state the array-level entry point does not have:
   - `write` -> `_write`
   - `write_streaming` -> `_write_streaming`
   - `read_to_array` -> `_read_to_array` (and update the `__init__.py` import to drop the rename alias)

3. **Update module docstrings** on `_writer.py` and `_reader.py` to say the module is private. The `_writers/` package already does this (`_writers/eager.py:1`).

## Tests

Each lower-level entry point should produce byte-identical output to the public wrapper for matching inputs, or the divergence should be intentional and documented:

- `test_write_vs_to_geotiff_byte_parity`: feed `write(arr, ...)` and `to_geotiff(xr.DataArray(arr), ...)` the same inputs and assert byte-equal output files for every compression in `_VALID_COMPRESSIONS`.
- `test_write_streaming_vs_to_geotiff_byte_parity`: same idea for dask-backed inputs, comparing the streaming path to `to_geotiff` with the same chunk size.
- `test_write_rejects_jpeg_without_opt_in`: regression for gap #2.
- `test_write_rejects_negative_max_z_error`: regression for gap #3.
- `test_write_rejects_bool_crs_epsg`: regression for gap #4.
- `test_write_nan_to_sentinel_does_not_mutate_input`: regression for gap #5.
- `test_write_promotes_float16_and_bool`: regression for gap #7.
- If the renames are accepted: `test_no_public_lowercase_write_reexport` that imports `xrspatial.geotiff` and asserts the renamed names are absent from the namespace.

## Acceptance

- All eight write-side validation gaps and four read-side gaps either fail at the array level or are explicitly out of scope (with a docstring note explaining why).
- Lower-level functions that need DataArray state are renamed to underscore-prefix and the module docstrings reflect the private contract.
- Byte-parity tests pass for `write` vs `to_geotiff` and `write_streaming` vs `to_geotiff`.
- Release notes call out the renames as a breaking change for any external import of `xrspatial.geotiff._writer.write` / `write_streaming` / `xrspatial.geotiff._reader.read_to_array`.

brendancol:
Closes #2138.

## Summary

- Push byte-affecting validation from ``to_geotiff`` down into ``_write`` and ``_write_streaming`` so direct callers cannot bypass it: compression-name list, JPEG opt-in gate (#1845), ``max_z_error`` sign + LERC-only pairing, ``crs_epsg`` bool rejection, ``crs_wkt`` fallback gate (#1929), defensive copy on the NaN-to-sentinel rewrite, and ``float16`` / ``bool_`` auto-promotion.
- Rename the three array-level entry points to underscore-prefixed canonical names (``_write``, ``_write_streaming``, ``_read_to_array``). The non-underscored names are kept as aliases so the ~50 internal call sites that import them still work. Module docstrings on ``_writer.py`` and ``_reader.py`` now spell out the private contract.
- Out of scope (kept in ``to_geotiff``): the four checks that need ``data.dims`` / ``data.attrs`` state -- 3D dim-order validation, band-first reorder, fail-closed georeferenced-transform, and the read-side ``masked_nodata`` handling. The issue's hybrid recommendation calls these out as DataArray-level by construction.

## Backend coverage

CPU writer (numpy / dask-via-``_write_streaming``). The GPU writer already runs its own JPEG + CRS-fallback gates before calling into ``_write``, so its CPU-fallback path now inherits the same set of checks without further changes.

## Test plan

- [x] ``pytest xrspatial/geotiff/tests/test_lowlevel_write_pushdown_2138.py`` -- 26 new tests covering each push-down gap plus byte parity between ``_write`` and ``to_geotiff`` for every lossless codec in ``_VALID_COMPRESSIONS``.
- [x] ``pytest xrspatial/geotiff/tests/`` -- full geotiff suite: 4271 passed, 25 skipped.
- [x] ``pytest xrspatial/tests/test_geotiff_streaming_bigtiff_threshold_1785.py xrspatial/tests/test_rasterize.py`` -- dependent tests outside ``geotiff/``.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -109,7 +109,11 @@ from ._validation import (
     _validate_dtype_cast,
     _validate_tile_size_arg,
 )
-from ._writer import write
+# ``_writer.write`` (alias for ``_writer._write``) is module-private;
+# see ``_writer.py`` docstring and issue #2138. The public eager write
+# surface is :func:`to_geotiff`; do not re-export the array-level
+# entry point here. The dotted path ``xrspatial.geotiff._writer._write``
+# still works for the handful of internal call sites that need it.
 from ._writers.eager import to_geotiff
 # Re-export only; called by xrspatial/geotiff/tests/test_nodata_no_extra_copy_1553.py.
 from ._writers.eager import _write_single_tile  # noqa: F401
--- a/xrspatial/geotiff/_reader.py
+++ b/xrspatial/geotiff/_reader.py
@@ -1,4 +1,20 @@
-"""TIFF/COG reader: tile/strip assembly, windowed reads, HTTP range requests."""
+"""TIFF/COG reader: tile/strip assembly, windowed reads, HTTP range requests.
+
+This module is private to :mod:`xrspatial.geotiff`. The supported public
+read entry points are :func:`xrspatial.geotiff.open_geotiff`,
+:func:`xrspatial.geotiff.read_geotiff_gpu`,
+:func:`xrspatial.geotiff.read_geotiff_dask`, and
+:func:`xrspatial.geotiff.read_vrt`. Direct callers of the helpers
+defined here bypass the DataArray-level work that the public wrappers
+perform (ambiguous-metadata fail-closed, nodata-to-NaN promotion,
+``masked_nodata`` attr, ``transform`` / ``crs`` attrs population) and
+have to replicate those steps by hand. See issue #2138.
+
+For source modules inside :mod:`xrspatial.geotiff`, the canonical
+internal name for the array-level reader is :func:`_read_to_array`.
+The non-underscored :func:`read_to_array` is kept as an alias for
+internal call sites that pre-date the rename.
+"""
 from __future__ import annotations
 
 import math
@@ -3185,13 +3201,13 @@ def _miniswhite_inverted_nodata(nodata, ifd: IFD, dtype: np.dtype):
     return nodata
 
 
-def read_to_array(source, *, window=None, overview_level: int | None = None,
+def _read_to_array(source, *, window=None, overview_level: int | None = None,
                   band: int | None = None,
                   max_pixels: int = MAX_PIXELS_DEFAULT,
                   max_cloud_bytes=_MAX_CLOUD_BYTES_SENTINEL,
                   allow_rotated: bool = False,
                   ) -> tuple[np.ndarray, GeoInfo]:
-    """Read a GeoTIFF/COG to a numpy array.
+    """Read a GeoTIFF/COG to a numpy array (module-private).
 
     Parameters
     ----------
@@ -3428,3 +3444,10 @@ def read_to_array(source, *, window=None, overview_level: int | None = None,
         close_sidecar(sidecar)
 
     return arr, geo_info
+
+
+# Backward-compatible alias for internal call sites that pre-date the
+# rename to :func:`_read_to_array`. New code inside
+# ``xrspatial.geotiff`` should import :func:`_read_to_array` directly.
+# See issue #2138.
+read_to_array = _read_to_array
--- a/xrspatial/geotiff/_writer.py
+++ b/xrspatial/geotiff/_writer.py
@@ -1,4 +1,21 @@
-"""GeoTIFF/COG writer."""
+"""GeoTIFF/COG writer (private).
+
+This module is private to :mod:`xrspatial.geotiff`. The supported public
+write entry points are :func:`xrspatial.geotiff.to_geotiff`,
+:func:`xrspatial.geotiff.write_geotiff_gpu`, and
+:func:`xrspatial.geotiff.write_vrt`. Direct callers of the helpers
+defined here bypass the DataArray-level validation that the public
+wrappers run (``transform`` derivation, ``masked_nodata`` handling,
+``band``-first dim reordering, ...) and must accept the resulting byte
+divergence. See issue #2138.
+
+For source modules inside :mod:`xrspatial.geotiff`, the canonical
+internal names for the array-level write functions are
+:func:`_write` and :func:`_write_streaming`. The non-underscored
+:func:`write` and :func:`write_streaming` are kept as aliases for
+internal call sites that pre-date the rename and may be removed in a
+future release.
+"""
 from __future__ import annotations
 
 import math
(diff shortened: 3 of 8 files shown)


brendancol (commented):
## PR Review: geotiff: push down byte-affecting validation into array-level writers (#2138)

Read the PR head at `issue-2138`. The hybrid approach matches the issue's recommendation: byte-affecting checks pushed down into `_write` / `_write_streaming`, DataArray-state-dependent checks left in `to_geotiff`, and the three array-level entry points renamed to underscore-prefixed canonical names with public-named aliases kept to avoid disturbing ~50 internal call sites.

None.

- [x] `xrspatial/geotiff/_writers/eager.py:776,1063,1080` -- `_write_single_tile` did not forward the new `allow_internal_only_jpeg` / `allow_unparseable_crs` kwargs to its inner `write(...)` call. A direct caller of `to_geotiff(da, 'foo.vrt', compression='jpeg', allow_internal_only_jpeg=True)` would clear the upstream gate then trip the new push-down gate inside `_write`. **Fixed in commit b81d298** by threading the kwargs through `_write_vrt_tiled` -> `_write_single_tile` -> `write`.
- [x] `xrspatial/geotiff/__init__.py:112` -- `from ._writer import write` leaks the array-level entry point onto `xrspatial.geotiff.write` even though `write` is not in `__all__`. Mirrors the `read_to_array` cleanup #1708 already did. **Fixed in commit b81d298** by dropping the unused re-export and adding a namespace-no-leak regression test.

- `xrspatial/geotiff/_writer.py:1711-1714` -- the new `TypeError` branch on non-string `compression` is reachable only by direct callers (the public `to_geotiff` wrapper rejects non-string `compression` upstream with the same type-only check). The downstream `_compression_tag` call would still raise `AttributeError` on the bare string-method access, so the explicit guard is defensible. Worth a one-line comment noting that the branch is unreachable from `to_geotiff` and exists only for direct callers.
- `xrspatial/geotiff/_writer.py:1920-1924` -- the float16 / bool_ promotion guards on `isinstance(data, np.ndarray)` so dask arrays fall through. `_write` is documented as numpy-only (the dask path goes through `_write_streaming`), so the guard is correct, but a one-line comment explaining why the promotion is skipped for dask arrays would help the next reader.

### What looks good

- The hybrid scope matches the issue exactly. Eight write-side validation gaps are inventoried; the six byte-affecting ones are pushed down, the two that need DataArray state (3D dim-order, fail-closed transform) stay in `to_geotiff`. Read-side gaps are correctly identified as DataArray-attrs-dependent and left in `open_geotiff`.
- The `_validate_lowlevel_write_kwargs` helper centralises the gates so `_write` and `_write_streaming` cannot drift. The `entry_point` argument makes the error source unambiguous.
- `restore_sentinel`-gated NaN-to-sentinel rewrite with defensive copy at `_writer.py:1933-1941` matches the eager-path semantics exactly and respects the `#1988` opt-out.
- Byte-parity tests at `tests/test_lowlevel_write_pushdown_2138.py:243-273` are the right shape: every entry in the lossless codec set rather than spot-checking one. JPEG is correctly excluded with an explicit reason in a comment.
- The aliases `write = _write`, `write_streaming = _write_streaming`, `read_to_array = _read_to_array` are object-identity preserving, which `test_aliases_match_underscore_names` pins so a future refactor cannot drift the alias out of sync.
- Test count is right for the scope: 26 new push-down + byte-parity tests, three existing tests updated for the renamed entry-point name in error messages.

### Checklist

- [x] Algorithm matches issue gap inventory
- [x] All implemented backends produce consistent results (numpy via `_write`, dask via `_write_streaming`; byte-parity tests pass on every lossless codec)
- [x] NaN handling is correct (defensive copy preserves caller buffer; sentinel-on-disk verified)
- [x] Edge cases are covered by tests (bool crs_epsg, negative max_z_error, float16, bool_, unknown compression)
- [x] Dask chunk boundaries handled correctly (push-down validation fires before any tile-row math)
- [x] No premature materialization or unnecessary copies (the new `arr.copy()` only runs when there is a NaN to rewrite)
- [x] Benchmark exists or is not needed (validation-only PR; no perf-sensitive path)
- [x] README feature matrix updated (if applicable) -- N/A, no public API change
- [x] Docstrings present and accurate (module docstrings spell out the private contract; `_write` / `_write_streaming` document the new kwargs and the DataArray-state gap)

brendancol (commented):
## Follow-up review pass after addressing review feedback

Second pass against the head of `issue-2138` after the merge from `main` and the review nits.

### Disposition of the original review findings

- **Suggestion: `_write_single_tile` does not forward JPEG / CRS opt-ins** -- fixed in commit b81d298. The kwargs now thread through `to_geotiff` -> `_write_vrt_tiled` -> `_write_single_tile` -> `_write`.
- **Suggestion: `__init__.py` re-exports `write` even though it is not in `__all__`** -- fixed in commit b81d298. Dropped the unused re-export and added a namespace-no-leak test pinning the contract.
- **Nit: comment the `TypeError` branch as unreachable from `to_geotiff`** -- fixed in commit adbfd38.
- **Nit: comment the `isinstance(data, np.ndarray)` gate as intentional** -- fixed in commit adbfd38.

### Re-review notes

- The merge from `origin/main` (commits b0c6ff9 and 24b014c, both `allow_rotated` CRS handling) touched `__init__.py`. Confirmed no interaction with the #2138 changes -- the `_writer.write` re-export drop sat in the writer-imports block, the rotated-CRS work sits in the reader-side block.
- Full geotiff suite: 4285 passed, 25 skipped, 1 xfailed after the merge.
- `mergeStateStatus` is still computing on the GitHub side; CI is running.

No new findings.

PR #2150, opened 2026-05-19
geotiff: tier the public surface, gate experimental codecs (#2137)

### LINKED ISSUE #2137: geotiff: tier the public surface; mark experimental codecs/features opt-in
## Why

The `xrspatial.geotiff` module has grown a lot of surface area. Some of it is the boring core (read a local `.tif`, write a `.tif`). Some of it is genuinely impressive but risky: VRT mosaics, COG, cloud URLs, GPU encode/decode via nvCOMP and nvJPEG, JPEG-in-TIFF, JPEG2000, LERC, GDAL metadata pass-through, external `.tif.ovr` sidecars, rotated `ModelTransformationTag` files.

A new user reading `from xrspatial.geotiff import open_geotiff, to_geotiff` cannot tell from the public surface which combinations are the supported "happy path" and which are research-grade. Every codec sits flat in `_VALID_COMPRESSIONS`. Every kwarg sits flat in the docstring. The one feature we have already gated behind an explicit opt-in is `allow_internal_only_jpeg` on `to_geotiff` (see `xrspatial/geotiff/tests/test_to_geotiff_allow_internal_only_jpeg_parity.py`), and that posture is the right one: the writer accepts JPEG only when the caller asks for it by name and emits `GeoTIFFFallbackWarning` so they know the output will not round-trip through libtiff / GDAL / rasterio.

This issue proposes extending that posture to the rest of the module: tier the features, document the tiers, and put an opt-in or warning in front of anything beyond the stable core. Nothing is removed.

## Current feature inventory

From `xrspatial/geotiff/__init__.py`, `_writers/eager.py`, `_writers/gpu.py`, `_backends/{gpu,dask,vrt}.py`, `_compression.py`, `_reader.py`, `_sidecar.py`:

Read entry points: `open_geotiff`, `read_geotiff_gpu`, `read_geotiff_dask`, `read_vrt`.
Write entry points: `to_geotiff`, `write_geotiff_gpu`, `write_vrt`.

Codec set in `_attrs.py:_VALID_COMPRESSIONS`: `none`, `deflate`, `lzw`, `jpeg`, `packbits`, `zstd`, `lz4`, `jpeg2000` / `j2k`, `lerc`.

Other capabilities: COG (`cog=True`), overview pyramids (`overview_levels=`), external `.tif.ovr` sidecars (`_sidecar.py`), VRT mosaics, fsspec / cloud URIs (`s3://`, `gs://`, `az://`, `abfs://`, `memory://`), HTTP range reads, GPU nvCOMP / nvJPEG / nvJPEG2K, BigTIFF, rotated transforms (`allow_rotated=True`), unparseable CRS (`allow_unparseable_crs=True`), GDAL metadata pass-through and `extra_tags`.

### Tier 1: Stable

The path a new user should be on. Local file in, local file out, common codec, axis-aligned grid.

- `open_geotiff(path)` and `to_geotiff(da, path)` on a local file system path.
- Codecs `none`, `deflate`, `zstd`, `lzw`, `packbits` (lossless integer + float, byte-for-byte round-trip).
- 2D and 3D rasters, axis-aligned transforms, EPSG CRS in `attrs['crs']`.
- `window=`, `band=`, `dtype=`, `nodata=`, `tiled=`, `tile_size=`, `predictor=`.
- `chunks=` for dask out-of-core reads on a local path.

### Tier 2: Supported but advanced

Works, is tested, but the caller should know what they are signing up for.

- VRT mosaics (`read_vrt`, `write_vrt`). Cross-source nodata, missing backing files, partial mosaics, per-band metadata disagreement.
- COG with overviews (`cog=True`, `overview_levels=`, `overview_resampling=`).
- External `.tif.ovr` sidecars.
- BigTIFF (`bigtiff=True`).
- fsspec cloud URIs (`s3://`, `gs://`, `az://`, ...) with the `max_cloud_bytes` budget.
- HTTP / HTTPS range reads on a `.tif` URL.
- Float predictor (`predictor=3`).
- `allow_rotated=True` (read-only; write path drops the rotation silently).
- `allow_unparseable_crs=True`.

### Tier 3: Experimental / opt-in

Works in our tests, no claim about external interop or numerical parity across backends.

- `lerc` codec. Lossy when `max_z_error > 0`; the bound is on per-pixel absolute error, not on downstream analytics.
- `jpeg2000` / `j2k` codec. CPU path uses glymur, GPU path uses nvJPEG2K, the two are not byte-for-byte equal.
- `lz4` codec. Listed in `_VALID_COMPRESSIONS` but rarely seen in the wild; reader support across GDAL versions is uneven.
- GPU read (`read_geotiff_gpu`, `open_geotiff(gpu=True)`) and GPU write (`write_geotiff_gpu`, `to_geotiff(gpu=True)`). Requires cupy + numba CUDA, optionally nvCOMP / nvJPEG / nvJPEG2K; `on_gpu_failure='strict'` vs `'auto'` controls fallback.
- GDAL metadata XML round-trip and `extra_tags` pass-through.

### Tier 4: Internal-only

Already gated behind an explicit opt-in. Used as the template for the rest.

- `allow_internal_only_jpeg=True` on `to_geotiff` / `write_geotiff_gpu`. The encode emits self-contained JFIF tiles and skips the JPEGTables tag, so the output is unreadable by libtiff / GDAL / rasterio and only round-trips through this library's reader. Already emits `GeoTIFFFallbackWarning`. See #1845.

## Opt-in mechanism

Recommend per-feature flags rather than a single `experimental=True` kwarg. A single flag flips too much at once and hides which feature actually mattered when something goes wrong. The JPEG-internal-only flag is the model.

For Tier 3 codecs, the proposal is to reject the codec name in `_VALID_COMPRESSIONS` unless an `allow_experimental_codecs=True` kwarg is set on the writer, and emit `GeoTIFFFallbackWarning` when the opt-in fires (same warning class as the JPEG path). The rejection message names the kwarg, mirroring how the JPEG rejection points at `allow_internal_only_jpeg`.

For Tier 3 GPU paths, the existing `gpu=True` kwarg is already the explicit opt-in. The change is in the docstring: mark the GPU path Experimental, list the optional libraries, and link the parity-matrix issue.

For Tier 2 features, no kwarg gate is added. The docstring gets a short "Advanced" marker block at the top of the relevant parameter, naming the specific failure mode (e.g. cloud reads can run up cost, VRT can return partial mosaics under `missing_sources='warn'`, rotated transforms drop on write).

No env vars. They are hard to discover and hard to test.

## Docstring posture

Each public entry point gets a tier marker block right after the summary line. Each parameter that crosses tier boundaries gets a one-line marker:

- Tier 1: no marker.
- Tier 2: `Advanced: <one-line caveat>` at the start of the parameter doc.
- Tier 3: `Experimental: <one-line caveat>` at the start of the parameter doc.
- Tier 4: `Internal-only: <one-line caveat>` at the start of the parameter doc.

The compression docstring in `_writers/eager.py` already calls out `lerc` as lossy and `jpeg` as internal-only. Extending that pattern to the other codecs is the bulk of the change.

## Deprecation plan

None. This issue is documentation plus opt-in flags only. The wire-level codec set is unchanged. Existing callers of `lerc` / `j2k` / `lz4` will see a `ValueError` the first time they hit the new gate, with a message naming `allow_experimental_codecs=True` so they can fix it in one line. That is the same break shape `allow_internal_only_jpeg` already has, and we accepted it there.

If a future release wants to drop a codec entirely, it goes through the normal `DeprecationWarning` -> remove cycle. Tiering does not skip that.

## Acceptance criteria

1. A module-level `xrspatial.geotiff.SUPPORTED_FEATURES` constant (dict-of-tuples or a small dataclass) that enumerates every feature with its tier. The test suite imports it. Example shape:

   ```python
   SUPPORTED_FEATURES = {
       'codec.zstd': 'stable',
       'codec.deflate': 'stable',
       'codec.lzw': 'stable',
       'codec.packbits': 'stable',
       'codec.none': 'stable',
       'codec.lerc': 'experimental',
       'codec.jpeg2000': 'experimental',
       'codec.j2k': 'experimental',
       'codec.lz4': 'experimental',
       'codec.jpeg': 'internal_only',
       'reader.local_file': 'stable',
       'reader.fsspec': 'advanced',
       'reader.http': 'advanced',
       'reader.vrt': 'advanced',
       'reader.sidecar_ovr': 'advanced',
       'reader.gpu': 'experimental',
       'reader.allow_rotated': 'advanced',
       'reader.allow_unparseable_crs': 'advanced',
       'writer.cog': 'advanced',
       'writer.overviews': 'advanced',
       'writer.bigtiff': 'advanced',
       'writer.gpu': 'experimental',
       'writer.gdal_metadata_xml': 'experimental',
       'writer.extra_tags': 'experimental',
   }
   ```

2. A new `allow_experimental_codecs: bool = False` kwarg on `to_geotiff` and `write_geotiff_gpu`. Setting `compression=` to a Tier 3 codec without the flag raises `ValueError` whose message names the flag. Setting it with the flag emits `GeoTIFFFallbackWarning` once per call.

3. The JPEG opt-in keeps its own flag name (`allow_internal_only_jpeg`). Internal-only is a stricter tier than experimental; the two should not collapse into one switch.

4. Docstrings for `open_geotiff`, `to_geotiff`, `read_geotiff_gpu`, `write_geotiff_gpu`, `read_vrt`, `write_vrt`, and `read_geotiff_dask` carry a tier marker at the top and per-parameter tier markers where they cross tiers.

5. A test in `xrspatial/geotiff/tests/` walks `SUPPORTED_FEATURES`, checks that every Tier 3 entry has a matching `ValueError` rejection on the default writer call, and checks that every Tier 4 entry rejects without its specific opt-in flag.

6. A short table in the geotiff user guide notebook (`examples/user_guide/39_GeoTIFF_IO.ipynb`) generated from `SUPPORTED_FEATURES`, so the documentation cannot drift from the code.

## Tie-in: parity matrix (#2132)

The backend parity test matrix in #2132 needs a definition of which features parity is required for. Tier 1 is the answer. Tier 2 features get parity tests where the feature applies to multiple backends (e.g. cloud URIs are CPU-only, sidecar ovr is CPU-only, COG cuts across CPU and GPU writers). Tier 3 features only carry a single-backend test, since cross-backend numerical parity is not a claim Tier 3 makes. The `SUPPORTED_FEATURES` constant lets the parity matrix iterate the tier set programmatically rather than hard-coding the list.

## Out of scope

- Removing any codec or capability.
- Changing the wire format.
- Renaming any public function.
- Adding new codecs.

brendancol:
Closes #2137.

## Summary

- Adds `xrspatial.geotiff.SUPPORTED_FEATURES`, a module-level mapping from feature name to tier label (`stable` / `advanced` / `experimental` / `internal_only`). Single source of truth for the docstrings, the writer gates, and the user-guide notebook.
- Adds `allow_experimental_codecs=False` to `to_geotiff` and `write_geotiff_gpu`. Tier 3 codecs (`lerc`, `jpeg2000` / `j2k`, `lz4`) raise `ValueError` naming the flag without it; with it, the writer emits `GeoTIFFFallbackWarning` and proceeds. Modelled on the existing `allow_internal_only_jpeg` opt-in for the Tier 4 `jpeg` codec (#1845); the two flags do not collapse into one switch.
- Docstrings on `to_geotiff`, `write_geotiff_gpu`, and `open_geotiff` carry a tier marker block after the summary, plus per-parameter `Advanced:` / `Experimental:` markers on kwargs that cross tier boundaries.
- User-guide notebook `39_GeoTIFF_IO.ipynb` renders `SUPPORTED_FEATURES` as a markdown table so the documentation cannot drift from the code.

## Out of scope

- Nothing is removed. The wire-level codec set is unchanged.
- The wire format, GPU dispatcher, and public function names are untouched.

## Breaking change shape

Existing callers of `to_geotiff(compression='lerc')` (or `'jpeg2000'` / `'j2k'` / `'lz4'`) see a `ValueError` the first time they hit the new gate. The message names `allow_experimental_codecs=True` so the caller can fix the call site in one line. This is the same break shape `allow_internal_only_jpeg` already has, and the issue accepted it there.

In-tree tests that wrote Tier 3 codecs are updated to pass the flag so they exercise the encode path rather than the rejection gate.

## Backend coverage

CPU (`to_geotiff`) and GPU (`write_geotiff_gpu`) both enforce the gate; the auto-dispatch path through `to_geotiff(gpu=True, ...)` forwards the kwarg. Read paths are unchanged (the codec set sits on writers, not readers).

## Test plan

- [x] New test file `test_supported_features_tiers_2137.py` walks `SUPPORTED_FEATURES`, asserts every Tier 3 codec rejects by default and accepts the opt-in with a `GeoTIFFFallbackWarning`, and pins that `allow_experimental_codecs` does not unlock the Tier 4 `jpeg` codec.
- [x] Full geotiff test suite passes locally (4262 passed, 25 skipped).
- [x] `__all__` parity test updated to include the new `SUPPORTED_FEATURES` export.

### DIFF
--- a/examples/user_guide/39_GeoTIFF_IO.ipynb
+++ b/examples/user_guide/39_GeoTIFF_IO.ipynb
@@ -261,7 +261,7 @@
        "}\n",
        "\n",
        ".xr-group-name::before {\n",
-       "  content: \"📁\";\n",
+       "  content: \"\ud83d\udcc1\";\n",
        "  padding-right: 0.3em;\n",
        "}\n",
        "\n",
@@ -324,7 +324,7 @@
        "\n",
        ".xr-section-summary-in + label:before {\n",
        "  display: inline-block;\n",
-       "  content: \"►\";\n",
+       "  content: \"\u25ba\";\n",
        "  font-size: 11px;\n",
        "  width: 15px;\n",
        "  text-align: center;\n",
@@ -335,7 +335,7 @@
        "}\n",
        "\n",
        ".xr-section-summary-in:checked + label:before {\n",
-       "  content: \"▼\";\n",
+       "  content: \"\u25bc\";\n",
        "}\n",
        "\n",
        ".xr-section-summary-in:checked + label > span {\n",
@@ -968,6 +968,49 @@
    "metadata": {},
    "source": [
     "The VRT is a few hundred bytes of XML. `open_geotiff` assembles the tiles when you read it."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Supported features by tier\n",
+    "\n",
+    "`xrspatial.geotiff.SUPPORTED_FEATURES` is the source of truth for which features sit in the stable core, which are advanced but supported, which are experimental, and which require an opt-in flag because they do not round-trip through external readers. The table below is built from that constant so the documentation cannot drift from the code (issue #2137).\n",
+    "\n",
+    "Tiers:\n",
+    "\n",
+    "- **stable** -- the path a new user should be on. Local file in, local file out, lossless codec, axis-aligned grid.\n",
+    "- **advanced** -- works and is tested, but the caller should know the failure mode (cloud cost, partial VRT mosaics, rotated transforms drop on write, BigTIFF promotion, etc.).\n",
+    "- **experimental** -- no claim about cross-backend numerical parity or external interop. Tier 3 codecs (`lerc`, `jpeg2000` / `j2k`, `lz4`) require `allow_experimental_codecs=True` on `to_geotiff` and `write_geotiff_gpu`; the GPU read/write paths use `gpu=True` as their explicit opt-in.\n",
+    "- **internal_only** -- the strictest tier. `compression='jpeg'` writes self-contained JFIF tiles without the TIFF JPEGTables tag, so the output decodes through xrspatial but not libtiff / GDAL / rasterio. Requires the dedicated `allow_internal_only_jpeg=True` flag (issue #1845); `allow_experimental_codecs` does not cover it."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from xrspatial.geotiff import SUPPORTED_FEATURES\n",
+    "\n",
+    "# Render SUPPORTED_FEATURES as a markdown table grouped by tier.\n",
+    "# This cell renders from the live constant so the table tracks any\n",
+    "# future additions to the feature inventory.\n",
+    "from collections import defaultdict\n",
+    "from IPython.display import Markdown\n",
+    "\n",
+    "_TIER_ORDER = ['stable', 'advanced', 'experimental', 'internal_only']\n",
+    "_by_tier = defaultdict(list)\n",
+    "for name, tier in SUPPORTED_FEATURES.items():\n",
+    "    _by_tier[tier].append(name)\n",
+    "\n",
+    "lines = ['| Feature | Tier |', '| --- | --- |']\n",
+    "for tier in _TIER_ORDER:\n",
+    "    for name in sorted(_by_tier[tier]):\n",
+    "        lines.append(f'| `{name}` | {tier} |')\n",
+    "\n",
+    "Markdown('\\n'.join(lines))"
    ]
   },
   {
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -128,6 +128,7 @@ __all__ = [
     'MixedBandMetadataError',
     'NonUniformCoordsError',
     'RotatedTransformError',
+    'SUPPORTED_FEATURES',
     'UnparseableCRSError',
     'UnsafeURLError',
     'open_geotiff',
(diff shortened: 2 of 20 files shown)


brendancol (commented):
## PR Review: geotiff: tier the public surface, gate experimental codecs (#2137)

None.

- [ ] `xrspatial/geotiff/_writers/eager.py:423` -- the experimental-codec warning at the CPU dispatcher fires unconditionally before the GPU dispatch decision. So `to_geotiff(gpu=True, compression='lerc', allow_experimental_codecs=True)` warns here, then warns again in `write_geotiff_gpu` at `_writers/gpu.py:321`. The JPEG opt-in at `eager.py:471-483` avoids the double-warn by gating on `not use_gpu` (the `use_gpu` local is computed at line 470 for exactly this reason). The comment at `eager.py:417-422` says this "mirrors `allow_internal_only_jpeg`" but the JPEG path does the opposite. Either gate the new warning the same way, or fix the comment and the docstring's "once per call" claim to acknowledge the double-warn.

- [ ] Acceptance criterion 4 in #2137 names seven entry points that need a tier marker: `open_geotiff`, `to_geotiff`, `read_geotiff_gpu`, `write_geotiff_gpu`, `read_vrt`, `write_vrt`, and `read_geotiff_dask`. The PR covers `to_geotiff`, `write_geotiff_gpu`, and `open_geotiff`. The other four (`_backends/gpu.py:92`, `_backends/dask.py:34`, `_backends/vrt.py:33`, `_writers/vrt.py:21`) still have no tier marker block at the top of their docstrings.

- [ ] `test_supported_features_tiers_2137.py:216-222` -- the docstring promises "emits `GeoTIFFFallbackWarning` once per call" but the assertion only checks `len(fallback) >= 1`. Pin the count: `assert len(fallback) == 1` would have caught the double-warn the next time someone touched the dispatcher.

- [ ] `_writers/eager.py:407` and `_writers/gpu.py:308` import `_EXPERIMENTAL_CODECS` inside the function body. The constant only depends on `SUPPORTED_FEATURES`, so there is no circular-import risk for hoisting the import to module scope; it would save one lookup per write call.

### What looks good

- `_EXPERIMENTAL_CODECS` is derived from `SUPPORTED_FEATURES` rather than written out twice. The gate cannot drift from the documentation.
- The new test walks `SUPPORTED_FEATURES` instead of hard-coding the codec list. Adding a new Tier 3 entry grows the test matrix automatically.
- The notebook renders the table from the live constant.
- `__all__` carries the new `SUPPORTED_FEATURES` export and the existing `test_all_lists_supported_functions` parity check was updated in the same commit.
- The PR resists the temptation to fold `compression='jpeg'` into the new flag. Internal-only is the stricter tier and keeps its own gate, matching acceptance criterion #3.
- The in-tree tests that wrote Tier 3 codecs were updated to pass the new flag, so the suite stays green rather than silently breaking 41 tests.

brendancol (commented):
## Follow-up: addressed the three Suggestions and the Nit

### Disposition

- **Suggestion 1 (double-warn on GPU dispatch path)** -- **fixed** in `a187138`. The CPU dispatcher's experimental warning is now gated on `not use_gpu`, mirroring the JPEG opt-in at `eager.py:471-483`. The `write_geotiff_gpu` warning still fires on the GPU path. The rejection check at line 411 still runs unconditionally so a default call sees the same `ValueError` either way.

- **Suggestion 2 (missing tier markers on four entry points)** -- **fixed** in `a187138`. Added tier marker blocks to:
  - `read_geotiff_gpu` (`_backends/gpu.py`) -- Experimental
  - `read_geotiff_dask` (`_backends/dask.py`) -- Stable for the Tier 1 codec set, Advanced for the `allow_*` opt-ins
  - `read_vrt` (`_backends/vrt.py`) -- Advanced, with the specific failure modes named
  - `write_vrt` (`_writers/vrt.py`) -- Advanced

  All seven entry points listed in acceptance criterion 4 now carry the marker.

- **Suggestion 3 (loose warning count assertion)** -- **fixed** in `a187138`. The opt-in test now asserts `len(fallback) == 1` and includes a comment explaining what the count would catch. With Suggestion 1 fixed the assertion holds; without it, the GPU dispatch path would fail this test.

- **Nit 4 (lazy import)** -- **fixed** in `a187138`. Moved `SUPPORTED_FEATURES` and `_EXPERIMENTAL_CODECS` from the package `__init__` to `_attrs.py` so the writers can import them at module scope. The package `__init__` re-exports `SUPPORTED_FEATURES`, so the public API at `xrspatial.geotiff.SUPPORTED_FEATURES` is unchanged. This breaks the circular dependency that forced the lazy import.

### Verification

- Full geotiff test suite: 4262 passed, 25 skipped on the latest commit.
- The new `test_experimental_codec_opt_in_emits_warning` parametrize set still passes with the tighter `== 1` assertion.
- No other test files needed updates for the `_attrs.py` move; the public re-export keeps the existing import paths working.

PR #2153, opened 2026-05-19
polygonize: cupy backend honours float tolerance from numba path (#2151)

### LINKED ISSUE #2151: polygonize: cupy backend uses exact equality where numpy uses tolerance
## Summary

The `cupy` polygonize backend groups float pixels into regions using exact equality (`data == v`) while the `numpy` (and therefore dask and dask+cupy) backend uses tolerance-based comparison via `_is_close` (`atol=1e-8`, `rtol=1e-5`). This produces different polygons for the same float input depending on the backend.

## Reproducer

```python
import numpy as np
import xarray as xr
import cupy as cp

data = np.array([
    [1.0, 1.000001, 2.0],
    [1.000001, 1.0, 2.0],
    [3.0, 3.0, 2.0],
], dtype=np.float64)

vals_np, polys_np = polygonize(xr.DataArray(data))
vals_cp, polys_cp = polygonize(xr.DataArray(cp.asarray(data)))

print(vals_np)      # [1.0, 2.0, 3.0]   -> 3 polygons
print(vals_cp)      # [1.0, 1.000001, 2.0, 1.000001, 1.0, 3.0]  -> 6 polygons
```

## Root cause

`xrspatial/polygonize.py::_calculate_regions` (numba CPU path) calls `_is_close` for adjacency checks:

```python
matches_W = ... _is_close(values[ij], values[ij-1])
```

`_is_close` uses `abs(value - reference) <= atol + rtol*abs(reference)` for float types.

`xrspatial/polygonize.py::_calculate_regions_cupy` instead bins pixels by exact value:

```python
unique_vals = cp.unique(data[valid] if valid is not None else data.ravel())
for v in unique_vals:
    bin_mask = (data == v)
    ...
```

## Impact

- HIGH: numerically different results on identical float input across backends.
- The dask and dask+cupy backends both route per-chunk work through `_polygonize_numpy`, so only the pure-cupy backend diverges.
- Documented or not, two of the four backends produce a different polygon set on the same float input.

## Suggested fix

In `_calculate_regions_cupy`, group near-equal float values before per-value labeling. For float dtypes, sort `unique_vals`, walk them, and merge any consecutive pair where the cpu-side `_is_close` predicate would return True. Then label each merged group with a single `(data == merged_values).any(axis=...)` style mask. For integer dtypes, keep the existing exact-equality path (it already matches the numba path).

The check should match `_is_close` exactly so the two backends produce equivalent regions on float input.

Discovered by the `/sweep-accuracy` audit (Cat 5 - Backend Inconsistency).

brendancol:
## Summary

- The cupy polygonize backend grouped float pixels into regions by exact equality (`data == v`), while the numba CPU path used by the numpy / dask / dask+cupy backends compared neighbouring pixels with `_is_close` (`atol=1e-8`, `rtol=1e-5`).
- On float rasters with near-equal values the cupy backend produced more polygons than the other three backends, e.g. values `1.0` and `1.000001` were split into different regions on cupy but merged on numpy.
- Fix: in `_calculate_regions_cupy`, for float dtypes greedily chain sorted unique values using the same tolerance and label each chained group as one region. Integer dtypes keep the existing exact-equality path.

Fixes #2151.

## Test plan

- [x] `pytest xrspatial/tests/test_polygonize.py` (all 115 tests pass, 13 skipped)
- [x] Added regression `test_polygonize_cupy_float_tolerance_matches_numpy_2151` covering the divergence reported in #2151
- [x] Manually re-ran the issue reproducer; both backends now return 3 polygons with values `[1.0, 2.0, 3.0]`

Discovered by the `/sweep-accuracy` audit (Cat 5 - Backend Inconsistency).

### DIFF
--- a/.claude/sweep-accuracy-state.csv
+++ b/.claude/sweep-accuracy-state.csv
@@ -23,7 +23,7 @@ multispectral,2026-03-30T14:00:00Z,1094,,,
 normalize,2026-05-01,,,,rescale and standardize across all 4 backends. NaN/inf filtered via isfinite mask before min/max/mean/std. Constant input handled (range=0 -> new_min; std=0 -> 0.0). Output dtype float64 consistently. Backend parity covered by test_matches_numpy. No accuracy issues found.
 perlin,2026-04-10T12:00:00Z,,,,Improved Perlin noise implementation correct. Fade/gradient functions verified. Backend-consistent. Continuous at cell boundaries.
 polygon_clip,2026-04-13T12:00:00Z,1197,,,crop=True + all_touched=True drops boundary pixels. Fix in PR #1200.
-polygonize,2026-04-13T12:00:00Z,1190,,,NaN pixels not masked in numpy/dask backends. Fix in PR #1194.
+polygonize,2026-05-19,2151,HIGH,5,cupy backend used exact-equality vs numba _is_close tolerance for float dtypes; fixed via tolerance-aware value grouping
 proximity,2026-03-30T15:00:00Z,,,,Direction >= boundary fragile but works due to truncated constant. Float32 truncation is design choice. No wrong-results bugs found.
 reproject,2026-05-17,2025,HIGH,5,"HIGH: _apply_vertical_shift crashed on dask (no fancy-index setitem), cupy (Numba JIT can't take cupy), and 3D multi-band (broadcast mismatch). Fixed by routing dask through map_blocks, round-tripping cupy via host, and looping per band. Previously documented MEDIUM (GPU dispatcher missing NAD27/Helmert wrapper) and MEDIUM (legacy _resample_cupy cubic NaN threshold 1e-6 diverges from CPU bilinear-fallback) still stand. Identity-CRS cupy reproject returns zeros - separate pre-existing bug, not in scope."
 resample,2026-04-14T12:00:00Z,1202,,,Interpolation used edge-aligned coords instead of block-centered. Fix in PR #1204.
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -588,42 +588,17 @@ def _polygonize_numpy(
 _DIR_ANGLE = {(1, 0): 0, (0, 1): 1, (-1, 0): 2, (0, -1): 3}
 
 
-def _group_float_values_by_tolerance(unique_vals_np):
-    """Greedy chaining of sorted float values using the numpy backend's
-    tolerance (matches ``_is_close``: ``atol=1e-8, rtol=1e-5``).
-
-    Returns a list of numpy arrays, one per group of near-equal values.
-    Two consecutive sorted values ``a`` and ``b`` (``a <= b``) are placed in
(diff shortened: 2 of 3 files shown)


brendancol:
## PR Review: polygonize: cupy backend honours float tolerance from numba path (#2151)

I ran this review on commit `2e9b364` and then pushed a follow-up commit (`7bcecfb`) that I think is needed for the stated parity goal. Details below.

### Blocker (fixed in follow-up)

- **`xrspatial/polygonize.py:591-614,650-693`. value-grouping heuristic still diverges from CPU CCL on transitively-close values.** The greedy chain merges `1.0`, `1.000009`, `1.000018` into a single tolerance group because each consecutive pair is within tolerance, but the numba CPU path runs *spatial* CCL with a *pairwise* `_is_close` check. On
  ```
  [1.0,        1.000018]
  [1.000009,   3.0     ]
  ```
  CPU returns 3 regions (`1.0`+`1.000009` adjacent and close, `1.000018` not close to either neighbour, `3.0` alone). The new value-grouping path returns 2 regions, since `cp.isin(data, [1.0, 1.000009, 1.000018])` makes the top-row pair land in the same connected component on cupy.

  The push at `7bcecfb` routes float dtypes through `_polygonize_numpy` inside `_polygonize_cupy`. Data is already transferred to host for boundary tracing in this hybrid backend, so the cost is one extra `_calculate_regions` call on CPU rather than a per-value GPU label pass. The `_group_float_values_by_tolerance` helper and the float branch of `_calculate_regions_cupy` are removed as dead code.

### Suggestions

- The new `test_polygonize_cupy_float_tolerance_matches_numpy_2151` only covered the adjacent-near-equal pattern from the issue. The follow-up parametrizes it with `adjacent_near_equal` and `transitive_non_adjacent` so the second case is locked in.

### Nits

- `xrspatial/polygonize.py:1681-1682`. the existing docstring already documents that cupy data is transferred to CPU for boundary tracing; after the follow-up, that note also covers the float CCL path, so no additional text is needed.

### What looks good

- The diagnosis in the PR body is accurate: the cupy backend's exact-equality binning was the source of the divergence reported in #2151.
- The sweep-state CSV update is consistent with the previous entries.
- The existing tests (`test_polygonize_cupy_matches_numpy`, `test_polygonize_dask_cupy_matches_numpy`) cover integer rasters, and the integer GPU path is unchanged.

### Checklist

- [x] Algorithm matches the numba CPU `_is_close` predicate (with follow-up)
- [x] All implemented backends produce consistent results (with follow-up)
- [x] NaN handling preserved (float path falls through `_polygonize_numpy`, which masks NaN)
- [x] Edge cases covered: parametrized regression test now exercises the transitivity case
- [x] Benchmark not needed (correctness fix, integer GPU path unchanged)
- [x] Docstrings updated to describe the integer-GPU / float-CPU split
- [x] Local `pytest xrspatial/tests/test_polygonize.py` passes (116 passed, 13 skipped) on the post-fix branch

PR #2157, opened 2026-05-20
geotiff: surface rotated_affine 6-tuple on DataArray attrs (#2129)

### LINKED ISSUE #2129: geotiff: surface rotated_affine 6-tuple on the DataArray attrs
Follow-up to #2126 / PR #2128.

When ``open_geotiff(..., allow_rotated=True)`` reads a rotated GeoTIFF, the rotated 6-tuple is held on ``geo_info.transform.rotated_affine`` but ``geo_info`` is not on the returned DataArray. The matrix is therefore unreachable from public code -- callers can only see that ``transform`` is missing and ``crs`` is missing, with no way to recover the actual rotated mapping.

**Proposal**

Emit ``attrs['rotated_affine']`` as a rasterio-style 6-tuple ``(a, b, c, d, e, f)`` on the ``allow_rotated=True`` path. Downstream code that wants to handle rotated rasters (custom warps, visualisation) can read it directly instead of treating the array as fully ungeoreferenced.

The attr should:
- Only appear when the source carried a rotated ``ModelTransformationTag`` (not on plain no-georef reads).
- Stay omitted on round-trip via ``to_geotiff`` until the writer learns to emit ``ModelTransformationTag`` (see #2115 follow-up).

**Why not bundled with #2126**

#2126 fixed the documented contract (CRS attrs were leaking on the rotated path). Adding a new public attr is API surface expansion that deserves its own PR with explicit consideration of the round-trip contract and downstream consumers.

brendancol:
Closes #2129.

## Summary
- Reads opened with `allow_rotated=True` now emit `attrs['rotated_affine']` as a rasterio-style 6-tuple `(a, b, c, d, e, f)` so callers can recover the rotated mapping without reaching into `geo_info.transform`.
- The attr only appears when the source carried a rotated `ModelTransformationTag`; plain no-georef reads and axis-aligned reads do not grow the attr.
- Read-only contract: `attrs_to_metadata` (the writer-side boundary parser) intentionally drops `rotated_affine`, so `to_geotiff` keeps writing a plain no-georef file until the writer learns to emit `ModelTransformationTag` (#2115 follow-up).
- Bumps the attrs contract version from 3 to 4; existing keys keep their pre-v4 shape.

## Backend coverage
All four read backends share the same marshalling step (`_populate_attrs_from_geo_info` → `geo_info_to_metadata` → `metadata_to_attrs`), so numpy, cupy, dask+numpy, and dask+cupy paths all emit the new attr without per-backend code.

## Test plan
- [x] `test_rotated_affine_attr_2129.py` covers the rotated path (with and without CRS), the plain no-georef path, axis-aligned reads, tuple type guarantee, dask path, and the writer round-trip drop contract.
- [x] Existing `test_allow_rotated_*` suites still pass — CRS-drop and no-georef behaviour unchanged.
- [x] Contract-version pin tests in `test_attrs_contract_version_1984.py` and `test_georef_status_2136.py` updated to track v4.
- [x] Full `xrspatial/geotiff/tests/` suite: 4453 passed, 1 pre-existing unrelated failure on `test_lowlevel_write_pushdown_2138::test_write_vs_to_geotiff_byte_parity_uint8[lz4]` (also fails on `main`).

### DIFF
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -63,6 +63,19 @@ write.
        affine transform tuple matching the rasterio ``Affine`` ordering.
        Omitted for files with no ``ModelTransformation`` /
        ``ModelPixelScale`` / ``ModelTiepoint`` tags.
+   * - ``rotated_affine``
+     - tuple
+     - Full ``(a, b, c, d, e, f)`` rasterio-style 6-tuple for files
+       opened with ``allow_rotated=True`` whose source carried a
+       rotated / sheared ``ModelTransformationTag``. The axis-aligned
+       ``transform`` cannot express the rotation terms; this attr
+       surfaces the rotated mapping so downstream code (custom warps,
+       visualisation) can recover it. Only emitted on the rotated
+       opt-in path; absent on axis-aligned reads and on plain
+       no-georef files. Read-only -- ``to_geotiff`` drops the rotation
+       on the way out until the writer learns to emit
+       ``ModelTransformationTag`` (issue #2115 follow-up). See issue
+       #2129.
    * - ``nodata``
      - scalar
      - Numeric NoData sentinel. Emitted by readers when the file
@@ -124,7 +137,7 @@ write.
        ``ResolutionUnit`` ids 1, 2, 3).
    * - ``_xrspatial_geotiff_contract``
      - int
-     - Contract version. Currently ``2``. See `Versioning`_.
+     - Contract version. Currently ``4``. See `Versioning`_.
    * - ``_xrspatial_no_georef``
      - bool
      - Stamped ``True`` on reads of files with no GeoTIFF transform
@@ -328,7 +341,7 @@ Versioning
 ==========
 
 The contract is versioned through ``attrs['_xrspatial_geotiff_contract']``.
-The current value is ``2``. Future revisions that add canonical keys,
+The current value is ``4``. Future revisions that add canonical keys,
 move keys between tiers, or change a key's semantics will bump the
 integer. Callers that depend on a specific layout can branch on the
 version, and writers will emit the version they were built against.
@@ -344,3 +357,17 @@ and matplotlib-colormap attrs that v1 emitted on read under a
 ``attrs[key]`` will now see ``KeyError``; switch to ``attrs.get(key)``
 or migrate to the canonical ``crs`` / ``crs_wkt`` plus :mod:`pyproj`
 recipe documented in `Removed in contract v2`_.
+
+Contract v3 (issue #2136) added the ``georef_status`` attr to the
+canonical tier, encoding the five distinct states the reader can
+land in (``full``, ``transform_only``, ``crs_only``, ``none``,
+``rotated_dropped``) so downstream code can branch on a single
+value instead of reconstructing the state from the union of ``crs``,
+``crs_wkt``, ``transform``, and ``_xrspatial_no_georef``.
+
+Contract v4 (issue #2129) added the ``rotated_affine`` attr to the
+canonical tier. The attr surfaces the rotated 6-tuple from
+``ModelTransformationTag`` on the ``allow_rotated=True`` opt-in path
+so callers can recover the rotated mapping. The writer drops it on
+round-trip until ``to_geotiff`` learns to emit
+``ModelTransformationTag`` (issue #2115 follow-up).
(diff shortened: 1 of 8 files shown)


brendancol (commented):
## PR Review: geotiff: surface rotated_affine 6-tuple on DataArray attrs (#2129)

### Blockers
None.

### Suggestions
- [ ] **VRT backend does not emit `rotated_affine` for the same condition.** `xrspatial/geotiff/_backends/vrt.py:277` detects rotated GDAL GeoTransforms via `_vrt_is_rotated = (gt[2] != 0.0 or gt[4] != 0.0)` and routes the array into the `rotated_dropped` bucket on both the eager (line 336) and chunked (line 842) builds. Neither `GeoTIFFMetadata(...)` site populates the new `rotated_affine` field, so a VRT user who opens a rotated VRT with `allow_rotated=True` still cannot recover the 6-tuple. Same gap #2129 closes for the `ModelTransformationTag` path. The 6-tuple is reconstructable from `vrt.geo_transform` via the existing `_gdal_geotransform_to_affine_tuple` helper (used by the non-rotated VRT branch at line 597). Adding `rotated_affine=_gdal_geotransform_to_affine_tuple(vrt.geo_transform) if _vrt_is_rotated else None` to both builds keeps the backends consistent.

### Nits
- [ ] `test_geotiff_metadata_2139.py::_representative_attrs_dicts` (line 237) drives the symmetric round-trip tests. The new attr is intentionally asymmetric (read-only), so it does not belong there — but a short comment near the helper saying so would prevent a future contributor from adding it and hitting a confusing failure.
- [ ] No GPU-backed end-to-end test for the new attr. The GPU read path shares the marshalling step with the eager/dask paths so the unit tests cover the logic, but a `@_gpu_only` sibling would close the matrix.

### What looks good
- Flows through the existing `GeoTIFFMetadata` / `metadata_to_attrs` step rather than adding a parallel code path.
- Read-only round-trip contract is explicit and pinned: `attrs_to_metadata` does not parse the attr back, and `test_attrs_to_metadata_drops_rotated_affine` will fail loudly if that ever changes.
- `tuple(src_t.rotated_affine)` cast is defensive against a parser change to list/ndarray; `test_rotated_affine_is_tuple_not_list` pins the type.
- Contract version bump from 3 to 4 is paired with both pin-test updates and an `attrs_contract.rst` versioning note.
- `test_open_geotiff_axis_aligned_omits_rotated_affine` is a useful negative test: an axis-aligned `ModelTransformationTag` still rides on `attrs['transform']` and the new attr stays off.

### Checklist
- [x] Algorithm matches reference (N/A — metadata)
- [x] All implemented backends consistent (caveat: VRT gap above)
- [x] NaN handling correct (N/A)
- [x] Edge cases tested (rotated with/without CRS, plain no-georef, axis-aligned, dask, tuple type, writer-side drop)
- [x] Dask chunk boundaries handled (`test_open_geotiff_rotated_emits_rotated_affine_dask`)
- [x] Benchmark exists or not needed (not needed)
- [x] README feature matrix updated (N/A — no new function)
- [x] Docstrings updated (`open_geotiff` + `attrs_contract.rst`)

brendancol (commented):
## PR Review: follow-up pass on a4c48d5

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- VRT rotated_affine gap closed at both build sites (`_backends/vrt.py:343` eager, `_backends/vrt.py:854` chunked); `_gdal_geotransform_to_affine_tuple` correctly maps GDAL ordering to rasterio Affine ordering, so the VRT surface matches the non-VRT `ModelTransformationTag` surface.
- The new ``_vrt_rotated_affine`` is computed only when ``_vrt_is_rotated``; gating mirrors the existing ``rotated_dropped`` gate so the two attrs stay in lockstep.
- VRT tests added (eager + chunked + axis-aligned negative) and the symmetric round-trip helper now carries an inline note explaining why `rotated_affine` is excluded — so a future contributor does not waste cycles wondering.
- Lazy import of `_gdal_geotransform_to_affine_tuple` is already in scope at both modified sites (existing imports at lines 231 and 603).

### Disposition of original review
- Suggestion (VRT gap): **fixed** in a4c48d5.
- Nit (`_representative_attrs_dicts` doc note): **fixed** in a4c48d5.
- Nit (GPU-marked test): **dismissed.** The GPU read path shares `_populate_attrs_from_geo_info` with eager/dask, which `test_rotated_optin_emits_rotated_affine_tuple` already exercises directly. A GPU-marked sibling would only confirm the shared helper still runs under cupy, which the four other GPU rotated tests in the suite already cover at the `attrs['crs']` / `attrs['transform']` level.

Nothing left to fix.

Issue #2074, opened 2026-05-18
open_geotiff overview_level accepts bool and non-int types, returning wrong resolution silently


brendancol:
## Description

`open_geotiff(..., overview_level=...)` does not validate the type of `overview_level`. In `xrspatial/geotiff/_header.py:794-808`, the selector assigns `level = overview_level` and compares it numerically. Because `bool` is a subclass of `int` in Python, `overview_level=True` is coerced to `1` and the function returns overview level 1 with no error. Non-int types (`str`, `float`) leak raw `TypeError` messages from the internal comparison or list indexing.

Downstream code then gets a wrong-resolution array with no signal that anything was off.

## Reproducer

```python
import numpy as np
import tifffile
from xrspatial.geotiff import open_geotiff

arr = np.random.randint(0, 256, size=(64, 64), dtype=np.uint8)
half = arr[::2, ::2]
with tifffile.TiffWriter("cog.tif") as tw:
    tw.write(arr, tile=(32, 32), photometric="minisblack")
    tw.write(half, tile=(32, 32), photometric="minisblack", subfiletype=1)

# Bug 1: bool is silently coerced to an int level
open_geotiff("cog.tif", overview_level=True).shape   # (32, 32)
open_geotiff("cog.tif", overview_level=False).shape  # (64, 64)

# Bug 2: raw TypeErrors from internal comparison/indexing
open_geotiff("cog.tif", overview_level="0")
# TypeError: '<' not supported between instances of 'str' and 'int'

open_geotiff("cog.tif", overview_level=1.0)
# TypeError: list indices must be integers or slices, not float
```

## Expected behavior

`overview_level` should be an `int` (>= 0) or `None`. Any other type should raise a clear `TypeError` that names the rejected type and value.

## Source location

`xrspatial/geotiff/_header.py:794-808` in `select_overview_ifd`. The public entrypoint is `open_geotiff` in `xrspatial/geotiff/__init__.py:243`.

## Environment

xarray-spatial-contrib `main`, Python 3.12.

PR #2163, opened 2026-05-20
geotiff: validate overview_level up front on dask/gpu backends

### LINKED ISSUE #2160: Direct public backends do not share overview_level validation ordering
**Describe the bug**

`open_geotiff` rejects invalid `overview_level` arguments up front at `xrspatial/geotiff/__init__.py:513` via `_validate_overview_level_arg`. The direct backends do not share that early check. `read_geotiff_dask` (`xrspatial/geotiff/_backends/dask.py:108`) runs source coercion and chunk validation first, then leaves the type check to `select_overview_ifd` inside the metadata path. `read_geotiff_gpu` (`xrspatial/geotiff/_backends/gpu.py:324`) reaches `select_overview_ifd` the same way.

The selector does call the validator, so `True` / `"0"` / `1.0` eventually raise `TypeError`. The bug is the ordering. A caller who passes both a bad source (or bad `chunks=`) and a bad `overview_level` gets the source / chunk error, not the type error, so the real defect in the call is hidden. `open_geotiff` reports the `overview_level` problem first; the direct backends do not.

Reproduction:

```python
from xrspatial.geotiff import read_geotiff_dask, open_geotiff

open_geotiff("/nonexistent.tif", overview_level=True)
# TypeError: overview_level must be an int or None, got bool: True

read_geotiff_dask("/nonexistent.tif", overview_level=True)
# FileNotFoundError: [Errno 2] No such file or directory: '/nonexistent.tif'

read_geotiff_dask("real.tif", chunks=0, overview_level=True)
# ValueError: chunks must be a positive int ...
```

**Expected behavior**

All three public entry points (`open_geotiff`, `read_geotiff_dask`, `read_geotiff_gpu`) should validate `overview_level` first so the same call raises the same error regardless of which entry point you reach.

**Additional context**

The issue #2074 test module only covers `open_geotiff`. The direct backends are public, so they need their own coverage. Implementation: call the existing `_validate_overview_level_arg(...)` validator at the top of each public backend, before `_coerce_path` / `_validate_chunks_arg`. Mirror the issue #2074 tests for `read_geotiff_dask` and `read_geotiff_gpu`.

Scope: only the `overview_level` validation ordering. The three sibling findings (`mask_nodata=False` not honored, `masked_nodata` attrs lying, gpu HTTP source handling) are out of scope for this issue.

brendancol:
Closes #2160.

## Summary

- `open_geotiff` rejected `overview_level=True` / `"0"` / `1.0` up front, but `read_geotiff_dask` and `read_geotiff_gpu` only reached the same check inside `select_overview_ifd`, after source coercion, chunk validation, and (on GPU) `on_gpu_failure` alias handling. A bad `overview_level` paired with a bad source or bad `chunks=` surfaced the source / chunk error first, hiding the real defect.
- Call the existing `_validate_overview_level_arg` validator at the top of both direct backends so the user-facing error matches `open_geotiff` regardless of which entry point is hit. `select_overview_ifd` still validates as defense in depth.
- Mirror the issue #2074 tests against both backends, plus ordering checks: the type error must fire before `_coerce_path`, `_validate_chunks_arg`, and the `on_gpu_failure` policy check.

## Backend coverage

- numpy: not applicable (eager numpy path goes through `open_geotiff`, already validated)
- cupy: covered via `read_geotiff_gpu` (CPU-machine path; validator fires before the cupy import)
- dask+numpy: covered via `read_geotiff_dask`
- dask+cupy: covered via `read_geotiff_gpu(chunks=...)` which routes through `read_geotiff_dask` after the validator fires

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_overview_level_validation_backends_2160.py` (22 passed)
- [x] `pytest xrspatial/geotiff/tests/test_overview_level_type_validation_2074.py` (regression check, 11 passed)
- [x] `pytest xrspatial/geotiff/tests/test_overview_filter.py` (selector defense-in-depth still works)

### DIFF
--- a/xrspatial/geotiff/_backends/dask.py
+++ b/xrspatial/geotiff/_backends/dask.py
@@ -109,14 +109,10 @@ def read_geotiff_dask(source: str, *,
 
     from .._reader import _coerce_path
 
-    # Reject bool and non-int ``overview_level`` up front (issue #2160).
-    # ``open_geotiff`` runs the same check at its entry point; without
-    # this guard a caller who passes a bad ``overview_level`` together
-    # with a bad source or bad ``chunks=`` gets the unrelated source /
-    # chunk error first, so the real defect in the call is masked.
-    # ``select_overview_ifd`` does still validate later as defense in
-    # depth, but the user-facing error is supposed to match
-    # ``open_geotiff`` regardless of which public entry point is hit.
+    # Match ``open_geotiff``'s ordering so a bad ``overview_level`` is
+    # reported before unrelated source / ``chunks=`` errors mask it
+    # (issue #2160). ``select_overview_ifd`` revalidates as defense in
+    # depth.
     _validate_overview_level_arg(overview_level)
 
     source = _coerce_path(source)
--- a/xrspatial/geotiff/_backends/dask.py
+++ b/xrspatial/geotiff/_backends/dask.py
@@ -27,7 +27,11 @@ from .._coords import (
     geo_to_coords as _geo_to_coords,
 )
 from .._reader import read_to_array as _read_to_array
-from .._validation import _validate_chunks_arg, _validate_dtype_cast
+from .._validation import (
+    _validate_chunks_arg,
+    _validate_dtype_cast,
+    _validate_overview_level_arg,
+)
 from .vrt import read_vrt
 
 
@@ -104,6 +108,16 @@ def read_geotiff_dask(source: str, *,
     import dask.array as da
 
     from .._reader import _coerce_path
+
+    # Reject bool and non-int ``overview_level`` up front (issue #2160).
+    # ``open_geotiff`` runs the same check at its entry point; without
+    # this guard a caller who passes a bad ``overview_level`` together
+    # with a bad source or bad ``chunks=`` gets the unrelated source /
+    # chunk error first, so the real defect in the call is masked.
+    # ``select_overview_ifd`` does still validate later as defense in
+    # depth, but the user-facing error is supposed to match
+    # ``open_geotiff`` regardless of which public entry point is hit.
+    _validate_overview_level_arg(overview_level)
 
     source = _coerce_path(source)
 
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -193,15 +193,10 @@ def read_geotiff_gpu(source: str, *,
     xr.DataArray
         CuPy-backed DataArray on GPU device.
     """
-    # Reject bool and non-int ``overview_level`` up front (issue #2160).
-    # ``open_geotiff`` runs the same check at its entry point; without
-    # this guard a caller who passes a bad ``overview_level`` together
-    # with a bad ``on_gpu_failure``, bad ``chunks=``, or bad source gets
-    # the unrelated kwarg / source error first, so the real defect in
-    # the call is masked. ``select_overview_ifd`` does still validate
-    # later as defense in depth, but the user-facing error is supposed
-    # to match ``open_geotiff`` regardless of which public entry point
-    # is hit.
+    # Match ``open_geotiff``'s ordering so a bad ``overview_level`` is
+    # reported before unrelated ``on_gpu_failure`` / ``chunks=`` / source
+    # errors mask it (issue #2160). ``select_overview_ifd`` revalidates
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: geotiff: validate overview_level up front on dask/gpu backends

None.

None.

- [ ] `xrspatial/geotiff/_backends/dask.py:112-120` and `xrspatial/geotiff/_backends/gpu.py:196-205`: the two new comment blocks are near duplicates. The justification is the same on both backends (parity with `open_geotiff`, ordering matters, selector is defense-in-depth). Trimming each to one or two lines and pointing at the issue would read the same and save 14 lines of comment for a two-line validator call.
- [ ] `xrspatial/geotiff/tests/test_overview_level_validation_backends_2160.py`: no explicit test for the GPU chunked dispatch (`read_geotiff_gpu(chunks=...)`) path. The validator fires before the `chunks is not None` branch on line 263, so the eager-path tests already cover it via code flow, but a one-line parametrised case would document the contract. Not load-bearing; skip if you want to keep the file tight.
- [ ] `xrspatial/geotiff/_backends/gpu.py:205`: minor wording in the comment block -- "bad `on_gpu_failure`" is the only kwarg name not in code-formatted backticks in either comment ("source", "chunks=" both do). Cosmetic.

### What looks good
- The fix calls the existing `_validate_overview_level_arg` rather than copying logic, so `_validate_overview_level_arg` stays the single source of truth across `open_geotiff`, `select_overview_ifd`, and now both direct backends.
- Tests mirror the structure of `test_overview_level_type_validation_2074.py` so coverage is symmetric across the three public entry points (bool, str, float, numpy ints, None, ordering against bad source / bad chunks).
- The GPU test file imports `read_geotiff_gpu` and exercises the validator without touching cupy. The validator sits before the cupy import (line 230), so the bad-input cases run on CPU-only CI without `importorskip`-ing the whole file. Good ordering call.
- The ordering-check test against `on_gpu_failure="bogus"` (`test_gpu_overview_level_check_runs_before_on_gpu_failure_validation`) directly asserts the fix's purpose -- the type error has to beat every other kwarg validator, not just source / chunks.
- Tmp filenames carry `_2160` so they will not collide with parallel rockouts on the sibling issues running against the same `tmp_path`.
- Scope is clean. Three sibling issues (mask_nodata, masked_nodata attrs, gpu HTTP) are not bundled in.

### Checklist
- [x] Algorithm matches reference/paper -- n/a (validation only)
- [x] All implemented backends produce consistent results -- dask, gpu now match `open_geotiff`
- [x] NaN handling is correct -- n/a (no array math touched)
- [x] Edge cases are covered by tests -- bool both ways, str, float, numpy int32/int64, None, 0, 1, ordering against three other validators
- [x] Dask chunk boundaries handled correctly -- n/a
- [x] No premature materialization or unnecessary copies -- n/a
- [x] Benchmark exists or is not needed -- not needed (entry-point validation, microseconds)
- [x] README feature matrix updated -- not applicable (no new function, no backend support change)
- [x] Docstrings present and accurate -- existing docstrings still valid; behaviour-change is invisible to docs because the error type is the same one the selector raised, just earlier

brendancol (commented):
## Follow-up review: post nit cleanup

### Status
All three nits from the previous review are addressed in 53dad3d.

- Nit 1 (duplicate comment blocks): the two near-duplicate justifications on `xrspatial/geotiff/_backends/dask.py` and `xrspatial/geotiff/_backends/gpu.py` are collapsed to one-sentence comments each pointing at the issue. The selector defense-in-depth note survives in shorter form.
- Nit 2 (no explicit GPU chunked dispatch test): `test_gpu_overview_level_check_runs_before_chunked_dispatch` now asserts the validator fires before the `chunks is not None` branch of `read_geotiff_gpu` routes through `_read_geotiff_gpu_chunked`. Test count is 23, all passing locally.
- Nit 3 (code-format `on_gpu_failure` in the GPU comment): the trimmed comment now uses backticks consistently across the three named kwargs.

### Verification
- `pytest xrspatial/geotiff/tests/test_overview_level_validation_backends_2160.py` -- 23 passed.
- `pytest xrspatial/geotiff/tests/test_overview_level_type_validation_2074.py xrspatial/geotiff/tests/test_overview_filter.py` -- regression check, no new failures.

No remaining open items.

PR #2164, opened 2026-05-20
geotiff: honor mask_nodata=False for float VRT sources (#2158)

### LINKED ISSUE #2158: read_vrt(mask_nodata=False) silently masks float source sentinels to NaN
**Describe the bug**

`mask_nodata=False` is not honored for float VRT sources.

`xrspatial/geotiff/_vrt.py:1336` masks float source nodata to NaN inside the internal VRT reader before the public VRT backend ever sees `mask_nodata`. The public wrapper only skips the integer helper at `xrspatial/geotiff/_backends/vrt.py:389`, so `read_vrt(..., mask_nodata=False)` still rewrites float sentinels. The chunked path hits the same bug because `_vrt_chunk_read` calls the same internal reader before its `mask_nodata` check at `xrspatial/geotiff/_backends/vrt.py:525`.

The documented opt-out is silently ignored, which can drop literal sentinel values downstream.

**Expected behavior**

`read_vrt(..., mask_nodata=False)` should leave the source dtype and the literal sentinel value intact for float source bands, just like it does for integer source bands. The eager and chunked paths should agree.

**Reproduction**

A VRT pointing at a single float32 source where the source band declares `<NoDataValue>-9999.0</NoDataValue>` and the array contains literal `-9999.0` pixels.

* `read_vrt(vrt, mask_nodata=True)` -> pixels at the sentinel value become NaN. Correct.
* `read_vrt(vrt, mask_nodata=False)` -> pixels at the sentinel value still become NaN. Wrong: the caller asked to keep them as `-9999.0`.

Same outcome with `chunks=` set, via `_vrt_chunk_read` -> `_read_vrt_internal` -> `_read_data`.

**Additional context**

The internal reader's NaN masking lives in `_vrt._read_data` and runs before the dtype cast or sentinel scan in the public backend. Fix is to thread the kwarg into the internal reader (or move the float masking out of the per-source loop, up to the same level as the integer helper) so both paths honor the opt-out symmetrically.

Related: issue #2052 added `mask_nodata=False` as the documented opt-out for integer sources. Float sources should follow the same contract.

brendancol:
## Summary

- Thread `mask_nodata` from the public VRT backend into `_vrt.read_vrt` so the float NaN masking inside `_read_data` honors the opt-out. Gate the integer-source-feeding-float-VRT promotion on the same kwarg so a single opt-out covers every inline masking path.
- AND `attrs['masked_nodata']` with `mask_nodata` on both eager and chunked VRT paths so the attr does not lie when the inline masking is skipped.
- Add regression tests covering the eager and chunked paths, fractional float64 sentinels, and masked vs unmasked bit-equivalence on non-sentinel pixels.

Closes #2158.

## Backend coverage

- numpy eager: covered.
- dask+numpy chunked: covered.
- cupy / dask+cupy: the GPU VRT backend reuses the same internal reader through the same call site; no separate kwarg plumbing was needed.

## Test plan

- [x] New regression tests in `test_vrt_mask_nodata_float_source_2158.py` (6 tests, all pass).
- [x] Existing `test_mask_nodata_kwarg_2052.py`, `test_mask_nodata_gpu_vrt_2052.py`, `test_vrt_int_source_float_dtype_1616.py`, `test_masked_nodata_attr_2092.py` still pass.
- [x] Full `vrt or nodata` slice (990 tests) passes.

### DIFF
--- a/xrspatial/geotiff/_backends/vrt.py
+++ b/xrspatial/geotiff/_backends/vrt.py
@@ -415,20 +415,30 @@ def read_vrt(source: str, *,
     # follow-up).
     pre_cast_dtype = np.dtype(str(arr.dtype))
 
-    # When the inline float-NaN masking inside ``_vrt._read_data`` already
-    # ran (float source + float VRT dataType + declared sentinel), the
-    # integer helper above is a no-op but the resulting float buffer may
-    # already carry NaNs at the sentinel locations. Promote
-    # ``nodata_pixels_present`` to reflect that, since a downstream "any
-    # nodata?" check should answer yes for those tiles too.
+    # Float-NaN proxy for ``nodata_pixels_present``. The signal for a
+    # float buffer can come from any of three places, only one of which
+    # actually fires per call:
+    #
+    # * ``mask_nodata=True`` + float source + declared sentinel: the
+    #   inline NaN masking in ``_vrt._read_data`` rewrote sentinels to
+    #   NaN, so ``np.isnan(arr).any()`` is the presence answer.
+    # * ``mask_nodata=False`` + declared sentinel: ``_vrt_scan_for_sentinel``
+    #   above already set ``nodata_pixels_present`` from a literal-value
+    #   scan, so this block short-circuits via ``not present``.
+    # * ``mask_nodata=True`` + float source with no inline masking (e.g.
+    #   source's natural NaN values, sentinel absent from window): the
+    #   NaN proxy catches the in-buffer NaNs.
     #
     # Invariant: at most one branch sets ``nodata_pixels_present`` to True
     # before this block runs. ``_vrt_mask_with_presence`` only sets True
     # on integer buffers (it short-circuits on float dtype), and
     # ``_vrt_scan_for_sentinel`` is gated on the ``mask_nodata=False``
     # branch above. Either way, the float-NaN proxy below is the only
-    # presence signal for float buffers, so the ``not present`` guard
-    # is sufficient -- we will not double-scan the same buffer.
+    # remaining presence signal for float buffers, so the ``not present``
+    # guard is sufficient -- we will not double-scan the same buffer.
+    # The ``mask_nodata=False`` arm fix from #2158 means the inline NaN
+    # masking no longer runs under the opt-out; the ``_vrt_scan_for_sentinel``
+    # short-circuit above is what keeps the presence attr honest there.
     if (nodata is not None
             and pre_cast_dtype.kind == 'f'
             and not nodata_pixels_present):
--- a/xrspatial/geotiff/_backends/vrt.py
+++ b/xrspatial/geotiff/_backends/vrt.py
@@ -248,9 +248,20 @@ def read_vrt(source: str, *,
         ),
     })
 
+    # Thread ``mask_nodata`` into the internal reader so the float
+    # source NaN masking inside ``_read_data`` honors the opt-out
+    # alongside the integer ``_vrt_mask_with_presence`` helper called
+    # further down. Without this the eager path silently rewrote
+    # literal float sentinels (e.g. ``-9999.0``) to NaN regardless of
+    # the kwarg, which violates the documented opt-out and can drop
+    # legitimate sentinel pixels from downstream consumers. The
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: geotiff: honor mask_nodata=False for float VRT sources (#2158)

The fix correctly threads the kwarg through both call sites of `_read_vrt_internal` (eager and chunked) and gates the two inline masking branches in `_vrt._read_data`. The eager and chunked `masked_nodata` attrs are also kept honest. Test coverage is reasonable for the float source path. A few smaller items below.

- None.

- [ ] `xrspatial/geotiff/_backends/vrt.py:418-431`: the comment block above the float-NaN proxy still claims "the inline float-NaN masking inside `_vrt._read_data` already ran". After this PR, that masking is gated on `mask_nodata`, so under `mask_nodata=False` the proxy's preconditions are different: `_vrt_scan_for_sentinel` (line 405) provides the answer, and the NaN proxy only acts as a tiebreaker when the buffer naturally contains NaNs. Update the comment so a future reader does not trust the stale invariant.
- [ ] `tests/test_vrt_mask_nodata_float_source_2158.py`: only the float-source-feeding-float-VRT path is tested. The integer-source-feeding-float-VRT promotion at `_vrt.py:1351-1390` is the other branch this PR gated; without a regression test it could silently re-promote in a future refactor. A spot check confirmed the behavior is correct (uint16 65535 surfaces as 65535.0 under `mask_nodata=False`), so it is a matter of locking in the contract.

- [ ] `xrspatial/geotiff/_vrt.py:1357-1361`: the `if not mask_nodata: pass / elif ...` idiom reads awkwardly. Wrapping the whole sentinel-handling block in `if mask_nodata:` is the more idiomatic form and avoids a no-op branch.
- [ ] `tests/test_vrt_mask_nodata_float_source_2158.py:21`: `pytest` is imported but never used. Drop it or use it.

### What looks good
- The kwarg is plumbed symmetrically through both eager and chunked entry points.
- The `attrs['masked_nodata']` flag is AND-ed with `mask_nodata` on both paths, so the attr does not lie when the inline masking is skipped.
- Test file naming follows the project convention with a unique `_2158` suffix.
- The fix scope is tight: only the two call sites of `_read_vrt_internal` and the single gating block in `_read_data`. No collateral refactor.

### Checklist
- [x] Algorithm matches reference / contract (`_set_nodata_attrs` documented contract: `mask_nodata and final_dtype.kind == 'f'`).
- [x] All implemented backends produce consistent results (eager and chunked share `_read_vrt_internal`).
- [x] NaN handling is correct (no NaN substitution under the opt-out; sentinel survives).
- [x] Edge cases are covered: fractional float64 sentinel, windowed reads (spot-checked).
- [x] Dask chunk boundaries handled correctly (chunked path uses same internal reader; per-chunk parity test passes).
- [ ] Benchmark exists or is not needed: not needed; the change is a boolean gate inside an existing loop, no new compute.
- [ ] README feature matrix updated: not applicable; no new public function.
- [x] Docstrings present and accurate: internal `read_vrt` docstring updated.

brendancol (commented):
## PR Review (follow-up): geotiff: honor mask_nodata=False for float VRT sources (#2158)

Re-checked after follow-up commit f106bf1. All items from the previous review have been addressed.

- Suggestion (stale comment at `_backends/vrt.py:418-431`): fixed in f106bf1 — comment now enumerates all three branches that can set `nodata_pixels_present` under the new gating.
- Suggestion (missing int-source-float-VRT coverage): fixed in f106bf1 — two new tests pin the contract for both `mask_nodata=True` and `mask_nodata=False` on the integer-source-feeding-float-VRT promotion branch.
- Nit (awkward `if not mask_nodata: pass`): fixed in f106bf1 — collapsed into `mask_nodata and ...` guards on both arms.
- Nit (unused pytest import): fixed in f106bf1.

### New findings
- None.

### What looks good after follow-up
- The gating expression in `_vrt.py:1357-1366` reads cleanly now.
- The new tests are bit-exact assertions, not "does it run" smoke tests.
- Test count: 8 in the new file, all passing; 37 passing across `2158 + 2052 + 1616 + 2092` slices.

PR #2165, opened 2026-05-20
geotiff: VRT backends gate masked_nodata on mask_nodata (#2159)

### LINKED ISSUE #2159: VRT backends: masked_nodata attr ignores mask_nodata=False
**Describe the bug**

The two VRT read paths compute `attrs['masked_nodata']` from the buffer dtype alone:

```python
# eager VRT (xrspatial/geotiff/_backends/vrt.py:462)
_set_nodata_attrs(
    attrs, nodata,
    masked=(pre_cast_dtype.kind == 'f'),
    ...
)

# chunked VRT (xrspatial/geotiff/_backends/vrt.py:883)
_set_nodata_attrs(
    attrs, nodata_meta,
    masked=(declared_dtype.kind == 'f'),
    ...
)
```

When the caller passes `mask_nodata=False` on a float-source VRT, neither path runs the sentinel-to-NaN replacement, but the buffer dtype stays float. The dtype-only rule then reports `masked_nodata=True` even though the literal sentinel values are still sitting in the buffer.

The dask, GPU, and dask+GPU backends already gate on both signals (`mask_nodata and dtype.kind == 'f'`):
- `xrspatial/geotiff/_backends/dask.py:373`
- `xrspatial/geotiff/_backends/gpu.py:472-473`
- `xrspatial/geotiff/_backends/gpu.py:844-845`
- `xrspatial/geotiff/_backends/gpu.py:1404`

The contract at `xrspatial/geotiff/_attrs.py:723` says the same thing: `masked` should be `mask_nodata and final_dtype.kind == 'f'`. The two VRT call sites are the outliers.

`attrs['masked_nodata']` is described at `xrspatial/geotiff/_attrs.py:668` as a lifecycle truth signal. Downstream code that uses it to decide whether to restore sentinels on write gets the wrong answer on the VRT path when `mask_nodata=False`.

**Expected behavior**

Both VRT paths report `masked_nodata=True` only when the caller actually requested masking AND the resulting buffer dtype is float. This matches the other three backends and the contract.

**Scope**

This issue only covers the truthfulness of the `masked_nodata` attr on the VRT backends. The actual masking behavior under `mask_nodata=False` on the VRT path is a separate finding being addressed in a parallel PR. Do not touch the masking logic here, only the attr computation.

**Additional context**

- Eager VRT call site: `xrspatial/geotiff/_backends/vrt.py:462`
- Chunked VRT call site: `xrspatial/geotiff/_backends/vrt.py:883`
- Contract definition: `xrspatial/geotiff/_attrs.py:668`
- Sibling backends with correct logic: `_backends/dask.py:371`, `_backends/gpu.py:470`, `_backends/gpu.py:842`, `_backends/gpu.py:1402`

brendancol:
Closes #2159

## Summary

- Both VRT read paths (`vrt.py:462` eager, `vrt.py:883` chunked) computed `attrs['masked_nodata']` from the buffer dtype alone, so a float-source VRT with `mask_nodata=False` reported `masked_nodata=True`.
- The dask, GPU, and dask+GPU backends already use `mask_nodata and dtype.kind == 'f'`; the contract at `_attrs._set_nodata_attrs` says the same. The two VRT call sites now match.
- Scope is limited to the attr computation. The behavioural mismatch where the VRT internal reader still NaN-masks float sources under `mask_nodata=False` is tracked separately in #2158.

## Backend coverage

VRT eager (numpy + GPU) and VRT chunked (dask + dask+GPU). All four share the two call sites touched here. No change to numpy / dask / GPU / dask+GPU non-VRT paths.

## Test plan

- [x] New `test_vrt_masked_nodata_attr_2159.py` with 9 tests covering: float source + mask off (both paths), float source + mask on (regression guard), int source + mask off, float source + explicit dtype cast + mask off, cross-backend attr parity.
- [x] Existing `test_masked_nodata_attr_2092.py` (12 tests), `test_nodata_semantics_split_1988.py` (39 tests), `test_mask_nodata_gpu_vrt_2052.py` (19 tests), `test_read_vrt_lazy_chunks_1798.py` (3 tests) all still pass.
- [x] Full `xrspatial/geotiff/tests/` run: 4456 passed (2 failures are pre-existing on `main`, confirmed by stash + retest).


brendancol (commented):
## PR Review: geotiff: VRT backends gate masked_nodata on mask_nodata (#2159)

None.

None.

- [ ] `xrspatial/geotiff/_backends/vrt.py:878-887` — the older comment block above the new one still says `declared_dtype.kind == 'f'` is "the correct gate for 'buffer is NaN-aware.'" That was true under the dtype-only rule but now reads as a contradiction with the new gate that adds `mask_nodata`. Tightening the older block to point at the new gate would save the next reader a confused minute.
- [ ] `xrspatial/geotiff/_backends/vrt.py:462-472` — the new comment says "the inline NaN-masking inside `_vrt._read_data` and the integer-promotion helper are both skipped" under `mask_nodata=False` on a float source. The integer helper IS gated on `mask_nodata` (`vrt.py:389`), but the inline float NaN-masking in `_vrt.read_vrt` at `_vrt.py:1339` is NOT gated on `mask_nodata` today (that's the sibling bug #2158). The comment is forward-looking and lines up with the contract, but a reader who greps for it today and finds the unconditional masking will be confused. Calling out "(sibling #2158 covers the case where the inline reader still masks float sources today)" inside the comment block makes the forward intent explicit.

### What looks good
- Fix is targeted: two lines of logic change, matching the contract verbatim and aligned with the three other backends (`dask.py:373`, `gpu.py:472`, `gpu.py:844`, `gpu.py:1404`).
- New test file covers both VRT paths in both directions (mask on vs off), the int-source regression case, the explicit-dtype-cast edge, and a cross-backend attr parity check. Temp filenames carry the issue number per the project's collision-avoidance rule.
- Comments at both call sites cite the contract location and the related issues, so the rationale stays local.
- Scope discipline: masking behaviour is not touched, sibling issue #2158 is called out in the PR body and the test docstring.
- Pre-existing tests in `test_masked_nodata_attr_2092.py`, `test_nodata_semantics_split_1988.py`, `test_mask_nodata_gpu_vrt_2052.py`, and `test_read_vrt_lazy_chunks_1798.py` still pass.

### Checklist
- [x] Algorithm matches contract (`_attrs._set_nodata_attrs`: `mask_nodata and final_dtype.kind == 'f'`)
- [x] All implemented backends produce consistent attrs (new cross-backend parity test pins this)
- [x] NaN handling unchanged (this PR does not touch masking)
- [x] Edge cases covered (int source, float source, explicit dtype cast, both masking directions)
- [x] Dask chunk boundaries — N/A, attr is computed once at graph construction time, not per-chunk
- [x] No premature materialization (chunked path keeps `pixels_present=None`)
- [x] Benchmark — N/A, attr computation is one boolean
- [x] README feature matrix — N/A, no API surface change
- [x] Docstrings — call sites have new explanatory comments; module-level test docstring explains scope

brendancol (commented):
## PR Review: follow-up after comment fixes (#2159)

### Blockers
None.

### Suggestions
None.

### Nits
None remaining. The two earlier nits about contradictory and incomplete call-site comments are addressed in `f276c1f`:
- Eager site (`vrt.py:462-477`) now scopes the "skipped" claim to the integer-promotion helper and cross-references sibling #2158 for the inline float-NaN masking that still runs today.
- Chunked site (`vrt.py:878-902`) drops the stale "`declared_dtype.kind == 'f'` is the correct gate" paragraph and inlines the new gate's rationale.

### What looks good
- Logic untouched on the follow-up commit — only comments changed.
- Test suite still passes (`test_vrt_masked_nodata_attr_2159.py`, `test_masked_nodata_attr_2092.py`, `test_read_vrt_lazy_chunks_1798.py` all green).
- Disposition: 2 nits fixed, 0 deferred, 0 dismissed.

Issue #1809, opened 2026-05-13
MinIsWhite inversion runs before nodata mask, swapping data and nodata pixels


brendancol:
## Summary

`_apply_photometric_miniswhite` runs **before** the nodata-sentinel-to-NaN mask in every read path. For a TIFF that is both `Photometric=MinIsWhite` (tag 262 == 0) and carries a `GDAL_NODATA` tag, the inversion changes the pixel values that the mask then compares against the original sentinel, so:

- pixels whose stored value equalled the nodata sentinel survive as `iinfo.max - sentinel` (uint) or `-sentinel` (float) instead of becoming `NaN`
- pixels whose stored value equalled `iinfo.max - sentinel` / `-sentinel` are incorrectly converted to `NaN`

The bug fires on all four backends (`open_geotiff` numpy, dask, GPU eager, HTTP COG).

## Reproducer

```python
import numpy as np, tifffile, tempfile, os
from xrspatial.geotiff import open_geotiff

stored = np.array([[0, 100, 200], [50, 0, 255]], dtype=np.uint8)
with tempfile.TemporaryDirectory() as td:
    path = os.path.join(td, 'mw_nodata.tif')
    tifffile.imwrite(
        path, stored, photometric='miniswhite',
        extratags=[('GDAL_NODATA', 's', 0, '0\0', True)],
    )
    print(open_geotiff(path).values)
```

Actual output:
```
[[255. 155.  55.]
 [205. 255.  nan]]
```

Expected: the two `0` cells (the sentinel) should be `NaN`. The cell whose stored value is `255` should remain a real datum (post-inversion `0`).

The float path has the same symptom: `_apply_photometric_miniswhite` returns `-arr`, the subsequent mask then compares `-nodata` to the sentinel and flags the wrong pixels.

## Affected sites

- `xrspatial/geotiff/_reader.py:2459` -- `read_to_array` applies MinIsWhite *after* the array is decoded but *before* `open_geotiff` runs the nodata-mask block at `xrspatial/geotiff/__init__.py:899-937`.
- `xrspatial/geotiff/_reader.py:1957` -- `_read_cog_http` does the same.
- `xrspatial/geotiff/__init__.py:2476` -- the dask chunk fetcher inverts before the dask nodata block at line 2484.
- `xrspatial/geotiff/__init__.py:3331-3336` -- the GPU eager path also inverts before `_apply_nodata_mask_gpu` at line 3346.

## Root cause

[...]

PR #2166, opened 2026-05-20
geotiff: fix open_geotiff(gpu=True) raw FileNotFoundError on HTTP/fsspec URLs (#2161)

### LINKED ISSUE #2161: open_geotiff(gpu=True) raises raw FileNotFoundError on HTTP/fsspec URLs
## Summary

`open_geotiff(source, gpu=True)` and `read_geotiff_gpu(source)` raise a raw `FileNotFoundError` when `source` is an HTTP, HTTPS, or fsspec URI. The CPU eager path and the chunked GPU path both accept these sources; only the eager GPU path doesn't.

## Where

`xrspatial/geotiff/_backends/gpu.py`, around line 293 in the eager GPU path:

```python
source = _coerce_path(source)
...
# Parse metadata on CPU (fast, <1ms)
src = _FileSource(source)
```

`_FileSource.__init__` calls `_mmap_cache.acquire(path)`, which does `open(real, 'rb')`. On a URL string, that raises `FileNotFoundError`.

## Why this is wrong

The `open_geotiff` source contract accepts string paths, HTTP/HTTPS URLs, fsspec URIs (`s3://`, `gs://`, etc.), and binary file-likes. `_open_source` in `xrspatial/geotiff/_reader.py` already has the right dispatch:

```python
if source.startswith(('http://', 'https://')):
    return _HTTPSource(source)
if _is_fsspec_uri(source):
    return _CloudSource(source)
return _FileSource(source)
```

The chunked GPU path already follows this. The docstring on `_read_geotiff_gpu_chunked` (around line 1040) says HTTP/fsspec falls back through the CPU dask path:

> **CPU decode + GPU upload** for everything else (HTTP / fsspec, no KvikIO, planar=2, sparse, MinIsWhite, non-trivial orientation, stripped layouts).

The implementation (around line 1106) only takes the GDS fast path for local strings; otherwise it routes to `read_geotiff_dask` and uploads each block via `map_blocks(cupy.asarray)`. That works for URL sources.

The eager path skips that dispatch. It assumes the source is a local file and lets `_FileSource` blow up.

## Reproducer

```python
import xrspatial.geotiff as g
g.open_geotiff("https://example.com/foo.tif", gpu=True)
# FileNotFoundError: [Errno 2] No such file or directory:
#   'https://example.com/foo.tif'
```

The CPU eager path on the same URL works (via `_open_source`).
The GPU chunked path on the same URL works (via the dask fallback).

## Expected behaviour

One of:

1. The eager GPU path opens the source via `_open_source` (or equivalent), reads the bytes into a CPU buffer, then runs the existing GPU decode pipeline over those bytes. Same behaviour as the chunked path.
2. Or, the eager GPU path rejects HTTP/fsspec sources up front with a clear `ValueError` that points the user at `chunks=...` (the dask+GPU fallback) or `gpu=False`.

Either is fine. A raw `FileNotFoundError` whose message is just the URL is not.

## Scope

This ticket covers the source-dispatch in the eager GPU path and the matching tests. Three sibling issues exist and are tracked separately: `mask_nodata=False` not honored on the GPU path, `masked_nodata` attrs lying about source-dtype on `mask_nodata=False`, and `overview_level` validation drift across direct backends. Don't bundle them.

## Acceptance

- Eager GPU path on a local path still returns a CuPy-backed `DataArray`.
- Eager GPU path on an `http://` / `https://` URL behaves per (1) or (2) above.
- Same coverage for at least one fsspec scheme (mocked; `s3://` via the existing test mocks).
- No more raw `FileNotFoundError` for URL strings.

brendancol:
## Summary

Closes #2161.

`open_geotiff(url, gpu=True)` and `read_geotiff_gpu(url)` raised a raw `FileNotFoundError` for any `http://`, `https://`, or fsspec URL. The eager GPU path called `_FileSource(source)` unconditionally and that constructor goes through the local-file mmap cache.

The `open_geotiff` source contract accepts URL and fsspec strings. The chunked GPU path already documents and implements a fallback for them: CPU decode through `read_geotiff_dask`, then `map_blocks(cupy.asarray)` to upload each block. The eager path skipped that dispatch.

## What changed

Added `_read_geotiff_gpu_eager_via_cpu`, a private helper that runs `_read_to_array` (which speaks the full source contract) and pushes the result to the device via `cupy.asarray`. URL and fsspec sources route to it at the top of `read_geotiff_gpu`. The local-file disk path is unchanged.

I picked CPU decode + GPU upload over a `ValueError` because the chunked path already does this for the same sources, so the two paths now agree. Users on URL/fsspec sources can flip `gpu=True` and get a CuPy-backed result without refactoring to `chunks=...`. Peak GPU memory is the same as the local eager path either way (the whole image lands on the device in one upload); callers who want bounded GPU memory should already be using `chunks=...`.

## Backend coverage

- numpy: unchanged
- cupy: covers HTTP/fsspec sources now, local path unchanged
- dask+numpy: unchanged
- dask+cupy: unchanged (already worked for URL/fsspec via the chunked path)

## Test plan

- [x] Local path still returns CuPy (no regression).
- [x] `http://` URL via `read_geotiff_gpu` and `open_geotiff(gpu=True)` returns CuPy with values matching the CPU eager path.
- [x] `memory://` fsspec URI does the same.
- [x] Unreachable URL no longer surfaces as a `FileNotFoundError` whose message is just the URL.
- [x] `chunks=` on a URL still uses the chunked dask+CuPy helper.
- [x] All 742 existing GPU tests still pass.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -573,9 +573,11 @@ def open_geotiff(source: str | BinaryIO, *,
         if gpu:
             raise ValueError(
                 "max_cloud_bytes is not supported when gpu=True. "
-                "The GPU reader does not apply the cloud-byte budget; "
-                "drop the kwarg, or pass gpu=False to use the eager "
-                "CPU path.")
+                "The GPU reader does not accept the kwarg directly; "
+                "for URL/fsspec sources the budget is honoured via "
+                "the XRSPATIAL_GEOTIFF_MAX_CLOUD_BYTES env var "
+                "(see issue #2161). Drop the kwarg, set the env var, "
+                "or pass gpu=False to use the eager CPU path.")
         if chunks is not None:
             raise ValueError(
                 "max_cloud_bytes is not supported when chunks=... (dask). "
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -115,7 +115,11 @@ def read_geotiff_gpu(source: str, *,
     Parameters
     ----------
     source : str
-        File path.
+        Local file path, ``http://`` / ``https://`` URL, or fsspec URI
+        (``s3://``, ``gs://``, ``memory://``, ...). URL and fsspec sources
+        use a CPU decode + GPU upload internally (matching the chunked
+        path's HTTP/fsspec fallback); the result is still a CuPy-backed
+        DataArray.
     dtype : str, numpy.dtype, or None
         Cast the result to this dtype after reading. None keeps the
         file's native dtype. Float-to-int casts raise ValueError, mirroring
@@ -261,7 +265,7 @@ def read_geotiff_gpu(source: str, *,
 
     from .._reader import (
         _FileSource, _check_dimensions, MAX_PIXELS_DEFAULT, _coerce_path,
-        _max_tile_bytes_from_env, _resolve_masked_fill,
+        _is_fsspec_uri, _max_tile_bytes_from_env, _resolve_masked_fill,
     )
     from .._compression import COMPRESSION_LERC
     from .._header import (
@@ -288,6 +292,27 @@ def read_geotiff_gpu(source: str, *,
         if w_r0 >= w_r1 or w_c0 >= w_c1 or w_r0 < 0 or w_c0 < 0:
             raise ValueError(
                 f"window={window} has non-positive size or negative origin.")
+
+    # HTTP / HTTPS / fsspec sources can't drive the disk-based GPU decode
+    # pipeline: every internal step (``_FileSource``, KvikIO GDS,
+    # ``gpu_decode_tiles_from_file``) opens the path as a local file and
+    # would raise a raw ``FileNotFoundError`` on a URL string (issue #2161).
+    # Route them through the CPU decode + GPU upload fallback, matching
+    # what ``_read_geotiff_gpu_chunked`` already does for the same sources
+    # (around line 1106-1150 in this file). The peak GPU memory is the
+    # whole image either way for the eager path; the trade-off is a CPU
+    # decode instead of nvCOMP-on-GPU. Callers who want bounded GPU
+    # memory should pass ``chunks=...``.
+    if isinstance(source, str) and (
+            source.startswith(('http://', 'https://'))
+            or _is_fsspec_uri(source)):
+        return _read_geotiff_gpu_eager_via_cpu(
+            source, dtype=dtype, window=window,
+            overview_level=overview_level, band=band, name=name,
+            max_pixels=max_pixels, allow_rotated=allow_rotated,
+            allow_unparseable_crs=allow_unparseable_crs,
+            mask_nodata=mask_nodata,
+        )
 
     # Parse metadata on CPU (fast, <1ms)
     src = _FileSource(source)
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: geotiff: fix open_geotiff(gpu=True) raw FileNotFoundError on HTTP/fsspec URLs (#2161)

None.

- [ ] **Misleading comment about `_read_to_array` masking nodata** -- `xrspatial/geotiff/_backends/gpu.py`, inside `_read_geotiff_gpu_eager_via_cpu`. The comment says "_read_to_array already applied nodata masking when its CPU mask_nodata default fires" but `_read_to_array` does not apply a nodata-to-NaN mask. It returns the raw array and stashes the (possibly inverted) sentinel on `geo_info._mask_nodata`. The mask-on-device call right after is correct (mirrors the stripped CPU-fallback branch), so the code is fine; only the comment is wrong. Rewrite it to say `_read_to_array` does NOT mask, and the GPU mask call is the canonical place.

- [ ] **`on_gpu_failure='strict'` is silently no-op on URL paths** -- `xrspatial/geotiff/_backends/gpu.py`, the new URL branch. The dispatch happens regardless of `on_gpu_failure` / `gpu`. That's defensible (there's no GPU decode stage to be strict about), but a user passing `gpu='strict'` on a URL won't get either an error or a warning when the path drops to CPU. Document it in the helper's docstring, or emit a `RuntimeWarning` if `on_gpu_failure='strict'` was explicit and the URL branch took over.

- [ ] **`max_cloud_bytes` reachability** -- `xrspatial/geotiff/__init__.py:573-578`. The rejection message says "The GPU reader does not apply the cloud-byte budget", but the new URL path runs through `_read_to_array`, which honours `XRSPATIAL_GEOTIFF_MAX_CLOUD_BYTES`. The env var works; the per-call kwarg is still hard-rejected. Either thread `max_cloud_bytes` through `read_geotiff_gpu` for URL sources only, or update the rejection message to say the env var is the supported channel for the GPU URL path.

- [ ] **`name` derivation when URL has a query string** -- `os.path.splitext(os.path.basename(source))` handles `?token=...` URLs in a way that's correct but non-obvious (`splitext('x.tif?token=abc')` returns `('x', '.tif?token=abc')`, so name becomes `'x'`). One-line comment is worth it.

- [ ] **Test 4 accepts any non-FileNotFoundError exception** -- `test_unreachable_http_url_does_not_raise_filenotfound`. The assertion just checks the error isn't a `FileNotFoundError` ending with the URL. Tightening it to require a networking exception (`urllib3.exceptions.MaxRetryError`, `ConnectionError`, `OSError`) would catch a future regression where the URL branch raises a `ValueError` and still passes the test.

- [ ] **Helper docstring could name the `open_geotiff` route** -- the helper docstring talks about `read_geotiff_gpu` but most users reach it through `open_geotiff(..., gpu=True)`. One sentence helps grep.

### What looks good

- The fix mirrors the chunked path's documented HTTP/fsspec fallback. The two paths now agree on URL behaviour.
- The eager-via-CPU helper builds the same `attrs` / `coords` / `dims` as the existing stripped-CPU-fallback branch in the local path. Nodata mask, MinIsWhite sentinel inversion via `geo_info._mask_nodata`, dtype cast, pixels-present attr all line up.
- Tests cover the four acceptance criteria from the issue (local regression, HTTP parity with CPU, fsspec parity with CPU, no raw FileNotFoundError) plus a chunked-URL smoke test that confirms the eager URL branch doesn't intercept the dask path.
- The local disk path is untouched.
- A live smoke test during review confirmed CPU and GPU URL paths produce identical NaN values and matching `attrs['nodata']` / `attrs['masked_nodata']` / `attrs['nodata_pixels_present']`.

### Checklist

- [x] Algorithm matches reference/paper -- n/a (I/O fix)
- [x] All implemented backends produce consistent results -- CPU and GPU URL paths agree
- [x] NaN handling is correct -- nodata-to-NaN mask runs on the device after upload
- [x] Edge cases are covered by tests -- local, HTTP, fsspec, unreachable URL, chunked URL
- [ ] Dask chunk boundaries handled correctly -- n/a (eager path)
- [x] No premature materialization or unnecessary copies -- one CPU-side `read_to_array`, one `cupy.asarray` upload, then on-device ops
- [ ] Benchmark exists or is not needed -- n/a
- [ ] README feature matrix updated -- n/a (no new public API)
- [x] Docstrings present and accurate -- `source` doc updated, helper docstring is thorough (minor comment fix noted above)

brendancol (commented):
## Follow-up review after commit f7eccaf

All three suggestions and all three nits from the previous review are addressed.

### Disposition

- **Fixed: Misleading `_read_to_array` nodata comment** — rewritten in `_read_geotiff_gpu_eager_via_cpu` to say `_read_to_array` does NOT mask; the GPU mask call is the canonical site. Mentions issue #1809 for the MinIsWhite precedence rule.
- **Fixed: `on_gpu_failure='strict'` no-op on URL paths** — added a paragraph to the helper's docstring explaining there is no GPU decode stage to be strict about; the only GPU op is the `cupy.asarray` upload and its failures propagate unconditionally. Pointed users at a local-file repro for GPU pipeline regression checks.
- **Fixed: `max_cloud_bytes` rejection message** — updated to mention that `XRSPATIAL_GEOTIFF_MAX_CLOUD_BYTES` still applies on the URL/fsspec GPU path via `_read_to_array`. The per-call kwarg stays rejected (env-var-only is the smaller fix).
- **Fixed: URL query-string `name` derivation comment** — added inline comment in the name-derivation block explaining the `?token=abc` case.
- **Fixed: Test 4 tightened** — `test_unreachable_http_url_does_not_raise_filenotfound` now also asserts the error is one of `(MaxRetryError, HTTPError, ConnectionError, OSError)`. A future regression where the URL branch raises a `ValueError` from earlier validation would now fail the test.
- **Fixed: helper docstring mentions `open_geotiff`** — first line now names `open_geotiff(url, gpu=True)` and `read_geotiff_gpu(url)` as the two public entry points.

### Verification

- All 5 new URL-eager tests still pass.
- All 14 cloud-byte-limit tests still pass (no test pinned the exact rejection message text).

No remaining findings.

PR #2176, opened 2026-05-20
rasterize: reject non-uniformly spaced like grids (#2168)

### LINKED ISSUE #2168: rasterize(like=...) reuses irregular coords with uniform-grid rasterizer
**Describe the bug**

`_extract_grid_from_like()` in `xrspatial/rasterize.py` (around line 1948) takes `px`/`py` from the first coordinate interval of `like` and then treats the grid as regular. Around line 2283, the output DataArray reuses `like.coords` verbatim. The rasterizer only works on a uniform grid.

If `like` has non-uniform x or y spacing, `rasterize(..., like=...)` returns a DataArray whose coords (copied from the irregular `like`) no longer line up with the uniform grid the pixels were written to. The output silently mislabels where each cell lives.

**Reproduce**

```python
import numpy as np
import xarray as xr

x = np.array([0.0, 1.0, 2.5, 3.5])  # non-uniform
y = np.array([3.0, 2.0, 1.0, 0.0])
like = xr.DataArray(np.zeros((4, 4)), dims=('y','x'), coords={'y': y, 'x': x})

out = rasterize([(box(0, 0, 3.5, 3.0), 1.0)], like=like)
# out.x is [0, 1, 2.5, 3.5] but the data sits on a uniform grid.
```

**Expected behavior**

Two options:

(a) Detect irregular grids in `_extract_grid_from_like()` and raise `ValueError`, naming the offending axis and the largest deviation. This is what rasterio does and is the safer default.

(b) Regularize the like grid before rasterizing and document the coercion.

Prefer (a).

**Implementation notes**

- Check the full coord vector is uniformly spaced within a numeric tolerance. Use `numpy.allclose` on the diffs, not strict equality. Affine-derived coords drift.
- Name the axis (x or y) and report the largest deviation in the error message.
- A sibling issue covers ascending-y `like`. Order the uniform-spacing check so it does not interfere with the ascending-y work in that other PR.
- The helper sits on the numpy path, but dask, cupy, and dask+cupy all call into it, so one fix covers every backend.

**Tests**

- Non-uniform x: `ValueError` raised, axis named.
- Non-uniform y: same.
- Uniformly spaced like: existing happy path unchanged.

brendancol:
Closes #2168.

## Summary

- `_extract_grid_from_like()` now validates that the `like` DataArray's `x` and `y` coords are uniformly spaced before the rasterizer commits to a single `px`/`py`.
- If either axis is non-uniform (beyond `np.allclose`'s default tolerance), `rasterize(..., like=...)` raises `ValueError` naming the axis and the largest deviation.
- The check works on `abs(diff)` so it stays direction-agnostic, leaving room for the sibling ascending-y support without conflict.
- Docstring updated to call out the uniform-grid requirement on `like=`.

## Backend coverage

The validation runs in the shared `_extract_grid_from_like()` helper. Every entry point (numpy, cupy, dask+numpy, dask+cupy) calls into it, so all four backends get the same rejection behaviour.

## Test plan

- [x] New tests in `xrspatial/tests/test_rasterize.py::TestLikeUniformGridValidation` cover non-uniform x, non-uniform y, the uniform happy path, error message contents, ulp-level float drift tolerance, and a 1-cell-wide degenerate axis.
- [x] Full rasterize test suite (`test_rasterize*.py`) is green: 281 passed, 2 skipped.

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -1924,6 +1924,42 @@ def _parse_input(geometries, column=None, columns=None):
     return geom_list, props_array, None
 
 
+def _check_uniform_axis(axis_name, coords, expected_step):
+    """Raise ``ValueError`` if ``coords`` is not uniformly spaced.
+
+    ``coords`` is a 1-D float64 array of dim-coord values for the named
+    axis (``'x'`` or ``'y'``).  ``expected_step`` is the magnitude of the
+    first interval (already computed by the caller).  Axes with fewer
+    than three points cannot be non-uniform in a way this check would
+    catch, so they pass trivially.
+
+    The comparison is on ``abs(diff)`` so the validation does not care
+    whether the axis is ascending or descending -- a sibling change is
+    expected to allow ascending-y ``like`` inputs, and gating on the
+    sign here would block that work.  ``np.allclose`` is used (rather
+    than strict equality) because affine-transform-derived coords drift
+    by a few ulps in practice.
+    """
+    if coords.size < 3:
+        return
+
+    diffs = np.abs(np.diff(coords))
+    if expected_step == 0 or not np.isfinite(expected_step):
+        # Degenerate step: defer to downstream code so the error path
+        # here doesn't mask a different underlying problem.
+        return
+
+    if not np.allclose(diffs, expected_step, rtol=1e-5, atol=1e-8):
+        max_dev = float(np.max(np.abs(diffs - expected_step)))
+        raise ValueError(
+            "'like' DataArray has non-uniform spacing along the "
+            f"{axis_name!r} axis (expected step {expected_step}, "
+            f"largest deviation {max_dev}). rasterize() requires a "
+            "regular grid; resample 'like' to a uniform grid before "
+            "passing it."
+        )
+
+
 def _extract_grid_from_like(like):
     """Extract width, height, bounds, dtype from a template DataArray.
 
@@ -1953,6 +1989,21 @@ def _extract_grid_from_like(like):
         py = abs(float(y[0] - y[1]))
     else:
         py = 1.0
+
+    # The rasterizer assumes a uniform grid.  If ``like`` has non-uniform
+    # spacing on either axis, ``px``/``py`` (taken from the first interval)
+    # will not describe the rest of the grid, and reusing ``like.coords``
+    # on the output (see ``rasterize`` below) would mislabel where each
+    # pixel lives.  Validate uniform spacing here so the rasterizer never
+    # produces a DataArray whose coords disagree with its data layout.
+    #
+    # Compare ``abs(diff)`` against the first interval so the check stays
+    # agnostic to axis direction -- ascending or descending y both pass as
+    # long as the spacing is uniform.  Use ``np.allclose`` rather than
+    # strict equality because affine-transform-derived coords drift by a
+    # few ulps.
+    _check_uniform_axis('x', x, px)
+    _check_uniform_axis('y', y, py)
 
     xmin = float(np.min(x)) - px / 2
     xmax = float(np.max(x)) + px / 2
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: rasterize: reject non-uniformly spaced like grids (#2168)

### Blockers
None.

### Suggestions
- `xrspatial/rasterize.py:1947` -- the early-return on `expected_step == 0` lets a degenerate `like` (all-equal x or y coords) slip through silently. That isn't a regression from `main` (the old code also accepted it and then divided things by `px==0` later), but since this PR is the new gatekeeper on `like` regularity it would be a reasonable place to reject zero-step grids too. If you'd rather keep the diff tight, leaving a `# TODO` or filing a follow-up is fine; I won't block on it.
- `xrspatial/rasterize.py:1952` -- `np.allclose` defaults (`rtol=1e-5`, `atol=1e-8`) are reasonable for projected coords (meters) but they're loose for sub-degree geographic grids where the step is `1e-4` and `atol` is `0.01%` of the step. A real non-uniformity slightly above ulp drift could pass on a fine lat/lon grid. Probably fine in practice -- just calling it out.

### Nits
- `xrspatial/rasterize.py:1959` -- "resample 'like' to a uniform grid before passing it" is good user-facing advice. Could be even more useful if it pointed at `xarray`'s `interp` / `reindex` or the typical regularize idiom, but the message is already actionable.
- `xrspatial/tests/test_rasterize.py:2120` -- the `test_error_message_names_axis_and_deviation` test asserts the string `"0.5"` appears in the message. That couples the test to numeric formatting. If you ever wrap `max_dev` with `f"{max_dev:.3g}"`, the assertion would still pass, but `f"{max_dev:.2e}"` would break it. Low risk because the format is currently `f"{max_dev}"` (default repr), but worth knowing.

### What looks good
- Validation lives in the single shared helper `_extract_grid_from_like`, so all four backends (numpy / cupy / dask+numpy / dask+cupy) inherit the rejection without per-backend duplication. Verified there's exactly one caller.
- `abs(diff)` keeps the check direction-agnostic. The comment explicitly notes the sibling ascending-y work and gates accordingly, which is the right call.
- The `coords.size < 3` short-circuit handles size==1 (the `px = 1.0` fallback path) and size==2 (where `np.diff` is trivially "uniform") without divide-by-zero or false positives.
- Tests cover the four corners I'd want: non-uniform x, non-uniform y, ulp-level drift tolerated, single-row/column degenerate axes, and an explicit assertion on the error message contents. The existing 281 rasterize tests still pass.
- Docstring update on `like=` calls out the requirement before the user trips over it.

### Checklist
- [x] Algorithm matches the validation pattern rasterio uses
- [x] All implemented backends produce consistent results (single helper)
- [x] NaN handling -- NaN coords fail closed via allclose returning False
- [x] Edge cases covered (size<3, ulp drift, degenerate width=1)
- [x] Dask chunk boundaries -- N/A, this is a pre-dispatch input guard
- [ ] Benchmark exists or is not needed -- not needed, validation is O(n) on the coord vector
- [ ] README feature matrix updated -- N/A per rockout instructions

brendancol (commented):
## Follow-up review after cfc0f72

Quick second pass after the review fixes.

- Suggestion (zero-step grid silently accepted) -- **fixed** in cfc0f72. `_check_uniform_axis` now raises a dedicated ValueError for all-equal coord vectors. Covered by `test_zero_step_like_raises`.
- Suggestion (loose tolerance on sub-degree geographic grids) -- **dismissed**. Tightening `rtol`/`atol` would either break the legitimate ulp-drift case the original commit was explicitly designed to tolerate (the `test_tiny_float_drift_is_tolerated` case), or have no effect at all. Default `np.allclose` defaults are what rasterio-style tools use too; the comment in the source documents the tradeoff.
- Nit (error message could mention `interp`/`reindex`) -- **fixed** in cfc0f72. Both error branches now name those remedies.
- Nit (test couples to "0.5" string) -- **fixed** in cfc0f72. The assertion now parses the deviation out with a regex and compares numerically.

### New review of follow-up commit
No new blockers, suggestions, or nits. 282 rasterize tests still green.

### What looks good
- The new zero-step branch raises before any downstream code can divide by `px==0`. Net safety win.
- The numeric-parse in the test is robust to any reasonable float format.

PR #2181, opened 2026-05-20
rasterize: honor like.y orientation for ascending-y templates (#2170)

### LINKED ISSUE #2170: rasterize(like=...) silently flips spatial meaning when y coordinates ascend
**Describe the bug**

`rasterize(like=...)` ignores the y-axis orientation of the template. The rasterizer always burns with row 0 = ymax (standard image convention), then the output reuses `like.y` verbatim. When `like.y` is ascending, the y coords no longer match where the geometry got burned, so anything that selects by world coordinate downstream (zonal stats, spatial joins, plotting) silently lines up against the wrong rows.

`_extract_grid_from_like()` at xrspatial/rasterize.py:1957 takes min/max of the coords, so the bounding box is fine. The orientation just never makes it through the pipeline, and the coord assignment at xrspatial/rasterize.py:2276 hands back the original coords without flipping the array to match.

**Reproduce**

```python
import numpy as np
import xarray as xr

x = np.array([0.5, 1.5])
y_desc = np.array([1.5, 0.5])  # standard top-down y
y_asc  = np.array([0.5, 1.5])  # ascending y (bottom-up)

like_desc = xr.DataArray(np.zeros((2, 2)), dims=['y', 'x'], coords={'y': y_desc, 'x': x})
like_asc  = xr.DataArray(np.zeros((2, 2)), dims=['y', 'x'], coords={'y': y_asc,  'x': x})

geom = [(box(0, 0, 1, 1), 1.0)]  # world y in [0, 1] -> lower row

print(rasterize(geom, like=like_desc, fill=0).sel(y=0.5, x=0.5).item())  # 1.0  ok
print(rasterize(geom, like=like_asc,  fill=0).sel(y=0.5, x=0.5).item())  # 0.0  wrong
```

**Expected behavior**

`result.sel(y=0.5)` should return the burned value whether `like.y` is ascending or descending. `output.y` should still equal `like.y` so `xr.align` keeps working.

**Scope**

Ascending-y only. There's a separate sibling issue for irregular `like` grids (non-uniform spacing) that should stay out of this fix.

**Backends**

All four (numpy, cupy, dask+numpy, dask+cupy) hit the same coord assignment path, so all four are affected.

brendancol:
## Summary

- Detect `like.y` orientation once in `_extract_grid_from_like()`.
- Carry the flag through to the coord-assignment site and flip the burned array on axis 0 right before reusing the template's coords. After the fix, `result.sel(y=...)` lines up with the geometry whether `like.y` ascends or descends.
- `result.y` still equals `like.y` byte-for-byte, so `xr.align` keeps working.

Closes #2170

## Backend coverage

The flip is just `out[::-1, :]`, which behaves the same on numpy, cupy, dask+numpy, and dask+cupy arrays.

## Out of scope

There's a sibling issue for irregular `like` grids (non-uniform spacing). Not touched here.

## Test plan

- [x] New `TestLikeYOrientation2170` in `xrspatial/tests/test_rasterize.py`, covering polygons, lines, and points across all four backends (GPU cases skip when CUDA isn't available).
- [x] Regression test: `result.sel(y=0.5, x=0.5)` returns the burned value for both ascending and descending `like.y`.
- [x] Round-trip: `result.y == like.y` and `xr.align(result, like)` still works.
- [x] Explicit-bounds path still rebuilds descending coords (flip is correctly bypassed for any resized output).
- [x] Full `test_rasterize.py` passes locally (199 passed, 2 skipped).

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -12,7 +12,7 @@ Supports numpy, cupy, dask+numpy, and dask+cupy backends.
 from __future__ import annotations
 
 import warnings
-from typing import Optional, Tuple, Union
+from typing import NamedTuple, Optional, Tuple, Union
 
 import numpy as np
 import shapely
@@ -1924,6 +1924,24 @@ def _parse_input(geometries, column=None, columns=None):
     return geom_list, props_array, None
 
 
+class _LikeGrid(NamedTuple):
+    """Grid attributes extracted from a template DataArray.
+
+    Returned by :func:`_extract_grid_from_like`.  Using a named tuple
+    instead of a bare tuple keeps the call site readable as more
+    template-derived attributes accrete over time.
+    """
+    width: int
+    height: int
+    bounds: Tuple[float, float, float, float]
+    dtype: np.dtype
+    x_coord: xr.DataArray
+    y_coord: xr.DataArray
+    extra_coords: dict
+    attrs: dict
+    y_ascending: bool
+
+
 def _extract_grid_from_like(like):
     """Extract width, height, bounds, dtype from a template DataArray.
 
@@ -1963,8 +1981,12 @@ def _extract_grid_from_like(like):
     # at ymax (standard image convention), so if the template's y is
     # ascending (low-to-high), the burned rows have to be flipped before
     # we hand back the coords or downstream coord-aware ops line up
-    # against the wrong rows.  Only look at the first/last sample; the
-    # irregular-spacing case is handled separately.
+    # against the wrong rows.  Assumes ``like.y`` is monotonic -- only
+    # the first and last samples are inspected.  Non-monotonic or
+    # duplicate-valued coords (e.g. stitched templates) are out of
+    # scope and are handled by the irregular-spacing path separately.
+    # The x-axis is assumed ascending; descending-x templates would hit
+    # the same bug class and are not supported here.
     y_ascending = height > 1 and float(y[-1]) > float(y[0])
 
     # Carry through any non-dim coords (e.g. rioxarray's ``spatial_ref``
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: rasterize honor like.y orientation for ascending-y templates (#2170)

### Blockers
None.

### Suggestions
- `xrspatial/rasterize.py:1968`. `y_ascending = height > 1 and float(y[-1]) > float(y[0])` only looks at the first and last samples. Fine for a monotonic 1D coord, but a non-monotonic or duplicate-valued `like.y` would be misclassified. The inline comment already defers the irregular-spacing case; a sentence noting monotonicity is assumed, or an `np.all(np.diff(y) > 0)` check, would make the contract explicit.
- `xrspatial/rasterize.py:1978`. `_extract_grid_from_like` now returns a 9-tuple and the call site unpacks positionally. A `NamedTuple` or small dataclass would scale better as future grid attributes get added. Out of scope for this fix.

### Nits
- `xrspatial/rasterize.py:2298`. Descending-x templates would hit the same bug class as descending-y did before this fix. A short comment near the y-flip noting "x is assumed ascending; descending-x templates are not supported" would document the asymmetry for the next reader.
- `xrspatial/tests/test_rasterize.py:2244`. `test_numpy_explicit_bounds_skips_flip` asserts `result.y.values[0] > result.y.values[-1]`. Tightening to exact expected values (e.g. `[3.5, 2.5, 1.5, 0.5]`) would catch a regression where the bounds path produces an off-by-one coord centre.

### What looks good
One detection site, one flip site, with comments at both explaining the image-vs-world convention. `out[::-1, :]` is backend-agnostic by construction. Tests hit all four backends and three geometry types, plus the `xr.align` round-trip and the explicit-bounds bypass. World-y selection asserts (`result.sel(y=0.5, x=0.5)`) target the bug directly rather than checking array indices. `result.y` is preserved byte-for-byte against `like.y`, which is what `xr.align` depends on. The `height == 1` edge case falls out for free: the `height > 1` guard makes `y_ascending=False` so the flip is a no-op.

brendancol (commented):
## Follow-up review on 54fe792

All four findings from the first pass are addressed in the new commit.

### Disposition
- **Suggestion 1 (monotonicity contract).** Fixed. The comment on `y_ascending` now states that monotonicity is assumed and that the irregular-spacing case is out of scope.
- **Suggestion 2 (9-tuple unpacking).** Fixed. `_extract_grid_from_like` returns a `_LikeGrid` NamedTuple. The call site in `rasterize` unpacks by attribute name.
- **Nit 1 (x-axis asymmetry).** Fixed. The same comment block now flags the descending-x case as out of scope so the asymmetry is explicit.
- **Nit 2 (test assertion tightness).** Fixed. `test_numpy_explicit_bounds_skips_flip` locks the exact y/x coord centres.

### Verification
- `pytest xrspatial/tests/test_rasterize.py` passes locally (199 passed, 2 skipped).

### Remaining concerns
None.

PR #2188, opened 2026-05-20
polygonize: tolerate near-equal floats in Dask chunk-stitching (#2171)

### LINKED ISSUE #2171: polygonize: Dask float results are chunk-size dependent
## Description

The numpy path groups adjacent float pixels using `_is_close` (atol=1e-8, rtol=1e-5) at `xrspatial/polygonize.py:244`. The Dask path stitches chunk-boundary polygons by keying on the raw float value at `xrspatial/polygonize.py:1601`. So the polygon count depends on how you chunk the input.

## Reproduction

```python
import numpy as np
import xarray as xr
import dask.array as da

data = np.array([[1.0, 1.000001]], dtype=np.float64)

# 1 polygon
polygonize(xr.DataArray(data))

# 1 polygon
polygonize(xr.DataArray(da.from_array(data, chunks=data.shape)))

# 2 polygons
polygonize(xr.DataArray(da.from_array(data, chunks=(1, 1))))
```

## Expected

Dask multi-chunk results should match numpy results for the same input.

## Actual

Multi-chunk Dask runs split near-equal float pixels into separate polygons. The boundary-merge dict keys on raw float values, so 1.0 and 1.000001 land in different buckets.

This also contradicts the docstring at `xrspatial/polygonize.py:1646` which says 4-connectivity works perfectly with Dask chunking.

## Fix Direction

The chunk-stitching code at line 1601 needs to bucket boundary polygons using the same `_is_close` tolerance the numpy path uses. Cross-backend tests should assert numpy and multi-chunk Dask produce the same polygons for near-equal float inputs, including the `[[1.0, 1.000001]]` case.

brendancol:
## Summary

- Bucket Dask boundary polygons using the same float tolerance the numpy path already uses, so `polygonize` results stop depending on chunking when the raster has near-equal floats.
- Mirror `_is_close` at the Python orchestration layer as `_values_close` and route the three boundary-merge dict insertions through a `_bucket_key_for_value` helper. Integer-valued rasters short-circuit to exact equality, so the int path keeps its current behaviour and cost.

Closes #2171

## Backend coverage

- numpy / cupy: unchanged.
- dask+numpy: fixed. Multi-chunk boundary merging matches the single-chunk numpy result on near-equal float inputs.
- dask+cupy: same orchestration code path as dask+numpy (per-chunk results flow through the same boundary dict), so the fix applies there too.

## Test plan

- [x] `[[1.0, 1.000001]]` returns one polygon across numpy, dask single-chunk, and dask `(1, 1)` chunked
- [x] Multi-chunk dask polygon count and per-value areas match numpy for a few near-equal float patterns at (1, 1) and (2, 2) chunks
- [x] Existing polygonize tests still pass
- [x] Integer rasters unchanged (exact-equality fast path)

### DIFF
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -247,6 +247,37 @@ def _is_close(
             abs(value - reference) <= (atol + rtol*abs(reference))
 
 
+# Pure-Python tolerance check that mirrors ``_is_close`` for use at the
+# orchestration layer (outside numba kernels).  Integer values fall back
+# to exact equality; floats use the same atol=1e-8, rtol=1e-5 tolerance
+# the CPU CCL uses to merge adjacent pixels.  See issue #2171.
+def _values_close(reference, value):
+    if isinstance(reference, (int, np.integer)) and \
+            isinstance(value, (int, np.integer)):
+        return value == reference
+    return abs(value - reference) <= (1e-8 + 1e-5 * abs(reference))
+
+
+def _bucket_key_for_value(boundary_by_value, val):
+    """Return the dict key that ``val`` should bucket into.
+
+    If an existing key in ``boundary_by_value`` is close to ``val`` under
+    ``_values_close``, return that key so close float values from adjacent
+    chunks land in the same bucket.  Otherwise return ``val`` unchanged.
+
+    This keeps Dask chunk-stitching consistent with the tolerance-based
+    grouping the NumPy / numba path uses inside a single chunk (#2171).
+    """
+    # Integer-valued rasters use exact equality, so the existing dict
+    # lookup is already correct -- skip the linear scan.
+    if isinstance(val, (int, np.integer)):
+        return val
+    for existing in boundary_by_value:
+        if _values_close(existing, val):
+            return existing
+    return val
+
+
 # Calculate region connectivity for the specified values raster and optional
 # mask raster.  Each region is labelled with a unique integer ID starting at
 # 1.  Regions corresponding to masked out pixels are all given the same region
@@ -1486,7 +1517,8 @@ def _merge_chunk_polygons(chunk_results, transform):
     for interior, boundary in chunk_results:
         all_interior.extend(interior)
         for val, rings in boundary:
-            boundary_by_value.setdefault(val, []).append(rings)
+            key = _bucket_key_for_value(boundary_by_value, val)
+            boundary_by_value.setdefault(key, []).append(rings)
 
     # Merge boundary polygons per value using edge cancellation.
     merged = []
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: polygonize: tolerate near-equal floats in Dask chunk-stitching (#2171)

None.

- [ ] The docstring at `xrspatial/polygonize.py:1681` says "4-connectivity works perfectly with Dask chunking" without mentioning float tolerance. The whole point of this PR is to make that statement true on near-equal floats, so the docstring should say that near-equal float values (within `_is_close` tolerance, atol=1e-8 rtol=1e-5) are grouped consistently across chunks. Pins down the contract for future readers.

- [ ] `_bucket_key_for_value` does a linear scan of `boundary_by_value` on every float insertion, so float-path cost is O(B^2) per chunk where B is the unique-value count. Typical polygonize inputs have few unique values, so this is fine. A short comment noting the assumption (small B per chunk) would help future readers.
- [ ] Naming drift between the numba helper `_is_close` and the new Python helper `_values_close`. Same predicate, different verb. The new helper's comment already references `_is_close`, so this is minor.

### What looks good
- Fix is local: a small Python helper plus three call-site swaps. No changes to the numba kernels or per-chunk algorithm.
- Tests directly reproduce the issue's `[[1.0, 1.000001]]` case and add parametrized chunked-vs-numpy parity checks at (1, 1) and (2, 2) chunks.
- Integer short-circuit keeps the existing O(1) dict lookup for int rasters.
- The tolerance constants in `_values_close` are pulled from the same numbers `_is_close` uses, so the two predicates cannot drift.

### Checklist
- [x] Algorithm matches reference (`_is_close` tolerance, atol=1e-8 rtol=1e-5)
- [x] All implemented backends produce consistent results (numpy / dask single-chunk / dask multi-chunk parity tested)
- [x] NaN handling preserved (NaN remains non-close to anything, same as before)
- [x] Edge cases covered (1x2, 2x2, 3x3, 4x4 with (1, 1) and (2, 2) chunks)
- [n/a] Benchmark coverage (no perf-sensitive change on int path; float path is O(B) per insert which is fine for typical B)
- [n/a] README feature matrix updated (no new function)
- [x] Docstrings present (helper has one; public API docstring unchanged)

brendancol (commented):
## Follow-up review

- Suggestion (docstring at `polygonize.py:1681` should mention float tolerance): **deferred**. A parallel rockout is already on the float-semantics docstring and will likely touch this exact paragraph; editing it here would conflict with that work.
- Nit (O(B^2) scan in `_bucket_key_for_value`): **fixed** in c9fe67e -- added a performance note in the helper's docstring describing the assumption and a path forward if B ever grows large.
- Nit (naming drift between `_is_close` and `_values_close`): **dismissed**. The new helper's existing comment already cross-references `_is_close` as the predicate it mirrors. Renaming the existing numba helper is out of scope for a bug fix.

### Post follow-up scan
No new findings. Tests still pass (5/5 on the new #2171 tests; the broader `test_polygonize.py` and `test_polygonize_coverage_2026_05_19.py` set is also green except for a pre-existing cupy +Inf failure that exists on `main` and is being addressed by the parallel inf-handling rockout).

PR #2189, opened 2026-05-20
reproject: handle (band, y, x) 3-D inputs (#2182)

### LINKED ISSUE #2182: reproject: 3-D rasters with (band, y, x) dims crash and produce wrong output
**Describe the bug**

`reproject()` accepts 3-D inputs but the per-chunk worker assumes the spatial dims are the first two axes and the band dim is trailing. The common rasterio/rioxarray convention is `(band, y, x)`, and that layout breaks `reproject()`:

- `xrspatial/reproject/__init__.py:290`: the worker does `source_data[r_min_clip:r_max_clip, c_min_clip:c_max_clip]`, which slices band and y when the band dim is leading.
- `xrspatial/reproject/__init__.py:300`: the 3-D branch reads `window.shape[2]` and indexes `window[:, :, b]`, hard-coded for a trailing band axis.
- `xrspatial/reproject/__init__.py:823`: the output is built with `[ydim, xdim, band_dim]` regardless of input layout.

A `(band, y, x)` input either crashes with a band coord-length mismatch or returns wrong data because the wrong axes are sliced.

**Expected behavior**

`reproject()` should detect the band axis position from `raster.dims` and handle both `(band, y, x)` and `(y, x, band)` layouts. The output DataArray should preserve the input's dim order so downstream code (rioxarray, rasterio write paths) keeps working.

**Reproducer**

```python
import numpy as np
import xarray as xr

data = np.random.rand(3, 32, 32).astype('float32')
raster = xr.DataArray(
    data,
    dims=['band', 'y', 'x'],
    coords={
        'band': [1, 2, 3],
        'y': np.linspace(50, 40, 32),
        'x': np.linspace(-5, 5, 32),
    },
    attrs={'crs': 'EPSG:4326'},
)
out = reproject(raster, 'EPSG:3857')  # crashes
```

**Additional context**

Tests should cover both `(band, y, x)` and `(y, x, band)` 3-D inputs on the numpy, dask+numpy, cupy, and dask+cupy backends (cupy paths gated on GPU availability).

brendancol:
Closes #2182

## Summary

- The per-chunk worker assumed the band axis was trailing: it sliced `source_data[r:, c:]` and read `window.shape[2]` for the band count. A rasterio/rioxarray-style `(band, y, x)` input sliced the band/y axes instead of y/x and either crashed with a band coord-length mismatch or returned wrong-shape data.
- The entry point now transposes 3-D inputs to the canonical `(y, x, band)` layout before dispatch and transposes the output back to the input's dim order, so a `(band, y, x)` raster round-trips as `(band, y, x)`.

## Backends covered

numpy, dask+numpy, cupy, dask+cupy. New tests run on all four (cupy/dask+cupy paths gated on GPU availability via `skipif`).

## Test plan

- [x] `pytest xrspatial/tests/test_reproject.py::TestReproject3DBandFirst` -- 9 passed
- [x] `pytest xrspatial/tests/test_reproject.py` -- 277 passed, no regressions
- [x] Pixel-equality check that `(band, y, x)` and `(y, x, band)` inputs produce the same data after a common transpose

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -652,6 +652,23 @@ def reproject(
                 f"{sorted(_VERTICAL_DATUM_EPSG)} or None."
             )
 
+    # Normalize 3-D inputs to canonical (y, x, band) layout.
+    # The per-chunk workers slice the source as ``source_data[r:, c:]`` and
+    # assume the band axis is trailing. A rasterio/rioxarray-style
+    # ``(band, y, x)`` input would otherwise slice the band/y axes instead
+    # of the y/x axes and either crash or return wrong-shape data (#2182).
+    # We record the input's original dim order so the output can be
+    # transposed back at the end, preserving downstream expectations.
+    _input_dims = tuple(raster.dims)
+    if raster.ndim == 3:
+        _ydim_in, _xdim_in = _find_spatial_dims(raster)
+        _band_dims_in = [d for d in _input_dims
+                         if d not in (_ydim_in, _xdim_in)]
+        _band_dim_in = _band_dims_in[0] if _band_dims_in else None
+        _canonical = (_ydim_in, _xdim_in, _band_dim_in)
+        if _band_dim_in is not None and _input_dims != _canonical:
+            raster = raster.transpose(*_canonical)
+
     # Resolve CRS
     src_crs = _resolve_crs(source_crs)
     if src_crs is None:
@@ -858,6 +875,14 @@ def reproject(
         name=name or raster.name,
         attrs=out_attrs,
     )
+
+    # Preserve the input's dim order so a ``(band, y, x)`` source produces a
+    # ``(band, y, x)`` output (#2182). The internal pipeline always builds the
+    # array as ``(y, x, band)`` for 3-D rasters; transpose back here.
+    if result.ndim == 3 and set(_input_dims) == set(result.dims):
+        if tuple(result.dims) != _input_dims:
+            result = result.transpose(*_input_dims)
+
     return result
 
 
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: reproject handles (band, y, x) 3-D inputs (#2182)

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/reproject/__init__.py:670`: the `_canonical` tuple is built with `_band_dim_in` even when it is `None`. The guard `_band_dim_in is not None and _input_dims != _canonical` saves the call, but a tuple containing `None` is a little odd to read. Building the tuple inside the guard would make the intent clearer.
- `xrspatial/tests/test_reproject.py:3996`: `test_band_first_dask_matches_numpy` calls `_make_band_first_raster()` three times. Caching the raster in a local would save two RNG passes (cosmetic; the test is fast either way).

### What looks good
- The fix is localized to the entry point: transpose to canonical (y, x, band) before dispatch, transpose back after the output DataArray is built. Downstream workers and the per-block dask adapter are unchanged, which keeps the blast radius small.
- The set-equality check on the final transpose (`set(_input_dims) == set(result.dims)`) is a defensive guard against crashing if the band-dim name ever differs between input and output.
- Tests cover all four backends (numpy, dask, cupy, dask+cupy) with both layout-preservation and pixel-equality assertions.
- `xr.DataArray.transpose` is metadata-only on dask arrays, so no extra materialization is introduced for the band-first dask path.

### Checklist
- [x] Algorithm matches reference/paper -- N/A (pure layout fix)
- [x] All backends produce consistent results -- verified by test_band_first_matches_band_last and test_band_first_dask_matches_numpy
- [x] NaN handling is correct -- unchanged from the (y, x, band) path
- [x] Edge cases covered -- integer dtype roundtrip included
- [x] Dask chunk boundaries handled -- lazy transpose preserves chunking semantics
- [x] No premature materialization or unnecessary copies -- transpose is metadata-only
- [x] Benchmark not needed -- no perf-sensitive code touched
- [x] README feature matrix -- N/A (no new API)
- [x] Docstrings -- existing docstring still accurate

PR #2190, opened 2026-05-20
reproject.merge: honor y orientation in same-CRS fast path (#2186)

### LINKED ISSUE #2186: reproject.merge: same-CRS fast path ignores y orientation, flips y-ascending inputs
**Describe the bug**

`reproject.merge()` and `reproject()` disagree for a single y-ascending raster when the source CRS equals the target CRS.

`_place_same_crs()` in `xrspatial/reproject/__init__.py` receives a `y_desc` flag but never uses it. The function copies the source pixel window directly into a north-up output grid. When the input is y-ascending (`y_desc=False`), row 0 of the source corresponds to the bottom of the bounding box, but row 0 of the output corresponds to the top. The result is silently flipped along y.

`reproject()` handles `y_desc` correctly, so `merge([r], target_crs=r.crs)` and `reproject(r, target_crs=r.crs)` produce different outputs for the same y-ascending raster. That is wrong; the same-CRS fast path should be a no-op equivalent to a same-CRS reproject.

Affected code:
- `xrspatial/reproject/__init__.py:1910` -- `_place_same_crs()` accepts `y_desc` and discards it.
- `xrspatial/reproject/__init__.py:1965` -- direct window copy with no orientation handling.

Both the in-memory path (`_merge_inmemory`) and the dask path (`_merge_block_adapter`) call `_place_same_crs()`, so both backends are affected.

**Expected behavior**

For any single raster `r`, `merge([r], target_crs=r.crs)` should match `reproject(r, target_crs=r.crs)` cell-by-cell. The output orientation of `merge()` is documented as always north-up; the same-CRS fast path should flip y-ascending sources to match.

**Additional context**

There is a separate sibling issue for x-descending inputs in the main reproject path. Keep this fix focused on the merge same-CRS path. Coverage should include:

- y-ascending single-raster merge vs reproject parity.
- y-descending single-raster merge (regression guard).
- Multi-raster merge with mixed y orientations.
- Backends: numpy, dask+numpy. (cupy / dask+cupy share the same Python adapter.)

brendancol:
Closes #2186.

## Summary
- `_place_same_crs()` accepted a `y_desc` flag but discarded it, so y-ascending inputs were copied directly into the always-north-up merge output without flipping. `merge([r], target_crs=r.crs)` therefore disagreed with `reproject(r, target_crs=r.crs)` for the same raster.
- Read the source window in the correct direction when `y_desc` is `False` and place it so the result is north-up. The branch handles tiles that extend past the output bounds on either side.
- Both backends (`_merge_inmemory`, `_merge_block_adapter`) call the shared helper, so numpy and dask paths get the fix together.

## Test plan
- [x] `pytest xrspatial/tests/test_reproject.py::TestMergeSameCrsYOrientation -v`
  - Single y-ascending raster: `merge` vs `reproject` parity
  - Single y-ascending raster: latitude-encoded gradient appears north-up
  - Single y-descending raster regression guard
  - Multi-raster merge with mixed orientations (constant values)
  - Multi-raster merge with mixed orientations (gradient parity vs reproject)
  - Dask vs numpy parity for y-ascending input
- [x] Full reproject suite: `pytest xrspatial/tests/test_reproject.py` -- 274 passed.

## Backends
- numpy
- dask + numpy
- cupy / dask + cupy share the same Python adapter, so the fix flows through; no GPU-only kernel path involved.

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -1913,6 +1913,14 @@ def _place_same_crs(src_data, src_bounds, src_shape, y_desc,
 
     No reprojection needed -- just index the output rows/columns that
     overlap with the source tile and copy the data.
+
+    The output grid is always north-up (row 0 is the top of
+    ``out_bounds``). When ``y_desc`` is ``False`` the source is
+    y-ascending (row 0 is the bottom of ``src_bounds``); in that case
+    the source window is flipped along y before being written so the
+    placed data has the same north-up orientation as the rest of the
+    output. Without this, ``merge([r])`` of a y-ascending raster
+    silently differs from ``reproject(r, target_crs=r.crs)`` (#2186).
     """
     out_h, out_w = out_shape
     src_h, src_w = src_shape
@@ -1924,7 +1932,8 @@ def _place_same_crs(src_data, src_bounds, src_shape, y_desc,
     s_res_x = (s_right - s_left) / src_w
     s_res_y = (s_top - s_bottom) / src_h
 
-    # Output pixel range that this tile covers
+    # Output pixel range that this tile covers. The output is always
+    # north-up so ``row_start`` is measured from ``o_top`` downward.
     col_start = int(round((s_left - o_left) / o_res_x))
     col_end = int(round((s_right - o_left) / o_res_x))
     row_start = int(round((o_top - s_top) / o_res_y))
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: reproject.merge same-CRS y orientation fix (#2186)

### Blockers
None.

### Suggestions
- `xrspatial/reproject/__init__.py:1985-2004`: the y-ascending branch is correct but pretty dense. A one-line worked example in the comment (something like `src row r maps to output row row_end - 1 - r`) would save the next reader from re-deriving the index math.
- `xrspatial/tests/test_reproject.py:802-812`: `test_mixed_orientation_multi_raster_merge` only asserts tile values inside `if valid_l.any()` / `if valid_r.any()`. If the output grid ever shifts and both masks come up empty, the test passes with zero assertions. Worth hoisting the `.any()` checks into unconditional asserts so it can't quietly turn into a no-op.

### Nits
- `xrspatial/reproject/__init__.py:1957-1962`: `n_cols` is used once. Inlining it keeps the column branch shaped like the two row branches.
- `xrspatial/tests/test_reproject.py:746`: long test names. Not really a problem, just noting it.

### What looks good
- The diagnosis lines up with the code. `_place_same_crs` accepted `y_desc` and ignored it; `reproject()` honored it. `merge([r])` vs `reproject(r)` was always going to disagree on a y-ascending input.
- One helper, two callers: `_merge_inmemory` and `_merge_block_adapter` both go through `_place_same_crs`, so numpy and dask pick up the fix from a single edit.
- I walked the y-ascending branch through the three interesting clip cases (top, bottom, exact-fit). The `out_row_start = row_end - src_hi` offset comes out right in each one.
- The test mix is well chosen for this bug: y-ascending vs reproject parity, y-descending regression guard, multi-raster mixed orientation, and dask vs numpy parity. The latitude-encoded gradient catches the original bug instantly.
- Counterfactual check: with the source change stashed and the tests kept, 3 of the 6 new tests fail with orientation-flip signatures. The tests are exercising the bug, not just the API surface.

### Checklist
- [x] Output contract honored (always north-up)
- [x] Numpy and dask agree (`test_dask_y_ascending_matches_numpy`)
- [x] NaN handling unchanged from prior code
- [x] Edge cases covered (top-clip, bottom-clip, exact-fit)
- [x] Dask chunk boundaries unchanged (same helper, same shapes)
- [x] No new materialization or extra copies
- [ ] No benchmark needed (bug fix on existing code)
- [ ] No README update (no new API, no backend change)
- [x] Docstring updated with the orientation contract

brendancol (commented):
## PR Review (follow-up): reproject.merge same-CRS y orientation fix (#2186)

- Suggestion (worked-example comment in y-ascending branch): fixed in `13f49db` -- the comment now spells out the `src_h=4, row_end=4` flip case.
- Suggestion (hoist `.any()` checks in `test_mixed_orientation_multi_raster_merge`): fixed in `13f49db` -- both masks now hit explicit asserts up front.
- Nit (inline `n_cols`): fixed in `13f49db`.
- Nit (long test names): dismissed. Matches the explicit, behavior-describing style used by the surrounding `TestMerge*` tests, so renaming would create local inconsistency.

### Blockers
None.

### Suggestions
None remaining.

### Nits
None remaining (one prior nit dismissed with reason above).

### What looks good
- Same as the first pass: orientation fix lands in the shared helper, numpy and dask paths share the change, edge cases in the y-ascending branch all line up, and 3 of 6 new tests fail without the fix.
- The follow-up commit keeps the diff small (13 insertions, 9 deletions) and does not touch behavior.

### Checklist
- [x] All findings dispositioned (3 fixed, 1 dismissed with reason)
- [x] Tests still green locally (274 passed)

PR #2192, opened 2026-05-20
reproject: validate source coords are monotonic and regularly spaced (#2184)

### LINKED ISSUE #2184: reproject: irregular x/y coordinates accepted but interpreted as a regular grid
**Describe the bug**

`reproject()` and `merge()` in `xrspatial.reproject` accept rasters with any x/y coordinate arrays, but the pixel math downstream assumes a uniform grid. `_source_bounds()` in `xrspatial/reproject/__init__.py` infers the source pixel size from just the first two coordinate samples:

```python
res_y = abs(float(y[1] - y[0]))
...
res_x = abs(float(x[1] - x[0]))
```

If the coords are non-monotonic, or monotonic but irregularly spaced, that single inferred value is then used as `src_res_x` / `src_res_y` for every pixel in `_reproject_chunk_numpy` / `_reproject_chunk_cupy`. The output looks reasonable but is georeferenced wrong, and no error is raised.

**Expected behavior**

Validate the input coords at the top of `reproject()` and `merge()`, before any CRS resolution or allocation:

1. x and y must be strictly monotonic (all ascending or all descending).
2. Spacing must be regular within a small relative tolerance, e.g. `max|diff - median(diff)| < rtol * |median(diff)|`.

If either check fails, raise a `ValueError` naming the axis and the worst step deviation. No silent coercion.

**Reproduction**

```python
import numpy as np
import xarray as xr

x = np.linspace(-5, 5, 64)
x[32] += 0.1  # perturb one sample
y = np.linspace(50, 40, 64)
data = np.random.rand(64, 64)
raster = xr.DataArray(
    data, dims=['y', 'x'], coords={'y': y, 'x': x},
    attrs={'crs': 'EPSG:4326'},
)
# Today: silently produces wrong georeferencing.
# Wanted: ValueError naming the x axis and the worst step.
reproject(raster, 'EPSG:3857')
```

**Additional context**

`merge()` has the same problem since it calls `_source_bounds` on every input raster.

brendancol:
Closes #2184

## Summary
- `_source_bounds()` inferred pixel resolution from only the first two coord samples and reused that single value for the whole grid, so irregular or non-monotonic x/y silently produced wrong georeferencing.
- Added `_validate_source_coords()` and called it at the top of `reproject()` and `merge()` before any CRS resolution or grid math. Per axis it checks finiteness, strict monotonicity, and that every step is within `1e-6 * |median(diff)|` of the median. Failures raise `ValueError` naming the axis, median step, worst deviation, offending index, and the two coord values at that index. Sub-ULP drift from real GeoTIFFs still passes.
- `merge()` runs the check per raster so the message reports which input failed.

## Backend coverage
Validation is pure NumPy on `coords.values` and runs before backend dispatch, so it fires identically for numpy, cupy, dask+numpy, and dask+cupy inputs. Tests cover all four.

## Test plan
- [x] Regular ascending and descending coords pass.
- [x] Sub-ULP drift from a typical GeoTIFF affine still passes.
- [x] Single-pixel rasters pass (nothing to validate).
- [x] One perturbed sample in x raises with axis, median step, deviation, index, and surrounding coord values.
- [x] One perturbed sample in y raises with the same message shape.
- [x] Non-monotonic coords raise.
- [x] Repeated values raise.
- [x] NaN in coords raises.
- [x] Validation fires before CRS resolution.
- [x] `merge()` reports which raster is bad.
- [x] dask, cupy, and dask+cupy backends all raise from the same path.

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -135,8 +135,15 @@ def _validate_regular_axis(coords, axis_name, func_name, rtol=_REGULAR_COORD_RTO
             f"{func_name}(): coordinate '{axis_name}' contains non-finite "
             f"values (NaN or inf)."
         )
-    diffs = np.diff(arr.astype(np.float64))
+    # np.asarray skips the copy when arr is already float64; np.diff promotes
+    # ints to int64, which is fine but we want float steps for the median /
+    # tolerance math below.
+    diffs = np.diff(np.asarray(arr, dtype=np.float64))
     # Strict monotonicity: every step has the same sign and is non-zero.
+    # `diffs > 0` AND `diffs < 0` are both False for zero steps (repeated
+    # coords), so the combined check rejects them. Do NOT replace this with
+    # a sign-only test like `np.all(np.sign(diffs) == np.sign(diffs[0]))` --
+    # that variant accepts zero steps and lets a repeated coord through.
     if not (np.all(diffs > 0) or np.all(diffs < 0)):
         raise ValueError(
             f"{func_name}(): coordinate '{axis_name}' must be strictly "
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -88,6 +88,86 @@ def _find_spatial_dims(raster):
     return dims[-2], dims[-1]
 
 
+# Default tolerance for the regular-spacing check. Coordinates loaded from
+# real GeoTIFFs can drift a few ULPs from perfectly uniform after pixel-to-
+# world transforms, so 1e-6 (relative) is loose enough to accept those while
+# still catching the single-pixel perturbation case in #2184.
+_REGULAR_COORD_RTOL = 1e-6
+
+
+def _validate_regular_axis(coords, axis_name, func_name, rtol=_REGULAR_COORD_RTOL):
+    """Validate that 1-D coordinate array is strictly monotonic and regular.
+
+    Pixel-resolution math in `_source_bounds` and the chunk workers assumes
+    a uniform grid. Without this check, irregular or non-monotonic coords
+    silently produce wrong georeferencing (see #2184).
+
+    Parameters
+    ----------
+    coords : array-like
+        1-D coordinate values along one axis.
+    axis_name : str
+        Name of the axis ("x" or "y") for the error message.
+    func_name : str
+        Calling function name for the error prefix.
+    rtol : float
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: reproject: validate source coords are monotonic and regularly spaced (#2184)

None.

None.

- `xrspatial/reproject/__init__.py:140` - the strict-monotonicity check uses `np.all(diffs > 0) or np.all(diffs < 0)`. For a 2-element coord with identical values, `diffs = [0.0]` falls through both branches and gives the right error. A short inline comment would stop a future reader from "simplifying" it to a sign check that accepts zero steps.
- `xrspatial/tests/test_reproject.py:2010` - `test_reproject_rejects_repeated_coord` matches the loose pattern `r"strictly"`. Tightening to `r"must be strictly monotonic"` would catch a future reword that drops the word "strictly" elsewhere in the sentence.
- `xrspatial/reproject/__init__.py:138` - `arr.astype(np.float64)` always copies, even when `arr` is already float64. `np.asarray(arr, dtype=np.float64)` skips the copy in that case. Coord arrays are small so the cost is negligible either way.

### What looks good
- Validation runs at the very top of `reproject()` and `merge()`, before CRS resolution or any grid math. The dedicated ordering test (`test_reproject_rejects_irregular_before_crs_resolution`) locks that in.
- The error message names the axis, the median step, the worst absolute deviation, the offending step index, and the two coord values at that index. Enough to find the bad sample without re-running diff.
- `merge()` runs the check per raster and tags the error with `rasters[i]`, so callers know which input failed.
- The 1e-6 relative tolerance is loose enough to accept sub-ULP drift from real GeoTIFFs (covered by `test_reproject_accepts_tiny_floating_drift`) and tight enough to reject the 0.1 single-sample perturbation in the issue reproducer.
- All four backends are covered: numpy (implicit), dask, cupy, dask+cupy. The validator runs on `coords.values`, which is materialized NumPy regardless of the data array backend, so backend parity is structural.
- Single-pixel rasters skip the check rather than falling into an `np.diff` on a zero-length array.
- Non-finite check uses `np.isfinite`.

### Checklist
- [x] Algorithm matches the spec in the issue.
- [x] All implemented backends produce consistent results.
- [x] NaN handling is correct.
- [x] Edge cases covered by tests.
- [x] Benchmark not needed (defensive validation, not a perf-critical path).
- [x] README feature matrix update not applicable (no new public API).

PR #2194, opened 2026-05-20
polygonize: expose atol/rtol, align docstring with float tolerance

### LINKED ISSUE #2173: polygonize: float-tolerance grouping contradicts public docstring
**Describe the bug**

The `polygonize` docstring at `xrspatial/polygonize.py:1624` says regions group cells that "share the same pixel value". The implementation at `xrspatial/polygonize.py:246` does something different for floats: it groups them with an `_is_close` tolerance (`atol=1e-8`, `rtol=1e-5`).

Integer inputs use strict equality. Floats use tolerance. The docstring doesn't mention this split.

**Reproduction**

```python
import numpy as np
import xarray as xr

raster = xr.DataArray(np.array([[1.0, 1.000009, 1.000018]], dtype=np.float64))
column, polys = polygonize(raster)
print(column)
# Actual:    [1.0]    -> one polygon
# Docstring would imply: three polygons (three distinct values)
```

**Expected behavior**

Docstring and code should agree. For a classified float raster (where the float value is the category label and categories are close in magnitude), the current behavior is silent data corruption: distinct categories collapse into one polygon, and the value reported is the first one encountered.

**Why a docstring-only fix isn't enough**

Both behaviors have real users. Tolerance grouping is the right call for rasters that came out of resampling, projection, or raster algebra, where strict equality is fragile. Strict grouping is the right call for classified float rasters where the value is the label.

So the fix needs to:

- Update the docstring to describe what actually happens (tolerance for floats, strict for ints).
- Expose the tolerance as a parameter so callers can pass `atol=0` and opt into strict equality.
- Document the default tolerance value, where the constant lives, and why it's set where it is.

The default stays backward-compatible: tolerance for floats, strict for ints.

**Additional context**

Source references:

- Docstring: `xrspatial/polygonize.py:1624`
- `_is_close` and the tolerance constants: `xrspatial/polygonize.py:236-247`
- CuPy float path routes through the CPU `_polygonize_numpy`, so it inherits whatever the CPU does (`xrspatial/polygonize.py:719-732`).

brendancol:
Closes #2173.

## Summary

- The docstring said "share the same pixel value" but the float path silently grouped pixels with `_is_close` (`atol=1e-8`, `rtol=1e-5`). Integer path used strict equality. Two different behaviors, one undocumented.
- This PR keeps the tolerance default (changing it would be a silent behavior change for everyone using polygonize on resampled / projected float rasters) and adds `atol` / `rtol` kwargs to `polygonize()` so callers with classified float rasters can pass `atol=0.0, rtol=0.0` for strict equality.
- The docstring now describes both behaviors and points at the new parameters. The default constants live as `xrspatial.polygonize._DEFAULT_ATOL` and `_DEFAULT_RTOL`.

## Decision: keep tolerance default

Two reasons:

- Backward compatibility. Tolerance has been the float behavior since the float path was introduced. Flipping the default would silently change polygon counts for any pipeline that processes resampled / projected / arithmetically derived float rasters.
- The CuPy float path already routes through `_polygonize_numpy` to inherit the tolerance behavior (per #2151). Flipping the default would force a different conversation about CuPy parity.

Callers who want strict equality opt in with `atol=0.0, rtol=0.0`.

## Backend coverage

- numpy: parameter wired through `_polygonize_numpy` -> `_calculate_regions`.
- dask+numpy: parameter wired through `_polygonize_dask` -> `_polygonize_chunk` -> `_polygonize_numpy`.
- cupy: integer GPU path uses strict equality (atol/rtol do not apply); float path routes through `_polygonize_numpy` and honors them.
- dask+cupy: same dispatch as dask+numpy.

## Tests

- Default tolerance merges `[[1.0, 1.000009, 1.000018]]` to one polygon.
- `atol=0.0, rtol=0.0` splits the same input into three polygons.
- Integer raster `[[1, 2, 3]]` stays at three polygons even with `atol=10.0, rtol=10.0`.
- Custom intermediate tolerances split or merge as expected.
- Negative atol/rtol raise `ValueError`.
- Dask single-chunk and per-pixel multi-chunk both honor default and strict behavior.

## Test plan

- [x] `pytest xrspatial/tests/test_polygonize.py` (135 passed, 13 skipped)
- [x] Manual repro from the issue runs both ways.
- [ ] CI green.

## Skipped rockout steps

- Step 6 (user guide notebook): this is a small API addition, not a new feature.
- Step 7 (README feature matrix): no new function, no backend support change.

### DIFF
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -1,6 +1,8 @@
 # Polygonize algorithm creates vector polygons for connected regions of pixels
-# that share the same pixel value in a raster.  It is a raster to vector
-# converter.
+# in a raster that group together by pixel value.  It is a raster to vector
+# converter.  Integer rasters group by strict equality; float rasters group
+# by an absolute / relative tolerance (see _DEFAULT_ATOL and _DEFAULT_RTOL,
+# and the public polygonize() docstring for the atol / rtol parameters).
 #
 # Algorithm here uses compass directions for clarity, so +x direction is East
 # and +y direction is North.  2D arrays are flattened to 1D to make maths of
@@ -229,21 +231,34 @@ def _follow(
     return region, points
 
 
+# Default absolute and relative tolerances used by the float pathway of the
+# value-equality predicate.  These values mirror numpy.isclose's defaults
+# and have been the historical defaults of polygonize since the float path
+# was introduced.  A user can override them via the polygonize() public
+# atol / rtol parameters, including passing atol=0.0 and rtol=0.0 to get
+# strict equality for float rasters.
+_DEFAULT_ATOL = 1e-8
+_DEFAULT_RTOL = 1e-5
+
+
 # Generator of numba-compatible comparison functions for values.
 # If both values are integers use a fast equality operator, otherwise use a
-# slower floating-point comparison like numpy.isclose.
+# slower floating-point comparison like numpy.isclose with caller-supplied
+# tolerances.  Passing atol=0.0 and rtol=0.0 reduces the float branch to
+# strict equality.
 @generated_jit(nogil=True, nopython=True)
 def _is_close(
     reference: Union[int, float],
     value: Union[int, float],
+    atol: float,
+    rtol: float,
 ) -> bool:
     if (isinstance(reference, nb.types.Integer) and
             isinstance(value, nb.types.Integer)):
-        return lambda reference, value: value == reference
+        # Integer raster: tolerance does not apply, use strict equality.
+        return lambda reference, value, atol, rtol: value == reference
     else:
-        atol = 1e-8
-        rtol = 1e-5
-        return lambda reference, value: \
+        return lambda reference, value, atol, rtol: \
             abs(value - reference) <= (atol + rtol*abs(reference))
 
 
@@ -273,6 +288,8 @@ def _calculate_regions(
     connectivity_8: bool,
     nx: int,
     ny: int,
+    atol: float,
+    rtol: float,
 ) -> np.ndarray:  # _regions_dtype, shape (nx*ny,)
     # Array of regions to return, integers starting at zero.
     regions = np.zeros_like(values, dtype=_regions_dtype)
@@ -293,7 +310,7 @@ def _calculate_regions(
             matches_W = \
                 (ij % nx > 0 and                         # i > 0
                     (mask is None or mask[ij-1]) and     # W pixel in mask
-                    _is_close(values[ij], values[ij-1]))
+                    _is_close(values[ij], values[ij-1], atol, rtol))
 
             if matches_W:
                 region_W = regions[ij-1]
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: polygonize: expose atol/rtol, align docstring with float tolerance

### Blockers
None.

### Suggestions
- `xrspatial/polygonize.py:1813-1816`: the non-negative checks on `atol`/`rtol` don't catch NaN. `np.nan < 0` is False, so `polygonize(raster, atol=np.nan)` slips past validation and silently produces all singletons (`abs(x) <= nan + ...` is always False). Reject NaN explicitly. One `math.isnan` per arg, no test cost.

### Nits
- `test_polygonize_dask_multi_chunk_default_tolerance` (`xrspatial/tests/test_polygonize.py`): the assertion `np.isclose(v, _REPRO_2173[0]).any()` is looser than it needs to be. It would pass if every chunk emitted a polygon valued exactly `_REPRO_2173[0, 0]` and the close-but-distinct neighbours got dropped. `len(values) == 3` plus a check that each value is in `_REPRO_2173[0]` describes the per-chunk-tolerance contract better. Not blocking; the area=3 assertion already covers total coverage.
- `xrspatial/polygonize.py:1816-1818`: the comment on `float(atol)` / `float(rtol)` is correct but terse. Numba's `_is_close` dispatches on the dtype of `reference`/`value`, not on `atol`/`rtol`. The cast exists so a Python literal `0` doesn't get inferred as int and pick a wrong-typed lambda specialization. One extra line in the comment saves the next reader a trip through `_is_close`.

### What looks good
- Default preserved; `atol=0.0, rtol=0.0` is the strict-equality opt-out the issue asked for.
- Tests hit the repro for default and strict, check the integer path (atol doesn't leak in), and cover dask single-chunk and multi-chunk.
- Scope is honest. Multi-chunk merge keying by exact value is left to a separate rockout and the test comments mention it.
- `_DEFAULT_ATOL` / `_DEFAULT_RTOL` exported as module attributes.
- CuPy float dtypes already routed through `_polygonize_numpy`, so the kwargs flow through for free. `_polygonize_cupy`'s docstring updated to match.

### Checklist
- [x] Algorithm matches intent
- [x] Implemented backends consistent (numpy, dask+numpy, dask+cupy, cupy)
- [x] NaN handling for raster values unchanged; NaN for `atol`/`rtol` flagged above
- [x] Edge cases covered (strict, default, integer with bogus atol, intermediate atol, negative-rejected, dask single + multi-chunk)
- [x] No new function, so no benchmark or README matrix update needed

brendancol (commented):
## Follow-up review (after commit 1b73ea8)

All three findings from the first pass are addressed.

### Disposition
- Suggestion: NaN slipping past validation — **fixed**. `np.isfinite(atol)` now gates both `atol` and `rtol`, and `+/-inf` is rejected for the same reason as NaN. Error message updated to "non-negative finite number".
- Nit: loose multi-chunk assertion — **fixed**. The test now asserts `len(values) == 3` and an exact sorted value match against `_REPRO_2173[0]`.
- Nit: terse comment on `float(atol)` / `float(rtol)` — **fixed**. The comment now explains the cast is for the tolerance args themselves (so a Python `0` doesn't pick the int-typed lambda specialization), not for `_is_close`'s outer dispatch.

### New test
`test_polygonize_nan_atol_raises` exercises NaN and +/-inf for both `atol` and `rtol`.

### Remaining
None.

Tests still green (136 passed, 13 skipped locally).

PR #2195, opened 2026-05-20
polygonize: emit +inf and -inf cells as their own polygons (#2174)

### LINKED ISSUE #2174: polygonize: Inf cells silently absorbed into finite polygons
### Describe the bug

`polygonize` silently absorbs `+inf` and `-inf` cells into adjacent finite polygons on the numpy and dask paths. No Inf polygons come out. The existing tests pin this broken behavior instead of fixing it.

Root cause is in `_is_close` at `xrspatial/polygonize.py:236`. The float branch is:

```python
abs(value - reference) <= (atol + rtol*abs(reference))
```

When `reference = inf` and `value = 1.0`: `abs(1.0 - inf) = inf` and `atol + rtol*abs(inf) = inf`, so `inf <= inf` is True. A finite cell tests "close" to an Inf neighbor, so `_calculate_regions` merges them.

For two `+inf` cells, `abs(inf - inf) = nan` and `nan <= x = False`, so two adjacent `+inf` cells do NOT merge into one polygon either. Both behaviors are wrong.

### Evidence

- Source-bug note: `xrspatial/tests/test_polygonize_coverage_2026_05_19.py:369`
- Pinned assertions for the broken output: lines 400-472 (numpy, dask, dask+cupy)

Reproducer:

```python
import numpy as np
import xarray as xr

data = np.array([
    [1.0, np.inf, 1.0],
    [-np.inf, 1.0, -np.inf],
    [1.0, np.inf, 1.0],
], dtype=np.float64)
v, p = polygonize(xr.DataArray(data), connectivity=4)
print(v)  # [1.0, 1.0, 1.0] -- Inf cells absorbed, no inf polygons
```

The eager-CuPy backend (`_polygonize_cupy`) handles Inf correctly via `_calculate_regions_cupy`.

### Expected behavior

- `+inf` cells form `+inf` polygons.
- `-inf` cells form `-inf` polygons.
- `+inf` is not equal to `-inf`.
- A finite value never equals an Inf.
- NaN handling is out of scope for this fix.

### Fix direction

- Update `_is_close` so Inf comparisons behave correctly: `+inf == +inf`, `-inf == -inf`, `+inf != -inf`, finite never equals Inf.
- Apply to both the numpy single-array path and the dask chunk-stitching path.
- Replace the pinned tests at `test_polygonize_coverage_2026_05_19.py:369-472` with assertions for correct behavior.
- Add tests for: pure `+inf` input, pure `-inf` input, mixed `+inf`/`-inf`/finite, and dask multi-chunk versions of each.

### Affected backends

- numpy: broken, needs fix
- dask+numpy: broken, needs fix
- dask+cupy: broken (uses the numpy chunk path), needs fix
- cupy: already correct

### Out of scope

- NaN polygon semantics
- Float-tolerance docstring clarity
- Chunk-size-dependent float merge issue
- 8-connectivity chunk-corner issue

brendancol:
Closes #2174.

## Summary

- `_is_close` returned True for `(inf, 1.0)` because the tolerance check collapses to `inf <= inf`, so finite cells next to Inf cells were merged into the Inf region. Two same-sign Inf cells also failed to merge because `abs(inf - inf) = nan` and `nan <= x` is False.
- Short-circuit Inf comparisons to exact equality: `+inf == +inf`, `-inf == -inf`, `+inf != -inf`, finite never equals Inf. NaN semantics are unchanged.
- The fix lives in the shared `_is_close` helper used by the numpy path and `_polygonize_chunk`, so it covers all four backends.

## Backend coverage

- numpy: fixed
- dask+numpy: fixed (shares `_is_close` via `_polygonize_chunk`)
- dask+cupy: fixed (also routes per-chunk through the numpy helper)
- cupy: already correct; tests now assert the same shape of result for parity

## What changed in tests

The pinned-buggy tests in `xrspatial/tests/test_polygonize_coverage_2026_05_19.py` (the old `TestInfPins` class at lines 369-472) are replaced with `TestInfPolygons`, `TestPureInfRaster`, and `TestInfBlocks`. Coverage now includes:

- Mixed `+inf` / `-inf` / finite 3x3 raster on numpy, cupy, dask single-chunk, dask multi-chunk, dask+cupy single-chunk, dask+cupy multi-chunk.
- Pure `+inf` and pure `-inf` rasters across all four backends.
- `+inf` block adjacent to `-inf` block with a finite stripe between them, including dask multi-chunk to exercise the boundary merge.

## Test plan

- [x] `pytest xrspatial/tests/test_polygonize_coverage_2026_05_19.py` (72 pass)
- [x] `pytest xrspatial/tests/test_polygonize.py` (124 pass, 13 skipped)
- [x] `pytest xrspatial/tests/test_polygon_clip.py` (23 pass)
- [ ] CI on all configured Python / OS combos

### DIFF
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -232,6 +232,15 @@ def _follow(
 # Generator of numba-compatible comparison functions for values.
 # If both values are integers use a fast equality operator, otherwise use a
 # slower floating-point comparison like numpy.isclose.
+#
+# Inf handling (issue #2174):  the plain tolerance check above misbehaves on
+# infinities.  ``abs(1.0 - inf) = inf`` and ``atol + rtol*abs(inf) = inf`` so
+# ``inf <= inf`` is True, which would merge a finite cell into an adjacent
+# Inf region.  ``abs(inf - inf) = nan`` and ``nan <= x = False`` would also
+# split two same-sign Inf cells.  Branch on whether either operand is
+# infinite and fall back to exact equality there: +inf == +inf, -inf == -inf,
+# +inf != -inf, and finite never equals inf.  NaN semantics are unchanged
+# (any comparison with NaN is False).
 @generated_jit(nogil=True, nopython=True)
 def _is_close(
     reference: Union[int, float],
@@ -243,8 +252,14 @@ def _is_close(
     else:
         atol = 1e-8
         rtol = 1e-5
-        return lambda reference, value: \
-            abs(value - reference) <= (atol + rtol*abs(reference))
+
+        def impl(reference, value):
+            # Exact equality short-circuit handles ±inf == ±inf correctly
+            # and finite-vs-inf falls through as not close.
+            if np.isinf(reference) or np.isinf(value):
+                return value == reference
+            return abs(value - reference) <= (atol + rtol*abs(reference))
+        return impl
 
 
 # Calculate region connectivity for the specified values raster and optional
--- a/xrspatial/tests/test_polygonize_coverage_2026_05_19.py
+++ b/xrspatial/tests/test_polygonize_coverage_2026_05_19.py
@@ -399,6 +399,8 @@ def _check_mixed_inf_output(v, p, total_size):
     finite_vals = [val for val in v if np.isfinite(val)]
     plus_inf = [val for val in v if np.isposinf(val)]
     minus_inf = [val for val in v if np.isneginf(val)]
+    # Four 1.0 corners + 1.0 center, each isolated under 4-connectivity.
+    assert len(finite_vals) == 5, f"finite polygon count wrong: {v}"
     assert all(val == 1.0 for val in finite_vals)
     assert len(plus_inf) == 2, f"+inf polygon count wrong: {v}"
     assert len(minus_inf) == 2, f"-inf polygon count wrong: {v}"
--- a/xrspatial/tests/test_polygonize_coverage_2026_05_19.py
+++ b/xrspatial/tests/test_polygonize_coverage_2026_05_19.py
@@ -364,22 +364,23 @@ class TestAllNanRaster:
 
 
 # ---------------------------------------------------------------------------
-# Cat 2 HIGH: Inf inputs
+# Cat 2 HIGH: Inf inputs (issue #2174 fixed)
 #
-# !!! Source-bug pin (issue #2155) !!!
-# The numpy/dask boundary-tracing backend silently absorbs +Inf and -Inf
-# pixels into adjacent regions instead of emitting them as their own
-# polygons.  This is because _is_close (polygonize.py:240) reduces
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: polygonize +inf / -inf as their own polygons (#2174)

### Blockers
None.

### Suggestions
- `xrspatial/tests/test_polygonize_coverage_2026_05_19.py:392-411` — `_check_mixed_inf_output` asserts that every finite value is 1.0 and that the +inf and -inf polygon counts are 2 each, but it does not assert the count of finite polygons. The docstring claims five 1.0 polygons (four corners + center). A regression that collapses finite cells into fewer polygons while keeping the total area at 9 would still pass. Add `assert len(finite_vals) == 5` to catch that.

### Nits
- `xrspatial/polygonize.py:258-261` — The inner `impl` could short-circuit on the common finite-pair case to skip the two `np.isinf` calls. Numba may already constant-fold this; only worth doing if a microbench shows it helps the hot loop.

### What looks good
- Inf cases all behave correctly: `+inf == +inf`, `-inf == -inf`, `+inf != -inf`, finite never equals Inf. NaN behavior is preserved because `np.isinf(NaN)` is False and the tolerance branch already returns False for NaN.
- One point of repair: `_is_close` is shared by `_calculate_regions` and the dask chunk worker via `_polygonize_chunk -> _polygonize_numpy`, so all four backends pick up the fix.
- Boundary stitching in `_merge_from_separated` keys polygons on the Python float value. `float('inf')`, `float('-inf')`, and finite floats hash and compare correctly, so the chunk-seam merge groups Inf rings by sign.
- Test coverage spans the four backends with mixed, pure-positive, pure-negative, and block-adjacency rasters, plus dask multi-chunk variants. The old pinned-buggy tests at lines 369-472 are replaced with assertions for the corrected behavior.

### Checklist
- [x] Inf semantics correct on all four cases
- [x] NaN handling unchanged
- [x] Edge cases covered (pure inf, mixed sign, chunk boundaries)
- [x] Benchmark not needed (bug fix, no new perf-sensitive code)
- [x] README feature matrix not affected
- [x] Inline comment in `_is_close` documents why the Inf branch exists

brendancol (commented):
## Follow-up review (after fa627b0)

- Suggestion (finite polygon count assertion in `_check_mixed_inf_output`): fixed in fa627b0. The helper now asserts `len(finite_vals) == 5` so a regression that collapses 1.0 cells without changing total area would fail loudly.
- Nit (short-circuit `np.isinf` for finite pairs): dismissed. Numba most likely constant-folds the `or` chain on a non-inf scalar, and adding micro-optimizations without a benchmark would just complicate the helper. Happy to revisit if a profile flags `_is_close` as hot.

No further issues. The new commit re-ran the inf tests (18 pass) and does not touch source.

PR #2196, opened 2026-05-20
rasterize: supercover boundary burn for all_touched=True (#2169)

### LINKED ISSUE #2169: rasterize: all_touched=True misses pixels on shallow/diagonal polygon boundaries
## Describe the bug

When `all_touched=True`, `rasterize` first does the usual center-fill pass and then burns the polygon ring with a plain Bresenham line. Bresenham steps one pixel per dominant-axis step, so it skips the off-axis cells a shallow or diagonal edge passes through. That is not the "supercover" behavior the name `all_touched` implies.

I compared the output to `rasterio.features.rasterize(..., all_touched=True)` on four polygons. Three of them differed. The worst was a diagonal strip: xrspatial burned 10 pixels, rasterio burned 21. Downstream code that relies on these masks for conservative coverage (zone masks, training masks, footprint masks) is undercounting.

## Expected behavior

`all_touched=True` should match `rasterio.features.rasterize(..., all_touched=True)` pixel-for-pixel, up to rasterization tie-breaking on shared edges.

The boundary burn should use a supercover algorithm (Amanatides & Woo style grid traversal), so every cell a ring segment passes through is burned.

## Location

- `xrspatial/rasterize.py:891` (CPU numpy path)
- `xrspatial/rasterize.py:1568` (dask tile path)
- GPU path goes through the same line-burn helper at `xrspatial/rasterize.py:1142`.

## Additional context

The CUDA line and point kernels reuse `_burn_lines_gpu`. The fix should add a separate supercover burn used only for the polygon-boundary path so the existing line/point behavior is not affected.

brendancol:
## Summary
- Fixes #2169. With `all_touched=True`, polygon boundaries were drawn with a Bresenham line, which only steps one cell per dominant-axis step. Shallow and diagonal edges then skipped the off-axis cells the segment crossed, so `xrspatial.rasterize(...)` returned fewer pixels than `rasterio.features.rasterize(..., all_touched=True)` on the same input.
- The boundary burn now uses an Amanatides & Woo grid traversal in float pixel space. Endpoints are clipped to the raster with Liang-Barsky, then the walker advances one cell at a time along whichever axis hits its next boundary first. Every cell the mathematical segment crosses is burned.
- All four backends (numpy, cupy, dask+numpy, dask+cupy) share the supercover path so they agree pixel-for-pixel with each other and with rasterio for polygons whose vertices do not sit exactly on a grid line. Polygon fill (when `all_touched=False`), line burning, and point burning are unchanged.

## Test plan
- [x] New `xrspatial/tests/test_rasterize_all_touched_supercover_2169.py` pins parity against `rasterio.features.rasterize(all_touched=True)` for four polygons (diagonal strip, thin triangle, edges on pixel centres, concave subpixel) on all four backends. 18 tests pass locally (4 backends x 4 polygons + 2 regression cases).
- [x] Existing `xrspatial/tests/test_rasterize_accuracy.py` rasterio reference tests still pass.
- [x] Full rasterize suite (298 tests) green locally.

Closes #2169

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -1464,31 +1464,29 @@ def _ensure_gpu_kernels(merge_fn, should_write):
         else:
             step_y = 0
 
-        INF = 1.0e308
-
         if dx != 0.0:
             t_delta_x = 1.0 / (dx if dx > 0.0 else -dx)
         else:
-            t_delta_x = INF
+            t_delta_x = math.inf
 
         if dy != 0.0:
             t_delta_y = 1.0 / (dy if dy > 0.0 else -dy)
         else:
-            t_delta_y = INF
+            t_delta_y = math.inf
 
         if dx > 0.0:
             t_max_x = (float(cx + 1) - x0) / dx
         elif dx < 0.0:
             t_max_x = (float(cx) - x0) / dx
         else:
-            t_max_x = INF
+            t_max_x = math.inf
 
         if dy > 0.0:
             t_max_y = (float(cy + 1) - y0) / dy
         elif dy < 0.0:
             t_max_y = (float(cy) - y0) / dy
         else:
-            t_max_y = INF
+            t_max_y = math.inf
 
         max_steps = abs(end_cx - cx) + abs(end_cy - cy) + 2
         for _ in range(max_steps):
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -11,6 +11,7 @@ Supports numpy, cupy, dask+numpy, and dask+cupy backends.
 """
 from __future__ import annotations
 
+import math
 import warnings
 from typing import Optional, Tuple, Union
 
@@ -490,6 +491,13 @@ _EMPTY_LINES = (np.empty(0, np.int32), np.empty(0, np.int32),
                 np.empty(0, np.int32), np.empty(0, np.int32),
                 np.empty(0, np.int32))
 
+# Float-coordinate variant used by the supercover boundary burn.
+# Endpoints are kept in subpixel space so grid traversal can pick up
+# every cell a segment crosses, not just the dominant-axis steps.
+_EMPTY_LINES_FLOAT = (np.empty(0, np.float64), np.empty(0, np.float64),
+                      np.empty(0, np.float64), np.empty(0, np.float64),
+                      np.empty(0, np.int32))
+
 
 def _extract_line_segments(geometries, bounds, height, width):
     """Parse LineString/MultiLineString geometries into pixel-space segments.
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: rasterize: supercover boundary burn for all_touched=True (#2169)

None.

None.

- `xrspatial/rasterize.py:1467` -- the GPU kernel uses `INF = 1.0e308` instead of `math.inf`. numba.cuda accepts `math.inf` and reads clearer about intent. Not blocking; the magic constant still saturates the `t_max_x < t_max_y` comparisons correctly.
- `xrspatial/rasterize.py:780` -- `np.array([-dx, dx, -dy, dy])` builds a 4xN temporary on every call. For very large polygon counts, `np.stack(..., axis=0)` or computing each plane inline would shave a copy. Probably not measurable for typical polygon counts; mentioning it only because the line is repeated from the existing line-clipping path.
- The new test file `test_rasterize_all_touched_supercover_2169.py` runs all four backends per case, but the cupy/dask+cupy classes silently skip on CI without a GPU. That matches existing test conventions in this repo so it is fine; just noting that GPU parity is covered locally only until CI grows a GPU runner.

### What looks good
- Algorithm: Amanatides & Woo grid traversal is implemented correctly. `t_max` initialization handles both positive and negative `dx`/`dy`, the `t_delta` formula uses `1/|d|`, and the tie-break case advances both axes together so the supercover sweep does not over-burn diagonal-neighbour cells.
- Bounds: world-space Liang-Barsky clipping happens before the pixel-space walk, so the kernel never iterates over far-away cells.
- The internal bounds check (`0 <= cy < height and 0 <= cx < width`) is a sensible guard for floating-point endpoints that floor to `width` or `height`.
- `max_steps = abs(end_cx - cx) + abs(end_cy - cy) + 2` is a tight upper bound for the cell-walk length and guards against pathological floating-point loops.
- All four backends (numpy, cupy, dask+numpy, dask+cupy) call the same float-coordinate extraction and the same supercover walker, so backend parity is structural rather than coincidental.
- The non-all-touched path (scanline fill), point burning, and line burning are unchanged. The Bresenham `_burn_lines_cpu` and `_burn_lines_gpu` kernels are still used for `LineString`/`MultiLineString` geometries.
- Docstring update on `rasterize(... all_touched=True ...)` accurately describes the new behaviour and references rasterio for parity expectations.
- Test coverage: 4 polygon shapes (diagonal strip, thin triangle, edges-on-pixel-centers, concave subpixel) x 4 backends, plus a regression-specific "diagonal was undercounted before" case and a dask-matches-numpy parity test. All 18 pass locally. Broader rasterize suite (298 tests) still green.

### Checklist
- [x] Algorithm matches reference (Amanatides & Woo 1987, "A Fast Voxel Traversal Algorithm for Ray Tracing")
- [x] NaN handling unchanged from existing path
- [x] Edge cases covered by tests (subpixel vertices, edges on pixel centres, concave shapes)
- [x] Dask chunk boundaries handled (segments clipped per-tile, parity test asserts dask == eager numpy)
- [x] No benchmark needed (bug fix, no new public API)
- [x] README feature matrix unchanged (rasterize was already listed)
- [x] Docstring updated

brendancol (commented):
## Follow-up review

Dispositions for the previous review:
- Nit 1 (`INF = 1.0e308` -> `math.inf`): fixed in 5795f96. Tests still pass on all four backends locally.
- Nit 2 (`np.array([-dx, dx, -dy, dy])` 4xN allocation): deferred. The same pattern lives in the older `_extract_lines_vectorized` Liang-Barsky code; rewriting only the new copy at line 780 would leave the codebase inconsistent. Better handled as a follow-up sweep across both call sites.
- Nit 3 (CI lacks a GPU runner for cupy/dask+cupy classes): dismissed. The test gating uses the same pattern as the rest of the repo, and CI GPU coverage is out of scope for this PR.

No new findings on the latest commit.

PR #2197, opened 2026-05-20
polygonize: fix Dask 8-connectivity diagonal merge at chunk corners (#2172)

### LINKED ISSUE #2172: polygonize: Dask 8-connectivity splits polygons at chunk corners
**Describe the bug**

The Dask backend of `polygonize` does not match the NumPy polygon count under 8-connectivity. The docstring at `xrspatial/polygonize.py:1646` already notes that extra polygon splits happen at chunk corners.

This isn't a cosmetic mismatch. Region counts, joins, and area statistics computed downstream will disagree between the two backends.

**Reproduction**

```python
import numpy as np
import xarray as xr
import dask.array as da

arr = np.array([[1, 0], [0, 1]], dtype=np.int32)

# NumPy backend: 2 polygons
raster_np = xr.DataArray(arr)
vals_np, polys_np = polygonize(raster_np, connectivity=8)
print(len(vals_np))  # 2

# Dask backend with chunks (1, 1): 4 polygons
raster_dk = xr.DataArray(da.from_array(arr, chunks=(1, 1)))
vals_dk, polys_dk = polygonize(raster_dk, connectivity=8)
print(len(vals_dk))  # 4
```

Under `connectivity=8`, the two `1` cells are diagonally adjacent and should merge into a single polygon. Same for the two `0` cells. NumPy reports 2 polygons. Dask with `chunks=(1, 1)` reports 4 because each cell sits in its own chunk and the diagonal connection across the chunk corner is not detected.

**Expected behavior**

Dask should match NumPy's polygon count and geometry for the same input and `connectivity` value, regardless of chunking.

**Fix direction**

The Dask chunk-stitching logic handles edges between adjacent chunks today, which covers 4-connectivity. For 8-connectivity it also needs a corner-merge pass over the four cells meeting at every interior chunk corner. The pass should look at the 2x2 corner neighborhood at every interior `(chunk_y_boundary, chunk_x_boundary)` and merge the two diagonal cells when they share the same value, reusing the existing float tolerance helper.

The pass needs to stay chunked. Materializing the full array would defeat the point of the Dask path on large rasters.

**Test coverage to add**

Compare polygon count and per-region area between NumPy and Dask (multi-chunk) under `connectivity=8` on:
- The repro `[[1, 0], [0, 1]]`
- A larger checkerboard
- A diagonal stripe that crosses several chunk corners

**Docstring**

Update or remove the caveat at `xrspatial/polygonize.py:1646` once the fix lands.

brendancol:
## Summary

Closes #2172.

- Fixes Dask `polygonize` with `connectivity=8` so polygon counts and per-region areas match NumPy when same-value cells are only diagonally adjacent across a chunk boundary.
- Cause: `_pick_next_edge` always took the smallest CCW turn at degree-4 vertices, which splits the figure-8 ring NumPy produces for diagonal-only adjacency into two separate squares.
- `connectivity_8` is now plumbed through `_merge_polygon_rings` and `_trace_rings` into `_pick_next_edge`. Under 8-connectivity it picks the largest forward-going CCW turn at degree-4 vertices instead, which preserves figure-8 rings across chunk boundaries.
- Drops the docstring caveat that warned about extra polygon splits at chunk corners under Dask 8-connectivity.

## Backend coverage

- numpy: unchanged.
- cupy: unchanged.
- dask+numpy: bug fixed.
- dask+cupy: same Dask code path, fix applies.

## Test plan

- [x] New `xrspatial/tests/test_polygonize_issue_2172.py` covers:
  - Original repro `[[1, 0], [0, 1]]` with `chunks=(1, 1)`.
  - 4x4 and 6x6 checkerboards across several chunk shapes.
  - Diagonal stripes that cross multiple chunk corners.
  - Float and NaN inputs.
  - Random 20x20 across several chunk shapes.
  - 4-connectivity counts unchanged (separate test).
- [x] Existing `test_polygonize.py` (124 tests) still passes.
- [x] `test_polygonize_coverage_2026_05_19.py` still passes except for one pre-existing CuPy +inf test owned by a parallel rockout.

## Out of scope

This PR does not touch the chunk-size-dependent float merge (#2171), Inf handling (#2174), or the float-tolerance docstring. Parallel rockouts are handling those.

### DIFF
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -853,11 +853,17 @@ def _pick_next_edge(adj, prev_vertex, current_vertex,
     polygons that touch only at a corner as separate rings, matching
     NumPy 4-connectivity semantics.
 
-    For ``connectivity_8=True``, pick the largest CCW turn (smallest
-    CW turn) at degree-4 vertices.  This pairs up the crossing edges
-    diagonally, producing a single figure-8 ring across the shared
-    vertex — exactly what NumPy 8-connectivity produces when two
-    diagonally adjacent cells share a value.
+    For ``connectivity_8=True``, prefer "straight through" (rel == 0)
+    when available, otherwise pick the largest CCW turn (smallest CW
+    turn) at degree-4 vertices.  Within a single ring, ``rel == 0``
+    cannot occur on a grid because the simplify pass folds consecutive
+    same-direction edges; but ``_merge_polygon_rings`` operates on
+    multiple rings sharing a vertex, where one ring can pass straight
+    through V while another corners at V.  Preferring "straight
+    through" keeps such rings continuous; falling back to the largest
+    turn pairs up the diagonal crossings into a single figure-8 ring,
+    matching the NumPy 8-connectivity output for two diagonally
+    adjacent same-value cells.
     """
     targets = adj[current_vertex]
     if len(targets) == 1:
@@ -869,22 +875,24 @@ def _pick_next_edge(adj, prev_vertex, current_vertex,
 
     best = None
     if connectivity_8:
-        # Prefer the LARGEST forward-going CCW turn so that crossings
-        # produced by diagonal-only adjacency stay merged into figure-8
-        # rings.  ``rel == 0`` (straight ahead) is impossible on a grid
-        # because consecutive ring edges never go in the same direction
-        # without the trace having already turned, but we still avoid
-        # 180° u-turns when other options exist.
+        # Priority order at a degree-4 vertex:
+        #   1. ``rel == 0`` (continue straight) -- keeps a ring that
+        #      passes through V on the same trajectory.
+        #   2. ``rel == 3`` (90 deg CW) -- pairs the diagonal-only
+        #      crossing into a figure-8 ring.
+        #   3. ``rel == 1`` (90 deg CCW) -- the 4-connectivity choice,
+        #      used only if 0 and 3 are unavailable.
+        #   4. ``rel == 2`` (180 deg u-turn) -- last resort.
+        priority = {0: 4, 3: 3, 1: 2, 2: 1}
         best_rel = -1
         for target in targets:
             dx = target[0] - current_vertex[0]
             dy = target[1] - current_vertex[1]
             out_angle = _DIR_ANGLE[(dx, dy)]
             rel = (out_angle - incoming_angle) % 4
-            if rel == 2:
-                rel = -1  # 180° u-turn -> last resort
-            if rel > best_rel:
-                best_rel = rel
+            score = priority[rel]
+            if score > best_rel:
+                best_rel = score
                 best = target
     else:
         best_rel = 5
@@ -1515,8 +1523,12 @@ def _merge_polygon_rings(polys_list, connectivity_8=False):
     return _group_rings_into_polygons(simplified)
 
 
-def _merge_chunk_polygons(chunk_results, transform):
-    """Merge polygons from all chunks and return final output."""
+def _merge_chunk_polygons(chunk_results, transform, connectivity_8=False):
+    """Merge polygons from all chunks and return final output.
+
+    ``connectivity_8`` is forwarded to :func:`_merge_polygon_rings` to
+    select the degree-4-vertex pairing rule used by the trace step.
+    """
     all_interior = []
     boundary_by_value = {}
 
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: polygonize: fix Dask 8-connectivity diagonal merge at chunk corners (#2172)

None.

- `xrspatial/polygonize.py:884`: the 8-conn branch in `_pick_next_edge` keeps `rel == 0` (continue straight) in the max search with weight 0, which loses to `rel == 3` (right turn). At a degree-4 vertex where a straight-through ring meets a corner ring, the trace can cross from the straight-through ring into the corner ring instead of continuing straight. The current tests don't hit this case because all degree-4 vertices in the inputs are crossings (both incoming directions are turns). Suggest giving `rel == 0` priority over `rel == 3`, or excluding `rel == 0` from the max search. Add a regression test where a same-value polygon spans a chunk-boundary mid-edge (not a corner) and another same-value polygon has a corner at the same point.

- `xrspatial/tests/test_polygonize_issue_2172.py:152`: the random-20x20 test loops `for chunks in [...]`. On failure pytest reports only the last value. Parametrize the chunks list so each shape is its own test case.

- `xrspatial/polygonize.py:1561`: `_merge_chunk_polygons` (unused now) still calls `_merge_polygon_rings(polys_list)` without `connectivity_8`. Delete it if dead, otherwise plumb `connectivity_8` through so the bug isn't reintroduced if it gets revived.

- `xrspatial/polygonize.py:858`: docstring uses an em dash ("vertex — exactly what NumPy"). Rest of the file uses ASCII punctuation; replace with a comma or hyphen.

- `xrspatial/polygonize.py:879`: comment says `rel == 0` "is impossible on a grid because consecutive ring edges never go in the same direction without the trace having already turned". That's true within a single ring, not across multiple rings merged into one edge set, which is exactly the case `_merge_polygon_rings` handles. Fix the comment to match actual behaviour.

- `xrspatial/tests/test_polygonize_issue_2172.py:53`: `pytest.approx(area_dk[val], abs=1e-9)` is tighter than needed. Polygon areas come back as exact integers for integer input and exact floats for float input; `abs=0` works. Not load-bearing.

### What looks good

- Fix is minimal and surgical. `_pick_next_edge`, `_trace_rings`, `_merge_polygon_rings`, and `_merge_from_separated` gain a `connectivity_8` flag with default `False`, so 4-conn behaviour is preserved bit-for-bit.
- Docstring caveat at the old line 1646 is removed.
- Tests cover the original repro, multiple chunk shapes, checkerboards, diagonal stripes, NaN inputs, and float-tolerance paths.
- No per-cell allocation. Zero extra memory cost relative to the existing path. No OOM concern.

### Checklist
- [x] Algorithm matches reference (NumPy backend).
- [x] All implemented backends consistent (NumPy, Dask+NumPy verified; Dask+CuPy unchanged code path).
- [x] NaN handling unchanged (covered by `test_float_with_nan`).
- [x] Edge cases covered (checkerboard, stripe, NaN, multiple chunk shapes).
- [ ] Benchmark: none added; fix sits in a non-perf-critical merge step.
- [ ] README feature matrix: N/A (no new function).
- [x] Docstrings updated.

brendancol (commented):
## Follow-up review after c822056

### Disposition of the previous review

- Suggestion: rel==0 priority -- fixed in `_pick_next_edge` via explicit priority table {0: 4, 3: 3, 1: 2, 2: 1}.  rel==0 (straight through) now wins, rel==3 (right turn / figure-8 pair) is second, rel==1 (4-conn left turn) is third, rel==2 (u-turn) is last.
- Suggestion: parametrize random-20x20 test -- done; each chunk shape is its own test case.
- Suggestion: plumb connectivity_8 through `_merge_chunk_polygons` -- done.  The function is unused in the current call graph but the parameter is in place if it gets revived.
- Nit: em dash in docstring -- removed.
- Nit: misleading "rel==0 is impossible" comment -- rewritten to reflect that rel==0 is possible across multiple rings sharing a vertex (which is exactly the case `_merge_polygon_rings` handles).
- Nit: pytest.approx tolerance -- tightened from 1e-9 to 1e-12 (areas are exact for integer inputs).

### New tests

`TestMergePolygonRingsDirect` pins the merge behaviour directly without going through the full polygonize path:
- Two diagonally-adjacent unit squares merge to a single figure-8 ring under 8-conn.
- Same input under 4-conn keeps them separate.
- Two edge-sharing squares still cancel and merge under 8-conn.

### Residual findings

None.

### Checklist
- [x] Algorithm matches reference.
- [x] All implemented backends consistent.
- [x] NaN handling unchanged.
- [x] Edge cases covered.
- [ ] Benchmark: N/A.
- [ ] README feature matrix: N/A.
- [x] Docstrings updated.

PR #2198, opened 2026-05-20
Use CUDA atomics in GPU rasterize kernels (#2167)

### LINKED ISSUE #2167: GPU point/line rasterize kernels have racy overlap writes
**Describe the bug**

`_apply_merge_gpu` at xrspatial/rasterize.py:1038 does non-atomic read-modify-write on `out`, `written`, and `order`. The point and line kernels (`_burn_points_gpu`, `_burn_lines_gpu` at xrspatial/rasterize.py:1128 and xrspatial/rasterize.py:1141) launch one CUDA thread per point or segment.

When two threads land on the same pixel, the writes race. That happens with overlapping points, crossing lines, duplicate segments, or `all_touched=True` polygon boundaries that share pixels. The `last`, `first`, `sum`, `count`, `min`, and `max` aggregators on cupy can return nondeterministic results that disagree with the numpy backend.

Polygon fill by itself is mostly safe because the scanline fill runs one thread per row and walks columns serially, so the same pixel is not touched by two threads inside a single polygon. Mixed point/line workloads and shared boundaries between polygons and other geometries are not.

**Expected behavior**

Deterministic results across runs that match the numpy backend for all six built-in aggregators (`last`, `first`, `sum`, `count`, `min`, `max`) on cupy, even when geometries overlap.

**Reproducer**

Rasterize two coincident points with `merge='sum'` on cupy. The pixel should equal `props_a + props_b`. Today it can be either single value, or the sum, depending on thread interleaving.

**Additional context**

Fix needs CUDA atomics on the per-pixel state:

- `sum` / `count`: `cuda.atomic.add`
- `min` / `max`: `cuda.atomic.min` / `cuda.atomic.max` on the value array
- `first` / `last`: deterministic conflict resolution via `cuda.atomic.min` / `cuda.atomic.max` on a per-pixel winning-index array, then a second pass that materializes the winner's value

The numpy / numba CPU path is unaffected and should stay as-is.

brendancol:
## Summary

- The GPU point and line rasterize kernels used non-atomic
  read-modify-write on per-pixel state, so overlapping geometries
  returned values that disagreed with the numpy backend and varied
  between runs. Switch them to CUDA atomics for the six built-in
  aggregators (last, first, sum, count, min, max).
- first / last use a two-pass scheme: pass 1 resolves the winning
  input index with atomic min/max on an index buffer; pass 2 stamps
  the value in.
- Add regression tests covering coincident points, crossing line
  segments, and duplicate segments, with both cross-backend parity
  and per-run determinism checks.

Backends touched: cupy, dask+cupy. numpy and dask+numpy paths are
unchanged.

## Test plan

- [x] New tests in `xrspatial/tests/test_rasterize_gpu_race_2167.py`
      pass on the local GPU.
- [x] Existing rasterize tests still pass on the local GPU.

Closes #2167

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -1022,28 +1022,103 @@ def _check_gpu_edge_cap(row_ptr):
         )
 
 
-def _ensure_gpu_kernels(merge_fn, should_write):
+#: Built-in merges whose GPU path uses CUDA atomics so overlapping
+#: writes are deterministic.  User-supplied merge callables fall back to
+#: the non-atomic closure path (the public ``merge_fn`` signature has no
+#: slot for atomicity, and we cannot synthesise atomics for an arbitrary
+#: Python function).
+_GPU_ATOMIC_MERGES = ('sum', 'count', 'min', 'max', 'first', 'last')
+
+
+def _ensure_gpu_kernels(merge_fn, should_write, merge_name=None):
     """Compile CUDA kernels for the given merge + predicate pair and cache.
 
     Each unique pair produces a separate set of kernels because CUDA kernels
     cannot accept function arguments; both functions are captured by closure
     at compile time.
+
+    When ``merge_name`` is one of the built-in merges (``sum``, ``count``,
+    ``min``, ``max``, ``first``, ``last``), the returned kernel set uses
+    CUDA atomics on ``out`` / ``order`` so that overlapping point writes,
+    crossing line segments, and shared polygon-boundary pixels produce
+    deterministic results matching the CPU backend.  Otherwise (callable
+    or unknown merge) the kernels keep the legacy non-atomic
+    read-modify-write path, which is racy on overlap but preserved for
+    user callables whose signature has no atomic equivalent.
     """
-    key = (id(merge_fn), id(should_write))
+    key = (id(merge_fn), id(should_write), merge_name)
     if key in _gpu_kernel_cache:
         return _gpu_kernel_cache[key]
 
     from numba import cuda
 
+    atomic_mode = merge_name if merge_name in _GPU_ATOMIC_MERGES else None
+
+    # The per-pixel write strategy depends on ``atomic_mode``.  Numba
+    # treats free Python variables in a ``cuda.jit`` closure as compile-
+    # time constants, so the if/elif ladder below is folded away and
+    # each compiled kernel contains only the relevant atomic call.
+    #
+    # Initialisation contract (launcher fills ``out`` / ``order`` before
+    # the kernel runs):
+    #   sum, count    -> out = 0.0           , order = MIN_INT64
+    #   min           -> out = +inf          , order = MIN_INT64
+    #   max           -> out = -inf          , order = MIN_INT64
+    #   first         -> out = fill          , order = MAX_INT64
+    #   last          -> out = fill          , order = MIN_INT64
+    #   None (user)   -> out = fill          , order = -1 (legacy)
+    #
+    # For sum/count/min/max, ``order`` is repurposed as a touched mask:
+    # any non-sentinel value means the pixel was written by at least
+    # one geometry.  A host-side post-pass blends in ``fill`` where
+    # the pixel is still untouched, and substitutes ``fill`` for the
+    # numerical sentinels used by min/max init.
+    #
+    # For first/last, the kernel is two-pass: pass 1 resolves the
+    # winning input index atomically into ``order``; pass 2 stamps the
+    # winner's ``props[0]`` value into ``out``.  ``pass_id`` selects
+    # which branch executes (kernels are compiled once per pass).
+
     @cuda.jit(device=True)
     def _apply_merge_gpu(out, written, order, r, c, props, new_idx):
-        is_first = np.int64(written[r, c] == 0)
-        cur_idx = order[r, c]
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Use CUDA atomics in GPU rasterize kernels (#2167)

None.

- `xrspatial/rasterize.py:1569-1577` -- In `_run_cupy` the boundary-burn block uses `poly_geoms` and `poly_ids` from the polygon-fill block above. The boundary path runs whenever `all_touched and poly_geoms` is truthy, and `poly_geoms` comes from `_classify_geometries` earlier, so it's defined even if no edges crossed the raster. The same shape appears in `_rasterize_tile_cupy:2008-2031`, where `poly_geoms` is bound inside `if poly_wkb:` and only used when `poly_wkb and all_touched`. Safe today, fragile to future refactors. Worth nesting the boundary block inside the polygon block, or adding a comment, so the dependency is explicit.
- `xrspatial/rasterize.py:1083-1109` -- `_apply_merge_gpu` now branches on `atomic_mode` at compile time, so each merge produces a separate compiled kernel. Fine. But `written` is passed into every kernel and only read in the legacy user-callable branch. For atomic modes it's an `(H, W) int8` allocation that the kernels never touch. Worth skipping the `written` allocation for atomic modes. Memory is small but it's a clean win.

- `xrspatial/rasterize.py:1079-1080` -- The comment block says `` `pass_id` selects which branch executes (kernels are compiled once per pass)``. The implementation has no `pass_id` parameter; pass 1 and pass 2 are separate kernels named `..._gpu` and `..._gpu_pass2`. Update the comment to match the code.
- `xrspatial/tests/test_rasterize_gpu_race_2167.py` -- An `all_touched=True` scenario with overlapping polygon boundaries would be a useful addition. The cross-backend parity test would already catch a regression there, but an explicit scenario makes the coverage clearer for the next reader.

### What looks good

- The atomic strategy follows the fix sketched in issue #2167: atomic add for sum/count, atomic min/max on the value buffer for min/max, two-pass index-resolution for first/last.
- The numpy and dask+numpy paths are untouched.
- The kernel cache key now includes `merge_name`, so atomic and non-atomic kernels coexist without collision.
- Tests cover all six aggregators across three overlap scenarios, with both cross-backend parity and per-run determinism.
- The legacy non-atomic closure path is kept for user callables, which is the right call given the public `merge_fn` signature.
- `_gpu_finalize_buffers` handles the +inf / -inf initialisers for min/max so the user-visible fill semantics stay consistent with numpy.

### Checklist

- [x] Algorithm matches the strategy described in issue #2167.
- [x] All six built-in aggregators produce results consistent with the numpy backend (verified by tests).
- [x] NaN / fill handling is correct (finalize blends fill into untouched pixels).
- [x] Edge cases are covered (coincident points, crossing lines, duplicate segments).
- [n/a] Dask chunk boundaries: tiles partition the output grid, so per-tile atomicity is sufficient. dask+cupy uses the same kernel set via `_rasterize_tile_cupy`.
- [x] No premature materialization. Device transfers are staged once and reused for pass 2.
- [n/a] Benchmark: this is a correctness fix, not a perf change.
- [n/a] README feature matrix: no public API changed.
- [x] Docstrings updated (`_ensure_gpu_kernels` and `_run_cupy` describe the atomic path).

brendancol (commented):
## Follow-up review (after 2842709)

All four findings from the prior review are addressed.

### Disposition

- Suggestion 1 (fragile `poly_geoms` dependency in boundary block) -- **fixed**. `_rasterize_tile_cupy` now stages the boundary launch inside the `if poly_wkb:` block so the dependency is local. `_run_cupy` got an explicit comment noting that `poly_geoms` comes from `_classify_geometries` unconditionally and the same shape is safe there.
- Suggestion 2 (`written` allocation wasted for atomic modes) -- **fixed**. Atomic modes now get a `(1, 1) int8` placeholder; the kernels never touch it for those modes, and the legacy user-callable path still gets a full `(H, W)` allocation.
- Nit 1 (stale `pass_id` comment) -- **fixed**. Comment now describes the actual two-kernel layout.
- Nit 2 (no `all_touched=True` polygon scenario) -- **fixed**. Added shared-boundary polygon tests covering cross-backend parity and per-run determinism for all six aggregators.

### Verification

- 50 tests pass in `xrspatial/tests/test_rasterize_gpu_race_2167.py` (up from 38 in the first pass).
- 325 tests pass across the full rasterize suite (`test_rasterize.py`, `test_rasterize_accuracy.py`, `test_rasterize_coverage_2026_05_17.py`, `test_rasterize_tile_props_slice_2020.py`, `test_rasterize_gpu_race_2167.py`).

brendancol:
Heads up on CI: the three failing `run (..., 3.14)` jobs are failing on `test_polygonize_dask_multi_chunk_default_tolerance` and `test_polygonize_dask_multi_chunk_strict_float` in `xrspatial/tests/test_polygonize.py` (`assert 2 == 3`).

These two failures are pre-existing on `main`, not introduced by the atomics work in this PR. The same two failures show up on the unrelated PR #2196 against the same `main` head. Confirmed by checking recent `main` runs — the polygonize tests have been red there independently.

Everything in this PR's surface area (the new GPU rasterize race tests, the existing rasterize suite) passes locally on the merged branch. Happy to address the polygonize failures in a separate change if useful, but they did not regress here.

Issue #2180, opened 2026-05-20
GeoTIFF: migrate VRT backend to shared finalization helpers (PR E of #2162)


brendancol:
Parent: #2162
Depends on: #2177 (helpers), #2179 (eager helper signature finalized), #2178 (lazy helper signature finalized)

## Goal

Bring the VRT backend into the shared finalization. Today VRT builds attrs by calling `metadata_to_attrs(GeoTIFFMetadata(...))` directly and uses `_compute_georef_status_from_parts()` inline, bypassing `_validate_read_geo_info` and `_populate_attrs_from_geo_info`.

#2162 allows documented divergence where the helper genuinely cannot apply ("treat VRT masking drift as a follow-up bug unless it is cheap to fix during extraction"). This PR should attempt full migration first, and only accept divergence where the cost is large.

## Scope

Two call sites in `xrspatial/geotiff/_backends/vrt.py`:

1. `vrt.py:357-505` — VRT eager finalization. Migrate to `_finalize_eager_read`. VRT builds `GeoTIFFMetadata` from `vrt.bands[].nodata`, `vrt.crs_wkt`, `vrt.geo_transform`. Either (a) construct a synthetic `geo_info` shape from those parts and feed the helper, or (b) document why a specific aspect must diverge and call the helper for the rest.
2. `vrt.py:882-942` — VRT chunked finalization. Migrate to `_finalize_lazy_read_attrs`. Same approach.

Where divergence is unavoidable, leave an inline comment naming the constraint (per-band nodata, missing_sources synthesis, etc.) so it does not read as oversight.

Preserve all current behavior:

- `band_nodata` per-band selection still works.
- `missing_sources` synthesis still works.
- VRT eager and chunked both pass `test_vrt_band_nodata_1598.py` and `test_open_geotiff_missing_sources_1810.py`.

## Tests

Add `xrspatial/geotiff/tests/test_vrt_finalization_parity_2162.py`. Assert:

- VRT eager attrs match eager numpy attrs (where applicable) for equivalent inputs.
- VRT chunked attrs match dask numpy attrs (where applicable) for equivalent inputs.
- `band_nodata` paths produce the right per-band attrs.
- `missing_sources` paths still work.
- `georef_status` matches across VRT and non-VRT paths for the five states.

## Files

- `xrspatial/geotiff/_backends/vrt.py` (rewrite 357-505 and 882-942)
[...]

Issue #2179, opened 2026-05-20
GeoTIFF: migrate eager backends to shared eager finalization helper (PR D of #2162)


brendancol:
Parent: #2162
Depends on: #2177 (adds the eager helper)

## Goal

Replace duplicated finalization in the four eager paths with `_finalize_eager_read` from #2177.

## Scope

Four call sites:

1. `xrspatial/geotiff/__init__.py:711-867` — eager numpy in `open_geotiff`. About 150 lines.
2. `xrspatial/geotiff/_backends/gpu.py:470-508` — GPU CPU-fallback eager (stripped or bps-mismatch path).
3. `xrspatial/geotiff/_backends/gpu.py:861-880` — GPU local eager (main path).
4. `xrspatial/geotiff/_backends/gpu.py:980-1016` — GPU HTTP eager.

Each site is a near-copy of: validate geo_info, populate attrs, mask using a sentinel, compute `nodata_pixels_present`, cast dtype, record `nodata_dtype_cast`, call `_set_nodata_attrs`, construct DataArray.

The three GPU sites derive `mask_sentinel` three different ways:

- `_backends/gpu.py:470-508`: from `_cpu_fallback_geo._mask_nodata`.
- `_backends/gpu.py:861-880`: from `_mw_mask_nodata` local.
- `_backends/gpu.py:980-1016`: from raw `nodata`.

The helper takes `mask_sentinel` as an explicit parameter. Callers pass whichever they have.

Preserve all current behavior:

- `attrs['nodata_pixels_present']` is set as a bool from a real scan of the array.
- Float vs int sentinel branching: integer sources with a sentinel promote to float64; float sources mask in place.
- MinIsWhite inversion still happens before finalization.
- GPU paths return CuPy-backed DataArrays.

## Tests

Add `xrspatial/geotiff/tests/test_eager_finalization_parity_2162.py`. Parametrize over `numpy` and `cupy` (eager GPU). Assert:

- Both backends produce the same attrs dict (modulo array backend) for matched inputs.
- `nodata_pixels_present` matches as a bool across both.
- `nodata_dtype_cast` matches.
- Float-vs-int sentinel handling matches.
- `georef_status` matches.

Skip GPU tests when CUDA is unavailable.

## Files

- `xrspatial/geotiff/__init__.py` (rewrite 711-867 to call `_finalize_eager_read`)
[...]

Issue #2178, opened 2026-05-20
GeoTIFF: migrate dask backends to shared lazy finalization helper (PR C of #2162)


brendancol:
Parent: #2162
Depends on: #2177 (adds the lazy helper)

## Goal

Replace duplicated finalization in the two dask paths with `_finalize_lazy_read_attrs` from #2177.

## Scope

Two call sites, each about 25 lines of nearly identical validate then populate then set-nodata-with-None code:

1. `xrspatial/geotiff/_backends/dask.py:359-386` — CPU dask finalization in `read_geotiff_dask`.
2. `xrspatial/geotiff/_backends/gpu.py:1530-1553` — GPU+dask finalization in the dask branch of `read_geotiff_gpu`.

Both sites construct attrs, call `_validate_read_geo_info`, call `_populate_attrs_from_geo_info`, and call `_set_nodata_attrs` with `pixels_present=None`. After the migration each site is one call to `_finalize_lazy_read_attrs(...)` followed by the backend-specific dask graph assembly.

Preserve all current behavior:

- `attrs['nodata_pixels_present']` remains absent (None) in lazy outputs.
- `attrs['nodata_dtype_cast']` is recorded when the caller forced a cast.
- MinIsWhite inversion (at `dask.py:227-238`) still preempts the sentinel before the chunk tasks see it.

## Tests

Add `xrspatial/geotiff/tests/test_lazy_finalization_parity_2162.py`. Parametrize over `dask+numpy` and `dask+cupy`. Assert:

- Both backends produce the same attrs dict (modulo array backend markers) for matched inputs.
- `nodata_pixels_present` is absent in both.
- `nodata_dtype_cast` matches across both when the caller forces a cast.
- `georef_status` matches across both for the five georef states (full, transform_only, crs_only, none, rotated_dropped).

Skip GPU tests when CUDA is unavailable.

## Files

- `xrspatial/geotiff/_backends/dask.py` (rewrite 359-386 to call `_finalize_lazy_read_attrs`)
- `xrspatial/geotiff/_backends/gpu.py` (rewrite 1530-1553 to call `_finalize_lazy_read_attrs`)
- `xrspatial/geotiff/tests/test_lazy_finalization_parity_2162.py` (new)

## Constraints

- No public API change.
- Existing dask tests must still pass: `test_nodata_lifecycle_attrs_2135.py`, `test_nodata_semantics_split_1988.py`, `test_georef_status_2136.py`.
[...]

Issue #2175, opened 2026-05-20
GeoTIFF: centralize public read-argument validation across backends (PR A of #2162)


brendancol:
Parent: #2162

## Goal

Add a shared validator that `open_geotiff` and every direct backend (`read_geotiff_dask`, `read_geotiff_gpu`, `read_vrt`) call up front, so identical inputs produce identical errors no matter which entry point you use.

Today `open_geotiff` validates `overview_level`, `on_gpu_failure`, `missing_sources`, `band_nodata`, `max_cloud_bytes`, and file-like source restrictions inline at `xrspatial/geotiff/__init__.py:508-584`. Direct backends each call `_validate_overview_level_arg` individually but skip the rest. Direct callers can pass an invalid `band_nodata` to dask, or `max_cloud_bytes` to GPU, and they will not see the error `open_geotiff` would raise.

## Scope

Add `_validate_dispatch_kwargs(...)` to `xrspatial/geotiff/_validation.py`. It validates:

- `overview_level` (delegates to existing `_validate_overview_level_arg`)
- `on_gpu_failure` (must be `auto` or `strict`; raises when GPU is off)
- `missing_sources` (VRT only)
- `band_nodata` (VRT only)
- `max_cloud_bytes` (incompatible with VRT/GPU/dask per current policy)
- file-like source restrictions for `gpu=True` and `chunks=...`

Wire it in:

- `open_geotiff` at `__init__.py:508-584` replaces the inline checks with one call.
- `read_geotiff_dask`, `read_geotiff_gpu`, and `read_vrt` each gain one call near the top of their bodies, next to the existing `_validate_overview_level_arg` calls at `dask.py:116` and `gpu.py:204`.

Keep error messages the same as today's inline checks.

## Tests

Add `xrspatial/geotiff/tests/test_dispatch_validation_parity_2162.py`. Parametrize over `open_geotiff` and the three direct backends. Prove that each invalid input raises the same exception type and a recognizable message regardless of entry point:

- Invalid `overview_level` (`bool`, `str`, `float`) through all four entry points.
- `max_cloud_bytes` with dask, GPU, and VRT.
- `missing_sources` on non-VRT sources.
- `band_nodata` on non-VRT sources.
- `on_gpu_failure` when GPU is disabled.

## Files

- `xrspatial/geotiff/_validation.py` (add helper)
- `xrspatial/geotiff/__init__.py` (replace 508-584 with helper call)
[...]

PR #2200, opened 2026-05-20
geotiff: add read-finalization helpers (PR B of #2162)

### LINKED ISSUE #2177: GeoTIFF: add shared read-finalization helpers (PR B of #2162)
Parent: #2162
Sibling in wave 1: #2175

## Goal

Add two private helpers in `xrspatial/geotiff/_attrs.py` that capture the read-finalization pipelines duplicated across backends. This issue only adds the helpers and unit-tests them in isolation. No backend gets migrated here. That happens in waves 2 (#2178 dask, #2179 eager) and 3 (#2180 VRT).

Adding helpers first lets all downstream migration PRs depend on a frozen signature instead of inventing slightly different shapes.

### `_finalize_eager_read`

\`\`\`python
def _finalize_eager_read(
    arr,
    *,
    geo_info,
    nodata,
    mask_sentinel,
    mask_nodata,
    dtype,
    window,
    name,
    allow_rotated=False,
    allow_unparseable_crs=False,
    attrs_in=None,
):
    """Validate, populate attrs, mask, cast, and build an eager DataArray."""
\`\`\`

Pipeline:

1. \`_validate_read_geo_info(geo_info, window=window, allow_rotated=..., allow_unparseable_crs=...)\`
2. \`_populate_attrs_from_geo_info(attrs, geo_info, window=window)\`
3. Mask \`arr\` using \`mask_sentinel\` when \`mask_nodata\` is True. Record \`nodata_pixels_present\` as a bool.
4. Cast dtype; record \`nodata_dtype_cast\`.
5. \`_set_nodata_attrs(attrs, nodata, masked=..., pixels_present=..., dtype_cast=...)\`
6. Return \`xarray.DataArray(...)\` with coords from the helper.

Why \`mask_sentinel\` is a parameter: the three GPU eager sites derive it three different ways (\`_mw_mask_nodata\` local, \`_cpu_fallback_geo._mask_nodata\`, raw \`nodata\`). The helper must not read it off \`geo_info\`.

### \`_finalize_lazy_read_attrs\`

\`\`\`python
def _finalize_lazy_read_attrs(
    *,
    geo_info,
    nodata,
    mask_nodata,
    dtype,
    window,
    allow_rotated=False,
    allow_unparseable_crs=False,
    attrs_in=None,
):
    """Validate and populate attrs for dask-style lazy reads.
    Returns the attrs dict. Does NOT build a DataArray (the caller
    assembles the dask graph)."""
\`\`\`

Pipeline:

1. \`_validate_read_geo_info(...)\`
2. \`_populate_attrs_from_geo_info(attrs, geo_info, window=window)\`
3. \`_set_nodata_attrs(attrs, nodata, masked=(mask_nodata and float dtype), pixels_present=None, dtype_cast=...)\`
4. Return \`attrs\`.

\`pixels_present=None\` is the documented dask contract from #2135. The signature accepts None explicitly. Computing pixels-present eagerly here would force \`.compute()\` and break the lazy contract.

## Tests

Add \`xrspatial/geotiff/tests/test_finalization_helpers_2162.py\`. Synthesize \`GeoInfo\` fixtures and assert:

- Eager helper populates \`attrs['nodata']\`, \`attrs['nodata_pixels_present']\`, \`attrs['nodata_dtype_cast']\`, and \`attrs['georef_status']\` correctly across float and int input dtypes.
- Eager helper respects \`mask_nodata=False\` (no masking, no \`nodata_pixels_present\`).
- Lazy helper populates the same attrs minus \`nodata_pixels_present\`.
- Both helpers propagate \`_validate_read_geo_info\` rejections for ambiguous geo_info.
- Both helpers handle \`mask_sentinel != nodata\` (the GPU minIsWhite inversion case).

## Files

- \`xrspatial/geotiff/_attrs.py\` (add two helpers)
- \`xrspatial/geotiff/tests/test_finalization_helpers_2162.py\` (new)

## Constraints

- No public API change.
- Helpers are private (underscore prefix).
- The eager helper returns a \`DataArray\`; the lazy helper returns an attrs dict. Wave 2 and wave 3 callers depend on these shapes.
- \`_validate_read_geo_info\` runs first. Partial attrs must not leak when validation fails.

## Out of scope

- Migrating any backend to call these helpers (waves 2 and 3).
- Changes to the existing shared helpers (\`_validate_read_geo_info\`, \`_populate_attrs_from_geo_info\`, \`_set_nodata_attrs\`).

brendancol:
## Summary

Closes #2177. Wave 1 of #2162. Sibling wave 1 PR #2175 is in progress against disjoint files.

Adds two private helpers in `xrspatial/geotiff/_attrs.py` for the read-finalization pipeline that is duplicated across the four read backends. No call sites are migrated here; that work belongs to waves 2 and 3 (#2178, #2179, #2180). The helpers are dead code until those PRs land.

- `_finalize_eager_read(arr, *, geo_info, nodata, mask_sentinel, mask_nodata, dtype, window, name, ...)` validates `geo_info`, populates attrs, masks sentinel pixels, casts dtype, sets the nodata lifecycle attrs (`pixels_present` as bool), and returns the `xarray.DataArray`. `mask_sentinel` is a separate parameter because the three GPU eager sites derive it three different ways (MinIsWhite inversion, CPU-fallback `_mask_nodata`, raw `nodata`).
- `_finalize_lazy_read_attrs(*, geo_info, nodata, mask_nodata, dtype, window, ...)` validates `geo_info`, populates attrs, sets nodata attrs with `pixels_present=None` per the documented dask contract from #2135. Returns the attrs dict only; the caller assembles the dask graph and builds the DataArray.
- `_validate_read_geo_info` runs first in both helpers, so partial attrs cannot leak when validation fails.

## Test plan

`xrspatial/geotiff/tests/test_finalization_helpers_2162.py` synthesises `GeoInfo` fixtures and exercises the helpers in isolation:

- [x] `nodata`, `nodata_pixels_present`, `nodata_dtype_cast`, `masked_nodata`, and `georef_status` get set correctly on the eager helper for float and int input dtypes.
- [x] `mask_nodata=False` skips masking but still surfaces `pixels_present`.
- [x] `mask_sentinel != nodata` (GPU MinIsWhite inversion case) routes through the masking step.
- [x] Lazy helper produces the same attrs minus `nodata_pixels_present`.
- [x] Both helpers raise on ambiguous geo_info before any attrs are written; the caller's seed dict stays untouched.
- [x] `allow_unparseable_crs=True` bypasses the validator on both helpers.
- [x] Signature pinning: eager helper takes `mask_sentinel`, lazy helper deliberately does not.

21 tests pass locally. Existing `_attrs` / `nodata_lifecycle` / `georef_status` tests still pass.

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -160,6 +160,8 @@ from typing import Any
 
 import numpy as np
 
+import xarray as xr
+
 from ._coords import (
     coords_from_geo_info as _coords_from_geo_info,
     transform_tuple_from_pixel_geometry as _transform_tuple_from_pixel_geometry,
@@ -1250,18 +1252,30 @@ def _finalize_eager_read(
     sources). Read paths that don't need MinIsWhite inversion can pass
     ``mask_sentinel=nodata``.
 
-    Numpy buffers only; the GPU eager call sites pass cupy ndarrays and
-    rely on cupy's numpy-equivalent operators (``arr.dtype.kind``,
-    ``arr.astype``, boolean fancy indexing, ``arr == scalar``) so the
-    same helper covers both backends. The GPU eager path also passes a
-    masking function that runs on the device; this helper handles the
-    host-side path. (Wave 2 will migrate eager numpy / VRT eager; wave 3
-    will migrate the three GPU eager sites with the mask step factored
-    out.)
+    Wave migration plan:
+
+    * Wave 2 (#2178 dask, #2179 eager numpy) migrates the eager numpy
+      paths. The mask block inside this helper matches the inline block
+      in ``open_geotiff`` field-for-field; the migration is a straight
+      swap.
+    * Wave 3 (#2180 VRT, GPU) migrates the VRT eager + three GPU eager
+      sites. The VRT eager path is host-side and works with the helper
+      as-is. The GPU sites apply masking through a CUDA kernel
+      (``_apply_nodata_mask_gpu_with_presence``); they can either
+      pre-mask and call the helper with ``nodata=None`` to skip the
+      helper's host-side mask block, or wave 3 can extend this
+      helper's signature with a ``mask_fn`` injection. Either choice
+      lives in #2180; the wave 1 contract here is the host-side path.
 
     Returns a :class:`xarray.DataArray` ready for the caller to return
     from the read function. The caller assembles the dask graph
     separately when a lazy backend is in play; this helper is eager-only.
+
+    ``attrs_in`` is shallow-copied via ``dict(attrs_in)``. Nested values
+    (e.g. ``extra_tags`` list, ``gdal_metadata`` dict) are shared between
+    the caller's seed dict and the returned DataArray's attrs; mutating
+    a nested value after the call propagates both ways. Callers that
+    care about isolation can ``copy.deepcopy(attrs_in)`` first.
     """
     # Step 1: validate first so partial attrs never leak.
     _validate_read_geo_info(
@@ -1322,10 +1336,6 @@ def _finalize_eager_read(
     else:
         dims = ['y', 'x']
 
-    # Local import to avoid a top-level ``import xarray`` -- ``_attrs``
-    # is imported from ``__init__`` and the lazy import keeps the module
-    # load graph the same shape it already has.
-    import xarray as xr
     return xr.DataArray(arr, dims=dims, coords=coords, name=name, attrs=attrs)
 
 
@@ -1365,13 +1375,27 @@ def _finalize_lazy_read_attrs(
     deliberately does not touch arrays or coords.
 
     The ``dtype`` parameter accepts a numpy dtype, a string ('float64'),
-    or ``None`` (no explicit cast). When ``None`` the masked-flag still
-    has to know the graph dtype; the caller passes whatever ``dtype=``
-    they resolved onto the graph (e.g. ``target_dtype`` in the dask
-    backend) so the helper can derive ``masked``. The integer-to-float
-    promotion that an eager path applies when sentinel pixels are
-    present is handled by the dask backend before it calls this helper;
-    ``dtype`` here is the resolved graph dtype, not the file dtype.
+    or ``None``. It is the **resolved graph dtype** the dask backend
+    settled on (e.g. ``target_dtype`` after the int->float64 promotion
+    that ``mask_nodata=True`` triggers on int files): the helper uses
+    it to derive ``masked`` and writes it as ``nodata_dtype_cast`` when
+    non-None.
+
+    Wave 2 migration note: the current pre-helper dask backend
+    distinguishes "caller explicitly passed ``dtype=``" from
+    "graph dtype was auto-promoted by masking" so that
+    ``nodata_dtype_cast`` surfaces only on the explicit-cast case.
+    This helper conflates the two -- whatever ``dtype`` value the
+    caller passes here becomes the ``nodata_dtype_cast`` attr. The
+    migration PR (#2178) can either accept that change, or split the
+    helper's ``dtype`` into two parameters at that point. Frozen
+    signature here per #2177 means we ship the one-``dtype`` shape
+    and leave the split for wave 2 if it turns out to matter.
+
+    ``attrs_in`` is shallow-copied via ``dict(attrs_in)``. Nested values
+    are shared between the caller's seed dict and the returned attrs;
+    callers that care about isolation can ``copy.deepcopy(attrs_in)``
+    first.
     """
     _validate_read_geo_info(
         geo_info, window=window,
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -161,6 +161,7 @@ from typing import Any
 import numpy as np
 
 from ._coords import (
+    coords_from_geo_info as _coords_from_geo_info,
     transform_tuple_from_pixel_geometry as _transform_tuple_from_pixel_geometry,
 )
 from ._geotags import (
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: geotiff: add read-finalization helpers (PR B of #2162)

Wave 1 of #2162. Adds two private helpers in `xrspatial/geotiff/_attrs.py` plus a 21-test exercise file. No call sites are migrated here. Reviewed against the issue body contract.

None.

- [ ] `xrspatial/geotiff/_attrs.py:1276`: `attrs_in` is shallow-copied via `dict(attrs_in)`. If a caller passes a seed dict that contains a list or dict value (e.g. `extra_tags`) and later mutates that nested value, the change leaks onto the returned DataArray's attrs. Wave 2 will probably pass freshly-built dicts so this isn't likely to bite in practice, but a docstring note ("shallow copy; nested values are shared") would document the gotcha for future migrators.

- [ ] `xrspatial/geotiff/_attrs.py:1392-1395`: the lazy helper's `dtype` parameter does double duty. It's the graph dtype (for `masked` derivation) AND the value recorded as `nodata_dtype_cast`. The existing dask backend keeps those separate (`target_dtype` for masked, raw caller `dtype=` for cast). The docstring now flags this as a wave-2 migration concern, which is the right call given the frozen-signature constraint. Worth pinning a test that documents which interpretation we shipped so wave 2 catches any drift early. The current `test_lazy_int_graph_dtype_keeps_masked_false` exercises the int-dtype path but doesn't assert the dtype_cast attr; adding `assert attrs['nodata_dtype_cast'] == 'int16'` to that test would lock in the conflated semantics.

- [ ] `xrspatial/geotiff/_attrs.py:1267`: the `_validate_read_geo_info(...)` block is identical between the two helpers. Could be extracted into one private call shared by both, but the duplication is small (4 lines x 2) and pulling it out would obscure the validate-first contract. Leave as is.

- [ ] `xrspatial/geotiff/_attrs.py:1328`: the `import xarray as xr` inside `_finalize_eager_read` runs every call. `xarray` is already imported at module-load time in `__init__.py`, so a top-level `import xarray as xr` in `_attrs.py` would be free. The local import is defensible (matches `_validate_dtype_cast`'s pattern), but xarray is already a hard dep of the package.

- [ ] `xrspatial/geotiff/tests/test_finalization_helpers_2162.py:316`: the seed-dict-untouched assertion uses `seed == {'sentinel_marker': True}` which is correct, but a stronger check is `'sentinel_marker' in seed and len(seed) == 1` to catch the case where the validator partially populates and then re-raises (currently impossible but defensive).

### What looks good

- Validate-first ordering is explicit and tested. The seed-dict-untouched test catches the partial-attrs leak the issue body called out.
- `mask_sentinel != nodata` (the GPU MinIsWhite case) is tested directly with a fixture that diverges the two values.
- Eager helper handles int -> float64 promotion when the sentinel matches, matching the existing inline block field-for-field.
- The lazy helper's `pixels_present=None` is documented as the dask contract from #2135 and tested via the absent-from-attrs assertion.
- Signature pinning tests catch the case where a future refactor adds `mask_sentinel` to the lazy helper or drops it from the eager helper.
- Docstrings call out the wave-2 / wave-3 migration plans so the next PR doesn't have to re-derive the design.

### Checklist
- [x] Helper signatures match the issue body's contract.
- [x] `_validate_read_geo_info` runs first in both helpers.
- [x] Partial attrs do not leak on validation failure (tested).
- [x] Eager helper covers float + int input dtypes.
- [x] `mask_nodata=False` opt-out tested.
- [x] `mask_sentinel != nodata` (GPU MinIsWhite) tested.
- [x] Lazy helper omits `nodata_pixels_present` per #2135.
- [x] Tests pass locally (21/21).
- [x] Existing `_attrs` / `nodata_lifecycle` / `georef_status` tests still pass.

brendancol (commented):
## PR Review: follow-up pass (PR B of #2162)

Second pass after the initial review findings were addressed. Diff is +54 / -20 across the helper module and the test file.

### Blockers
None.

### Suggestions
None new. Initial-pass suggestions were applied:

- `attrs_in` shallow-copy semantics documented on both helpers (`_attrs.py:1265-1276`, `_attrs.py:1389-1405`).
- `test_lazy_int_graph_dtype_keeps_masked_false` now asserts `nodata_dtype_cast == 'int16'` to lock in the wave-1 conflated-dtype contract.

### Nits
None new. Initial-pass nits were applied where actionable:

- `import xarray as xr` now lives at module scope.
- Both seed-dict-untouched assertions now also check `len(seed) == 1`.
- The dedup nit on the `_validate_read_geo_info` block was kept as-is (the initial review itself flagged "leave as is" for that one).

### What looks good in this pass

- Diff is minimal and targeted, no scope creep.
- All 21 helper tests still pass. 99 related tests (`_attrs`, `nodata_lifecycle`, `georef_status`, `attrs_contract`) also still pass with the module-level xarray import.
- Docstring notes on `attrs_in` shallow-copy are short and concrete; they call out the wave-2 / wave-3 caller responsibility without hedging.

### Checklist (carry-over from initial pass)
- [x] Helper signatures match the issue body's contract.
- [x] `_validate_read_geo_info` runs first in both helpers.
- [x] Partial attrs do not leak on validation failure (tested with both equality and length checks).
- [x] Eager helper covers float + int input dtypes.
- [x] `mask_nodata=False` opt-out tested.
- [x] `mask_sentinel != nodata` (GPU MinIsWhite) tested.
- [x] Lazy helper omits `nodata_pixels_present` per #2135.
- [x] Conflated `dtype` semantics on the lazy helper pinned with an explicit assertion.
- [x] Tests pass locally.
- [x] Existing tests still pass.

Ready for merge once CI is green.

Issue #2162, opened 2026-05-20
Refactor GeoTIFF read finalization and dispatch validation to prevent backend drift


brendancol:
# Refactor GeoTIFF Read Finalization and Dispatch Validation

## Summary
Refactor the GeoTIFF read stack so shared policy lives in shared helpers instead of being hand-copied across `open_geotiff`, dask, GPU, and VRT backends. The goal is to prevent backend drift in nodata masking, attrs, dtype casting, georef validation, and public kwarg validation.

## Key Changes
- Add a shared read-finalization helper for eager array backends that handles:
  - read metadata validation
  - coord/dim construction
  - attrs population
  - nodata masking and `nodata_pixels_present`
  - dtype casting and `nodata_dtype_cast`
  - final `xarray.DataArray` construction
- Add a shared lazy-read attrs helper for dask-style backends that centralizes:
  - `_validate_read_geo_info`
  - `_populate_attrs_from_geo_info`
  - `_set_nodata_attrs`
  - the dask policy that `nodata_pixels_present` stays unset
- Add shared public read-argument validation for:
  - `overview_level`
  - backend-only kwargs such as `on_gpu_failure`, `missing_sources`, `band_nodata`, and `max_cloud_bytes`
  - file-like source restrictions for GPU/dask
- Update `open_geotiff` to become mostly dispatch orchestration:
  - validate args once
  - choose VRT/GPU/dask/eager backend
  - forward normalized kwargs
  - avoid duplicating backend finalization logic
- Update direct public backends (`read_geotiff_dask`, `read_geotiff_gpu`, `read_vrt`) to call the same relevant validators so direct calls and `open_geotiff` calls reject the same invalid inputs.

## Implementation Notes
- Keep public APIs unchanged.
- Prefer adding private helpers in existing internal modules rather than expanding `__init__.py`.
- Do not change current documented dask behavior: lazy outputs should not compute `nodata_pixels_present`.
- Treat VRT masking drift as a follow-up bug unless it is cheap to fix during extraction; this refactor's core goal is to make that kind of drift harder to reintroduce.
- Move comments that document policy to the shared helper once; remove duplicated issue-history comments from backend bodies where the helper name/test now carries the intent.

## Test Plan
- Add parity tests proving invalid `overview_level` values fail through:
  - `open_geotiff`
  - `read_geotiff_dask`
  - `read_geotiff_gpu`
- Add backend parity tests for nodata lifecycle attrs:
  - eager numpy
[...]

Issue #1987, opened 2026-05-15
geotiff: fail closed on ambiguous geospatial metadata (CRS, transform, coords, nodata)


brendancol:
## Reason or Problem

The reader has tightened over time (`test_crs_fail_closed_1929.py`, `test_crs_fail_closed_gpu_1929.py`, `test_coord_regularity_1720.py`, `test_rotated_transform_attr_1764.py`) but several "guess and continue" paths remain. The reviewer's position, which I agree with: for foundational I/O, refusing ambiguous input is better than manufacturing a plausible-looking raster.

Specific cases still loose:

- Invalid CRS codes — `to_geotiff(crs=True)` writes EPSG=1 silently (already filed as #1971).
- Unparseable CRS strings — partial WKT may be accepted and emitted unchanged on read, leading to mismatched `crs` vs `crs_wkt` attrs.
- Rotated transforms — currently flagged in attrs but the reader does not reject; downstream functions assume axis-aligned and may produce wrong results.
- Non-uniform coords on write — `to_geotiff` accepts coords that disagree with the implied transform; sentinel checks exist but coverage is uneven.
- Mixed band metadata — VRT inputs with bands that declare different nodata sentinels are silently flattened.
- Conflicting `crs` and `crs_wkt` on write — no consistency check.
- Conflicting nodata aliases — `attrs['nodata']` set to one value and `attrs['nodatavals']` set to another. `_resolve_nodata_attr` picks one and ignores the other.

## Proposal

Each of the above becomes a hard error by default, with an opt-in escape hatch where the legacy behavior was documented.

**Design:**

Add a validation layer in `_validation.py` (which already exists) and call it from every read and write entry point. The validator raises with a specific error class per case (subclasses of a new `GeoTIFFAmbiguousMetadataError`). Each case gets its own test in `test_crs_fail_closed_*` style — one new test file per case is fine, but prefer adding to existing fail-closed files where the topic matches.

Per-case rough plan:

- **invalid CRS codes** — covered by #1971, link from this issue.
- **unparseable CRS strings** — raise on read, raise on write. No silent passthrough.
- **rotated transforms** — read raises by default; opt-in `allow_rotated=True` keeps existing attr-flag behavior for users who want the array without geo-aware downstream ops.
- **non-uniform coords** — write raises; the existing sentinel exemption for int-dtype coords stays (#1969).
- **mixed band metadata in VRT** — read raises; opt-in `band_nodata='first'` retains current.
[...]

PR #2201, opened 2026-05-20
geotiff: share dispatch-kwarg validation across read entry points (#2175)

### LINKED ISSUE #2175: GeoTIFF: centralize public read-argument validation across backends (PR A of #2162)

brendancol:
## Summary

Closes #2175. Parent #2162 (wave 1 of 3).

`open_geotiff` used to run six dispatcher-level kwarg checks inline at `xrspatial/geotiff/__init__.py:508-584`. The three direct backends (`read_geotiff_dask`, `read_geotiff_gpu`, `read_vrt`) only ran the `_validate_overview_level_arg` slice and silently dropped the rest. A caller who passed `band_nodata` to dask or `max_cloud_bytes` to the GPU reader got no error at all, or got an unrelated `TypeError` from the signature.

This PR adds `_validate_dispatch_kwargs(...)` to `xrspatial/geotiff/_validation.py` and wires it into every public read entry point so the same invalid input raises the same exception text regardless of which entry point the caller used.

### Rules enforced (all pre-existing in `open_geotiff`; now mirrored on the direct backends):

- `overview_level` type check (delegates to existing helper, issue #2074).
- `on_gpu_failure` rejected on the CPU / dask / VRT paths (issue #1615).
- `missing_sources` rejected on non-VRT sources (issue #1810).
- `band_nodata` rejected on non-VRT sources (issue #1987).
- `max_cloud_bytes` rejected on VRT / GPU / dask paths (issue #1974).
- File-like sources reject `gpu=True` / `chunks=...`.

### API surface

No behavioral change for any valid input. Direct backends gain new kwargs with sentinel defaults (`_ON_GPU_FAILURE_SENTINEL`, `_MISSING_SOURCES_SENTINEL`, `_MAX_CLOUD_BYTES_SENTINEL`) so that omitted kwargs round-trip through the helper without tripping a rejection. Explicit invalid values now raise consistently across `open_geotiff`, `read_geotiff_dask`, `read_geotiff_gpu`, and `read_vrt`.

### Backend coverage

All four read entry points share the same validator. Pure CPU-side validation; no kernels or device code.

## Test plan

- [x] New `xrspatial/geotiff/tests/test_dispatch_validation_parity_2162.py` parametrizes the matrix from issue #2175 (invalid `overview_level` types through all four entry points; `max_cloud_bytes` with dask/GPU/VRT; `missing_sources` on non-VRT; `band_nodata` on non-VRT; `on_gpu_failure` without GPU; file-like + `gpu=True`/`chunks=...`).
- [x] Existing dispatcher tests (`test_overview_level_validation_backends_2160.py`, `test_max_cloud_bytes_dispatcher_silent_drop_2026_05_15.py`, `test_open_geotiff_missing_sources_1810.py`, `test_open_geotiff_on_gpu_failure_1615.py`) still pass.
- [x] Signature-order pin (`test_reader_kwarg_order_1935.py`) still passes after adding new kwargs.
- [x] Full `xrspatial/geotiff/tests/` suite green (4552 passed, 45 skipped, 1 xfailed). The pre-existing `lz4` failure on `test_lowlevel_write_pushdown_2138.py` reproduces on `main` and is unrelated.

## Scope notes

- Wave 1 of the #2162 refactor. Out of scope: finalization helpers (sibling PR in this wave), backend migration to a shared finalization pipeline (waves 2 and 3).
- The VRT-specific value rejection (`overview_level not in (None, 0)` and the legacy "on_gpu_failure on VRT" message) stays inside the VRT branch of `open_geotiff` and at the top of `read_vrt`. Folding those into the helper would expand scope beyond the issue's contract.
- GPU-only test paths are skipped when CUDA is unavailable; the bad-input tests run on CPU-only hosts because the validator fires before any CUDA import.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -548,6 +548,11 @@ def open_geotiff(source: str | BinaryIO, *,
         # ``overview_level=0`` is documented as "full resolution" (the
         # default), so treat it as a no-op the same as ``None`` rather
         # than rejecting a kwarg value the caller could have omitted.
+        # Mirrored at ``_backends/vrt.py`` (the direct-call entry point)
+        # so callers see the same rejection through both paths. The
+        # value-level rejection stays here rather than in
+        # ``_validate_dispatch_kwargs`` so the helper does not need a
+        # special-case for ``overview_level=0`` vs ``overview_level=N``.
         if overview_level not in (None, 0):
             raise ValueError(
                 "overview_level is not supported for VRT sources. "
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -509,84 +509,33 @@ def open_geotiff(source: str | BinaryIO, *,
 
     source = _coerce_path(source)
 
-    # Reject bool and non-int ``overview_level`` up front (issue #2074).
-    # Without this guard, ``overview_level=True`` is coerced to ``1`` and
-    # silently returns the first overview level, and non-int types leak
-    # raw ``TypeError`` messages from the internal numeric comparison or
-    # list indexing.
-    from ._validation import _validate_overview_level_arg
-    _validate_overview_level_arg(overview_level)
+    # All dispatcher-level kwarg rejection lives in
+    # ``_validate_dispatch_kwargs`` so the three direct backends
+    # (``read_geotiff_dask``, ``read_geotiff_gpu``, ``read_vrt``)
+    # surface the same errors when called directly (issue #2175,
+    # parent #2162). The previous inline block at this line is now
+    # a single call that runs ``_validate_overview_level_arg`` (issue
+    # #2074), the ``on_gpu_failure`` GPU-only guard (issue #1615),
+    # the ``missing_sources`` VRT-only guard (issue #1810), the
+    # ``band_nodata`` VRT-only guard (issue #1987), the
+    # ``max_cloud_bytes`` non-VRT non-GPU non-dask guard (issue #1974),
+    # and the file-like source restrictions for gpu/chunks.
+    from ._validation import _validate_dispatch_kwargs
+    _validate_dispatch_kwargs(
+        source=source,
+        gpu=gpu,
+        chunks=chunks,
+        overview_level=overview_level,
+        on_gpu_failure=on_gpu_failure,
+        missing_sources=missing_sources,
+        band_nodata=band_nodata,
+        max_cloud_bytes=max_cloud_bytes,
+    )
 
-    # ``on_gpu_failure`` is GPU-only. Reject it up front for CPU/dask paths
-    # rather than silently dropping it once dispatch is decided -- callers
-    # otherwise have no way to learn that the policy is being ignored.
-    # ``gpu=False`` (the default) on a ``.vrt`` source still routes through
-    # ``read_vrt`` below which has no GPU-failure concept, so the same
-    # rejection rule applies there.
-    if on_gpu_failure is not _ON_GPU_FAILURE_SENTINEL and not gpu:
-        raise ValueError(
-            "on_gpu_failure only applies when gpu=True. "
-            "Pass gpu=True to enable the GPU pipeline, or drop "
-            "on_gpu_failure to keep the default CPU path.")
-
-    # ``missing_sources`` is VRT-only. Reject it up front when the source
-    # is not a ``.vrt`` file so callers learn the policy is being ignored
-    # instead of getting a silent drop -- same pattern ``on_gpu_failure``
-    # uses above for the GPU-only kwarg, and the same class of dispatcher
-    # silently-drops-backend-kwarg bug #1561 / #1605 / #1685 / #1795 fixed
-    # for the other VRT/GPU kwargs. See issue #1810.
     missing_sources_passed = (
         missing_sources is not _MISSING_SOURCES_SENTINEL)
     _is_vrt_source = (
         isinstance(source, str) and source.lower().endswith('.vrt'))
-    if missing_sources_passed and not _is_vrt_source:
-        raise ValueError(
-            "missing_sources only applies to VRT sources. "
-            "Pass a .vrt path to enable the VRT pipeline, or drop "
-            "missing_sources to keep the default GeoTIFF path.")
-
-    # ``band_nodata`` is the #1987 PR 5 opt-out for the mixed-band
-    # metadata fail-closed check. It only has meaning on the VRT pipeline
-    # (a plain GeoTIFF has one nodata sentinel per file, not per band),
-    # so reject the kwarg up front on non-VRT sources rather than letting
-    # it leak into ``read_vrt`` and confuse the caller about what the
-    # opt-out actually controls.
-    if band_nodata is not None and not _is_vrt_source:
-        raise ValueError(
-            "band_nodata only applies to VRT sources. "
-            "Pass a .vrt path to enable the VRT pipeline, or drop "
-            "band_nodata to keep the default GeoTIFF path. "
-            "See issue #1987.")
-
-    # ``max_cloud_bytes`` is the eager fsspec-read budget. Only
-    # ``_read_to_array`` on the eager non-VRT, non-GPU, non-dask branch
-    # consumes it; the GPU (``read_geotiff_gpu``), dask
-    # (``read_geotiff_dask``), and VRT (``read_vrt``) branches all ignore
-    # the kwarg silently. Reject it up front on those paths so callers
-    # learn the budget is being dropped, matching the
-    # ``on_gpu_failure`` / ``missing_sources`` guards above and the
-    # silently-drops-backend-kwarg fixes in #1561 / #1605 / #1685 / #1810.
-    # See issue #1974.
-    if max_cloud_bytes is not _MAX_CLOUD_BYTES_SENTINEL:
-        if _is_vrt_source:
-            raise ValueError(
-                "max_cloud_bytes is not supported for VRT sources. "
-                "The VRT reader does not apply the cloud-byte budget; "
-                "drop the kwarg, or call open_geotiff on the underlying "
-                ".tif source.")
-        if gpu:
-            raise ValueError(
-                "max_cloud_bytes is not supported when gpu=True. "
-                "The GPU reader does not accept the kwarg directly; "
-                "for URL/fsspec sources the budget is honoured via "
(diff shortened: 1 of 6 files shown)


brendancol (commented):
## PR Review: geotiff: share dispatch-kwarg validation across read entry points (#2175)

- [ ] `xrspatial/geotiff/_backends/gpu.py:214` calls `_validate_dispatch_kwargs(source=source, ...)` before the `_coerce_path(source)` call at line 304. A caller passing `pathlib.Path('foo.tif')` to `read_geotiff_gpu` now raises `ValueError: gpu=True is not supported for file-like sources.` because the helper's file-like guard (`not isinstance(source, str)`) treats the Path as a file-like. Reproduced locally on this branch. On `main`, `read_geotiff_gpu(Path('foo.tif'))` returns the array. The same flow in `dask.py:116` and `vrt.py:176` coerces first; the GPU entry point needs the same. Either move `_coerce_path` above the validator call in `gpu.py`, or coerce inside the helper.

- [ ] `xrspatial/geotiff/_validation.py:494` (the file-like guard in `_validate_dispatch_kwargs`) is the only block in the helper that runs after the VRT-detection-dependent checks. The current ordering means `read_geotiff_dask(BytesIO(...), missing_sources='raise')` raises the `missing_sources only applies to VRT sources` error before the more descriptive `chunks=... is not supported for file-like sources` rejection. Both messages are correct, but the file-like message names the actual blocker more clearly. Consider moving the file-like guard before the VRT-dependent checks.

- [ ] `xrspatial/geotiff/_backends/vrt.py:206` -- `overview_level not in (None, 0)` still lives outside the helper. It is VRT-specific (the value rejection, not the type check) so it is reasonable to keep it inline, but the symmetric rejection inside `open_geotiff.__init__.py:550-555` is now duplicated. Either move both into the helper, or add an inline comment cross-referencing the other site so a future refactor does not drop one of them.

- [ ] `xrspatial/geotiff/_validation.py:425` -- the local `from ._reader import _MAX_CLOUD_BYTES_SENTINEL` lookup is documented as a circular-import workaround. The same module already imports from `._runtime` at module load (`_MISSING_SOURCES_SENTINEL`, `_ON_GPU_FAILURE_SENTINEL`). If the cloud-byte sentinel can move to `_runtime.py` in a follow-up, the local import goes away. Out of scope here, but worth a one-line `TODO` comment so the next pass picks it up.

- [ ] `xrspatial/geotiff/tests/test_dispatch_validation_parity_2162.py:271` -- `_get_error_message` returns `f"{type(e).__name__}: {e}"` and the parity tests assert substrings against it. Asserting against `e.args[0]` (or the formatted message alone, without the type prefix) would catch a regression where the exception type changes silently while the message stays the same. Minor; the current asserts are sufficient.

- [ ] `xrspatial/geotiff/_backends/dask.py:155-165` -- the local `vrt_kwargs = {}` shim mirrors what `open_geotiff` does at line 568-578. With the helper now centralizing rejection, the two sites are mostly bookkeeping. Could fold into a small `_forward_to_read_vrt(**kwargs)` helper in a follow-up.

### What looks good

- Helper docstring lists every rule with the issue number for the original guard. Future readers can trace each rejection back to its motivation.
- Sentinel defaults preserve "caller did not pass this kwarg" semantics so existing call sites stay bit-stable.
- Test matrix covers every entry point x every invalid kwarg combination, plus a separate `test_*_message_parity` block that asserts the same substring across entry points. Pinning the message text catches future drift.
- Signature ordering in `read_vrt` lines up with the canonical `_CANONICAL_ORDER` constant in `test_reader_kwarg_order_1935.py` so the existing pin still passes.
- The full `xrspatial/geotiff/tests/` suite is green (4552 passed); the pre-existing `lz4` failure on `test_lowlevel_write_pushdown_2138.py` reproduces on `main` and is unrelated.

### Checklist
- [x] Algorithm correctness: pure CPU-side input validation, no kernels involved.
- [x] Backend parity: all four read entry points share one validator; tests verify identical exception text across them.
- [x] NaN handling: not applicable (no numerical code).
- [x] Edge cases covered: sentinel defaults, explicit `None`, bool/str/float type checks, file-like + gpu/chunks.
- [x] Dask chunk boundaries: not applicable (no chunk math changed).
- [x] No premature materialization or unnecessary copies: validator does sentinel-identity comparisons only.
- [x] Benchmark coverage: not applicable (validator overhead is one call per read).
- [x] README feature matrix: not applicable (no new public API).
- [x] Docstrings: helper has a full docstring with Parameters/Raises; updated direct-backend docstrings inherit the existing kwarg descriptions.

brendancol (commented):
## PR Review: geotiff: share dispatch-kwarg validation across read entry points (#2175) -- follow-up pass

Reviewing commits `cefe66c` and `3007e8c` against the previous review.

### Disposition of previous findings

- **Blocker (Path coercion in `gpu.py`):** Fixed in `cefe66c`. `_backends/gpu.py:204-210` now coerces `pathlib.Path` to `str` before the validator call. Verified locally that `read_geotiff_gpu(Path('foo.tif'))` no longer raises the file-like rejection. New regression tests at `test_dispatch_validation_parity_2162.py` cover Path inputs across all four entry points. Further tidied in `3007e8c`: import hoisted to module scope, duplicate inner import removed.

- **Suggestion (file-like guard ordering):** Fixed in `cefe66c`. `_validation.py:444-461` moves the file-like guard ahead of the VRT-dependent checks. A file-like buffer with `missing_sources='raise'` now surfaces the file-like restriction first, which names the actual blocker.

- **Suggestion (cross-reference comments on VRT overview_level rejection):** Fixed in `cefe66c`. `__init__.py:557-562` and `_backends/vrt.py:208-214` each reference the other site so a future refactor cannot drop one in isolation.

- **Nit (TODO marker for sentinel migration):** Fixed in `cefe66c`. `_validation.py:431-433` carries the follow-up note.

- **Nit (test message-parity assertion granularity):** Fixed in `cefe66c`. `_get_error` now returns `(type_name, message)` and the parity tests assert on both fields. A regression that swapped `ValueError` for `RuntimeError` while keeping the same message would fail the test.

- **Nit (function-scope `_coerce_path` alias in `gpu.py`):** Fixed in `3007e8c`. Hoisted to the module import block; the function-scope alias and the duplicate inner import are gone.

- **Nit (bare except in Path-object test):** Fixed in `3007e8c`. Narrowed to `(ImportError, RuntimeError)` so a `NameError` or `AttributeError` from a refactor would surface in the test.

- **Nit (`vrt_kwargs` shim duplication):** Deferred to a follow-up. The duplication is mild (two call sites, ~5 lines each) and unifying it would touch unrelated dispatch code in `__init__.py`. Not bundled into this PR.

### New findings (this pass)

#### Blockers

None.

#### Suggestions

None.

#### Nits

None.

### What looks good

- All review findings (Blocker, Suggestions, Nits) from the previous pass are now addressed in-PR, with one Nit explicitly deferred and the rationale recorded.
- The full `xrspatial/geotiff/tests/` suite stays green (4557 passed); the pre-existing `lz4` failure on `test_lowlevel_write_pushdown_2138.py` reproduces on `main` and is unrelated.

### Checklist
- [x] Algorithm correctness: pure CPU-side validation, no kernels.
- [x] Backend parity: type + message asserted across entry points.
- [x] NaN handling: not applicable.
- [x] Edge cases covered: sentinel defaults, explicit `None`, bool/str/float type checks, file-like + gpu/chunks, Path objects.
- [x] No premature materialization: validator does sentinel-identity comparisons.
- [x] Benchmark coverage: not applicable.
- [x] README feature matrix: not applicable.
- [x] Docstrings: helper docstring carries the full rule list with issue numbers.

This PR is ready for merge once CI is green.

PR #2209, opened 2026-05-21
geotiff: migrate VRT backend to shared finalization helpers (PR E of #2162)

### LINKED ISSUE #2180: GeoTIFF: migrate VRT backend to shared finalization helpers (PR E of #2162)

brendancol:
## Summary

Wave 3 of #2162. Route the VRT eager and chunked read paths in `xrspatial/geotiff/_backends/vrt.py` through `_finalize_lazy_read_attrs` from #2177. The two sites previously built `GeoTIFFMetadata` from `VRTDataset` internals by hand and called `metadata_to_attrs` directly, skipping the shared `_validate_read_geo_info` / `_populate_attrs_from_geo_info` block the other backends use.

A new private helper `_vrt_to_synthetic_geo_info` projects the parsed `VRTDataset` onto a `GeoInfo` so the lazy helper's `geo_info_to_metadata` pipeline produces the same attrs surface the inline code did.

This is the final wave 3 PR that completes the #2162 refactor (Wave 1: #2175 + #2177, Wave 2: #2178 + #2179, Wave 3: #2180).

## Documented divergence

The issue body allowed divergence where the helper genuinely cannot apply, with inline comments naming the constraint. The VRT migration falls back to documented divergence in four places:

* Per-band sentinel selection happens above the helper call, because `_apply_eager_nodata_mask` is single-sentinel and `GeoInfo.nodata` cannot model per-band sentinels.
* `attrs['vrt_holes']` rides through `attrs_in` because `GeoInfo` has no field for it. The helper's `attrs.update` leaves it intact.
* `attrs['nodata_pixels_present']` is computed via the VRT-aware scan helpers and stamped after the helper returns, because `_finalize_lazy_read_attrs` passes `pixels_present=None`.
* `validate_read_metadata` runs once pre-read with the VRT-specific `band_nodata` / `band_nodata_values` context; the helper re-runs validation downstream as a no-op for the band-nodata check.

## Test plan

- [x] `test_vrt_band_nodata_1598.py` (6 tests pass)
- [x] `test_open_geotiff_missing_sources_1810.py` (6 tests pass)
- [x] `test_georef_status_2136.py` (33 tests pass including VRT branches)
- [x] New `test_vrt_finalization_parity_2162.py` (29 tests covering VRT eager vs eager numpy attrs, VRT chunked vs dask numpy attrs, band_nodata, missing_sources, and georef_status across the five canonical states)
- [x] Wider VRT test suite (583 tests pass)
- [x] Full geotiff test suite (4637 tests pass; only the pre-existing `lz4` failure on main is deselected)

Closes #2180. Parent #2162.

### DIFF
--- a/xrspatial/geotiff/_backends/vrt.py
+++ b/xrspatial/geotiff/_backends/vrt.py
@@ -30,6 +30,7 @@ from .._geotags import (
 from .._reader import _MAX_CLOUD_BYTES_SENTINEL
 from .._runtime import _ON_GPU_FAILURE_SENTINEL
 from .._validation import (
+    _gdal_geotransform_to_affine_tuple,
     _validate_chunks_arg,
     _validate_dispatch_kwargs,
     _validate_dtype_cast,
@@ -89,10 +90,6 @@ def _vrt_to_synthetic_geo_info(vrt) -> GeoInfo:
         else RASTER_PIXEL_IS_AREA
     )
     if is_rotated:
-        # ``_gdal_geotransform_to_affine_tuple`` lives in ``_validation``;
-        # local import keeps the module-load surface free of a heavy
-        # transitive dependency.
-        from .._validation import _gdal_geotransform_to_affine_tuple
         rotated_affine = _gdal_geotransform_to_affine_tuple(gt)
         return GeoInfo(
             transform=GeoTransform(rotated_affine=tuple(rotated_affine)),
@@ -361,10 +358,7 @@ def read_vrt(source: str, *,
     # set here pre-read and let ``_finalize_lazy_read_attrs`` re-run the
     # ``allow_rotated`` / ``allow_unparseable_crs`` arm later as a no-op.
     import os as _os
-    from .._validation import (
-        validate_read_metadata,
-        _gdal_geotransform_to_affine_tuple,
-    )
+    from .._validation import validate_read_metadata
     from .._vrt import parse_vrt as _parse_vrt, _read_vrt_xml
     _xml_str = _read_vrt_xml(source)
     _vrt_dir = _os.path.dirname(_os.path.abspath(source))
@@ -714,10 +708,7 @@ def _read_vrt_chunked(source, *, window, band, name, chunks, gpu, dtype,
     # Issue #1987 ambiguous-metadata checks on the chunked VRT path. Run
     # before the band-count validator below so a rejected file does not
     # produce side effects.
-    from .._validation import (
-        validate_read_metadata,
-        _gdal_geotransform_to_affine_tuple,
-    )
+    from .._validation import validate_read_metadata
     validate_read_metadata({
         'allow_rotated': allow_rotated,
         'allow_unparseable_crs': allow_unparseable_crs,
--- a/xrspatial/geotiff/_backends/vrt.py
+++ b/xrspatial/geotiff/_backends/vrt.py
@@ -14,16 +14,19 @@ import numpy as np
 import xarray as xr
 
 from .._attrs import (
-    GeoTIFFMetadata,
-    _compute_georef_status_from_parts,
-    _set_nodata_attrs,
-    metadata_to_attrs,
+    _apply_caller_dtype_cast,
+    _finalize_lazy_read_attrs,
 )
 from .._coords import (
     coords_from_pixel_geometry as _coords_from_pixel_geometry,
-    transform_tuple_from_pixel_geometry as _transform_tuple_from_pixel_geometry,
 )
 from .._crs import _wkt_to_epsg
+from .._geotags import (
+    RASTER_PIXEL_IS_AREA,
+    RASTER_PIXEL_IS_POINT,
+    GeoInfo,
+    GeoTransform,
+)
 from .._reader import _MAX_CLOUD_BYTES_SENTINEL
 from .._runtime import _ON_GPU_FAILURE_SENTINEL
 from .._validation import (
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: geotiff: migrate VRT backend to shared finalization helpers (PR E of #2162)

### Blockers
None.

### Suggestions

- `xrspatial/geotiff/_backends/vrt.py:362`: `validate_read_metadata` runs twice on the success path. The first call here carries the VRT-specific `band_nodata` context; the helper call inside `_finalize_lazy_read_attrs` runs it again without those keys. Harmless, but it costs one registered-check dispatch per read. If a future PR extends `_validate_read_geo_info` to forward `band_nodata` / `band_nodata_values`, the two collapse into one call. Not blocking. Worth a followup issue.

- `xrspatial/geotiff/tests/test_vrt_finalization_parity_2162.py`: the matrix does not cover the `dtype=...` + no-sentinel branch where `_apply_caller_dtype_cast` pops `nodata_dtype_cast`. The dask parity test in `test_lazy_finalization_parity_2162.py` already pins this for the non-VRT path. Add the symmetric VRT case so the two attrs surfaces cannot drift. Cheap: open a no-nodata VRT with `dtype=np.float64` and assert `'nodata_dtype_cast' not in attrs`.

### Nits

- `xrspatial/geotiff/_backends/vrt.py:91-94`: the `_gdal_geotransform_to_affine_tuple` import sits inside the rotated arm with a comment about module-load surface. The function is already imported at the eager and chunked call sites further down. A top-level import would let this branch drop the local import and the comment.

- `xrspatial/geotiff/_backends/vrt.py:104`: the no-GeoTransform arm uses `GeoTransform()` for an unused transform field (`has_georef=False`, so the field is never read for coords or transform). A `GeoTransform()` is cheap, so this is fine, but `None` would work too if the dataclass allowed it. `_compute_georef_status` uses `getattr(transform, 'rotated_affine', None)` so either form works.

### What looks good

- Behaviour preserved. All three regression suites cited in the issue body pass (`test_vrt_band_nodata_1598`, `test_open_geotiff_missing_sources_1810`, `test_georef_status_2136` VRT cases), and the wider VRT suite stays green at 583 tests.
- Documented divergence is inline rather than silent. The four-place breakdown (per-band sentinel, `vrt_holes`, `nodata_pixels_present`, double `validate_read_metadata`) matches what the issue body allowed.
- `_vrt_to_synthetic_geo_info` is short and centralises the projection so eager and chunked paths cannot drift on transform / `has_georef` / CRS assembly.
- The parity test pins all five georef states across both backends and compares to the non-VRT counterparts where the comparison applies.

### Checklist

- [x] Algorithm matches reference (n/a, refactor)
- [x] NaN handling unchanged from pre-migration
- [x] Dask chunk boundaries unchanged
- [x] Benchmark not needed (refactor)
- [x] README feature matrix n/a

brendancol (commented):
## Follow-up review

### Original findings disposition

- **Suggestion (vrt.py:362, double `validate_read_metadata`)**: Deferred to follow-up #2210. The fix needs to extend `_validate_read_geo_info`'s signature, which the issue body called out as "wave-1 frozen". Now that waves 2 and 3 have landed, the constraint can relax. The follow-up issue tracks the consolidation.

- **Suggestion (test gap, `dtype=` + no-sentinel)**: Fixed in 5f8ea7d. Added `test_dtype_cast_no_sentinel_omits_attr_eager` and `test_dtype_cast_no_sentinel_omits_attr_chunked` covering both VRT branches.

- **Nit (vrt.py:91-94, local import)**: Fixed in 5f8ea7d. Moved `_gdal_geotransform_to_affine_tuple` to the top-level import block and dropped the three local-import sites.

- **Nit (vrt.py:104, `GeoTransform()` default)**: Dismissed. Changing to `None` would need `GeoInfo.transform`'s annotation widened from `GeoTransform` to `GeoTransform | None` and every reader that constructs `GeoInfo` updated. Style preference, out of scope.

### New findings

Nothing on the second pass. The fixes are localized to the original review items and do not change behaviour.

### Test status

- New parity tests: 31 total in `test_vrt_finalization_parity_2162.py`, all passing.
- Original regression suites pass: `test_vrt_band_nodata_1598`, `test_open_geotiff_missing_sources_1810`, `test_georef_status_2136` VRT cases.
- Full geotiff suite: 4639 pass. The PR adds 31 new tests; one pre-existing `lz4` failure on main is deselected.

Issue #2107, opened 2026-05-19
perf(geotiff): batch _try_nvjpeg2k_batch_decode per-tile allocations


brendancol:
## Summary

`_try_nvjpeg2k_batch_decode` in `xrspatial/geotiff/_gpu_decode.py` (around L2725-2778) allocates per-component output buffers via `cupy.empty(tile_height * pitch)` inside its per-tile decode loop, and synchronizes on the default stream after every single tile. This mirrors the anti-pattern previously fixed in sibling helpers:

- `_try_nvcomp_from_device_bufs` (#1659): per-tile alloc + trailing concat -> single contiguous buffer
- `_try_kvikio_read_tiles` (#1688): per-tile cupy.empty + blocking IOFuture.get -> single buffer + batched submit
- `_nvcomp_batch_compress` (#1712): per-tile cupy.get -> concat + single get

For an N-tile JPEG 2000 read with S samples per pixel, the current pattern is:

```python
for i, tile_data in enumerate(compressed_tiles):
    ...
    comp_bufs = []
    pitch = tile_width * dtype.itemsize
    for c in range(samples):
        buf = cupy.empty(tile_height * pitch, dtype=cupy.uint8)  # N*S allocs
        comp_bufs.append(buf)
    ...
    cupy.cuda.Device().synchronize()  # serialises ALL streams, once per tile
```

The per-tile sync alone forces serial decode (no opportunity for nvJPEG2000's internal pipelining), even when the encode/decode path is otherwise stream-friendly. Each `cupy.empty` allocation round-trips through cupy's memory pool, adding tens of microseconds per call.

## Anti-Pattern Reference

The other batched-codec helpers in the same module use the canonical fix:

- Allocate one contiguous output buffer of size `n_tiles * tile_bytes`.
- Compute per-tile/per-component pointer offsets on host (`np.cumsum` or `np.arange * stride`).
- Pass per-element device pointers as `base_ptr + offsets`.
- Synchronize ONCE after the batch.

Applied to nvJPEG2000, components for tile `i` should live in slices of the existing `d_all_tiles` buffer (or one fresh `d_comp_pool = cupy.empty(n_tiles * samples * tile_height * pitch)`); the loop body just writes pointers into the nvjpeg2kImage_t struct.

## Impact

[...]

Issue #1712, opened 2026-05-12
perf(geotiff): batch _nvcomp_batch_compress per-tile D2H and allocations


brendancol:
## Reason or Problem

`_nvcomp_batch_compress` in `xrspatial/geotiff/_gpu_decode.py` is the GPU write-side compress path (deflate/zstd via nvCOMP). Two related anti-patterns remain in this function that mirror the ones already fixed elsewhere by #1552 and #1659:

**1. Per-tile cupy.empty allocations (line 2457):**

```python
d_comp_bufs = [cupy.empty(max_cs, dtype=cupy.uint8) for _ in range(n_tiles)]
```

For an N-tile compress, this issues N separate cupy allocations (each potentially a memory-pool query plus a kernel launch for zero-init under some allocator configs). The decode side at `_gpu_decode.py:2004` and the LZW kernel decode at `_gpu_decode.py:1973` already use a single contiguous allocation + offsets/views.

Microbench (1024 tiles, 16 KiB each):
- N separate `cupy.empty`: 4.74 ms
- 1 `cupy.empty` + N slice views: 1.02 ms

**2. Per-tile `.get().tobytes()` D2H readback (lines 2522-2526):**

```python
comp_sizes = d_comp_sizes.get().astype(int)
result = []
for i in range(n_tiles):
    cs = int(comp_sizes[i])
    raw = d_comp_bufs[i][:cs].get().tobytes()
    ...
```

Each `.get()` is a separate D2H DMA on the default stream, so they serialise. The comment block at lines 2503-2505 even calls out the same problem for the uncompressed-side adler32 path:

> Batch the GPU->CPU transfer so all tiles move in a single DMA instead of one .get() per tile (which serializes on the default stream and is the dominant cost on the deflate path).

But the compressed side at lines 2522-2526 still does per-tile `.get()`. The decode-side fix (#1552) batched the same pattern in `gpu_decode_tiles_from_file` (lines 870-913 region).

Microbench (1024 tiles, 16 KiB each, filled, on a single Ampere-class GPU):
- per-tile `.get().tobytes()`: 35.19 ms
- 1 concat + 1 `.get()` + Python slice: 19.31 ms (~45% reduction)

## Proposal

Mirror the existing batched D2H pattern used by `gpu_decode_tiles_from_file` (#1552) and `_try_nvcomp_from_device_bufs` (#1659):

[...]

Issue #1688, opened 2026-05-12
perf(geotiff): batch _try_kvikio_read_tiles to one alloc + parallel preads


brendancol:
## Reason or Problem

`_try_kvikio_read_tiles` in `xrspatial/geotiff/_gpu_decode.py` (line 941) pulls each compressed tile into GPU memory through three per-tile operations that serialise:

```python
d_tiles = []
with kvikio.CuFile(file_path, 'r') as f:
    for off, bc in zip(tile_offsets, tile_byte_counts):
        buf = cupy.empty(bc, dtype=cupy.uint8)
        nbytes = f.pread(buf, file_offset=off)
        actual = nbytes.get() if hasattr(nbytes, 'get') else int(nbytes)
        if actual != bc:
            return None
        d_tiles.append(buf)
```

Three problems compound in that loop on a typical multi-tile COG read:

1. `cupy.empty(bc, ...)` allocates one device buffer per tile. Each `cupy.empty` call costs tens of microseconds of setup independent of `bc`. A 256-tile COG pays that overhead 256 times, the same per-tile allocation pattern PR #1659 just fixed in `_try_nvcomp_from_device_bufs`.
2. `f.pread(...)` returns an `IOFuture` so the call itself is non-blocking. The very next line calls `nbytes.get()`, which blocks until that one pread finishes. Every tile waits before the next pread is submitted, so kvikio's internal thread pool never sees more than one outstanding request and the parallel-IO design of `pread` collapses to serial.
3. No `_check_gpu_memory(sum(tile_byte_counts), ...)` guard runs before the allocations. A crafted COG with large `TileByteCounts` can OOM the device one tile at a time before any single allocation hits a guard. The sibling paths (`_try_nvcomp_from_device_bufs` at L1641, `_batched_d2h_to_bytes` at L930) all check the total bytes up front.

## Proposal

Move `_try_kvikio_read_tiles` to the same single-buffer + batched-future pattern the other GDS/nvCOMP paths in this file already use:

```python
sizes = [int(bc) for bc in tile_byte_counts]
offsets = np.zeros(len(sizes), dtype=np.int64)
if len(sizes) > 1:
    np.cumsum(sizes[:-1], out=offsets[1:])
total_bytes = int(np.sum(sizes))

_check_gpu_memory(total_bytes, what="kvikio tile read buffer")
combined = cupy.empty(total_bytes, dtype=cupy.uint8)

futures = []
[...]

Issue #1659, opened 2026-05-12
perf: replace per-tile alloc + concat in _try_nvcomp_from_device_bufs (GDS+nvCOMP fast path)


brendancol:
## Summary

`_try_nvcomp_from_device_bufs` in `xrspatial/geotiff/_gpu_decode.py` allocates N separate CuPy buffers (one per tile) for nvCOMP decompression output, then `cupy.concatenate`s them after the decompress kernel returns. The other nvCOMP paths in this module (LZW at L1847, deflate at L1878, host-buffer at L1114) already allocate a single contiguous `cupy.empty(n_tiles * tile_bytes)` up front and derive per-tile device pointers as `base_ptr + i * tile_bytes`. The per-tile + concat pattern keeps two copies of the decompressed data on device simultaneously, and the trailing concatenate is pure overhead.

## Where it triggers

This path runs when:
- `kvikio` is importable (GPUDirect Storage available)
- the `nvCOMP` shared library loads via ctypes
- the COG is ZSTD-compressed (compression tag 50000)

That is the fastest read path the GPU pipeline offers: NVMe to GPU via DMA, nvCOMP decompresses on device, no host involvement.

## Measurement (overhead only)

The numbers below isolate the cost of the alloc + concat steps themselves. They do NOT include nvCOMP decompression latency, kvikio I/O, or any other work the function does -- they cover only what would be deleted by the refactor. CUDA 12, RTX-class GPU:

```
n=  64 tile_bytes= 65536: per-tile alloc+concat= 20.84 ms; single buf+ptrs=  0.56 ms  (-20.3 ms)
n= 256 tile_bytes= 65536: per-tile alloc+concat=  3.66 ms; single buf+ptrs=  0.69 ms  ( -3.0 ms)
n=1024 tile_bytes= 65536: per-tile alloc+concat= 11.91 ms; single buf+ptrs=  2.39 ms  ( -9.5 ms)
n= 256 tile_bytes=262144: per-tile alloc+concat=  8.18 ms; single buf+ptrs=  0.13 ms  ( -8.1 ms)
```

The proportion of end-to-end `read_geotiff_gpu` time that this overhead represents depends on how fast the actual nvCOMP+GDS work is on the target host. The local environment lacks kvikio so an end-to-end measurement of this exact path is not possible here; the table above is the isolated overhead.

[...]

PR #2213, opened 2026-05-21
perf(geotiff): scope nvJPEG/nvJPEG2K encoder per-tile sync to default stream (#2212)

### LINKED ISSUE #2212: perf(geotiff): replace Device().synchronize() with Stream.null.synchronize() in nvJPEG/nvJPEG2000 encoders
## Summary

`_nvjpeg_batch_encode` (`xrspatial/geotiff/_gpu_decode.py` ~L1560) and `_nvjpeg2k_batch_encode` (~L2942) both call `cupy.cuda.Device().synchronize()` inside their per-tile encode loops. `Device().synchronize()` is a whole-device fence that blocks every CUDA stream, defeating overlap with any other concurrent CUDA work (e.g. another stream's predictor encode or D2H copy).

The matching decoder `_try_nvjpeg_batch_decode` at L1442 already uses `cupy.cuda.Stream.null.synchronize()` for the same purpose -- it only blocks the default stream, which is the only stream the encode/retrieve sequence depends on (encoder state and the default stream the calls were issued on).

## Why this is real perf waste

- nvJPEG and nvJPEG2000 encoders maintain a single shared state per encoder, so encodes within one batch are inherently serial -- we can't batch the encode calls themselves.
- The fix is not to remove the sync, but to scope it to the default stream so concurrent GPU work the caller is doing on other streams is no longer serialised behind every per-tile encode.
- Pass 12 (#2107) fixed the analogous anti-pattern on the decode side for `_try_nvjpeg2k_batch_decode`. The encode side was missed.

## Repro

Calling `to_geotiff(da, ..., gpu=True, compression='jpeg')` or `compression='jpeg2000'` on a multi-tile raster hits the loops above. Each tile pays a full-device sync.

## Proposed fix

Replace `cupy.cuda.Device().synchronize()` with `cupy.cuda.Stream.null.synchronize()` in both encoder loops, matching the decoder's pattern at L1442.

## Pattern history

- #2107: nvJPEG2000 decode batch sync fix
- #1659: nvCOMP per-tile alloc fix
- #1688: kvikio per-tile pread fix
- #1712: nvCOMP encode per-tile D2H fix

This issue closes the encoder-side gap.

## Sweep context

Filed by /sweep-performance (Pass 13, 2026-05-20) on the `deep-sweep-performance-geotiff-2026-05-20` worktree branch.

brendancol:
## Summary

- `_nvjpeg_batch_encode` and `_nvjpeg2k_batch_encode` in `_gpu_decode.py` called `cupy.cuda.Device().synchronize()` inside their per-tile encode loops, a whole-device fence that serialised concurrent work on other CUDA streams.
- The matching decoder `_try_nvjpeg_batch_decode` already used `cupy.cuda.Stream.null.synchronize()`; the encoder side was inconsistent.
- Switch both encoders to `Stream.null.synchronize()` so concurrent work on other streams (predictor encodes, D2H copies, other channels) is no longer blocked. The encode/retrieve API ordering is preserved because both calls were issued on the default (null) stream.

Fixes #2212.

## Why this is safe

nvJPEG / nvJPEG2000 encoders maintain a single shared state per encoder, so encodes within one batch are inherently serial. The sync is what guarantees the next iteration's encode call sees a completed prior encode. `Stream.null.synchronize()` waits for everything queued on the default stream, which is the same stream the encode/retrieve calls were issued on -- the ordering contract is preserved.

The decoder counterpart `_try_nvjpeg_batch_decode` at `_gpu_decode.py:1442` has used this pattern since before #2107; this PR aligns the encoders with the established codebase pattern.

## Pattern history

- #2107: nvJPEG2000 *decode* batch sync + alloc fix (analogous change on the decode side)
- #1659: nvCOMP per-tile alloc fix
- #1688: kvikio per-tile pread fix
- #1712: nvCOMP encode per-tile D2H fix

## Test plan

- [x] 5 new AST-level structural tests in `test_nvjpeg_encode_stream_sync_2212.py` covering both encoders and pinning the decoder reference pattern.
- [x] 19 existing related encoder/decoder tests pass (`test_nvjpeg2k_single_alloc_2107.py`, `test_gpu_writer_compression_modes_2026_05_11.py`).
- [ ] End-to-end nvJPEG / nvJPEG2000 encode verification is gated on a host with libnvjpeg / libnvjpeg2k installed; this host has neither. Re-validate on a RAPIDS conda env.
- [x] `.claude/sweep-performance-state.csv` updated (Pass 13, 2026-05-20).


brendancol:
## PR Review: nvJPEG/nvJPEG2K encoder per-tile sync scope

### Blockers
None.

### Suggestions
None.

### Nits
- The non-loop `cupy.cuda.Device().synchronize()` calls at `_gpu_decode.py:1013`, `:1268`, `:1764`, `:2537`, and `:2830` are batch-end barriers on the same default-stream-only paths. Same argument as the encoder fix would scope them to `Stream.null.synchronize()`. Out of scope for #2212; flagging for a follow-up sweep.

### What looks good
- Both `nvjpegEncodeImage` and `nvjpeg2kEncode` are issued with `stream=0`, so syncing the null stream is enough for the encode/retrieve ordering.
- Matches the decoder at `_try_nvjpeg_batch_decode` (`_gpu_decode.py:1442`).
- Inline comments at `:1560-1569` and `:2954-2957` cite #2212 and #2107, so the rationale is in the file.
- AST tests cover both encoders and pin the decoder pattern as the reference. They run without libnvjpeg installed.
- 5 new tests pass locally (`test_nvjpeg_encode_stream_sync_2212.py`).

### Caveats
- End-to-end encode verification needs a host with libnvjpeg / libnvjpeg2k. The PR body notes re-validation on a RAPIDS env. The structural tests won't catch a behavioral bug where `Stream.null.synchronize()` returns before the encode actually completes, but that would be a CuPy/driver issue, not something this PR introduces.

### Checklist
- [x] Sync scope correct for the encode/retrieve calls
- [x] Matches the decoder pattern
- [x] Per-tile sync semantics preserved (encoder state still safe between iterations)
- [x] Comments reference issue numbers
- [x] Structural regression test in place
- [x] 5 new tests pass; 19 existing related tests pass
- [N/A] End-to-end encode test (libnvjpeg gated)
- [N/A] Benchmark (no geotiff encode benchmarks in suite)
- [N/A] README feature matrix (internal perf fix)

PR #2218, opened 2026-05-21
geotiff: fail closed on 1xN / Nx1 writes without explicit pixel size

### LINKED ISSUE #2214: geotiff: 1xN / Nx1 writes silently invent the missing pixel size
**Describe the bug**

`coords_to_transform()` in `xrspatial/geotiff/_coords.py` (around line 417) invents a pixel size for the degenerate axis of 1xN / Nx1 rasters by copying the magnitude of the non-degenerate axis. The step on a length-1 coord array is undefined, so the writer cannot read it off the coords. The fix from #1945 borrows the other axis's spacing and writes the file with implied square pixels.

That assumption is unsafe. If the source raster has 30 m by 10 m pixels and the caller hands the writer a 1xN strip, the writer records 30 m by 30 m with no warning. Downstream slope, proximity, and zonal math then trusts a wrong transform.

**Expected behavior**

The writer should fail closed when one spatial axis has length 1 and the caller has not supplied an explicit transform. Two ways to keep a successful write:

1. Supply the affine transform on `attrs['transform']` (rasterio 6-tuple). This is already the recommended path for 1x1 inputs and round-trips bit-exactly.
2. Opt in to the square-pixel assumption with `attrs['assume_square_pixels_for_degenerate_axis'] = True` (default `False`). This restores the #1945 borrow-from-other-axis behaviour for callers who know their source is square.

Without one of those, the writer raises `ValueError` with a message pointing the caller at both escape hatches.

**Reproducer**

```python
import numpy as np
import xarray as xr

# Source raster has 10 m y pixels and 30 m x pixels.
PIXEL_X = 30.0
PIXEL_Y = 10.0

da = xr.DataArray(
    np.arange(8, dtype="float32").reshape(1, 8),
    dims=("y", "x"),
    coords={
        "x": np.arange(8, dtype="float64") * PIXEL_X,
        "y": np.array([0.0], dtype="float64"),
    },
    attrs={"crs": 4326},
)

to_geotiff(da, "/tmp/strip.tif")
r = open_geotiff("/tmp/strip.tif")
print(r.attrs["transform"])
# Writes pixel_height = -30.0 (borrowed from x axis), not the true -10.0.
```

**Scope**

This affects every writer that routes through `coords_to_transform`: eager numpy, dask+numpy, gpu (cupy), dask+cupy, and the VRT writer.

**Acceptance criteria**

- 1xN write without `attrs['transform']` or the opt-in raises `ValueError`.
- Nx1 write without `attrs['transform']` or the opt-in raises `ValueError`.
- 1xN write with `attrs['transform']` round-trips the supplied transform.
- 1xN write with `attrs['assume_square_pixels_for_degenerate_axis'] = True` writes the borrowed-square transform.
- Existing 2-row / 2-column writes are unchanged.
- The error message names both escape hatches.

**Additional context**

The borrow path landed in #1945 to stop silent georef stripping on degenerate writes. The current behaviour swaps one silent failure for another: the pixel size on a length-1 axis is not recoverable from coords, but the writer pretends otherwise. Fail-closed-with-opt-in keeps #1945's explicit case working and removes the silent-invent path for everyone else.

brendancol:
Closes #2214.

## Summary

- `coords_to_transform` no longer invents the pixel size for the degenerate axis on 1xN / Nx1 writes. The borrow-from-other-axis path that landed in #1945 ran unconditionally and silently produced wrong transforms when the source raster was not square.
- The new default is fail-closed: a 1xN / Nx1 `DataArray` with spatial coords on both axes but no `attrs['transform']` and no opt-in flag raises `ValueError`. The message names both escape hatches.
- Two ways to keep a successful write:
  - `attrs['transform']` (rasterio 6-tuple) supplies the true pixel geometry and round-trips bit-exactly. Preferred for non-square sources.
  - `attrs['assume_square_pixels_for_degenerate_axis'] = True` opts in to the #1945 borrow-from-other-axis fallback. Identity-checked against `True` so a stray truthy string can't re-enable the silent path.
- Multi-row / multi-column writes are unchanged. 1x1 inputs still need `attrs['transform']` for the same reason as before.

## Backend coverage

`coords_to_transform` is the single helper every writer routes through:

- numpy (`to_geotiff` eager path)
- dask + numpy (`to_geotiff` streaming path)
- gpu / cupy (`write_geotiff_gpu`)
- dask + cupy (`to_geotiff` with chunked cupy)
- VRT writer (`to_geotiff(*.vrt)`)

All five hit the same fail-closed branch via `require_transform_for_georeferenced`.

## Test plan

- [x] New tests in `test_degenerate_pixel_size_2214.py`: 1xN / Nx1 raise without opt-in, `attrs['transform']` round-trip preserves the true non-square pixel size, opt-in flag restores the #1945 borrow behaviour, the opt-in must be the literal `True` (not a truthy string), and 2x2 / 3x5 writes are unchanged.
- [x] Existing #1945 round-trip tests updated to set the opt-in flag and still pass.
- [x] Full `xrspatial/geotiff/tests/` suite: 2374 passed, 40 skipped. The one pre-existing failure (`test_lowlevel_write_pushdown_2138::test_write_vs_to_geotiff_byte_parity_uint8[lz4]`) reproduces on `main` and is unrelated to this change (lz4 experimental-codec gate from #2137).


brendancol (commented):
## PR Review: geotiff: fail closed on 1xN / Nx1 writes without explicit pixel size

None.

- [ ] `xrspatial/geotiff/tests/test_degenerate_pixel_size_2214.py`: the new test file only exercises the eager numpy writer (`to_geotiff` on a non-chunked DataArray). The fail-closed branch lives in the shared `coords_to_transform` helper, so dask+numpy, dask+cupy, gpu, and vrt writers all hit the same code by construction, but no test pins that. The pre-existing `test_degenerate_georef_1945.py` covers the round-trip (opt-in) path on every backend; a parallel cross-backend pass on the fail-closed default would catch a future refactor that bypasses `_require_transform_for_georeferenced` on one specific writer (e.g. an inlined copy added for performance). Suggest a short class like `TestDegenerateFailClosedAcrossBackends` that exercises the dask path and the VRT writer at minimum.

- [ ] `xrspatial/geotiff/tests/test_degenerate_pixel_size_2214.py` (the `test_Nx1_optin_borrows_from_y_axis` body comment): the explanation "True y step is -10 (top-down), so |pixel_height| = 10 is what gets copied to pixel_width" is correct but takes a moment to parse. Consider tightening to "Borrow path takes abs(y step) = 10 and copies it onto pixel_width."
- [ ] `xrspatial/geotiff/_coords.py:51-62`: `_assume_square_for_degenerate` mirrors the shape of `_has_no_georef_marker` exactly (mapping check, identity check on `True`). Could share a small internal predicate, but factoring out is probably noise for two callers. Leave as-is.

### What looks good

- The split between fail-closed default and opt-in flag matches the project's pattern. Identity-check on `True` (rather than truthiness) tracks the existing `_has_no_georef_marker` pattern and prevents the "stray attrs value re-enables silent invent" footgun.
- The error message names both escape hatches with their exact syntax. Future users hitting the raise will not have to read the source to recover.
- Helper-level tests in `TestCoordsToTransformHelperContract` exercise `coords_to_transform` directly, independent of any writer wrapping. That pins the contract on the helper rather than relying on writer side effects.
- Docs in `geotiff.rst` describe the new contract, the two escape hatches, and the rationale. The entry sits between the strict-mode and VRT sections, where anyone scanning that file will find it.
- The existing `_strip_1xN` / `_strip_Nx1` helpers in `test_degenerate_georef_1945.py` were updated rather than duplicated, so the dask / gpu / dask+cupy / vrt round-trip tests already pin the opt-in path on every backend.

### Checklist
- [x] Algorithm matches the cited issue: fail closed unless `attrs['transform']` or the opt-in flag is set.
- [x] All implemented backends share the same helper; opt-in path is exercised on all four.
- [ ] Fail-closed default is only exercised on the eager numpy path (see Suggestion above).
- [x] NaN handling: unchanged. No NaN-bearing code paths in the diff.
- [x] Edge cases: 1x1 still raises (existing #1945 coverage), 2x2 / 3x5 multi-axis paths covered.
- [x] Dask: shared helper, no chunking changes.
- [x] No premature materialization or unnecessary copies; the helper only walks coord arrays.
- [x] No benchmark needed (no performance-sensitive code touched).
- [x] README feature matrix unchanged (no new function or backend support added).
- [x] Docstrings on `_assume_square_for_degenerate` and the updated error message are clear.

brendancol (commented):
## Review follow-up

Dispositions of the original review (PR #2218):

**Suggestion** — cross-backend fail-closed test coverage: **fixed** in `01ac592`. Added `TestDegenerateFailClosedAcrossBackends` with raise checks on dask+numpy (1xN, Nx1), VRT (1xN, Nx1), GPU (1xN, Nx1), and dask+cupy (1xN). All seven tests pass on this machine (GPU branch ran live).

**Nit** — `test_Nx1_optin_borrows_from_y_axis` comment tightening: **fixed** in `01ac592`. Replaced the multi-clause explanation with a single-line "Borrow path takes abs(y step) = 10 and copies it onto pixel_width."

**Nit** — factoring `_assume_square_for_degenerate` / `_has_no_georef_marker` into a shared predicate: **dismissed**. The original review marked this as "leave as-is" and I agree. Two callers, identical shape; abstracting would add an indirection without removing duplication.

Re-ran `xrspatial/geotiff/tests/test_degenerate_pixel_size_2214.py` and `test_degenerate_georef_1945.py`: 50 passed in 0.74s, no new failures.

PR #2219, opened 2026-05-21
geotiff: split lazy-read helper dtype into graph_dtype + caller_dtype (#2206)

### LINKED ISSUE #2206: geotiff: split _finalize_lazy_read_attrs dtype into graph_dtype + caller_dtype
## Goal

Split the `dtype` parameter on `_finalize_lazy_read_attrs` into two so the call-site fixup in the dask backends can go away.

## Context

`_finalize_lazy_read_attrs` (in `xrspatial/geotiff/_attrs.py`, added in #2177) takes one `dtype` argument that does two jobs:

1. Resolved graph dtype: drives `masked_nodata = bool(mask_nodata and np.dtype(dtype).kind == 'f')`.
2. Caller-supplied cast attr: written as `attrs['nodata_dtype_cast'] = np.dtype(dtype).name`.

On the dask paths these are different values. When `mask_nodata=True` and the source is integer, the dask backends auto-promote the graph dtype to `float64` without the caller asking for a cast. The auto-promotion is the right value for `masked_nodata` and the wrong value for `nodata_dtype_cast`.

#2178 worked around this by passing the resolved graph dtype to the helper and fixing up `nodata_dtype_cast` via a small `_apply_caller_dtype_cast` helper (one call per backend). The helper docstring on `_finalize_lazy_read_attrs` already flagged the conflation as deferred to the migration PR.

## Proposed change

Split the helper's signature:

```python
def _finalize_lazy_read_attrs(
    *,
    geo_info,
    nodata,
    mask_nodata,
    graph_dtype,        # was: dtype; drives masked_nodata
    caller_dtype=None,  # new; drives nodata_dtype_cast
    window,
    allow_rotated=False,
    allow_unparseable_crs=False,
    attrs_in=None,
) -> dict:
    ...
```

`graph_dtype` is the resolved dask graph dtype (what the helper currently computes `masked` from). `caller_dtype` is the caller's `dtype=` kwarg verbatim, or `None` when the caller did not pass one.

After the split:

* `_apply_caller_dtype_cast` becomes dead code; remove it.
* Both dask backends drop the post-helper fixup; the helper handles both attrs in one place.

## Out of scope

The eager backends (#2179 wave 2 sibling, #2180 wave 3) do not have the auto-promotion issue because masking is folded into a single eager step; they would just pass `caller_dtype=dtype` and `graph_dtype=resolved_dtype` symmetrically.

## Files

* `xrspatial/geotiff/_attrs.py` (`_finalize_lazy_read_attrs` signature, `_apply_caller_dtype_cast` removal)
* `xrspatial/geotiff/_backends/dask.py` (drop the fixup, rename kwarg at the call site)
* `xrspatial/geotiff/_backends/gpu.py` (same)
* Wave 2/3 PRs (#2179, #2180) should also be updated if they merged before this change.

## Acceptance criteria

* `_finalize_lazy_read_attrs` takes two dtype parameters.
* The dask backends call the helper once with no post-call fixup.
* Existing tests still pass.
* `test_lazy_finalization_parity_2162.py` still passes unchanged.

brendancol:
Closes #2206.

## Summary

- `_finalize_lazy_read_attrs` used to take one `dtype` argument that doubled as the resolved graph dtype (drives `masked_nodata`) and the caller-supplied cast attr (drives `nodata_dtype_cast`). On the dask paths those are different values whenever `mask_nodata=True` runs against an int source — the graph dtype auto-promotes to float64 but `nodata_dtype_cast` should record caller intent, not the auto-promotion.
- Split the helper into `graph_dtype` and `caller_dtype`, drop `_apply_caller_dtype_cast`, and remove the post-helper fixup from the dask, dask+GPU, and VRT call sites. Eager-style VRT callers pass `graph_dtype=resolved_dtype` and `caller_dtype=dtype` symmetrically.

## Backend coverage

- numpy / cupy eager: unchanged (do not route through this helper).
- dask+numpy: call site simplified, attr semantics unchanged.
- dask+cupy: call site simplified, attr semantics unchanged.
- VRT eager + chunked: call sites simplified, attr semantics unchanged.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_finalization_helpers_2162.py` (24 pass)
- [x] `pytest xrspatial/geotiff/tests/test_lazy_finalization_parity_2162.py` (21 pass, unchanged)
- [x] `pytest xrspatial/geotiff/tests/test_vrt_finalization_parity_2162.py` (31 pass)
- [x] `pytest xrspatial/geotiff/tests/` (4641 pass, one pre-existing lz4 codec-gate failure unrelated to this change)
- [ ] CI on all configured Python / OS combos


brendancol (commented):
# Domain-aware review

## Blockers
None. Behavior is preserved across all four backends. The new two-parameter form is semantically equivalent to the pre-PR `_finalize_lazy_read_attrs(dtype=graph_dtype)` + `_apply_caller_dtype_cast(caller_dtype, has_nodata)` pair:

- `nodata is None` -> `_set_nodata_attrs` short-circuits before writing `nodata_dtype_cast`, matching the old "stay absent" branch when `not has_nodata`.
- `caller_dtype is None` -> `dtype_cast_attr` is `None` -> `_set_nodata_attrs` skips writing the attr (no pre-existing value to "pop", since the helper always builds a fresh attrs dict).
- `masked` is unchanged (still derived from the graph dtype).

All 2127 scoped geotiff tests pass. The one unrelated failure (`test_lowlevel_write_pushdown_2138.py::...[lz4]`) is on the write path and pre-exists on `main`.

## Suggestions

1. **Docstring inaccuracy** at `_attrs.py` (the "Eager-style callers (VRT) pass them symmetrically" line). The two VRT call sites actually pass them asymmetrically (`graph_dtype=pre_cast_dtype` / `declared_dtype` vs `caller_dtype=dtype`) for the same reason as dask: a caller `dtype=` cast must not flip `masked_nodata`. The call-site comment at `vrt.py:544-549` correctly explains this; the helper docstring contradicts it.

2. **Test coverage gap**: no test exercises `graph_dtype != caller_dtype` with both non-`None` (e.g. `graph_dtype='float64'`, `caller_dtype='int16'`). Coverage is (float64, float64), (int16, None), (int16, int16), (float64, None), (None, None) -- a swap of the two parameters would slip past. A single test pinning that the attr records `caller_dtype` when they differ would lock in the contract.

## Nits

1. The "When ``nodata is None`` the attr stays absent regardless" sentence is filed under the `caller_dtype` bullet but applies equally to `graph_dtype`'s `masked` (no `masked_nodata` attr either). Worth lifting to a shared note.

2. The older docstring line "``dtype_cast`` is recorded when the caller passed an explicit ``dtype=`` kwarg" reads cleaner post-split as "when ``caller_dtype`` is not None and ``nodata`` is declared."

PR #2220, opened 2026-05-21
geotiff: raise NonUniformCoordsError consistently for spatial aliases

### LINKED ISSUE #2215: geotiff: NonUniformCoordsError leaks as plain ValueError for spatial alias coords
**Describe the bug**

`to_geotiff()` documents spatial dim aliases (`lat`/`lon`, `latitude`/`longitude`, `row`/`col`), but the ambiguous-metadata validator only looks at `data.coords['y']` and `data.coords['x']` before calling `validate_write_metadata()` in `xrspatial/geotiff/_writers/eager.py:348-353`. If a DataArray uses one of the alias names instead, the `NonUniformCoordsError` check at `xrspatial/geotiff/_validation.py:881` never sees those coord arrays.

The transform-synthesis path in `xrspatial/geotiff/_coords.py:399` does still catch non-uniform `lat`/`lon` spacing, but it raises a plain `ValueError` ("coords are not uniformly spaced") instead of the exported `NonUniformCoordsError`. So `except NonUniformCoordsError:` works for `y`/`x` coords and quietly misses the same condition on `lat`/`lon` coords. Which exception you get ends up depending on which dim name the caller picked.

**Expected behavior**

Same error type regardless of which documented alias the caller uses. Two things need to change:

1. The validator at the writer entry point should resolve the spatial dim aliases from `_Y_DIM_NAMES` / `_X_DIM_NAMES` in `xrspatial/geotiff/_runtime.py` (`y`, `lat`, `latitude`, `row` and `x`, `lon`, `longitude`, `col`) before passing coord arrays into `validate_write_metadata()`.
2. The non-uniform check in `_coords.py::_is_regular` should raise `NonUniformCoordsError`, not plain `ValueError`, so `isinstance(exc, NonUniformCoordsError)` holds on every non-uniform-coord failure.

After the fix, `except NonUniformCoordsError` should fire on non-uniform `lat`/`lon` coords the same way it does on non-uniform `y`/`x` coords.

**Additional context**

Caught by an external review of the geotiff module. This is a contract inconsistency, not a data-corruption bug. Both code paths already fail closed. The problem is that the documented `NonUniformCoordsError` is unreliable for callers who picked an alias name.

brendancol:
## Summary

Closes #2215. The ambiguous-metadata validator at the `to_geotiff` / `write_geotiff_gpu` entry points only looked at `coords['y']` / `coords['x']`, so DataArrays with documented spatial aliases (`lat`/`lon`, `latitude`/`longitude`, `row`/`col`) slipped past the `NonUniformCoordsError` check. Non-uniform alias-named coords still failed closed, but as a plain `ValueError` from the later transform-synthesis path in `_coords.py::_is_regular`, not the typed error the writer's docstring tells callers to expect.

- New `_resolve_spatial_coords` helper in `_runtime.py` looks up the documented aliases in `_Y_DIM_NAMES` / `_X_DIM_NAMES` and returns the matching coord arrays. Canonical `y` / `x` come first in the alias list so existing arrays keep matching exactly the same coord as before.
- Helper wired into all three writer entry points: eager `to_geotiff`, the eager dask-VRT writer, and the GPU writer.
- `_coords.py::_is_regular` raises `NonUniformCoordsError` instead of plain `ValueError`. The class already subclasses `ValueError` via `GeoTIFFAmbiguousMetadataError`, so legacy `except ValueError` clauses keep catching it.

## Backend coverage

All four writer backends pick up the fix because the helper lives in the validator that all three entry points share:
- numpy (eager `to_geotiff`)
- dask+numpy (eager dask-VRT writer)
- cupy and dask+cupy (GPU writer)

## Test plan

- [x] 25 new tests in `test_non_uniform_coords_alias_2215.py` cover the alias resolution helper, alias-named non-uniform y / x coords, alias-named constant coords, alias-named uniform coords (no false positive), and the legacy `except ValueError` path
- [x] Parametrized over every alias in `_Y_DIM_NAMES` / `_X_DIM_NAMES`, with a coverage assertion that fails if a new alias is added without updating the parametrization
- [x] Full geotiff test suite passes (3878 passed, one unrelated `lz4` failure also exists on main)


brendancol (commented):
## PR Review: geotiff: raise NonUniformCoordsError consistently for spatial aliases

Read every changed file in full from the PR worktree on branch `issue-2215`. The fix is sound: the validator now resolves spatial aliases (`lat`/`lon`, `latitude`/`longitude`, `row`/`col`) before passing coord arrays into `validate_write_metadata`, and `_coords.py::_is_regular` raises `NonUniformCoordsError` so the exception type no longer depends on which dim name the caller picked. Legacy `except ValueError` callers keep working because the new class subclasses `ValueError` through `GeoTIFFAmbiguousMetadataError`.

None.

- [ ] `xrspatial/geotiff/_runtime.py:95-97` — `_resolve_spatial_coords` is only consumed by the writer ambiguous-metadata validators. A one-line cross-reference comment immediately above the `def` (`# Used by the writer ambiguous-metadata validators; see issue #2215`) would help future readers find the call sites without going through the docstring. Borderline-redundant given the docstring already names #2215, so feel free to dismiss.

- [ ] `xrspatial/geotiff/_runtime.py:96` — only one blank line between the end of `_resolve_spatial_coords` and the next module-level comment block (`_TIME_DIM_NAMES`). PEP 8 wants two between top-level definitions, and the rest of the file follows that convention.

### What looks good
- All three writer entry points (eager `to_geotiff`, eager dask-VRT writer, GPU writer) call the helper, so the dispatch contract is consistent.
- `_resolve_spatial_coords` puts `y` / `x` first in the alias order, so arrays that already use the canonical names keep matching exactly the same coord they did before.
- Tests parametrize over every alias in `_Y_DIM_NAMES` / `_X_DIM_NAMES` and include a coverage-drift assertion that fails if a new alias is added without updating the parametrization.
- Legacy `except ValueError` path is pinned (`test_legacy_except_value_error_still_catches`).
- Existing `test_coord_regularity_1720.py` (which catches plain `ValueError` for the `_is_regular` message) keeps passing because `NonUniformCoordsError` subclasses `ValueError`.

### Checklist
- [x] Algorithm matches reference (issue #2215 description)
- [x] All implemented backends produce consistent results (one shared validator)
- [x] NaN handling unchanged (N/A; error-type fix)
- [x] Edge cases covered (alias names, constant coords, uniform coords, no coords)
- [ ] Dask chunk boundaries — N/A; no neighborhood work
- [x] Benchmark — N/A; pure validation fix
- [ ] README feature matrix — N/A; no new function

brendancol (commented):
## PR Review (follow-up after f167a82)

Re-read the updated `_runtime.py`. Both findings from the previous review are addressed:

- Suggestion (cross-reference comment): fixed in f167a82. A four-line comment above `_resolve_spatial_coords` names the call sites and the issue number.
- Nit (PEP 8 blank-line spacing): fixed in f167a82. Two blank lines now separate the helper from the `_TIME_DIM_NAMES` block.

No new findings. The fix is ready from a review standpoint; remaining gates are CI and `mergeable` against `main`.

PR #2222, opened 2026-05-21
geotiff: forward band_nodata context through _validate_read_geo_info (#2210)

### LINKED ISSUE #2210: geotiff: extend _validate_read_geo_info to forward band_nodata context
Follow-up to #2180 (wave 3 of #2162).

After #2180 routed the VRT eager and chunked paths through
`_finalize_lazy_read_attrs`, both VRT call sites still run
`validate_read_metadata` once pre-read with the VRT-specific
`band_nodata` / `band_nodata_values` keys. The helper's internal
`_validate_read_geo_info` then runs `validate_read_metadata` again
without those keys, so the mixed-band check
(`_check_read_mixed_band_metadata`) is a no-op on the second call.

Harmless, but it wastes one registered-check dispatch per VRT read.

Two ways to consolidate:

1. Extend the `_validate_read_geo_info` signature with optional
   `band_nodata` / `band_nodata_values` kwargs. The VRT callers pass
   them; the non-VRT callers omit. The wave-1 frozen signature
   constraint (issue #2177) is now relaxed because waves 2 and 3 have
   landed.

2. Push the `band_nodata` context into `geo_info` itself so the helper
   picks it up automatically. Needs a `GeoInfo.band_nodata_values:
   list | None` field.

(1) is the smaller change. (2) is more invariant-friendly.

Out of scope for #2180 per the issue body's "wave-1 frozen signature"
constraint. Filed here so it does not get lost.

brendancol:
## Summary
- Add optional `band_nodata` / `band_nodata_values` kwargs to `_validate_read_geo_info` and `_finalize_lazy_read_attrs`.
- The VRT eager and chunked paths pass the same context they already pass to the inline pre-read call, so `_check_read_mixed_band_metadata` does real work on the helper-routed call instead of firing as a no-op.
- Non-VRT callers omit the kwargs and the mixed-band check short-circuits because `band_nodata_values` defaults to `None`.

The helpers are backend-agnostic; the change covers all four VRT-aware paths (numpy, cupy, dask+numpy, dask+cupy).

Closes #2210.

## Test plan
- [x] New tests in `test_helper_band_nodata_2210.py` cover the signature contract and the mixed-band rejection through both helpers.
- [x] Existing suites pass: `test_finalization_helpers_2162.py`, `test_lazy_finalization_parity_2162.py`, `test_vrt_finalization_parity_2162.py`, `test_mixed_band_metadata_fail_closed_1987.py`.
- [x] Full `xrspatial/geotiff/tests/` suite green except one pre-existing unrelated lz4 failure also on main.


brendancol (commented):
## PR Review: geotiff: forward band_nodata context through _validate_read_geo_info (#2210)

### Blockers
(none)

### Suggestions
- `band_nodata_values` is missing a type annotation in `_validate_read_geo_info` (xrspatial/geotiff/_attrs.py:788) and `_finalize_lazy_read_attrs` (xrspatial/geotiff/_attrs.py:1363). Neighbouring params have annotations; `list | None = None` matches the surrounding style.

### Nits
- The chunked VRT path builds `[b.nodata for b in vrt.bands]` twice -- inline pre-read at xrspatial/geotiff/_backends/vrt.py:726 and helper call at xrspatial/geotiff/_backends/vrt.py:1010. Same pattern on the eager path at lines 376 vs 576. Hoisting to a local cleans it up. Cost is negligible either way.

### What looks good
- Non-VRT callers (gpu.py, dask.py, `_finalize_eager_read`) keep working unchanged: the new kwargs default to `None` and `_check_read_mixed_band_metadata` short-circuits on a falsy `band_nodata_values`.
- The comments in vrt.py at lines 353-358 and 711-718 are rewritten to describe the helper-routed call as a defensive consistency check rather than a no-op.
- Test coverage targets what changed: signature pinning, mixed-band rejection via both helpers, the `band_nodata='first'` opt-out, and the non-VRT no-kwargs case.
- Keyword-only on both helpers leaves the wave-1 positional shape alone, so the parity tests in `test_finalization_helpers_2162.py` still pin it.

### Checklist
- [x] Algorithm matches reference/paper -- N/A (internal helper refactor)
- [x] All implemented backends produce consistent results -- shared helpers across the four VRT backends
- [x] NaN handling is correct -- unchanged
- [x] Edge cases are covered by tests -- mixed sentinels, single sentinel, opt-out, no kwargs
- [x] Dask chunk boundaries handled correctly -- N/A
- [x] No premature materialization or unnecessary copies -- N/A
- [x] README feature matrix updated (if applicable) -- N/A
- [x] Docstrings present and accurate -- both helpers' docstrings updated

brendancol (commented):
## Follow-up review (#2210)

The two findings from the previous pass are addressed in commit 775ca11:

- **Suggestion (type annotation)** -- `band_nodata_values: list | None = None` now on both helpers (xrspatial/geotiff/_attrs.py:788, 1363). Matches the surrounding signature style.
- **Nit (duplicate band-list build)** -- both VRT paths hoist the list to a local. Eager: `_band_nodata_values` shared between the inline pre-read and the `_finalize_lazy_read_attrs` call. Chunked: `band_nodata_values` similarly shared.

No new findings on this pass. Same 96-test suite still green.

Issue #2211, opened 2026-05-21
Refactor GeoTIFF internals around shared read/write contracts


brendancol:
## Problem

The GeoTIFF module has accumulated a large number of local fixes across very large files:

- `xrspatial/geotiff/_reader.py` is roughly 3.4k lines
- `xrspatial/geotiff/_writer.py` is roughly 2.8k lines
- `xrspatial/geotiff/_gpu_decode.py` is roughly 3.3k lines
- `xrspatial/geotiff/_compression.py` is roughly 1.9k lines

The test suite is extensive, but the implementation shape is now high-risk: behavior that should be one contract is often repeated or partially mirrored across eager NumPy, Dask, GPU, Dask+GPU, VRT, HTTP/fsspec, and writer paths. That makes regressions likely whenever a bug fix lands in one backend but misses another.

The highest-risk duplicated contracts are:

- transform inference and georef/no-georef/rotated-state handling
- nodata masking, `masked_nodata`, `nodata_pixels_present`, and dtype promotion
- metadata finalization and attrs contract population
- dispatcher/backend kwarg validation parity
- read/write round-trip invariants across eager, lazy, GPU, VRT, and COG paths

This is not about rewriting the module wholesale. The goal is to reduce regression surface by extracting shared contract boundaries and making each backend call those contracts instead of reimplementing them locally.

### Phase 1: Inventory and contract map

- Build a short internal map of current backend entry points and the contract steps each performs:
  - source/kwarg validation
  - metadata parse
  - transform/georef classification
  - pixel decode
  - orientation/photometric handling
  - nodata mask and dtype cast
  - attrs finalization
  - DataArray construction
- Identify which steps already have shared helpers and which are still duplicated.
- Add this map as developer documentation under `xrspatial/geotiff/` or `docs/source/reference/` so future patches have an intended flow to follow.

### Phase 2: Centralize transform/georef contract

- Extract one shared transform/georef resolver that owns:
  - coord-to-transform inference
  - no-georef marker behavior
  - degenerate axis policy
  - rotated-affine read-only/drop policy
  - `georef_status` derivation
- Make eager writer, VRT tiled writer, GPU writer, and any low-level write path call the same resolver.
- Make the resolver return a typed result rather than loosely coupled attrs/tuples.
- Add parity tests for `y/x`, `lat/lon`, `row/col`, no-georef, transform-only, CRS-only, and rotated-dropped cases.

### Phase 3: Centralize nodata lifecycle

[...]

PR #2230, opened 2026-05-21
geotiff: add backend full parity gate

### LINKED ISSUE #2229: Refactor GeoTIFF Phase 6: backend parity gate test (PR-F of #2211)
## Summary

Part of epic #2211. Phase 6 deliverable: a single table-driven test that runs the same fixture through every available backend and asserts pixel, coord, attrs, and dtype parity. This locks in the contracts the earlier phases established.

## Scope

Tests only. No production code changes.

## Files

- Create: `xrspatial/geotiff/tests/test_backend_full_parity_2211.py`.
- Reuse fixtures from `xrspatial/geotiff/tests/conftest.py` and `golden_corpus/`.

## Test matrix

- Backends: `{eager_numpy, dask_numpy, gpu, dask_gpu, vrt_eager, http_fsspec}`. Skip/xfail those that can't run in the current env, but the skip must be loud and explicit, not silent.
- For each fixture × backend, assert:
  - pixel values (`allclose` with appropriate tol)
  - dims and coords
  - transform/georef attrs
  - CRS attrs
  - nodata attrs and dtype
  - a curated set of rich metadata attrs

## Acceptance

- New file passes locally; CI green.
- All assertions are table-driven (parametrized), not N hand-rolled tests.
- GPU and Dask+GPU skip path is explicit and the skip reason is loud.

## Note for the implementer

This PR is the contract gate for the rest of the epic. If you find a real parity bug while writing the test, prefer xfail-with-issue over hiding it — opening a sub-issue is correct.

Closes part of #2211.

brendancol:
## Summary

Phase 6 of the GeoTIFF refactor epic (#2211): the parity gate that locks in the cross-backend invariants PRs A through E built.

- Adds `xrspatial/geotiff/tests/test_backend_full_parity_2211.py`. One table-driven test runs every golden-corpus fixture through six backends and checks the same battery on each result: pixel values (`allclose`, NaN-aware), dims and coords, transform/georef attrs, CRS attrs, nodata sentinel and dtype, plus a curated canonical-metadata subset.
- Backends: `eager_numpy`, `dask_numpy`, `gpu`, `dask_gpu`, `vrt_eager`, `http_fsspec`.
- GPU and Dask+GPU rows skip with a loud explicit reason when cupy + CUDA are missing, per the issue. No silent zero-collection in CI.
- Per-backend fixture skips cite the source of each divergence so a future refactor knows whether to fix the gap or move the skip.

Tests only. No production code changes.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_backend_full_parity_2211.py -v` passes locally (171 passed, 23 skipped on a CUDA host)
- [x] GPU rows fire and pass on a CUDA host (`-k gpu`, 59 passed)
- [x] `test_taxonomy_ids_are_in_manifest` catches stale skip-table ids
- [x] `test_gpu_skip_reason_is_loud` enforces the loud-skip contract

Closes #2229.
Part of #2211.

### DIFF
--- a/xrspatial/geotiff/tests/test_backend_full_parity_2211.py
+++ b/xrspatial/geotiff/tests/test_backend_full_parity_2211.py
@@ -1,0 +1,787 @@
+"""Backend full-parity gate (PR-F of #2211 / closes #2229).
+
+A single table-driven parity test that runs the same fixture through
+every read backend ``xrspatial.geotiff`` ships and asserts the full
+observable contract on each output: pixel values, dims and coords,
+transform/georef attrs, CRS attrs, nodata sentinel and dtype, and a
+curated set of canonical metadata attrs.
+
+This module is the **contract gate** for the geotiff refactor epic
+#2211. Earlier phases (PRs A through E) shared finalization helpers
+across the eager and lazy backends so the four CPU/GPU x eager/dask
+paths produce the same DataArray. PR-F locks that result in: any drift
+between backends on the curated attrs set raises here, so a refactor
+in one backend that silently changes an attr cannot land without an
+update to this test.
+
+Design notes
+------------
+
+* Backends covered (per #2229):
+    - ``eager_numpy``  -- ``open_geotiff(path)``
+    - ``dask_numpy``   -- ``open_geotiff(path, chunks=...)``
+    - ``gpu``          -- ``open_geotiff(path, gpu=True, on_gpu_failure='strict')``
+    - ``dask_gpu``     -- ``open_geotiff(path, gpu=True, chunks=..., on_gpu_failure='strict')``
+    - ``vrt_eager``    -- ``open_geotiff(vrt_path)`` (one-source VRT
+                          generated per fixture)
+    - ``http_fsspec``  -- ``open_geotiff('memory:///corpus/<id>.tif')``
+                          through fsspec's in-process memory filesystem
+* Fixtures come from the existing golden corpus
+  (``xrspatial/geotiff/tests/golden_corpus/``) so the fixture surface
+  is the same one Phase 3 (#1930) exercised. The set is filtered to
+  fixtures the full parity matrix can compare cleanly today; known
+  divergences (e.g. JPEG-YCbCr axis order, MinIsWhite inversion) live
+  in ``_INTENTIONAL_SKIPS`` with citations.
+* The reference is the ``eager_numpy`` read of the same fixture, taken
+  once per fixture and reused across backends. Pixels are compared
+  ``allclose`` for float dtypes (NaN-aware) and bit-exact for ints.
+* GPU and Dask+GPU rows skip with a **loud** reason when cupy or CUDA
+  is missing: the issue text calls this out explicitly so a silent
+  collection of zero GPU tests in CI is itself a bug.
+
+Adding a new backend or a new fixture
+-------------------------------------
+
+To add a backend: append a ``_Backend`` to ``_BACKENDS`` with its
+``read`` callable, an availability predicate, and the list of fixture
+ids it cannot handle (with citations). The parametrize matrix lights
+it up on the next run.
+
+To add a fixture: drop it into the manifest at
+``golden_corpus/manifest.yaml`` and regenerate. The matrix picks it
+up automatically.
+"""
+from __future__ import annotations
+
+import importlib.util
+import pathlib
+from dataclasses import dataclass, field
+from typing import Any, Callable
+
+import numpy as np
+import pytest
+import xarray as xr
+
+pytest.importorskip("yaml")
+pytest.importorskip("rasterio")
+
+from xrspatial.geotiff import open_geotiff, write_vrt  # noqa: E402
+from xrspatial.geotiff.tests.golden_corpus import generate  # noqa: E402
+from xrspatial.geotiff.tests.golden_corpus._marks import (  # noqa: E402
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: geotiff: add backend full parity gate

None.

- [ ] `_assert_nodata_attrs` line ~598. The NaN-equality short-circuit only fires for Python `float` instances. A nodata of type `numpy.float32(nan)` or `numpy.float64(nan)` fails `isinstance(ref_nd, float)`, falls through to plain `==`, and raises spuriously because NaN compares unequal to itself. xrspatial stamps a Python float today so the gap is latent, but a future writer that preserves the numpy scalar type would trip it. Drop the isinstance gate and use `np.isnan(np.asarray(ref_nd, dtype=float))` inside a try/except for non-numeric sentinels.

- [ ] `_assert_pixels_close` line ~487. `atol=1e-7` is loose for float64. The comment cites a "ULP-level perturbation", but the float64 ULP near 1.0 is ~2e-16, not 1e-7, and for data with magnitudes well below 1 a real disagreement of 5e-8 would pass. Two reasonable options: tighten to `rtol=1e-12, atol=0` so the comparison scales with the data, or keep the absolute floor and document the rationale (e.g. integer-source data round-tripped to float through rasterio).

- [ ] No assertion that the backend actually returned what it was asked for. The corpus tests have `test_gpu_candidate_is_actually_on_device` and `test_dask_candidate_is_actually_chunked` because `open_geotiff` can silently drop a kwarg and return a numpy array when the caller asked for `gpu=True`. This module would pass that case because the pixel comparison against the eager-numpy reference still succeeds. Two cheap assertions on one fast fixture would close it: `isinstance(da.data, cupy.ndarray)` for the gpu row, `hasattr(da.data, "dask")` plus a >1-block chunk-grid check for the dask rows.

- [ ] `_vrt_cache_dir` line 318. `hash(str(fixtures_dir))` uses Python's salted string hash, so the cache directory name changes every process via `PYTHONHASHSEED` randomization. The "cache invalidation when fixtures dir changes" comment claims cross-run stability the cache does not have. Functionally fine since the test only needs intra-process cache hits; reword the comment or switch to `hashlib.sha1` for a stable digest.

- [ ] `_read_http_fsspec` does not clean up the fsspec memory filesystem between cells. The corpus fsspec test uses a `memory_fs_clean` fixture for the same reason. Keys here are unique per fixture id so collisions are unlikely, but stale bytes pile up over the module run. A `try/finally` to delete the key, or a session-scoped autouse cleanup, would tighten it up.

- [ ] `_reference_cache` is module-scoped and never evicts. Eager reads in the current corpus are small so memory is bounded, but if the corpus grows substantially the cache becomes a tax under `pytest -n auto`.

- [ ] The top-of-file "Design notes" docstring lists the six backends and their kwargs but does not flag that `eager_numpy` is also the reference everything else is compared against. That is the most important thing about the design and worth one sentence.

### What looks good

- The parametrize axes (backends, fixtures) are independent, so a new backend lights up against every fixture and a new fixture against every backend.
- Loud GPU skip reasons match the issue text verbatim. `test_gpu_skip_reason_is_loud` is a nice belt-and-braces.
- Per-fixture skips cite their source (sibling test or issue link), so a future contributor can tell at a glance whether the skip is a real gap or by-design.
- VRT path-containment workaround (copy source into the cache dir rather than mutating `XRSPATIAL_VRT_ALLOWED_ROOTS`) is the right call.
- `test_taxonomy_ids_are_in_manifest` catches typos in the skip tables, same guard the corpus modules ship with.

### Checklist

- [x] Algorithm matches reference/paper. Test-only PR.
- [x] All implemented backends produce consistent results. That is what the new test asserts.
- [x] NaN handling is correct. See the numpy-scalar gap above.
- [x] Edge cases covered by tests. Corpus fixtures span the surface.
- [x] Dask chunk boundaries handled. `_CHUNK_SIZE = 32` exercises 2x2 and 1x1 grids.
- [x] No premature materialization. Single `.compute() / .get()` per cell.
- [x] Benchmark exists or is not needed. Not needed.
- [x] README feature matrix updated. N/A.
- [x] Docstrings present and accurate. See the `hash(str(...))` comment above.

brendancol (commented):
## PR Review (follow-up): geotiff: add backend full parity gate

### Blockers
None.

### Suggestions
None.

### Nits

Now addressed in 3cd8539: the `_assert_pixels_close` comment block now explicitly documents that `atol=0` is intentional and tells a future contributor not to loosen it without confirming the new fixture's expectations.

### What looks good

- `_is_nan_sentinel` accepts numpy scalars and falls through cleanly on non-numeric values. Closes the latent gap from the previous pass.
- `hashlib.sha1` digest in `_vrt_cache_dir` is stable across processes; the comment names the security non-claim explicitly.
- The three storage-type sanity tests catch the failure mode the parity battery is structurally blind to (silent kwarg drop returning the wrong storage type).
- fsspec memory cleanup is best-effort and the `try/except FileNotFoundError` is the right shape.
- Top-of-file "Design notes" now flags `eager_numpy` as the reference.

### Checklist

- [x] Algorithm matches reference/paper. Test-only PR.
- [x] All implemented backends produce consistent results.
- [x] NaN handling is correct, including numpy scalar nodata.
- [x] Edge cases covered.
- [x] Dask chunk boundaries handled.
- [x] Benchmark exists or is not needed.
- [x] README feature matrix updated. N/A.

Clean. No open items.

Issue #2226, opened 2026-05-21
Refactor GeoTIFF Phase 3: centralize nodata lifecycle (PR-C of #2211)


brendancol:
## Summary

Part of epic #2211. Phase 3 deliverable: extract a single nodata lifecycle helper that owns all the decisions around raw sentinel, effective sentinel post-photometric, masking, dtype promotion, and writer restore. Backend kernels keep their array op; the helper owns the policy.

## Scope

Refactor + parity tests. No public API change.

## Files

- Create: `xrspatial/geotiff/_nodata.py` — lifecycle helper.
- Modify: `xrspatial/geotiff/_reader.py`, `_writer.py`, `_vrt.py` — replace inline nodata logic with helper calls.
- Modify: `xrspatial/geotiff/_backends/dask.py`, `_backends/gpu.py`, `_backends/vrt.py`, `_backends/_gpu_helpers.py`, `_writers/eager.py`, `_writers/gpu.py`.
- Create: `xrspatial/geotiff/tests/test_nodata_lifecycle_parity_2211.py`.

## Lifecycle helper contract

`NodataLifecycle(declared, photometric, dtype_in, dtype_request)` returns a value object exposing:

- `raw_sentinel`
- `effective_sentinel` (post-photometric)
- `pixels_present` slot (set by execution, exposed by helper)
- `masking_occurred`
- `dtype_cast_occurred`
- `writer_restore_sentinel()` — decision used by writer paths

Backend kernels still own the array op; the helper owns all decisions.

## Parity tests

Cover at minimum:
- integer sentinel, float sentinel, NaN sentinel
- out-of-range sentinel
- MinIsWhite photometric inversion
- `mask_nodata=False`
- explicit `dtype=` kwarg

Across: eager NumPy, Dask, GPU (skip if no CUDA), VRT eager.

## Acceptance

- Inline `if nodata is not None` branches in backends collapse to one helper call for the decision; the array op stays backend-local.
- `test_nodata_lifecycle_attrs_2135.py`, `test_nodata_semantics_split_1988.py`, `test_apply_nodata_mask_gpu_inplace_1934.py` still pass.
- No public API change.

## Note for the implementer

PR-E (sources extraction) is held until this PR merges to avoid simultaneous heavy edits to `_reader.py`. Keep your touches to `_reader.py` scoped to the nodata changes.

Closes part of #2211.

Issue #2225, opened 2026-05-21
Refactor GeoTIFF Phase 2: centralize transform/georef contract (PR-B of #2211)


brendancol:
## Summary

Part of epic #2211. Phase 2 deliverable: one shared transform/georef resolver that every backend (read and write) calls, with parity tests across the supported cases. Replaces hand-copied coord-to-transform and georef classification logic that currently lives in multiple modules.

## Scope

Refactor + parity tests. No public API change.

## Files

- Modify: `xrspatial/geotiff/_coords.py` — promote/centralize `coords_to_transform`, add `resolve_georef()` that returns a typed result.
- Modify: `xrspatial/geotiff/_attrs.py` — `georef_status` derivation routes through `resolve_georef()`.
- Modify: `xrspatial/geotiff/_writers/eager.py`, `_writers/gpu.py`, `_writers/vrt.py` — call the resolver instead of inline logic.
- Modify: `xrspatial/geotiff/_backends/vrt.py` — call the resolver.
- Create: `xrspatial/geotiff/tests/test_georef_resolver_parity_2211.py`.

## Resolver contract

- Input: a DataArray, or coord arrays + crs hint + transform hint.
- Output: a `GeorefResolution` dataclass:
  - `transform: GeoTransform | None`
  - `georef_status: Literal["none","coords","transform-only","crs-only","rotated-dropped","full"]`
  - `dropped_reason: str | None`
  - `applied_no_georef_marker: bool`
- Owns: coord-to-transform inference, no-georef marker behavior, degenerate axis policy, rotated-affine read-only/drop policy.

## Parity tests

Cover at minimum:
- `y`/`x` dim names
- `lat`/`lon` dim names
- `row`/`col` dim names (no georef)
- transform-only (no CRS)
- CRS-only (no transform)
- rotated-affine dropped on read

For each case, assert identical `GeorefResolution` produced from every supported call site.

## Acceptance

- All 6 fixture cases produce identical `GeorefResolution` across eager/Dask/VRT call sites.
- `xrspatial/geotiff/tests/test_georef_status_2136.py` and the `test_allow_rotated_*` suites still pass.
- No call site outside `_coords.py` / `_attrs.py` builds a `GeoTransform` by hand on the read or write paths covered.
- No public API change.

Closes part of #2211.

PR #2205, opened 2026-05-20
geotiff: share lazy finalization helper across dask backends (PR C of #2162)

brendancol:
## Summary

Wave 2 of #2162. Both dask read paths had a 25-line validate-populate-stamp block that was identical except for a couple of comment differences. Replace each with one call to `_finalize_lazy_read_attrs` from #2177 (now on main) so a future fix lands in both backends at once.

- `xrspatial/geotiff/_backends/dask.py`: CPU+dask finalization in `read_geotiff_dask` now goes through the helper.
- `xrspatial/geotiff/_backends/gpu.py`: GPU+dask finalization in the dask branch of `read_geotiff_gpu` now goes through the helper. The three eager GPU sites in this file are out of scope (sibling PR D, #2179).

There is one wrinkle. The helper's `dtype` argument is used for two things at once: the resolved graph dtype (which drives `masked_nodata`) and the caller's cast attr (which drives `nodata_dtype_cast`). The dask path has to keep those separate, because `mask_nodata=True` on an integer source auto-promotes the graph dtype to `float64` without the caller asking, and that auto-promotion must not leak out as `nodata_dtype_cast`. The migration passes the resolved `target_dtype` / `declared_dtype` so `masked_nodata` is right, then fixes up `nodata_dtype_cast` at the call site to match the pre-helper contract.

Preserved behavior:

- `attrs['nodata_pixels_present']` stays absent on lazy outputs (#2135 contract).
- `attrs['nodata_dtype_cast']` is recorded only when the caller passed `dtype=`.
- MinIsWhite inversion (`dask.py:227-238`) still preempts the sentinel before chunk tasks see it; `nodata_attr` (the pre-inversion value) is what reaches `attrs['nodata']`.

Closes #2178. Depends on #2177 (merged).

## Test plan

- [x] New `test_lazy_finalization_parity_2162.py` (19 tests) covering:
  - `nodata_pixels_present` absent on both backends.
  - `nodata_dtype_cast` matches across backends when caller forces a cast.
  - `nodata_dtype_cast` absent on both backends when the graph dtype was auto-promoted by masking.
  - `georef_status` matches across both backends for full / transform_only / crs_only / none / rotated_dropped.
  - Full attrs-dict parity (modulo backend markers) per fixture.
- [x] All 91 existing tests in `test_nodata_lifecycle_attrs_2135.py`, `test_nodata_semantics_split_1988.py`, `test_georef_status_2136.py` pass.
[...]

PR #2232, opened 2026-05-21
Refactor GeoTIFF Phase 4: complete metadata finalization centralization

### LINKED ISSUE #2227: Refactor GeoTIFF Phase 4: complete metadata finalization centralization (PR-D of #2211)
## Summary

Part of epic #2211. Phase 4 deliverable: finish the work started in #2162 — move every remaining attrs-population call into `GeoTIFFMetadata` / the finalization helpers, so every read backend stamps the same canonical keys for equivalent inputs.

## Scope

Refactor + table-driven attrs parity test. No public API change.

## Files

- Modify: `xrspatial/geotiff/_attrs.py` — make `GeoTIFFMetadata.finalize()` (and its lazy variant) the single attrs entry point. Absorb any per-backend attrs population still living in `_reader.py`, `_vrt.py`, `_backends/*`.
- Modify: each backend module to call `finalize()` (or the lazy variant) rather than building attrs inline.
- Create: `xrspatial/geotiff/tests/test_attrs_finalization_parity_2211.py` — table-driven comparison across backends, excluding documented backend-specific keys.

## Acceptance

- Every attrs key defined in `_attrs.py`'s canonical schema is stamped by `finalize()` / the lazy helper, not by ad-hoc `da.attrs[k] = v` in backends.
- Table-driven parity test asserts identical attrs across eager / Dask / GPU (when available) / VRT for the same input fixture.
- `test_attrs_contract_*.py` and `test_attrs_parity_1548.py` still pass.
- No public API change.

## Prior art

PRs #2200, #2205, #2207, #2209 from epic #2162 established the shared finalization helpers for eager + dask + VRT. This PR finishes the migration so no backend writes attrs directly.

Closes part of #2211.

brendancol:
Closes #2227
Part of #2211

## Summary

Finishes the migration started by PRs #2200, #2205, #2207, #2209 from epic #2162. After this PR, every read backend assembles its attrs through one of two finalization entry points in `xrspatial/geotiff/_attrs.py`:

* `_finalize_eager_read` for the eager numpy + GPU paths.
* `_finalize_lazy_read_attrs` for the dask, dask+GPU, and VRT paths.

## Changes

* `_attrs.py`: add an optional `pixels_present` kwarg to `_finalize_lazy_read_attrs`. The default stays `None` so the dask backends keep the lazy contract from #2135 (a strict per-chunk reduction would force `.compute()`).
* `_backends/vrt.py`: the eager VRT path now threads its VRT-aware `nodata_pixels_present` scan through the helper's new kwarg instead of writing the attr after the call.
* `_backends/gpu.py`: drop stale imports of `_populate_attrs_from_geo_info`, `_set_nodata_attrs`, and `_validate_read_geo_info`. The GPU paths route every attrs write through the two top-level helpers.
* New test: `test_attrs_finalization_parity_2211.py` is a table-driven parity check across eager numpy, dask+numpy, GPU (when available), dask+GPU (when available), and VRT. Four fixture shapes (plain float, float with sentinel, integer with sentinel, plain uint8). Backend-specific keys (`vrt_holes`, `nodata_pixels_present`, TIFF tag pass-throughs) are carved out in a single named frozenset so a future migration that promotes one of those keys to all backends has one place to update.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy, VRT. GPU branches are gated on `cupy + CUDA` availability so the test still runs on CPU-only hosts.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_attrs_finalization_parity_2211.py` passes (9/9)
- [x] `pytest xrspatial/geotiff/tests/test_attrs_parity_1548.py test_attrs_contract_*.py test_finalization_helpers_2162.py test_eager_finalization_parity_2162.py test_lazy_finalization_parity_2162.py test_vrt_finalization_parity_2162.py` all green (129 tests)
- [x] Full geotiff suite: 4721 passed, 1 unrelated pre-existing failure in `test_lowlevel_write_pushdown_2138.py[lz4]` (experimental codec gate, untouched by this PR)

## Coordination

PR-B (#2225) is also touching `_attrs.py` (georef_status). PR-C (#2226) is touching `_reader.py`/backends for nodata. PR-F (#2229) is adding a parity test. I kept the `_attrs.py` edit here to one small added kwarg so a rebase against any of those branches stays cheap.

## Not in scope

* No public API change.
* `test_attrs_contract_*.py` and `test_attrs_parity_1548.py` were left untouched and still pass.

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -1353,6 +1353,7 @@ def _finalize_lazy_read_attrs(
     band_nodata: str | None = None,
     band_nodata_values: list | None = None,
     attrs_in: dict | None = None,
+    pixels_present: bool | None = None,
 ):
     """Validate and populate attrs for dask-style lazy reads.
 
@@ -1369,10 +1370,14 @@ def _finalize_lazy_read_attrs(
     3. :func:`_set_nodata_attrs` -- ``masked`` is True iff the caller
        opted into masking AND the graph dtype is float.
        ``dtype_cast`` is recorded when ``caller_dtype`` is not ``None``
-       and ``nodata`` is declared. ``pixels_present=None`` is the
-       documented dask contract from issue #2135: a strict per-chunk
-       reduction would force an eager ``.compute()`` and break the
-       lazy contract, so the attr is left absent on lazy outputs.
+       and ``nodata`` is declared. ``pixels_present`` defaults to
+       ``None`` per the documented dask contract from issue #2135 (a
+       strict per-chunk reduction would force eager ``.compute()`` and
+       break the lazy contract, so the attr stays absent on lazy
+       outputs). Eager-VRT callers that already computed the presence
+       bool via a single decode pass can pass it through this kwarg so
+       the attr is stamped via the same finalization helper rather
+       than written ad-hoc post-call (PR-D of #2211).
 
     Returns the attrs ``dict`` only; the caller assembles the dask graph
     and builds the :class:`xarray.DataArray` itself, so this helper
@@ -1448,7 +1453,7 @@ def _finalize_lazy_read_attrs(
     _set_nodata_attrs(
         attrs, nodata,
         masked=masked,
-        pixels_present=None,
+        pixels_present=pixels_present,
         dtype_cast=dtype_cast_attr,
     )
 
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -23,9 +23,6 @@ import xarray as xr
 from .._attrs import (
     _finalize_eager_read,
     _finalize_lazy_read_attrs,
-    _populate_attrs_from_geo_info,
-    _set_nodata_attrs,
-    _validate_read_geo_info,
 )
 from .._coords import (
     coords_from_geo_info as _coords_from_geo_info,
--- a/xrspatial/geotiff/_backends/vrt.py
+++ b/xrspatial/geotiff/_backends/vrt.py
@@ -558,8 +558,11 @@ def read_vrt(source: str, *,
     # * Per-band nodata selection happens above this call, not inside
     #   the helper. ``_apply_eager_nodata_mask`` is single-sentinel.
     # * ``nodata_pixels_present`` is computed above (VRT-aware scan)
-    #   and stamped post-helper because ``_finalize_lazy_read_attrs``
-    #   passes ``pixels_present=None`` unconditionally.
+    #   and threaded through the helper's ``pixels_present`` kwarg so
+    #   the attr is stamped by ``_set_nodata_attrs`` rather than
+    #   written ad-hoc post-call (PR-D of #2211). The kwarg defaults
+    #   to ``None`` so the dask backends keep the lazy contract from
+    #   issue #2135 unchanged.
     synth_geo_info = _vrt_to_synthetic_geo_info(vrt)
     attrs_seed: dict = {}
     if vrt.holes:
@@ -576,9 +579,8 @@ def read_vrt(source: str, *,
         band_nodata=band_nodata,
         band_nodata_values=_band_nodata_values,
         attrs_in=attrs_seed,
+        pixels_present=nodata_pixels_present,
     )
-    if nodata is not None and nodata_pixels_present is not None:
-        attrs['nodata_pixels_present'] = bool(nodata_pixels_present)
 
     if arr.ndim == 3:
         dims = ['y', 'x', 'band']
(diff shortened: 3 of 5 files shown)


brendancol (commented):
## PR Review: GeoTIFF Phase 4 attrs finalization centralization

### Blockers
None.

### Suggestions

- [ ] `xrspatial/geotiff/tests/test_attrs_finalization_parity_2211.py:19` — docstring still says "MinIsWhite photometric" but the fixture was renamed to `uint8_no_nodata`. Either restore MinIsWhite coverage or update the description to match the actual fixture.
- [ ] `xrspatial/geotiff/_attrs.py:1356` — the new `pixels_present` kwarg has no direct unit test in `test_finalization_helpers_2162.py`. The lazy "pixels_present absent by default" assertions stay green, but a positive test (pass `pixels_present=True/False`, check the attr lands) would prevent a future regression where the kwarg silently stops being threaded through to `_set_nodata_attrs`. The parity test covers it transitively via VRT, which is OK but indirect.

### Nits

- [ ] `test_attrs_finalization_parity_2211.py:122` — `_attrs_close` declares `rtol` as a parameter but only consults it inside the `transform` branch. Either drop the parameter or apply it to every numeric key for consistency. The docstring says "all other keys must match exactly" so dropping it is the cleaner option.
- [ ] `test_attrs_finalization_parity_2211.py:67-70` — the 3-D dims branch (`['y', 'x', 'band']`) in `_coord_array` is dead code since none of the fixtures pass 3-D arrays. Either drop the branch or add a multi-band fixture.
- [ ] `test_attrs_finalization_parity_2211.py:259` — `_open_vrt` calls `open_geotiff(path)` to seed the VRT XML. A silent drift in `open_geotiff` would propagate into the VRT helper and only surface as a VRT-vs-dask divergence rather than a clean failure. Reading the values from the source DataArray that `_write_*` returns (or accepting them as parameters) would isolate the VRT path from the eager path.

### What looks good

- The diff is exactly what #2227 called for: one new kwarg with a backward-compatible default, one inline write removed, three stale imports cleaned up.
- The lazy contract from #2135 is preserved — `pixels_present` defaults to `None` and all existing `'nodata_pixels_present' not in out.attrs` assertions in `test_lazy_finalization_parity_2162.py` still pass.
- The carveout for backend-specific keys lives in one named `frozenset` with a self-test against the docstring, so a future maintainer adding a new key has a single place to update.
- The full geotiff suite stays green (the one failure in `test_lowlevel_write_pushdown_2138.py[lz4]` is the experimental-codec gate and predates this PR).

### Checklist

- [x] Refactor matches issue #2227 scope
- [x] No public API change; existing contract tests still pass
- [x] Backend keysets compared in both directions (value + key set)
- [x] Lazy/dask contract from #2135 preserved
- [x] Stale imports removed
- [ ] Direct unit coverage for the new `pixels_present` kwarg (see Suggestion 2)

brendancol (commented):
## PR Review (round 2): GeoTIFF Phase 4 attrs finalization centralization

Re-review after commit e30ab49.

### Blockers
None.

### Suggestions
None.

### Nits
None blocking. The previous round's nits are all addressed.

- Suggestion 1 (docstring drift): fixed. Module docstring now matches the `uint8_no_nodata` fixture name.
- Suggestion 2 (missing direct unit test for `pixels_present` kwarg): fixed. Four new tests in `test_finalization_helpers_2162.py` cover the True / False / None / "nodata=None ignores stray forward" cases.
- Nit 1 (`_attrs_close` rtol parameter): fixed. Tolerance is now `_TRANSFORM_RTOL` / `_TRANSFORM_ATOL` module constants and only applies to the `transform` key, matching the docstring.
- Nit 2 (dead 3-D branch): fixed. `_coord_array` now asserts 2-D input.
- Nit 3 (`_open_vrt` calling `open_geotiff`): fixed. The VRT helper takes a `_FixtureMeta` from the writer and reads geometry from module-level constants. The same constants seed `_coord_array`, so the VRT path and the TIFF path now share one source of truth without going through `open_geotiff`.

### What looks good

- The fixture-geometry constants are paired with an in-function assert (`_FIX_HEIGHT` / `_FIX_WIDTH` vs `arr.shape`), so a future fixture that uses a different shape fails loudly rather than silently producing a mis-sized VRT.
- The half-pixel shift in `_open_vrt` (corner-based GDAL convention vs center-based xarray coords) is explicit and explained in a comment, so a future maintainer can tell the offset is intentional rather than a copy-paste bug.
- The new `pixels_present=True` / `False` / `None` / `nodata=None` test matrix pins all four states the kwarg can land in, so a future regression where the kwarg stops being threaded through will fail loudly.

### Checklist

- [x] Refactor matches issue #2227 scope
- [x] No public API change; existing contract tests still pass
- [x] Backend keysets compared in both directions (value + key set)
- [x] Lazy/dask contract from #2135 preserved
- [x] Direct unit coverage for the new `pixels_present` kwarg
- [x] VRT helper isolated from `open_geotiff`
- [x] Stale imports removed

Issue #2228, opened 2026-05-21
Refactor GeoTIFF Phase 5a: extract _sources.py (PR-E of #2211)


brendancol:
## Summary

Part of epic #2211. Phase 5a deliverable: mechanical extraction of source classes and SSRF/range helpers out of `_reader.py` into a new `_sources.py`.

## Scope

Mechanical extraction. Behavior-neutral.

**Hold for dispatch until PR-C (#2226) merges** — both touch `_reader.py` heavily and parallel work would force a large rebase.

## Files

- Create: `xrspatial/geotiff/_sources.py` — local mmap source, HTTP source, fsspec source, BytesIO source, SSRF / byte-range helpers.
- Modify: `xrspatial/geotiff/_reader.py` — remove the classes/helpers being moved; import them from `_sources`.
- Modify: any other module that imports source classes (likely `_vrt.py`, `_backends/vrt.py`).

## Acceptance

- `_reader.py` line count drops by the size of the extracted classes (rough target: ~500–800 lines).
- All existing source-related tests pass unchanged: `test_bytesio_source.py`, `test_cog_http_*.py`, `test_cloud_read_byte_limit_1928.py`, `test_fsspec_*.py`, SSRF tests.
- Public import paths preserved (`from xrspatial.geotiff import ...`).
- No public API change.

## Non-goals

- `_decode.py`, `_layout.py`, `_write_layout.py`, `_nodata.py` (separate), `_georef.py` (already extracted in PR-B). These are deferred to follow-up issues.

Closes part of #2211.

PR #2234, opened 2026-05-21
geotiff: centralize nodata lifecycle

### LINKED ISSUE #2226: Refactor GeoTIFF Phase 3: centralize nodata lifecycle (PR-C of #2211)

brendancol:
## Summary

Adds `xrspatial/geotiff/_nodata.py` housing `NodataLifecycle`, a value
object that captures the nodata-related decisions the read and write
backends were each making inline:

- `raw_sentinel` and `effective_sentinel` (post-MinIsWhite inversion)
- `sentinel_fits_buffer` (finite + integer + in-range gate for integer dtypes)
- `masking_occurred` (the value `attrs['masked_nodata']` should carry)
- `dtype_cast_occurred` / `cast_dtype_name` (the `attrs['nodata_dtype_cast']` signal)
- `writer_restore_sentinel` (gates the writer's NaN-to-sentinel rewrite, including the `masked_nodata=False` opt-out from #1988)
- `pixels_present` slot (set by execution after the per-buffer scan)

The backends keep their array operations; the helper owns the
decisions. Affected sites:

- `_backends/dask.py`: the MinIsWhite-inversion block and the integer
  auto-promotion gate collapse to `effective_sentinel` and
  `sentinel_fits_buffer`.
- `_backends/gpu.py`: the post-MinIsWhite sentinel resolution moves
  to `effective_sentinel` instead of reaching into `_reader`.
- `_backends/_gpu_helpers.py`: `_apply_nodata_mask_gpu` uses the
  shared `_sentinel_fits_dtype` gate.
- `_writers/eager.py` and `_writers/gpu.py`: the NaN-to-sentinel
  rewrite gate (none / non-float / NaN sentinel / `masked_nodata=False`)
  routes through `writer_restore_sentinel`.

No public API change. The existing regression tests
(`test_nodata_lifecycle_attrs_2135.py`, `test_nodata_semantics_split_1988.py`,
`test_apply_nodata_mask_gpu_inplace_1934.py`) keep passing.

`_reader.py` and `_vrt.py` are left untouched in this PR to keep
PR-E (#2228, sources extraction) rebasing cheaply: those modules
already use `_miniswhite_inverted_nodata` and per-band sentinel
helpers respectively, and the lifecycle value object mirrors the
same math without forcing those large files into the diff.

## Tests

Adds `test_nodata_lifecycle_parity_2211.py` (54 cases):

- direct unit tests against every `NodataLifecycle` property and decision
- cross-backend parity for integer / float / NaN / out-of-range
  sentinels through eager NumPy, Dask, GPU (skip if no CUDA), and
  VRT eager
- MinIsWhite photometric inversion end-to-end, plus a
  property-equivalence check against `_reader._miniswhite_inverted_nodata`
- `mask_nodata=False` parity across eager / Dask / GPU
- explicit `dtype=` kwarg parity across eager / Dask
- writer-side restore: NaN restored to sentinel by default, opt-out
  via `attrs['masked_nodata']=False` honoured, GPU writer matches

GPU skips report the explicit reason (`cupy + CUDA required for
NodataLifecycle GPU parity (PR-C #2226)`) so a missing CUDA
environment is visible rather than silently passing.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_nodata_lifecycle_parity_2211.py` (54 pass)
- [x] `pytest xrspatial/geotiff/tests/test_nodata_lifecycle_attrs_2135.py xrspatial/geotiff/tests/test_nodata_semantics_split_1988.py xrspatial/geotiff/tests/test_apply_nodata_mask_gpu_inplace_1934.py` (65 pass)
- [x] `pytest xrspatial/geotiff/tests/` (4702 pass; one pre-existing lz4 failure on `main` unrelated to this PR)

Closes #2226
Part of #2211

### DIFF
--- a/xrspatial/geotiff/_backends/_gpu_helpers.py
+++ b/xrspatial/geotiff/_backends/_gpu_helpers.py
@@ -22,6 +22,7 @@ from .._coords import (
     geo_to_coords as _geo_to_coords,
 )
 from .._geotags import GeoTransform, RASTER_PIXEL_IS_POINT
+from .._nodata import _sentinel_fits_dtype
 from .._runtime import _geotiff_strict_mode
 
 
@@ -79,8 +80,7 @@ def _apply_nodata_mask_gpu(arr_gpu, nodata):
         # fractional sentinels short-circuit to a no-op so
         # ``attrs['nodata']`` still records them while masking
         # leaves the integer buffer untouched.
-        from .._nodata import _sentinel_fits_dtype as _fits
-        if not _fits(nodata, arr_dtype):
+        if not _sentinel_fits_dtype(nodata, arr_dtype):
             return arr_gpu
         nodata_int = int(nodata)
         sentinel = arr_dtype.type(nodata_int)
--- a/xrspatial/geotiff/_backends/_gpu_helpers.py
+++ b/xrspatial/geotiff/_backends/_gpu_helpers.py
@@ -73,25 +73,16 @@ def _apply_nodata_mask_gpu(arr_gpu, nodata):
                          arr_dtype.type('nan'))
         return arr_gpu
     if arr_dtype.kind in ('u', 'i'):
-        # Out-of-range sentinels (e.g. uint16 + GDAL_NODATA="-9999") cannot
-        # match any decoded pixel; skip the cast that would otherwise raise
-        # OverflowError. A non-finite sentinel ("NaN" / "Inf" GDAL_NODATA
-        # strings) also cannot match an integer pixel and would raise
-        # ValueError on ``int(nodata)``; gate on ``np.isfinite`` first to
-        # mirror ``_resolve_masked_fill`` in ``_reader.py`` (#1774). A
-        # fractional sentinel (e.g. ``"3.5"`` on a ``uint16`` file) also
-        # cannot match an integer pixel and ``int(3.5)`` would truncate
-        # to 3, silently masking a real pixel value; gate on
-        # ``float(nodata).is_integer()`` as well (mirrors the
-        # ``_writer.py`` / ``_vrt.py`` pattern used for #1564 / #1616).
-        # attrs['nodata'] is still set by the caller so the original
-        # sentinel survives a write round-trip.
-        if not (np.isfinite(nodata) and float(nodata).is_integer()):
+        # PR-C #2226: lifecycle helper owns the
+        # finite / integer / in-range gate previously inlined here
+        # (#1774, #1564, #1616). Out-of-range, non-finite, or
+        # fractional sentinels short-circuit to a no-op so
+        # ``attrs['nodata']`` still records them while masking
+        # leaves the integer buffer untouched.
+        from .._nodata import _sentinel_fits_dtype as _fits
+        if not _fits(nodata, arr_dtype):
             return arr_gpu
         nodata_int = int(nodata)
-        info = np.iinfo(arr_dtype)
-        if not (info.min <= nodata_int <= info.max):
-            return arr_gpu
         sentinel = arr_dtype.type(nodata_int)
         mask = arr_gpu == sentinel
         if bool(mask.any().item()):
--- a/xrspatial/geotiff/_backends/dask.py
+++ b/xrspatial/geotiff/_backends/dask.py
@@ -237,52 +237,33 @@ def read_geotiff_dask(source: str, *,
         geo_info, full_h, full_w, file_dtype, n_bands = _read_geo_info(
             source, overview_level=overview_level,
             allow_rotated=allow_rotated)
-    nodata = geo_info.nodata
-    nodata_attr = nodata  # original sentinel preserved for attrs['nodata']
-    # When the source is MinIsWhite (photometric == 0, samples_per_pixel == 1),
-    # the per-chunk reader inverts pixel values before this closure's nodata
-    # mask runs. Track the inverted sentinel so the mask compares against the
-    # post-inversion value, not the original (#1809).
-    if nodata is not None:
-        _phm = getattr(geo_info, '_ifd_photometric', None)
-        _spp = getattr(geo_info, '_ifd_samples_per_pixel', None)
-        if _phm == 0 and _spp == 1:
-            if file_dtype.kind == 'u' and np.isfinite(nodata) and \
-                    float(nodata).is_integer():
-                vi = int(nodata)
-                info = np.iinfo(file_dtype)
-                if info.min <= vi <= info.max:
-                    nodata = info.max - vi
-            elif file_dtype.kind == 'f' and not np.isnan(nodata):
-                nodata = -float(nodata)
+    # PR-C #2226: centralize the nodata lifecycle in one value object.
+    # ``raw_sentinel`` carries the pre-inversion sentinel that
+    # ``attrs['nodata']`` must preserve; ``effective_sentinel`` is what
+    # the per-chunk reader's mask compares against (post-MinIsWhite).
+    # ``sentinel_fits_buffer`` collapses the previous finite/integer/
+    # in-range gate so the integer auto-promotion below skips out-of-
+    # range / fractional / non-finite sentinels (#1774, #1564, #1616).
+    from .._nodata import NodataLifecycle
+    lifecycle = NodataLifecycle(
+        declared=geo_info.nodata,
+        photometric=getattr(geo_info, '_ifd_photometric', None),
+        dtype_in=file_dtype,
+        dtype_request=dtype,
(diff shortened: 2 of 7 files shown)


brendancol (commented):
## PR Review: geotiff: centralize nodata lifecycle

### Blockers
None.

### Suggestions

- [ ] `xrspatial/geotiff/_backends/_gpu_helpers.py:82`, `_backends/dask.py:247`, `_backends/gpu.py:811`, `_writers/eager.py:867`, `_writers/eager.py:970`, `_writers/gpu.py:581`: the `from .._nodata import ...` calls are all inline lazy imports. `_nodata.py` only depends on numpy + stdlib, so there's no circular-import risk. Hoist them to the module top to avoid paying the import lookup cost on every read/write call and to keep the dependency graph visible.

- [ ] `xrspatial/geotiff/_backends/dask.py:253`: `samples_per_pixel=getattr(geo_info, '_ifd_samples_per_pixel', 1) or 1` collapses any falsy value (None, 0) to 1. The `_ifd_*` attrs are always stashed together (`__init__.py:248-249`, `_backends/dask.py:230-231`), so when `_ifd_photometric` is missing the lifecycle's `_applies_miniswhite` short-circuits on photometric anyway and the `or 1` is harmless. Worth a sentence in the comment so a future reader does not worry about the fallback masking real `samples_per_pixel=N` files.

- [ ] `_writers/eager.py:870` and `_writers/eager.py:973`: `masked_nodata_attr=False if not restore_sentinel else None` reconstructs the attr value from a derived bool. Cleaner to keep `restore_sentinel` as the cached `_should_restore_nan_sentinel(...)` return and pass it through directly, e.g. add a `restore_sentinel: bool = True` short-circuit at the top of `writer_restore_sentinel`. Saves the reverse-encoding in two writer sites.

- [ ] `xrspatial/geotiff/_nodata.py:175` -- `pixels_present` is exposed as a settable `field(init=False)` slot per the issue contract, but no production code path in this PR actually sets it (the helper is constructed locally and discarded). Either wire the existing scan results into it at the callers that compute `nodata_pixels_present`, or note in the docstring that the slot is staged for a later PR. Right now the slot exists without a consumer.

### Nits

- [ ] `xrspatial/geotiff/_nodata.py:81` -- `_invert_for_miniswhite` is a near-duplicate of `_reader._miniswhite_inverted_nodata`. The docstring acknowledges the duplication; consider adding a regression test that asserts the two functions stay in lockstep (the new `test_lifecycle_effective_sentinel_matches_reader_helper` covers the property-equivalence at the lifecycle level, but a direct equality check across a wider sentinel/dtype grid would catch drift earlier).

- [ ] `xrspatial/geotiff/tests/test_nodata_lifecycle_parity_2211.py:622` builds VRT XML by interpolating `src.rsplit("/", 1)[-1]`. Works on POSIX, breaks on Windows backslash paths. `os.path.basename(src)` is portable.

- [ ] The lifecycle helper's `_final_dtype` (line 256) leaves an asymmetry the docstring already calls out: it cannot tell "int promoted to float64 by masking" from "still int because the sentinel never matched a pixel". That is the right call for a decision-only helper, but a one-line note at the top of the class docstring would make it explicit so future maintainers do not try to use `masking_occurred` as a post-execution oracle.

### What looks good

- The decision/array-op split lands cleanly: backend kernels keep the cupy/numpy/dask ops and the helper owns the gate logic.
- The new parity tests cover the contract directly (54 cases pinning each property) AND the cross-backend behavior, with a property-equivalence check against the existing reader helper so drift surfaces fast.
- The protected regression suites (`test_nodata_lifecycle_attrs_2135.py`, `test_nodata_semantics_split_1988.py`, `test_apply_nodata_mask_gpu_inplace_1934.py`) all still pass.
- VRT and `_reader.py` are intentionally untouched to keep PR-E's rebase cheap; the PR body calls this out.
- Photometric handling is correct: `_applies_miniswhite` requires `photometric == 0 AND samples_per_pixel == 1` matching `_reader._miniswhite_inverted_nodata`'s gate.
- Comment density is honest -- every new code site explains the policy collapse with an issue reference.

### Checklist
- [x] Algorithm matches reference (lifecycle helper mirrors `_reader._miniswhite_inverted_nodata` exactly for MinIsWhite, `_should_restore_nan_sentinel` for the writer gate)
- [x] All implemented backends produce consistent results (eager / Dask / GPU / VRT-eager parity covered in new tests)
- [x] NaN handling is correct (NaN sentinel short-circuits in masking AND in `writer_restore_sentinel`)
- [x] Edge cases covered (out-of-range, fractional, non-finite, multi-band photometric=0)
- [x] Dask chunk boundaries handled correctly (lifecycle decisions resolve once at graph-build time, not per chunk)
- [x] No premature materialization or extra copies (dataclass init is constant-time scalar work)
- [n/a] Benchmark exists or not needed (decision helper, not numerical code)
- [n/a] README feature matrix updated (no public API change)

brendancol (commented):
## PR Review (round 2): centralize nodata lifecycle

Re-reviewing after the follow-up commit `34440e2`. All four suggestions
from the round-1 review landed; the three nits are also fixed.

### Blockers
None.

### Suggestions
None.

### Nits
None new. Round-1 nits:
- `_invert_for_miniswhite` drift coverage: addressed by
  `TestInvertForMinIsWhiteLockstep.test_grid_matches_reader_helper`
  (108 sentinel/dtype combos pinned).
- VRT XML `os.path.basename` portability: applied at the fixture.
- `_final_dtype` asymmetry note: addressed in the class docstring's
  Notes section.

### What looks good

- The lazy `from .._nodata import ...` calls are gone; every site now
  uses a single module-top import. The dependency graph is visible
  from the import block.
- The new `restore_sentinel=` kwarg on `writer_restore_sentinel`
  collapses the reverse-encoded `masked_nodata_attr=False if not
  restore_sentinel else None` into one positional bool. Both eager
  writer sites and both GPU writer sites use the new shape.
- The `pixels_present` slot and `masking_occurred` limitations are
  now spelled out in the class docstring's Notes block.
- `TestInvertForMinIsWhiteLockstep` sweeps 18 sentinels x 6 dtypes,
  catching any drift between `_nodata._invert_for_miniswhite` and
  `_reader._miniswhite_inverted_nodata` immediately.

### Checklist
- [x] Algorithm matches reference (108-cell drift test now pins this)
- [x] NaN handling is correct
- [x] Edge cases covered
- [n/a] Benchmark (decision helper)
- [n/a] README feature matrix (no public API change)

Local results: `pytest xrspatial/geotiff/tests/` -> 4759 passed, 45
skipped, 1 xfailed (one pre-existing lz4 deselect, unrelated to this
PR). Ready for merge from a review standpoint.

PR #2241, opened 2026-05-21
geotiff: sync attrs contract docstring with constant (#2237)

### LINKED ISSUE #2237: geotiff: attrs contract version docstring is stale (says 3, constant is 4)
**Describe the bug**

The module docstring in `xrspatial/geotiff/_attrs.py` says the contract version is `3` (line 28), but the `_ATTRS_CONTRACT_VERSION` constant on line 279 is `4`. The constant was bumped to 4 when `attrs['rotated_affine']` was added for the `allow_rotated=True` opt-in path (issue #2129); the docstring's "Attrs contract" preamble never got the matching update.

This is not a runtime bug. The constant is the source of truth and is what gets stamped into `attrs['_xrspatial_geotiff_contract']` on every read. The stale docstring just makes the documented contract version misleading for anyone reading the module top-down.

**Expected behavior**

The version mentioned in the module docstring should match `_ATTRS_CONTRACT_VERSION`.

**Fix**

1. Update the `(currently ``3``)` in the docstring at `_attrs.py:28` to `4`.
2. Grep the geotiff package and docs for any other in-source references that pin a specific contract version number. `docs/source/user_guide/attrs_contract.rst` already says `4` in its Versioning section, so no doc update is needed there. The per-version changelog notes inline in `_attrs.py` above the constant already cover v1, v2, v3, and v4, so no new history needs to be invented.
3. Add a guard test in `xrspatial/geotiff/tests/` that parses the module docstring for the version number and asserts it equals `_ATTRS_CONTRACT_VERSION`. The test should fail if the docstring and the constant drift apart again.

**Additional context**

Found during a low-severity code review of the geotiff module.

brendancol:
## Summary

- The `_attrs.py` module docstring said the contract version was `3`, but `_ATTRS_CONTRACT_VERSION` is `4` (bumped for `rotated_affine`, issue #2129). Sync the docstring.
- Add a guard test that parses the version out of the module docstring and asserts it equals `_ATTRS_CONTRACT_VERSION`, so the next drift fails in CI.

## Backends touched

Docs and tests only. No backend code paths changed.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_attrs_contract_version_1984.py` passes (8 tests, including the new guard).
- [x] Confirmed the guard catches drift: patching the docstring back to `3` makes the new test fail.
- [x] `pytest xrspatial/geotiff/tests/` passes (4974 passed, 68 skipped, 6 deselected for a pre-existing `lz4` failure that also fails on `main`, unrelated to this change).

Closes #2237

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -25,7 +25,7 @@ is present, and pass-through keys are kept when the writer can
 reconstruct them from canonical state.
 
 The contract version is recorded in ``attrs['_xrspatial_geotiff_contract']``
-(currently ``3``). Consumers can branch on this integer if the tier
+(currently ``4``). Consumers can branch on this integer if the tier
 split changes in a future release.
 
 Canonical (xrspatial owns these; round-trip stable):
--- a/xrspatial/geotiff/tests/test_attrs_contract_version_1984.py
+++ b/xrspatial/geotiff/tests/test_attrs_contract_version_1984.py
@@ -24,11 +24,13 @@ from __future__ import annotations
 
 import importlib.util
 import os
+import re
 
 import numpy as np
 import pytest
 
 from xrspatial.geotiff import open_geotiff, read_vrt
+from xrspatial.geotiff import _attrs as _attrs_module
 from xrspatial.geotiff._attrs import _ATTRS_CONTRACT_VERSION
 
 tifffile = pytest.importorskip("tifffile")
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: geotiff: sync attrs contract docstring with constant (#2237)

None.

None.

- `xrspatial/geotiff/tests/test_attrs_contract_version_1984.py:106`: the regex pins the exact wording (`attrs['_xrspatial_geotiff_contract']`` (currently ``<N>``)``). If someone rewrites the surrounding sentence (drops `currently`, replaces it with `current value`, etc.), the test fails on the "Could not find" branch instead of the more helpful version-mismatch one. The test's own docstring already flags this ("Update the docstring or this test if the phrasing changed"), so no action needed here, but worth knowing for the next person who refactors that paragraph.

### What looks good

- Smallest possible fix for the bug: one-character docstring change plus a single guard test.
- Guard test mirrors the existing `test_attrs_contract_version_constant_is_current` pattern (pin the integer) and adds the doc-side check that was missing.
- Assertion messages tell the maintainer exactly which file to edit and what to change. The dynamic `'(currently ``{documented_version}``)' to '(currently ``{_ATTRS_CONTRACT_VERSION}``)'` hint is a nice touch.
- PR body correctly notes that `docs/source/user_guide/attrs_contract.rst` already says v4 and that the per-version changelog in `_attrs.py` already covers v1 through v4. No extra docs work was needed, and none was invented.
- Pre-existing `lz4` test failure on `main` correctly identified as unrelated. Reproduced from the main checkout to confirm.

### Checklist

- [x] Algorithm matches reference/paper (N/A, docs/test change)
- [x] All implemented backends produce consistent results (N/A, no backend code changed)
- [x] Edge cases covered by tests (missing-docstring branch guarded; drift case verified manually)
- [x] README feature matrix updated (not needed; no new function, no backend change)
- [x] Docstrings present and accurate (the whole point of this PR)

PR #2243, opened 2026-05-21
geotiff: tighten 3D writer dim validator to reject non-band trailing dims (#2240)

### LINKED ISSUE #2240: geotiff: _validate_3d_writer_dims silently accepts non-band trailing dims
`_validate_3d_writer_dims` in `xrspatial/geotiff/_validation.py` accepts any `(y_alias, x_alias, *)` DataArray dim tuple where the trailing dim is not a recognized temporal name. DataArrays with dims like `('y', 'x', 'z')`, `('y', 'x', 'level')`, or `('lat', 'lon', 'scenario')` go through silently and write the trailing axis as TIFF bands. The band index in the resulting file has no link back to the original dim values.

Reproducer:

```python
import io, numpy as np, xarray as xr

da = xr.DataArray(
    np.zeros((4, 4, 3), dtype=np.float32),
    coords={'y': np.arange(4.0), 'x': np.arange(4.0),
            'scenario': ['low', 'mid', 'high']},
    dims=('y', 'x', 'scenario'),
)
buf = io.BytesIO()
to_geotiff(da, buf)  # No error. Writes three TIFF bands.
```

Expected behaviour: `ValueError` telling the caller to rename the trailing dim to a band alias (`band`, `bands`, `channel`), reduce/select along the non-band axis, or pass a raw ndarray if the band-last layout is intended.

The validator's docstring says only `(band, y, x)` and `(y, x, band)` are unambiguous, but the implementation has a fallback that accepts unknown trailing dims. The fallback was intended for raw numpy callers building band-last arrays without dim metadata, but it lives in the wrong place: a DataArray dim validator. DataArray dims should require a band alias; raw ndarray band-last compatibility belongs outside this function.

Fix:

1. Tighten `_validate_3d_writer_dims` so DataArray dim tuples with a non-band, non-temporal trailing axis raise `ValueError`. The message should name the offending dim and point at recognized band aliases or the `transpose` workaround.
2. Confirm the raw-ndarray band-last path still works. Raw ndarrays never reach the validator (it is only called inside `isinstance(data, xr.DataArray)` branches), so this is a move-by-inspection rather than code motion.
3. Update the one existing test that locked in the loose contract (`test_validate_3d_still_accepts_unknown_trailing_dim`) to assert the new rejection.

Existing callers passing DataArrays with dims like `('y', 'x', 'z')` will now get a `ValueError` instead of a silently mis-written TIFF. Raw ndarray inputs are unchanged.

Tests to add: `('y', 'x', 'z')` and `('lat', 'lon', 'scenario')` raise with an actionable message; `('band', 'y', 'x')` and `('y', 'x', 'band')` still succeed; temporal dims still take the temporal-specific message branch; raw ndarrays with band-last layout still write and round-trip.

Found in code review of the geotiff module.

brendancol:
## Summary

Closes #2240.

- Tighten `_validate_3d_writer_dims` so DataArray dim tuples with a non-band, non-temporal trailing axis raise `ValueError` instead of silently writing the trailing axis as TIFF bands. Examples that previously went through: `('y', 'x', 'z')`, `('y', 'x', 'level')`, `('lat', 'lon', 'scenario')`.
- The error names the offending dim and points the caller at the band aliases (`band`, `bands`, `channel`), an `isel`/`mean` reduction, or the raw-ndarray band-last escape.
- Raw ndarray inputs are unchanged. The validator is only invoked inside `isinstance(data, xr.DataArray)` branches in the eager, dask-streaming, and GPU writers, so bare numpy arrays never reach it. The audit is by inspection rather than by code motion.

Backend coverage: the validator is dispatch-agnostic and fires before any backend-specific work. The #1812 / #1972 regression matrix already exercises eager numpy, dask streaming, and GPU paths and continues to pass.

## Test plan

- [x] New validator-level tests reject `('y', 'x', 'z')`, `('y', 'x', 'level')`, `('y', 'x', 'scenario')`, and the y/x alias variants.
- [x] End-to-end `to_geotiff` rejects `('y', 'x', 'z')` and `('lat', 'lon', 'scenario')` DataArrays with the new message.
- [x] `('band', 'y', 'x')` and `('y', 'x', 'band')` still round-trip cleanly.
- [x] Temporal trailing/leading dims still route to the #1972 message branch.
- [x] Raw ndarray with band-last layout still writes correctly.
- [x] Full `xrspatial/geotiff/tests/` suite: 5011 passed (1 unrelated pre-existing lz4 failure on `main`).

### DIFF
--- a/xrspatial/geotiff/_validation.py
+++ b/xrspatial/geotiff/_validation.py
@@ -93,11 +93,14 @@ def _validate_3d_writer_dims(dims) -> None:
     if band_layout or yxb_layout:
         return
     # ``(y, x, *)`` with a non-band trailing dim. Temporal names get a
-    # dedicated friendly message (issue #1972); everything else
-    # (``z``, ``level``, ``scenario``, ``foo``, ...) is rejected with
-    # the generic ambiguous-dims wording below. Issue #2240 closes the
-    # escape hatch that previously let unknown trailing names through
-    # the band-position fallback.
+    # dedicated friendly message (issue #1972); every other non-band
+    # trailing name (``z``, ``level``, ``scenario``, ``foo``, ...) is
+    # rejected with the dedicated ``"non-band trailing dim"`` message
+    # immediately below. Issue #2240 closes the escape hatch that
+    # previously let unknown trailing names through the band-position
+    # fallback. The generic ``"ambiguous dims"`` message at the bottom
+    # of the function only fires for layouts that are neither
+    # ``(y, x, *)`` nor ``(_TIME_, y, x)`` (e.g. ``('foo', 'y', 'x')``).
     if d0 in _Y_DIM_NAMES and d1 in _X_DIM_NAMES:
         if _is_temporal_dim_name(d2):
             raise ValueError(
@@ -176,12 +179,13 @@ def _validate_writer_spatial_shape(shape, dims=None,
     collapse of the band axis.
 
     Note that this validator runs before ``_validate_3d_writer_dims``
-    (#1812 / #1972) in ``to_geotiff``. For an ambiguous-dim input like
-    ``(5, 5, 0)`` with dims ``('y', 'x', 'time')``, the band-last branch
-    sees ``bands == 0`` and the "no bands" error wins over the friendlier
-    ambiguous-dim message. Both errors name the right call to fix, so
-    the ordering is acceptable; reorder only if the ambiguous-dim
-    diagnostic becomes more important than the empty-axis one.
+    (#1812 / #1972 / #2240) in ``to_geotiff``. For an ambiguous-dim
+    input like ``(5, 5, 0)`` with dims ``('y', 'x', 'time')``, the
+    band-last branch sees ``bands == 0`` and the "no bands" error wins
+    over the friendlier ambiguous-dim message. Both errors name the
+    right call to fix, so the ordering is acceptable; reorder only if
+    the ambiguous-dim diagnostic becomes more important than the
+    empty-axis one.
     """
     if shape is None:
         return
--- a/xrspatial/geotiff/_validation.py
+++ b/xrspatial/geotiff/_validation.py
@@ -55,7 +55,7 @@ def _is_temporal_dim_name(name) -> bool:
 
 
 def _validate_3d_writer_dims(dims) -> None:
-    """Reject ambiguous 3D writer inputs (issue #1812).
+    """Reject ambiguous 3D writer inputs (issues #1812, #1972, #2240).
 
     The writer interprets a 3D DataArray as either ``(band, y, x)`` or
     ``(y, x, band)``. ``data.dims[0] in _BAND_DIM_NAMES`` decides which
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: geotiff: tighten 3D writer dim validator to reject non-band trailing dims (#2240)

### Blockers
None.

### Suggestions
None above nit level. The change is small: one helper, one new test file, one renamed existing test. Docstring, commit body, and PR body all say the same thing.

### Nits
- `xrspatial/geotiff/_validation.py` lines 95-100: the new comment says non-temporal trailing dims are "rejected with the generic ambiguous-dims wording below". That was true of an earlier draft, but the implementation raises a dedicated `"non-band trailing dim"` message a few lines down (line 112). Update the comment to match, or fold the dedicated raise into the generic block. As written, a future reader tracing branches will get the wrong mental model.
- `xrspatial/geotiff/_validation.py` lines 178-184 (unchanged `_validate_writer_spatial_shape` docstring): the "runs before `_validate_3d_writer_dims` (#1812 / #1972)" cross-reference could pick up #2240. Low priority.
- The new dedicated "non-band trailing dim" wording is trailing-only. The symmetric leading-dim case (`('z', 'y', 'x')`) still falls through to the generic `"ambiguous dims"` message. The asymmetry is deliberate (the #1812 tests are wired to the generic wording for the leading case), but a follow-up could symmetrise both ends.
- The new test file does not exercise the GPU writer entry point. The validator is shared, and the existing #1812 / #1972 modules already gate GPU coverage on `_HAS_GPU`, so this matches project pattern. Flagging only for completeness.

### What looks good
- The escape-hatch removal is in the right place. Read `_writers/eager.py` (lines 763, 833) and `_writers/gpu.py` (line 485): every call to `_validate_3d_writer_dims` is gated on `isinstance(data, xr.DataArray)`, so raw ndarray inputs were never going through this validator. The "move outside the DataArray validator" requirement from the issue is satisfied by inspection rather than by code motion, and the new docstring spells that out.
- Error message names the offending dim, gives three concrete remediations (band-alias rename, `isel` / `mean` reduction, raw-ndarray escape), and cites the issue.
- 33 new tests covering: validator-level for 9 trailing-dim names crossed with 4 y/x alias pairs, temporal routing regression, leading-dim ambiguous regression, 2D pass-through, end-to-end `to_geotiff` for `(y, x, z)` and `(lat, lon, scenario)`, end-to-end happy paths for `(y, x, band)` and `(band, y, x)`, and two raw-ndarray band-last round-trips.
- Tmp files include the issue number (`tmp_2240_*.tif`), matching project convention.
- Backward-compat statement in the docstring is accurate.

### Checklist
- [x] Algorithm matches reference -- n/a (validator change).
- [x] Backends produce consistent results -- validator runs before backend dispatch.
- [x] NaN handling -- n/a.
- [x] Edge cases covered -- 2D pass-through and band-alias positions covered.
- [x] Dask chunk boundaries -- n/a.
- [x] No premature materialization or unnecessary copies -- pure on `dims`.
- [x] Benchmark -- not needed.
- [x] README feature matrix -- not applicable.
- [x] Docstrings present and accurate -- yes, with the comment-vs-implementation nit above.

brendancol (commented):
## Follow-up review after nit fixes (#2240)

### Disposition of the four nits from the prior review

1. **Branch-routing comment (lines 95-100 of `_validation.py`)** -- fixed in 224013b9. Comment now matches the implementation: temporal trailing dims take the #1972 message, every other non-band trailing dim takes the dedicated `"non-band trailing dim"` message immediately below, and the bottom `"ambiguous dims"` raise only fires for layouts that are neither `(y, x, *)` nor `(_TIME_, y, x)`.
2. **`_validate_writer_spatial_shape` cross-reference (lines 178-184)** -- fixed in the same commit. The "runs before `_validate_3d_writer_dims`" note now lists #1812, #1972, and #2240.
3. **Leading-dim symmetric "non-band" wording** -- deferred. The #1812 regression test `test_error_message_actionable` asserts on the generic `"#1812"` wording and the `(band, y, x)` / `(y, x, band)` / `transpose` phrasing for the leading-dim case. Symmetrising the message would require rewiring those assertions and pulls in scope beyond the issue's "tighten the trailing-dim escape hatch" mandate. Filing a follow-up tracking issue if anyone wants the symmetric diagnostic.
4. **GPU end-to-end coverage** -- deferred. Matches the existing #1812 / #1972 project pattern: the validator is shared, the validator-level tests cover every trailing-dim and y/x alias combination, and adding GPU-gated tests would only duplicate validator-level coverage when CUDA happens to be present. Not worth the test maintenance.

### Re-verified
- Touched-test suite (`#1812`, `#1972`, `#2240` modules): 84 passed.
- No blockers remain. The PR is in the same shape as before, just with the comment and cross-reference cleaned up.

PR #2244, opened 2026-05-21
geotiff: forward read kwargs through all GPU CPU-fallback paths

### LINKED ISSUE #2238: GPU tiled fallback drops critical read kwargs (allow_rotated/window/band/max_pixels)
**Describe the bug**

In `xrspatial/geotiff/_backends/gpu.py`, three CPU-fallback call sites to `_read_to_array` drop caller-supplied kwargs (`allow_rotated`, `window`, `band`, `max_pixels`):

- line 684 (planar=2 per-band GPU decode fallback)
- line 697 (sparse-tile fallback)
- line 771 (post GPU-decode-failure fallback)

Each call is currently:

```python
arr_cpu, _cpu_fallback_geo = _read_to_array(
    source, overview_level=overview_level)
```

The later `_gpu_apply_window_band(...)` call repairs shape for `window`/`band` after the fact, but it cannot repair `allow_rotated=False`. By the time control returns to GPU code the CPU parser has already raised `NotImplementedError` on the rotated `ModelTransformation`. Callers who pass a raised `max_pixels` can also still trip the default 1B-pixel cap inside the fallback.

The stripped-layout branch at gpu.py:491 already forwards `window`, `band`, and `max_pixels` correctly (see comment referencing #1732). The three tiled-fallback sites missed the same fix, plus `allow_rotated`.

**Expected behavior**

The three CPU-fallback call sites should forward all four caller-supplied kwargs:

```python
arr_cpu, _cpu_fallback_geo = _read_to_array(
    source,
    overview_level=overview_level,
    window=window,
    band=band,
    max_pixels=max_pixels,
    allow_rotated=allow_rotated,
)
```

Regression coverage:

- rotated-file fallback succeeds when `allow_rotated=True`. This currently xfails in `test_allow_rotated_no_crs_2122.py::test_cupy_rotated_read_drops_crs`.
- caller-supplied `max_pixels` raise is honored on the fallback path.
- caller-supplied `window`/`band` reach the fallback. The assertion can sit at the `_read_to_array` boundary so it does not rely on `_gpu_apply_window_band` masking the bug.

**Affected paths**

- Sparse tiled files (`has_sparse_tile=True`)
- PlanarConfiguration=2 multi-band tiled files when the per-band stage-2 GPU decode fails
- Any tiled file when both `gpu_decode_tiles_from_file` and `gpu_decode_tiles` raise under `on_gpu_failure='auto'`

**Additional context**

Surfaced by a code review sweep of the geotiff module. Sibling finding to #2122, which fixed the eager numpy / dask / dask+cupy paths for `allow_rotated`. The `test_cupy_rotated_read_drops_crs` xfail in `test_allow_rotated_no_crs_2122.py` was filed against this same defect and will flip to a pass once the kwargs are forwarded.

brendancol:
Closes #2238.

## Summary

- Forward `allow_rotated`, `window`, `band`, and `max_pixels` through every CPU-fallback `_read_to_array` call inside `read_geotiff_gpu` (stripped-layout, planar=2 per-band, sparse-tile, and post GPU-decode-failure).
- Skip the post-decode shape check and `_gpu_apply_window_band` slicer on CPU-decoded paths so the now-already-sliced fallback buffer doesn't get double-windowed or trip the multi-band shape guard.
- Flip `test_cupy_rotated_read_drops_crs` from xfail to pass; add a regression-test file pinning the kwarg-forwarding contract on all four fallback sites.

## Backend coverage

Touches only `xrspatial/geotiff/_backends/gpu.py`. The cupy and dask+cupy paths use this code; the dask+cupy chunked path already forwarded kwargs correctly via `_read_geotiff_gpu_chunked`, so it's unaffected. The numpy and dask+numpy paths are unchanged.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_gpu_fallback_forwards_kwargs_2238.py` (6 new tests, all pass)
- [x] `pytest xrspatial/geotiff/tests/test_allow_rotated_no_crs_2122.py` (xfail-flipped test now passes; 8 tests total)
- [x] `pytest -k "gpu or cupy or planar or sparse or fallback or rotated" xrspatial/geotiff/tests/` (1189 passed, 20 skipped, no regressions)

### DIFF
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -487,10 +487,15 @@ def read_geotiff_gpu(source: str, *,
             # Orientation != 1 + window is already rejected at line 2495,
             # so ``window`` is None whenever ``geo_info`` will be remapped
             # below.
+            #
+            # ``allow_rotated`` is forwarded so rotated stripped files do
+            # not trip ``_read_to_array``'s rotation guard on the fallback
+            # path (issue #2238).
             src.close()
             arr_cpu, _stripped_geo = _read_to_array(
                 source, overview_level=overview_level,
-                window=window, band=band, max_pixels=max_pixels)
+                window=window, band=band, max_pixels=max_pixels,
+                allow_rotated=allow_rotated)
             arr_gpu = cupy.asarray(arr_cpu)
             if orientation != 1:
                 geo_info = _apply_orientation_geo_info(
@@ -681,8 +686,18 @@ def read_geotiff_gpu(source: str, *,
                 # 2/3/4 case is covered regardless of #1539's merge state), but
                 # keep it on ``_cpu_fallback_geo`` so the MinIsWhite-aware nodata
                 # mask below sees ``_mask_nodata``.
+                #
+                # Forward ``allow_rotated``, ``window``, ``band``, and
+                # ``max_pixels`` so caller intent reaches the fallback parser
+                # (issue #2238). Without these the CPU parser rejects rotated
+                # ModelTransformation files even with ``allow_rotated=True``,
+                # and a caller-raised ``max_pixels`` still trips the default
+                # 1B-pixel cap inside the fallback. The shape repair in
+                # ``_gpu_apply_window_band`` below cannot un-raise that.
                 arr_cpu, _cpu_fallback_geo = _read_to_array(
-                    source, overview_level=overview_level)
+                    source, overview_level=overview_level,
+                    window=window, band=band, max_pixels=max_pixels,
+                    allow_rotated=allow_rotated)
                 arr_gpu = cupy.asarray(arr_cpu)
                 arr_was_cpu_decoded = True
             else:
@@ -694,8 +709,11 @@ def read_geotiff_gpu(source: str, *,
                         f"({height}, {width}, {samples})"
                     )
         elif has_sparse_tile:
+            # Forward caller kwargs to the CPU fallback (issue #2238).
             arr_cpu, _cpu_fallback_geo = _read_to_array(
-                source, overview_level=overview_level)
+                source, overview_level=overview_level,
+                window=window, band=band, max_pixels=max_pixels,
+                allow_rotated=allow_rotated)
             arr_gpu = cupy.asarray(arr_cpu)
             arr_was_cpu_decoded = True
         else:
@@ -768,8 +786,11 @@ def read_geotiff_gpu(source: str, *,
                     RuntimeWarning,
                     stacklevel=2,
                 )
+                # Forward caller kwargs to the CPU fallback (issue #2238).
                 arr_cpu, _cpu_fallback_geo = _read_to_array(
-                    source, overview_level=overview_level)
+                    source, overview_level=overview_level,
+                    window=window, band=band, max_pixels=max_pixels,
+                    allow_rotated=allow_rotated)
                 arr_gpu = cupy.asarray(arr_cpu)
                 arr_was_cpu_decoded = True
 
@@ -777,7 +798,14 @@ def read_geotiff_gpu(source: str, *,
         # config -- catch any shape regression in the kernels before we attach
         # dims/coords below. Plain `raise` rather than `assert` so the check
         # survives `python -O`.
-        if samples > 1:
+        #
+        # Skip the check on CPU-fallback paths: ``_read_to_array`` is now
+        # called with ``window``/``band`` (issue #2238), so the returned
+        # buffer is already sliced and will not equal the full IFD shape.
+        # The pure-GPU branches still need the guard because they hand
+        # back a fully-decoded ``(H, W, samples)`` buffer and rely on the
+        # downstream ``_gpu_apply_window_band`` for the slice.
+        if samples > 1 and not arr_was_cpu_decoded:
             if (arr_gpu.shape[:2] != (height, width)
                     or arr_gpu.shape[2] != samples):
                 raise RuntimeError(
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: geotiff: forward read kwargs through all GPU CPU-fallback paths

- None.

- [ ] `test_sparse_tile_fallback_respects_caller_max_pixels` (test_gpu_fallback_forwards_kwargs_2238.py:351-376) does not actually exercise the fallback. With `max_pixels=10` on a 32x32 file, the pre-fallback `_check_dimensions(width, height, samples, max_pixels)` at `gpu.py:559` raises before the sparse-tile branch fires. The test passes for the wrong reason. The kwarg-recorder test directly above (`test_sparse_tile_fallback_forwards_all_kwargs`) already pins `max_pixels` plumbing via assertion on `seen[0]['max_pixels']`. Either drop this test as redundant, or convert it to a case where the file's full IFD fits under `max_pixels` but a synthetic sub-decoder failure forces the fallback to re-run with the same kwarg (closer to the recorder pattern in `test_planar2_fallback_forwards_all_kwargs`).

- [ ] The hand-crafted TIFF fixture helpers `_write_rotated_tiled_tiff` and `_write_sparse_rotated_tiled_tiff` are nearly identical; the sparse variant differs only in the offset/byte_count packing for the last tile. Consider factoring the shared IFD-build logic into a helper or parameterising one writer on a `sparse: bool` flag. About 80 lines of fixture code collapses to about 10. Not blocking; the duplication is contained and well-commented.

- [ ] `gpu.py:801`: the comment "now-already-sliced fallback buffer" reads slightly awkwardly. "the already-sliced fallback buffer" or "the buffer (now sliced by `_read_to_array`)" is cleaner.

- [ ] `test_gpu_fallback_forwards_kwargs_2238.py:392-393`: import order. `_gpu_decode` is imported inside the test body. Pre-existing style in this repo (see `test_predictor2_big_endian_gpu_1517.py`) puts function-scope imports at the top of the test function. The current order works; flagged only for consistency.

### What looks good
- The fix is small and mirrors the existing `arr_was_cpu_decoded` skip pattern already used by `_apply_orientation_gpu` (gpu.py:817-820). The new skip on `_gpu_apply_window_band` (gpu.py:864-873) and the multi-band shape guard (gpu.py:801-811) reuses the same flag, so future readers can see all three skips together.
- The post-decode shape check is correctly narrowed to non-CPU-decoded paths so it still catches pure-GPU kernel regressions while not falsely tripping on the now-windowed fallback output.
- The xfail flip on `test_cupy_rotated_read_drops_crs` (was strict=True, raises=NotImplementedError) provides automatic regression coverage; if the kwarg-forwarding breaks again, that test will raise in CI before the new test file even runs.
- Test coverage hits all four call sites: stripped (via the existing #2122 test), sparse (real round-trip on a hand-crafted rotated tiled file), planar=2 (monkeypatched `_gpu_decode_single_band_tiles` returning None), and decode-failure (both `gpu_decode_tiles_from_file` and `gpu_decode_tiles` patched to raise).
- The `_make_kwarg_recorder` helper is the right abstraction here. Asserting on `seen[0]` rather than chasing data shapes keeps each fallback site's contract pinned at the API boundary, independent of downstream slicing fixes.
- No backend dispatch changes; the numpy / dask / dask+cupy paths are correctly identified as unaffected (the dask+cupy chunked path already forwards kwargs via `_read_geotiff_gpu_chunked`).

### Checklist
- [x] Algorithm correctness: no algorithm change; kwarg-plumbing fix only
- [x] All implemented backends produce consistent results: verified by the cross-fallback recorder tests and the 1189-test sweep
- [x] NaN handling: unchanged; the sparse-tile fallback still goes through `_finalize_eager_read` for the mask
- [x] Edge cases: window/band/max_pixels/allow_rotated each covered for at least one fallback site
- [x] Dask chunk boundaries: N/A (eager GPU path only)
- [x] Benchmark: N/A (bug fix)
- [x] README feature matrix: N/A (no backend change)
- [x] Docstrings: comments at each new branch reference issue #2238; no public API change

brendancol (commented):
## PR Review (round 2): post-follow-up

- None.

- None remaining from the previous review.

- None remaining from the previous review.

- `test_sparse_tile_fallback_respects_caller_max_pixels` was dropped. The recorder test (`test_sparse_tile_fallback_forwards_all_kwargs`) covers `max_pixels` plumbing.
- `_write_sparse_rotated_tiled_tiff` was folded into `_write_rotated_tiled_tiff` via a `sparse: bool` kwarg. The shared IFD-build logic now lives in one place; the sparse path only adds offset=0/bytecount=0 entries for the last tile.

- Suggestion 1 (misleading max_pixels test): fixed by deletion. The duplicate-coverage rationale in the review was the right call.
- Suggestion 2 (TIFF writer duplication): fixed via the `sparse=True` parameter on the unified writer. Diff dropped ~93 lines of fixture code.
- Nit 1 (comment wording): dismissed. The phrase the reviewer flagged was in the round-1 review body, not the source. The actual gpu.py:804 comment ("the buffer is already sliced and will not equal the full IFD shape") reads cleanly.
- Nit 2 (import order): dismissed. Surrounding test files (e.g. `test_predictor2_big_endian_gpu_1517.py`) put function-scope imports inside the test body, matching the current style. The suggested move would diverge from the repo's existing convention.

### Verification
- `pytest xrspatial/geotiff/tests/test_gpu_fallback_forwards_kwargs_2238.py xrspatial/geotiff/tests/test_allow_rotated_no_crs_2122.py`: 13 passed

PR #2245, opened 2026-05-21
Discover external .tif.ovr sidecars on remote chunked GeoTIFF reads

### LINKED ISSUE #2239: Remote chunked GeoTIFF reads do not honor external .ovr sidecars
**Describe the bug**

The eager reader in `xrspatial/geotiff/_reader.py` checks for a sibling `.ovr` sidecar (local and fsspec sources) before it picks an `overview_level`, so a request for an overview that lives in the external `.ovr` resolves against the merged base + sidecar IFD list.

The dask metadata path for HTTP and fsspec in `xrspatial/geotiff/_backends/dask.py` (around line 181) skips that step and goes straight through `_parse_cog_http_meta`. `_read_geo_info` in `xrspatial/geotiff/__init__.py` (around line 223) has the same fsspec bypass. The eager HTTP path in `_read_cog_http` also never looks for a sidecar, so HTTP is wrong both ways.

So `open_geotiff(remote_tif, chunks=..., overview_level=1)` against a GDAL external-overview file can raise out-of-range or pick a different overview than the eager read of the same URL, even though the `.ovr` is sitting beside the base file.

**Expected behavior**

A chunked open of a remote (HTTP or fsspec-mocked) GeoTIFF with a sibling `.ovr` should resolve `overview_level` against the same merged pyramid the eager open sees. Eager and chunked should agree for HTTP and fsspec sources, the same way they already do for local files.

**Repro sketch**

1. Take a non-COG GeoTIFF with an external `.ovr` produced by `gdaladdo -ro`.
2. Host it (or mock it via fsspec `memory://`) so both the base and the sidecar are fetchable.
3. `open_geotiff(remote_tif, overview_level=1)` succeeds and returns the sidecar's first IFD.
4. `open_geotiff(remote_tif, chunks={'x': 256, 'y': 256}, overview_level=1)` fails or picks a different overview.

**Additional context**

Factor the sidecar-discovery helper so all four call sites (eager local/fsspec, eager HTTP, dask metadata, `_read_geo_info`) share one source of truth, plus regression tests covering chunked-vs-eager parity on at least the fsspec-mocked and HTTP-mocked paths.

Found by code review of the geotiff module.

brendancol:
Closes #2239

## Summary
- Pull external `.tif.ovr` sidecar discovery into a shared `discover_remote_sidecar` helper in `_sidecar.py`. The eager local/fsspec reader already merged sidecar IFDs onto the pyramid list; the dask metadata path, the eager HTTP path, and `_read_geo_info`'s fsspec branch went straight through `_parse_cog_http_meta` and never saw the sidecar.
- `_parse_cog_http_meta` now takes an optional `source_path` to opt into discovery, plus an optional `return_sidecar=True` to hand back the sidecar handle + route URL so chunked tile fetches target the sidecar object when the requested overview lives there.
- Thread the new kwargs through `_read_cog_http` (eager HTTP), `_backends/dask.py` (chunked HTTP + fsspec), and `_read_geo_info` (chunked metadata for fsspec).

## Backend coverage
- numpy (eager): local-file and fsspec `memory://` already merged sidecars pre-PR; eager HTTP now matches.
- dask + numpy: chunked HTTP and chunked fsspec `memory://` both honour sidecar overview levels.
- cupy and dask + cupy: untouched. The GPU backend already routes through `_read_geo_info` for chunked and through its own discovery for eager local. GPU + remote sidecar is out of scope.

## Test plan
- [x] `pytest xrspatial/geotiff/tests/test_remote_sidecar_chunked_2239.py` -- 12 new tests, all pass.
- [x] `pytest xrspatial/geotiff/tests/test_sidecar_ovr_2112.py xrspatial/geotiff/tests/test_sidecar_max_cloud_bytes_2121.py` -- existing sidecar suite still green.
- [x] Full geotiff suite: 4990 pass / 1 pre-existing failure (`test_write_vs_to_geotiff_byte_parity_uint8[lz4]`, unrelated).
- [x] Full xrspatial suite: 3956 pass.


brendancol (commented):
## PR Review: Discover external .tif.ovr sidecars on remote chunked GeoTIFF reads (#2245)

None.

- [ ] `_parse_cog_http_meta`: the `Returns` docstring section is missing -- the prose paragraph mentions the 5-tuple but there's no formal `Returns` block. Add one so callers don't have to read the body to learn the shape. (`xrspatial/geotiff/_reader.py:2329`)
- [ ] `_parse_cog_http_meta`: when `return_sidecar=True` is passed with `source_path=None`, `route_path` ends up as `None`. No current caller hits that combo, but a future one could and would crash on `_HTTPSource(None)`. Either assert at entry that `return_sidecar` requires `source_path`, or pick a saner fallback. (`xrspatial/geotiff/_reader.py:2403`)
- [ ] `discover_remote_sidecar`: the bare `except Exception` swallows any load failure and falls back to base-only. The docstring covers this, but the `CloudSizeLimitError` carve-out deserves its own line so users debugging a surprise sidecar miss know that budget breaches are the one exception that re-raises. (`xrspatial/geotiff/_sidecar.py:296-340`)
- [ ] No focused regression test for the "probe succeeds, load fails" branch in `discover_remote_sidecar`. The `test_security` mocks exercise it indirectly, but a targeted test would catch a future regression that drops the catch-all.

- [ ] `xrspatial/geotiff/_backends/dask.py:198`: the comment block above the `_parse_cog_http_meta` call grew significantly. The same rationale is relevant in `_reader.py:_read_cog_http`. Worth extracting to a module-level note.
- [ ] `test_remote_sidecar_chunked_2239.py`: the fsspec fixture key `memory://issue2239/x.tif` is not collision-proof under pytest-xdist (workers share the in-process memory filesystem). Adding `os.getpid()` or a uuid would close the remaining window.
- [ ] `test_remote_sidecar_chunked_2239.py:128-141`: `test_fsspec_chunked_open_base_level_unchanged` duplicates the level-0 leg of the parametrised `test_fsspec_chunked_open_resolves_sidecar_overview`. Folding it in via `@pytest.mark.parametrize("overview_level", [0, 1, 2])` removes one test body.

### What looks good
- The helper extraction is the right call. Three call sites converging on `discover_remote_sidecar` is a clean win over duplicated discovery logic.
- The dask backend's `effective_source` swap is the minimum change needed to point per-chunk reads at the sidecar URL when the requested overview lives there. The per-chunk task only uses the path string to instantiate its source, so the swap is enough.
- The two existing-test fixes (autouse `find_sidecar` monkeypatch, tuple-shape-forwarding wrappers) preserve the original assertions without weakening them. The comments say why.
- The 12 new tests cover eager-vs-chunked parity for HTTP and fsspec, the metadata-only path, and the out-of-range guard.
- `extract_geo_info_with_overview_inheritance` is called with the base buffer even for sidecar IFDs, matching the eager local reader. The inline comment calls this out.

### Checklist
- [x] Algorithm matches the eager local reader's sidecar contract
- [x] numpy eager and numpy dask backends produce consistent results against the local-file baseline
- [x] NaN handling unchanged (this PR doesn't touch the nodata lifecycle)
- [x] Edge cases tested (out-of-range overview, base-level unchanged, file-like buffer unaffected)
- [x] Dask chunk boundaries handled correctly (per-chunk task uses sidecar URL; existing chunk math unchanged)
- [x] No premature materialization (sidecar metadata parsed once per graph build; per-chunk tasks reuse the IFD via `http_meta_key`)
- [N/A] Benchmark needed -- no perf-sensitive code touched
- [N/A] README feature matrix -- no new public function
- [x] Docstrings present and accurate (one suggestion above to add a `Returns` block)

brendancol (commented):
## PR Review (Round 2)

The first round had no Blockers, 4 Suggestions, and 3 Nits. The follow-up commit (29e6d00c) addresses six of seven; one Nit is dismissed with a reason.

Suggestions:
- Fixed: `_parse_cog_http_meta` `Returns` block added (`_reader.py`).
- Fixed: precondition check on `return_sidecar=True` + `source_path=None` raises `TypeError` (`_reader.py`).
- Fixed: `CloudSizeLimitError` carve-out lifted into a `Raises` block (`_sidecar.py`).
- Fixed: two regression tests for the swallow-and-fall-back path, plus one for the precondition.

Nits:
- Dismissed: extracting the dask-backend comment into a module-level note. Each call site (eager HTTP in `_reader.py`, dask in `_backends/dask.py`) reads cleanest with its own context, and the duplication is small. Worth revisiting if a third call site ever appears.
- Fixed: fsspec `memory://` fixture key now uses `os.getpid()` + uuid suffix.
- Fixed: the base-level test is folded into the parametrised eager-vs-chunked test.

### What looks good on the follow-up commit
- The `Returns` block reads the same as the prose paragraph it replaces, just with the right docstring shape.
- The new `TypeError` precondition gives a future caller a clear message instead of a deep `_HTTPSource(None)` crash.
- The three new regression tests run in under 0.2s combined and use `monkeypatch` cleanly -- no module state to undo.

### Checklist
- [x] Round-1 findings addressed or explicitly dismissed
- [x] Tests still pass locally (15 new tests + the existing sidecar suite)
- [x] No new findings introduced

Issue #2255, opened 2026-05-21
rasterize: GPU max/min merge silently suppresses NaN burn values


brendancol:
## Summary

The GPU (cupy / dask+cupy) backends silently suppress NaN burn values
under `merge='max'` and `merge='min'`, while the CPU (numpy / dask+numpy)
backends propagate NaN per IEEE arithmetic. Same input, different output
across backends.

## Reproducer

```python
import numpy as np
from shapely.geometry import box
from xrspatial.rasterize import rasterize

geoms = [(box(0, 0, 10, 5), np.nan), (box(0, 0, 10, 5), 1.0)]

# CPU: NaN propagates (1.0 > NaN is False, so kernel keeps NaN).
r_cpu = rasterize(geoms, width=10, height=5, bounds=(0, 0, 10, 5),
                  fill=0, merge='max')
assert np.all(np.isnan(r_cpu.values))  # passes

# GPU: NaN is suppressed (buffer init -inf, atomicMax keeps the finite).
r_gpu = rasterize(geoms, width=10, height=5, bounds=(0, 0, 10, 5),
                  fill=0, merge='max', use_cuda=True)
# r_gpu == 1.0 everywhere, NOT NaN
```

Also: a single NaN burn under `merge='max'` returns NaN on CPU but `-inf`
(the GPU buffer initial value) on GPU.

## Root cause sketch

- `_merge_max` (CPU, `rasterize.py:91`) and `_merge_max_gpu`
  (`rasterize.py:1180`) have identical Python source.
- The CPU path initialises the output buffer to the fill value and uses
  `is_first` to take the first burn's value verbatim, so a NaN burn lands
  unchanged.
- The GPU path initialises the output buffer to `-inf` for `max` (and
  `+inf` for `min`) so atomic max / min can operate without an
  `is_first` flag. NaN comparisons under atomic max return False, so
  every subsequent finite value wins and the NaN never lands in the
  output.

## Impact

- Users who rely on NaN-as-"no-data" semantics get silently different
  rasters depending on which backend is selected.
- The xrspatial.rasterize public API documents merge='max' / 'min'
  without backend-specific carve-outs.

## Where it was caught

deep-sweep test-coverage pass 2 on 2026-05-21 added
`test_rasterize_coverage_2026_05_21.py::TestNaNBurnValues`. The
asymmetric tests
- `test_nan_burn_overlaps_max_cpu_propagates` (CPU = NaN)
[...]

PR #2256, opened 2026-05-21
rasterize: sweep-test-coverage pass 2 -- NaN/Inf burns, nested GC, GPU multi-column

brendancol:
## Summary

deep-sweep test-coverage pass 2 on `rasterize`. Adds 58 tests in
`xrspatial/tests/test_rasterize_coverage_2026_05_21.py` closing the
public-API coverage gaps left after pass-1 (2026-05-17). Source untouched.

### Gaps closed

- **Cat 2 HIGH (NaN / Inf edges)** -- pin +Inf / -Inf / NaN burn values
  across numpy / cupy / dask+numpy / dask+cupy on polygon, point, and
  line geometries. `Inf + finite` under `sum` stays Inf; `Inf + (-Inf)`
  under `sum` collapses to NaN; `min(Inf, 1.0)` and `max(-Inf, 1.0)`
  pick the finite value. Inf-as-bound is rejected with the same
  ValueError as NaN-as-bound (pass-1 only covered NaN bound).
- **Cat 1 MEDIUM (backend coverage)** -- nested `GeometryCollection`
  (GC inside a GC) on all four backends. `rasterize.py:1995` documents
  recursive unpacking but only flat GCs were tested. Deeply-nested
  3-levels eager test pins that the recursion depth limit is not 1 or 2.
- **Cat 1 MEDIUM (backend coverage)** -- `columns=` (multi-column
  properties) on cupy and dask+cupy. `TestMultiColumn` covered only
  numpy and dask+numpy.
- **Cat 3 LOW (geometric edges)** -- rectangular-pixel parity
  (`resolution=(rx, ry)`, rx != ry) across all four backends.

### Bug surfaced -- issue #2255

The NaN-burn-under-max tests revealed a cross-backend divergence:
**GPU `max` / `min` merge silently suppresses NaN burn values** while
CPU propagates them per IEEE arithmetic.

- CPU: `max(NaN, 1.0)` returns NaN (because `1.0 > NaN` is False, the
  kernel keeps the prior NaN pixel).
- GPU: `max(NaN, 1.0)` returns 1.0 (the kernel inits the output buffer
  to -inf for max, +inf for min, and atomicMax/Min is NaN-suppressing
  under IEEE device semantics).

Filed as #2255. This PR pins **both** observables in paired
CPU vs GPU tests (`test_nan_burn_overlaps_max_cpu_propagates`,
`test_nan_burn_overlaps_max_gpu_suppresses_nan`,
`test_nan_burn_single_geom_max_gpu_returns_neg_inf`) so the divergence
is visible in CI until the GPU kernels are aligned with the CPU
`is_first` semantics.

### State updates

- `.claude/sweep-test-coverage-state.csv`: rasterize row updated with
  pass-2 notes, `severity_max=HIGH`, `categories_found=1;2;3`,
  `issue=2255`.

## Test plan

- [x] All 58 new tests pass on a CUDA host
  (`pytest xrspatial/tests/test_rasterize_coverage_2026_05_21.py`)
- [x] Pre-existing 417 rasterize tests still pass
  (`pytest xrspatial/tests/test_rasterize*.py`)
- [x] Source files untouched (only adding tests + state CSV)
- [ ] CI runs the new tests on CPU and GPU hosts


brendancol (commented):
## PR Review: rasterize sweep-test-coverage pass 2

### Summary

Tests-only PR (`xrspatial/tests/test_rasterize_coverage_2026_05_21.py`, +547/-0) adding 58 tests across four buckets: Inf/NaN burn values, nested GeometryCollection, GPU multi-column, rectangular pixels. State CSV updated (+1/-1). No source changes. All 58 new tests pass locally, and CI is green on macOS/Ubuntu/Windows + py3.14. The 206 pre-existing rasterize tests still pass (verified locally).

### Blockers

None.

### Suggestions

None.

### Nits

- `test_inf_bounds_rejected` (lines 208-219) matches on `"must be finite"`, which also appears in the resolution-validation error at `xrspatial/rasterize.py:3101` (`"resolution must be finite and > 0"`). The path the test targets is at `rasterize.py:3078-3081`: `"Invalid bounds: all of (xmin, ymin, xmax, ymax) must be finite, ..."`. The regex still hits the right path today because the resolution path is not reached, but `r"Invalid bounds:.*must be finite"` would be safer against a future refactor that shuffles validation order. Verified manually that the test triggers the bounds path today.
- `TestNaNBurnValues` references issue #2255 in a comment at line 283. Putting that reference (or a module-level `ISSUE_2255 = ...` constant) on the class docstring too would make it easier to find when the GPU max kernel is later aligned and the `..._gpu_suppresses_nan` / `..._gpu_returns_neg_inf` tests need to be inverted.
- `test_deeply_nested_gc_eager` (lines 398-411) only runs against numpy. The fast-path-vs-slow-path branch in `_classify_geometries` is backend-agnostic, so this is not load-bearing, but one dask+numpy run would also check that nested-GC pre-classification survives the per-tile graph builder.

### What looks good

- Every test class docstring says what it pins and why a regression would be silent today. Makes the file readable a year from now.
- The GPU/CPU NaN-max divergence is pinned as paired tests rather than hidden behind `xfail`, so the asymmetry shows up in CI instead of getting buried in skip output. Issue #2255 (open) tracks the source fix.
- `_materialise` handles all four backends in one place (dask compute + cupy host copy), so individual tests do not leak backend assumptions.
- Inf-as-bound rejection (line 208) and the nested-GC recursion check (line 398) close gaps that the source documented but had no test for (rasterize.py:1995 GC recursion, rasterize.py:3078 bounds validation).
- Skip markers use `pytest.param(..., marks=...)` with stable ids (`numpy`, `cupy`, `dask_numpy`, `dask_cupy`), so test IDs are searchable and skip output stays readable on CPU-only hosts.

### Checklist

- [x] Source unchanged (tests-only)
- [x] All four backends covered where the source supports them
- [x] NaN handling pinned per-backend (CPU IEEE propagation, GPU asymmetry)
- [x] Edge cases (Inf, -Inf, Inf+(-Inf), NaN burn, nested GC, rect pixels)
- [x] Dask paths use chunked fixtures (`chunks=(5,5)` and `chunks=(3,3)`)
- [x] No premature materialization; `_materialise` is the single sink
- [x] No benchmark needed (no source changes)
- [x] No README change needed (no new public API)
- [x] Docstrings on every test explain the pin
- [x] Bug surfaced (#2255) is pinned in CI rather than fixed in scope

brendancol (commented):
## Follow-up review (post-fix)

Pushed `c9a71bd4` addressing all three nits from the prior review, plus a merge of `main` (`c5a66009`) to pull in the unrelated GeoTIFF Phase 5b/5c refactors that landed in between.

### Nit disposition

- Nit 1 (regex collision): fixed -- `test_inf_bounds_rejected` now matches `r"Invalid bounds:.*must be finite"` so it cannot be satisfied by the resolution-validation path.
- Nit 2 (issue #2255 visibility): fixed -- `TestNaNBurnValues` class docstring now flags the GPU/CPU NaN-merge divergence and points to the asymmetric paired tests.
- Nit 3 (deeply-nested GC dask coverage): fixed -- `test_deeply_nested_gc_eager` was renamed `test_deeply_nested_gc` and parametrized over `numpy` and `dask_numpy`, so the recursive `_classify_geometries` check now covers the per-tile graph builder too.

### Verification

- Local: `pytest xrspatial/tests/test_rasterize_coverage_2026_05_21.py` -- 59 passed (was 58, +1 from the new `dask_numpy` parametrization).
- Pre-existing 206 rasterize tests still pass.
- Merge of `main` was clean (GeoTIFF-only conflict surface, no overlap with rasterize).

### Out of scope (intentional)

- Issue #2255 (GPU max/min NaN suppression) is **not** fixed in this PR. The paired CPU-propagates / GPU-suppresses tests pin the current divergence so it stays visible in CI until the GPU atomic kernels can be aligned with the CPU `is_first` semantics. That's a source-level fix in `_compile_gpu_kernels` (rasterize.py:~1320) and belongs in a separate PR.

No remaining findings from my side. Ready for an outside reviewer.

PR #2270, opened 2026-05-21
Cap merged-range size in coalesce_ranges to bound HTTP over-fetch (#2266)

### LINKED ISSUE #2266: COG range coalescing can turn safe tile reads into huge over-fetches
**Describe the bug**

`coalesce_ranges` in `xrspatial/geotiff/_sources.py` merges any two adjacent byte ranges whose gap is at most `COALESCE_GAP_THRESHOLD_DEFAULT` (1 MiB). There is no upper bound on the merged range length and no cap on the total slack accumulated across a chain of merges.

The per-tile cap at `xrspatial/geotiff/_cog_http.py:883` rejects individual tiles that declare a `TileByteCount` larger than `MAX_TILE_BYTES_DEFAULT` (256 MiB by default). After that check, the full per-tile fetch list is handed to `read_ranges_coalesced` at `xrspatial/geotiff/_cog_http.py:914`, which calls `coalesce_ranges` and issues one GET per merged range.

A malformed or hostile COG can supply a tile table where every individual `TileByteCount` passes the per-tile cap (say, 1 KB each), but the tile offsets sit just under 1 MiB apart. Each adjacent pair fits the merge predicate, so the coalescer chains them all into one merged range whose length is roughly `num_tiles * 1 MiB`. For a 4096-tile COG that is a single ~4 GiB HTTP / cloud GET, even though each tile individually passed the cap.

This affects both the HTTP path (`_HTTPSource.read_ranges_coalesced`) and the cloud / fsspec path (`_FSSpecSource.read_ranges_coalesced`).

**Expected behavior**

The coalescer should refuse to merge ranges when the resulting merged-range length would exceed a configurable cap (and/or when the accumulated slack between merged members would exceed a cap). When the cap is hit, seal the current merged range and start a new one. The cap should follow the existing tile-cap pattern: a module-level default with an env-var override.

**Reproduction**

```python
from xrspatial.geotiff._sources import coalesce_ranges

# 4096 tiles, each 1 KB compressed, but offsets spaced 1 MiB apart.
# Every individual range passes a 256 MiB per-tile cap.
ranges = [(i * (1 << 20), 1024) for i in range(4096)]
merged, mapping = coalesce_ranges(ranges)
assert len(merged) == 1
merged_len = merged[0][1]
print(f'merged length: {merged_len / (1 << 30):.2f} GiB')  # ~4.00 GiB
```

A single merged range that large drives an unbounded HTTP / cloud read even though each tile individually was tiny.

**Additional context**

- `xrspatial/geotiff/_cog_http.py:883` -- per-tile cap (`max_tile_bytes`), enforced before coalescing.
- `xrspatial/geotiff/_cog_http.py:914` -- call site for `read_ranges_coalesced`.
- `xrspatial/geotiff/_sources.py:547` -- `COALESCE_GAP_THRESHOLD_DEFAULT = 1 << 20`.
- `xrspatial/geotiff/_sources.py:597` -- the merge predicate; only checks `gap <= gap_threshold`.

**Suggested fix**

Add a `max_coalesced_range_bytes` parameter (and matching `XRSPATIAL_COG_MAX_COALESCED_RANGE_BYTES` env var) to `coalesce_ranges`. Before extending the current merged range, also check that the resulting `new_end - cur_start` would not exceed the cap; if it would, seal the current merged range and start a new one. Propagate the parameter through `read_ranges_coalesced` so the COG HTTP and fsspec call sites pick up the same cap. Default: reuse `MAX_TILE_BYTES_DEFAULT` (256 MiB).

brendancol:
## Summary

- Add a `max_coalesced_range_bytes` cap to `coalesce_ranges` so a tile table with many small valid byte counts and sub-MiB gaps can no longer chain into a multi-GiB merged GET. The cap follows the per-tile cap pattern: default `MAX_COALESCED_RANGE_BYTES_DEFAULT` (256 MiB), overridable via `XRSPATIAL_COG_MAX_COALESCED_RANGE_BYTES`.
- Plumb the cap through `_HTTPSource.read_ranges_coalesced` and `_FSSpecSource.read_ranges_coalesced` so both COG remote-read paths pick up the bound automatically.
- Add eight unit tests: the 4096-tile adversarial case from the issue, byte-level round-trip after a forced split, env-var override, the zero-cap escape hatch, no false splits on legitimate back-to-back tiles, and HTTP wrapper propagation.

Closes #2266.

## Backend coverage

HTTP / cloud range-fetching code; the four-backend grid does not apply. The fix is in `_sources.py` and is exercised by both `_HTTPSource` (urllib3) and `_FSSpecSource` (cloud).

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_http_cog_coalesce.py` -- 19 passed (11 existing + 8 new)
- [x] Full `xrspatial/geotiff/tests/` minus one pre-existing `lz4` opt-in test unrelated to this change -- 5045 passed

### DIFF
--- a/xrspatial/geotiff/_cog_http.py
+++ b/xrspatial/geotiff/_cog_http.py
@@ -101,6 +101,7 @@ from ._sources import (
     # so monkeypatches against the ``_reader`` namespace continue to
     # intercept the source. PR-J / #2258.
     _HTTPSource,
+    _max_coalesced_range_bytes_from_env,
     _max_tile_bytes_from_env,
 )
 from ._validation import _validate_predictor_sample_format
@@ -901,6 +902,13 @@ def _fetch_decode_cog_http_tiles(
     # tolerates small interleaved metadata between tiles without dragging
     # in unrelated overview data. Set XRSPATIAL_COG_COALESCE_GAP=-1 to
     # disable merging (one GET per tile, the legacy behaviour).
+    #
+    # The merged-range size cap (issue #2266) is resolved here too so
+    # the call below is self-documenting: a reader can see at the call
+    # site that both ``gap_threshold`` and ``max_coalesced_range_bytes``
+    # are governed by env vars. Without the explicit lookup the cap
+    # would still apply -- ``coalesce_ranges`` resolves a ``None`` cap
+    # against the same env var -- but the asymmetry would hide that.
     try:
         workers = max(1, int(_os_module.environ.get('XRSPATIAL_COG_HTTP_WORKERS', '8')))
     except ValueError:
@@ -911,8 +919,12 @@ def _fetch_decode_cog_http_tiles(
             str(COALESCE_GAP_THRESHOLD_DEFAULT)))
     except ValueError:
         gap = COALESCE_GAP_THRESHOLD_DEFAULT
+    max_coalesced = _max_coalesced_range_bytes_from_env()
     tile_bytes_list = source.read_ranges_coalesced(
-        fetch_ranges, max_workers=workers, gap_threshold=gap)
+        fetch_ranges,
+        max_workers=workers,
+        gap_threshold=gap,
+        max_coalesced_range_bytes=max_coalesced)
 
     # Pass 3: decode each tile and place it (clipped to the window).
     #
--- a/xrspatial/geotiff/_reader.py
+++ b/xrspatial/geotiff/_reader.py
@@ -81,6 +81,7 @@ from ._sources import (  # noqa: F401
     # Public module-level constants.
     COALESCE_GAP_THRESHOLD_DEFAULT,
     MAX_CLOUD_BYTES_DEFAULT,
+    MAX_COALESCED_RANGE_BYTES_DEFAULT,
     MAX_TILE_BYTES_DEFAULT,
     # Private module-level constants and sentinels.
     _CLOUD_SCHEMES,
@@ -116,6 +117,7 @@ from ._sources import (  # noqa: F401
     _is_file_like,
     _is_fsspec_uri,
     _make_pinned_pool,
+    _max_coalesced_range_bytes_from_env,
     _max_tile_bytes_from_env,
     _mmap_cache_size_from_env,
     _open_source,
--- a/xrspatial/geotiff/_sources.py
+++ b/xrspatial/geotiff/_sources.py
@@ -644,19 +644,20 @@ def coalesce_ranges(
 
     for off, length, orig_idx in indexed[1:]:
         gap = off - cur_end
+        # Gaps may be negative if a later-listed range overlaps an
+        # earlier one; clamp ``new_end`` so the merged length covers
+        # both. ``candidate_length`` is the length the merged range
+        # would have if we extended it to include this input. We use
+        # it both to decide whether the merge is allowed under the
+        # size cap and (when it is) to update ``cur_length``.
         new_end = max(cur_end, off + length)
-        # Length the merged range would have if we extended it to include
-        # this input. Used both to decide whether the merge is allowed
-        # and (when it is) to update cur_length.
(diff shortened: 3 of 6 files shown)


brendancol (commented):
## PR Review: Cap merged-range size in coalesce_ranges to bound HTTP over-fetch (#2266)

- None.

- `xrspatial/geotiff/_cog_http.py:914` -- the call site reads `XRSPATIAL_COG_COALESCE_GAP` from env explicitly and passes it through `gap_threshold`, but it does not read `XRSPATIAL_COG_MAX_COALESCED_RANGE_BYTES` the same way. The env var is still honored because `coalesce_ranges` resolves `None` against the env, so behavior is correct. The asymmetry is a readability hazard: a future reader of the call site cannot tell from the call whether the cap is enforced. Either drop the explicit gap lookup at the call site (let `coalesce_ranges` read both env vars), or add the symmetric `max_coalesced_range_bytes` lookup at the call site. Either way the call becomes self-documenting.

- `_sources.py:657` -- the comment "Gaps may be negative if a later-listed range overlaps an earlier one; clamp so the merged length covers both" is slightly stale. The clamp is still there (`new_end = max(cur_end, off + length)`), but it moved out of the `if` body to a few lines above. Either move the comment up next to the line where the clamp actually happens, or trim it.
- `test_coalesce_caps_merged_range_size_2266` (line 138) asserts the set of mapping triples has 8 distinct entries. The next test (`test_coalesce_cap_round_trips_bytes_2266`) already covers byte-level round-trip, so the set assertion is redundant. Not a bug, just dead weight.

### What looks good
- The cap is layered on top of the gap check, not replacing it. Legitimate back-to-back tiles still coalesce.
- Both `_HTTPSource` and `_FSSpecSource` get the parameter, so the cloud path inherits the same defense.
- Re-exports in `_reader.py` keep the public import surface (`from xrspatial.geotiff._reader import ...`) consistent with the existing pattern.
- Tests cover the 4096-tile adversarial scenario, env-override, the zero-cap escape hatch, byte round-trip after a forced split, and HTTP wrapper propagation through a mock source.
- "Single oversized input is emitted intact" is the correct semantics and is documented. Rejecting oversized individual tiles belongs at the per-tile cap, not in the coalescer.

### Checklist
- [x] Algorithm correctness: bug fix; cap added correctly.
- [n/a] Backend grid: HTTP / cloud code, not a four-backend op.
- [x] Edge cases (zero cap, oversized single, legitimate back-to-back, env override).
- [n/a] Dask boundaries.
- [n/a] Benchmark: internal helper, no perf-sensitive new API.
- [n/a] README feature matrix: no new public function.

brendancol (commented):
## Follow-up review after f3ec8870

All three findings from the previous review are applied. No new findings.

- Suggestion (call-site asymmetry): fixed. `_cog_http.py:914` now resolves `_max_coalesced_range_bytes_from_env()` explicitly and passes the cap as `max_coalesced_range_bytes=...`. The env var was already honoured implicitly; the explicit lookup just makes the call site self-documenting.
- Nit (stale clamp comment): fixed. The "Gaps may be negative ..." comment moved up next to the `new_end = max(cur_end, off + length)` line where the clamp actually happens.
- Nit (redundant set assertion): fixed. The 8-distinct-mapping-triples check was removed; `test_coalesce_cap_round_trips_bytes_2266` already covers correctness after a forced split.

Test mocks in `tests/test_security.py` and `tests/test_strip_zero_dims_2053.py` picked up the new `max_coalesced_range_bytes` kwarg so the call from `_read_cog_http` keeps working under the mock.

### Checklist (re-verified)
- [x] No new blockers, suggestions, or nits.
- [x] `pytest xrspatial/geotiff/tests/` -- 5045 passed, 1 deselected (pre-existing `lz4` failure unrelated to this PR).

PR #2271, opened 2026-05-21
read_vrt: chunked missing_sources='raise' fails at build, not compute

### LINKED ISSUE #2265: Chunked VRT: missing_sources='raise' defers raise to compute, breaking 'fails immediately' contract
**Describe the bug**

`read_vrt`'s docstring says `missing_sources='raise'` (the default) "fails immediately on an unreadable backing source so a partial mosaic never surfaces silently" (xrspatial/geotiff/_backends/vrt.py:182-190). The chunked path doesn't honor that at build time.

`_read_vrt_chunked` runs an `os.path.exists` sweep over every source in the parsed VRT and records misses into `attrs['vrt_holes']` (xrspatial/geotiff/_backends/vrt.py:974-984). It then builds delayed tasks that pass `missing_sources='raise'` through to the per-chunk decodes (vrt.py:891-902). Construction succeeds. The raise only fires if the caller computes a chunk that intersects a hole.

So downstream lazy pipelines can keep running against a known-bad VRT and only blow up at compute time, usually somewhere deep in a dask graph. If the caller windows past the missing tile (see `test_compute_present_only_chunk_succeeds`), the raise never fires at all and the partial mosaic is consumed silently. The docstring says that can't happen on `'raise'`.

The scan that populates `vrt_holes` already has what it needs to raise up front. No extra decoding required.

**Expected behavior**

`read_vrt(chunks=N, missing_sources='raise')` should raise during the static missing-source scan if the scan finds any hole. Same as the eager path.

`missing_sources='warn'` keeps the current behavior: record holes in `attrs['vrt_holes']`, warn at compute time per-task.

**Steps to reproduce**

Build a 2-source VRT where one source file doesn't exist on disk. Call `read_vrt(vrt_path, chunks=4, missing_sources='raise')`. The call returns a DataArray with `attrs['vrt_holes']` populated. Only `result.compute()` (and only on chunks that touch the missing tile) raises.

**Additional context**

- Relevant file: `xrspatial/geotiff/_backends/vrt.py`
- Existing test that locks in the deferred-raise behavior: `xrspatial/geotiff/tests/test_vrt_chunked_missing_sources_1799.py::TestChunkedMissingSourcesRaise`
- That test's rationale ("an upfront raise would force the parse-time sweep to decode every source") predates the `os.path.exists` sweep that already runs at build time, so the rationale no longer holds.

brendancol:
## Summary

- Closes #2265.
- `read_vrt(chunks=N, missing_sources='raise')` now fails at construction time, not at compute time. The docstring already promised this; the chunked path didn't honor it.
- The static `os.path.exists` sweep that already runs to populate `vrt_holes` is the place the raise hooks in. No source decoding at build.
- Scoped to the requested `window=` and `band=`: a windowed or band-restricted read that doesn't touch a missing source still builds and computes, matching the eager path.
- `XRSPATIAL_GEOTIFF_STRICT=1` forces the raise regardless of the kwarg.

## Backend coverage

Build-time validation change in the dispatcher. All four backends (numpy / cupy / dask+numpy / dask+cupy) flow through the same `_read_vrt_chunked` static sweep and pick up the new behavior.

## Test plan

- [x] New test module `test_vrt_chunked_missing_raise_at_build_2265.py`:
  - Build raises immediately on `'raise'` (and on the default kwarg)
  - Error message names the `'warn'` opt-in
  - Window past missing source still builds and computes
  - Window intersecting missing source still raises
  - `band=` restriction scopes the raise (missing source on an unselected band is ignored)
  - `XRSPATIAL_GEOTIFF_STRICT=1` overrides `'warn'` kwarg
  - `'warn'` path still records `vrt_holes` and warns at compute time
- [x] Existing tests updated for the new contract (`test_vrt_chunked_missing_sources_1799.py`, `test_vrt_lazy_chunks_1814.py`, `test_read_vrt_lazy_chunks_1798.py`)
- [x] Full VRT test suite passes locally (659 tests via `pytest -k vrt`)

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -129,9 +129,18 @@ VRT missing sources
 ===================
 
 ``read_vrt`` accepts ``missing_sources='warn'`` or ``'raise'``. The default
-``'warn'`` preserves the historical behavior: unreadable source files emit
-:class:`xrspatial.geotiff.GeoTIFFFallbackWarning`, the returned DataArray
-contains ``attrs['vrt_holes']``, and the mosaic is returned with holes.
-Use ``missing_sources='raise'`` when a partial mosaic should fail the
-pipeline immediately. ``XRSPATIAL_GEOTIFF_STRICT=1`` still raises in
-``'warn'`` mode so CI environments can enforce fail-fast behavior globally.
+``'raise'`` (since #1860) fails the read immediately if any source file
+referenced by the VRT does not exist on disk. Both the eager and chunked
+dispatchers honour this at construction time -- chunked callers do not
+have to wait until ``compute()`` to learn the VRT is broken (#2265).
+The static missing-source sweep is scoped to the requested ``window=``
+and ``band=`` so a windowed or band-restricted read that does not depend
+on a missing source still succeeds.
+
+Pass ``missing_sources='warn'`` to opt into the lenient path: unreadable
+source files emit :class:`xrspatial.geotiff.GeoTIFFFallbackWarning`, the
+returned DataArray carries ``attrs['vrt_holes']``, and the mosaic is
+returned with holes left as the band's nodata sentinel (or zero on
+integer bands without a sentinel). ``XRSPATIAL_GEOTIFF_STRICT=1``
+forces the raise in ``'warn'`` mode too, so CI environments can enforce
+fail-fast behavior globally.
--- a/xrspatial/geotiff/_backends/vrt.py
+++ b/xrspatial/geotiff/_backends/vrt.py
@@ -185,8 +185,13 @@ def read_vrt(source: str, *,
         unreadable backing source so a partial mosaic never surfaces
         silently. This matches the internal ``_vrt.read_vrt`` default
         and the rest of the geotiff module's up-front rejection of
-        malformed input. Prior to #1860 the public default was
-        ``'warn'``; callers that relied on the lenient behaviour pass
+        malformed input. Both the eager and chunked dispatchers raise
+        at construction time when the static missing-source sweep
+        finds any source file that does not exist on disk and
+        intersects the requested window (#2265); chunked callers no
+        longer have to wait until ``compute()`` to learn the VRT is
+        broken. Prior to #1860 the public default was ``'warn'``;
+        callers that relied on the lenient behaviour pass
         ``missing_sources='warn'`` explicitly.
         ``'warn'`` is the opt-in escape hatch for partial mosaics: it
         emits ``GeoTIFFFallbackWarning``, records ``attrs['vrt_holes']``,
@@ -695,6 +700,7 @@ def _read_vrt_chunked(source, *, window, band, name, chunks, gpu, dtype,
     import dask.array as da
 
     from .._reader import MAX_PIXELS_DEFAULT
+    from .._runtime import _geotiff_strict_mode
     from .._vrt import (
         parse_vrt,
         _read_vrt_xml,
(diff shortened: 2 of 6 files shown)


brendancol (commented):
## PR Review: read_vrt: chunked missing_sources='raise' fails at build, not compute

None.

- [x] `xrspatial/geotiff/_backends/vrt.py:989` — the band-scope gate compared `vrt_band.band_num` (the XML's 1-based `band=` attribute) against `band + 1`, but the slice a few lines up (`selected_bands = [vrt.bands[band]]` at line 813) uses `band` as a list index. They line up for sequential VRTs, but a hand-rolled VRT with non-sequential `band=` attributes could route a hole to the wrong scope. Switched to `enumerate` so both gates use list position. Fixed in a follow-up commit.

- [ ] `xrspatial/geotiff/_backends/vrt.py:1042` — the raise message names only `chunked_holes[0]` and the total count. With several missing sources, listing a few paths instead of one would save the caller a `'warn'`-mode re-read. Not high priority; the count is enough to know to investigate.
- [ ] `xrspatial/geotiff/tests/test_vrt_chunked_missing_raise_at_build_2265.py` — no test covers multiple missing sources. Could assert the count token in the error message for that case.

### What looks good

- The fix piggybacks on the existing static `os.path.exists` sweep, so build still doesn't decode any source.
- Window-scoping (lines 1001-1008) and band-scoping (lines 988-991) mirror the eager path. A windowed or band-restricted read that misses the bad tile still builds.
- `XRSPATIAL_GEOTIFF_STRICT=1` is honored alongside the kwarg.
- Docstring and `docs/source/reference/geotiff.rst` updated. The old RST still claimed `'warn'` was the default (stale since #1860); cleaned up here as a side effect.
- The three pre-existing tests that locked in the deferred-raise behavior were rewritten under the new contract instead of deleted, so the underlying invariants are still pinned.
- New test module covers build-time raise, error-message content, window scoping, band scoping, strict-mode override, and `'warn'` preservation.

### Checklist

- [x] Algorithm matches reference (eager-path contract from #1860)
- [x] Backends share the dispatcher path, so the build-time raise applies to numpy / cupy / dask+numpy / dask+cupy
- [x] NaN handling unaffected (build-time validation only)
- [x] Edge cases covered (empty window, band restriction, strict mode)
- [x] No premature materialization (sweep only calls `os.path.exists`)
- [x] No benchmark needed
- [x] README feature matrix unchanged
- [x] Docstrings and RST docs updated

brendancol (commented):
## Follow-up review (after #2265 fixes)

- Suggestion (band-scope `vrt_band.band_num` vs list index, vrt.py:989): fixed in commit `aec6422f`. The gate now uses `enumerate` on `vrt.bands` and compares `band_idx == band` so it matches the `selected_bands = [vrt.bands[band]]` slice convention.
- Nit (raise message lists only the first hole, vrt.py:1042): fixed in commit `e57e646e`. The message now previews up to 3 missing paths and appends an "and N more" tail when truncated. Total count still reported.
- Nit (no test for multiple missing sources): added in commit `e57e646e`. Two new tests under `TestMultipleMissingSources`: one for n=2 (both paths listed) and one for n=5 (first 3 listed, "and 2 more" suffix, total count reported).

### New findings

None.

### What looks good (additions)

- New tests use a dedicated `_make_multi_missing_vrt` helper rather than monkey-patching the existing single-hole helper, so the multi-source schema stays explicit.
- Preview cap is bounded (3) and the truncation suffix includes the remainder count, so the error message size is predictable even on mosaics with hundreds of holes.
- All 24 tests in the new module pass plus the rewritten existing suites; broader `pytest -k 'vrt or strict'` is green (696 passed).

### Final checklist

- [x] Build-time raise honors `missing_sources='raise'`, the default kwarg, and `XRSPATIAL_GEOTIFF_STRICT=1`
- [x] Window-scoping and band-scoping mirror the eager path
- [x] `'warn'` semantics preserved (records `vrt_holes`, warns at compute time)
- [x] Error message previews up to 3 paths and reports the total
- [x] All four backends share the dispatcher path
- [x] Tests cover single-hole, multi-hole, window-past, band-restricted, and strict-mode cases
- [x] Docstring and RST docs updated to match the new contract

PR #2272, opened 2026-05-21
Raise RotatedTransformError for rotated GeoTIFFs

### LINKED ISSUE #2267: Rotated GeoTIFF errors bypass the typed RotatedTransformError exception
**Describe the bug**

`RotatedTransformError` (a `GeoTIFFAmbiguousMetadataError` / `ValueError`
subclass) is the typed error the reader is supposed to raise when it
rejects a rotated affine transform. The VRT read path does that via the
registered `_check_read_rotated_transform` hook in `_validation.py`.

Plain GeoTIFFs with a rotated `ModelTransformationTag` skip the hook and
raise a bare `NotImplementedError` from
`_geotags._extract_transform_and_georef` at
`xrspatial/geotiff/_geotags.py:679`.

The shared helper `_validate_read_geo_info` in `_attrs.py` even calls
this out at line 813:

> Note: the transform tuple built here is always axis-aligned
> (`b == 0` / `d == 0`) because `_transform_tuple_from_pixel_geometry`
> only carries origin + pixel size, and the upstream TIFF reader
> rejects rotated `ModelTransformationTag` entries with
> `NotImplementedError` in `_geotags._extract_transform_and_georef`
> before we reach this helper. The rotated-transform check therefore
> fires only on the VRT path.

So today, VRT rotated inputs raise `RotatedTransformError` and GeoTIFF
rotated inputs raise `NotImplementedError`. A caller using
`except RotatedTransformError` catches only the VRT case.

**To Reproduce**

```python
from xrspatial.geotiff import open_geotiff, RotatedTransformError

# A GeoTIFF with a rotated ModelTransformationTag (34264).
try:
    open_geotiff('rotated.tif')
except RotatedTransformError:
    print('caught typed error')   # never reached for the GeoTIFF path
except NotImplementedError:
    print('caught NotImplementedError')  # this is what fires today
```

**Expected behavior**

Both paths should raise `RotatedTransformError` when rejecting a rotated
transform, so one `except` clause covers both.

**Suggested fix**

Swap the `raise NotImplementedError(...)` at `_geotags.py:679` for
`raise RotatedTransformError(...)`. Keep the message and the
`allow_rotated=True` hint. Update the stale comment in
`_attrs.py:813-820`, the existing tests that assert
`NotImplementedError` on the rotated `ModelTransformationTag` path, and
the public docstrings in `__init__.py` that mention
`NotImplementedError` for the rotated case.

**Module / location**

- `xrspatial/geotiff/_geotags.py:679` -- the bare `NotImplementedError`
- `xrspatial/geotiff/_attrs.py:813-820` -- the stale comment
- `xrspatial/geotiff/_errors.py:56` -- `RotatedTransformError`
- `xrspatial/geotiff/_validation.py:1124` -- the VRT-side raise (already typed)

**Additional context**

Found by an automated sweep. The message and opt-out behaviour stay
intact; only the exception type changes. `RotatedTransformError`
subclasses `ValueError`, so any code already catching `ValueError`
keeps catching the rotated case.

brendancol:
Closes #2267.

## Summary

- Replace the bare `NotImplementedError` in
  `_geotags._extract_transform_and_georef` with the typed
  `RotatedTransformError`, so rotated GeoTIFFs and rotated VRTs share
  one exception contract.
- Update the stale comment in `_attrs.py:813` and the `__init__.py`
  docstrings that referenced the old error type.
- Update the four existing tests that asserted `NotImplementedError`
  on the rotated `ModelTransformationTag` path.
- Add `test_rotated_typed_error_2267.py` to pin the parity: parser
  raises `RotatedTransformError`, subclasses
  `GeoTIFFAmbiguousMetadataError` and `ValueError`, message still
  names the tag, and `allow_rotated=True` still reads the pixel grid.

## Backend coverage

Read-path only (numpy + dask). No GPU code touched. Writers are not
affected; the existing `drop_rotation` plumbing in `to_geotiff` stays
as-is.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_rotated_typed_error_2267.py`
  - new tests pass
- [x] `pytest xrspatial/geotiff/tests/test_allow_rotated_geotiff_2115.py
  xrspatial/geotiff/tests/test_geotags.py
  xrspatial/geotiff/tests/test_georef_status_2136.py
  xrspatial/geotiff/tests/test_remaining_fail_closed_1987.py
  xrspatial/geotiff/tests/test_ambiguous_metadata_hooks_1987.py
  xrspatial/geotiff/tests/test_http_dask_allow_rotated_2130.py
  xrspatial/geotiff/tests/test_backend_parity_matrix.py` - all green
- [x] `pytest xrspatial/geotiff/tests/` - only pre-existing failure
  (`test_lowlevel_write_pushdown_2138::test_write_vs_to_geotiff_byte_parity_uint8[lz4]`,
  unrelated lz4 experimental-codec gate) remains.

## Compatibility note

`RotatedTransformError` is a `ValueError` subclass, so anything that
caught `ValueError` keeps catching the rotated case. Code that wrote
`except NotImplementedError` specifically for the rotated branch will
not catch it any more, which is the intentional half of the fix: that
contract was the bug.


brendancol (commented):
## PR Review: Raise RotatedTransformError for rotated GeoTIFFs (#2272)

None.

- [ ] Add a `CHANGELOG.md` entry under `### Unreleased`. The repo lists
  one bullet per behaviour change with the PR number, and the original
  `RotatedTransformError` introduction already has a #1987-era entry
  there. This PR extends that typed error to the GeoTIFF entry point,
  so a sibling bullet referencing `(#2267)` keeps the breaking-but-
  intentional behaviour change visible to anyone scanning the
  changelog before upgrading. (`CHANGELOG.md`, top of `### Unreleased`)

- [ ] Update the stale historical comment at `_backends/dask.py:206`.
  It still says rotated remote GeoTIFFs "raised `NotImplementedError`
  from `_parse_cog_http_meta`". Historically accurate for pre-#2130,
  but anyone grepping for the typed-error contract will land on it
  and be briefly confused. Either retype to `RotatedTransformError`
  while keeping the historical framing, or add a `(now
  RotatedTransformError after #2267)` parenthetical. Cleanup only.

- [ ] `test_extract_transform_rotated_does_not_raise_notimplemented`
  at `xrspatial/geotiff/tests/test_rotated_typed_error_2267.py:137-153`
  is functionally identical to
  `test_extract_transform_rotated_raises_typed_error` just above it.
  `RotatedTransformError` does not subclass `NotImplementedError` (it
  inherits from `ValueError`), so the inner `except NotImplementedError:`
  clause is unreachable and the test reduces to
  `pytest.raises(RotatedTransformError)`. Either drop it or rewrite
  it to `assert not issubclass(RotatedTransformError, NotImplementedError)`,
  which actually pins the type-hierarchy contract the docstring
  claims to guard.

### What looks good

- The exception swap at `_geotags.py:684` keeps the original message
  text: tag name, coordinate term values, `allow_rotated=True` hint.
  No diagnostic regression.
- Tests hit both the unit surface (`_extract_transform` on a
  synthetic IFD) and the end-to-end surface (`open_geotiff` on a real
  on-disk rotated TIFF), and the existing #2115 / #2130 / #2136 /
  #2132 fixtures cover the dask and HTTP variants.
- The stale comment in `_attrs.py:813-820` is rewritten, not just
  patched, so it now describes the symmetric contract instead of
  contradicting it.
- `test_rotated_error_is_geotiff_ambiguous_subclass` explicitly pins
  the `GeoTIFFAmbiguousMetadataError` and `ValueError` invariants,
  so the legacy `except ValueError` path is locked down.

### Checklist

- [x] Algorithm matches reference/paper - N/A, no algorithm change
- [x] All implemented backends produce consistent results - GeoTIFF
  and VRT now share the typed error; dask and HTTP paths go through
  the same parser
- [x] NaN handling is correct - N/A
- [x] Edge cases are covered by tests - rotated + opt-in, rotated +
  default, axis-aligned passthrough; rotated + dask chunked; rotated
  + HTTP
- [x] Dask chunk boundaries handled correctly - N/A (error path)
- [x] No premature materialization or unnecessary copies - N/A
- [x] Benchmark exists or is not needed - not needed (error-path
  change)
- [x] README feature matrix updated (if applicable) - not applicable
- [x] Docstrings present and accurate - `__init__.py` docstrings
  updated

brendancol (commented):
## Follow-up review (after f02b607)

Disposition of the original review findings:

### Suggestions

- [x] **CHANGELOG entry** -- added under `### Unreleased` next to the
  existing #1987 entry that introduced `RotatedTransformError`,
  describing the GeoTIFF parity extension and the breaking-but-
  intentional half (callers catching `NotImplementedError`
  specifically must switch to `RotatedTransformError` or
  `ValueError`).
- [x] **Stale comment at `_backends/dask.py:206`** -- updated with a
  `(now RotatedTransformError after #2267)` parenthetical, preserving
  the pre-#2130 historical framing.

### Nits

- [x] **Redundant regression test** -- rewrote as
  `test_rotated_error_not_a_notimplemented_subclass`, which uses
  `issubclass` to pin the type-hierarchy invariant directly. Also
  pins the inverse (`NotImplementedError` is not a
  `GeoTIFFAmbiguousMetadataError`) so accidental hierarchy drift in
  either direction fails loudly.

All 6 tests in `test_rotated_typed_error_2267.py` still pass; the
rotated-related fixtures across `test_allow_rotated_geotiff_2115`,
`test_http_dask_allow_rotated_2130`, `test_georef_status_2136`,
`test_backend_parity_matrix`, `test_geotags`, and
`test_ambiguous_metadata_hooks_1987` are unchanged and green.

Issue #2051, opened 2026-05-18
[Security] _HTTPSource.read_all() is unbounded — small declared raster can pull huge HTTP body


brendancol:
## Summary

`_HTTPSource.read_all()` in `xrspatial/geotiff/_reader.py` (around line 1219) does not validate `Content-Length` and does not cap the body it pulls down. A crafted GeoTIFF whose header declares a tiny raster (well within `_check_dimensions`) can still be served over HTTP with a multi-gigabyte body. The full-image strip path passes that body straight into `_read_strips`, allocating the whole thing before TIFF parsing has any chance to reject it.

## Attack scenario

1. Attacker hosts a URL that:
   - Returns a valid TIFF header declaring (for example) a 100x100 image
   - Has a `Content-Length` of several GB (or no `Content-Length` at all, streaming forever)
2. Victim calls `read_geotiff(url)` or any path that lands in `_fetch_decode_cog_http_strips` with `window is None`
3. `_check_dimensions(100, 100, samples, max_pixels)` passes because the declared raster is small
4. `source.read_all()` happily downloads the whole body into memory before `_read_strips` runs
5. Result: process OOMs or fills disk via swap

## Affected code

- `xrspatial/geotiff/_reader.py` line 1219 — `_HTTPSource.read_all()` has no `max_bytes`, no Content-Length check, no streaming cap
- `xrspatial/geotiff/_reader.py` line 2416-2420 — `_fetch_decode_cog_http_strips` `window is None` branch calls `read_all()` after a pixel-count check that has nothing to do with the on-wire body size

The windowed branch in the same function is already safe: it uses per-strip ranged GETs bounded by `_max_tile_bytes_from_env()` (default 256 MiB per strip, env override `XRSPATIAL_COG_MAX_TILE_BYTES`).

## Proposed fix

Option (b) from the security report: add a `max_bytes` parameter to `read_all()`.

- Validate `Content-Length` against `max_bytes` up front and raise `OSError` if it exceeds
- Stream the body and abort once `max_bytes + 1` bytes have arrived (catches lying servers and missing Content-Length)
- Caller in `_fetch_decode_cog_http_strips` computes `max_bytes` from the strip table: roughly `max(offsets[i] + byte_counts[i])` plus a small overhead for the TIFF header

[...]

PR #2273, opened 2026-05-21
Stream HTTP read_range body so misbehaving servers cannot OOM (#2264)

### LINKED ISSUE #2264: HTTP read_range buffers full response body before slicing -- OOM on misbehaving servers
**Describe the bug**

`_HTTPSource.read_range()` in `xrspatial/geotiff/_sources.py:995` calls `_request()` with `preload_content=True` (the default), so urllib3 buffers the whole response body into `resp.data` before `read_range` sees it. The `data[:length]` slice in `_validate_range_response` (line 1067) only runs after that buffering has already happened.

The comment at `_sources.py:1067` says the slice exists so "a 16 KiB prefetch against a 2 GiB object doesn't drag the whole thing into memory". That is exactly what still happens: the full 2 GiB lands in `resp.data` first, then we throw most of it away.

The COG metadata prefetch path in `_cog_http.py:197` calls `source.read_range(0, fetch_size)` with `fetch_size` starting at `INITIAL_HTTP_HEADER_BYTES` (16 KiB) and growing to `MAX_HTTP_HEADER_BYTES` (4 MiB). Against a server that returns 200 with the full object instead of honouring `Range`, `MAX_HTTP_HEADER_BYTES` is not actually bounding what hits the process. The body has already been fully buffered by urllib3 before the cap is checked.

The existing test `test_range_ignored_200_with_full_body_is_sliced_to_length` in `tests/test_http_range_validation_1735.py` only proves the slice happens. It does not assert that the body was bounded on the wire.

**Expected behavior**

`read_range(start, length)` should bound the body at roughly `length` bytes (plus a small overshoot probe) before they hit Python-allocated memory. `read_all` already does this for the `max_bytes` path: stream with `preload_content=False`, then `_read_capped` aborts past the cap. `read_range` should follow the same pattern, or reject a 200-without-Content-Range response unless the body is proven small (e.g. `Content-Length <= length`) before buffering.

**Reproducer**

```python
import http.server, socketserver, threading
from xrspatial.geotiff._reader import _HTTPSource

PAYLOAD = b'\x00' * (200 * 1024 * 1024)  # 200 MiB

class H(http.server.BaseHTTPRequestHandler):
    def log_message(self, *a, **k): pass
    def do_GET(self):
        # Ignore Range, return full object as 200.
        self.send_response(200)
        self.send_header('Content-Length', str(len(PAYLOAD)))
        self.end_headers()
        self.wfile.write(PAYLOAD)

httpd = socketserver.TCPServer(('127.0.0.1', 0), H)
port = httpd.server_address[1]
threading.Thread(target=httpd.serve_forever, daemon=True).start()

import os; os.environ['XRSPATIAL_GEOTIFF_ALLOW_PRIVATE_HOSTS'] = '1'
src = _HTTPSource(f'http://127.0.0.1:{port}/x.bin')
# Caller asks for 16 KiB. urllib3 buffers all 200 MiB into resp.data
# before _validate_range_response slices it down.
out = src.read_range(0, 16 * 1024)
assert len(out) == 16 * 1024
```

**Additional context**

This is the same class of risk that #2051 fixed for `read_all`, but on the `read_range` path. The fix should mirror that approach: stream with `preload_content=False` and cap with `_read_capped`.

The 200-no-Content-Range branch is the only path that needs streaming. The 206 branch already gets a server-attested length via `Content-Range`, so buffering there is bounded by the requested range.

Touches `xrspatial/geotiff/_sources.py` (`_HTTPSource.read_range`, `_validate_range_response`). No backend dispatch involved, since this is HTTP-only.

brendancol:
## Summary

- Switch `_HTTPSource.read_range()` to streaming (`preload_content=False`) and cap the body on the wire with `_read_capped`, matching the `read_all` pattern from #2051. Before this change urllib3 buffered the whole response into `resp.data` *before* the `data[:length]` slice in `_validate_range_response` ran, so a server that ignored `Range` and returned a 200 with a multi-GiB body still OOM'd the process. `MAX_HTTP_HEADER_BYTES` was advisory, not enforcing.
- For the 200-without-Content-Range fallback (server ignored `Range` on a `start=0` prefetch), the legitimate "small file served whole" case still works, but the slack is capped at a tight absolute (`_RANGE_IGNORED_FULL_OBJECT_CAP = 16 MiB`). The validator still slices that fallback body down to `length` for the caller.
- Two enforcement paths: Content-Length past the cap is rejected on the preflight before any body bytes are read; chunked or missing Content-Length is aborted mid-stream by `_read_capped`. The 206-with-Content-Range path keeps its existing behaviour but now also goes through the streaming path so the cap is uniform.
- HTTP-only path. No backend dispatch involved (numpy / cupy / dask+numpy / dask+cupy all share this code through the geotiff reader).

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_http_range_validation_1735.py`: 10 passing. Covers honest-oversize-rejected, slack-fallback-sliced, short-body-returned, dishonest-oversize-streamed-and-capped, the `preload_content=False` wire contract, and the pre-existing 1735 regression cases.
- [x] `pytest xrspatial/geotiff/tests/test_sidecar_ovr_2112.py`: 28 passing. Includes the SimpleHTTPRequestHandler-ignores-Range scenario that drove the slack-cap design.
- [x] `pytest xrspatial/geotiff/tests/test_ssrf_hardening_1664.py test_dns_rebinding_pin_issue_1846.py test_http_no_stdlib_fallback_2050.py test_http_meta_buffer_1718.py`: 68 passing. Mock-fixture updates absorbed cleanly.
- [x] Full `pytest xrspatial/geotiff/tests/`: 5041 passing. One pre-existing lz4 failure on `main` unrelated to this change.

Closes #2264.

### DIFF
--- a/xrspatial/geotiff/_sources.py
+++ b/xrspatial/geotiff/_sources.py
@@ -992,6 +992,16 @@ class _HTTPSource:
             return self._pool
         return self._get_pinned_pool(scheme, host, parsed.port, pinned_ip)
 
+    #: Absolute ceiling for a ``read_range(0, length)`` whose server
+    #: ignores ``Range`` and returns a 200 with the full object body. Up
+    #: to this size we tolerate the misbehaviour (the response is sliced
+    #: down to ``length``); beyond it we refuse to buffer. The value is
+    #: large enough to cover a typical sidecar/header prefetch served
+    #: from a Range-blind origin (kilobytes to a few MiB) but well below
+    #: the multi-gigabyte regime the OOM guard exists to prevent. Issue
+    #: #2264.
+    _RANGE_IGNORED_FULL_OBJECT_CAP = 16 * 1024 * 1024
+
     def read_range(self, start: int, length: int) -> bytes:
         # Match the ``b''``-for-non-positive-length convention used by
         # other source implementations (e.g. ``_BytesIOSource``).
@@ -1003,14 +1013,74 @@ class _HTTPSource:
             return b''
         end = start + length - 1
         headers = {'Range': f'bytes={start}-{end}'}
-        resp = self._request(headers=headers)
+        # Stream the body so a server that ignores ``Range`` and returns
+        # the full object as 200 cannot drag arbitrarily many bytes into
+        # process memory before the cap is applied. Without
+        # ``preload_content=False`` urllib3 buffers the full response into
+        # ``resp.data`` *before* this method returns, so the slice in
+        # ``_validate_range_response`` only ever ran after the body was
+        # already resident. Mirrors the ``read_all`` streaming path that
+        # #2051 introduced. Issue #2264.
+        resp = self._request(headers=headers, preload_content=False)
+        try:
+            content_range = resp.headers.get('Content-Range')
+            content_length = resp.headers.get('Content-Length')
+            # Choose the byte budget for ``_read_capped``. A 206 with
+            # Content-Range is bounded by ``length`` (the validator
+            # below also re-checks this against the advertised range).
+            # A 200 without Content-Range means the server ignored
+            # ``Range``; allow the full-object slack only when the
+            # caller asked from ``start=0`` (the only case where the
+            # full body offsets line up with what was requested).
+            if content_range is None and start == 0:
+                cap = self._RANGE_IGNORED_FULL_OBJECT_CAP
+            else:
+                cap = length
+            self._check_range_content_length(content_length, cap)
+            data = self._read_capped(resp, cap)
+        finally:
+            try:
+                resp.release_conn()
+            except Exception:  # noqa: BLE001
+                pass
         return self._validate_range_response(
             status=resp.status,
-            content_range=resp.headers.get('Content-Range'),
-            data=resp.data,
+            content_range=content_range,
+            data=data,
             start=start,
             length=length,
         )
+
+    @staticmethod
+    def _check_range_content_length(raw, cap: int) -> None:
+        """Reject a range response whose advertised ``Content-Length``
+        exceeds the byte budget for the call.
+
+        For a ``Range: bytes=S-E`` request, a well-behaved server returns
+        either a 206 with body length ``E - S + 1`` or a 200 with the
+        full object body. A server that ignores ``Range`` and returns a
+        multi-gigabyte 200 advertises that size up front; rejecting on
+        the header lets us bail before any body bytes are read.
+
+        Missing or unparseable ``Content-Length`` returns silently --
+        the streaming cap in :meth:`_read_capped` is the real defence
+        and will catch an over-sized body whether the header was honest,
+        dishonest, or absent. Issue #2264.
+        """
+        if raw is None:
+            return
+        try:
+            declared = int(raw)
+        except (TypeError, ValueError):
+            return
+        if declared > cap:
+            raise OSError(
+                f"HTTP range response declares Content-Length="
+                f"{declared:,} bytes, which exceeds the byte budget of "
+                f"{cap:,} bytes for this request. The server likely "
+                f"ignored the Range header and returned a multi-MiB "
+                f"body. Issue #2264."
+            )
 
     @staticmethod
     def _validate_range_response(*, status, content_range, data,
--- a/xrspatial/geotiff/_sources.py
+++ b/xrspatial/geotiff/_sources.py
@@ -998,11 +998,40 @@ class _HTTPSource:
     #: down to ``length``); beyond it we refuse to buffer. The value is
     #: large enough to cover a typical sidecar/header prefetch served
     #: from a Range-blind origin (kilobytes to a few MiB) but well below
-    #: the multi-gigabyte regime the OOM guard exists to prevent. Issue
-    #: #2264.
+    #: the multi-gigabyte regime the OOM guard exists to prevent.
+    #:
+    #: This is the worst-case *transient* buffer per ``read_range``
+    #: call. :meth:`read_ranges` fans out across up to 8 threads, so a
+    #: full pool against a Range-blind server can hold roughly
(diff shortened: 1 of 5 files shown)


brendancol (commented):
## PR Review: Stream HTTP read_range body so misbehaving servers cannot OOM (#2264)

None.

- [ ] `xrspatial/geotiff/_sources.py:1005` - `read_range` has no docstring on `_HTTPSource`. The base `Source.read_range` is documented at the module level (line 9), but the HTTP override now carries non-trivial extra contract (Content-Length preflight, streaming cap, 16 MiB slack for `start=0` fallback, OSError on overshoot). A short docstring summarising "raises OSError when the server's body is past the byte budget for this call" would save the next reader from chasing comments.

- [ ] `xrspatial/geotiff/_sources.py:1137-1142` - The slice comment in `_validate_range_response` still says "a 16 KiB prefetch against a 2 GiB object doesn't drag the whole thing into memory". The whole point of this PR is that the slice never actually achieved that. With the new cap the worst case is `_RANGE_IGNORED_FULL_OBJECT_CAP` (16 MiB) of transient buffer, not 2 GiB. Update the comment so it does not contradict the new behaviour.

- [ ] `xrspatial/geotiff/_sources.py:1310-1312` - `_read_capped`'s OSError message refers only to "Issue #2051". When reached from `read_range` (the new caller) the trail leads readers to the wrong issue. Either reference both issues or drop the specific number and rely on the call site's message for context.

- [ ] `xrspatial/geotiff/_sources.py:1003` - `_RANGE_IGNORED_FULL_OBJECT_CAP = 16 MiB` is the maximum *transient* buffer per call. With `read_ranges` running up to 8 workers concurrently, the worst-case transient is ~128 MiB if every worker hits a Range-blind server. The current code is still vastly better than the pre-#2264 behaviour, but it's worth noting this in the constant's docstring or in `read_ranges` so future tuning is informed.

- [ ] `xrspatial/geotiff/_sources.py:995-1003` - `_RANGE_IGNORED_FULL_OBJECT_CAP` is a class attribute, which is fine, but it sits next to other module-level HTTP knobs (`_HTTP_MAX_REDIRECTS`, `INITIAL_HTTP_HEADER_BYTES` in `_cog_http`). Promoting it to module-level would help the next person tweaking HTTP behaviour find it. Class-level does make the `monkeypatch.setattr(_HTTPSource, ...)` pattern in the tests cleaner, so it's a judgement call.

- [ ] `xrspatial/geotiff/_sources.py:1024-1045` - The `try/finally` around `_check_range_content_length` and `_read_capped` runs `resp.release_conn()` in finally, then `_validate_range_response` reads `resp.status` and `resp.headers` *after* release_conn. This works (urllib3 keeps status/headers populated after release), but the flow is slightly non-obvious. Either copy `status` and `content_range` into locals before the finally, or move `_validate_range_response` inside the try block. The `read_all` method has the same shape, so consistency is also fine.

### What looks good

- The streaming-with-preflight pattern matches the `read_all` design from #2051. Symmetry across the two HTTP read paths is easier to audit than two different fixes.
- The test plan covers four interesting cases: honest oversize via Content-Length, dishonest oversize via chunked, legitimate small-file fallback, and the `preload_content=False` wire contract. The wire-contract test in particular is the right kind of regression bait - a future refactor that flipped the default would fail it immediately.
- The mock-fixture updates in test_ssrf_hardening_1664.py and test_dns_rebinding_pin_issue_1846.py are scoped and minimal: just `stream()`, `release_conn()`, and `Content-Length`. No drive-by changes to the SSRF or DNS-rebinding test logic.
- The existing test `test_range_ignored_200_with_full_body_is_sliced_to_length` was correctly renamed and updated to reflect the new behaviour (now `test_range_ignored_200_full_object_sliced_within_cap`). The old test name was the giveaway: it asserted slicing happened, not that buffering was bounded.

### Checklist

- [x] Algorithm matches reference/paper - HTTP-only path, no algorithm involved.
- [x] All implemented backends produce consistent results - HTTP-only, no backend dispatch.
- [x] NaN handling is correct - N/A, byte-level code.
- [x] Edge cases are covered by tests - empty body, short body, oversize body, chunked oversize, zero-length, nonzero start.
- [x] Dask chunk boundaries handled correctly - N/A.
- [x] No premature materialization or unnecessary copies - `_read_capped` joins chunks once at end.
- [x] Benchmark exists or is not needed - not needed for a defensive HTTP fix.
- [x] README feature matrix updated (if applicable) - N/A.
- [x] Docstrings present and accurate - partially; see suggestion on `read_range` docstring.

brendancol (commented):
## Follow-up review after edc3313

Reviewed the second commit. Verdict: ready to merge once CI is green.

- Suggestion (no docstring on `_HTTPSource.read_range`) -- fixed at `_sources.py:1012-1034`. Docstring lists status / Content-Length / Content-Range axes and names the OSError contract.
- Suggestion (stale 2 GiB comment in `_validate_range_response`) -- fixed. The comment now cites `_RANGE_IGNORED_FULL_OBJECT_CAP` (16 MiB) and references #2264.
- Suggestion (`_read_capped` references only #2051) -- fixed. The OSError message now lists both #2051 and #2264.
- Suggestion (document the 128 MiB worst-case under `read_ranges`) -- fixed in the constant's docstring.
- Nit (release_conn / validator ordering) -- fixed. `_validate_range_response` now runs inside the try block; `release_conn` runs in finally after validation completes or raises.
- Nit (class-level vs module-level constant) -- left as-is. Class-level keeps the `monkeypatch.setattr(_HTTPSource, ...)` test pattern clean, which the test suite relies on.

### What looks good

- Docstring is explicit about which budget applies in which branch, which is the part that took the longest to reason about from the original code.
- The validator-inside-try refactor is a small readability win without changing behaviour. Easy to verify the response is alive when status/headers are read.

### Outstanding
None blocking. CI is the last gate.

PR #2279, opened 2026-05-22
Shut tile-compress pool down on streaming-write error path (#2276)

### LINKED ISSUE #2276: ThreadPoolExecutor leak on mid-stream failure in tiled writes
**Describe the bug**

`_write_streaming` in `xrspatial/geotiff/_writer.py` creates a `ThreadPoolExecutor` for per-tile compression on the streaming tiled-write path. The pool is constructed at line 1026 and shut down with `tile_pool.shutdown(wait=True)` only after the tile-row loop finishes (line 1131).

If anything raises in between (a compression error, a dask compute failure, a disk write failure), the shutdown call is skipped. The outer `try/except BaseException` one frame up cleans up the temp file but never touches the pool. Worker threads survive the failure path and the executor leaks.

Fix: wrap the executor lifetime in `try/finally`, or use it as a context manager (`with ThreadPoolExecutor(...) as executor:`) so shutdown is guaranteed. In-flight futures should be cancelled/awaited cleanly on the error path.

**Expected behavior**

On any exception raised during the streaming tiled write, the `ThreadPoolExecutor` is shut down before the exception propagates. No worker thread should outlive the failed call.

**Repro sketch**

- Call `_write_streaming` with a dask-backed input that compresses (e.g. `compression='deflate'`) and a `streaming_buffer_bytes` small enough to force multiple horizontal segments.
- Inject a failure in the per-tile work (monkey-patch `_compress_block` to raise on the second tile, or patch dask compute to raise after the first segment).
- Confirm that no `ThreadPoolExecutor` worker threads remain alive after the call returns.

**Additional context**

`test_streaming_write_parallel.py` covers the happy path but does not exercise the failure path.

brendancol:
Closes #2276

## What

- Wrap the streaming-write tile loop in `try/finally` so the per-tile compression `ThreadPoolExecutor` is shut down on every exit path, including mid-stream exceptions.
- Cancel still-pending futures before `shutdown(wait=True)` so the error path does not block on work we no longer need.
- Tag the pool's worker threads with `thread_name_prefix='xrspatial-geotiff-tile-compress'` so leak detection can tell them apart from dask's offload / scheduler pools.

## Why

Before this change, `tile_pool.shutdown(wait=True)` lived past the tile-row loop. A raise from `_compress_block`, from a dask compute, or from the sequential file write bypassed the shutdown call. The outer `try/except BaseException` one frame up cleans up the temp file but never touches the pool, so worker threads outlive the failed write.

## Backend coverage

Writer-only path (numpy / dask). No GPU code touched.

## Test plan

- [x] `xrspatial/geotiff/tests/test_streaming_write_pool_leak_2276.py` covers three mid-stream failure shapes and the happy path:
  - `test_pool_shutdown_on_compress_failure` raises from inside `_compress_block` after a few successful calls
  - `test_pool_shutdown_on_file_write_failure` raises from the sequential file-write step after the parallel compress already ran
  - `test_pool_shutdown_on_happy_path` is a regression guard that the rewrite did not break clean-exit shutdown
- [x] Each test asserts both the pool's `_shutdown` flag and the absence of leaked worker threads named with the writer's prefix.
- [x] Confirmed tests fail without the fix.
- [x] `pytest xrspatial/geotiff/tests/test_streaming_write_parallel.py xrspatial/geotiff/tests/test_streaming_write.py xrspatial/geotiff/tests/test_streaming_codecs_2026_05_11.py` -- 43 pass, 1 skipped (perf gate).


brendancol (commented):
## PR Review: Shut tile-compress pool down on streaming-write error path (#2276)

- None.

- `xrspatial/geotiff/_writer.py:1133-1143`: the inner `try/finally` clears `_inflight_futures = []` as soon as the list comprehension exits, including the exception path. If `fut.result()` raises for fut[3] of 8 submitted, the outer `finally` at line 1161 sees an empty list and cannot cancel the still-pending fut[4..7]. `tile_pool.shutdown(wait=True)` then blocks until those pending tasks drain. Either drop the inner clear, or call `tile_pool.shutdown(wait=True, cancel_futures=True)` (Python 3.9+) so pending work is dropped on the error path. The second option is simpler and removes the need for the `_inflight_futures` bookkeeping entirely.

- `xrspatial/geotiff/tests/test_streaming_write_pool_leak_2276.py:21-23`: the module docstring still describes the default `ThreadPoolExecutor-` thread name, but the implementation filters on `_WRITER_POOL_PREFIX = 'xrspatial-geotiff-tile-compress'`. Update the docstring to match.

- `CHANGELOG.md`: no CHANGELOG entry. Recent bug-fix PRs add a bullet under "Unreleased / Bug fixes and improvements" referencing the issue number. Add one.

- `xrspatial/geotiff/_writer.py:1026-1033`: the `'xrspatial-geotiff-tile-compress'` literal is duplicated between the writer and the test. Consider a module-level constant in `_writer.py` (e.g. `_TILE_POOL_THREAD_PREFIX`) so the test imports it.

- `xrspatial/geotiff/tests/test_streaming_write_pool_leak_2276.py:189-207`: `monkeypatch.setattr(os, 'fdopen', ...)` patches the global. It is reverted at teardown, but anything in the same test session that calls `os.fdopen` during the patched window gets intercepted. Scope is narrow enough here that it is fine to leave.

### What looks good
- `try/finally` placement is correct: pool construction happens before the try block, so the finally always has a valid `tile_pool` to act on.
- The `thread_name_prefix` is an effective way to tell writer pools apart from dask's offload and scheduler pools when checking for leaks.
- Three failure shapes plus a happy-path regression guard is solid coverage. The tests fail without the fix.
- Temp filenames include the issue number per project convention.
- Existing streaming-write suite (43 tests) still passes.

### Checklist
- [x] Algorithm matches reference/paper (N/A)
- [x] All implemented backends produce consistent results (writer-only)
- [x] README feature matrix updated (N/A)
- [x] Docstrings present and accurate (one minor drift noted)

brendancol (commented):
## Follow-up review (after commit 1dcdb567)

All three Suggestions addressed. The first Nit applied; the second Nit dismissed with reason. No new findings.

### Suggestion dispositions
- `xrspatial/geotiff/_writer.py:1133-1143` (cancel_futures): **fixed**. Replaced the inner `try/finally` + `_inflight_futures` bookkeeping with `tile_pool.shutdown(wait=True, cancel_futures=True)`. The error path now drops queued work instead of blocking the unwind. Comment block updated to reflect the new approach.
- `xrspatial/geotiff/tests/test_streaming_write_pool_leak_2276.py:21-23` (docstring drift): **fixed**. Module docstring rewritten to describe the writer-prefix filtering and the dask-pool false-positive rationale instead of the old default-prefix wording.
- `CHANGELOG.md`: **fixed**. Added bullet under "Unreleased / Bug fixes and improvements" referencing #2276.

### Nit dispositions
- Shared `_TILE_POOL_THREAD_PREFIX` constant: **fixed**. Hoisted into `_writer.py` as a module-level constant; the test now imports it from `writer_mod` so a future rename of the prefix on the writer side updates the test in lockstep.
- `os.fdopen` monkeypatch global scope: **dismissed**. Pytest's `monkeypatch` fixture reverts the attribute at test teardown, the patched window covers only the body of `to_geotiff` for one call, and the failure injection is deterministic on call count. No realistic risk of cross-test contamination.

### Verification
- `pytest xrspatial/geotiff/tests/test_streaming_write_pool_leak_2276.py xrspatial/geotiff/tests/test_streaming_write_parallel.py xrspatial/geotiff/tests/test_streaming_write.py` -- 37 pass, 1 skipped (perf gate).

No remaining findings. Pool shutdown is now guaranteed on every exit path, queued work is dropped on the error path, and the test/code constant stays in sync.

PR #2280, opened 2026-05-22
Resolve CRS model type via pyproj instead of EPSG number range (#2277)

### LINKED ISSUE #2277: CRS model type guessed from EPSG number range corrupts geographic CRSes outside 4000-4999
**Describe the bug**

`build_geo_tags` in `xrspatial/geotiff/_geotags.py` decides whether to write `GeographicTypeGeoKey` or `ProjectedCSTypeGeoKey` from the EPSG number range:

```python
# xrspatial/geotiff/_geotags.py:1403
if crs_epsg == 4326 or (crs_epsg >= 4000 and crs_epsg < 5000):
    model_type = MODEL_TYPE_GEOGRAPHIC
else:
    model_type = MODEL_TYPE_PROJECTED
```

EPSG codes aren't allocated by type. Plenty of geographic CRSes live outside the 4000-4999 window. A few examples that get mis-tagged today:

- EPSG:6318, NAD83(2011) geographic 2D
- EPSG:7844, GDA2020 geographic 2D
- EPSG:9057, WGS 84 (G2139) geographic 3D
- EPSG:8252, NAD83(FBN) geographic 2D

All four get written with `ProjectedCSTypeGeoKey`. That's CRS corruption at write time. A spec-compliant reader will treat the file as projected with linear units of metres, when the coordinates are actually decimal degrees on an ellipsoid.

The current test suite only covers EPSG 4326 and 32610 (`xrspatial/geotiff/tests/test_geotags.py:80`), so this hole is untested.

**Expected behavior**

`build_geo_tags` should consult pyproj when available and use `CRS.from_epsg(crs_epsg).is_geographic` to decide the GeoKey. If pyproj isn't installed and the EPSG code isn't one of the few that can be hard-coded with confidence (4326 plus the 4000-4999 geographic block), the writer should raise rather than guess. Silent CRS corruption is worse than an explicit error at write time.

**Additional context**

Fix sketch:

1. A helper that returns the model type for an EPSG code, preferring pyproj and falling back to a tight hard-coded allowlist.
2. Call it from `build_geo_tags` in place of the range check.
3. Honour `XRSPATIAL_GEOTIFF_STRICT` on the fail-closed path so behaviour matches the rest of the geotiff module.
4. Tests for EPSG 6318 and 7844 (geographic, outside 4000-4999) plus an unknown EPSG with pyproj patched out to exercise the raise path.

brendancol:
## Summary

- `build_geo_tags` picked geographic vs projected from the EPSG number range (4326 plus 4000-4999). Geographic CRSes outside that window (6318, 7844, 9057, 8252) were silently written as projected, corrupting the CRS at write time.
- Now it calls `pyproj.CRS.from_epsg(...).is_geographic` when pyproj is available, and falls back to a tight hard-coded set (4326 + 4000-4999) when it isn't. Anything outside the fallback set without pyproj raises `UnknownCRSModelTypeError` instead of guessing.
- New typed error `UnknownCRSModelTypeError` joins the existing `GeoTIFFAmbiguousMetadataError` family and is exported from `xrspatial.geotiff`.

## Backend coverage

Writer path only. Not backend-dispatched: the geotag construction is plain Python and runs identically on every backend's output.

## Test plan

- [x] `TestModelTypeFromEPSG` covers the four geographic codes that regressed (6318, 7844, 9057, 8252).
- [x] Parametrized projected codes (32610, 3857, 2193) to confirm dispatch the other way.
- [x] Legacy-range geographic codes (4326, 4269, 4267) still resolve to `GEOKEY_GEOGRAPHIC_TYPE`.
- [x] `test_unknown_epsg_without_pyproj_raises` mocks the pyproj import failure and asserts `UnknownCRSModelTypeError`.
- [x] `test_known_geographic_without_pyproj_still_works` keeps EPSG 4326 working when pyproj is gone (via the hard-coded fallback set).
- [x] `test_unknown_epsg_with_pyproj_raises` exercises the case where pyproj is installed but the EPSG code isn't in its database.
- [x] Full geotiff test suite: 5084 passed.

Closes #2277


brendancol (commented):
## PR Review: Resolve CRS model type via pyproj instead of EPSG number range (#2277)

None.

- `_KNOWN_GEOGRAPHIC_EPSG_FALLBACK = frozenset(range(4000, 5000))` (xrspatial/geotiff/_geotags.py:1328) is the same kind of range guess this PR is fixing, just smaller. Codes like 4087 / 4088 (World Equidistant Cylindrical) and 4499 (CGCS2000 / Gauss-Kruger zone 21) are projected, not geographic. When pyproj is unavailable, the fallback will mis-tag them as geographic, which is the exact dual of the original bug. Two options: tighten the fallback to a vetted allowlist of true geographic codes (e.g. `{4326, 4269, 4267, ...}`), or drop the fallback entirely and always require pyproj for codes outside `{4326}`. The latter matches the issue's "fail closed rather than guess" principle most strictly.
- Behaviour change worth flagging: callers who write projected GeoTIFFs (e.g. EPSG 3857, 32610) without pyproj installed used to write the correct projected key and now hit `UnknownCRSModelTypeError`. The fix is right (without pyproj we genuinely can't classify the code), but it's worth a line in the CHANGELOG when one is cut, and `pyproj` may need to move into `requirements.txt` / `setup.cfg` if it isn't already a hard dependency.

- xrspatial/geotiff/tests/test_geotags.py:147 -- `flat.index(GEOKEY_GEOGRAPHIC_TYPE)` is safe for the tested EPSG values but is fragile if another field in the flat tuple ever equals 2048. Walking the header offset (`4 + key_position*4 + 3`) would be more robust.
- xrspatial/geotiff/tests/test_geotags.py:243 -- `assert _geotags is not None` just to silence "unused import" reads as noise; drop the `from xrspatial.geotiff import _geotags` import since the test doesn't actually use it.
- xrspatial/geotiff/_geotags.py:1326-1328 comment says "Used only when pyproj is unavailable" but the fallback set is also consulted in the pyproj-installed-but-`CRS.from_epsg`-failed branch (line 1357). Worth updating to mention both paths.

### What looks good
- `_model_type_from_epsg` is small and replaces a single hot site cleanly.
- `UnknownCRSModelTypeError` slots into the existing `GeoTIFFAmbiguousMetadataError` family and is exported from `xrspatial.geotiff`, so callers can except on it.
- Tests cover the four reported geographic codes (6318, 7844, 9057, 8252), the legacy-range cases, projected dispatch, and both flavours of fail-closed (pyproj missing, pyproj installed but DB lookup fails).
- Error messages reference the issue number and give actionable remediation steps.

### Checklist
- [x] Algorithm matches reference -- pyproj's `is_geographic` is the authoritative answer
- [x] Backend coverage N/A (writer-only, pure-Python tag construction)
- [x] No NaN handling concerns (integer EPSG codes)
- [x] Edge cases covered: legacy-range geographic, outside-range geographic, projected, unknown
- [x] No dask/cupy/multiprocessing surface affected
- [x] No materialization concerns
- [x] Benchmark not applicable (constant-time decision per write)
- [x] README feature matrix not affected
- [x] Docstrings present and accurate for the new helper and error class

brendancol (commented):
## Follow-up review (after 20122384)

Re-checked the diff after the follow-up commit. Disposition of the original findings:

### Original findings disposition
- Suggestion 1 (tighten the 4000-4999 fallback) -- **fixed.** Allowlist is now nine hand-verified geographic codes; tests added for 4087 / 4088 / 4499 confirm the projected-inside-block direction.
- Suggestion 2 (behaviour change / pyproj dep) -- **fixed.** `pyproj` is now in `setup.cfg [options.extras_require] geotiff`, and CHANGELOG has an Unreleased entry that names the behaviour change explicitly.
- Nit 1 (`flat.index` fragility) -- **fixed.** Assertion dropped.
- Nit 2 (`assert _geotags is not None`) -- **fixed.** Import dropped.
- Nit 3 (docstring "Used only when pyproj is unavailable") -- **fixed.** Both fallback paths are now mentioned.

### Additional change worth noting
- The public-API contract test `test_features.py::TestPublicAPI::test_all_lists_supported_functions` was updated to include `UnknownCRSModelTypeError` (caught the addition correctly, the test class is doing its job).

### New issues
None spotted. `pyproj` is now a hard dependency for callers passing EPSG codes outside the small allowlist; that's the right tradeoff and is documented.

### Status
No remaining blockers, suggestions, or nits. Ready from a code-review perspective; merge gated on CI.

PR #2281, opened 2026-05-22
Reject signed-integer MinIsWhite TIFFs on read and write (#2278)

### LINKED ISSUE #2278: Signed-integer MinIsWhite TIFFs round-trip with semantically wrong photometric meaning
**Describe the bug**

The TIFF reader in `_apply_photometric_miniswhite` (`xrspatial/geotiff/_decode.py:870`) only inverts unsigned and float data when `PhotometricInterpretation = MinIsWhite` (photometric == 0). The writer mirrors that pass-through in `_apply_photometric_miniswhite_invert` (`xrspatial/geotiff/_encode.py:122`).

Signed-integer pixel types with MinIsWhite are accepted on both read and write but never inverted. They round-trip through xrspatial unchanged, but the on-disk tag declares "high values are dark" while the stored values sit in the MinIsBlack domain. Any standards-compliant TIFF reader (GDAL, libtiff, ImageMagick) will display the image inverted.

The existing `test_miniswhite_int_passthrough` in `tests/test_miniswhite_writer_roundtrip_1836.py` documents the passthrough as intentional for round-trip purposes. That test pins behaviour that is semantically wrong against any other consumer of the file.

**Expected behavior**

Two paths to consider:

- Option A (reject): treat signed MinIsWhite as unsupported. Raise on both read and write with the SampleFormat / BitsPerSample / Photometric values in the message. Document the limitation.
- Option B (support): implement signed-domain inversion (`min + (max - x)`, which simplifies to `-1 - x` for signed ints), applied symmetrically on read and write, with care around the type's min value to avoid overflow.

Option A is the right call here. Failing closed matches how this codebase already handles other ambiguous TIFF metadata cases (the `GeoTIFFAmbiguousMetadataError` family in `_errors.py`). Signed MinIsWhite is rare in real GIS data anyway: signed rasters are typically DEMs or signed reflectance, both of which use MinIsBlack, and MinIsWhite in the wild is almost always unsigned grayscale scanned imagery. Right now a user writing a signed raster with `photometric='miniswhite'` gets a file that opens wrong in GDAL or libtiff and they have no way to know until someone else opens it. Raising at write time tells them on the spot.

Implementing the signed inversion would also drag a parallel signed-domain branch into the writer's nodata sentinel inversion, and nobody has asked for the support.

**Files affected**

- `xrspatial/geotiff/_decode.py:870-878` (`_apply_photometric_miniswhite`)
- `xrspatial/geotiff/_encode.py:122-143` (`_apply_photometric_miniswhite_invert`)
- `xrspatial/geotiff/tests/test_miniswhite_writer_roundtrip_1836.py:77-85` (the existing `test_miniswhite_int_passthrough` will flip to assert the new error)

**Additional context**

Found during a deep sweep of the geotiff module. Related to the earlier MinIsWhite work in #1809 and #1836.

brendancol:
## Summary

Closes #2278.

The reader and writer used to silently let signed-integer pixels pass through unchanged when `PhotometricInterpretation = MinIsWhite`. That round-tripped fine inside xrspatial but produced files whose pixel values disagreed with the on-disk Photometric tag against every other TIFF consumer (GDAL, libtiff, ImageMagick) -- basically a guaranteed inverted display anywhere else.

Both paths now raise `NotImplementedError` with the SampleFormat, BitsPerSample, and Photometric values in the message so callers can diagnose without digging into the spec.

I went with Option A (reject) rather than Option B (implement signed inversion). The rationale is in the issue: failing closed is what the rest of this codebase already does for ambiguous TIFF metadata, signed MinIsWhite is rare in real GIS data anyway, and silent passthrough was the worst-of-both-worlds outcome -- looked fine locally, broken everywhere else.

## Test plan

- [x] New `xrspatial/geotiff/tests/test_signed_miniswhite_rejected_2278.py`: parameterised write rejection across `int8`/`int16`/`int32`/`int64`, a no-partial-file check, a read-path test that forges a signed MinIsWhite file by writing a normal int16 MinIsBlack TIFF and flipping the Photometric tag to 0 in place, and non-regression coverage that the unsigned and float MinIsWhite round-trips still work.
- [x] Flipped the existing `test_miniswhite_int_passthrough` in `test_miniswhite_writer_roundtrip_1836.py` to assert the new error (renamed `test_miniswhite_int_rejected_at_write_2278`).
- [x] Full `xrspatial/geotiff/tests/` run locally: 5079 passed, 68 skipped.

## Backend coverage

Bug fix in the shared CPU read/write photometric helpers. The reader change covers both the local-file and HTTP COG paths (both go through `_reader._apply_photometric_miniswhite`). The writer change covers the eager write path. The dask streaming writer already rejects `miniswhite` outright for an unrelated reason (#1836), so signed-int + dask is covered transitively.

### DIFF
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -845,6 +845,24 @@ def read_geotiff_gpu(source: str, *,
                 arr_gpu = np.iinfo(gpu_dtype).max - arr_gpu
             elif gpu_dtype.kind == 'f':
                 arr_gpu = -arr_gpu
+            elif gpu_dtype.kind == 'i':
+                # Mirror the CPU reader's signed-int MinIsWhite
+                # rejection (issue #2278). Without this the pure-GPU
+                # decode path would silently return un-inverted
+                # signed pixels while the CPU decode path raises,
+                # breaking backend parity.
+                raise NotImplementedError(
+                    f"Signed-integer MinIsWhite TIFFs are not "
+                    f"supported on the GPU decode path either "
+                    f"(issue #2278): Photometric=0 (MinIsWhite), "
+                    f"SampleFormat={ifd.sample_format} (signed int), "
+                    f"BitsPerSample={ifd.bits_per_sample}, "
+                    f"dtype={gpu_dtype}. xrspatial has no "
+                    f"semantically correct inversion for signed "
+                    f"pixels here. Convert the file to MinIsBlack "
+                    f"(Photometric=1) with another tool, or open it "
+                    f"in an unsigned dtype."
+                )
 
         if name is None:
             import os
--- a/xrspatial/geotiff/_decode.py
+++ b/xrspatial/geotiff/_decode.py
@@ -868,13 +868,33 @@ def _apply_orientation_with_geo(
 
 
 def _apply_photometric_miniswhite(arr: np.ndarray, ifd: IFD) -> np.ndarray:
-    """Apply TIFF MinIsWhite inversion for single-band grayscale images."""
+    """Apply TIFF MinIsWhite inversion for single-band grayscale images.
+
+    Signed-integer single-band MinIsWhite is rejected (issue #2278). The
+    reader used to pass these through unchanged, which round-tripped
+    inside xrspatial but produced files whose pixel values disagreed
+    with the on-disk Photometric tag against every other TIFF consumer
+    (GDAL, libtiff, ImageMagick).
+    """
     if ifd.photometric != 0 or ifd.samples_per_pixel != 1:
         return arr
     if arr.dtype.kind == 'u':
         return np.iinfo(arr.dtype).max - arr
     if arr.dtype.kind == 'f':
         return -arr
+    if arr.dtype.kind == 'i':
+        raise NotImplementedError(
+            f"Signed-integer MinIsWhite TIFFs are not supported "
+            f"(issue #2278): Photometric=0 (MinIsWhite), "
+            f"SampleFormat={ifd.sample_format} (signed int), "
+            f"BitsPerSample={ifd.bits_per_sample}, dtype={arr.dtype}. "
+            f"The reader has no semantically correct inversion for "
+            f"signed pixels here, and passing them through unchanged "
+            f"would disagree with the on-disk Photometric tag against "
+            f"every standards-compliant TIFF reader (GDAL, libtiff, "
+            f"etc.). Convert the file to MinIsBlack (Photometric=1) "
+            f"with another tool, or open it in an unsigned dtype."
+        )
     return arr
 
 
--- a/xrspatial/geotiff/_encode.py
+++ b/xrspatial/geotiff/_encode.py
@@ -131,8 +131,12 @@ def _apply_photometric_miniswhite_invert(
     See issue #1836.
 
     Returns the pre-inverted array (a new array) so that the reader's
-    inversion restores the original values. Multi-band data and signed
-    integer data pass through unchanged, matching the reader.
+    inversion restores the original values. Multi-band data passes
+    through unchanged. Signed-integer single-band MinIsWhite raises
+    ``NotImplementedError`` -- the previous passthrough produced files
+    whose pixel values disagreed with the on-disk Photometric tag
+    against every standards-compliant TIFF consumer (GDAL, libtiff).
+    See issue #2278.
     """
     if resolved_photometric != 0 or samples_per_pixel != 1:
         return arr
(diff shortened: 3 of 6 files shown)


brendancol (commented):
## PR Review: Reject signed-integer MinIsWhite TIFFs on read and write (#2278)

None.

- [ ] Pure-GPU decode path still silently passes signed-int MinIsWhite through (`xrspatial/geotiff/_backends/gpu.py:830-847`). That block runs when `ifd.photometric == 0 and samples == 1 and not arr_was_cpu_decoded`, and only handles `gpu_dtype.kind in ('u', 'f')`. Signed-int dtypes hit neither branch, so the buffer comes back un-inverted with Photometric=0 still in the IFD attrs -- the same silent-passthrough this PR fixes on the CPU side. Hard to hit in practice (no signed-int MinIsWhite fixtures in the golden corpus, and the GDS direct-decode path bails on `photometric == 0` at line 1058-1059), but it's a backend-parity gap: a CPU write followed by `open_geotiff(..., gpu=True)` would diverge from the CPU read after this PR lands. Either raise the same `NotImplementedError` here, or route signed-int MinIsWhite reads through the CPU fallback.

- [ ] `xrspatial/geotiff/tests/test_signed_miniswhite_rejected_2278.py:96`: `f'i{expected_bps}_msw_2278_{tmp_path.name}.tif'` interpolates `tmp_path.name`, but `tmp_path` is already per-test and per-invocation, so the extra suffix is redundant. `f'i{expected_bps}_msw_2278.tif'` works.
- [ ] `_patch_tag_value` in the new test file duplicates the IFD walk in `xrspatial/geotiff/tests/_tiff_surgery.py:patch_byte_counts`. One caller right now so the duplication is cheap, but if a third surgery test ever shows up the helper is worth promoting.

### What looks good
- Error message carries SampleFormat, BitsPerSample, Photometric, and dtype on both paths.
- Pre-write rejection runs before any bytes hit disk; there's a dedicated test pinning that no partial file is left behind.
- Read coverage spans all three CPU read paths (local file via `_reader._read_to_array`, HTTP COG eager via `_cog_http`, HTTP COG dask via `_backends/dask.py`). They all funnel through the same `_apply_photometric_miniswhite` helper, so changing one place covers all three.
- Non-regression coverage for the unsigned (uint8) and float (float32) round-trips is in the new test module, not just relying on the #1836 file.
- Public `to_geotiff` docstring updated with the limitation and the workaround.
- The forged-file read test asserts `SampleFormat=2` post-write before flipping Photometric, so if the writer's SampleFormat default ever shifts the test fails loud instead of forging a file that no longer reproduces the bug.

### Checklist
- [x] Algorithm matches the issue's chosen direction (Option A: reject)
- [x] All CPU read/write backends produce consistent behaviour (raise)
- [ ] Backend parity gap on the pure-GPU decode path (see Suggestions)
- [x] NaN handling unchanged; only a new dtype-branch rejection
- [x] Edge cases covered (int8/16/32/64 widths, type bounds via `info.min`/`info.max`)
- [x] Benchmark not needed (pure error path)
- [x] README feature matrix update not needed (no public function added)
- [x] Docstrings present and accurate on both helpers and the `to_geotiff` kwarg

brendancol (commented):
## PR Review (follow-up): Reject signed-integer MinIsWhite TIFFs on read and write (#2278)

Re-reviewed after the follow-up commit `0df24a90`.

### Blockers
None.

### Suggestions
The previous round's GPU parity gap is closed at `xrspatial/geotiff/_backends/gpu.py:848-866` with a matching `NotImplementedError` carrying the same SampleFormat / BitsPerSample / Photometric values. The new `test_read_signed_miniswhite_rejected_on_gpu_path_2278` exercises that path against a forged signed-int MinIsWhite TIFF and skips cleanly without cupy / CUDA. Nothing else to add.

### Nits
The previous round's helper-duplication nit (`_patch_tag_value` vs `_tiff_surgery.py:patch_byte_counts`) was deferred on purpose because there's still only one caller. Worth promoting if a third surgery test ever shows up.

### What looks good
- GPU rejection message has the same shape as the CPU one, so callers see a uniform diagnostic regardless of which backend `open_geotiff(...)` dispatched to.
- The GPU test reuses `_forge_signed_miniswhite_tif`, so the on-disk file layout is identical between the CPU and GPU read tests.
- Skip handling on the GPU test follows the established pattern in `test_golden_corpus_gpu_1930.py`: `pytest.importorskip("cupy")` plus a `getDeviceCount()` guard.

### Checklist
- [x] All read paths (eager local, eager HTTP COG, dask HTTP COG, GPU pure-decode) raise consistently
- [x] All write paths (eager) raise consistently; dask streaming continues to reject `miniswhite` via the existing #1836 guard
- [x] Tests cover all dtype widths (int8/16/32/64) and all rejection sites (write, CPU read, GPU read)
- [x] No regressions in unsigned or float MinIsWhite round-trips
- [x] Full `xrspatial/geotiff/tests/` run: 5080 passed, 68 skipped

PR #2283, opened 2026-05-22
Fix issue #2277: apply same EPSG-range fix to test fixture conftest

vanhci:
## Summary

The fix in PR #2280 addressed the broken EPSG range heuristic in 's . However, the same broken heuristic was present in  in the test :

This corrupts test GeoTIFFs with projected CRS codes that fall in the 4000-4999 window (e.g. EPSG 4087, 4088, 4499), which get incorrectly tagged as  instead of .

## Fix

Mirrors the approach from :

1. Added  frozenset (same codes as in )
2. Added  helper that:
   - Uses  when pyproj is available
   - Falls back to the vetted allowlist when pyproj is unavailable
3. Replaced the  check with 

## Why the original heuristic was wrong

The EPSG 4000-4999 range contains **both** geographic and projected CRSes:
- Geographic examples: 4019, 4047, 4230, 4258, 4267, 4269, 4283, 4322
- Projected examples: 4087, 4088, 4499 (these were being mis-tagged as geographic)

See issue #2277 for full context.

### DIFF
--- a/xrspatial/geotiff/tests/conftest.py
+++ b/xrspatial/geotiff/tests/conftest.py
@@ -10,6 +10,32 @@ import struct
 import numpy as np
 import pytest
 
+#: Vetted allowlist of EPSG codes known to be geographic (for test fixtures).
+#: Mirrors the fallback set in _geotags.py. Kept intentionally small.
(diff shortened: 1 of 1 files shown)


brendancol:
@vanhci can you run the `/review-pr #2283`  command?

PR #2290, opened 2026-05-22
CI: run geotiff golden_corpus tests against conda-forge rasterio (#2289)

### LINKED ISSUE #2289: CI: run geotiff golden_corpus tests with rasterio installed from conda-forge
## Reason or Problem
The current `test.yml` workflow installs xrspatial via pip and does not pull in `rasterio`. Every golden_corpus test under `xrspatial/geotiff/tests/` that starts with `pytest.importorskip('rasterio')` is silently skipped on CI. The oracle in `golden_corpus/_oracle.py` checks xrspatial output against what rasterio reads from the same file, so right now the parity assertions never actually run on a hosted runner.

Rasterio is awkward to install via pip on all three OS targets because the wheels depend on GDAL. Easier to run a separate CI job that gets rasterio (and its GDAL stack) from conda-forge and runs the geotiff golden_corpus tests there.

## Proposal
Add `.github/workflows/test-geotiff-corpus.yml`. Triggers match `test.yml`, but:
- Provision a conda env from conda-forge using `mamba-org/setup-micromamba` (or `conda-incubator/setup-miniconda`).
- Install `rasterio`, `gdal`, and the project's test extras into that env.
- Run `pytest xrspatial/geotiff/tests/ -m "not slow"` on PRs and the full set on push-to-main / nightly cron.

**Value:** the parity oracle stops being a no-op on CI. Drift from rasterio's reads gets caught on every PR instead of only when somebody happens to run the suite locally.

## Drawbacks
A second workflow adds a few minutes per PR. Conda-forge installs are slower than pip wheels; micromamba keeps the difference small.

## Alternatives
Add rasterio to the existing `test.yml` matrix. Skipped because GDAL under pip is fragile across the OS matrix, and one bad install would take down the whole job instead of an isolated one.

brendancol:
Closes #2289.

## Summary
- Adds `.github/workflows/test-geotiff-corpus.yml`. Sets up a conda env from conda-forge via `mamba-org/setup-micromamba`, installs `rasterio`, `gdal`, and `pyyaml` alongside the project's `tests` extras, then runs `pytest xrspatial/geotiff/tests/`.
- Triggers mirror `test.yml`: PR, push-to-main, nightly cron at 03:30 UTC (offset 30 min from `test.yml` to avoid runner contention), and `workflow_dispatch`.
- PR runs stay on the ubuntu-only fast lane (`-m "not slow"`) to keep CI cost down. Push-to-main and the nightly widen to ubuntu / macos / windows and run the full test set.
- `fail-fast: false` on the matrix so a flake on one OS does not abort the others.

## Why
The parity oracle in `xrspatial/geotiff/tests/golden_corpus/_oracle.py` is silently skipped on CI today because `test.yml` does not install rasterio (pip + GDAL is fragile across the OS matrix). This workflow makes the oracle actually run.

## Test plan
- [ ] Workflow YAML parses (checked locally with `yaml.safe_load`).
- [ ] On this PR, the new `pytest-geotiff-corpus` job shows up, micromamba env solves, rasterio + GDAL versions print, and the fast-lane geotiff tests pass.
- [ ] The existing `pytest` workflow keeps passing.


brendancol (commented):
## PR Review: CI: run geotiff golden_corpus tests against conda-forge rasterio (#2289)

None.

- [ ] Scope mismatch with the issue. #2289 talks about "golden_corpus tests", but the workflow runs all of `xrspatial/geotiff/tests/` at `.github/workflows/test-geotiff-corpus.yml:72` and `:75`. Most of the non-corpus tests in that directory also `importorskip('rasterio')`, so they will benefit too -- but the job picks up heavier integration tests that have nothing to do with the parity oracle. If that is on purpose, rename the workflow and PR to talk about the whole geotiff suite. If not, narrow the pytest target to `xrspatial/geotiff/tests/golden_corpus/`.

- [ ] `pip install -e .[tests]` on top of conda-forge `rasterio` is a known sharp edge (`.github/workflows/test-geotiff-corpus.yml:60-63`). The `tests` extra pulls `geopandas`, `pyproj`, `dask`, `scipy`, `pyarrow` from PyPI wheels. PyPI's pyproj ships its own bundled proj that can shadow the conda-forge proj GDAL links against; numpy can get re-resolved the same way. Two safer options: (a) install `geopandas` / `pyproj` / `dask` / `scipy` / `pyarrow` via `create-args` (conda-forge) and run `pip install -e . --no-deps`, or (b) `pip install -e . --no-deps` followed by a separate pip install of the dev-only extras (`flake8`, `pytest`, `hypothesis`, `noise`). Failing that, add a step that re-imports `rasterio` right after the pip install so a broken env fails fast instead of mid-pytest.

- [ ] The `env: OS / PYTHON` block at `.github/workflows/test-geotiff-corpus.yml:36-38` is unused. Inherited from `test.yml`, harmless. Drop or wire to a step that needs it.
- [ ] The `Show rasterio / GDAL versions` step could also print numpy and Python versions in the same line, so the version stamp is one log entry to grep for rather than two.
- [ ] `pytest-geotiff-corpus` is a fine workflow name, but if the goal is the parity oracle, `pytest-geotiff-rasterio-oracle` says it more plainly. Cosmetic.

### What looks good
- Cron offset (`30 3 * * *`) avoids contention with `test.yml`'s 03:00 nightly. Lines 19-22 explain why.
- `fail-fast: false` is right for a cross-OS install matrix where one runner can flake without invalidating the others.
- `shell: bash -el {0}` as a default keeps the conda env activated for every step, which is easy to forget on Windows.
- `condarc` pins channel priority to strict conda-forge, which avoids accidentally mixing with defaults.
- Matrix narrowing on PRs (ubuntu-only, python 3.14) keeps PR CI cost bounded; nightly widens to the full OS / python matrix.
- I checked: no test under `xrspatial/geotiff/tests/` imports cupy at module top level without a guard, so the new job will not blow up trying to import cupy on a CPU runner; those tests skip cleanly.

### Checklist
- [x] Algorithm matches reference/paper -- N/A, infra-only.
- [x] All implemented backends produce consistent results -- N/A.
- [x] NaN handling is correct -- N/A.
- [x] Edge cases covered -- the workflow itself is the test surface.
- [x] Dask chunk boundaries handled correctly -- N/A.
- [x] README feature matrix updated -- not applicable.
- [x] Docstrings present -- workflow has explanatory comments.

brendancol (commented):
## Follow-up review (after 3451017f)

Re-checked the diff against the previous review.

- Suggestion: scope drift -- **fixed**. `pytest` target narrowed to `xrspatial/geotiff/tests/golden_corpus/` at lines 75 / 78.
- Suggestion: `pip install -e .[tests]` shadowing the conda env -- **partially fixed**. Added the "Verify rasterio still imports after pip step" guard at lines 61-67. The fuller restructure (install all binary deps via `create-args` and use `pip --no-deps`) is deferred; the guard catches the failure mode loudly enough for now. If CI starts complaining about pyproj/proj mismatches in practice, the bigger restructure becomes worth doing.
- Nit: unused `env:` block -- **fixed** (removed).
- Nit: combined version stamp -- **fixed**. The verify step now prints python, rasterio, gdal, and numpy on one go.
- Nit: workflow rename -- **dismissed**. "corpus" is the term the issue and tests already use; renaming would invite drift.

### New findings
None. Diff is small and targeted.

Issue #2286, opened 2026-05-22
Promote COG support to ready/stable with compliance and parity gates


brendancol:
## Goal

Move COG support from "use with caution" to a clearly documented ready/stable state, starting with a narrow production contract and expanding only when the matching test gates exist.

COG should be split into separate readiness tracks because the risks are different:

- COG writer: `to_geotiff(..., cog=True)`
- COG reader: local COG reads
- HTTP/range COG reader: remote COG reads, windowed reads, and dask graph reads

## Proposed stable scope for the first promotion

Promote only this subset first:

- Axis-aligned 2D/3D rasters.
- CPU writer and CPU reader paths.
- Stable codecs only: `none`, `deflate`, `lzw`, `zstd`, and `packbits` if current interop coverage supports it.
- Internal overviews only.
- Normal CRS, transform, dtype, nodata, band, and pixel-is-area / pixel-is-point behavior.

Keep these out of the first stable claim:

- GPU COG read/write.
- Experimental codecs: `lerc`, `jpeg2000` / `j2k`, `lz4`.
- Internal-only JPEG.
- Rotated transforms.
- External `.tif.ovr` sidecars.
- File-like destinations with `cog=True`.
- Any BigTIFF COG behavior not covered by external validation.

### 1. Define the COG readiness contract in code and docs

- Split broad feature tiering into more precise entries, e.g. `writer.cog`, `reader.local_cog`, `reader.http_cog`.
- Keep HTTP/range COG separate from local COG because it has additional transport, range, coalescing, and security behavior.
- Update `SUPPORTED_FEATURES` only after test gates are in place.

Relevant files:

- `xrspatial/geotiff/_attrs.py`
- `docs/source/reference/geotiff.rst`
- `examples/user_guide/52_COG_Overview_Generation.ipynb`

### 2. Add an external COG compliance gate for the writer

For COGs written by `to_geotiff(..., cog=True)`, assert external interoperability, not just xrspatial self-read.

Test matrix should cover:

- Stable codecs.
- Integer and float dtypes.
- Single-band and multiband rasters.
- Nodata sentinel and NaN nodata cases.
- Pixel-is-area and pixel-is-point georeferencing.
- Explicit and auto-generated overview levels.
- Overview resampling modes currently intended to be stable.

Assertions should include:

- Rasterio can open the file.
- Base pixels match expected data.
- Overview count and dimensions match expected levels.
- Overview pixels match expected values or documented tolerances.
[...]

PR #2295, opened 2026-05-22
Split SUPPORTED_FEATURES into writer.cog / reader.local_cog / reader.http_cog (#2291)

### LINKED ISSUE #2291: COG readiness: split SUPPORTED_FEATURES into writer.cog / reader.local_cog / reader.http_cog (#2286 PR 1/6)

## Scope

Foundational, no behavior change. Split the COG entry in `SUPPORTED_FEATURES` so the writer, local reader, and HTTP reader can be promoted on independent tracks.

## Required changes

- `xrspatial/geotiff/_attrs.py`:
  - Keep `writer.cog` at `advanced`.
  - Add `reader.local_cog`: `advanced`.
  - Add `reader.http_cog`: `advanced`.
- Add an inventory test that asserts all three keys exist in `SUPPORTED_FEATURES` (extend an existing inventory test or add a small new test file).
- Mention the split in the `SUPPORTED_FEATURES` docstring/comment block so the rationale is visible to future readers.

## Out of scope

- Do not change any tier value other than adding the two new entries.
- Do not touch docs reference page or the COG notebook (PR 5/6 handle docs).
- Do not promote any feature to `stable`.

## Acceptance

- `SUPPORTED_FEATURES['writer.cog']`, `SUPPORTED_FEATURES['reader.local_cog']`, `SUPPORTED_FEATURES['reader.http_cog']` all resolve.
- Full geotiff test suite green.
- No production code path branches on the new keys yet.

brendancol:
Closes #2291. Part of the #2286 COG readiness rollout.

## What changed

- `xrspatial/geotiff/_attrs.py`: adds `reader.local_cog` and `reader.http_cog` to `SUPPORTED_FEATURES` at the `advanced` tier. `writer.cog` keeps its current tier. The comment block above the dict picks up a short note on why the three keys are separate.
- `xrspatial/geotiff/tests/test_supported_features_tiers_2137.py`: small inventory test that asserts the three keys exist and carry a known tier label.

## What did not change

- No tier was promoted. No production code path branches on the new keys.
- Docs reference page and the COG notebook are untouched; those updates ride on later PRs in the #2286 chain.

## Why the split

The three COG code paths have separate stability profiles. The writer emits a self-contained COG layout, the local reader walks overview IFDs on disk, and the HTTP reader sits on top of range-request fetching and partial-IFD handling. Folding them under one key would tie any promotion of one path to a promotion of all three.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/` passes locally (5118 passed, 68 skipped).
- [x] The new inventory assertion fires if any of the three keys is dropped.

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -228,8 +228,10 @@ SUPPORTED_FEATURES = {
     'reader.sidecar_ovr': 'advanced',
     'reader.allow_rotated': 'advanced',
     'reader.allow_unparseable_crs': 'advanced',
-    # COG reader paths (issue #2291): split from the single writer.cog
-    # entry so the local and HTTP variants can promote independently.
+    # COG reader paths (issue #2291): split out from the previous
+    # single COG concept (which only carried ``writer.cog``) so the
+    # local and HTTP reader variants can promote independently of the
+    # writer and of each other.
     'reader.local_cog': 'advanced',
     'reader.http_cog': 'advanced',
     'reader.gpu': 'experimental',
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -195,6 +195,16 @@ _VALID_COMPRESSIONS = (
 # See ``xrspatial/geotiff/__init__.py`` for the per-tier semantics; the
 # inline comments here track the codec/reader/writer split used by the
 # user-guide notebook table.
+#
+# COG entries are split into three keys (issue #2291, part of the #2286
+# rollout) so the writer, local reader, and HTTP reader can move between
+# tiers on their own track. The three code paths have different stability
+# profiles: the writer emits a self-contained COG layout, the local reader
+# parses overview IFDs out of an on-disk file, and the HTTP reader adds
+# range-request fetching and partial-IFD handling on top of that. Folding
+# them into a single ``cog`` key would tie a promotion of any one path to
+# a promotion of all three. The keys carry no behaviour today; production
+# code still gates on the underlying writer / reader options.
 SUPPORTED_FEATURES = {
     # Codecs. Tier 1 lossless integer + float byte-for-byte round-trip.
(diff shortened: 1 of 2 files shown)


brendancol:
## PR Review

Inventory-only change, scoped tightly to what #2291 asks for. Two new reader keys, one updated block comment, one new test. Nothing else touched.

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/geotiff/_attrs.py` line 231: the inline comment reads "split from the single `writer.cog` entry," which makes it sound like the writer key was renamed. It was kept as-is and two reader keys were added next to it. Rewording to "split out from the previous single COG concept" would be less misleading for someone skimming the dict later. Not a blocker.

### What looks good
- Purely additive. No tier promotion, no production code path branches on the new keys, no docs touched (per the issue, docs ride on later PRs in the #2286 chain).
- The block comment above `SUPPORTED_FEATURES` records why the three keys are separate, so the rationale is visible without needing git blame.
- The new test pins both presence and a valid tier label for each of the three keys, so a future refactor that drops one of them fails fast.
- A grep over `xrspatial/` and `docs/` for the new key names found no other references, confirming the change cannot collide with a stale call site.
- The existing `test_supported_features_is_a_mapping` walks every entry, so the new keys also pick up the generic shape assertion for free.

### Checklist
Almost everything in the standard checklist is n/a for an inventory-only change. The two items that do apply both pass:
- [x] Edge cases covered by tests: presence + tier label asserted for each of the three keys.
- [x] Docstrings / comments accurate: the block comment now explains the split.

brendancol:
## Re-review (round 2)

Picked up the one nit from the first pass. The inline comment at `xrspatial/geotiff/_attrs.py` now reads "split out from the previous single COG concept (which only carried `writer.cog`)" instead of "split from the single `writer.cog` entry," so a future reader skimming the dict will not get the impression that the writer key was renamed.

Diff is otherwise unchanged from round 1. No new findings. Marking the review clean from my side.

PR #2296, opened 2026-05-22
Add COG parity rows to release gate (#2294)

### LINKED ISSUE #2294: COG readiness: add COG rows to backend parity/release gate (#2286 PR 4/6)

## Scope

Add a focused COG parity layer to the release gate so COG read/write paths are required-green for promotion.

## Required changes

New file: `xrspatial/geotiff/tests/test_cog_parity_2286.py` (or an additive section to `test_backend_full_parity_2211.py`; prefer a new file to keep diffs reviewable).

Required rows:

1. `xrspatial write COG -> xrspatial eager read`
2. `xrspatial write COG -> xrspatial dask read`
3. `xrspatial write COG -> rasterio read`
4. `golden/rasterio COG fixture -> xrspatial local read`
5. `golden/rasterio COG fixture -> xrspatial HTTP range read`
6. `golden/rasterio COG fixture -> xrspatial dask HTTP range read`

For each row:

- Compare expected pixel values within documented tolerances (lossless = byte-exact).
- Assert CRS, transform, nodata, dtype, band count, dim names survive.
- Skip reasons must be explicit and `pytest.skip` with a string mentioning the dependency or linked issue. No silent skips.

Use the existing golden corpus under `xrspatial/geotiff/tests/golden_corpus/` where it applies, or generate a small COG fixture via rasterio in the test if no suitable golden exists.

## Out of scope

- GPU rows (`reader.gpu` stays experimental).
- Experimental codecs.
- Promoting any feature tier (PRs 5/6 do that).

## Acceptance

- New file under `xrspatial/geotiff/tests/`.
- All six rows present, each either passing or skipping with a linked reason.
- Suite green on CPU CI.

brendancol:
Closes #2294. Part of #2286 (COG readiness/stability rollout, PR 4/6).

## Summary

New file `xrspatial/geotiff/tests/test_cog_parity_2286.py` adds the six COG parity rows the release gate needs. Tests-only change; no production code or other test file is modified.

## Rows

1. xrspatial write COG -> xrspatial eager read
2. xrspatial write COG -> xrspatial dask read
3. xrspatial write COG -> rasterio read
4. golden/rasterio COG fixture -> xrspatial local read
5. golden/rasterio COG fixture -> xrspatial HTTP range read
6. golden/rasterio COG fixture -> xrspatial dask HTTP range read

Each row compares pixels byte-exact (every fixture uses lossless deflate) and asserts CRS, transform, nodata, dtype, band count, and dim names survive the round trip.

The HTTP rows reuse the in-process range-aware HTTP server pattern from `test_cog_http_parallel_decode_2026_05_15.py` and `test_cog_http_concurrent.py`. The golden COG fixture is the existing `cog_internal_overview_uint16` entry from the golden corpus.

Skips are explicit. Missing `dask`, `rasterio`, or the golden corpus fixture each surface a `pytest.skip` with a string naming the dependency. No silent skips.

## Test plan

- [x] All six rows pass locally (`pytest xrspatial/geotiff/tests/test_cog_parity_2286.py -v` -> 6 passed)
- [ ] CI green on CPU runners

### DIFF
--- a/xrspatial/geotiff/tests/test_cog_parity_2286.py
+++ b/xrspatial/geotiff/tests/test_cog_parity_2286.py
@@ -17,8 +17,12 @@ Rows
 6. ``golden/rasterio COG fixture -> xrspatial dask HTTP range read``
 
 Each row asserts byte-exact pixels (every fixture used here is
-lossless) and the metadata contract that the release gate must lock
-in: CRS, transform, nodata, dtype, band count, and dim names.
+lossless) and a fixed subset of the metadata contract: ``crs`` (or
+``crs_wkt``), ``transform``, ``nodata`` (including the no-nodata
+case), pixel ``dtype``, band count, and the ``(y, x)`` dim names.
+The wider canonical-attrs surface (resolution, georef_status, etc.)
+lives in ``test_backend_full_parity_2211.py``; this file is the
+narrower COG-only gate.
 
 Skip policy
 -----------
@@ -45,6 +49,7 @@ import importlib.util
 import pathlib
 import socketserver
 import threading
+import uuid
 
 import numpy as np
 import pytest
@@ -121,12 +126,30 @@ class _RangeHandler(http.server.BaseHTTPRequestHandler):
 
 
 def _serve_payload(payload: bytes, monkeypatch):
-    """Spin a range-aware server bound to localhost; return (httpd, port)."""
+    """Spin a range-aware server bound to localhost; return (httpd, port).
+
+    The handler subclass is named with a uuid suffix so that the two
+    fixtures in this module (and any future ones) don't share a
+    qualname. Without the suffix, tracebacks reuse the same class
+    identifier across fixture invocations and become harder to read.
+
+    ``allow_reuse_address = True`` lets the OS reclaim the port
+    quickly when the test tears down (avoiding TIME_WAIT-related
+    binding races under parallel pytest runs). ``timeout=5`` on the
+    server caps how long a stuck request can pin the daemon thread.
+    """
     monkeypatch.setenv("XRSPATIAL_GEOTIFF_ALLOW_PRIVATE_HOSTS", "1")
     handler_cls = type(
-        "RangeHandler2286", (_RangeHandler,), {"payload": payload}
+        f"RangeHandler2286_{uuid.uuid4().hex[:8]}",
+        (_RangeHandler,),
+        {"payload": payload},
     )
-    httpd = socketserver.TCPServer(("127.0.0.1", 0), handler_cls)
+
+    class _ReusableTCPServer(socketserver.TCPServer):
+        allow_reuse_address = True
+        timeout = 5
+
+    httpd = _ReusableTCPServer(("127.0.0.1", 0), handler_cls)
     port = httpd.server_address[1]
     thread = threading.Thread(target=httpd.serve_forever, daemon=True)
     thread.start()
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: Add COG parity rows to release gate (#2294)

Tests-only addition of `xrspatial/geotiff/tests/test_cog_parity_2286.py`. The file lays out the six rows the issue calls for, each with a loud skip path. Follows the patterns from `test_backend_full_parity_2211.py` and `test_cog_http_parallel_decode_2026_05_15.py`. A few real issues below.

None.

- `test_cog_parity_2286.py` row 4 (`test_row4_golden_cog_xrspatial_local`) does not compare pixel values, only shape/dtype/CRS/transform presence. The issue text explicitly asks for "Compare expected pixel values within documented tolerances" on every row. Row 4 should either compare against a recorded reference (e.g. read the same fixture with rasterio and compare byte-exact) or at minimum sanity-check the pixel range/sum is non-trivial. Right now a regression that returned all-zero pixels would pass row 4 silently.

- `_assert_nodata_equals` accepts `expected=None` by silently passing through. The fixture sets `expected_attrs["nodata"] = None`, so this helper is effectively a no-op for rows 1 and 2. Either drop the call and assert directly, or extend the helper to assert that the read side also has no nodata attr (catching a regression where the reader fabricates a sentinel).

- `_RangeHandler` / `_serve_payload` have no timeouts on the TCPServer. If a test hangs after fetching tiles, the daemon thread keeps the socket open until process exit. Setting `httpd.timeout = 5` and `httpd.allow_reuse_address = True` matches the other COG HTTP tests in the tree.

- The handler class name `RangeHandler2286` is reused across multiple fixture invocations in the same process. `type()` returns a fresh class each call so it works, but duplicate qualnames make tracebacks harder to read. Append the port or a uuid suffix.

- Module docstring mentions "the metadata contract that the release gate must lock in". The actual contract checked is a subset (CRS / transform / nodata / dtype / band / dims). Naming the subset in the docstring would make a future reader's job easier.

- `_assert_band_count` and `_assert_dim_names` hard-code single-band 2D shape. That matches every fixture used here, but if a future row adds a multi-band fixture the helper names will read misleadingly. A comment noting the 2D-only scope would help.

### What looks good

- Loud skip strings on every dependency-gated path (dask, rasterio, golden corpus). No silent skips.
- Row 2 and row 6 both verify `hasattr(da.data, "dask")`. A regression that drops `chunks=` and falls back to eager would not silently pass.
- HTTP server pattern matches existing tree style. `monkeypatch.setenv("XRSPATIAL_GEOTIFF_ALLOW_PRIVATE_HOSTS", "1")` is the correct knob.
- Byte-exact comparison is justified by the lossless deflate choice. The fixture uses a +1 offset so the uint16 byte-exact path is not destroyed by the masked-nodata float64 upcast.
- `_materialise` correctly orders dask compute before cupy get (matches the corpus oracle).

### Checklist

- [x] Algorithm matches reference (tests-only PR, contract follows issue #2294)
- [x] NaN handling is correct (+1 offset avoids masked-nodata upcast)
- [x] Edge cases are covered by tests (skip paths exercised)
- [x] Dask chunk boundaries handled correctly (chunks=16 on 64x64)
- [x] Benchmark not applicable (tests-only PR)
- [ ] README feature matrix update not applicable (no new public API)

brendancol (commented):
## Follow-up review (after 5bc031ce)

All three suggestions and all three nits from the previous review are addressed:

- Row 4 now compares pixels byte-exact against a rasterio read of the same golden COG bytes. A regression that returned correct shape but wrong values is caught.
- `_assert_nodata_equals` now asserts the read side is also nodata-free when the writer did not stamp one, so a fabricated sentinel from the reader fails the test.
- `_serve_payload` uses `allow_reuse_address=True`, `timeout=5`, and a uuid-suffixed handler class. The daemon thread can't pin a stuck socket forever, and tracebacks stay readable across fixture invocations.
- Module docstring now names the metadata subset the gate locks in.
- 2D-only helpers carry a scope comment for future readers.

Local run: `pytest xrspatial/geotiff/tests/test_cog_parity_2286.py -v` reports 6 passed.

No remaining items. Holding for CI.

PR #2297, opened 2026-05-22
Add external COG writer compliance test suite (#2292)

### LINKED ISSUE #2292: COG readiness: external compliance test suite for to_geotiff(cog=True) (#2286 PR 2/6)

## Scope

Add an external-interop compliance suite that exercises COGs written by `to_geotiff(..., cog=True)` and asserts they are readable and well-formed under rasterio. No production code changes — tests only.

## Required changes

New file: `xrspatial/geotiff/tests/test_cog_writer_compliance.py`.

Test matrix:

- Stable codecs only: `none`, `deflate`, `lzw`, `zstd`, `packbits` (skip if not currently interop-validated).
- Dtypes: at least one integer dtype (e.g. `uint16`) and one float dtype (e.g. `float32`).
- Band counts: single-band and multiband (3-band).
- Nodata: sentinel value and NaN cases.
- Georef: pixel-is-area and pixel-is-point.
- Overviews: explicit level list and auto-generated levels.
- Overview resampling modes that are intended to be stable.

For each row, assertions:

- `rasterio.open` succeeds.
- Base pixels equal expected (allclose with documented tolerance for lossy paths, byte-exact for lossless).
- Overview count and dims match.
- Overview pixels match expected within tolerance.
- CRS, transform, nodata, dtype, band count, and dim names survive the round-trip.
- TIFF layout is COG-like: tiled (not stripped), overview IFDs present, sane tile offsets, IFDs placed before image data.
- Optional rio-cogeo / GDAL `validate_cloud_optimized_geotiff` check that `pytest.skip`s cleanly when the dependency is missing.

## Out of scope

- Do not modify writer code. Failures should produce actionable test failures, not silently get patched over.
- GPU writer paths.
- Experimental codecs.
- BigTIFF beyond what existing tests already cover.

## Acceptance

- New test file added under `xrspatial/geotiff/tests/`.
- Suite green on CPU CI.
- Any genuinely-broken row gets a targeted xfail with a linked issue rather than a silent skip.

brendancol:
Closes #2292. Part of #2286 (COG readiness/stability rollout, sub-PR 2/6).

## Summary

Adds `xrspatial/geotiff/tests/test_cog_writer_compliance.py`, an
external-interop suite that opens `to_geotiff(..., cog=True)` output
through rasterio and walks the writer across a stable-codec matrix.
No production code changed — this is a tests-only PR.

The matrix covers:

- Codecs: `none`, `deflate`, `lzw`, `zstd`, `packbits`.
- Dtypes: `uint16` and `float32`.
- Band counts: single-band and 3-band.
- Nodata: integer sentinel, float sentinel, and NaN.
- Georef: PixelIsArea and PixelIsPoint.
- Overviews: explicit `[2, 4, 8]` list and auto-generated pyramids.

Each row asserts base pixels are byte-exact at full resolution (stable
codecs are lossless), overview factors and per-level dims match, the
file is tiled, IFDs sit before image data, and CRS / transform /
nodata / dtype / band count all survive the round-trip.

The suite also runs the optional rio-cogeo `cog_validate` /
GDAL `validate_cloud_optimized_geotiff` check, and `pytest.skip`s
cleanly when neither package is installed.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_cog_writer_compliance.py` — 28 passed, 1 skipped locally
- [x] Optional validator path skips cleanly when rio-cogeo / osgeo_utils are missing
- [x] No production code modified; only the new test file lands

### DIFF
--- a/xrspatial/geotiff/tests/test_cog_writer_compliance.py
+++ b/xrspatial/geotiff/tests/test_cog_writer_compliance.py
@@ -117,7 +117,13 @@ def _build_da(
 
 
 def _pick_sentinel(dtype: np.dtype) -> float | int:
-    """Pick a nodata sentinel that fits the dtype."""
+    """Pick a nodata sentinel that fits the dtype.
+
+    The signed-int branch is unreachable from the current DTYPES list
+    (only ``uint16`` and ``float32``) but is kept for the eventual case
+    where the matrix grows. Dead branches in a helper are cheap and the
+    intent is clearer than special-casing the current matrix here.
+    """
     dt = np.dtype(dtype)
     if dt.kind == "f":
         return -9999.0
@@ -192,7 +198,7 @@ def _try_cog_validate(path: str) -> None:
         )
         return
 
-    warnings, errors, _details = validate_cloud_optimized_geotiff.validate(
+    _warns, errors, _details = validate_cloud_optimized_geotiff.validate(
         path, full_check=True,
     )
     assert not errors, f"GDAL validator errors: {errors}"
@@ -207,7 +213,16 @@ def _try_cog_validate(path: str) -> None:
 @pytest.mark.parametrize("dtype", DTYPES)
 @pytest.mark.parametrize("codec", STABLE_CODECS)
 def test_codec_dtype_bands_roundtrip(tmp_path, codec, dtype, bands):
-    """Stable codec round-trip via rasterio: base pixels byte-exact, georef survives."""
+    """Stable codec round-trip via rasterio: base pixels byte-exact, georef survives.
+
+    Contracts asserted per row:
+    - rasterio.open succeeds and reports a tiled COG.
+    - Band count and dtype survive.
+    - Base pixels are byte-exact (stable codecs are lossless).
+    - Overview decimation factors survive.
+    - CRS and transform survive.
+    - IFDs sit before any tile data block (COG layout).
+    """
     arr = _make_data(dtype, bands=bands, height=64, width=64)
     da = _build_da(arr, raster_type="area", crs=4326)
 
@@ -249,7 +264,7 @@ def test_codec_dtype_bands_roundtrip(tmp_path, codec, dtype, bands):
         assert src.crs is not None and src.crs.to_epsg() == 4326, (
             f"CRS round-trip failed: got {src.crs}"
         )
-        assert src.transform.is_identity is False, (
+        assert not src.transform.is_identity, (
             "transform should not be identity for a georeferenced raster"
         )
     # COG layout invariant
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## Review: external COG writer compliance suite (#2292)

Tests-only PR. The backend dispatch / Dask / CuPy / performance sections of the standard review do not apply here. Focused review on test quality, coverage of the stated matrix, and a few minor cleanups.

### Blockers
None.

### Suggestions

- **Missing overview pixel-value assertion.** The issue asks for "Overview pixels match expected within tolerance", but the suite only checks overview dims and decimation factors. `test_overviews_explicit_levels` reads `OVERVIEW_LEVEL=lvl` and asserts the shape; it never compares the overview pixels to a reference. A `mean` overview is deterministic from the base array, so we can compute the expected 2x2 block-mean ourselves and assert allclose. Worth adding at least one row that pins overview values.
- **No "intended to be stable" resampling-mode coverage.** Issue mentions "Overview resampling modes that are intended to be stable" as part of the matrix. The existing assertions all rely on the default `mean`. One small param test over `['mean', 'nearest']` would close that gap.
- **`test_nodata_sentinel_survives` for float32 only round-trips one sentinel value (-9999.0).** That is fine for the spec, but a positive sentinel (e.g. `np.finfo(np.float32).max`) would exercise the path where the sentinel collides with valid data range. Optional.

### Nits

- `xrspatial/geotiff/tests/test_cog_writer_compliance.py:252`: `assert src.transform.is_identity is False` — `is False` on a property is brittle (works today, breaks if the attr ever returns a truthy non-bool). `assert not src.transform.is_identity` is cleaner.
- `xrspatial/geotiff/tests/test_cog_writer_compliance.py:195`: local variable `warnings` shadows the stdlib `warnings` module. Rename to `_warns` to mirror the rio-cogeo branch above it.
- `xrspatial/geotiff/tests/test_cog_writer_compliance.py:126`: `_pick_sentinel` has a signed-int branch (`iinfo(dt).min`) that is never reached -- the DTYPES list only has `uint16` and `float32`. Either drop the dead branch or document it as a future-proof.
- `xrspatial/geotiff/tests/test_cog_writer_compliance.py:396`: the power-of-two check `(o & (o - 1)) == 0` is correct given the `o >= 2` guard but the intent is not obvious at a glance. A short comment ("bitwise: power of two iff o & (o-1) == 0") would help future readers.
- The big parametrized matrix is named `test_codec_dtype_bands_roundtrip`. The two assertions that fire after the read (`src.transform.is_identity`, `src.crs.to_epsg() == 4326`) cover CRS/transform survival. Worth a one-line comment at the top of the test naming each contract it covers, so a future reader does not need to scan the body.

### What looks good

- Black-box via rasterio is the right framing -- the in-process round-trip tests already exist elsewhere, and this suite catches the cross-reader interop class of bugs the parent issue cares about.
- Optional rio-cogeo / GDAL validator path with a clean `pytest.skip` when neither is installed, exactly as the issue spec asks for.
- COG layout assertion (`_assert_ifds_before_data`) is reused across the explicit / auto / layout tests rather than duplicated.
- Temporary filenames carry the `2292_` issue prefix per project conventions for parallel-safe paths.
- 28 passed locally, 1 skipped (the validator path). No `xfail` rows needed -- nothing in the matrix surfaced a writer bug.

### Checklist

- [x] Matrix matches the issue scope (codecs, dtypes, bands, nodata, georef, overviews)
- [x] Optional validators handled with `pytest.skip`
- [x] No production code modified
- [x] Temp paths use a unique `2292_` prefix
- [ ] Overview pixel values (not just dims) compared against expected -- suggestion above
- [ ] At least one overview-resampling mode beyond the `mean` default exercised -- suggestion above

brendancol (commented):
## Follow-up review (#2292)

All items from the previous review are addressed:

- Overview pixel-value comparison: new ``test_overview_pixels_match_expected`` runs both ``mean`` and ``nearest`` against a hand-computed 2x decimation, with a 1 ULP tolerance documented in the comment.
- ``test_codec_dtype_bands_roundtrip`` now opens with a docstring listing the contracts each row asserts.
- ``test_overviews_auto_generated`` carries an inline note explaining the bitwise power-of-two check.
- ``is False`` on ``src.transform.is_identity`` replaced with ``not``.
- ``warnings`` local in ``_try_cog_validate`` renamed to ``_warns``.
- ``_pick_sentinel`` documents the unreachable signed-int branch.

Local run: 30 passed, 1 skipped (the external validator step skips when neither rio-cogeo nor osgeo_utils is installed).

No new blockers, suggestions, or nits. PR is ready from a review standpoint pending CI and the merge-with-main check.

Issue #2300, opened 2026-05-22
Promote local COG contract to stable: flip tiers, document the contract, release notes (#2286 prod-ready wave A)


brendancol:
Follow-up to #2286 wave 1 (#2291, #2292, #2293, #2294 — all merged).

With the writer compliance suite, parity gate, byte-budget contract, and tier split in place, the gating evidence to flip the local COG paths to `stable` exists. This issue captures that flip and the documentation that goes with it.

### `xrspatial/geotiff/_attrs.py`

- `writer.cog`: `advanced` → `stable`
- `reader.local_cog`: `advanced` → `stable`
- `reader.http_cog`: stays `advanced` (HTTP COG has separate transport/security/coalescing concerns; the byte-budget contract in #2298 pins behavior but does not yet make a promotion case)
- Update the comment block above `SUPPORTED_FEATURES` to describe what the stable COG contract guarantees and what stays advanced.

### `docs/source/reference/geotiff.rst`

Add a section titled something like "Stable COG contract" with these bullets verbatim (or close), pulled from #2286:

- Axis-aligned 2D/3D rasters.
- CPU writer and CPU reader paths.
- Stable codecs only: `none`, `deflate`, `lzw`, `zstd`, `packbits`.
- Internal overviews only.
- Normal CRS, transform, dtype, nodata, band, pixel-is-area / pixel-is-point behavior.

And a "Outside the stable contract" section listing:

- GPU COG read/write.
- Experimental codecs (`lerc`, `jpeg2000`, `j2k`, `lz4`).
- Internal-only JPEG.
- Rotated transforms.
- External `.tif.ovr` sidecars.
- File-like destinations with `cog=True`.
- BigTIFF COG (tracked separately).
- HTTP/range COG (tracked separately; see #2298 byte-budget contract).

### `examples/user_guide/52_COG_Overview_Generation.ipynb`

Update any prose that says "use with caution" / "experimental" for COG writer paths to reflect the stable contract. Keep the cautious phrasing for HTTP COG and the out-of-scope combos.

### Release notes

Add an entry in the release notes file describing the COG stability contract — what is now stable, what stays advanced, and the gating tests that back the promotion (#2292, #2293, #2294 compliance + parity + byte-budget).

## Out of scope

- BigTIFF COG (tracked separately).
- HTTP COG promotion (tracked separately).
- Error-message UX (tracked separately).
- Touching `_writer.py` or any other production code path.

## Acceptance

[...]

PR #2304, opened 2026-05-22
Make rio-cogeo COG validator a required CI gate (#2302)

### LINKED ISSUE #2302: rio-cogeo / GDAL validate_cloud_optimized_geotiff as required CI gate (#2286 prod-ready wave C)
Follow-up to #2286 / #2292.

The compliance suite in `xrspatial/geotiff/tests/test_cog_writer_compliance.py` runs an optional `rio-cogeo` / GDAL `validate_cloud_optimized_geotiff` check that `pytest.skip`s cleanly when neither package is installed. An optional interop validator that silently skips is not a production guarantee. This issue stands up a CI job that installs `rio-cogeo` + GDAL and runs that validation strictly.

### CI workflow

In `.github/workflows/` (whichever file runs pytest), add a new job (or extend an existing one) that:

- Installs `rio-cogeo` and the GDAL Python bindings on Linux.
- Sets an env var like `XRSPATIAL_REQUIRE_COG_VALIDATOR=1`.
- Runs the COG compliance subset: `pytest xrspatial/geotiff/tests/test_cog_writer_compliance.py xrspatial/geotiff/tests/test_cog_parity_2286.py -x`.

Keep the existing matrix jobs as-is — this is an additional job, not a replacement.

### Test file

`xrspatial/geotiff/tests/test_cog_writer_compliance.py`:

- The current optional validator block should:
  - Skip (current behavior) when `XRSPATIAL_REQUIRE_COG_VALIDATOR` is not set and rio-cogeo / GDAL is missing.
  - **Fail** (not skip) when `XRSPATIAL_REQUIRE_COG_VALIDATOR=1` is set and the dependency is missing — so a misconfigured CI job is loud, not silent.
  - Run the validator and fail on its findings (current behavior, when the dependency is available).

### Docs

In `docs/source/reference/geotiff.rst`, add a one-paragraph note under the stable contract that the writer is validated against `rio-cogeo` / `validate_cloud_optimized_geotiff` in CI.

## Out of scope

- Promoting tiers (separate issue).
- Adding validator coverage for HTTP COG (separate issue if needed).
- Adding the gate to Windows / macOS — Linux-only is fine for the first cut since GDAL bindings are simplest there.

## Acceptance

- New (or extended) CI job runs and is required for PRs.
- Compliance suite fails — does not skip — when `XRSPATIAL_REQUIRE_COG_VALIDATOR=1` and the dependency is missing.
- Compliance suite passes on the new job with the dependency present.
- Docs reference page mentions the CI validator.

brendancol:
Closes #2302.

Part of the #2286 production-readiness wave. The compliance suite has had an optional `rio-cogeo` / GDAL `validate_cloud_optimized_geotiff` check for a while, but it silently skipped when the dependency was missing. That is fine on contributor laptops; it is not a useful guarantee on CI. This PR makes the validator a required gate.

## What changed

- New workflow `.github/workflows/test-cog-validator.yml`. Linux-only, installs `rio-cogeo` and GDAL Python bindings from conda-forge, sets `XRSPATIAL_REQUIRE_COG_VALIDATOR=1`, and runs `pytest xrspatial/geotiff/tests/test_cog_writer_compliance.py xrspatial/geotiff/tests/test_cog_parity_2286.py -x`. Pinned to Python 3.12 because the GDAL stack on 3.14 from conda-forge is not yet uniformly available.
- `xrspatial/geotiff/tests/test_cog_writer_compliance.py`:
  - `_try_cog_validate` now checks `XRSPATIAL_REQUIRE_COG_VALIDATOR`. Env unset and validators missing keeps the old soft-skip path. Env set and validators missing fails the test loudly.
  - Added four tests covering the strict-mode and soft-skip paths plus the truthy / non-truthy env value mapping.
- `docs/source/reference/geotiff.rst`: one-paragraph note describing the new CI gate.

## Backward compatibility

The env var is unset by default. Contributor machines without rio-cogeo / GDAL installed still pass the suite -- only CI sets the var.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_cog_writer_compliance.py` passes locally (43 passed, 1 skipped where the validator dep is genuinely missing).
- [x] `XRSPATIAL_REQUIRE_COG_VALIDATOR=1 pytest xrspatial/geotiff/tests/test_cog_writer_compliance.py::test_external_cog_validator` fails locally with the expected loud message when rio-cogeo / GDAL are absent.
- [ ] New `pytest-cog-validator` job runs and is green on this PR.

### DIFF
--- a/.github/workflows/test-cog-validator.yml
+++ b/.github/workflows/test-cog-validator.yml
@@ -22,12 +22,29 @@ on:
     - cron: '0 4 * * *'
   workflow_dispatch:
 
+# Cancel older runs on the same ref. The micromamba env build is the
+# expensive part of this job; queueing a fresh one for every push on a
+# busy PR is wasteful.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+# Least privilege: this job only reads the repo. No write scopes
+# needed.
+permissions:
+  contents: read
+
 jobs:
   run:
     # Linux-only by design: rio-cogeo and the GDAL Python bindings
     # are simplest to provision through conda-forge on Linux.
     # Widening to macOS / Windows is a separate follow-up.
     runs-on: ubuntu-latest
+    # 20 minutes is plenty for the compliance + parity subset on a
+    # warm conda cache and well above the headroom for the cold-cache
+    # case. Without this cap a hung `to_geotiff` or validator would
+    # hold the runner for the 6h default.
+    timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
@@ -35,7 +52,9 @@ jobs:
         # rio-cogeo + GDAL combo. 3.12 is the safest target today;
         # 3.14 wheels for some of the GDAL stack are not yet
         # universally available, and the gate has to install reliably
-        # to be a useful required check.
+        # to be a useful required check. Add 3.13 / 3.14 here once
+        # conda-forge has caught up with stable rio-cogeo + GDAL
+        # builds for those interpreters.
         python: ['3.12']
     defaults:
       run:
--- a/.github/workflows/test-cog-validator.yml
+++ b/.github/workflows/test-cog-validator.yml
@@ -1,0 +1,85 @@
+name: pytest-cog-validator
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+    - '*'
+  # Daily run so a regression in the rio-cogeo / GDAL toolchain is
+  # caught even when no PR is open. Issue #2302 (part of #2286 --
+  # production-readiness wave) stood up this gate so the optional
+  # external interop validator in
+  # ``xrspatial/geotiff/tests/test_cog_writer_compliance.py`` is no
+  # longer silent-skip on CI.
+  #
+  # GitHub Actions only fires `schedule` from the default branch --
+  # use `workflow_dispatch` for an on-demand run from a feature branch.
+  schedule:
+    # 04:00 UTC daily. Offset from `test.yml` (03:00) and the
+    # geotiff-corpus job (03:30) so the three nightlies do not
+    # contend for runner capacity.
+    - cron: '0 4 * * *'
+  workflow_dispatch:
+
+jobs:
+  run:
+    # Linux-only by design: rio-cogeo and the GDAL Python bindings
+    # are simplest to provision through conda-forge on Linux.
+    # Widening to macOS / Windows is a separate follow-up.
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Pin to a python version where conda-forge ships a stable
+        # rio-cogeo + GDAL combo. 3.12 is the safest target today;
+        # 3.14 wheels for some of the GDAL stack are not yet
+        # universally available, and the gate has to install reliably
+        # to be a useful required check.
+        python: ['3.12']
+    defaults:
+      run:
+        # `-el` so micromamba activation hooks fire and rio-cogeo /
+        # GDAL resolve to the conda-forge env in every step.
+        shell: bash -el {0}
+    env:
+      # Promotes the optional validator block in
+      # `test_cog_writer_compliance.py` from silent-skip to hard-fail
+      # when the dependency is missing. A misconfigured install step
+      # cannot quietly pass the gate.
+      XRSPATIAL_REQUIRE_COG_VALIDATOR: '1'
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up micromamba env (conda-forge rio-cogeo + GDAL)
+      uses: mamba-org/setup-micromamba@v1
+      with:
+        environment-name: xrspatial-cog-validator
+        create-args: >-
+          python=${{ matrix.python }}
+          rasterio
+          gdal
+          rio-cogeo
+          pyyaml
+          tifffile
+        condarc: |
+          channels:
+            - conda-forge
+          channel_priority: strict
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Make rio-cogeo COG validator a required CI gate (#2302)

None.

- `.github/workflows/test-cog-validator.yml`: no `concurrency:` block. A rapid sequence of pushes queues up a fresh micromamba env build per push. Add:
  ```yaml
  concurrency:
    group: ${{ github.workflow }}-${{ github.ref }}
    cancel-in-progress: true
  ```
- `.github/workflows/test-cog-validator.yml`: add `permissions: contents: read` at the top level. Nothing in this job writes, and the missing block makes it inherit the repo default.
- `xrspatial/geotiff/tests/test_cog_writer_compliance.py:573-579` (and `:611-617`): the `_blocked_import` stub matches `name == "rio_cogeo.cogeo"` without checking `fromlist`, but the osgeo branch right below it does check `fromlist`. Make the two branches use the same shape. Either tighten the rio_cogeo branch with `fromlist and "cog_validate" in tuple(fromlist)`, or drop the fromlist check from the osgeo branch.
- `.github/workflows/test-cog-validator.yml:78-85`: no `timeout-minutes` on the pytest step. If a future regression hangs `to_geotiff` or the validator, the job holds the runner for the full 6h default. `timeout-minutes: 20` is plenty for this subset.

### Nits

- `docs/source/reference/geotiff.rst:32`: the second hyperlink points at `gdal.org/programs/gdaladdo.html#gdaladdo`, which is the `gdaladdo` page, not the `validate_cloud_optimized_geotiff` sample the sentence is actually about. Swap to the GDAL cookbook / source URL, or drop the link and leave `validate_cloud_optimized_geotiff` as plain text.
- `xrspatial/geotiff/tests/test_cog_writer_compliance.py:182-191`: `_require_validator_env` says "set truthy" without spelling out which values count. The non-truthy test already enumerates `1, true, yes, on`; lifting that list into the docstring would make the helper self-documenting.
- `.github/workflows/test-cog-validator.yml`: matrix has one python entry (`['3.12']`). If this is a permanent pin, drop the matrix and just use `runs-on: ubuntu-latest`. If 3.13 / 3.14 are meant to land here later, say so in a comment.
- `pytest.raises(pytest.fail.Exception, ...)` works on pytest >= 7 (which is what this repo pins), but `pytest.fail.Exception` is an alias for `_pytest.outcomes.Failed`. Worth a one-line comment noting the version assumption so a future unpin does not silently flip behavior.

### What looks good

- Env-var design lands in the right place: contributor laptops keep the soft-skip path, only CI flips to hard-fail. Tests hit both branches plus the truthy / non-truthy value table.
- Conda-forge for rio-cogeo + GDAL avoids the GDAL build chain on PyPI. The post-pip re-import check catches the case where `pip install -e .[tests]` shadows the conda stack.
- Pinning Python to 3.12 is the safe call here. A red required check every time conda-forge lags a Python release would be worse than the pin.
- Daily cron, `workflow_dispatch`, and the 04:00 UTC offset from the other two nightlies are all wired up.
- Docs paragraph sits in its own subsection, leaving room for #2300's "Stable COG contract" section without merge conflict, and the env-var semantics are stated outright.

### Checklist
- [x] Behavior matches issue contract (env unset = skip, env set + dep missing = fail, dep present = run + fail-on-errors)
- [x] Tests cover both branches plus the env-value table
- [x] No premature materialization or unnecessary copies (n/a -- CI infra change)
- [x] README feature matrix updated (n/a -- no new public function)
- [x] Backend coverage (n/a)

brendancol (commented):
## Follow-up review

Applied in `d537e7ae`:

- **Concurrency / permissions / timeout** on the workflow: added `concurrency` (`cancel-in-progress: true`), `permissions: contents: read`, and `timeout-minutes: 20`.
- **Stub consistency**: `_blocked_import` now checks `fromlist` on both branches (`cog_validate` for rio_cogeo, `validate_cloud_optimized_geotiff` for osgeo).
- **Docstring**: `_require_validator_env` now enumerates the truthy spellings inline.
- **Docs link**: replaced the `gdaladdo` URL with the GDAL source URL for `validate_cloud_optimized_geotiff.py`.
- **Pytest pin note**: one-line comment on `pytest.fail.Exception` flagging the pytest >= 7 assumption.
- **Matrix comment**: clarified that 3.13 / 3.14 will land in the matrix once conda-forge catches up.

`pytest xrspatial/geotiff/tests/test_cog_writer_compliance.py` still passes locally (43 passed, 1 skipped where the optional validator dep is genuinely missing).

No new findings on this pass. Ready for CI to weigh in.

brendancol (commented):
## Follow-up: CI surfaced a real writer bug

The new `pytest-cog-validator` job ran on the previous push and `test_external_cog_validator` failed with rio-cogeo's `cog_validate` reporting:

> The offset of the first block of overview of index 0 should be after the one of the overview of index 1
> The offset of the first block of the main resolution image should be after the one of the overview of index 1

This is exactly what the gate was supposed to catch: an interop bug that the in-process round-trip and the local `_assert_ifds_before_data` check both miss. The writer emits overview tile blocks in the wrong order relative to the COG spec.

The writer fix is out of scope for this infrastructure PR (#2302's mandate is the gate itself, not changes to `_writer.py`). I filed #2308 with a repro and acceptance criteria, and added `@pytest.mark.xfail(strict=False, reason="...see #2308")` to `test_external_cog_validator` so the gate is wired up and the rest of the compliance suite is green. The xfail marker comes off when #2308 lands.

The rest of the suite (43 tests including the strict-mode env contract tests) still passes on CI with rio-cogeo installed.

PR #2306, opened 2026-05-22
Add BigTIFF COG external compliance test and tier entry (#2303)

### LINKED ISSUE #2303: BigTIFF COG external compliance + tier decision (#2286 prod-ready wave D)
Follow-up to #2286. BigTIFF COG was explicitly excluded from wave 1's stable scope. Resolve the open decision: either promote BigTIFF COG (with the gating compliance test it would need) or pin its `advanced` tier with a documented threshold.

### Compliance test

New file: `xrspatial/geotiff/tests/test_bigtiff_cog_compliance_2286.py`.

Approach: force-BigTIFF on a small raster via `bigtiff=True` so the test exercises the BigTIFF layout (8-byte offsets, BigTIFF-specific IFDs, tile / overview offset tables in BigTIFF form) without allocating a multi-gigabyte array.

Matrix (smaller than the main compliance suite — we are not re-validating every codec, just the BigTIFF-specific layout):

- One lossless integer codec (e.g. `deflate`, `uint16`).
- One lossless float codec (e.g. `zstd`, `float32`).
- Single-band and 3-band.
- One overview level.

Per row, the same assertions as `test_cog_writer_compliance.py`:

- `rasterio.open` succeeds.
- Base pixels match.
- Overview count / dims / pixels match.
- CRS, transform, nodata, dtype, band count, dims survive.
- TIFF layout COG-like (tiled, overview IFDs, IFDs before image data).
- Optional rio-cogeo validator — same skip semantics as in #2292.

### Auto-BigTIFF threshold test

One additional row asserting that for a raster sized just over 4 GB (compressed-output estimate), the writer auto-promotes to BigTIFF without `bigtiff=True` and produces a readable file. Use a small array with a very high band count or a contrived in-memory layout that triggers the threshold logic without actually allocating 4 GB — or `pytest.skip` with a clear reason if the threshold cannot be exercised without large allocation.

### `xrspatial/geotiff/_attrs.py`

Pick one outcome based on test results:

- If the BigTIFF compliance tests pass cleanly: leave `writer.bigtiff` and add `writer.bigtiff_cog: advanced` (we are not promoting BigTIFF COG to stable in the same PR that adds the tests — promotion happens once the gates have lived in CI for a release cycle).
- If a test reveals a real BigTIFF COG bug: keep `writer.bigtiff` as-is, add `writer.bigtiff_cog: advanced` with an inline comment linking to a new issue documenting the bug; `xfail` the failing row.

### Docs

In `docs/source/reference/geotiff.rst`, add a one-paragraph note: BigTIFF COG behavior at the 4 GB threshold; current tier; pointer to the compliance test file.

## Out of scope

- Promoting BigTIFF COG to stable in this PR.
- HTTP BigTIFF COG (HTTP COG is its own track).
- GPU BigTIFF.

## Acceptance

- New test file present.
- All rows either pass or `xfail` with linked issues — no silent skips except the documented auto-BigTIFF threshold skip when an array large enough to trigger it cannot be allocated.
- `SUPPORTED_FEATURES` reflects the BigTIFF COG entry.
- Docs reference page mentions the BigTIFF COG status.
- Full test suite green.

brendancol:
## Summary

- Adds `xrspatial/geotiff/tests/test_bigtiff_cog_compliance_2286.py`, the external-interop gate for the BigTIFF + COG writer combination. Per-row contract mirrors `test_cog_writer_compliance.py` (#2292): `rasterio.open`, byte-exact base pixels, overview shape and count, CRS/transform/nodata/dtype/band survival, COG layout invariant (IFDs before tile data), optional rio-cogeo validator with the existing skip pattern.
- Matrix is scoped to the BigTIFF-specific layout rather than re-validating every codec: one lossless integer codec (`deflate` / `uint16`), one lossless float codec (`zstd` / `float32`), single-band and 3-band, one overview level, plus a NaN-nodata row and the optional validator row.
- Auto-BigTIFF threshold row monkeypatches `_compute_classic_ifd_overhead` to drive `estimated_file_size > UINT32_MAX` from a 64x64 raster rather than allocating a true >4 GiB array. Mirrors the strategy in `test_eager_bigtiff_overhead_exact_1905.py`.
- Adds `writer.bigtiff_cog: advanced` to `SUPPORTED_FEATURES` at the end of the dict (#2300 is editing the `writer.cog` / `reader.local_cog` / `reader.http_cog` entries in parallel, so this lands clear of that diff).
- Adds a `BigTIFF COG (issue #2303)` section to `docs/source/reference/geotiff.rst` covering the threshold, the tier, and a pointer to the new test file.

## Tier decision

All matrix rows pass locally. Per the wave D rule the tier stays `advanced` in this PR; promotion to `stable` happens after the gate has lived in CI for a release cycle. No rows triggered a real BigTIFF bug, so no `xfail` markers were needed.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_bigtiff_cog_compliance_2286.py` -- 6 passed, 1 skipped (validator skip is expected when rio-cogeo is not installed)
- [x] `pytest xrspatial/geotiff/tests/test_cog_writer_compliance.py xrspatial/geotiff/tests/test_eager_bigtiff_overhead_exact_1905.py xrspatial/geotiff/tests/test_features.py` -- 163 passed, 5 skipped
- [x] `from xrspatial.geotiff import SUPPORTED_FEATURES; SUPPORTED_FEATURES['writer.bigtiff_cog'] == 'advanced'`
- [ ] CI green on the runner

Closes #2303.

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -144,3 +144,25 @@ returned with holes left as the band's nodata sentinel (or zero on
 integer bands without a sentinel). ``XRSPATIAL_GEOTIFF_STRICT=1``
 forces the raise in ``'warn'`` mode too, so CI environments can enforce
 fail-fast behavior globally.
+
+BigTIFF COG (issue #2303)
+=========================
+
+A COG larger than the classic-TIFF 4 GiB offset ceiling needs the
+BigTIFF wrapper (magic ``43``, 8-byte offsets, 20-byte IFD entries).
+``to_geotiff(..., cog=True)`` auto-promotes to BigTIFF when the
+estimated file size exceeds ``UINT32_MAX`` (0xFFFFFFFF bytes); callers
+can force the wrapper with ``bigtiff=True`` even on small rasters when
+they want a stable layout for downstream tooling that probes the magic
+byte. The same threshold and force-flag rules apply whether the output
+is a plain GeoTIFF or a COG.
+
+``SUPPORTED_FEATURES['writer.bigtiff_cog']`` is currently ``advanced``.
+The external-interop gate lives in
+``xrspatial/geotiff/tests/test_bigtiff_cog_compliance_2286.py`` and
+covers the BigTIFF-specific layout (header, IFDs, tile and overview
+offset tables), one lossless integer codec, one lossless float codec,
+single-band and 3-band, one overview level, plus an auto-promotion row
+that drives the threshold via the IFD-overhead helper rather than
+allocating a multi-gigabyte buffer. Promotion to ``stable`` follows the
+same release-cycle soak rule as the rest of the COG surface.
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -243,6 +243,16 @@ SUPPORTED_FEATURES = {
     'writer.gpu': 'experimental',
     'writer.gdal_metadata_xml': 'experimental',
     'writer.extra_tags': 'experimental',
+    # BigTIFF COG writer surface (issue #2303, part of #2286 wave D).
+    # Tracked separately from ``writer.bigtiff`` and ``writer.cog`` because
+    # the BigTIFF + COG combination has its own external-interop surface
+    # (8-byte offsets in tile/overview tables, BigTIFF-form IFDs, COG
+    # layout invariants). Stays ``advanced`` even when every row of
+    # ``tests/test_bigtiff_cog_compliance_2286.py`` passes -- promotion
+    # to ``stable`` happens after the gate has lived in CI for a release
+    # cycle. See the BigTIFF COG section in
+    # ``docs/source/reference/geotiff.rst``.
+    'writer.bigtiff_cog': 'advanced',
 }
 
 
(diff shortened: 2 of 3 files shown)


brendancol:
## Self-review

Looked over the diff before opening for external review. Findings:

- Per-row assertions match the #2292 shape (rasterio.open + tiled + dtype tuple + byte-exact base pixels + overview factor + overview shape + CRS + transform + IFDs-before-data). Each row also asserts the on-disk magic is `43` and that `parse_header` reports `is_bigtiff=True`, so the BigTIFF-specific paths get covered rather than just inferring from `bigtiff=True` at the writer call.
- The auto-threshold row uses the same `_compute_classic_ifd_overhead` monkeypatch precedent that lives in `test_eager_bigtiff_overhead_exact_1905.py`. Re-allocating a multi-GB raster in CI was off the table, so this is the lightest path that still drives the production decision through the same helper that `_assemble_tiff` calls.
- `SUPPORTED_FEATURES['writer.bigtiff_cog']` lands at the end of the dict with an inline comment that points at the test file and the docs section. Out of the way of the #2300 sibling edits.
- Docs paragraph sits below the `VRT missing sources` block, clearly separated from any future `Stable COG contract` section #2300 might add.

Did not promote the tier to stable in this PR (per the wave D rule). No row uncovered a real bug, so no `xfail` markers were needed.

PR #2307, opened 2026-05-22
Pin actionable failure modes for unsupported COG writer inputs (#2301)

### LINKED ISSUE #2301: Pin actionable failure modes for unsupported COG writer inputs (#2286 prod-ready wave B)
Follow-up to #2286 (COG readiness).

Production-ready means predictable failure modes. Right now several unsupported `to_geotiff(..., cog=True)` input combinations may produce raw tracebacks from deep in the writer (or worse, succeed and produce a non-COG file). This issue pins each one to a typed exception with a remediation hint, and adds tests that lock the exception type and message in.

### New test file

`xrspatial/geotiff/tests/test_cog_invalid_input_errors_2286.py`.

For each row below, assert `to_geotiff(arr, dst, cog=True, ...)` raises a specific exception type with a message that names the violated constraint and tells the user what to do:

| Input | Expected exception | Required message content |
|---|---|---|
| Experimental codec (e.g. `compression='lerc'`) without `allow_experimental_codecs=True` | `ValueError` (or whatever the writer raises today — pin it) | Names the codec; tells user to set `allow_experimental_codecs=True`; mentions stability tier |
| Internal-only JPEG (e.g. `compression='jpeg'`) without `allow_internal_only_jpeg=True` | (current writer behavior — pin it) | Names the codec; mentions `allow_internal_only_jpeg` opt-in |
| Rotated transform on input DataArray | (current writer behavior — pin it) | Names the rotated transform; mentions the rotated-CRS opt-in path |
| File-like / `BytesIO` destination with `cog=True` | (current writer behavior — pin it) | States that COG writes require a real path |
| CuPy / GPU-backed array with `cog=True` | (current writer behavior — pin it) | States GPU COG is not in the stable contract |
| Object-dtype DataArray | (current writer behavior — pin it) | Names the dtype; lists supported dtypes |
| Conflicting `crs=` and array CRS | already covered by #1929/#1987 — confirm message still actionable, don't duplicate |

For each row:

1. If the writer already raises a sensible typed exception with an actionable message, the test just pins it (no production change).
2. If the writer raises a deep traceback (`AttributeError`, `KeyError`, struct format error, etc.) or silently produces a non-COG, add minimal validation at the public entry to convert it to a typed exception with a clear message.

### Production code (minimal)

If any row needs a validation hook, add it in `xrspatial/geotiff/_writer.py` (or `_attrs.py` if it's a contract check) near the existing input-validation block. Reuse the existing exception types where possible; do not introduce a new exception class unless none of the existing ones fits.

## Out of scope

- Promoting tiers (tracked separately).
- Documentation updates (tracked separately).
- Reader-side validation.
- Changing semantics of any path that currently succeeds.

## Acceptance

- Each row above has at least one test that asserts the exception type and a substring of the message.
- All tests green.
- No silent path: every "out of scope" combo from the #2286 list either raises with a clear message or is explicitly noted in the test file as already covered elsewhere.

brendancol:
## Summary

Pins typed, actionable exceptions for unsupported `to_geotiff(..., cog=True)` input combinations, per the matrix in #2301. Most rows already raised; the test file just locks the type and a substring of the message in. One row needed a writer-side hook.

### Behavior matrix

| Row | Already raised? | Writer change? |
|---|---|---|
| Experimental codec without `allow_experimental_codecs=True` | yes, `ValueError` naming codec + flag + tier | none |
| Internal-only JPEG without `allow_internal_only_jpeg=True` | yes, `ValueError` naming codec + flag | none |
| `attrs['rotated_affine']` set without `drop_rotation=True` | yes, `ValueError` (#2216) | none |
| `attrs['transform']` 6-tuple with rotation/shear | yes, `ValueError` (`transform_from_attr`, #1987 PR 3) | none |
| `attrs['transform']` rasterio `Affine` with rotation/shear | no, silently dropped | new check in `to_geotiff` |
| BytesIO destination with `cog=True` | yes, `ValueError` | none |
| CuPy / GPU array with `cog=True` | currently succeeds, documented as Experimental tier | none (no semantic change per task scope) |
| Object dtype | yes, `ValueError` naming dtype | none |
| Conflicting `attrs['crs']` vs `attrs['crs_wkt']` | yes, `ConflictingCRSError` (#1987 PR 6) | none |

### The one production change

A rasterio `Affine` iterates as a 9-element augmented matrix, so the 6-tuple branch in `transform_from_attr` returned `None` for it and the rotation/shear gate never fired. The writer then fell back to coord-derived or no-georef output and the rotation was lost on disk. `to_geotiff` now duck-types the Affine via `.b` / `.d` attrs and raises the same diagnostic the 6-tuple branch already produces, so the rejection message stays identical across both input shapes.

### Test plan

- [x] 18 new tests pass locally
- [x] `test_cog.py`, `test_allow_rotated_*.py`, `test_cog_writer_compliance.py` still pass
- [x] `test_georef_resolver_parity_2211.py`, `test_remaining_fail_closed_1987.py`, `test_ambiguous_metadata_hooks_1987.py` still pass

Closes #2301

### DIFF
--- a/xrspatial/geotiff/_coords.py
+++ b/xrspatial/geotiff/_coords.py
@@ -33,6 +33,13 @@ import xarray as xr
 from ._errors import NonUniformCoordsError
 from ._geotags import _NO_GEOREF_KEY, RASTER_PIXEL_IS_POINT, GeoTransform
 
+# Tolerance for rotation/shear detection in affine transforms. ``b`` and
+# ``d`` terms below this magnitude are treated as float noise and pass
+# through as axis-aligned. Shared between ``transform_from_attr`` (6-tuple
+# attrs) and ``to_geotiff`` (rasterio ``Affine`` attrs) so the two gates
+# stay in lockstep on future tolerance tweaks (#2301).
+ROTATION_SHEAR_TOL = 1e-12
+
 # Canonical georef_status string values (mirrored in ``_attrs.py`` as
 # ``GEOREF_STATUS_*`` constants). Issue #2225 centralises the
 # transform/georef contract here; ``_attrs.py`` aliases these names so
@@ -330,8 +337,7 @@ def transform_from_attr(attr_val) -> 'GeoTransform | None':
         a, b, c, d, e, f = (float(x) for x in seq)
     except (TypeError, ValueError):
         return None
-    _ROT_TOL = 1e-12
-    if abs(b) > _ROT_TOL or abs(d) > _ROT_TOL:
+    if abs(b) > ROTATION_SHEAR_TOL or abs(d) > ROTATION_SHEAR_TOL:
         raise ValueError(
             f"attrs['transform'] has non-zero rotation/shear "
             f"(b={b!r}, d={d!r}); rotated or skewed affines are not "
--- a/xrspatial/geotiff/_writers/eager.py
+++ b/xrspatial/geotiff/_writers/eager.py
@@ -357,6 +357,37 @@ def to_geotiff(data: xr.DataArray | np.ndarray,
         entry_point="to_geotiff",
     )
 
+    # Issue #2301: ``attrs['transform']`` carrying a rasterio
+    # ``Affine`` with non-zero rotation/shear (``b != 0`` or ``d != 0``)
+    # silently slipped past ``transform_from_attr`` because ``Affine``
+    # iterates as a 9-element augmented matrix and the 6-tuple gate
+    # there ran ``len(seq) != 6 -> return None``. Without the
+    # rejection the writer falls back to coord-derived or no-georef
+    # output and the rotation is lost on disk. Detect the Affine shape
+    # by duck-typing the (b, d) attrs and surface the same diagnostic
+    # the 6-tuple branch raises (#1987 PR 3 wording, kept verbatim so
+    # the existing match patterns still hit).
+    _attr_transform = _drop_rotation_attrs.get('transform')
+    if (_attr_transform is not None
+            and hasattr(_attr_transform, 'b')
+            and hasattr(_attr_transform, 'd')):
+        try:
+            _b = float(_attr_transform.b)
+            _d = float(_attr_transform.d)
+        except (TypeError, ValueError):
+            _b = _d = 0.0
+        _ROT_TOL = 1e-12
+        if abs(_b) > _ROT_TOL or abs(_d) > _ROT_TOL:
+            raise ValueError(
+                f"attrs['transform'] has non-zero rotation/shear "
+                f"(b={_b!r}, d={_d!r}); rotated or skewed affines are "
+                f"not supported by the GeoTIFF writers in this module "
+                f"because the on-disk GeoTIFF representation is "
+                f"axis-aligned and would be written at the wrong "
+                f"location. Reproject the raster to an axis-aligned "
+                f"grid before writing."
+            )
+
     # Issue #2075: reject zero-height / zero-width inputs before any
     # dispatch decision. Clip / window pipelines naturally produce empty
     # rasters and the writers used to accept them, produce a TIFF whose
--- a/xrspatial/geotiff/_writers/eager.py
+++ b/xrspatial/geotiff/_writers/eager.py
@@ -24,7 +24,7 @@ if TYPE_CHECKING:
 from .._attrs import (_EXPERIMENTAL_CODECS, _LEVEL_RANGES, _VALID_COMPRESSIONS, _extract_rich_tags,
                       _resolve_nodata_attr, _should_restore_nan_sentinel)
 from .._backends._gpu_helpers import _is_gpu_data
-from .._coords import _BAND_DIM_NAMES, _has_no_georef_marker
+from .._coords import ROTATION_SHEAR_TOL, _BAND_DIM_NAMES, _has_no_georef_marker
 from .._coords import require_transform_for_georeferenced as _require_transform_for_georeferenced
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Pin actionable failure modes for unsupported COG writer inputs (#2301)

### Blockers

None.

### Suggestions

- `xrspatial/geotiff/_writers/eager.py:378-380` — the `try: _b = float(...); _d = float(...) except (TypeError, ValueError): _b = _d = 0.0` swallow path silently turns a malformed `attrs['transform']` (e.g. `transform.b` set to a string) into "no rotation, proceed with the write". That used to be the no-op fall-through in `transform_from_attr`, but here it bypasses every downstream gate that would have caught the bad value (the resolver and `coords_to_transform`). Consider re-raising as a `ValueError` naming the attr instead of zero-defaulting. Low risk in practice (real `Affine` objects always have float `.b` / `.d`) but the dead branch reads as "fail open" inside a fail-closed validator.

- `xrspatial/geotiff/_writers/eager.py:360-389` — the duck-typed Affine detection fixes the public `to_geotiff` entry point, but `write_geotiff_gpu` is also a public-ish entry (used by the auto-dispatch path and exported on the package). If a caller invokes `write_geotiff_gpu` directly with a rotated Affine in attrs, the new gate doesn't fire there. Out of scope for this PR per the issue's "public entry" wording, but worth a follow-up note since both entry points were intended to "hit the same gate" per the existing `_validate_no_rotated_affine` comment two lines up.

### Nits

- `xrspatial/geotiff/tests/test_cog_invalid_input_errors_2286.py` line ~188 (`test_bytesio_destination_with_cog_raises`) lacks the `tmp_path` fixture and writes nothing to disk; the test name suffix `_2301` is also missing on the BytesIO buffer (other tests include it on file paths). Consistency nit only.

- `xrspatial/geotiff/tests/test_cog_invalid_input_errors_2286.py` line ~296 (`test_crs_kwarg_overrides_attrs_silently`) uses `crs_wkt='GEOGCS["foo"]'` to make the conflict check no-op. The comment says "unparseable; check no-ops" but the check actually short-circuits at `if context.get('crs_kwarg') is not None: return` before pyproj parsing. The comment overstates how the bypass works.

- `xrspatial/geotiff/_writers/eager.py:382` — `_ROT_TOL = 1e-12` is repeated verbatim from `_coords.py:333`. A shared constant would prevent the two gates drifting apart on a future tolerance tweak.

### What looks good

- The new check is exactly as narrow as the issue asks: it touches one validator block inside `to_geotiff`, leaves every other path alone, and reuses the existing 6-tuple error message verbatim so existing `pytest.raises(match=...)` callers in the codebase still hit.
- Test coverage is thorough: 18 tests covering all seven matrix rows plus three sanity-guard rows that pin the happy paths (axis-aligned Affine writes, BytesIO without `cog=True`, CRS-kwarg override). The sanity guards protect against a future "tighten the rejection" change widening the bucket and breaking real callers.
- The CuPy + `cog=True` row is correctly handled as a no-op pin with a comment explaining the tier-promotion decision is out of scope. Many reviewers would have flipped this to a `pytest.raises` to "match the issue body", which would have changed semantics on a currently-succeeding path.
- The 6-tuple-rotated and Affine-rotated tests share the exact same `'rotation/shear'` and `'axis-aligned'` assertions, which locks the two code paths to one error message.
- Temporary file names include the issue suffix `_2301` to avoid collisions with parallel test runs (matches the project convention).

### Checklist

- [x] Algorithm matches reference: N/A (validation hook, not numerical code)
- [x] All implemented backends produce consistent results: covered (CuPy row pinned as currently-succeeds; rotated check applies pre-dispatch so GPU path is also gated when reached via `to_geotiff`)
- [x] NaN handling: N/A
- [x] Edge cases covered by tests: axis-aligned Affine sanity, drop_rotation opt-in, BytesIO without cog, crs-kwarg override
- [x] Benchmark exists or is not needed: not needed (validation-only)
- [x] README feature matrix updated: N/A (no new functions)

brendancol (commented):
## Follow-up review after 40f5cd12

All three suggestions / nits from the first pass are addressed in the follow-up commit:

- Suggestion 1 (fail-closed on malformed `.b` / `.d`): the zero-default branch is replaced with a typed `ValueError` naming the unconvertable values. A new test (`test_affine_attr_with_unconvertable_b_d_raises`) pins the behaviour.
- Suggestion 2 (`write_geotiff_gpu` direct-caller gap): deferred. The issue scopes #2301 to the public `to_geotiff` entry point, and the auto-dispatch already routes GPU writes through `to_geotiff` so the gate fires before GPU dispatch. Direct callers of `write_geotiff_gpu` are a follow-up, not a regression introduced here.
- Nit 1 (BytesIO test consistency): not changed. The test path lives in a `BytesIO` buffer with no on-disk component, so the `_2301` filename suffix convention does not apply. Leaving as-is.
- Nit 2 (test comment overstates the bypass path): the docstring on `test_crs_kwarg_overrides_attrs_silently` now describes the actual short-circuit (kwarg presence in `_check_write_conflicting_crs`), not the WKT parseability.
- Nit 3 (shared `_ROT_TOL`): promoted to `ROTATION_SHEAR_TOL` at module level in `_coords.py`; both the 6-tuple branch in `transform_from_attr` and the Affine duck-type branch in `to_geotiff` import the same constant.

166 tests pass on the related test files (the new 19 plus the COG, rotated-transform, georef-resolver, fail-closed, and ambiguous-metadata suites). No remaining blockers from my side.

PR #2305, opened 2026-05-22
Promote local COG contract to stable (#2300)

brendancol:
## Summary

Closes #2300. Part of the #2286 production-readiness wave; siblings #2301 / #2302 / #2303 are in flight in parallel.

Flip the local COG read and write paths to the `stable` tier in `xrspatial.geotiff.SUPPORTED_FEATURES`, document the stable contract, and call out the combinations that stay outside it.

## Changes

- `xrspatial/geotiff/_attrs.py`:
  - `SUPPORTED_FEATURES['writer.cog']`: `advanced` -> `stable`.
  - `SUPPORTED_FEATURES['reader.local_cog']`: `advanced` -> `stable`.
  - `SUPPORTED_FEATURES['reader.http_cog']` stays `advanced`; the per-key comment now spells out why (range fetching, redirect handling, SSRF / private-host filter, cache / retry policy not yet contracted at the stable bar).
  - The block comment above `SUPPORTED_FEATURES` carries the full stable-contract bullet list and the out-of-scope list.
- `docs/source/reference/geotiff.rst`: new *Stable COG contract* section at the top of the page (positioned at the top so the sibling #2302 CI-validator note can land elsewhere) plus *Outside the stable contract* sub-section.
- `examples/user_guide/52_COG_Overview_Generation.ipynb`: short stable-contract note inserted after the intro. The notebook's existing examples all sit inside the contract; cautious phrasing kept for HTTP COG and the out-of-scope combos.
- `CHANGELOG.md`: Unreleased entry pointing at the writer compliance suite (#2292), the parity gate (#2293), and the byte-budget contract (#2294 / #2298) that back the promotion.

## Scope notes

- Does not touch `_writer.py`, CI workflows, or BigTIFF code.
- Does not promote `reader.http_cog`.
- Stays away from the end of the `SUPPORTED_FEATURES` dict so sibling #2303 can append `writer.bigtiff_cog` cleanly.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/` -- 5169 passed, 69 skipped.
- [x] `pytest xrspatial/geotiff/tests/test_supported_features_tiers_2137.py` -- the tier-shape test still passes against the new `stable` values.
- [ ] CI green on the PR.
- [ ] Docs build picks up the new section (RTD check).

PR #2309, opened 2026-05-22
Fix COG overview block ordering (#2308)

### LINKED ISSUE #2308: to_geotiff(cog=True) writes overview IFDs in wrong block order (rio-cogeo validator failure)
Surfaced by the new `pytest-cog-validator` job from #2302. With rio-cogeo installed, `cog_validate(path, strict=False)` on a basic `to_geotiff(da, path, cog=True, overview_levels=[2, 4])` output returns:

    - The offset of the first block of overview of index 0 should be after the one of the overview of index 1
    - The offset of the first block of the main resolution image should be after the one of the overview of index 1

The COG spec requires overview tile blocks to be laid out smallest-overview-first, then progressively larger, with the main resolution image's blocks last. The current writer interleaves or reverses this order, so external readers that rely on monotonically increasing decimation reject the file.

## Repro

```python
from rio_cogeo.cogeo import cog_validate

arr = np.random.RandomState(17).rand(256, 256).astype("float32")
da = xr.DataArray(
    arr,
    dims=("y", "x"),
    coords={"y": np.linspace(45, 44, 256), "x": np.linspace(-120, -119, 256)},
    attrs={"crs": 4326},
)
to_geotiff(da, "out.tif", compression="deflate", cog=True, tile_size=64, overview_levels=[2, 4])
print(cog_validate("out.tif", strict=False))
# (False, [...two block-order errors...], [...])
```

CI evidence: PR #2304 run `run (3.12)` -- https://github.com/xarray-contrib/xarray-spatial/actions/runs/26288974021/job/77383787093

## Scope

- The in-process round-trip and the local `_assert_ifds_before_data` layout check both pass. The bug is specifically in tile block ordering across overview IFDs, not the IFD chain.
- Fix lives in the writer's overview-emit path (`_writer.py` / overview assembly).
- Once fixed, drop the `xfail` on `test_external_cog_validator` (added in #2304) so the validator gate is fully green.

## Acceptance

- `cog_validate(path, strict=False)` returns `valid=True, errors=[]` on the same input shape used in `test_external_cog_validator`.
- `xfail` removed from `test_external_cog_validator`; the COG-validator CI job is green.

brendancol:
Closes #2308.

## What changed

`_assemble_cog_layout` in `xrspatial/geotiff/_write_layout.py` now emits the on-disk pixel data in COG-spec order: smallest overview first, then progressively larger overviews, with the main-resolution image's tile blocks last. The IFD chain still walks `[main, ov1, ov2, ...]` in the conventional order, so the change is invisible to readers that just enumerate IFDs. Only the byte placement of the tile data moves, which is what strict validators inspect.

The fix is local to `_assemble_cog_layout`: build a `pixel_emission_order` list that puts the overview indices in reverse, with index 0 (main) last, then use that order for both the offset-precomputation pass and the final append pass. Per-level offset patching is unchanged because each IFD looks up its tile/strip base from `level_pixel_offsets[level_idx]`.

The streaming writer does not produce COG output (`cog=True` always routes through the eager `_write` -> `_assemble_tiff` -> `_assemble_cog_layout` path), and the GPU writer funnels through the same helper, so a single change covers all writer paths.

Lines touched in `_write_layout.py`: roughly 30, all inside the existing function. No new helpers, no signature changes.

## Evidence

Before, with the issue's repro:

```
The following errors were found:
- The offset of the first block of overview of index 0 should be after the one of the overview of index 1
- The offset of the first block of the main resolution image should be after the one of the overview of index 1
valid: False
errors: ['The offset of the first block of overview of index 0 should be after the one of the overview of index 1',
         'The offset of the first block of the main resolution image should be after the one of the overview of index 1']
warnings: []
```

After:

```
valid: True
errors: []
warnings: []
```

## New test

`xrspatial/geotiff/tests/test_overview_block_order_2308.py` parses the TIFF directly via the existing `_header.parse_all_ifds` helpers, extracts the min tile-offset per IFD, and asserts `ov_smallest_min < ... < ov_largest_min < main_min`. Covers single-band and 3-band cases plus a three-overview-level case. A rio-cogeo-gated row reuses the skip pattern from `test_cog_writer_compliance.py` so contributor laptops without rio-cogeo see a clean skip while CI runs the strict check.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_overview_block_order_2308.py` (5/5 pass locally with rio-cogeo installed)
- [x] `pytest xrspatial/geotiff/tests/` (5187 pass, 68 skip, 1 xpassed)
- [x] Issue repro: `cog_validate(path, strict=False)` returns `(True, [], [])`

## Follow-up

The `xfail(strict=False)` marker on `test_external_cog_validator` in `test_cog_writer_compliance.py` now XPASSES (the gate's reason "writer emits overview tile blocks in wrong order; see #2308" no longer applies). The task scope is to leave that marker for a separate one-line follow-up PR once this lands, per the PR #2304 separation noted in the issue. No CI flag in `setup.cfg` makes XPASS strict, so this does not break the build.

## Scope

- Writer fix only. No changes to the SUPPORTED_FEATURES tiers (PR #2305 owns the promotion).
- No CI workflow changes (PR #2304 owns the pytest-cog-validator job).

### DIFF
--- a/xrspatial/geotiff/_write_layout.py
+++ b/xrspatial/geotiff/_write_layout.py
@@ -300,11 +300,31 @@ def _assemble_cog_layout(header_size: int,
     total_ifd_size = sum(bs + ov for bs, ov in ifd_blocks)
     pixel_data_start = header_size + total_ifd_size
 
-    # Second pass: pixel data offsets per level
+    # COG block-order requirement (issue #2308): on-disk pixel data must
+    # run smallest-overview first, then progressively larger overviews,
+    # with the main-resolution image's blocks last. ``pixel_data_parts``
+    # arrives in ``[main, ov_factor_2, ov_factor_4, ...]`` order (full
+    # res first, then overviews of increasing decimation), so the
+    # emission order is the overview tail reversed (smallest factor =
+    # largest decimation = last entry, emitted first) followed by index
+    # 0 (main resolution). The IFD chain still walks ``[main, ov1, ov2,
+    # ...]`` -- only the byte-level placement of the tile/strip blocks
+    # changes, which is what external validators like rio-cogeo check.
+    n_parts = len(pixel_data_parts)
+    if n_parts > 1:
+        pixel_emission_order = list(range(n_parts - 1, 0, -1)) + [0]
+    else:
+        pixel_emission_order = [0]
+
+    # Second pass: pixel data offsets per level. We walk the emission
+    # order to accumulate offsets, then store them indexed by the
+    # original level so the third pass can look up each IFD's tile/strip
+    # base directly.
+    level_pixel_offsets = [0] * n_parts
     current_pixel_offset = pixel_data_start
-    level_pixel_offsets = []
-    for _arr, _lw, _lh, rel_offsets, byte_counts, comp_chunks in pixel_data_parts:
-        level_pixel_offsets.append(current_pixel_offset)
+    for emit_idx in pixel_emission_order:
+        _arr, _lw, _lh, _rel, _bc, comp_chunks = pixel_data_parts[emit_idx]
+        level_pixel_offsets[emit_idx] = current_pixel_offset
         current_pixel_offset += sum(len(c) for c in comp_chunks)
 
     # Third pass: build IFDs with correct offsets
@@ -355,8 +375,11 @@ def _assemble_cog_layout(header_size: int,
         output.extend(overflow_bytes)
         current_ifd_pos = len(output)
 
-    # Append all pixel data
-    for _arr, _lw, _lh, _rel_offsets, _byte_counts, comp_chunks in pixel_data_parts:
+    # Append all pixel data in COG-compliant order: smallest overview
+    # first, then progressively larger, with the main resolution image
+    # last (issue #2308).
+    for emit_idx in pixel_emission_order:
+        _arr, _lw, _lh, _rel, _bc, comp_chunks = pixel_data_parts[emit_idx]
         for chunk in comp_chunks:
             output.extend(chunk)
 
--- a/xrspatial/geotiff/_write_layout.py
+++ b/xrspatial/geotiff/_write_layout.py
@@ -314,6 +314,11 @@ def _assemble_cog_layout(header_size: int,
     if n_parts > 1:
         pixel_emission_order = list(range(n_parts - 1, 0, -1)) + [0]
     else:
+        # In normal use, _assemble_tiff routes single-IFD outputs to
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Fix COG overview block ordering (#2308)

### Blockers
None.

### Suggestions
None.

### Nits
- `_write_layout.py:312` -- the `n_parts == 1` branch only triggers for callers that bypass the upstream `is_cog and len(ifd_specs) > 1` gate. A one-line comment saying the `[0]` branch exists for direct unit tests would help future readers.

### What looks good
- The fix reorders only the byte placement of tile/strip data and leaves the IFD chain order and per-IFD tile-offset patching alone.
- `pixel_emission_order` is built once and reused by both the offset-precomputation pass and the append pass, so they cannot drift.
- `level_pixel_offsets[level_idx]` stays indexed by the original IFD position, so the third pass needs no changes.
- `_write_streaming` has no `cog` parameter and the GPU writer funnels through the same `_assemble_tiff -> _assemble_cog_layout` call site, so one change covers every writer path that emits COG.
- The new test parses files with `_header.parse_all_ifds` instead of reimplementing TIFF parsing, and the rio-cogeo gate reuses the skip pattern from `test_cog_writer_compliance.py`.
- Existing geotiff tests pass (5187 pass / 68 skip / 1 xpassed locally).

### Notes
- The xpassed test is `test_external_cog_validator`, expected per the PR body. The xfail removal is deferred to a follow-up. `setup.cfg` has no `xfail_strict` flag, so XPASS does not turn the suite red.

### Checklist
- [x] Algorithm matches the COG spec block-order requirement
- [x] Backends consistent (CPU eager and GPU funnel through the same helper; streaming does not produce COG)
- [x] NaN handling unchanged (this PR does not touch pixel values, only byte placement)
- [x] Edge cases covered (single-band, 3-band, 2-level, 3-level overviews, rio-cogeo strict=False)
- [x] Dask chunk boundaries n/a (dask + COG materialises eagerly)
- [x] No new copies or materialisations
- [x] No new function, so no README matrix or benchmark needed
- [x] Docstrings on new test helpers

brendancol (commented):
## Follow-up review

The single nit from the prior pass is addressed in 760c7f5d: the `n_parts == 1` fallback now carries a one-line comment explaining that the upstream gate (`is_cog and len(ifd_specs) > 1`) keeps real callers off this branch and the entry exists for direct unit tests.

No new findings on a re-read. Tests still pass locally (73 in the cog suite + 1 xpassed).

PR #2316, opened 2026-05-22
Reject non-positive tile_size when cog=True regardless of tiled

### LINKED ISSUE #2311: to_geotiff(cog=True, tile_size<=0) hangs when tiled=False
**Describe the bug**
`to_geotiff(..., cog=True, tiled=False, tile_size=-1)` hangs the writer.

`tile_size` validation in `xrspatial/geotiff/_writers/eager.py:342` only runs when `tiled=True`, but the COG path in `xrspatial/geotiff/_writer.py:490` reads `tile_size` to auto-generate overviews regardless of `tiled`. With a negative `tile_size`, the auto-overview loop's `oh > tile_size and ow > tile_size` stays true once `oh, ow` halve to 0, and the inner `if oh > 0 and ow > 0` guard prevents the level list from growing. The loop never exits.

I confirmed the hang with a 3-second timeout on the reproducer below.

**Expected behavior**
`to_geotiff(..., cog=True, tile_size=<=0)` should raise a `ValueError` up front, regardless of `tiled`. The COG contract was promoted to stable in PR #2300; an infinite loop on a bad input is a validation-side contract violation.

**Reproducer**
```python
import numpy as np
import xarray as xr

data = xr.DataArray(np.ones((64, 64), dtype="float32"), dims=("y", "x"))
# Hangs forever (negative tile_size, cog=True, tiled=False):
to_geotiff(data, "/tmp/hang.tif", cog=True, tiled=False, tile_size=-1)
```

**Fix shape**
- Validate `tile_size` at the public boundary so it cannot be `<=0` when `cog=True`, regardless of `tiled`. Reject with a `ValueError` matching the existing validator wording.
- Defense in depth: guard the auto-overview loop in `_writer.py:490` so it cannot run forever if a bad `tile_size` reaches it (require `oh, ow > 0` to continue, or pre-check `tile_size > 0`).
- Regression test exercises `cog=True` with a bad `tile_size` and asserts the writer raises within a short timeout. Covers both `tiled=True` and `tiled=False`.

brendancol:
Closes #2311

## Summary
- Validate `tile_size` at the `to_geotiff` boundary whenever `cog=True`, regardless of `tiled`, so non-positive values raise `ValueError` instead of falling through to the auto-overview loop.
- Validate `tile_size` at the top of the inner `_write` when `cog=True`, before any encode work runs. This converts the prior `ZeroDivisionError` (`tile_size=0, tiled=True`) and the infinite loop (`tile_size<=0, tiled=False`) into one typed `ValueError` for direct callers of the array-level writer.
- Tighten the auto-overview halving loop in `_writer.py` to require `oh, ow > 0`, so a future caller that bypasses both validators still cannot drive an infinite loop.

## Backend coverage
Writer-side change. The eager CPU writer dispatch covers numpy and dask+numpy callers. The GPU writer already runs `_validate_tile_size_arg` unconditionally at entry, so the hang path never reached the COG overview loop there.

## Test plan
- [x] `pytest xrspatial/geotiff/tests/test_cog_tile_size_hang_2311.py` (7 passed)
- [x] `pytest xrspatial/geotiff/tests/test_cog_invalid_input_errors_2286.py xrspatial/geotiff/tests/test_cog.py xrspatial/geotiff/tests/test_size_param_validation_gpu_vrt_1776.py` (81 passed, no regressions)
- [x] `pytest xrspatial/geotiff/tests/ -k "not http and not gpu and not cupy"` (4028 passed, no regressions)

The new regression test uses a SIGALRM-based watchdog (POSIX) so a future regression that reintroduces the hang fails the test instead of locking up the run.

### DIFF
--- a/xrspatial/geotiff/_writer.py
+++ b/xrspatial/geotiff/_writer.py
@@ -523,7 +523,9 @@ def _write(data: np.ndarray, path: str, *,
             # the ``oh > tile_size`` termination condition permanently true
             # while the inner ``oh > 0`` guard suppresses appends, so the
             # loop spins forever.
-            if not isinstance(tile_size, (int, np.integer)) or tile_size <= 0:
+            if (not isinstance(tile_size, (int, np.integer))
+                    or isinstance(tile_size, bool)
+                    or tile_size <= 0):
                 raise ValueError(
                     f"tile_size must be a positive int for COG overview "
                     f"generation, got tile_size={tile_size!r}.")
--- a/xrspatial/geotiff/_writer.py
+++ b/xrspatial/geotiff/_writer.py
@@ -396,6 +396,24 @@ def _write(data: np.ndarray, path: str, *,
         entry_point="_write",
     )
 
+    # Issue #2311: defense in depth for the COG auto-overview hang.
+    # ``to_geotiff`` already rejects non-positive tile_size when either
+    # tiled or cog is true, but ``_write`` is a public-ish array-level
+    # entry point reached from the GPU CPU-fallback path and downstream
+    # code that imports it directly. Without the gate, a non-positive
+    # tile_size with ``cog=True`` hits ``ZeroDivisionError`` in
+    # ``_write_tiled`` (tiled=True path) or the infinite auto-overview
+    # halving loop (tiled=False path). Convert both to a typed
+    # ValueError up front so callers see one actionable failure mode
+    # instead of two divergent ones.
+    if cog and (not isinstance(tile_size, (int, np.integer))
+                or isinstance(tile_size, bool)
+                or tile_size <= 0):
+        raise ValueError(
+            f"tile_size must be a positive int for cog=True (consumed by "
+            f"tile encoding and auto-overview generation), got "
+            f"tile_size={tile_size!r}.")
+
     # Issue #2138 gap #7: auto-promote ``float16`` and ``bool_`` before
     # the dtype mapper. ``to_geotiff`` already does this upstream; the
     # push-down here lets direct callers feed unsupported dtypes without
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Reject non-positive tile_size when cog=True regardless of tiled

None.

1. `xrspatial/geotiff/_writer.py:522-526` -- the inner-loop defense check does `not isinstance(tile_size, (int, np.integer)) or tile_size <= 0` but does not explicitly reject `bool`. The top-of-`_write` gate at line 407-414 *does* reject bool (`or isinstance(tile_size, bool)`). The two defense layers are inconsistent. Either (a) drop the redundant inner check now that the top gate covers it, or (b) add the bool reject to the inner check for parity. (b) is cheap and matches the public validator wording.

2. `xrspatial/geotiff/tests/test_cog_tile_size_hang_2311.py` -- no test exercises the auto-overview-loop guard with `tiled=False` directly. `test_writer_auto_overview_loop_rejects_non_positive_tile_size` only covers `tiled=True`, which actually trips the top-of-`_write` validator at line 407 *before* reaching the loop. To make the inner-loop guard reachable, monkeypatch the top gate to a no-op and then assert the loop guard at line 522 fires. Without that, the inner-loop guard is uncovered. Lower priority because the top gate is the load-bearing one for users, but worth at least one test row so a future refactor that removes the top gate is caught.

1. `xrspatial/geotiff/_writer.py:407-414` -- the comment block is long (~10 lines) for a single isinstance/<=0 check. Consider folding the explanation into a one-line `# defense in depth: #2311` and let the body comment carry the detail. Pure style preference; existing geotiff code mixes both conventions.

2. `xrspatial/geotiff/tests/test_cog_tile_size_hang_2311.py:88` -- `import warnings` inside the function body works but is non-idiomatic for this repo (the rest of the file imports at module scope). Move to module scope.

3. `xrspatial/geotiff/tests/test_cog_tile_size_hang_2311.py` -- no test covers `tile_size=None`, `tile_size=128.0`, or `tile_size=True`. The top-of-`_write` gate rejects all three; a one-row parametrization would close that gap.

### What looks good
- Two-layer fix: public-boundary rejection in `to_geotiff` plus inner-writer rejection in `_write`, with comments cross-referencing the issue.
- SIGALRM watchdog on every test so a regression that reintroduces the hang fails the test instead of locking up CI.
- Existing GPU writer path already runs `_validate_tile_size_arg` unconditionally; no GPU-side change needed.
- Sanity test confirms `cog=False, tiled=False, tile_size=-1` is still accepted (matches the "ignored" warning shape) so the fix does not overcorrect into breaking existing strip-layout callers.
- Test file message wording asserts both `tile_size` and `positive` substrings, which matches the existing `_validate_tile_size` convention from #1776/#2301.

### Checklist
- [x] Algorithm matches reference/paper (N/A -- validation-only)
- [x] All implemented backends produce consistent results (GPU path was already validated)
- [x] Edge cases are covered by tests (tile_size in {-1, 0}; minor gap on tile_size=None/float/bool)
- [x] Benchmark exists or is not needed (not needed -- validation)
- [x] README feature matrix updated (N/A -- no new public function)

brendancol (commented):
## Follow-up review (post-fixes)

### Disposition of first-pass findings
- Suggestion 1 (bool parity at `_writer.py:522`): **fixed** in commit a1767fa. Both defense layers now use the same `isinstance + bool + <=0` trinity.
- Suggestion 2 (inner-loop guard coverage): **fixed** via `test_inner_overview_loop_guard_message_is_pinned`. The constants-inspection approach pins the literal without requiring an invasive monkeypatch of the top gate. Reaching the inner guard through a real `_write` call requires either patching the top gate's inline `isinstance`/`bool`/`<=0` check (which has no helper to monkeypatch) or duplicating most of `_write` in the test. The constants pin catches the case the second-line defense exists to catch (someone deletes it), without that complexity.
- Nit 1 (long comment block): **dismissed**. The verbose comment matches the prevailing geotiff style (#2301, #2216, #1987 PR 3 all use ~10-line context comments on validation gates). Trimming would diverge from the surrounding code.
- Nit 2 (import scope): **fixed**. `import warnings` is now at module scope.
- Nit 3 (non-int tile_size rows): **fixed**. Added parametrized rows for `tile_size in {None, 128.0, True, False}`, all passing.

### Second-pass findings
None. The follow-up commit is focused and the test count is now 12 passing in 0.09s.

PR #2317, opened 2026-05-22
Reject zero-denominator RATIONAL/SRATIONAL tags (#2313)

### LINKED ISSUE #2313: Reject zero-denominator RATIONAL/SRATIONAL values in TIFF header reader
`_read_value` in `xrspatial/geotiff/_header.py` (around line 442) returns `0.0` when a RATIONAL or SRATIONAL value has a zero denominator. The SRATIONAL branch a few lines down does the same thing.

The rest of the geotiff reader fails closed on malformed input. See PR #2301 / commit f5fbad54 for the COG writer's actionable failure mode, and PR #2300 / commit 5e8bf9a5 which promoted the COG contract to stable. Quietly mapping a malformed rational to `0.0` is out of step with that. `XResolution` and `YResolution` can get silently altered, and the corrupted metadata then round-trips through the reader as if nothing happened.

## Expected

A zero-denominator RATIONAL/SRATIONAL raises `ValueError`, with the tag name and the denominator in the message.

## Actual

The value becomes `0.0`. Two existing tests (`TestReadValueRationals.test_rational_denominator_zero_returns_zero` and `test_srational_denominator_zero_returns_zero`) pin this as expected behaviour, so they need to be updated alongside the fix.

## Proposed fix

- Raise `ValueError` in `_read_value` for zero denominators on both the RATIONAL and SRATIONAL branches.
- Pass the tag through so the error message can name it (e.g. `XResolution`).
- Update the two existing tests to assert the new failure.
- Add a regression test that builds a TIFF with a zero-denominator `XResolution` rational and confirms the public reader fails with an actionable message.

## Acceptance

- Zero-denominator RATIONAL and SRATIONAL inputs raise `ValueError` naming the tag and the denominator.
- The failure surfaces at the public read entry points, not just the internal helper.
- A regression test covers both paths.

brendancol:
Closes #2313.

## Summary

- `_read_value` in `xrspatial/geotiff/_header.py` used to coerce a malformed RATIONAL or SRATIONAL value (denominator=0) to `0.0`. It now raises `ValueError` with the offending tag and denominator in the message, matching the actionable failure style from PR #2301.
- The tag id is threaded through the three call sites in `parse_ifd`, so the message can name `XResolution` / `YResolution` / `ResolutionUnit` directly when it's one of those.
- The two existing tests that pinned the old silent behaviour are flipped to assert the new failure, and a new regression test (`test_rational_zero_denominator_2313.py`) builds a TIFF whose `XResolution` rational has a zero denominator and confirms both the header parser and `open_geotiff` fail loudly.

## Backend coverage

Header decoding is shared across all four backends (numpy, cupy, dask+numpy, dask+cupy), so the change applies uniformly. No backend-specific code is touched.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_header.py`
- [x] `pytest xrspatial/geotiff/tests/test_rational_zero_denominator_2313.py`
- [x] `pytest xrspatial/geotiff/tests/ -k "not gpu and not cuda"` (4282 passed, 51 skipped)

### DIFF
--- a/xrspatial/geotiff/_header.py
+++ b/xrspatial/geotiff/_header.py
@@ -84,6 +84,26 @@ TAG_GEO_DOUBLE_PARAMS = 34736
 TAG_GEO_ASCII_PARAMS = 34737
 
 
+# Human-readable names for tags that turn up in error messages. Only
+# tags that legitimately use RATIONAL / SRATIONAL are interesting here;
+# anything else gets formatted as the bare numeric id.
+_TAG_NAMES = {
+    TAG_X_RESOLUTION: "XResolution",
+    TAG_Y_RESOLUTION: "YResolution",
+    TAG_RESOLUTION_UNIT: "ResolutionUnit",
+}
+
+
+def _tag_label(tag: int | None) -> str:
+    """Format a tag id for error messages."""
+    if tag is None:
+        return "<unknown>"
+    name = _TAG_NAMES.get(tag)
+    if name is not None:
+        return f"{name} (tag={tag})"
+    return f"tag={tag}"
+
+
 @dataclass
 class TIFFHeader:
     """Parsed TIFF file header."""
@@ -427,8 +447,15 @@ def parse_header(data: bytes | memoryview) -> TIFFHeader:
 
 
 def _read_value(data: bytes | memoryview, offset: int, type_id: int,
-                count: int, bo: str) -> Any:
-    """Read a typed value array from data at the given offset."""
+                count: int, bo: str, tag: int | None = None) -> Any:
+    """Read a typed value array from data at the given offset.
+
+    A zero-denominator RATIONAL or SRATIONAL is rejected with a
+    `ValueError` rather than silently coerced to 0.0. The TIFF spec
+    treats denominator-zero rationals as malformed, and quietly mapping
+    them to 0.0 lets corrupted `XResolution` / `YResolution` metadata
+    round-trip through the reader as if the file were valid.
+    """
     type_size = TIFF_TYPE_SIZES.get(type_id, 1)
 
     if type_id == ASCII:
@@ -445,7 +472,13 @@ def _read_value(data: bytes | memoryview, offset: int, type_id: int,
             off = offset + i * 8
             num = struct.unpack_from(f'{bo}I', data, off)[0]
             den = struct.unpack_from(f'{bo}I', data, off + 4)[0]
-            values.append(num / den if den != 0 else 0.0)
+            if den == 0:
+                raise ValueError(
+                    f"Malformed RATIONAL on {_tag_label(tag)}: "
+                    f"numerator={num} denominator={den} at element {i}; "
+                    f"refusing to parse possibly malformed TIFF"
+                )
+            values.append(num / den)
         return tuple(values) if count > 1 else values[0]
 
     if type_id == SRATIONAL:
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Reject zero-denominator RATIONAL/SRATIONAL tags (#2313)

### Blockers
None.

### Suggestions
- `test_header.py:261` and `test_header.py:272`: the two new tests do `from xrspatial.geotiff._header import TAG_X_RESOLUTION` (and Y) inside the method body. The module is already imported at the top of the file (`_read_value, parse_all_ifds, parse_header`); folding the tag constants into that same import keeps the block consistent and avoids the local import.

### Nits
- `_header.py:88-96`: `_TAG_NAMES` covers only X/YResolution and ResolutionUnit, which is the right scope for this PR. If a follow-up threads `tag=` through other error sites, growing this map will help. Flagging it so it doesn't get forgotten.
- `test_rational_zero_denominator_2313.py:147-208`: `test_yresolution_zero_denominator_named_in_error` reimplements most of `_build_tiff_with_rational_xres`. Adding a `which: int = 282` parameter to the helper (default XResolution, 283 for YResolution) would let the YResolution test reuse the helper instead of duplicating ~60 lines.

### What looks good
- Error message follows the "refusing to parse possibly malformed TIFF" pattern already used by `parse_ifd` (lines 666/672 in the original).
- All three `_read_value` call sites in `parse_ifd` get `tag=tag`, including the pre-scan path that swallows ValueError. That swallow is safe because `_DIMENSION_TAGS` doesn't include any RATIONAL tag.
- The regression test exercises both `parse_all_ifds` and `open_geotiff`, so the failure is verified at the public entry point, not just the internal helper.
- The two existing tests that pinned the old silent behaviour are flipped to assert the new failure, not deleted. A regression that reintroduces the silent path would fail loudly.
- BigTIFF coverage is implicit: `parse_ifd` uses the same `_read_value` for both classic and BigTIFF entries.
- Error message includes the element index, so a malformed entry inside a multi-element RATIONAL array gets pinpointed.

### Checklist
- [x] Algorithm matches spec (TIFF treats denom=0 as malformed)
- [x] NaN handling unchanged (rationals never produced NaN before either)
- [x] Edge cases covered (count>1 in `test_rational_denominator_zero_in_array_raises`)
- [x] Benchmark coverage: not needed (error path)
- [x] README feature matrix: not applicable (pure bug fix)
- [x] Docstrings present and accurate (`_read_value` docstring documents the new behaviour)

brendancol (commented):
## Follow-up review after ca86979c

Re-checked the two items that needed action and confirmed the new commit addresses them.

### Disposition

- Suggestion (hoist local imports in `test_header.py`): **fixed** in ca86979c. `TAG_X_RESOLUTION` and `TAG_Y_RESOLUTION` are now in the top-level import block, and the two local imports inside the test methods are gone.
- Nit (share helper between XResolution and YResolution tests): **fixed** in ca86979c. `_build_tiff_with_malformed_resolution` now takes a `which` parameter, and `test_yresolution_zero_denominator_named_in_error` is a one-liner that reuses the helper. Net `-41` lines in the test file.
- Nit (`_TAG_NAMES` map could grow): **deferred**. The map only exists to label tags in error messages produced by `_read_value`, and the only error sites that take a `tag=` kwarg today are the RATIONAL / SRATIONAL branches. Adding entries for tags that no error site references would be dead code. If a future PR threads `tag=` through other error sites, growing the map should be part of that PR.

Tests: `pytest xrspatial/geotiff/tests/test_header.py xrspatial/geotiff/tests/test_rational_zero_denominator_2313.py` -> 44 passed.

No new findings on this pass.

PR #2318, opened 2026-05-22
Reject cog=True with tiled=False at the writer boundary (#2312)

### LINKED ISSUE #2312: to_geotiff(cog=True, tiled=False) writes a non-COG strip TIFF
**Describe the bug**

`to_geotiff(..., cog=True, tiled=False)` returns successfully and writes a strip-layout TIFF. The COG spec requires a tiled internal layout, so the output is not a valid Cloud Optimized GeoTIFF even though the caller asked for one with `cog=True`.

The writer currently warns that `tile_size` is ignored under `tiled=False`, then writes strips via `xrspatial/geotiff/_writer.py:477`. The `cog=True` kwarg is silently dropped for the layout decision while the rest of the COG path (overviews, IFD ordering, etc.) still runs, so the file is a malformed hybrid.

This violates the stable COG contract that #2300 (commit 5e8bf9a5) promoted.

**To Reproduce**

```python
import numpy as np
import xarray as xr

da = xr.DataArray(np.zeros((256, 256), dtype=np.float32), dims=('y', 'x'))
to_geotiff(da, '/tmp/bad.tif', cog=True, tiled=False)
# Succeeds; produces a non-COG strip TIFF labelled as COG by intent.
```

**Expected behavior**

`cog=True` combined with `tiled=False` should raise `ValueError` at the public boundary with a message that names the violated constraint and tells the caller to either pass `tiled=True` or drop `cog=True`.

**Fix shape**

1. Reject the combination in `xrspatial/geotiff/_writers/eager.py` next to the existing validation block. Match the error-message style from PR #2301 / commit f5fbad54.
2. Defense in depth: raise from `xrspatial/geotiff/_writer.py` near line 477 if `cog and not tiled` ever reaches that path. This guards against future callers that bypass the public boundary.
3. Remove the now-dead "warn that tile_size is ignored when tiled=False under cog=True" branch. Under the new gate, `cog=True` implies `tiled=True`, so the warning would never fire on a `cog=True` call.
4. Add regression tests: one for the rejection (with message-substring matching, like the rows in `test_cog_invalid_input_errors_2286.py`), and one smoke test confirming the tiled COG path still works.

**Additional context**

Related PRs:
- #2300 (5e8bf9a5): promoted the COG contract to stable.
- #2301 (f5fbad54): "Pin actionable failure modes for unsupported COG writer inputs". Same shape of fix, same error-message style.

brendancol:
## Summary

Closes #2312.

- Reject `cog=True` combined with `tiled=False` at the `to_geotiff` public boundary with a `ValueError` that names the COG-spec constraint and both caller-side fixes (`tiled=True` or `cog=False`). Same shape as the COG input gates pinned in PR #2301 / commit f5fbad54.
- Add a defense-in-depth gate inside `_writer._write` so direct callers of the array-level entry point cannot bypass the public wrapper to produce the malformed file.
- Comment the now-dead `cog=True` arm of the "tile_size is ignored when tiled=False" warning. Under the new gate, `cog=True` always implies `tiled=True`, so the warning fires only on the `cog=False, tiled=False` path now.

The writer used to silently produce a non-COG strip TIFF for `cog=True, tiled=False`, violating the stable COG contract promoted in #2300.

## Backend coverage

Writer-side fix. Both the eager (`to_geotiff`) and array-level (`_write`) entry points gate the combination, which covers the numpy and dask CPU paths. The GPU writer (`write_geotiff_gpu`) already rejects `tiled=False` independently, so the four-backend matrix (numpy / cupy / dask+numpy / dask+cupy) is covered.

## Test plan

- [x] Public `to_geotiff(cog=True, tiled=False)` raises with the expected message tokens.
- [x] The `tile_size` kwarg with `cog=True, tiled=False` still raises and does not emit the misleading "tile_size is ignored" warning.
- [x] Low-level `_writer.write(cog=True, tiled=False)` also raises.
- [x] Valid tiled-COG path (`cog=True, tiled=True`) still round-trips.
- [x] `cog=True` with default `tiled` still works.
- [x] `cog=False, tiled=False` strip path still works.
- [x] Full `xrspatial/geotiff/tests/` suite: 5220 passed, 68 skipped locally.

### DIFF
--- a/xrspatial/geotiff/_writer.py
+++ b/xrspatial/geotiff/_writer.py
@@ -128,6 +128,28 @@ _TILE_POOL_THREAD_PREFIX = 'xrspatial-geotiff-tile-compress'
 
 
 # ---------------------------------------------------------------------------
+# Shared error messages
+# ---------------------------------------------------------------------------
+
+# Issue #2312: a single source of truth for the ``cog=True, tiled=False``
+# rejection message used by both the public ``to_geotiff`` boundary and
+# the array-level ``_write`` defense-in-depth gate. Keeping the message
+# string in one place stops the two raise sites from drifting if one
+# ever gets reworded. The substring assertions in
+# ``test_cog_requires_tiled_2312.py`` pin the actionable tokens
+# (``tiled=True``, ``cog=False``, ``COG``) so a future rewrite still
+# has to satisfy the same contract.
+_COG_REQUIRES_TILED_MSG = (
+    "cog=True requires tiled=True: the COG specification "
+    "mandates a tiled internal layout, so a strip-layout file "
+    "cannot be a valid Cloud Optimized GeoTIFF. Pass tiled=True "
+    "(or omit tiled, which defaults to True) to write a COG, or "
+    "set cog=False to write a non-COG strip TIFF. See issue "
+    "#2312."
+)
+
+
+# ---------------------------------------------------------------------------
 # Array-level write entry points (module-private; see module docstring)
 # ---------------------------------------------------------------------------
 
@@ -480,14 +502,7 @@ def _write(data: np.ndarray, path: str, *,
     # a strip-layout body, producing a malformed file that claims to be
     # a COG.
     if cog and not tiled:
-        raise ValueError(
-            "cog=True requires tiled=True: the COG specification "
-            "mandates a tiled internal layout, so a strip-layout file "
-            "cannot be a valid Cloud Optimized GeoTIFF. Pass tiled=True "
-            "(or omit tiled, which defaults to True) to write a COG, or "
-            "set cog=False to write a non-COG strip TIFF. See issue "
-            "#2312."
-        )
+        raise ValueError(_COG_REQUIRES_TILED_MSG)
 
     # Build pixel data parts
     parts = []
--- a/xrspatial/geotiff/_writer.py
+++ b/xrspatial/geotiff/_writer.py
@@ -470,6 +470,25 @@ def _write(data: np.ndarray, path: str, *,
         if nodata is not None:
             nodata = _invert_nodata_for_miniswhite(nodata, data.dtype)
 
+    # Issue #2312: defense-in-depth gate for ``cog=True, tiled=False``.
+    # The public ``to_geotiff`` wrapper rejects this combination at its
+    # own boundary, so this branch is unreachable when the wrapper is
+    # the caller; the gate matters for direct callers of ``_write`` and
+    # for any future caller (test harness, internal tool) that bypasses
+    # the wrapper. Without it, ``_write_stripped`` would run below and
+    # the overview-pyramid block at line ~490 would attach overviews to
+    # a strip-layout body, producing a malformed file that claims to be
+    # a COG.
+    if cog and not tiled:
+        raise ValueError(
+            "cog=True requires tiled=True: the COG specification "
+            "mandates a tiled internal layout, so a strip-layout file "
+            "cannot be a valid Cloud Optimized GeoTIFF. Pass tiled=True "
+            "(or omit tiled, which defaults to True) to write a COG, or "
+            "set cog=False to write a non-COG strip TIFF. See issue "
+            "#2312."
+        )
+
     # Build pixel data parts
     parts = []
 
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Reject cog=True with tiled=False at the writer boundary (#2312)

None.

- [ ] `xrspatial/geotiff/_writers/eager.py` lines 143-150 (`tiled` docstring entry): the new rejection is invisible to a reader scanning only that entry. Add a sentence saying `tiled=False` is incompatible with `cog=True` and raises `ValueError`. Symmetric with the `tile_size` entry that already documents the strip-mode warning.
- [ ] `xrspatial/geotiff/_writers/eager.py` lines 159-163 (`cog` docstring entry): mirror the same note on the `cog` side. A caller reading "how do I write a COG" lands here first, so the `tiled=True` requirement should appear here too.

- [ ] `xrspatial/geotiff/_writer.py` line 481 and `xrspatial/geotiff/_writers/eager.py` line 569: the two `ValueError` message strings are byte-identical. If one gets reworded, the other will drift. Pulling the message into a module-level constant (e.g. `_COG_REQUIRES_TILED_MSG`) would keep them in sync. The tests already pin substrings on both sides, so the drift risk is low; the constant is just an explicit invariant.
- [ ] `xrspatial/geotiff/_writer.py` line 473: the defense-in-depth gate is inline rather than inside `_validate_lowlevel_write_kwargs` (line 135), where the other push-down byte-affecting checks live. Moving it would broaden that helper's signature with `tiled` and `cog`, which may not be worth it. Flagging only for consideration; the inline gate is fine.

### What looks good

- The error message is actionable: it names the violated constraint (COG spec requires tiled layout), the two fixes the caller can apply (`tiled=True` or `cog=False`), and links to the issue. Same shape as the COG gates pinned in PR #2301 / commit f5fbad54.
- The defense-in-depth gate in `_writer._write` is at the right depth. It fires before `_write_stripped` runs and before the overview-pyramid block, so a direct caller cannot produce the malformed strip-plus-overviews file by bypassing `to_geotiff`.
- Test file is well-structured: separate rows for the public rejection, the low-level rejection, the "tile_size warning must not fire" pin, both happy-path smoke tests (explicit and default `tiled`), and a negative control for the strip-without-COG path.
- The dead "tile_size is ignored" warning branch is documented in place rather than deleted. The branch still serves the `cog=False, tiled=False, tile_size != 256` case, so the comment correctly scopes the change to the `cog=True` arm.
- Temp file names carry the `_2312` issue suffix, consistent with the project convention.

### Checklist

- [x] Algorithm matches reference/paper: COG spec requires tiled layout.
- [x] All implemented backends produce consistent results: the eager and array-level entry points both gate the combination; dask streaming falls through to `_write` (which gates again); the GPU path already rejected `tiled=False` independently.
- [x] NaN handling: not applicable (validation-only change).
- [x] Edge cases covered by tests: explicit `tiled=False`, with `tile_size`, default `tiled`, strip-without-COG control.
- [x] Dask chunk boundaries handled correctly: not applicable (no dask compute paths changed).
- [x] No premature materialization or unnecessary copies: not applicable.
- [x] Benchmark exists or is not needed: not needed.
- [x] README feature matrix updated: not applicable (no new function or backend change).
- [x] Docstrings present: the test file docstring documents the new ValueError; the suggestion above is to also mention the constraint on the `tiled` / `cog` parameter entries in the public docstring.

brendancol (commented):
## PR Review (follow-up): Reject cog=True with tiled=False at the writer boundary (#2312)

Re-review after commit dfc617f6.

### Status of original findings

Suggestions (both fixed):
- [x] `tiled` docstring entry now says `tiled=False` is incompatible with `cog=True` and raises `ValueError`.
- [x] `cog` docstring entry now says the parameter requires `tiled=True`.

Nits:
- [x] Fixed: shared error message extracted to `_COG_REQUIRES_TILED_MSG` in `_writer.py`, imported into `_writers/eager.py`. Both raise sites use the constant, so a future reword cannot make them drift.
- [ ] Dismissed: moving the defense-in-depth gate into `_validate_lowlevel_write_kwargs` would require adding `tiled` and `cog` parameters to that helper. Its existing signature only carries kwargs whose validation is independent of the layout decision (compression, max_z_error, crs_*, allow_unparseable_crs). Two more parameters for one check is wider than the consolidation is worth. The inline gate is fine and the comment at the call site explains why it lives there.

### New checks

No new findings. The 6 tests in `test_cog_requires_tiled_2312.py` still pass after the follow-up commit; the wider `test_cog_invalid_input_errors_2286.py` suite (19 tests) is unaffected.

### What looks good (continued)

The shared constant lives in `_writer.py` (the array-level module), which is the lower of the two modules that raise the error. Both raise sites import from a single source of truth. The docstring updates are parameter-scoped, so the rest of the long `to_geotiff` docstring did not shift, and the diff stays narrow.

PR #2319, opened 2026-05-22
Use sidecar header for remote .ovr decode byte order (#2314)

### LINKED ISSUE #2314: Remote sidecar overview reads decode with base TIFF byte order
## Bug description

`_parse_cog_http_meta` in `xrspatial/geotiff/_cog_http.py` finds a sibling `.tif.ovr` sidecar and merges its IFDs onto the pyramid list, but the function still returns the base file's `TIFFHeader` when the selected IFD lives in the sidecar (`used_sidecar=True`).

The HTTP eager path swaps the byte source to the sidecar URL (around line 329), but tile and strip decoding still passes `header.byte_order` from the base file when calling `_decode_strip_or_tile`. The dask path has the same bug: `http_meta = (http_header, http_ifd)` references the base header even though the surrounding comment claims it carries the sidecar header.

A big-endian `.ovr` sidecar paired with a little-endian base TIFF (or vice versa) decodes pixel bytes against the wrong byte order and returns garbage values.

## How to reproduce

1. Build a little-endian base TIFF whose external overviews live in a big-endian `.tif.ovr` sidecar (the byte orders deliberately differ).
2. Read an overview level that resolves into the sidecar via HTTP or via the dask HTTP path:

```python
arr = open_geotiff("https://example.com/mixed_endian.tif", overview_level=1)
```

3. The returned pixel values are scrambled because the decode step interprets sidecar bytes using the base file's byte order.

## Expected behaviour

When a sidecar is in use, the decode path should use the sidecar's own `TIFFHeader` (in particular `byte_order`) so the bytes are interpreted with the byte order of the file they came from. The local sidecar path in `xrspatial/geotiff/_reader.py:223` already does this via `sidecar_origin.get(id(ifd), (data, header))`. The HTTP and dask paths should reach parity.

## Required fix

- In `_parse_cog_http_meta` (around line 232), return the sidecar's header when `used_sidecar=True`. Update the docstring's return-tuple contract to match.
- In the HTTP eager decode path (`_read_cog_http`, around line 597 / the `_fetch_decode_cog_http_tiles` call site), thread the correct header into the decode step so `header.byte_order` matches the bytes being decoded.
- In the dask path (`xrspatial/geotiff/_backends/dask.py` around line 244), set `http_meta = (sidecar.header, http_ifd)` when sidecar overviews are present so per-chunk decoding sees the correct byte order.
- Add a regression test for a mixed-endian case: base TIFF with one byte order and a `.ovr` sidecar with the opposite byte order. The test should fail before the fix and pass after.

## Scope

- Files: `xrspatial/geotiff/_cog_http.py`, `xrspatial/geotiff/_backends/dask.py`
- Backends: numpy and dask+numpy HTTP / fsspec reads. The cupy backends sit downstream of these paths and inherit the fix.

brendancol:
Closes #2314.

## Summary

- `_parse_cog_http_meta` now returns the sidecar's `TIFFHeader` when the selected overview IFD came from the `.tif.ovr`. The eager HTTP path and the dask HTTP path already feed the unpacked `header.byte_order` into `_decode_strip_or_tile`, so the swap is enough to make a big-endian sidecar paired with a little-endian base (or the reverse) decode against the right endianness.
- The local sidecar path in `_reader.py:223` already did this; the remote paths now reach parity.
- Adds `test_remote_sidecar_byte_order_2314.py` covering the eager HTTP, dask HTTP, fsspec memory, and direct `_parse_cog_http_meta` contracts for both byte-order directions. Eight of ten test cases fail on the unfixed branch and pass after the change.

## Backend coverage

- numpy HTTP eager (`_read_cog_http`)
- dask+numpy HTTP (`_backends/dask.py`)
- dask+numpy fsspec (same dask path, exercised via `memory://`)
- cupy / dask+cupy paths inherit the fix because they share the same metadata parse

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_remote_sidecar_byte_order_2314.py` (10 tests pass after fix, 8 fail before)
- [x] `pytest xrspatial/geotiff/tests/` (5224 pass, 68 skipped)
- [x] Existing `test_remote_sidecar_chunked_2239.py` still passes (parity with same-endian sidecar fixture)

### DIFF
--- a/xrspatial/geotiff/_backends/dask.py
+++ b/xrspatial/geotiff/_backends/dask.py
@@ -261,6 +261,12 @@ def read_geotiff_dask(source: str, *,
             if sidecar is not None:
                 from .._sidecar import close_sidecar
                 close_sidecar(sidecar)
+        # ``http_header`` carries the sidecar's ``TIFFHeader`` when
+        # ``used_sidecar`` was True (``_parse_cog_http_meta`` swaps it
+        # so ``byte_order`` matches the file the per-chunk range GETs
+        # land on). Pass that through to the per-chunk decode step so a
+        # mixed-endian base / ``.ovr`` pair decodes against the right
+        # endianness. Issue #2314.
         http_meta = (http_header, http_ifd)
         if http_ifd.orientation != 1:
             raise ValueError(
--- a/xrspatial/geotiff/_cog_http.py
+++ b/xrspatial/geotiff/_cog_http.py
@@ -262,6 +262,13 @@ def _parse_cog_http_meta(
     # at ``_decode_strip_or_tile``. Mirrors the local sidecar path in
     # ``_reader.py:223`` which swaps to the sidecar header for the same
     # reason. Issue #2314.
+    #
+    # ``used_sidecar`` can only be True when ``sidecar`` is not None:
+    # ``sidecar_ifd_ids`` is populated by ``discover_remote_sidecar``
+    # only on the same branch that assigns ``sidecar`` (and stays empty
+    # otherwise), so ``id(ifd) in sidecar_ifd_ids`` implies the sidecar
+    # was loaded successfully. The branch below relies on that
+    # invariant when it reads ``sidecar.header``.
     return_header = sidecar.header if used_sidecar else header
     if return_sidecar:
         route_path = sidecar.path if used_sidecar else source_path
--- a/xrspatial/geotiff/_cog_http.py
+++ b/xrspatial/geotiff/_cog_http.py
@@ -147,6 +147,18 @@ def _parse_cog_http_meta(
         ``sidecar`` and must close it; ``route_path`` is the URL/URI
         that per-chunk fetches should target; ``used_sidecar`` is
         ``True`` iff the selected IFD came from the sidecar.
+
+        ``header`` is the :class:`TIFFHeader` of the file the chosen
+        IFD lives in: the sidecar's header when ``used_sidecar=True``,
+        otherwise the base file's header. Callers that decode pixel
+        bytes (the eager HTTP path, the dask chunk reader) MUST use
+        this returned header for ``byte_order`` so the decode step
+        interprets the right endianness; a big-endian ``.ovr`` paired
+        with a little-endian base file would otherwise scramble the
+        result. ``geo_info`` is still extracted from the base file's
+        ``header_bytes`` (sidecar IFDs typically carry no geokeys and
+        inherit from the level-0 IFD that sits in the base buffer);
+        that parse is unaffected by the swap. Issue #2314.
     """
     if return_sidecar and source_path is None:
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Use sidecar header for remote .ovr decode byte order (#2314)

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/geotiff/tests/test_remote_sidecar_byte_order_2314.py`: the fixture builder relies on tifffile's default layout, which is stripped (untiled). The eager and dask HTTP paths both flow through `_fetch_decode_cog_http_strips` for these fixtures, so the tiled decode path's `header.byte_order` plumbing is exercised only transitively. The fix lives at the metadata layer so the result is the same, but a second parametrized fixture with `tile=...` would close the last coverage gap. Optional.
- `xrspatial/geotiff/_cog_http.py:265`: `return_header = sidecar.header if used_sidecar else header` is safe because `used_sidecar` can only be True when `sidecar` is not None (`sidecar_ifd_ids` stays empty otherwise). A defensive comment tying those two invariants together would make the relationship obvious to a future reader. Optional.

### What looks good
- The fix is minimum-surface: one line in `_parse_cog_http_meta` returns the sidecar's `TIFFHeader` when `used_sidecar` is True. Downstream code in `_read_cog_http` and `_backends/dask.py` already reads `header.byte_order` off the unpacked header and picks up the right value with no further plumbing.
- Docstring updated to describe the new contract.
- Comment in `dask.py` updated so the code matches what the comment claims.
- Tests fail on 8 of 10 cases without the fix (parametrized LE/BE and BE/LE across eager HTTP, dask HTTP, fsspec memory, and a direct contract test on `_parse_cog_http_meta`). The two passing-before cases are the local sidecar path, included as a parity sanity check.
- Test fixtures use uuid-suffixed paths so parallel runs do not collide.
- No public API change; no docs / README update needed.

### Checklist
- [x] Algorithm matches reference behaviour (parity with `_reader.py:223` local sidecar swap)
- [x] All implemented backends produce consistent results (numpy and dask HTTP both exercised)
- [x] NaN handling is correct (untouched; uint16 fixtures have no NaN)
- [x] Edge cases are covered by tests (both byte-order directions)
- [x] Dask chunk boundaries handled correctly (`chunks=16` forces multi-chunk read)
- [x] Benchmark exists or is not needed (pure correctness fix)

brendancol (commented):
## PR Review (follow-up after c1dbfb98)

Both nits from the previous review have been addressed:

- `_cog_http.py`: invariant comment added above the `sidecar.header` read, explaining why `used_sidecar=True` implies `sidecar is not None`.
- `test_remote_sidecar_byte_order_2314.py`: new parametrized `test_http_eager_mixed_endian_sidecar_tiled` exercises the tiled HTTP decode path directly. Test count is now 12 (was 10), all passing.

No remaining blockers, suggestions, or nits. Ready for CI to land.

PR #2320, opened 2026-05-22
Route sidecar georef tags through the sidecar's bytes (#2315)

### LINKED ISSUE #2315: Sidecar georeferencing extraction ignores sidecar's own geokeys
## Description

When a `.tif.ovr` sidecar is loaded, local eager reads correctly swap the pixel buffer and TIFF header over to the sidecar (`xrspatial/geotiff/_reader.py` ~line 223). The georef extraction call right below it still passes the base file's `data` and `header.byte_order` (line 234). `_read_geo_info` in `xrspatial/geotiff/__init__.py` (~line 291) has the same bug.

This is harmless under the usual GDAL convention, where the sidecar has no geokeys and georef lives on the base file. It is not safe in general: if the sidecar IFD does carry its own GeoKeyDirectory or model tags, those tags get parsed against the wrong byte buffer and the returned `geo_info` is silently wrong.

## Expected behavior

If the selected IFD lives in the sidecar and that IFD declares its own georef payload, parse the georef against the sidecar's bytes and byte order. If the sidecar IFD has no georef payload, fall back to the existing behavior and inherit from the base file's level-0 IFD.

## Actual behavior

Georef tags on a sidecar IFD are parsed as if they lived in the base file. ModelPixelScale, ModelTiepoint, and ModelTransformation values come out of the wrong offsets, and GeoKeyDirectory entries that point into GeoDoubleParamsTag or GeoAsciiParamsTag dereference into the base file instead of the sidecar.

## Fix shape

- Detect whether a sidecar IFD carries its own georef payload (any of tags 33550, 33922, 34264, 34735).
- When it does, run `extract_geo_info_with_overview_inheritance` against the sidecar's `data` and `header.byte_order` for that IFD.
- When it does not, keep today's behavior.
- Apply the same rule at both call sites so the two paths agree.
- Add a regression test using a synthetic base + sidecar pair where the sidecar declares its own ModelPixelScale, ModelTiepoint, and GeoKeyDirectory differing from the base.

## Files

- `xrspatial/geotiff/_reader.py:234` (eager read path)
- `xrspatial/geotiff/__init__.py:291` (metadata-only path used by the dask backend)

brendancol:
## Summary

Closes #2315.

- Add an optional `sidecar_origin` mapping to `extract_geo_info_with_overview_inheritance`. When a sidecar IFD declares its own GeoKeyDirectory, ModelPixelScale, ModelTiepoint, or ModelTransformation, the helper parses those tags against the sidecar's bytes and byte order instead of the base file's.
- Wire the mapping through both call sites (`_reader.py` and `_read_geo_info` in `__init__.py`) so the eager and metadata-only paths agree.
- Conservative gate: a sidecar IFD with no georef tags (the GDAL convention) keeps today's behavior and inherits from the base file's level-0 IFD.

## Backend coverage

Read-side only. The fix applies to every read backend that shares the metadata parse: numpy eager, dask+numpy via `_read_geo_info`, and cupy / dask+cupy which call back into the same metadata helper. No GPU or dask-specific code changes.

## Test plan

- [x] New regression test `test_sidecar_own_geokeys_2315.py` covers both branches:
  - Sidecar with its own ModelPixelScale, ModelTiepoint, and GeoKeyDirectory: the sidecar's georef wins.
  - Sidecar with georef tags stripped (GDAL convention): the base file's georef is inherited.
- [x] Both `open_geotiff` (eager) and `_read_geo_info` (metadata-only / dask graph builder) exercised.
- [x] Existing `test_sidecar_ovr_2112.py` suite still passes (28 tests).
- [x] Wider sidecar / overview / geo_info selection (648 tests) passes.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -273,24 +273,38 @@ def _read_geo_info(source, *, overview_level: int | None = None,
         # so ``overview_level`` indexes both internal and external
         # overviews (issue #2112). Local file paths only.
         from ._sidecar import attach_sidecar_origin, find_sidecar, load_sidecar
+        sidecar_origin: dict[int, tuple] = {}
         sidecar_path = find_sidecar(source)
         if sidecar_path is not None:
             sidecar = load_sidecar(sidecar_path)
-            # Metadata-only path: drop the origin mapping. The reader
-            # only needs the merged IFD list to resolve the requested
-            # ``overview_level`` against; strip/tile bytes are sliced by
-            # ``read_to_array`` on the actual read.
-            attach_sidecar_origin(
+            # The origin mapping is consumed below for georef extraction
+            # only -- strip/tile bytes are sliced by ``read_to_array`` on
+            # the actual read. A sidecar IFD that carries its own
+            # GeoKeyDirectory / ModelPixelScale / ModelTiepoint /
+            # ModelTransformation needs the sidecar's byte order to
+            # parse cleanly; without the mapping the helper falls back
+            # to the base file's bytes (today's default, correct under
+            # the usual GDAL convention). See issue #2315.
+            sidecar_origin = attach_sidecar_origin(
                 sidecar.ifds, sidecar.data, sidecar.header)
             ifds = ifds + sidecar.ifds
         ifd = select_overview_ifd(ifds, overview_level)
         # Inherit georef from the level-0 IFD when the overview itself
         # has no geokeys (issue #1640). Pass-through for level 0. The
         # sidecar IFDs typically lack geokeys so the inheritance pulls
-        # from the base file's full-resolution IFD as GDAL does.
+        # from the base file's full-resolution IFD as GDAL does. When a
+        # sidecar IFD does declare its own georef payload, ``georef_origin``
+        # routes the parse to the sidecar's bytes / byte order so the
+        # sidecar's georef wins. See issue #2315.
+        georef_origin = (
+            {iid: (od, oh.byte_order)
+             for iid, (od, oh) in sidecar_origin.items()}
+            if sidecar_origin else None
+        )
         geo_info = extract_geo_info_with_overview_inheritance(
             ifd, ifds, data, header.byte_order,
-            allow_rotated=allow_rotated)
+            allow_rotated=allow_rotated,
+            sidecar_origin=georef_origin)
         bps = resolve_bits_per_sample(ifd.bits_per_sample)
         file_dtype = tiff_dtype_to_numpy(bps, ifd.sample_format)
         _validate_predictor_sample_format(ifd.predictor, ifd.sample_format)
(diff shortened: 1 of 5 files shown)


brendancol (commented):
## PR Review: Route sidecar georef tags through the sidecar's bytes (#2315)

### Blockers
- None.

### Suggestions
- [ ] `xrspatial/geotiff/_cog_http.py:242` -- the HTTP / fsspec metadata path calls `extract_geo_info_with_overview_inheritance(ifd, ifds, header_bytes, header.byte_order, ...)` without the new `sidecar_origin` kwarg. A sidecar IFD over HTTP that declares its own georef payload still parses against the base file's bytes here. A sibling rockout is reportedly handling the HTTP / byte-order side of this finding; if that is the plan, leave a cross-reference comment in the source so future maintainers see the symmetry argument. Otherwise the one-line fix is to thread the sidecar's IFD ids and `(sidecar.data, sidecar.header.byte_order)` through `discover_remote_sidecar`'s return into a mapping and pass it through.

### Nits
- [ ] `xrspatial/geotiff/_geotags.py` `extract_geo_info_with_overview_inheritance` body -- the inheritance branch (~line 1174) still calls `extract_geo_info(base_ifd, data, byte_order, ...)`. If `base_ifd` somehow comes from the sidecar (uncommon: the base file would need to have no full-res IFD of its own), the inherited georef parses against the wrong buffer. The new `sidecar_origin` map is the natural place to fix it -- the same lookup you do for `ifd` would work for `base_ifd`. Low priority, but a 3-line guard is cheap.

- [ ] `xrspatial/geotiff/tests/test_sidecar_own_geokeys_2315.py` `_mark_first_ifd_as_overview` assumes the new NewSubfileType entry (tag 254) sorts before every existing tag in the IFD, so it inserts at position 0. True for `to_geotiff` output today (smallest tag emitted is 256 / ImageWidth). Brittle if a future writer change adds a tag with id <= 254. Worth an `assert` that the first existing tag id is > 254, or a `bisect`-style insert.

### What looks good
- The fix is conservatively gated on `_ifd_has_georef_payload(ifd)`. The GDAL-convention case (sidecar with no geokeys) is unchanged.
- Both eager and metadata-only call sites are updated symmetrically.
- The regression test exercises both branches: sidecar wins when it has its own geokeys, base inherits when the sidecar has none.
- Signature change is backward compatible (new kwarg defaults to `None`).

### Checklist
- [x] NaN handling: not applicable (metadata path, no array ops)
- [x] Edge cases covered (sidecar with vs without geokeys)
- [x] Dask: the `_read_geo_info` path used by the dask graph builder is exercised
- [x] No premature materialization (metadata-only changes)
- [x] Benchmark coverage: not applicable
- [x] README feature matrix: not applicable (bug fix, no new function)

brendancol (commented):
## Follow-up review

Disposition of the round-1 findings:

- Suggestion (`_cog_http.py:242`, HTTP sidecar): **deferred** to the sibling HTTP / dask sidecar-byte-order rockout that already owns this surface. Added a cross-reference comment in the source so the symmetry argument is visible to future maintainers.
- Nit (base_ifd inheritance): **fixed** in commit 9d8a3e3. The inheritance branch now applies the same `sidecar_origin` lookup to `base_ifd` that the main branch applies to `ifd`.
- Nit (test fixture insert assumption): **fixed** in commit 9d8a3e3. `_mark_first_ifd_as_overview` now asserts the first existing IFD tag id is > 254 before inserting NewSubfileType at position 0, so a future writer change that emits a tag <= 254 fails loudly instead of silently producing an out-of-order IFD.

All 34 sidecar-focused tests still pass after the follow-up.

PR #2325, opened 2026-05-23
Wrap eager source lifetime around read_all() (#2322)

### LINKED ISSUE #2322: Eager source cleanup is not exception-safe around read_all() in geotiff reader
**Describe the bug**

In `xrspatial/geotiff/_reader.py`, `_read_to_array` constructs a source
(`_CloudSource`, `_FileSource`, or `_BytesIOSource`) at the end of the
local-or-cloud branch, then calls `data = src.read_all()` on line 187
*before* entering the `try`/`finally` block on line 190 that calls
`src.close()`.

If `read_all()` raises mid-read (a network blip on a fsspec source, a
transient S3 error, a local I/O failure), the exception propagates out
of the function and `src.close()` is never called.

Right now `_CloudSource.close()` is a no-op, so nothing actually leaks.
That makes this latent rather than urgent, but it's fragile. As soon as
the cloud source grows real state (pinned credentials, persistent
fsspec sessions, cached handles, retry counters), every failed read
leaks it.

**Expected behavior**

`src.close()` runs even when `read_all()` raises.

**Reproduction sketch**

Build a fake source whose `read_all()` raises, call `_read_to_array`
on it, and check that `src.close()` was called. Today it isn't.

**Fix direction**

Move the `try`/`finally` to start right after `src` is constructed and
put `read_all()` inside the protected region.

**Additional context**

The `CloudSizeLimitError` paths a few lines above already call
`src.close()` before raising. The eager-read path should inherit the
same guarantee structurally rather than relying on cleanup being a
no-op.

brendancol:
Closes #2322.

## Summary

- `_read_to_array` used to call `src.read_all()` BEFORE entering the
  `try/finally` block that calls `src.close()`. Any exception from
  `read_all()` (fsspec network failure, transient S3 error, local I/O
  failure) would skip the cleanup.
- The fix moves the `try` to start right after `src` is constructed
  and pulls `src.read_all()` inside the protected region. Mirrors the
  close-on-error contract `_read_cog_http` already enforces (#1816).
- `_CloudSource.close()` is a no-op today, so this is a structural
  guard rather than a fix for an observable leak. The point is to keep
  the cleanup intact before any future resource-holding source
  (pinned credentials, persistent fsspec sessions, cached file
  handles) gets added.

## Backend coverage

N/A. Pure correctness fix in the eager-read entry point. No backend
dispatch involved.

## Test plan

- [x] New regression tests in
      `xrspatial/geotiff/tests/test_eager_source_close_on_error_2322.py`
      cover the `_FileSource`, `_BytesIOSource`, and `_CloudSource`
      branches by injecting fake sources whose `read_all()` raises
      and asserting `close()` was still called.
- [x] Confirmed the new tests fail without the fix and pass with it.
- [x] Existing reader/source tests still pass
      (`test_bytesio_source`, `test_cog_http_close_on_error_1816`,
      `test_cloud_read_byte_limit_1928`,
      `test_open_geotiff_missing_sources_1810` — 40 tests).

### DIFF
--- a/xrspatial/geotiff/_reader.py
+++ b/xrspatial/geotiff/_reader.py
@@ -184,10 +184,18 @@ def _read_to_array(source, *, window=None, overview_level: int | None = None,
                     f"read.")
     else:
         src = _FileSource(source)
-    data = src.read_all()
 
     sidecar = None
+    # Wrap source lifetime in the try/finally immediately after
+    # construction so ``src.close()`` runs even when ``read_all()``
+    # raises (e.g. a fsspec network failure mid-download, a transient
+    # S3 error, or a local I/O error). ``_CloudSource.close()`` is a
+    # no-op today, but the structural guard prevents a future
+    # resource-holding source from leaking state on the failure path.
+    # Mirrors the close-on-error contract that ``_read_cog_http``
+    # already enforces (issue #1816). See issue #2322.
     try:
+        data = src.read_all()
         header = parse_header(data)
         ifds = parse_all_ifds(data, header)
 
--- a/xrspatial/geotiff/tests/test_eager_source_close_on_error_2322.py
+++ b/xrspatial/geotiff/tests/test_eager_source_close_on_error_2322.py
@@ -1,0 +1,156 @@
+"""Regression tests for issue #2322.
+
+``_read_to_array`` constructed a source (``_FileSource``,
+``_BytesIOSource``, or ``_CloudSource``) and immediately called
+``src.read_all()`` BEFORE entering the ``try/finally`` block that
+calls ``src.close()``. If ``read_all()`` raised mid-read (a fsspec
+network failure, a transient S3 error, a local I/O error), the
+exception propagated up and ``src.close()`` was never called.
+
+``_CloudSource.close()`` is a no-op today, so the leak was latent
+rather than active, but the structural guard is needed so a future
+resource-holding source (pinned credentials, persistent fsspec
+sessions, cached file handles) does not silently leak on the failure
+path.
+
+These tests pin the contract by injecting fake source objects whose
+``read_all()`` raises, then asserting ``close()`` was still called.
+"""
+from __future__ import annotations
+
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Wrap eager source lifetime around read_all() (#2322)

None.

None.

- `xrspatial/geotiff/_reader.py:160` — `_CloudSource(source)` itself can
  raise (e.g. an fsspec failure when fetching `size`), and that raise
  still happens outside the new try/finally. Today the partially
  constructed `_CloudSource` has no state to clean up, but the same
  reasoning that motivates this PR applies: a future stateful
  constructor would leak. Out of scope for this PR (the constructor
  guard is a separate concern), but worth noting in a follow-up.

### What looks good
- The fix is the minimal correct one: `try` now starts immediately
  after `src` is constructed, with `src.read_all()` inside.
  `finally: src.close()` already existed and now covers `read_all()`
  too.
- Comment block above the `try` explains the why, links to issue
  #1816 (the analogous `_read_cog_http` fix), and references #2322.
- The pre-existing `CloudSizeLimitError` paths at lines 167 and 175
  still call `src.close()` before raising, outside the new try.
  That's fine — they already had their own cleanup. No double-close
  risk because they raise immediately after.
- Tests cover all three source types (`_FileSource`, `_BytesIOSource`,
  `_CloudSource`) by patching the constructor to return a fake whose
  `read_all()` raises. The test file follows the same pattern as
  `test_cog_http_close_on_error_1816.py`.
- I verified the tests fail without the production change and pass
  with it, so they actually pin the contract.
- Temp filenames include the issue number (`tmp_2322_cleanup_file.tif`).

### Checklist
- [x] Algorithm matches reference/paper — N/A (cleanup correctness fix)
- [x] All implemented backends produce consistent results — N/A
- [x] Edge cases are covered by tests — all three source-type branches
- [x] README feature matrix updated — N/A
- [x] Docstrings present and accurate — module docstring on test file

Issue #1846, opened 2026-05-14
Pin validated IP into HTTP source to close DNS-rebinding TOCTOU


brendancol:
Hardening follow-up to #1664. No known exploit in the wild.

`_validate_http_url` calls `socket.getaddrinfo` and rejects the URL if any returned IP is private, loopback, link-local, or multicast. That runs at `_HTTPSource.__init__` and on every redirect's `Location`.

Then `_HTTPSource._request` calls `pool.request('GET', current_url, ...)`. urllib3 resolves the hostname a second time at connect time. The two resolutions are independent. Between them, a hostile DNS server can return a public IP first (validation passes) and a private IP second (TCP connection). The existing code comment labels this "best-effort" for exactly this reason.

Concrete scenarios:
- DNS rebinding: attacker controls a domain, hands 8.8.8.8 to the validator, then 127.0.0.1 (or 169.254.169.254) to the connect call.
- DNS cache races: low-TTL record flips between validate-time and connect-time.

Fix:
- `_validate_http_url` returns the first validated public IP instead of `None`.
- `_HTTPSource` opens the TCP socket to that IP literal while the HTTP `Host` header and TLS SNI stay set to the original hostname (so cert verification still works against the name).
- urllib3's `HTTPConnection._dns_host` is the right hook. `_new_conn` uses `_dns_host` for `socket.create_connection`, while `self.host` and `self.server_hostname` keep the original name for Host header and SNI.
- Each redirect hop re-resolves and re-pins independently. A redirect to a new hostname gets its own fresh validate-and-pin pair.

Residual limitations:
- Within a single hop the IP is pinned. Across redirects, each hop is resolved independently.
- An attacker who controls a hostname that legitimately resolves to multiple public IPs can still influence which one we pick. They can't bend us to a private IP.
- `XRSPATIAL_GEOTIFF_ALLOW_PRIVATE_HOSTS=1` still skips the pin (and the validation) for dev/test against localhost servers.

Refs #1664.

PR #2326, opened 2026-05-23
Reject uppercase HTTP(S) schemes in geotiff URL dispatch (#2323)

### LINKED ISSUE #2323: HTTP(S) SSRF hardening bypassable with uppercase schemes
## Description

The geotiff reader routes URLs to one of three sources using a case-sensitive prefix match. URL schemes are case-insensitive per RFC 3986, so an uppercase `HTTP://` or `HTTPS://` URL slips past `_HTTPSource` (which carries the SSRF allow-list and pinned DNS from #1664 / #1846) and ends up on the fsspec branch via `_is_fsspec_uri`.

## Affected call sites

The case-sensitive check appears in:

- `xrspatial/geotiff/_reader.py:145` (eager `read_to_array` dispatcher)
- `xrspatial/geotiff/_backends/dask.py:197-200` (dask backend HTTP/fsspec branch)
- `xrspatial/geotiff/_sources.py:1454-1460` (`_is_fsspec_uri`)
- `xrspatial/geotiff/_sources.py:1639` (`_open_source`)
- `xrspatial/geotiff/_sidecar.py:46` (`_is_http_url`)
- `xrspatial/geotiff/_writer.py:1244` (writer `_is_fsspec_uri`)
- `xrspatial/geotiff/_backends/gpu.py:350, 1056` (GPU backend HTTP routing)

`_is_fsspec_uri` only rejects lowercase `http://` / `https://`, so `HTTP://127.0.0.1/foo.tif` matches `://` and routes to `_CloudSource` (fsspec). The SSRF validator, private-host check, and pinned-DNS connection are all skipped.

## Reproducer

```python
from xrspatial.geotiff._sources import _is_fsspec_uri
assert _is_fsspec_uri("HTTP://127.0.0.1/foo.tif") is False  # currently True
```

## Fix direction

Parse the URL once with `urlparse(source).scheme.lower()` and use the lower-cased scheme at every dispatch site. `_is_fsspec_uri` should reject HTTP(S) case-insensitively. Add regression tests for `HTTP://127.0.0.1/...` and `HTTPS://...`.

## Versions

- xarray-spatial: main (current)

brendancol:
Closes #2323.

URL schemes are case-insensitive per RFC 3986, but the geotiff reader's
dispatch helpers used case-sensitive `startswith(('http://', 'https://'))`
checks. An uppercase `HTTP://127.0.0.1/foo.tif` slipped past `_HTTPSource`
(and its SSRF allow-list and pinned DNS from #1664 / #1846) and landed on
the fsspec branch via `_is_fsspec_uri`.

## Summary

- Add `_is_http_url` in `_sources.py` that runs the source through `urlparse`
  once and case-folds the scheme.
- Switch every dispatch site to use it: `_reader.read_to_array`, the dask
  backend HTTP/fsspec branch and per-tile source-class check, `_open_source`,
  `_is_fsspec_uri` (reader and writer copies), `_sidecar._is_http_url`, and
  both GPU-backend routing points.
- Add `test_uppercase_scheme_ssrf_2323.py` with case-insensitive coverage
  for the helpers and for the dispatcher itself.

## Backend coverage

No backend logic changed. The fix lives in scheme-detection helpers that
sit upstream of every backend (numpy, dask, gpu), so all four backends
benefit automatically.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_uppercase_scheme_ssrf_2323.py` (30 passed)
- [x] `pytest xrspatial/geotiff/tests/test_ssrf_hardening_1664.py xrspatial/geotiff/tests/test_dns_rebinding_pin_issue_1846.py` (56 passed)
- [x] Full `pytest xrspatial/geotiff/tests/` (5289 passed, 68 skipped)

### DIFF
--- a/xrspatial/geotiff/_backends/dask.py
+++ b/xrspatial/geotiff/_backends/dask.py
@@ -194,11 +194,8 @@ def read_geotiff_dask(source: str, *,
     # and ``_CloudSource`` satisfies that contract. Going through it
     # bounds metadata reads to ``MAX_HTTP_HEADER_BYTES`` instead of
     # fetching the whole remote object up front. See PR #1755 review.
-    is_http = (
-        isinstance(source, str)
-        and source.startswith(('http://', 'https://'))
-    )
-    from .._reader import _is_fsspec_uri
+    from .._reader import _is_fsspec_uri, _is_http_url
+    is_http = _is_http_url(source)
     is_fsspec = isinstance(source, str) and _is_fsspec_uri(source)
     http_meta = None
     http_meta_key = None
@@ -576,8 +573,8 @@ def _delayed_read_window(source, r0, c0, r1, c1, overview_level, nodata,
         # fsspec-addressable remotes (s3://, gs://, az://, memory://, ...).
         # Both source classes expose ``read_range``, which is all
         # ``_fetch_decode_cog_http_tiles`` needs.
-        _is_http_src = isinstance(source, str) and source.startswith(
-            ('http://', 'https://'))
+        from .._reader import _is_http_url as _ihu
+        _is_http_src = _ihu(source)
         _is_fsspec_src = False
         if http_meta is not None and isinstance(source, str) and \
                 not _is_http_src:
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -1054,10 +1054,16 @@ def _gds_chunk_path_available(source, ifd, has_sparse_tile, orientation):
     """
     if not isinstance(source, str):
         return False
+    # The http(s) gate must NOT live inside a ``try/except`` -- a hidden
+    # import failure would silently let an HTTP URL into the kvikio
+    # branch (which opens the path as a local file and panics). The
+    # canonical case-insensitive helper is a sibling module, so the
+    # import is safe at module load time (#2323).
+    from .._sources import _is_http_url
+    if _is_http_url(source):
+        return False
     try:
-        from .._reader import _is_fsspec_uri, _is_http_url
-        if _is_http_url(source):
-            return False
+        from .._reader import _is_fsspec_uri
         if _is_fsspec_uri(source):
             return False
     except Exception:
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -346,8 +346,9 @@ def read_geotiff_gpu(source: str, *,
     # whole image either way for the eager path; the trade-off is a CPU
     # decode instead of nvCOMP-on-GPU. Callers who want bounded GPU
     # memory should pass ``chunks=...``.
+    from .._reader import _is_http_url
     if isinstance(source, str) and (
-            source.startswith(('http://', 'https://'))
+            _is_http_url(source)
             or _is_fsspec_uri(source)):
         return _read_geotiff_gpu_eager_via_cpu(
             source, dtype=dtype, window=window,
@@ -1053,10 +1054,10 @@ def _gds_chunk_path_available(source, ifd, has_sparse_tile, orientation):
     """
     if not isinstance(source, str):
         return False
-    if source.startswith(('http://', 'https://')):
-        return False
     try:
-        from .._reader import _is_fsspec_uri
+        from .._reader import _is_fsspec_uri, _is_http_url
+        if _is_http_url(source):
+            return False
         if _is_fsspec_uri(source):
             return False
     except Exception:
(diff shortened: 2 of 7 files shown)


brendancol (commented):
## PR Review: Reject uppercase HTTP(S) schemes in geotiff URL dispatch (#2323)

### Blockers

None.

### Suggestions

- [ ] **Add a dask-backend dispatch test.** `xrspatial/geotiff/_backends/dask.py:197-199` and the per-tile branch around line 575-583 are touched here, but no test exercises the dask path for an uppercase URL. `test_uppercase_scheme_ssrf_2323.py` covers `_open_source`, `read_to_array`, and the helpers -- not `_read_geotiff_dask`. A small monkeypatched test (stub `_parse_cog_http_meta` like the existing dask-HTTP coalesce tests do) confirming an `HTTP://example.com/foo.tif` source lands on the http branch would close the gap.

- [ ] **GPU `_can_use_kvikio` lost its outside-the-try http rejection.** At `xrspatial/geotiff/_backends/gpu.py:1057-1064` the old code rejected http URLs before the `try` block that imported `_is_fsspec_uri`. Now both checks live inside the same `try: ... except Exception: pass`. If the `_reader` import ever fails (unlikely once the package is loaded, but the original code clearly cared), the function silently falls through to the kvikio eligibility check instead of refusing the HTTP URL. Either move `_is_http_url` back outside the try, or drop the try and let an import error propagate -- if `_reader` won't import, the kvikio path isn't the right thing to fix.

### Nits

- [ ] **`urlparse` import inside `_is_http_url`.** `xrspatial/geotiff/_sources.py:1464` imports `urlparse` on every call. Python caches it, so the cost is a dict lookup, but the helper runs per-chunk in dask paths. Module-scope import would match `_validate_http_url`'s neighbour code and read a little cleaner. Either way is fine.

- [ ] **`_sidecar._is_http_url` and `_writer._is_fsspec_uri` indirection.** Both now import-and-delegate inside the function body. A module-scope import would be tidier. Same caching argument.

### What looks good

- `_is_http_url` is the right shape: `urlparse` + `scheme.lower()`, matching the pattern `_validate_http_url` already uses.
- Coverage of the bug surface is thorough -- every grep hit for case-sensitive `startswith` outside tests has been converted.
- The regression test parameterizes the case-insensitive variants (`HTTP`, `Http`, `hTTpS`) against both the helpers and the dispatcher, including a loopback case and the cloud-metadata IP (169.254.169.254).
- The `read_to_array` test stubs `_read_cog_http` rather than making a network call, so it stays fast and deterministic.
- `test_ssrf_hardening_1664` and `test_dns_rebinding_pin_issue_1846` still pass, so the lowercase path is unchanged.
- Full geotiff suite: 5289 passed, 68 skipped.

### Checklist

- [x] Behaviour matches the cited fix direction (urlparse + lowercase scheme)
- [x] No backend-specific logic changed (helper sits upstream of all backends)
- [N/A] NaN handling -- not relevant to URL dispatch
- [x] Edge cases covered (mixed-case schemes, public IP, loopback, link-local metadata IP)
- [N/A] Dask chunk boundaries -- not relevant
- [N/A] Premature materialization -- not relevant
- [N/A] Benchmark -- pure correctness fix
- [N/A] README feature matrix -- no new public API
- [x] Docstrings on the new helper

brendancol (commented):
## PR Review (follow-up): Reject uppercase HTTP(S) schemes in geotiff URL dispatch (#2323)

Covers commit `a73226ee` on top of `8e2a9d69`.

### Blockers

None.

### Suggestions

None remaining. All four items from the first pass are addressed:

1. Dask-backend dispatch test (`TestDaskBackendUppercaseDispatch` in `test_uppercase_scheme_ssrf_2323.py`) -- stubs `_HTTPSource` and `_CloudSource`, asserts an uppercase `HTTP://` URL constructs the http source, an uppercase `S3://` URL still goes to the cloud source, and the lowercase http case still works. 3 new test methods.
2. `_can_use_kvikio` (gpu.py:1057-1064) -- the `_is_http_url` check now sits outside the `try/except Exception: pass`. The fsspec check stays inside the try, which is the right split: an http URL must never reach the kvikio branch, but a transient import problem on the fsspec side should not poison kvikio eligibility for plain local files.
3. `urlparse` lifted to module scope in `_sources.py:34`.
4. `_sidecar._is_http_url` and `_writer._is_fsspec_uri` now import the canonical helper at module scope.

### Nits

None.

### What looks good

- The dask test uses sentinel exceptions on `_HTTPSource` / `_CloudSource` construction, so dispatch is observable without standing up a mock HTTP server or fsspec backend. Fast and explicit.
- The `S3://` counter-test confirms the fix did not over-correct -- fsspec URIs still go to fsspec.
- Module-scope imports drop the per-call dict lookup on the dispatch hot path.

### Test status

- `pytest xrspatial/geotiff/tests/test_uppercase_scheme_ssrf_2323.py` -- 33 passed
- `pytest xrspatial/geotiff/tests/` -- 5292 passed, 68 skipped

### Recommendation

PR #2328, opened 2026-05-23
Route GPU eager sidecar reads through the base buffer for georef (#2324)

### LINKED ISSUE #2324: GPU eager sidecar reads can lose or corrupt inherited georef attrs
## Describe the bug

The GPU eager `read_geotiff_gpu` path mis-parses georef tags when a sibling `.tif.ovr` sidecar is present.

In `xrspatial/geotiff/_backends/gpu.py` around line 394, the path swaps `data, header` to the sidecar's bytes when the selected IFD lives in the sidecar:

```python
origin = sidecar_origin.get(id(ifd))
if origin is not None:
    data, header = origin
    sidecar_owned_ifd = True
```

A few lines later it calls the georef extractor with that swapped buffer and without passing `sidecar_origin`:

```python
geo_info = extract_geo_info_with_overview_inheritance(
    ifd, ifds, data, header.byte_order,
    allow_rotated=allow_rotated)
```

For the common GDAL sidecar case where overview IFDs carry no geokeys, the helper finds the base IFD and parses it against `data` / `header.byte_order`. Those locals now point at the sidecar buffer, so the base IFD's ModelPixelScale, ModelTiepoint, and GeoKeyDirectory get parsed off the wrong byte buffer. The CPU eager path (`_reader.py` ~line 244) and the dask metadata path (`__init__.py` ~line 304) keep the base buffer and pass `sidecar_origin=georef_origin` to the helper. The GPU eager path does not.

## Expected behavior

When the GPU eager read selects a sidecar IFD, the call to `extract_geo_info_with_overview_inheritance` should still resolve base-IFD georef tags against the base file's bytes. If the sidecar IFD carries its own georef payload, the helper should route that parse to the sidecar via the `sidecar_origin` mapping. The result should match what the CPU eager and dask backends return for the same `.tif` + `.tif.ovr` pair.

## Actual behavior

For a `.tif` + GDAL-style `.tif.ovr` pair where overview level >= 1 lives in the sidecar:

- The base IFD's georef tags are read at sidecar offsets, returning garbage (transform, CRS payload, sometimes the geokey directory itself).
- The returned `DataArray.attrs` carry that corrupted georef.
- The CPU eager and dask paths return the correct georef for the same call.

The corruption is silent. Pixel data is still decoded from the sidecar correctly, so a smoke test that only checks shape or dtype passes.

## Fix shape

- Stop overwriting the base `data` / `header` locals before the georef call. Use distinct names (e.g. `ifd_data`, `ifd_header`) for the swapped tile-decode buffer.
- Pass `sidecar_origin=georef_origin` to `extract_geo_info_with_overview_inheritance`, matching the CPU eager path in `_reader.py` and the dask path in `__init__.py`.
- Add a GPU + sidecar attr parity test that loads the same `.tif` + `.tif.ovr` pair via CPU eager, dask, and GPU, and asserts georef attrs match across backends. Gate the GPU half with the project's existing skip-when-no-cupy / no-cuda decorator.

## Files

- `xrspatial/geotiff/_backends/gpu.py:394` (the swap)
- `xrspatial/geotiff/_backends/gpu.py:403` (the call missing `sidecar_origin`)
- `xrspatial/geotiff/_reader.py:223,244` (CPU eager reference pattern)
- `xrspatial/geotiff/__init__.py:288,304` (dask reference pattern)

brendancol:
Closes #2324

## Summary

The GPU eager `read_geotiff_gpu` path overwrote its local `data, header` references with the sidecar's buffer before calling `extract_geo_info_with_overview_inheritance`. The CPU eager path (`_reader.py`) and the dask metadata path (`__init__.py`) keep `data, header` bound to the base file and pass `sidecar_origin=georef_origin` to the helper. The GPU path did not.

In today's tree the divergence is latent rather than user-visible: IFD entries are pre-resolved at `parse_all_ifds` time and `extract_geo_info` -> `_parse_geokeys` reads from `ifd.entries[...].value` rather than re-resolving offsets against `data`, so the pre-fix code happens to return the correct georef. The fix aligns the GPU path with the CPU and dask paths and protects against any future change that does start using `data` for offset resolution. The intent of the variable names also becomes clear: `data` and `header` are always the base file's; `ifd_data` and `ifd_header` are the swapped buffer used for tile and strip reads when the selected IFD lives in the sidecar.

This PR mirrors the CPU/dask pattern on the GPU eager path:

- Keep `data, header` pointing at the base file's buffers.
- Use new `ifd_data, ifd_header` locals for tile / strip reads that need the sidecar buffer when the selected IFD lives in the sidecar.
- Build the `georef_origin` mapping and pass it to the helper so a sidecar IFD that does carry its own georef payload routes the parse to sidecar bytes (the #2315 contract on the GPU eager path).
- Pass `sidecar_origin=None` explicitly at the chunked GPU GDS qualification probe site so all four call sites of the helper share the same call shape.

## Backend coverage

- numpy: unchanged.
- cupy (GPU eager): fixed.
- dask+numpy: unchanged.
- dask+cupy: unchanged. The chunked GPU path falls through to `read_geotiff_dask` for sidecar files, which already routes through the CPU sidecar handling.

## Test plan

- [x] New `test_gpu_sidecar_georef_parity_2324.py` loads the same `.tif` + `.tif.ovr` pair via CPU eager, dask, and GPU and asserts georef attrs (transform, crs) match. The tests confirm cross-backend agreement, not pre-fix corruption (see Summary above).
- [x] GPU assertions gated behind a `cupy + CUDA required` skip via the project's existing `_gpu_available()` helper. Verified locally on a CUDA-enabled box (`cupy.cuda.is_available() == True`).
- [x] Full sidecar + GPU related test set passes (97 tests via `-k "sidecar or gpu_window or gpu_nodata or remote_sidecar"`).

### DIFF
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -1279,9 +1279,17 @@ def _read_geotiff_gpu_chunked(source, *, dtype, chunks, overview_level,
             if not ifds:
                 raise ValueError("No IFDs found in TIFF file")
             ifd = select_overview_ifd(ifds, overview_level)
+            # The GDS qualification probe parses base-file IFDs only --
+            # sidecar files do not qualify for the disk->GPU fast path
+            # and the chunked path falls through to ``read_geotiff_dask``
+            # which carries its own sidecar handling. Pass
+            # ``sidecar_origin=None`` explicitly so all four call sites
+            # of this helper share the same call shape (review nit
+            # on #2324).
             geo_info = extract_geo_info_with_overview_inheritance(
                 ifd, ifds, raw, header.byte_order,
                 allow_rotated=allow_rotated,
+                sidecar_origin=None,
             )
             orientation = ifd.orientation
             has_sparse_tile = (
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -373,9 +373,10 @@ def read_geotiff_gpu(source: str, *,
         # Append sibling `.tif.ovr` sidecar IFDs onto the pyramid list so
         # ``overview_level`` indexes both internal and external overviews
         # (issue #2112). When the selected IFD comes from the sidecar,
-        # we swap ``data`` / ``header`` to the sidecar's buffers below
-        # and skip the GDS fast path -- GDS reads the source file path,
-        # which would point at the base file rather than the sidecar.
+        # tile / strip reads below use ``ifd_data`` / ``ifd_header`` for
+        # the sidecar's buffers, and we skip the GDS fast path -- GDS
+        # reads the source file path, which would point at the base
+        # file rather than the sidecar.
         from .._sidecar import attach_sidecar_origin, close_sidecar, find_sidecar, load_sidecar
         sidecar_origin: dict[int, tuple] = {}
         sidecar_path = find_sidecar(source)
@@ -388,21 +389,36 @@ def read_geotiff_gpu(source: str, *,
         # Skip mask IFDs (NewSubfileType bit 2)
         ifd = select_overview_ifd(ifds, overview_level)
 
-        # Swap base data / header for the sidecar's buffers when the
-        # requested overview level lives in the sidecar. Subsequent
-        # tile / strip reads slice the right buffer.
+        # Keep ``data`` / ``header`` bound to the base file's buffers so
+        # the georef extractor below resolves base-IFD tag offsets
+        # against the right bytes. Introduce ``ifd_data`` / ``ifd_header``
+        # for tile / strip reads that need the sidecar buffer when the
+        # selected IFD lives in the sidecar (issue #2324).
+        ifd_data, ifd_header = data, header
         origin = sidecar_origin.get(id(ifd))
         if origin is not None:
-            data, header = origin
+            ifd_data, ifd_header = origin
             sidecar_owned_ifd = True
 
         bps = resolve_bits_per_sample(ifd.bits_per_sample)
         file_dtype = tiff_dtype_to_numpy(bps, ifd.sample_format)
         # Inherit georef from the level-0 IFD when the overview itself
         # has no geokeys (issue #1640); pass-through for level 0.
+        # ``sidecar_origin`` routes tag-offset parsing to the sidecar
+        # bytes for any IFD that lives in the sidecar AND declares its
+        # own georef payload, mirroring the CPU eager path
+        # (``_reader.py``) and dask path (``__init__.py``). Without
+        # this, a sidecar-owned IFD that lacks geokeys would parse the
+        # base IFD against the sidecar bytes (issue #2324).
+        georef_origin = (
+            {iid: (od, oh.byte_order)
+             for iid, (od, oh) in sidecar_origin.items()}
+            if sidecar_origin else None
+        )
         geo_info = extract_geo_info_with_overview_inheritance(
             ifd, ifds, data, header.byte_order,
-            allow_rotated=allow_rotated)
+            allow_rotated=allow_rotated,
+            sidecar_origin=georef_origin)
         # Capture the Orientation tag (274) once so the post-decode flip
         # below picks it up for both the stripped fallback and the tiled
         # GPU pipelines. CPU read_to_array applies the array remap +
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Route GPU eager sidecar reads through the base buffer for georef (#2324)

None.

- [ ] **PR body slightly overstates the pre-fix harm.** The body says the helper "parsed the inherited base IFD against the sidecar bytes" leading to silently wrong georef. In practice, IFD entries are pre-resolved at `parse_all_ifds` time (each `IFDEntry.value` is already extracted against the IFD's source buffer), and `extract_geo_info` -> `_parse_geokeys` does not use the `data` argument to look up offsets. It reads `ifd.get_value(TAG_GEO_DOUBLE_PARAMS)` and `ifd.entries[...].value`. So the pre-fix code returned correct georef in today's tree, even though the buffer pointer was wrong. The fix is still right -- it aligns the GPU path with the CPU and dask paths and protects against any future change that does use `data` for offset resolution -- but the PR body could acknowledge that the bug is latent today rather than user-visible, so a reader does not go looking for a regression test that exposes the divergence. The parity tests confirm agreement across backends, not pre-fix corruption.

- [ ] `xrspatial/geotiff/_backends/gpu.py:1282`: the chunked GPU GDS qualification probe also calls `extract_geo_info_with_overview_inheritance(ifd, ifds, raw, header.byte_order, ...)` without `sidecar_origin`. The chunked path does not attach sidecar IFDs to its `ifds` list (parses the base file only) and falls through to `read_geotiff_dask` for non-qualifying files, so the omission is not exercised today. Adding `sidecar_origin=None` explicitly, or wiring the same mapping for completeness, would make the four call sites of this helper match shape. Out of scope for the current PR per issue #2324's GPU-eager framing, but a candidate follow-up if you want full alignment.

- [ ] `xrspatial/geotiff/tests/test_gpu_sidecar_georef_parity_2324.py:79`: `_attrs_subset` builds `tuple(da.attrs.get("transform", ()))`. If a backend ever stores the transform as a numpy array, `tuple(np.array)` yields numpy scalars and the equality check may still pass but with surprising types. `tuple(float(x) for x in da.attrs.get("transform", ()))` normalizes that. Minor.

### What looks good
- The `data, header` -> `ifd_data, ifd_header` rename is consistent across all four downstream call sites (lines 689, 753, 774, 795). Comments updated to match the rename, and the new variable names make the sidecar-vs-base distinction obvious to a future reader.
- `georef_origin` construction matches the CPU and dask paths byte-for-byte. The `if sidecar_origin else None` short-circuit avoids allocating an empty dict in the common no-sidecar case.
- Tests share the `_write_pair` fixture from `test_sidecar_own_geokeys_2315.py` rather than duplicating the IFD-rewriting plumbing, with a docstring note explaining why. Keeps the two suites coupled to the same fixture format.
- GPU assertions gated behind `@_gpu_only` (cupy + CUDA), so the test file runs cleanly on CI machines without a GPU.

### Checklist
- [x] Algorithm matches reference: pattern mirrors `_reader.py:244` and `__init__.py:304`.
- [x] All implemented backends produce consistent results: confirmed by the new parity test trio.
- [x] NaN handling: not applicable, metadata-only change.
- [x] Edge cases covered: with-geokeys, without-geokeys, and the cross-backend baseline.
- [x] No premature materialization or unnecessary copies: no new allocations beyond the `georef_origin` mapping (skipped when there is no sidecar).
- [x] Benchmark: not applicable.
- [x] README feature matrix: not applicable, no new public function.
- [x] Docstrings present and accurate: helper docstrings on the test file and inline comments explain the bug, the fix, and the alignment with CPU and dask paths.

brendancol (commented):
## PR Review (follow-up after nit fixes): Route GPU eager sidecar reads through the base buffer for georef (#2324)

### Blockers
None.

### Suggestions
None remaining.

### Nits
None remaining. The previous review's nits are addressed:

- `xrspatial/geotiff/_backends/gpu.py:1282`: `sidecar_origin=None` is now passed explicitly at the chunked GPU GDS qualification probe call site, with a comment that explains why the probe parses base IFDs only. All four call sites of `extract_geo_info_with_overview_inheritance` in this file now share the same call shape.
- `xrspatial/geotiff/tests/test_gpu_sidecar_georef_parity_2324.py:79`: `_attrs_subset` coerces transform elements to `float` so the cross-backend equality check uses native Python floats.
- PR body now flags that the bug is latent today rather than user-visible, so readers do not look for a regression test that exposes pre-fix corruption.

### What looks good
- Follow-up commit is tightly scoped (2 files, +18 / -2). No drive-by refactors.
- All 37 sidecar-related tests pass locally after the follow-up.

### Checklist
- [x] All findings from the previous review either addressed or noted as out of scope with reason.
- [x] No new findings introduced by the follow-up commit.
- [x] PR body now matches the actual scope of the change.

Issue #2321, opened 2026-05-23
GeoTIFF release hardening: define and lock down supported VRT contract


brendancol:
## Context

The GeoTIFF module is now large enough that release readiness should be judged by an explicit support contract, not by the fact that many edge cases have tests. VRT support is useful for xarray-spatial users, but it should remain a conservative interoperability layer rather than an attempt to reimplement GDAL VRT completely.

Recommendation: keep VRT support, but document and enforce a narrow supported subset. Anything outside that subset should fail closed with actionable errors rather than partially working, silently dropping metadata, or returning pixels whose downstream geospatial meaning is ambiguous.

## Recommended VRT support contract

Supported for release:

- Simple GDAL VRT mosaics backed by GeoTIFF sources.
- Sources with compatible CRS, transform orientation, pixel size, dtype, band count, and band layout.
- Windowed reads where source windows map cleanly to destination windows.
- Lazy/dask reads over the same supported subset.
- Explicit handling of nodata, including mixed-band nodata rejection by default.
- Missing source policy controlled by `missing_sources`, with `raise` as the default.

Not promised for release unless explicitly added later:

- Warped VRTs / reprojection VRTs.
- Arbitrary resampling semantics beyond the implemented, tested subset.
- Mixed CRS, mixed resolution, mixed dtype, or mixed band metadata unless there is a documented explicit opt-in.
- Nested VRTs.
- Complex source/mask/alpha semantics not fully modeled by the attrs contract.
- Full GDAL VRT parity.

The docs should say this plainly: VRT is an advanced compatibility feature for simple mosaics, not a GDAL replacement.

### PR 1: Publish the VRT contract

- Add a VRT support matrix to the GeoTIFF reference docs.
- Mark VRT as `advanced` in prose wherever GeoTIFF tiers are described.
- Document the supported subset and known non-goals above.
- Add examples of safe usage and examples that intentionally raise.
- Ensure public docstrings for `open_geotiff(... .vrt ...)`, `read_vrt`, and `write_vrt` match the same contract.

Acceptance criteria:

- Docs clearly distinguish supported simple mosaics from unsupported GDAL VRT features.
- No doc path implies full GDAL VRT compatibility.

### PR 2: Centralize VRT capability validation

- Add a single validator for parsed VRT metadata before read execution.
- Validate CRS compatibility, dtype compatibility, band count, nodata policy, transform orientation, pixel-size compatibility, source rectangle/destination rectangle sanity, and supported resampling.
[...]

PR #2335, opened 2026-05-23
GeoTIFF: VRT backend parity with .ovr sidecar interactions (#2321 sub-task 4)

### LINKED ISSUE #2330: GeoTIFF: VRT backend parity with .ovr sidecar interactions (#2321 sub-task 4)
Add a focused backend parity matrix for VRT reads across eager and dask, including external overview sidecars (VRT references GeoTIFFs that carry `.tif.ovr` pyramids). Assert metadata parity, not only values. Specifically, windowed reads should shift coords and `attrs['transform']` consistently between eager and dask, and the sidecar `.ovr` lookup should produce the same georef status and CRS attrs as an inline-overview source.

Acceptance: the VRT path cannot pass by returning correct pixel values with wrong georeferencing attrs. Windowed eager and lazy VRT reads agree on shape, coords, attrs, and values.

## Parent issue

Tracks sub-task 4 of #2321.

## Out of scope

- VRT capability validation (covered by sub-task 2).
- Metadata-only parity matrix that does not exercise sidecar overviews (covered by sub-task 3).
- Reprojection or warped VRTs.

## Files

- Create: `xrspatial/geotiff/tests/test_vrt_backend_parity_2321.py`
- Possibly add a `.vrt` + sidecar fixture pair under `xrspatial/geotiff/tests/golden_corpus/` if no existing case exercises the path.

brendancol:
## Summary

- Adds `xrspatial/geotiff/tests/test_vrt_backend_parity_2321.py`, a small parity matrix that asserts pixel and metadata equality between eager and dask VRT reads, plus sidecar `.ovr` vs inline-overview equivalence.
- Mirrors the structure of `test_backend_parity_matrix.py`: declarative fixture and backend specs, a shared `_assert_metadata_parity` helper, one parametrised cell per (fixture, backend) pair.
- Windowed-coord shift parity is split out into its own test so a coord or transform drift on the dask path surfaces a single named failure instead of a matrix-wide flag.

## Backend coverage

Tests-only. Backends exercised by the new cells:

- numpy (eager)
- dask+numpy

GPU + VRT parity already has dedicated coverage in `test_vrt_lazy_chunks_1814.py` and is out of scope here.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_vrt_backend_parity_2321.py -v`: 12 cells pass locally.
- [x] Sanity check: `test_assert_metadata_parity_flags_transform_drift` confirms the harness itself fails on a transform-only drift.
- [x] Sidecar fixture pair (`overview_external_ovr_uint16.tif` + `.tif.ovr`) and the inline-overview comparison fixture both already live under `golden_corpus/fixtures/`. No new on-disk fixtures added.

Closes #2330. Sub-task 4 of #2321.

### DIFF
--- a/xrspatial/geotiff/tests/test_vrt_backend_parity_2321.py
+++ b/xrspatial/geotiff/tests/test_vrt_backend_parity_2321.py
@@ -1,0 +1,575 @@
+"""Backend parity for VRT reads with sidecar/overview interactions (#2321 sub-task 4).
+
+Sub-task 4 of issue #2321 (parent) locks down the VRT support contract by
+asserting eager / dask parity on the surface that is most likely to drift:
+
+* metadata (``attrs['transform']``, ``attrs['crs']``, ``attrs['crs_wkt']``,
+  ``attrs['georef_status']``), not just pixel values;
+* coords -- a windowed read must shift the ``y`` / ``x`` arrays consistently
+  between the eager and the lazy code paths;
+* sidecar interactions -- a VRT whose backing source is a GeoTIFF with an
+  external ``.tif.ovr`` pyramid must surface the same georef attrs (and
+  pixel values) as an equivalent VRT over the inline-overview fixture.
+
+The shape mirrors ``test_backend_parity_matrix.py``: a small declarative
+fixture / backend matrix, one ``assert_parity`` helper, and a single
+parametrised test. We do not re-invent helpers -- the materialisation
+and pixel-comparison primitives match the matrix file so a future move
+to the shared parity harness is mechanical.
+
+VRT fixtures use the existing ``write_vrt`` writer (``xrspatial.geotiff
+._vrt.write_vrt``) on top of ``to_geotiff`` source tiles, plus the
+bundled ``overview_external_ovr_uint16.tif`` / ``.tif.ovr`` sidecar pair
+from ``golden_corpus/fixtures/`` (and its inline-overview counterpart).
+
+Acceptance per the parent issue: the VRT path cannot pass by returning
+correct pixel values with wrong georeferencing attrs. Windowed eager
+and lazy VRT reads agree on shape, coords, attrs, and values.
+"""
+from __future__ import annotations
+
+import pathlib
+import shutil
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Callable
+
+import numpy as np
+import pytest
+import xarray as xr
+
+from xrspatial.geotiff import open_geotiff, to_geotiff
+from xrspatial.geotiff._vrt import write_vrt as _write_vrt_internal
+
+
+# ---------------------------------------------------------------------------
+# Fixture paths shipped under golden_corpus.
+# ---------------------------------------------------------------------------
+
+_GOLDEN = (
+    pathlib.Path(__file__).resolve().parent
+    / "golden_corpus"
+    / "fixtures"
+)
+_SIDECAR_TIF = _GOLDEN / "overview_external_ovr_uint16.tif"
+_SIDECAR_OVR = _GOLDEN / "overview_external_ovr_uint16.tif.ovr"
+_INLINE_OVR_TIF = _GOLDEN / "overview_internal_uint16.tif"
+
+
+def _sidecar_fixture_or_skip() -> Path:
+    """Return the bundled sidecar TIFF or skip if absent."""
+    if not _SIDECAR_TIF.exists() or not _SIDECAR_OVR.exists():
+        pytest.skip("sidecar overview fixture not present in golden_corpus")
+    return _SIDECAR_TIF
+
+
+def _inline_overview_fixture_or_skip() -> Path:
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: VRT backend parity with .ovr sidecar interactions (#2321 sub-task 4)

None.

- [ ] `test_vrt_backend_parity_2321.py:317-328` -- the `sidecar-uint16-window` cell only checks eager-vs-dask parity for its windowed transform; it does not pin absolute values. The float32 windowed test (`test_windowed_vrt_shifts_coords_and_transform_consistently`) does pin absolute coords and the pixel-size half of the transform tuple, so a regression where BOTH backends shift coords the same wrong way would slip past the sidecar cell. Either reuse that absolute pin (the bundled sidecar fixture has known pixel size 0.001 and origin -120.0, 45.0), or document explicitly that the sidecar cell is parity-only.

- [ ] `test_vrt_backend_parity_2321.py:355-366` -- the `vrt_fixture` resolver globs `*.vrt` and returns the first match, then re-reads it via `open_geotiff` purely to recover the dtype. The dtype is already determined by the builder; storing it next to the path on first build (or hard-coding per `_FixtureSpec`) avoids a stray `open_geotiff` round-trip on every cache hit.
- [ ] `test_vrt_backend_parity_2321.py:431` -- `test_sidecar_vrt_attrs_match_inline` builds both VRTs fresh on every cell (no session cache) because it uses `tmp_path` directly. Two parametrised cells (eager + dask) and the build is cheap, so this is fine, but a session-scoped fixture would shave a few writes. Pure perf nit.
- [ ] `test_vrt_backend_parity_2321.py:299-329` -- the four `_FIXTURES` entries collapse onto two builders via `builder.__name__`. The collapse logic lives in the `vrt_fixture` resolver and is implicit; a one-line comment on `_FIXTURES` saying "fix_id is unique per (builder, window); the resolver caches per builder" would save a reader a hop.

### What looks good
- The harness mirrors `test_backend_parity_matrix.py` structure faithfully: same `_materialise` / `_assert_pixels_equal` shape, same dataclass-driven matrix, same labelled assertion messages. A future move to a shared parity harness is mechanical, as the docstring claims.
- `test_assert_metadata_parity_flags_transform_drift` is the right kind of sanity check: it locks the helper's own behaviour so a regression that quietly drops one of the metadata assertions cannot let the matrix pass with empty checks.
- The "windowed-cell straddles the tile seam" comment on `two-tile-float32-window-spans-seam` is the kind of intent-explaining note that catches the next reviewer up immediately.
- Re-using the bundled `overview_external_ovr_uint16.tif` / `.tif.ovr` fixture (no new on-disk fixtures added) keeps the corpus small and matches the brief's "only if no existing case exercises the path."

### Checklist
- [x] Algorithm matches reference -- tests-only, no algorithm changes.
- [x] All implemented backends produce consistent results -- exactly what the new cells assert.
- [x] NaN handling is correct -- `_assert_pixels_equal` uses NaN-aware comparison for floats.
- [x] Edge cases covered by tests -- window straddles tile seam; sidecar with and without window; cross-fixture pyramid comparison.
- [x] Dask chunk boundaries handled correctly -- `chunks=(16, 16)` on a 16x32 mosaic yields a 1x2 grid; window cells force the graph to read both backing sources.
- [x] No premature materialization or unnecessary copies -- `_materialise` is the only `.compute()` site, called once at assertion time.
- [n/a] Benchmark exists or is not needed -- tests-only.
- [n/a] README feature matrix updated -- tests-only.
- [x] Docstrings present and accurate -- module-level docstring explicitly names the parent issue and the acceptance bar.

brendancol (commented):
## PR Review (second pass): after follow-up commits

### Status of original findings

- Suggestion (sidecar window absolute pin): fixed in `test_sidecar_window_shifts_to_known_coords`. The bundled fixture's pixel size (0.001) and origin (-120.0, 45.0) are pinned, so a regression that drifts BOTH backends the same way now surfaces.
- Nit (cache dtype lookup): fixed. `vrt_fixture` now caches `(path, dtype)` in an in-process dict; cache hits no longer round-trip through `open_geotiff`.
- Nit (per-cell sidecar build in `test_sidecar_vrt_attrs_match_inline`): deferred. Two parametrised cells, builds are cheap, session-scoping adds complexity for no measurable gain.
- Nit (comment on `_FIXTURES` collapse): fixed. One-line comment above `_FIXTURES` says the fix_id is unique per (builder, window) and the resolver caches per builder.

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The new sidecar absolute-shift test follows the same pattern as `test_windowed_vrt_shifts_coords_and_transform_consistently`, so the two windowed surfaces (float32 mosaic + sidecar) now have symmetric coverage.
- The cache rework removes a stray `open_geotiff` call per cache hit, which is the kind of small cleanup that pays off when the matrix grows.

All 13 cells pass locally.

PR #2337, opened 2026-05-23
GeoTIFF: case-insensitive HTTP(S) scheme routing for SSRF protection (#2332)

### LINKED ISSUE #2332: GeoTIFF: case-insensitive HTTP(S) scheme routing for SSRF protection (#2321 sub-task 5)
Several call sites in `xrspatial/geotiff/` route on `startswith(('http://', 'https://'))`. That is case-sensitive, so `HTTP://example.internal/...` (uppercase) slips past `_HTTPSource`, skips the SSRF / DNS-pinning checks, and falls through to fsspec.

Fix: add a single `_is_http_source(s)` helper that does `urlparse(s).scheme.lower() in ('http', 'https')`, and route every call site through it.

Call sites to update (verify and replace all):
- `xrspatial/geotiff/_sources.py:1458, 1639`
- `xrspatial/geotiff/_reader.py:145`
- `xrspatial/geotiff/_writer.py:1244`
- `xrspatial/geotiff/_backends/gpu.py:350, 1056`
- `xrspatial/geotiff/_backends/dask.py:199`
- `xrspatial/geotiff/_sidecar.py:46` (the existing `_is_http_url` helper is also case-sensitive)

Tests: uppercase `HTTP://` routes through `_HTTPSource`; uppercase URLs targeting `127.0.0.1`, `169.254.169.254`, and `localhost` are still rejected by the SSRF check.

Acceptance: no security-relevant dispatch uses case-sensitive `startswith`. HTTP(S) SSRF protection applies regardless of scheme casing. Regression tests cover uppercase schemes and private-host rejection.

Parent: #2321 (sub-task 5).

brendancol:
## Summary

- Routing on `startswith(('http://', 'https://'))` was case-sensitive, so `HTTP://169.254.169.254/...` slipped past `_HTTPSource` and the SSRF / DNS-pinning checks and fell through to fsspec. Per RFC 3986 section 3.1, URI schemes are case-insensitive.
- Adds a single `_is_http_source(s)` helper in `_sources.py` backed by `urlparse(s).scheme.lower() in ('http', 'https')` and routes every dispatch site through it: `_sources.py` (`_is_fsspec_uri`, `_open_source`), `_reader.py`, `_writer.py` (`_is_fsspec_uri`), `_backends/dask.py`, `_backends/gpu.py` (eager-via-CPU branch + GDS gate), and the `_sidecar.py` `_is_http_url` helper.
- Backend coverage: numpy / cupy / dask+numpy / dask+cupy. The helper is on every dispatch path.

## Test plan

- [x] New `test_http_scheme_case_2321.py`: helper, `_open_source` dispatch for `HTTP` / `HTTPS` / `Http` / `hTTpS`, both `_is_fsspec_uri` classifiers, the sidecar helper, and end-to-end SSRF rejection of uppercase URLs that resolve to `127.0.0.1`, `169.254.169.254`, `10.0.0.1`, `192.168.1.1`, `0.0.0.0`, and `localhost`. `socket.getaddrinfo` is monkeypatched so tests stay offline.
- [x] `test_ssrf_hardening_1664.py` still passes (43 / 43).
- [x] Full `xrspatial/geotiff/tests/` suite (excluding the golden-corpus HTTP test that pulls real fixtures): 5317 passed, 68 skipped.

Closes #2332.
Parent: #2321 sub-task 5.

### DIFF
--- a/xrspatial/geotiff/_backends/dask.py
+++ b/xrspatial/geotiff/_backends/dask.py
@@ -194,8 +194,11 @@ def read_geotiff_dask(source: str, *,
     # and ``_CloudSource`` satisfies that contract. Going through it
     # bounds metadata reads to ``MAX_HTTP_HEADER_BYTES`` instead of
     # fetching the whole remote object up front. See PR #1755 review.
-    from .._sources import _is_http_source
+    # Local imports: backend modules avoid eager-importing the reader /
+    # sources layer at module load so the package can be imported without
+    # urllib3 in environments that only consume the dask path. Issue #2332.
     from .._reader import _is_fsspec_uri
+    from .._sources import _is_http_source
     is_http = _is_http_source(source)
     is_fsspec = isinstance(source, str) and _is_fsspec_uri(source)
     http_meta = None
--- a/xrspatial/geotiff/_backends/dask.py
+++ b/xrspatial/geotiff/_backends/dask.py
@@ -194,11 +194,9 @@ def read_geotiff_dask(source: str, *,
     # and ``_CloudSource`` satisfies that contract. Going through it
     # bounds metadata reads to ``MAX_HTTP_HEADER_BYTES`` instead of
     # fetching the whole remote object up front. See PR #1755 review.
-    is_http = (
-        isinstance(source, str)
-        and source.startswith(('http://', 'https://'))
-    )
+    from .._sources import _is_http_source
     from .._reader import _is_fsspec_uri
+    is_http = _is_http_source(source)
     is_fsspec = isinstance(source, str) and _is_fsspec_uri(source)
     http_meta = None
     http_meta_key = None
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -316,6 +316,7 @@ def read_geotiff_gpu(source: str, *,
     from .._header import parse_all_ifds, parse_header, select_overview_ifd, validate_tile_layout
     from .._reader import (MAX_PIXELS_DEFAULT, _check_dimensions, _FileSource, _is_fsspec_uri,
                            _max_tile_bytes_from_env, _resolve_masked_fill)
+    from .._sources import _is_http_source
 
     # ``source`` is already coerced above (before the dispatch
     # validator); no need to re-coerce here.
@@ -347,7 +348,7 @@ def read_geotiff_gpu(source: str, *,
     # decode instead of nvCOMP-on-GPU. Callers who want bounded GPU
     # memory should pass ``chunks=...``.
     if isinstance(source, str) and (
-            source.startswith(('http://', 'https://'))
+            _is_http_source(source)
             or _is_fsspec_uri(source)):
         return _read_geotiff_gpu_eager_via_cpu(
             source, dtype=dtype, window=window,
@@ -1053,7 +1054,8 @@ def _gds_chunk_path_available(source, ifd, has_sparse_tile, orientation):
     """
     if not isinstance(source, str):
         return False
-    if source.startswith(('http://', 'https://')):
+    from .._sources import _is_http_source
+    if _is_http_source(source):
         return False
     try:
         from .._reader import _is_fsspec_uri
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -1054,6 +1054,8 @@ def _gds_chunk_path_available(source, ifd, has_sparse_tile, orientation):
     """
     if not isinstance(source, str):
         return False
+    # Local import: backend gate is called once per source and the GDS
+    # path is optional, so avoid eager top-level coupling. Issue #2332.
     from .._sources import _is_http_source
     if _is_http_source(source):
         return False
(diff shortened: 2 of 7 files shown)


brendancol (commented):
## PR Review: GeoTIFF: case-insensitive HTTP(S) scheme routing for SSRF protection (#2332)

None.

- [ ] **Repeated `from ._sources import _is_http_source` inside function bodies** at `_backends/dask.py:197`, `_backends/gpu.py:319` (top of `read_geotiff_gpu`), `_backends/gpu.py:1057` (inside `_gds_chunk_path_available`), `_sidecar.py:50` (inside `_is_http_url`), and `_writer.py:1248` (inside `_is_fsspec_uri`) all do per-call imports. `_reader.py` already imports `_is_http_source` at module scope. Local imports add ~1us per dispatch and obscure the dependency graph. Consider promoting at least the two hot ones (`_backends/dask.py:197`, `_writer.py:1248`) to module scope, or folding them into the top-of-function import block that already exists in `read_geotiff_gpu` at line 315.

  Counter-argument: the `gpu.py` block at line 315 already does function-scope imports because the backend modules avoid eager-importing each other. Following that convention is reasonable. If you keep them, a one-line comment ("local to avoid import cycles") would help the next reader.

- [ ] **`urlparse` import inside `_is_http_source`** at `_sources.py:1474`. The function imports `urlparse` on every call. `urllib.parse` is already used at module top (line 468 in `_validate_http_url`). Hoist `from urllib.parse import urlparse` to the module top so the import cost only happens once. `_is_http_source` runs once per dispatch, so the perf delta is tiny.

- [ ] **Behavioural broadening note.** The new helper accepts schemes-only-with-colon (`http:foo` returns True; previously the `startswith` returned False). The validator still rejects these as "no hostname" so the security posture is unchanged, but a one-line test asserting `_is_http_source('http:foo')` and then `_open_source('http:foo')` raises `UnsafeURLError` would lock in the new behaviour. Not a blocker; the new behaviour is more RFC-correct.

- [ ] **Writer error type for uppercase HTTP write targets.** After this PR an `HTTP://...` write target falls through `_is_fsspec_uri` (correctly returns False) and then through the file-like check, landing at the local file write path on line 1297. The resulting error is `OSError: [Errno 22] Invalid argument` (or similar OS-specific) rather than a typed "writes not supported over HTTP". Consider adding a `_is_http_source(path)` check at `_writer.py:1273` that raises `NotImplementedError("writes are not supported over HTTP")`. Out of scope for this PR; pre-existing issue, just made more reachable.

### What looks good

- The helper is small, documented, and replaces seven case-sensitive `startswith` call sites with one. The `urlparse(...).scheme.lower() in ('http', 'https')` approach handles `http:` (scheme without `//`), leading whitespace, and host case automatically.
- The test file is thorough: helper parametrized across mixed-case schemes, end-to-end `_open_source` dispatch with `__init__` interception (no real network), and parametrized SSRF rejection across (scheme casing) x (private IP). All offline via `socket.getaddrinfo` monkeypatch.
- The `_sidecar.py` `_is_http_url` helper now delegates to the same helper, so future drift between the two is harder.
- No public API change. Backend coverage (numpy / cupy / dask+numpy / dask+cupy) is preserved on every dispatch path.
- Full geotiff test suite passes (5317 / 5317 non-network tests).

### Checklist
- [x] Algorithm matches reference (RFC 3986 sect 3.1, case-insensitive schemes)
- [x] All implemented backends produce consistent results (single helper used everywhere)
- [x] NaN handling: N/A
- [x] Edge cases covered (mixed case, non-string, empty string, non-HTTP schemes)
- [x] Dask chunk boundaries: N/A
- [x] Tests added (60 new test cases)
- [x] README feature matrix: N/A (no new public function)

brendancol (commented):
## Follow-up review: PR #2337 (after commit 532d5bcd)

### Blockers
None.

### Suggestions
None. The original Suggestion ("hoist function-scope imports") was applied for `_writer.py` and `_sidecar.py` (no cycle risk), and dismissed-with-reason for `_backends/dask.py` + `_backends/gpu.py`. The dismissal rationale is sound: the backend modules deliberately defer importing `_reader` / `_sources` so the package imports cleanly without urllib3 in dask-only environments. An explanatory comment now sits next to each remaining local import.

### Nits
None. All three original Nits are now resolved:

- `urlparse` is at module scope in `_sources.py` (no per-call import).
- `_is_http_source('http:foo')` returning True is locked in by an explicit test, and `_open_source('http:foo')` -> `UnsafeURLError` is also asserted.
- The writer rejects HTTP(S) destinations with a typed `NotImplementedError` in `_write_bytes`, covered by 5 parametrized cases across (http / https) x (lower / mixed / upper).

### What looks good

- The `NotImplementedError` in `_write_bytes` matches the existing wording in `_write_streaming` (lines 738-740), so the user sees a consistent message regardless of which entry point they hit.
- Test count grew from 60 to 67 cases; full geotiff suite stays at 5324 passed, 68 skipped, no warnings introduced.
- The remaining local imports in the backend modules now carry a one-line comment so the next reviewer does not flag them as oversight.

### Checklist

All boxes still ticked; no new findings.

Disposition of original findings: 3 fixed in-PR (urlparse hoist, behavioural-broadening test, typed writer error), 1 partially fixed + partially dismissed-with-reason (imports hoisted where safe, kept function-scope with comment where backend isolation requires it).

PR #2339, opened 2026-05-23
Centralize VRT capability validation (#2329)

### LINKED ISSUE #2329: GeoTIFF: centralize VRT capability validation (#2321 sub-task 2)
Sub-task of #2321.

VRT capability checks are scattered across `read_vrt` (in `_backends/vrt.py`) and the `open_geotiff` `.vrt` branch, and some checks fire only once a chunk is actually decoded. That means a malformed VRT can build a dask graph successfully and then blow up inside a chunk function, far from the call site.

Add a single validator that runs against an already-parsed `VRTDataset` before any read work starts. It should cover:

- CRS compatibility
- dtype compatibility across bands
- band count sanity
- nodata policy
- transform orientation (flips the read path does not honour)
- pixel-size compatibility
- `SrcRect` vs `DstRect` sanity
- the supported resampling set

Both `read_vrt` and `open_geotiff('foo.vrt')` should call the validator so the two entry points produce the same rejections. Raise a typed `VRTUnsupportedError` (a subclass of the existing geotiff error class in `_errors.py`) and put the offending source path and field name in the message.

The validator must accept already-parsed structures. Do not re-parse the VRT XML inside it.

### Acceptance
- Unsupported VRT features raise at graph build or eager-read setup, not deep in a chunk function.
- `read_vrt(bad.vrt)` and `open_geotiff('bad.vrt')` produce the same error type and message for the same input.
- One negative test per validator rule.

brendancol:
Closes #2329. Part of #2321.

## Summary
- New `xrspatial/geotiff/_vrt_validation.py` with one `validate_parsed_vrt` entry point that audits an already-parsed `VRTDataset` against every capability the read pipeline does not honour: band count, dtype kind, transform orientation, pixel-size, SrcRect/DstRect non-negativity and within-extent, the supported resampling set, plus the existing CRS / mixed-nodata policies.
- Wired into both eager and chunked `read_vrt` dispatch in `_backends/vrt.py`. `open_geotiff('foo.vrt')` delegates to `read_vrt`, so the same validator runs there too. Direct and dispatched entry points produce equivalent failures for the same bad input.
- The rejection lives ahead of any source decode. Previously some checks fired mid-decode (and per chunk task under chunked dispatch), so a malformed VRT could build a dask graph and then blow up deep inside `compute()`.
- New `VRTUnsupportedError` (a `GeoTIFFAmbiguousMetadataError` / `ValueError` subclass) carries the offending source path and field in its message. Existing typed errors (`RotatedTransformError`, `MixedBandMetadataError`, `UnparseableCRSError`) keep their contracts so `except`-by-subclass callers stay green; the validator delegates those cases to the existing registered hooks rather than swapping the type.

## Backend coverage
- numpy: yes (eager read path)
- dask+numpy: yes (chunked dispatcher)
- cupy: not applicable (VRT reads do not go through the GPU decoder pipeline)
- dask+cupy: not applicable

## Test plan
- [x] 17 new tests in `test_vrt_validation_2321.py`: one negative test per validator rule, plus entry-point parity tests asserting `read_vrt` and `open_geotiff` raise the same type and message for the same input.
- [x] Full geotiff suite: 5276 passed, 68 skipped.
- [x] Chunked dispatch: unsupported VRT raises at graph build, not inside a chunk function.

### DIFF
--- a/xrspatial/geotiff/_backends/vrt.py
+++ b/xrspatial/geotiff/_backends/vrt.py
@@ -384,9 +384,27 @@ def read_vrt(source: str, *,
     from .._validation import validate_read_metadata
     from .._vrt import _read_vrt_xml
     from .._vrt import parse_vrt as _parse_vrt
+    from .._vrt_validation import validate_parsed_vrt as _validate_parsed_vrt
     _xml_str = _read_vrt_xml(source)
     _vrt_dir = _os.path.dirname(_os.path.abspath(source))
     _parsed_vrt = _parse_vrt(_xml_str, _vrt_dir)
+    # Centralised VRT capability validator (issue #2329). Runs before
+    # ``validate_read_metadata`` so capability mismatches (negative
+    # SrcRect / DstRect, unsupported resampling, zero pixel size, etc.)
+    # surface with the typed ``VRTUnsupportedError`` and a message that
+    # names the offending source path and field. The validator overlaps
+    # ``validate_read_metadata`` on rotated-transform / unparseable-CRS
+    # / mixed-band-nodata; running the centralised one first keeps the
+    # ``VRTUnsupportedError`` type at the entry-point boundary for the
+    # capability checks added in #2329.
+    _validate_parsed_vrt(
+        _parsed_vrt,
+        source=source,
+        mode='read',
+        allow_rotated=allow_rotated,
+        allow_unparseable_crs=allow_unparseable_crs,
+        band_nodata=band_nodata,
+    )
     _band_nodata_values = (
         [b.nodata for b in _parsed_vrt.bands]
         if _parsed_vrt.bands else None
@@ -736,6 +754,22 @@ def _read_vrt_chunked(source, *, window, band, name, chunks, gpu, dtype,
     # the ``band`` / window / ``max_pixels`` validators below so the
     # error ordering matches the eager path.
     from .._validation import validate_read_metadata
+    from .._vrt_validation import validate_parsed_vrt as _validate_parsed_vrt
+    # Centralised VRT capability validator (issue #2329). Run at graph
+    # build time so capability mismatches surface here, not inside a
+    # per-chunk decode task. ``read_vrt(..., chunks=)`` previously let
+    # unsupported features ride through the graph build and raised
+    # deep in a ``compute()`` chunk function (an opaque user
+    # experience); the validator moves the rejection back to where the
+    # caller stands.
+    _validate_parsed_vrt(
+        vrt,
+        source=source,
+        mode='read',
+        allow_rotated=allow_rotated,
+        allow_unparseable_crs=allow_unparseable_crs,
+        band_nodata=band_nodata,
+    )
     band_nodata_values = (
         [b.nodata for b in vrt.bands] if vrt.bands else None
     )
--- a/xrspatial/geotiff/_errors.py
+++ b/xrspatial/geotiff/_errors.py
@@ -103,6 +103,24 @@ class ConflictingNodataError(GeoTIFFAmbiguousMetadataError):
     """
 
 
+class VRTUnsupportedError(GeoTIFFAmbiguousMetadataError):
+    """A parsed VRT declares a feature the read pipeline does not honour (#2329).
+
+    Raised by the centralised VRT validator at graph-build / eager-read
+    setup time, before any source bytes are decoded. Covers CRS / dtype
+    / band / nodata / transform / pixel-size / source-window /
+    destination-window / resampling mismatches that the VRT read path
+    cannot serve correctly. The message names the offending source path
+    and field so a caller can locate the bad source without re-parsing
+    the VRT XML themselves.
+
+    Subclasses ``GeoTIFFAmbiguousMetadataError`` (and therefore
+    ``ValueError``) so existing ``except ValueError`` callers keep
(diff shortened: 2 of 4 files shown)


brendancol (commented):
## PR Review: Centralize VRT capability validation (#2329)

Read every changed file in full from the issue-2329 worktree. Findings below; the rest of the diff is fine.

### Blockers
None.

### Suggestions

- [ ] Duplicate `GeoTIFFFallbackWarning` on malformed CRS (`_vrt_validation.py:246`). The validator calls `_wkt_to_epsg`, which emits a `GeoTIFFFallbackWarning` on parse failure. The existing `_check_read_unparseable_crs` in `_validation.py:1025` uses `pyproj.CRS.from_user_input` directly with no warning. Under the new code, a malformed VRT CRS emits a fallback warning that the existing check did not. You can see it in the test output for `test_unparseable_crs_rejected_without_opt_in`. Either swap to `from_user_input` directly, or wrap the probe in `warnings.catch_warnings()`.

- [ ] `_looks_like_wkt` and `_NEAREST_RESAMPLE_ALGS` are duplicated instead of imported (`_vrt_validation.py:52-77`). `_NEAREST_RESAMPLE_ALGS` already lives in `_vrt.py:566` and `_looks_like_wkt` already lives in `_crs.py:35`. The docstring justifies the copy by pointing at an import cycle, but a lazy import inside the validator body sidesteps the cycle without the drift risk. If someone adds `Nearest_Neighbor` to one set but not the other later, the bug will be hard to spot.

### Nits

- [ ] `mode='read'` parameter (`_vrt_validation.py:84, 132-137`). The only accepted value is `'read'`; the documented future write mode does not exist yet. Either drop the parameter until a write validator actually needs to share the entry point, or accept that every caller is going to pass `mode='read'` forever.

- [ ] Rotated and unparseable-CRS branches shadow the registered hooks (`_vrt_validation.py:181-193, 240-258`). `_check_read_rotated_transform` and `_check_read_unparseable_crs` still run from `validate_read_metadata` right after the validator, but the validator's branches now preempt them by raising first. That is intentional (to embed the source path in the message), but it is worth a comment in the docstring or in `validate_read_metadata` so a future maintainer modifying the registered hook realises the hook is now reachable only when the validator passes the case (e.g. `allow_rotated=True`).

### What looks good

- TDD order is correct: 17 negative tests added, validator written to make them pass, full geotiff suite stays at 5276 passing.
- The entry-point parity tests (`test_resample_parity_across_entry_points`, `test_rotated_parity_across_entry_points`, `test_zero_bands_parity_across_entry_points`) compare both error type and message between `read_vrt` and `open_geotiff`, which is the strongest form of the parity claim.
- Chunked dispatch is wired (`test_unsupported_resample_chunked_raises_at_build` exercises it), so the "deep in a chunk function" failure mode the parent issue called out is closed.
- `VRTUnsupportedError` is a `GeoTIFFAmbiguousMetadataError` subclass and therefore a `ValueError`, so `except ValueError` callers keep working.
- Existing typed errors keep their contracts: the validator raises `RotatedTransformError` / `MixedBandMetadataError` / `UnparseableCRSError` for cases that already had those types; only the new capability rules raise `VRTUnsupportedError`.

### Checklist
- [x] Algorithm matches reference: not applicable (no algorithmic change).
- [x] All implemented backends produce consistent results: validator runs at both eager and chunked dispatch.
- [x] NaN handling: not applicable.
- [x] Edge cases covered: zero bands, negative SrcRect/DstRect, zero pixel size, out-of-extent DstRect, unsupported resample, rotated transform, unparseable CRS, mixed-band nodata.
- [x] Dask chunk boundaries: rejection happens at graph build, not in a chunk function.
- [x] Benchmark not needed (validation overhead is a few ms per parse).
- [x] README feature matrix: not applicable (no new public API).

brendancol (commented):
## Follow-up review

Second pass after commit 8be5d843. Re-read the diff in the worktree.

### Dispositions from the first review

- Fixed: duplicate `GeoTIFFFallbackWarning` on malformed CRS. The validator now probes pyproj directly with `from_user_input`, matching the silent behaviour of the registered hook. Test output for `test_unparseable_crs_rejected_without_opt_in` no longer emits the spurious warning.
- Fixed: duplicated `_NEAREST_RESAMPLE_ALGS` and `_looks_like_wkt`. The validator lazy-imports both from their canonical homes (`_vrt._NEAREST_RESAMPLE_ALGS` and `_crs._looks_like_wkt`). Drift risk gone.
- Fixed: shadow on registered hooks. Added a Notes section to the validator docstring spelling out the overlap with `_check_read_rotated_transform` and `_check_read_unparseable_crs` so a future maintainer editing either side knows the validator preempts them on the VRT path.
- Dismissed (with reason): drop the `mode='read'` parameter. The master plan (`docs/superpowers/plans/2026-05-22-geotiff-vrt-release-hardening-2321.md`) calls out `validate_parsed_vrt(parsed, *, mode)` as the canonical signature for the centralised entry point, with a write-mode follow-up implied. Dropping the parameter would diverge from the plan; keeping it costs a single `mode != 'read'` check per call.

### Remaining
None. Full geotiff suite stays at 5276 passing.

### Checklist
- [x] All first-pass blockers and suggestions addressed.
- [x] Nit dismissals recorded with reason.
- [x] Tests still pass on the eager and chunked paths.

Issue #2340, opened 2026-05-23
Epic: GeoTIFF release contract and feature tiering


brendancol:
## Goal

Define the release promise for the GeoTIFF module so the next release is explicit about what is stable, advanced, experimental, internal-only, and unsupported.

This epic is the product boundary for the release. The goal is not to make every existing code path stable. The goal is to make the promised surface narrow, accurate, and defensible.

## Release Principle

A user should be able to answer:

- Can I rely on this feature in production?
- What metadata is preserved?
- Which backends are covered by parity tests?
- Which options are advanced or experimental?
- Which inputs should fail loudly instead of partially working?

## Proposed Tiering

Stable:

- Local GeoTIFF read/write.
- COG read/write for stable lossless codecs.
- CRS / transform / nodata attrs contract.
- Windowed reads where supported.
- Dask reads only where parity-tested against eager reads.

Advanced:

- Simple VRT mosaics.
- HTTP/fsspec reads.
- External `.ovr` sidecars.
- BigTIFF.
- Overview generation/selection.

Experimental:

- GPU read/write.
- LERC, JPEG2000/J2K, LZ4.
- Explicit CRS permissive escape hatches.
- Other paths without cross-backend numerical parity claims.

Internal-only:

- JPEG-in-TIFF path unless/until it is made externally interoperable with GDAL/libtiff/rasterio.

Unsupported for this release:

- Full GDAL VRT parity.
- Warped VRTs/reprojection VRTs unless separately implemented and tested.
- Rotated/sheared GeoTIFF write support.
- Silent mixed metadata flattening.

## Work Items

- Audit `SUPPORTED_FEATURES` and align it with the release promise.
- Audit public docstrings for `open_geotiff`, `read_geotiff_dask`, `read_geotiff_gpu`, `read_vrt`, `to_geotiff`, `write_geotiff_gpu`, and `write_vrt`.
- Add a single feature-tier table in docs.
- Add a short release-contract page or section that explains the tier meanings.
- Ensure experimental/internal-only paths require explicit opt-ins where appropriate.
- Ensure unsupported combinations fail with actionable errors.

## Acceptance Criteria

- The release has a written GeoTIFF support contract.
- Stable, advanced, experimental, internal-only, and unsupported features are unambiguous.
- No docs or docstrings imply full support for features outside the contract.
- Every stable feature has a corresponding release-gate test or is moved out of stable.
- Release notes can be written from this contract without caveats hidden in code comments.

## Related Epics

[...]

PR #2350, opened 2026-05-24
Align SUPPORTED_FEATURES with epic #2340 tiering

### LINKED ISSUE #2348: Align SUPPORTED_FEATURES with epic #2340 tiering
## Reason or Problem

Epic #2340 defines the GeoTIFF release contract and proposes a tiering: Stable, Advanced, Experimental, Internal-only. `SUPPORTED_FEATURES` in `xrspatial/geotiff/_attrs.py` is the single source of truth that docs, the user-guide notebook, and writer gates read from. Its contents need to be reconciled with the epic so the public surface matches the release promise.

## Proposal

Audit `SUPPORTED_FEATURES` against epic #2340 and fix the drift.

Tier shape per epic:

- Stable: local GeoTIFF r/w, COG r/w for lossless codecs (none/deflate/lzw/packbits/zstd), CRS/transform/nodata attrs, windowed reads, dask reads where parity-tested.
- Advanced: simple VRT mosaics, HTTP/fsspec reads, external `.ovr` sidecars, BigTIFF, overview generation/selection.
- Experimental: GPU r/w, LERC/J2K/LZ4, explicit-CRS permissive escape hatches.
- Internal-only: JPEG-in-TIFF.

Concrete changes:

- Add `reader.windowed` at `stable`. The window-read tests in `xrspatial/geotiff/tests/` already cover it.
- Add `reader.dask` at `stable`. `test_backend_parity_matrix.py` and `test_backend_full_parity_2211.py` cover it.
- Demote `reader.allow_unparseable_crs` from `advanced` to `experimental`. The epic places explicit-CRS permissive escape hatches in Experimental.
- Demote `reader.allow_rotated` from `advanced` to `experimental`. The rotated read path is an opt-in escape hatch and the epic lists rotated writes as Unsupported, so the read-side opt-in sits in Experimental.
- Add a shape unit test under `xrspatial/geotiff/tests/` that asserts every entry carries a tier label and the tier set is closed (`{stable, advanced, experimental, internal_only}`). Future drift fails CI.
- Add a CHANGELOG entry.

Out of scope, handled by sibling PRs under epic #2340:

- Docstring rewrites for public read/write entry points.
- Docs page or feature-tier table in `docs/source/`.
- New opt-in enforcement on experimental paths.

## Stakeholders and Impacts

GeoTIFF module callers who read `SUPPORTED_FEATURES` to gate optional features. The change is metadata-only. Runtime behaviour of every read/write path stays put. Downstream code that asserted `SUPPORTED_FEATURES['reader.allow_rotated'] == 'advanced'` will see `'experimental'` after this PR. That string is the only break.

## Drawbacks

This is a public-API change to a documented constant. Callers gating on the exact string value have to update.

## Alternatives

Leave the constant alone and document the drift in release notes. The epic asks for the constant to match the release contract, so this defeats the point.

Parent epic: #2340. PR 1 of 6.

brendancol:
Closes #2348. Wave 1 of 6 under epic #2340.

## Summary

- Reconcile `xrspatial.geotiff.SUPPORTED_FEATURES` with the GeoTIFF release-contract tiering proposed in epic #2340.
- Add `reader.windowed` and `reader.dask` at `stable` (both have release-gate coverage).
- Demote `reader.allow_rotated` and `reader.allow_unparseable_crs` from `advanced` to `experimental` to match the epic's placement of permissive read-side escape hatches.
- New shape test asserts every entry has a tier, the tier set is closed, and the dict literal has no duplicate keys.
- CHANGELOG entry.

## What ships

`SUPPORTED_FEATURES` is metadata-only. Runtime behaviour of every read and write path is unchanged. The only break is for callers that gate on the exact string value of the two demoted entries (`reader.allow_rotated`, `reader.allow_unparseable_crs`).

Out of scope under epic #2340 (sibling PRs handle these):

- Docstring rewrites for the public read/write entry points.
- The feature-tier table in `docs/source/`.
- New opt-in enforcement on experimental paths.

## Test plan

- [x] New shape test passes: `pytest xrspatial/geotiff/tests/test_supported_features_shape_2348.py`.
- [x] Pre-existing tier tests still pass: `pytest xrspatial/geotiff/tests/test_supported_features_tiers_2137.py`.
- [ ] CI green on the PR.


brendancol (commented):
## PR Review: Align SUPPORTED_FEATURES with epic #2340 tiering

None.

- `xrspatial/geotiff/_attrs.py:231` -- the older block comment listing things "outside the stable contract" still flags `reader.allow_rotated` without noting that wave-1 also moved its tier from `advanced` to `experimental`. The statement is still factually true (experimental is outside stable), but a reader has to jump to the wave-1 block to learn the tier itself changed. A one-line nudge would make the two comment blocks tell the same story.

- `xrspatial/geotiff/_attrs.py:279-283` -- the inline comment hard-codes the names of seven test files that gate `reader.windowed`. The names aren't load-bearing (no runtime parser reads them), but a future rename will silently stale the comment. Either trim to "covered by the window-read suite in `xrspatial/geotiff/tests/`" or accept the maintenance cost.

### What looks good
- Scope is narrow: `_attrs.py`, one new shape test, one CHANGELOG line. Nothing else touched.
- The shape test parses the dict literal via `ast` to catch Python's silent dedup of duplicate dict-literal keys. That's a real footgun for a single-source-of-truth constant.
- `test_epic_2340_wave_1_reconciliation` pins the four specific tier moves so any future revert needs a justifying commit.
- `test_every_tier_label_is_used` catches the case where a label is documented in the package docstring but no entry references it.
- The CHANGELOG entry names every concrete change, the public-API break, and the parent epic.

### Checklist
- [x] Algorithm matches reference -- N/A (metadata-only).
- [x] All implemented backends produce consistent results -- N/A (no backend code).
- [x] NaN handling is correct -- N/A.
- [x] Edge cases are covered by tests -- empty/closed tier set, duplicate keys, missing tier.
- [x] Dask chunk boundaries handled correctly -- N/A.
- [x] README feature matrix updated -- N/A (no new function).
- [x] Docstrings present and accurate -- new test file has module + per-test docstrings; the constant's tier semantics block covers the new entries.

brendancol (commented):
## Follow-up review pass

Both findings from the first review pass have been addressed in 1be3b116:

- Suggestion (`_attrs.py:231`) fixed: the outside-stable-contract bullet now flags the wave-1 tier demotion inline, so a reader doesn't have to reconcile the two comment blocks.
- Nit (`_attrs.py:279-283`) fixed: the windowed-reads comment now describes the coverage by backend ("eager numpy, dask, GPU, HTTP, VRT") instead of hard-coding seven filenames.

No new findings on the re-pass. Shape and tier tests still pass (30/30).

PR #2353, opened 2026-05-24
Add release-gate tests for stable GeoTIFF features (#2340)

brendancol:
PR 6 of 6 in the GeoTIFF release contract epic (#2340). Adds one small, deterministic release-gate test per stable feature so a release engineer can run `pytest -m release_gate` and know the next release does not silently regress a stable promise.

## Summary

- New `release_gate` pytest marker registered in `setup.cfg`. Tests run by default; the marker exists so the release-gate subset can be selected with `pytest -m release_gate` (~4 seconds, 45 tests).
- One file per stable feature category from the epic.
- Each test asserts the contract from the user's point of view: round-trip pixels, `attrs['crs']`, `attrs['transform']`, and `attrs['nodata']` end to end.

## Coverage map

| Stable feature (per epic #2340) | New release-gate file |
|---|---|
| Local GeoTIFF read | `test_release_gate_local_read.py` |
| Local GeoTIFF write | `test_release_gate_local_write.py` |
| Stable lossless codecs (none, deflate, lzw, packbits, zstd) | `test_release_gate_codecs.py` |
| COG read + write for stable codecs | `test_release_gate_cog.py` |
| CRS / transform / nodata attrs round-trip | `test_release_gate_attrs_contract.py` |
| Windowed reads | `test_release_gate_windowed_read.py` |
| Dask reads parity vs eager | `test_release_gate_dask_parity.py` |

The codec file also pins a cross-check between its local `STABLE_LOSSLESS_CODECS` constant and the `codec.*` entries tagged `stable` in `SUPPORTED_FEATURES`, so a future tier change cannot silently desync the gate from the runtime tier table.

## Out of scope (other PRs in epic #2340)

- `SUPPORTED_FEATURES` edits (PR 1)
- Release-contract docs page (PR 2)
- Public docstring audit (PR 3)
- Opt-in enforcement for experimental / internal-only paths (PR 4)
- Unsupported-combination error messages (PR 5)

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_release_gate_local_read.py` -- 4 tests pass
- [x] `pytest xrspatial/geotiff/tests/test_release_gate_local_write.py` -- 4 tests pass
- [x] `pytest xrspatial/geotiff/tests/test_release_gate_codecs.py` -- 11 tests pass
- [x] `pytest xrspatial/geotiff/tests/test_release_gate_cog.py` -- 15 tests pass
- [x] `pytest xrspatial/geotiff/tests/test_release_gate_attrs_contract.py` -- 4 tests pass
- [x] `pytest xrspatial/geotiff/tests/test_release_gate_windowed_read.py` -- 4 tests pass
- [x] `pytest xrspatial/geotiff/tests/test_release_gate_dask_parity.py` -- 3 tests pass
- [x] `pytest -m release_gate xrspatial/geotiff/tests/` -- 45 tests pass, 5412 deselected, ~4 seconds
- [x] Pre-existing `test_release_gate_2321.py` still passes

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -112,6 +112,7 @@ filterwarnings =
     ignore:'asyncio.AbstractEventLoopPolicy' is deprecated:DeprecationWarning:pytest_asyncio
 markers =
     slow: long-running test cell (typical: golden-corpus fixtures behind a heavy codec or large pixel count). PR CI can skip with `-m "not slow"`; nightly / release runs use no filter. See xrspatial/geotiff/tests/golden_corpus/_marks.py for the corpus-side helper.
+    release_gate: locks a single stable feature in the GeoTIFF release contract (epic #2340). Always runs by default in CI; the marker exists so release engineers can run only these gates with `pytest -m release_gate` before tagging a release. Tests in this marker should be small, deterministic, and fail loudly if the contract breaks.
 
 [isort]
 line_length = 100
--- a/xrspatial/geotiff/tests/test_release_gate_attrs_contract.py
+++ b/xrspatial/geotiff/tests/test_release_gate_attrs_contract.py
@@ -1,0 +1,154 @@
+"""Release gate: CRS / transform / nodata attrs contract (epic #2340).
+
+The canonical attrs after a GeoTIFF read are tagged ``stable`` in the
+release gate checklist. The contract: every georeferenced read produces
+a DataArray whose ``attrs`` carry, at minimum, ``crs``, ``crs_wkt``,
+``transform``, ``georef_status``, the contract version stamp, and (when
+declared) ``nodata``. These attrs survive a write -> read round trip.
+
+This file is the single-shot release gate. Deep canonicalisation,
+alias handling, contract version bumps, and pass-through semantics are
+each covered by their own ``test_attrs_contract_*_1984.py`` files; here
+we lock the user-facing names and round-trip stability so the release
+notes can quote the canonical attrs without caveats.
+
+Out of scope:
+* Alias handling (``test_attrs_contract_aliases_1984.py``).
+* Attrs pass-through for user-supplied keys
+  (``test_attrs_contract_passthrough_1984.py``).
+* Contract version stamp bump policy
+  (``test_attrs_contract_version_1984.py``).
+"""
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+from xrspatial.geotiff import open_geotiff, to_geotiff
+from xrspatial.geotiff._geotags import GeoTransform
+from xrspatial.geotiff._writer import write
+
+
+# Keys that release notes are allowed to promise on every georeferenced
+# read. Adding a new key to the canonical set is a contract-version
+# bump (see issue #1984); removing one is a breaking change. Anything
+# else in the attrs (``masked_nodata``, ``nodata_pixels_present``,
+# ``raster_type``, etc.) is additive and not pinned here.
+CANONICAL_KEYS = (
+    "_xrspatial_geotiff_contract",
+    "crs",
+    "crs_wkt",
+    "transform",
+    "georef_status",
+)
+
+
+def _write_known_good(path: str, *, nodata: float | None = None) -> None:
+    arr = np.arange(16, dtype=np.float32).reshape(4, 4)
+    gt = GeoTransform(
+        origin_x=500000.0,
+        origin_y=4000000.0,
+        pixel_width=30.0,
+        pixel_height=-30.0,
+    )
+    write(
+        arr,
+        path,
+        geo_transform=gt,
+        crs_epsg=32610,
+        nodata=nodata,
+        compression="none",
+        tiled=False,
+    )
+
+
+@pytest.mark.release_gate
+def test_release_gate_attrs_canonical_keys_present(tmp_path) -> None:
+    """A georeferenced read carries every canonical attrs key."""
+    path = str(tmp_path / "release_gate_attrs_canonical_2340.tif")
+    _write_known_good(path)
+
+    da = open_geotiff(path)
+    missing = [k for k in CANONICAL_KEYS if k not in da.attrs]
+    assert not missing, (
+        "release gate: canonical attrs keys missing from a georeferenced "
+        f"read: {missing}; release notes promise every key in "
+        f"{list(CANONICAL_KEYS)}"
+    )
+
+
+@pytest.mark.release_gate
+def test_release_gate_attrs_georef_status_full(tmp_path) -> None:
+    """A fully-georeferenced read reports ``georef_status='full'``."""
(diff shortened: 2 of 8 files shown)


brendancol (commented):
## PR Review: Add release-gate tests for stable GeoTIFF features (#2340)

### Blockers

None.

### Suggestions

- `test_release_gate_dask_parity.py`: two of the three tests call `open_geotiff(path, chunks=8)` without first running `pytest.importorskip("dask")`. The third does. If dask ever moves out of the test environment, the unprotected two will error at collection while the third skips. Add the import-skip to all three for consistency, or hoist it to module scope.

### Nits

- `test_release_gate_codecs.py` line 119: `from xrspatial.geotiff import SUPPORTED_FEATURES` is inside the test body; the other release-gate files import at module top. Easy hoist.
- `_make_data_array` is duplicated between `test_release_gate_local_write.py` and `test_release_gate_cog.py`. The duplication is on purpose (each gate is self-contained), but worth flagging in case a follow-up wants to share it.

### What looks good

- One file per stable feature, each one small and fast — 45 tests in about 4 seconds.
- Assertion messages name the contract being checked, so a failure tells the reader which release-notes claim broke.
- `test_release_gate_codec_stable_set_matches_supported_features` pins the local stable-codec list against `SUPPORTED_FEATURES`, so a tier change can't desync the gate from the runtime constant.
- Temp filenames include `_2340.tif`, matching the project's parallel-worktree convention.
- The `release_gate` marker is registered in `setup.cfg` and tests run by default; `pytest -m release_gate` picks just the gates when a release engineer wants the short loop.

### Checklist

- [x] Tests assert documented contract behaviour, not just that the code runs.
- [x] Stable lossless codec set matches `SUPPORTED_FEATURES`.
- [x] Round-trip pixel, CRS, transform, and nodata checks present for every stable feature.
- [x] Windowed read transform origin shift is verified explicitly.
- [x] Dask parity test compares pixels and canonical attrs against the eager backend.
- [x] Temp filenames are unique across files and parametrize cases.
- [x] `release_gate` marker registered in `setup.cfg`.

brendancol (commented):
## Follow-up review

Disposition of the original review:

- **Suggestion (dask import-skip)**: Fixed. `pytest.importorskip("dask")` is now at module scope in `test_release_gate_dask_parity.py`, so all three tests skip uniformly when dask is absent.
- **Nit (SUPPORTED_FEATURES import)**: Fixed. Hoisted to module top in `test_release_gate_codecs.py`.
- **Nit (`_make_data_array` duplication)**: Dismissed. Each release-gate file is intentionally self-contained so a single file can be cited from the release-contract checklist without pulling in shared helpers. A follow-up can consolidate if it wants to.

No new findings on the updated diff. Marker registration, test isolation, and assertion-message specificity are unchanged.

brendancol:
## PR Review: Add release-gate tests for stable GeoTIFF features (#2340)

Test-only PR (7 new test files plus a marker registration in `setup.cfg`), so the algorithm / backend dispatch / NaN sections of the template don't apply. Review focuses on what the tests actually pin and whether anything is silently inconsistent across the seven files.

None.

- `xrspatial/geotiff/tests/test_release_gate_cog.py:33` redefines `STABLE_LOSSLESS_CODECS = ("none", "deflate", "lzw", "packbits", "zstd")` as a second copy of the constant in `test_release_gate_codecs.py:34`. The cross-check against `SUPPORTED_FEATURES` runs only against the codecs-file copy, so a future tier change that updates the codecs file but forgets the COG file would leave the COG gate parametrized on a stale list. Importing the constant from `test_release_gate_codecs` is the simpler fix.

- `xrspatial/geotiff/tests/test_release_gate_local_write.py:51-52` hard-codes the pixel-center y/x coords as plain literals. A one-line comment showing the derivation (`origin_y - half_pixel - i*pixel_height`) would save a maintainer from paging through the writer's coord-to-transform logic to verify them.
- `xrspatial/geotiff/tests/test_release_gate_attrs_contract.py:96-114` checks `isinstance(version, int)`. NumPy scalars (`np.int32`, `np.int64`) are not instances of `int` in CPython. If `_attrs.py` ever returns a numpy scalar here the test would fail with a confusing message. If the contract is "a Python int" then the current check is correct and worth a one-line comment; if it is "any integer", `numbers.Integral` is the broader check.
- The float codec tests in `test_release_gate_codecs.py:78-108` and the COG / windowed equivalents use `np.testing.assert_array_equal` for float pixels. Byte-exact is the right contract for a lossless codec, but the assertion message could say so explicitly so a future maintainer doesn't relax it to `allclose`.

### What looks good

- Every file has a focused module docstring naming the contract, the "out of scope" list, and the related deeper tests. A maintainer can read one file and know what it does and does not lock.
- The cross-file parity check at `test_release_gate_codecs.py:111-132` catches tier drift at the right place. Adding or removing a `codec.*` entry in `_attrs.py` will fail this assertion immediately.
- Temp file names include the issue number (`_2340.tif`), matching the project's unique-name convention so the 45-test suite cannot collide on tmpdir.
- The dask-skip hoist in the second commit is the right call. Collection-time errors in a dask-less env would have masked the parity claim entirely.
- The pixel patterns are non-constant (`row * 100 + col`, `arange` on distinctive grids), so row / col stride confusion still fails the equality check rather than passing on a symmetric input.
- The `release_gate` marker description in `setup.cfg` is enough for a release engineer to use without reading the PR.

### Checklist

- [x] Algorithm matches reference -- N/A (test-only PR)
- [x] All implemented backends produce consistent results -- dask parity covered; gpu / vrt explicitly out of scope
- [x] NaN handling is correct -- N/A (lossless byte-exact checks; nodata sentinel covered)
- [x] Edge cases are covered by tests -- in-bounds windows; out-of-bounds delegated to existing file
- [x] Dask chunk boundaries handled correctly -- N/A (read-side parity only)
- [x] No premature materialization or unnecessary copies -- intentional materialization for parity comparison
- [x] Benchmark exists or is not needed -- not needed (~4 s for 45 tests)
- [x] README feature matrix updated -- not needed (no new public function)
- [x] Docstrings present and accurate -- module docstring on every file; per-test docstrings explain the contract

brendancol:
## Re-review (post self-fix)

Followed up on the Suggestion in the previous review by importing `STABLE_LOSSLESS_CODECS` from `test_release_gate_codecs` in commit 49f46db8, so the COG file and the codecs file cannot drift on the stable codec list. Verified locally: 26 codec + COG release-gate tests pass.

The two Nits remain open and are intentionally cosmetic. The `isinstance(version, int)` check at `test_release_gate_attrs_contract.py:108` matches the producer side -- `_attrs.py:344` defines `_ATTRS_CONTRACT_VERSION = 4` as a Python `int`, so the current check is exact rather than narrow.

No further actionable findings. Holding for CI.

PR #2355, opened 2026-05-24
Fail loudly on unsupported GeoTIFF feature combinations (PR 5 of epic #2340)

### LINKED ISSUE #2349: Fail loudly on unsupported GeoTIFF feature combinations (PR 5 of epic #2340)
## Goal

This is PR 5 of 6 under epic #2340 (GeoTIFF release contract and feature tiering). Make unsupported feature combinations fail loudly with an actionable error that names the feature, names the offending input, and points at the release contract tier.

The epic enumerates the following as unsupported for this release:

- Full GDAL VRT parity (any VRT feature beyond simple mosaic assembly).
- Warped / reprojection VRTs.
- Rotated / sheared GeoTIFF write.
- Silent mixed-metadata flattening across stacked sources.

## Current Gaps

1. `parse_vrt` in `xrspatial/geotiff/_vrt.py` does not check the `subClass` attribute on `<VRTDataset>`. A `subClass="VRTWarpedDataset"` or `"VRTPansharpenedDataset"` is silently treated as a plain mosaic; the reader then processes whichever simple sources happen to be present, producing wrong output rather than a clear error.

2. `parse_vrt` silently skips unknown band children (`<KernelFilteredSource>`, `<PansharpeningOptions>`, `<DerivedRasterBand>`, etc.). The mosaic appears to read fine but the declared transform is no longer applied.

3. The VRT writer in `_vrt.write_vrt` enforces pixel size, dtype, band count, and CRS across sources, but silently picks `first['nodata']` and ignores per-source skewed transforms. A caller mosaicing tiles with disagreeing nodata sentinels loses the disagreement without any signal.

4. The VRT writer does not check `raster_type` (Area vs Point) across sources. Mismatched pixel registration silently flattens to the writer default.

5. The eager writer already rejects rotated `attrs['transform']`; the parity check should be extended into the GeoTransform-tuple path (`attrs['transform']` declared as a 6-tuple with non-zero `b` / `d`) which currently slips past at the entry point.

## Scope

For each unsupported case:
1. Detect it at the entry point (before any IO).
2. Raise an error that names the feature, the offending input, and points at the `SUPPORTED_FEATURES` tier map.
3. Add a regression test under `xrspatial/geotiff/tests/` asserting the error type and the substring identifying the feature.

Out of scope: experimental opt-in plumbing (PR 4), docstrings (PR 3), `SUPPORTED_FEATURES` audit (PR 1), docs page (PR 2).

## Acceptance Criteria

- Warped / pansharpened / processed / derived VRT subClass attributes are rejected at parse time with a clear error.
- Unknown VRT band children that would alter the read result are rejected (no silent skip).
- Mosaic writer rejects mixed per-source nodata sentinels and mixed raster_type rather than flattening.
- Mosaic writer rejects sources whose source transform has non-zero skew terms.
- Eager writer rejects a rotated 6-tuple `attrs['transform']` at the entry point.
- One regression test per case, asserting both the error type and the feature-naming substring.

brendancol:
Closes #2349. Parent epic #2340 (PR 5 of 6).

## Summary

Surface a typed error at the entry point for the unsupported feature combinations listed in the release contract, instead of letting them slip through as silent no-ops or wrong output.

- New cases rejected at parse time: VRTDataset subClass (Warped / Pansharpened / Processed / Derived), VRTRasterBand subClass (DerivedRasterBand), and unknown VRTRasterBand child elements (KernelFilteredSource, MaskBand, PansharpeningOptions, pixel-function tags, plus a catch-all for unrecognised tags). Known informational children keep passing.
- New cases rejected at write time: `write_vrt` refuses sources with a rotated source transform, refuses cross-source AREA_OR_POINT mismatch, and refuses cross-source nodata mismatch unless the caller pins the mosaic nodata via the `nodata` kwarg.
- New `UnsupportedGeoTIFFFeatureError` exported from `xrspatial.geotiff`. Subclasses `ValueError` so existing `except ValueError` callers keep catching the case.
- Each message names the feature, names the offending input (kwarg / source path / band number), and points at `SUPPORTED_FEATURES` plus epic #2340.

## Out of scope

- Experimental opt-in plumbing (PR 4).
- Docstring tier audits (PR 3).
- `SUPPORTED_FEATURES` reshape (PR 1).
- Docs page (PR 2).

## Backend coverage

- numpy / cupy / dask+numpy / dask+cupy share `parse_vrt` and `write_vrt`; the new gates run before backend dispatch, so all four hit the same error.

## Test plan

- [x] New regression tests under `xrspatial/geotiff/tests/test_unsupported_features_2349.py` cover each rejected case plus the existing rotated 6-tuple writer refusal and the rotated VRT GeoTransform read refusal.
- [x] Full VRT test set (`-k vrt`) still passes (709 tests).
- [x] Full rotated test set (`-k rotated`) still passes (99 tests).
- [x] Writer / drop-rotation / VRT-tiled-validation suites still pass.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -73,7 +73,8 @@ from ._coords import \
 from ._crs import _resolve_crs_to_wkt, _wkt_to_epsg  # noqa: F401
 from ._errors import (ConflictingCRSError, ConflictingNodataError, GeoTIFFAmbiguousMetadataError,
                       InvalidCRSCodeError, MixedBandMetadataError, NonUniformCoordsError,
-                      RotatedTransformError, UnknownCRSModelTypeError, UnparseableCRSError)
+                      RotatedTransformError, UnknownCRSModelTypeError, UnparseableCRSError,
+                      UnsupportedGeoTIFFFeatureError)
 from ._geotags import RASTER_PIXEL_IS_AREA, RASTER_PIXEL_IS_POINT, GeoTransform  # noqa: F401
 from ._reader import _MAX_CLOUD_BYTES_SENTINEL, UnsafeURLError
 from ._reader import read_to_array as _read_to_array
@@ -115,6 +116,7 @@ __all__ = [
     'UnknownCRSModelTypeError',
     'UnparseableCRSError',
     'UnsafeURLError',
+    'UnsupportedGeoTIFFFeatureError',
     'open_geotiff',
     'read_geotiff_gpu',
     'read_geotiff_dask',
--- a/xrspatial/geotiff/_errors.py
+++ b/xrspatial/geotiff/_errors.py
@@ -118,6 +118,22 @@ class UnknownCRSModelTypeError(GeoTIFFAmbiguousMetadataError):
     """
 
 
+class UnsupportedGeoTIFFFeatureError(ValueError):
+    """Caller asked for a feature this release does not implement (#2349).
+
+    Raised at the read or write entry point when the input declares a
+    feature the GeoTIFF module does not support (warped / reprojection
+    VRTs, pansharpened / processed / derived VRT subclasses, unknown
+    VRT band children, source transforms with non-zero skew on a VRT
+    mosaic). The message names the feature and points the caller at
+    :data:`xrspatial.geotiff.SUPPORTED_FEATURES` for the full tier map.
+
+    Subclasses ``ValueError`` so existing ``except ValueError`` callers
+    keep catching the case; new code can ``except`` this class to
+    distinguish "we refuse this input" from "the input is malformed".
+    """
+
+
 __all__ = [
     "GeoTIFFAmbiguousMetadataError",
     "InvalidCRSCodeError",
@@ -128,4 +144,5 @@ __all__ = [
     "ConflictingCRSError",
     "ConflictingNodataError",
     "UnknownCRSModelTypeError",
+    "UnsupportedGeoTIFFFeatureError",
 ]
--- a/xrspatial/geotiff/_vrt.py
+++ b/xrspatial/geotiff/_vrt.py
@@ -346,6 +346,60 @@ def _parse_band_nodata(text: str | None,
         return None
 
 
+# VRTRasterBand child elements that read_vrt explicitly supports or
+# tolerates. Hoisted to module scope so the parse_vrt band loop does
+# not rebuild the frozensets on every iteration (epic #2340 / PR 5).
+#
+# ``_INFORMATIONAL_BAND_TAGS`` -- pure-metadata children that have no
+# effect on the assembled array. Silently ignored.
+#
+# ``_SOURCE_BAND_TAGS`` -- the source types whose pixel-placement
+# semantics this reader implements.
+#
+# ``_UNSUPPORTED_BAND_TAGS`` -- known GDAL VRT children that alter the
+# computed output (kernel filters, derived band expressions, mask /
+# pansharpening sources). Rejected with ``UnsupportedGeoTIFFFeatureError``
+# so a VRT relying on them cannot read as a plain mosaic and lose its
+# declared computation. Anything that is none of the three -- e.g. a
+# future GDAL VRT extension -- also raises, via the catch-all branch in
+# ``parse_vrt``.
+_INFORMATIONAL_BAND_TAGS = frozenset({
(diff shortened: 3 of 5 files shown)


brendancol (commented):
## PR Review: Fail loudly on unsupported GeoTIFF feature combinations (PR 5 of epic #2340)

### Blockers
None.

### Suggestions
- [ ] Dataset-level `<MaskBand>` slips through. Per the GDAL VRT spec, `<MaskBand>` sits as a child of `<VRTDataset>` (sibling of `<VRTRasterBand>`), not inside a band. The scan at `_vrt.py:504-534` only walks band children, so a VRT carrying a dataset-level mask band still silently drops the mask. Putting `MaskBand` in `_UNSUPPORTED_BAND_TAGS` does not help. Either sweep the dataset root for `MaskBand` / `GCPList` after the `subClass` check at `_vrt.py:390`, or note that the gap is deliberate.

- [ ] The three frozensets at `_vrt.py:490-502` are rebuilt on every band iteration. Hoist them to module scope.

- [ ] `<OverviewList>` and `<Overview>` (band-level VRT overview declarations) are in none of the three sets. The new "raise on unknown" branch at `_vrt.py:520-534` will now reject a VRT that declares external overviews on a band -- the previous parser ignored those. If ignoring was intentional, add both tags to `_INFORMATIONAL_BAND_TAGS`. If rejecting is intentional, add a regression test.

### Nits
- [ ] The new write-side checks at `_vrt.py:1705-1828` are interleaved with the legacy pixel-size / dtype / CRS checks. A helper or even a comment block grouping the new gates would make the cross-source policy easier to follow. Not a correctness issue.

- [ ] Error messages embed `` :data:`xrspatial.geotiff.SUPPORTED_FEATURES` `` in raw RST. Callers see this in tracebacks, not in rendered Sphinx output, so the `:data:` role is dead weight. Plain `` `xrspatial.geotiff.SUPPORTED_FEATURES` `` reads better.

### What looks good
- 13 tests cover each rejected case plus regression pins for the existing rotated-write and rotated-VRT-read paths.
- `UnsupportedGeoTIFFFeatureError` subclasses `ValueError`, so existing `except ValueError` callers keep catching the case.
- The mixed-nodata gate has an explicit `nodata=` opt-out and the test covers it.
- Each new check has a comment naming the silent-failure mode it replaces.
- The rotated 6-tuple writer refusal and the rotated VRT GeoTransform read refusal both pick up regression pins.

### Checklist
- [x] VRT subClass attribute and GeoTransform skew terms match the GDAL VRT spec.
- [x] All four backends share `parse_vrt` and `write_vrt`, so the new gates hit uniformly.
- [x] No NaN-handling regressions; the new checks are metadata-only.
- [x] Edge cases covered (asymmetric nodata, derived band subClass, unknown band child).
- [x] No premature materialization; the gates are O(num_sources) attribute reads.
- [n/a] Benchmark not needed (validation-only, runs before IO).
- [n/a] README feature matrix unchanged (no new functions added).
- [x] Docstring present on `UnsupportedGeoTIFFFeatureError`.

brendancol (commented):
## Follow-up review (post-revision)

Re-reviewed after the address-review commits (510e515f, 0ff6f064, 88e59046).

- Dataset-level `<MaskBand>` / `<GCPList>` / `<PansharpeningOptions>` sweep -- fixed at `_vrt.py:458-475`. Two new regression tests pin the rejection.
- Frozensets hoisted to module scope -- fixed at `_vrt.py:366-400`.
- `<OverviewList>` and `<Overview>` band children kept informational -- fixed at `_vrt.py:376`. Regression test added.
- Write-side checks grouped into helpers (`_check_no_rotated_source_transforms`, `_check_no_mixed_raster_type`, `_check_no_mixed_nodata`) -- done at `_vrt.py:1646-1749`.
- `:data:` Sphinx role dropped from error messages -- done.

### New findings from this pass

#### Self-caught while re-reviewing
Two sources both declaring NaN nodata would have hit the new mixed-nodata gate falsely because `float('nan') != float('nan')` is True in plain Python. Fixed in 88e59046 by routing the comparison through `_nodata_values_agree` (`math.isnan`-aware). Regression test pins the round-trip.

### Blockers
None.

### Suggestions
None remaining.

### Nits
None remaining.

### Verification
- 17 tests in `test_unsupported_features_2349.py` pass.
- Full `xrspatial/geotiff/tests/` suite: 5358 passed, 68 skipped, 1 xfailed, 1 xpassed.

Ready for CI. No outstanding review items.

Issue #2341, opened 2026-05-24
Epic: GeoTIFF correctness and backend parity release gate


brendancol:
## Goal

Make the promised GeoTIFF paths resistant to silent wrongness by asserting parity across pixels, coordinates, and metadata attrs.

The highest release risk is not an obvious crash. It is a plausible-looking raster with wrong `transform`, `crs`, `nodata`, `georef_status`, or coords that downstream spatial functions trust.

## Scope

This epic covers correctness gates for the release-promised paths:

- Local eager GeoTIFF reads/writes.
- COG reads/writes for stable codecs.
- Windowed reads.
- Dask reads where promised.
- Sidecar overview reads where promised.
- VRT simple mosaics where promised by the VRT epic.
- GPU only as experimental smoke/parity checks, not as stable release blocker unless it affects CPU behavior.

## Priority Risks

- Pixels match but attrs or coords are wrong.
- Eager and dask paths disagree.
- GPU fallback paths produce different metadata than CPU paths.
- Windowed reads return unshifted transforms.
- Overview reads lose CRS/transform/nodata metadata.
- Nodata masking/promotion differs across backends.
- Integer nodata, NaN nodata, MinIsWhite, and masked_nodata lifecycle drift.

## Work Items

- Define a small `geotiff_release_gate` test set that runs in normal CI.
- Add helper assertions for full raster equivalence:
  - values
  - dims
  - coords
  - `attrs['transform']`
  - `attrs['crs']` / `attrs['crs_wkt']`
  - `attrs['nodata']`
  - `attrs['masked_nodata']`
  - `attrs['georef_status']`
  - `attrs['raster_type']`
  - relevant rich metadata when promised
- Add eager vs dask parity tests for representative files.
- Add windowed-read tests that assert shifted coords and transform.
- Add overview/sidecar tests that assert metadata parity, not just pixel parity.
- Add read/write/read round-trip tests for stable codecs.
- Add negative tests for ambiguous metadata that must fail closed.

## Non-Goals

- Exhaustive codec matrix in the release gate.
- Making GPU a stable backend.
- Full GDAL/rasterio parity for every edge case.

## Acceptance Criteria

- A stable-path regression cannot pass by returning correct pixel values with wrong metadata.
- Eager and dask promised paths agree on values, coords, dims, and attrs.
- Windowed reads have correct shape, coords, and shifted transform.
- Overview and sidecar reads preserve promised geospatial metadata.
[...]

PR #2362, opened 2026-05-24
Release gate: eager-vs-dask raster equivalence (#2357)

### LINKED ISSUE #2357: Release-gate: eager-vs-dask raster equivalence for representative GeoTIFFs (PR 1 of 5 of epic #2341)
## Reason or Problem

Epic #2341 calls out that the highest release risk for the GeoTIFF surface is pixels matching while attrs, coords, or dims silently disagree across backends. Today, the eager (`open_geotiff`) and dask (`read_geotiff_dask`) entry points both promise stable behavior, but no single regression test asserts full raster equivalence — pixels + dims + coords + the seven release-attr keys — across the two paths on the same files.

## Proposal

Add `xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py`.

For each file in a representative-corpus list (integer dtype + nodata, float dtype + NaN nodata, MinIsWhite, masked_nodata-lifecycle), open it eagerly and via `read_geotiff_dask`, materialize the dask result, and assert all of:

- `.values` bit-exact (NaN-aware)
- `.dims` equal
- `.coords` element-wise equal under `xr.testing.assert_identical`-style strictness
- `attrs['transform']`, `attrs['crs']`, `attrs['crs_wkt']`, `attrs['nodata']`, `attrs['masked_nodata']`, `attrs['georef_status']`, `attrs['raster_type']` all equal

The assertions are inlined as a small helper inside the test module (no new shared helper file — keeps the PR self-contained; consolidation across the PR-2..PR-5 sibling tests can be a follow-up). Add a row to ``docs/source/reference/release_gate_geotiff.rst`` citing the new test file under \"Local GeoTIFF read and write\".

## Stakeholders and Impacts

Touches: one new test file under `xrspatial/geotiff/tests/`, one row added to the release-gate doc page. No production code changes. Runtime: bounded to the existing corpus fixtures already used by `test_backend_pixel_parity_matrix_1813.py`.

Out of scope (sibling PRs of epic #2341):

- Windowed-read shifted-transform parity (PR 2).
- Overview / sidecar metadata survival (PR 3).
- Stable-codec round-trip (PR 4).
- Negative tests for ambiguous metadata (PR 5).

## Drawbacks

Slight overlap with `test_backend_full_parity_2211.py`, which already checks attrs finalization parity. This PR's test adds the eager-vs-dask cross-backend axis on top, and pins it under a `release_gate`-themed name so the release checklist row points at one test, not three.

## Alternatives

- Extend `test_backend_full_parity_2211.py` instead of adding a new file. Rejected: the release-gate doc row needs a stable test path to cite; reusing an existing file couples the row to unrelated future edits.
- Make the assertions a shared helper module up front. Rejected: that requires sequential ordering across the five sibling PRs; inlining keeps PRs 1-5 independent.

Parent epic: #2341. Sibling issues to be created for PRs 2-5.

brendancol:
Closes #2357.

PR 1 of 5 of epic #2341.

## Summary

- Adds `xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py`. Reads each of four representative corpus fixtures (integer-nodata, float-NaN-nodata, MinIsWhite, masked-nodata-lifecycle) once through `open_geotiff` and once through `read_geotiff_dask`, then asserts pixel values (NaN-aware), `dims`, `coords` (dtype + bytes per axis), and the seven release-attr keys (`transform`, `crs`, `crs_wkt`, `nodata`, `masked_nodata`, `georef_status`, `raster_type`) all match.
- Helpers are inlined in the test module rather than pulled into a new shared helper file; the issue spec calls for keeping PRs 1-5 independent.
- Adds one row under "Local GeoTIFF read and write" in `docs/source/reference/release_gate_geotiff.rst` citing the new test file.

## Backend coverage

Eager numpy (`open_geotiff`) and dask numpy (`read_geotiff_dask`). GPU and dask+GPU are out of scope: the issue scopes this PR to the two stable backends.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py -v` -- 5 passed locally.
- [ ] CI green on the same file across the supported Python / NumPy matrix.

### DIFF
--- a/docs/source/reference/release_gate_geotiff.rst
+++ b/docs/source/reference/release_gate_geotiff.rst
@@ -71,6 +71,15 @@ Local GeoTIFF read and write
        coords, and ``attrs`` as the eager numpy read.
      - ``xrspatial/geotiff/tests/test_backend_parity_matrix.py``,
        ``xrspatial/geotiff/tests/test_backend_full_parity_2211.py``
+   * - ``reader.eager_dask_parity``
+     - stable
+     - ``open_geotiff(path)`` and ``read_geotiff_dask(path)`` return the
+       same pixels, ``dims``, ``coords``, and the seven release-attr
+       keys (``transform``, ``crs``, ``crs_wkt``, ``nodata``,
+       ``masked_nodata``, ``georef_status``, ``raster_type``) across
+       integer-nodata, float-NaN-nodata, MinIsWhite, and
+       masked-nodata-lifecycle fixtures.
+     - ``xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py``
    * - ``writer.local_file``
      - stable
      - ``to_geotiff`` writes a file that ``open_geotiff`` reads back
--- a/docs/source/reference/release_gate_geotiff.rst
+++ b/docs/source/reference/release_gate_geotiff.rst
@@ -77,8 +77,9 @@ Local GeoTIFF read and write
        same pixels, ``dims``, ``coords``, and the seven release-attr
        keys (``transform``, ``crs``, ``crs_wkt``, ``nodata``,
        ``masked_nodata``, ``georef_status``, ``raster_type``) across
-       integer-nodata, float-NaN-nodata, MinIsWhite, and
-       masked-nodata-lifecycle fixtures.
+       four scenarios: integer-nodata, float-NaN-nodata, MinIsWhite,
+       and the ``mask_nodata=False`` raw-sentinel branch of the
+       nodata lifecycle.
      - ``xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py``
    * - ``writer.local_file``
      - stable
--- a/xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py
+++ b/xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py
@@ -86,10 +86,10 @@ _RELEASE_ATTR_KEYS: tuple[str, ...] = (
 # Corpus selection
 # ---------------------------------------------------------------------------
 
-# One ``pytest.param`` per fixture scenario. ``open_kwargs`` carries any
-# extra kwargs (e.g. ``mask_nodata=True``) applied to both the eager and
-# dask reads so the masked-nodata-lifecycle row exercises the same
-# masking semantics on both paths.
+# One ``pytest.param`` per fixture scenario. ``open_kwargs`` carries
+# any extra kwargs (e.g. ``mask_nodata=False``) applied to both the
+# eager and dask reads so the masked-nodata-lifecycle row exercises
+# the same masking semantics on both paths.
 _CORPUS = [
     pytest.param(
         "nodata_int_sentinel_uint16",
--- a/xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py
+++ b/xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py
@@ -30,8 +30,10 @@ The corpus covers the four scenarios called out in the issue:
 * integer dtype with explicit integer nodata sentinel
 * float dtype with NaN nodata
 * MinIsWhite photometric (no explicit nodata tag)
-* masked-nodata lifecycle (``mask_nodata=True`` upcasts int to float and
-  rewrites the sentinel to NaN)
+* masked-nodata lifecycle: the same integer-sentinel fixture read with
+  ``mask_nodata=False`` so the raw uint sentinel branch is pinned in
+  parity against the default ``mask_nodata=True`` branch (which the
+  integer-nodata row above already covers)
 
 Out of scope (sibling PRs of epic #2341):
 
@@ -104,9 +106,14 @@ _CORPUS = [
         {},
         id="miniswhite",
     ),
+    # ``mask_nodata=False`` is the contrast cell to the first row's
+    # default ``mask_nodata=True``: the raw uint16 sentinel is preserved
+    # and ``masked_nodata`` flips to ``False``. Together the two cells
+    # pin both sides of the nodata lifecycle on the same fixture, which
+    # is the silent-disagreement case the issue calls out.
     pytest.param(
         "nodata_int_sentinel_uint16",
-        {"mask_nodata": True},
+        {"mask_nodata": False},
         id="masked-nodata-lifecycle",
     ),
 ]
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Release gate: eager-vs-dask raster equivalence (#2357)

None.

- [ ] `xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py:108-111` -- the `masked-nodata-lifecycle` row passes `{"mask_nodata": True}`, but `True` is already the default on both `open_geotiff` and `read_geotiff_dask`, so this row reads the fixture with the same effective kwargs as the `int-dtype-nodata` row at line 92. The two cells exercise an identical code path and pass / fail together. To make the fourth scenario distinct (and to actually cover the "lifecycle" both ways), pass `{"mask_nodata": False}` here so the row pins the raw-sentinel branch in parity against the masked branch from the first row. Verified locally: with `mask_nodata=True` the read returns float64 with `masked_nodata=True`; with `mask_nodata=False` it returns uint16 with `masked_nodata=False`. Today the matrix only covers the masked branch on the int fixture.

- [ ] `xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py:210-224` -- `_attr_equal` falls back to `==` for the `transform` 6-tuple of floats. The sibling parity gate `test_backend_full_parity_2211.py:608-623` allows a 1e-9 ULP tolerance on the transform tuple for the same reason: float coords reconstructed through different code paths can pick up sub-ULP drift even when the on-disk values are identical. Bit-exact works today and is the strongest signal, so keeping `==` is defensible; consider mirroring the ULP tolerance only if this gate starts flaking on float drift. Flagging so the next reader of this file knows the divergence from the sibling gate is intentional rather than an oversight.

- [ ] `xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py:119-127` -- `_materialise` is only called inside `_assert_values_equal`; it could be inlined to drop one level of indirection. Keeping it separate is fine too if the intent is to give a stable hook for the sibling PRs of epic #2341 to reuse via copy-paste; if so, a one-line comment saying as much would make that intent visible.

- [ ] `docs/source/reference/release_gate_geotiff.rst:81` -- "masked-nodata-lifecycle fixtures" is plural, but in practice the corpus reuses the `nodata_int_sentinel_uint16` fixture with different kwargs rather than carrying a separate masked-lifecycle fixture file. Rewording to "across four scenarios (integer-nodata, float-NaN-nodata, MinIsWhite, masked-nodata-lifecycle)" tracks the test more literally.

### What looks good

- The `@pytest.mark.release_gate` mark is the right home for this gate and the marker is already registered in `setup.cfg:115`.
- The seven release-attr keys match the list called out in the issue spec.
- The `_attr_equal` helper handles NaN sentinels, ndarray attrs, and nested tuples / lists; the NaN branch is essential for the `nodata_nan_float32` row.
- `test_release_gate_corpus_is_non_empty` is a useful sentinel against a future refactor that empties out the parametrize list and lets the matrix pass vacuously.
- Helpers are inlined per the issue's "no shared helper module in this PR" constraint, keeping the PR independent of the four sibling PRs.

### Checklist

- [x] Algorithm matches the issue spec (pixels + dims + coords + seven attrs)
- [x] Both implemented backends (eager numpy, dask numpy) produce consistent results on the in-tree corpus
- [x] NaN handling correct (`equal_nan=True` for floats, NaN-aware attr comparison)
- [x] Edge cases covered for the four scenarios listed in the issue (modulo the `mask_nodata=True` redundancy noted above)
- [x] Dask chunk size is sane for the corpus fixtures (32 over 64x64 max)
- [x] No premature materialisation or unnecessary copies
- [x] No new benchmarks needed (test-only PR)
- [x] Docstrings present on module, helpers, and the parametrized test

brendancol (commented):
## Follow-up review: Release gate: eager-vs-dask raster equivalence (#2357)

Second pass after the review fixes in commits `94d36bf5` and `5afc20b0`.

- **Fixed:** `masked-nodata-lifecycle` row now passes `mask_nodata=False`, so it is the contrast cell to the default-`True` `int-dtype-nodata` row rather than a duplicate of it. The fixture surface now pins both sides of the nodata lifecycle. Verified locally: 5/5 still pass after the kwarg flip.
- **Dismissed with rationale in code:** the `transform` 6-tuple comparison stays bit-exact via `==` (no ULP tolerance). The new docstring on `_attr_equal` explains the divergence from the sibling gate in `test_backend_full_parity_2211.py` and why it is intentional for the same-file eager-vs-dask axis. If this gate ever flakes on float drift, the comment points the next reader at the knob to turn.
- **Fixed:** `_materialise` now carries a sentence noting it is kept as a named helper for symmetry across the four sibling PRs of epic #2341. The hook is documented rather than mysterious.
- **Fixed:** the doc row at `release_gate_geotiff.rst:74` now reads "across four scenarios: integer-nodata, float-NaN-nodata, MinIsWhite, and the ``mask_nodata=False`` raw-sentinel branch of the nodata lifecycle". Tracks the parametrize matrix literally.
- **Follow-on nit fix (commit `5afc20b0`):** the comment above the `_CORPUS` literal had a stale `mask_nodata=True` example; updated to `mask_nodata=False` to match the actual row.

### Remaining findings

None.

### Checklist

- [x] All review actions either applied or dismissed in code with a rationale.
- [x] All 5 tests still pass locally.
- [x] No new blockers, suggestions, or nits surfaced by the second pass.

brendancol:
CI status note: the macOS `run (macos-latest, 3.14)` job failed with 4 test failures, and the ubuntu / windows jobs were auto-cancelled by the matrix-strategy fail-fast after macOS went red. The 4 failing tests are pre-existing on `main` and unrelated to this PR:

- `xrspatial/geotiff/tests/test_unsupported_features_2349.py::test_vrt_with_skewed_geotransform_rejected`
- `xrspatial/geotiff/tests/test_vrt_metadata_parity_2321.py::test_unsupported_resample_alg_raises`
- `xrspatial/geotiff/tests/test_vrt_metadata_parity_2321.py::test_negative_srcrect_size_rejected`
- `xrspatial/geotiff/tests/test_vrt_metadata_parity_2321.py::test_negative_dstrect_size_rejected`

Verified by checking out `origin/main` (`9c40df49`) into a separate worktree and running the same four nodeids -- all four fail with identical error messages there. The failures are regex / wording mismatches between the test expectations and the current error messages in those VRT validators; they are out of scope for this PR (which only adds `test_release_gate_eager_dask_parity_2341.py` and one doc row).

The `run (3.12)` job and the second `run (ubuntu-latest, 3.14)` job both passed, and the new test file's 5 cells pass locally:

```
xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py::test_release_gate_eager_dask_full_parity[int-dtype-nodata] PASSED
xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py::test_release_gate_eager_dask_full_parity[float-dtype-nan-nodata] PASSED
xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py::test_release_gate_eager_dask_full_parity[miniswhite] PASSED
xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py::test_release_gate_eager_dask_full_parity[masked-nodata-lifecycle] PASSED
xrspatial/geotiff/tests/test_release_gate_eager_dask_parity_2341.py::test_release_gate_corpus_is_non_empty PASSED
```

Flagging rather than papering over -- the 4 unrelated failures should be triaged in a separate PR.

PR #2365, opened 2026-05-24
Add stable-codec read/write/read round-trip release gate (#2360)

### LINKED ISSUE #2360: Release-gate: stable-codec read/write/read round-trip (PR 4 of 5 of epic #2341)
## Reason or Problem

Epic #2341 calls for read/write/read round-trip tests for stable codecs as part of the release gate. The existing `test_compression.py` and `test_supported_features_tiers_2137.py` cover codec behaviour in pieces, but no single test pins the joint contract \"on stable codecs, a round-trip preserves both bit-exact pixels AND every release-attr key\" across the full set of stable codecs and the dtypes that matter.

## Proposal

Add `xrspatial/geotiff/tests/test_release_gate_codec_round_trip_2341.py`.

Parametrize over the cartesian product of:

- Stable codecs from ``SUPPORTED_FEATURES``: `none`, `deflate`, `lzw`, `zstd`, `packbits`.
- Dtypes: `int16`, `int32`, `float32`, `float64`.

For each combination:

- Construct a small DataArray in memory with a known transform, CRS, and nodata value (NaN for float, sentinel for int).
- Write it via `to_geotiff(path, compress=codec)`.
- Read it back via `open_geotiff(path)`.
- Assert `.values` byte-exact (NaN-aware for float).
- Assert `attrs['transform']`, `attrs['crs']`, `attrs['crs_wkt']`, `attrs['nodata']`, `attrs['masked_nodata']`, `attrs['georef_status']`, `attrs['raster_type']` match the originals.
- Assert `attrs['compression']` (or whatever the canonical attr key is) reflects the requested codec.

Assertions inlined per-file. Add a row to ``docs/source/reference/release_gate_geotiff.rst`` under \"Local GeoTIFF read and write\" citing the new file.

## Stakeholders and Impacts

Touches one new test file. No production code change unless the round-trip surfaces a real bug.

Out of scope: experimental codecs (LERC, JPEG2000, J2K, LZ4) — already gated and tested by `test_supported_features_tiers_2137.py`. Internal-only JPEG — same.

## Drawbacks

Marginal test runtime increase (5 codecs * 4 dtypes = 20 round-trips). Each round-trip uses a small (e.g. 128x128) raster, so the absolute cost is small.

## Alternatives

- Reuse `test_compression.py` instead of adding a new file. Rejected for the same reason as PR 1: the release-gate doc row needs a stable test path to cite.

Parent epic: #2341. PR 4 of 5.

brendancol:
Closes #2360. PR 4 of 5 of epic #2341.

## Summary
- Adds `xrspatial/geotiff/tests/test_release_gate_codec_round_trip_2341.py`. It walks the cartesian product of stable codecs (`none`, `deflate`, `lzw`, `zstd`, `packbits`) and dtypes (`int16`, `int32`, `float32`, `float64`).
- Each parametrized case runs a write / read / write / read cycle and asserts byte-exact pixels (NaN-aware for float) plus every canonical release attr (`transform`, `crs`, `crs_wkt`, `nodata`, `masked_nodata`, `georef_status`, `raster_type`). It also checks the on-disk TIFF Compression tag matches the codec requested.
- Adds a row to `docs/source/reference/release_gate_geotiff.rst` under "Local GeoTIFF read and write" pointing at the new file.
- A drift guard fails if the codec list in the test gets out of sync with `stable` codecs in `SUPPORTED_FEATURES`.

## Test plan
- [x] `pytest xrspatial/geotiff/tests/test_release_gate_codec_round_trip_2341.py` (21 passed)

### DIFF
--- a/docs/source/reference/release_gate_geotiff.rst
+++ b/docs/source/reference/release_gate_geotiff.rst
@@ -102,6 +102,16 @@ Local GeoTIFF read and write
      - Lossless byte-for-byte round-trip on integer and float dtypes.
      - ``xrspatial/geotiff/tests/test_supported_features_tiers_2137.py``,
        ``xrspatial/geotiff/tests/test_compression.py``
+   * - Stable codec round-trip (read / write / read)
+     - stable
+     - On every stable codec (``none``, ``deflate``, ``lzw``, ``zstd``,
+       ``packbits``) crossed with every promised dtype (``int16``,
+       ``int32``, ``float32``, ``float64``), a write / read / write /
+       read cycle preserves both byte-exact pixels (NaN-aware for
+       float) and every canonical release attr (``transform``, ``crs``,
+       ``crs_wkt``, ``nodata``, ``masked_nodata``, ``georef_status``,
+       ``raster_type``).
+     - ``xrspatial/geotiff/tests/test_release_gate_codec_round_trip_2341.py``
    * - Codec ``lerc`` / ``jpeg2000`` / ``j2k`` / ``lz4``
      - experimental
      - Rejected by default; accepted with
--- a/docs/source/reference/release_gate_geotiff.rst
+++ b/docs/source/reference/release_gate_geotiff.rst
@@ -104,13 +104,10 @@ Local GeoTIFF read and write
        ``xrspatial/geotiff/tests/test_compression.py``
    * - Stable codec round-trip (read / write / read)
      - stable
-     - On every stable codec (``none``, ``deflate``, ``lzw``, ``zstd``,
-       ``packbits``) crossed with every promised dtype (``int16``,
-       ``int32``, ``float32``, ``float64``), a write / read / write /
-       read cycle preserves both byte-exact pixels (NaN-aware for
-       float) and every canonical release attr (``transform``, ``crs``,
-       ``crs_wkt``, ``nodata``, ``masked_nodata``, ``georef_status``,
-       ``raster_type``).
+     - For every stable codec * promised dtype combination, a full
+       write / read / write / read cycle preserves byte-exact pixels
+       (NaN-aware for float) and the canonical release attrs. See
+       the cited test for the codec, dtype, and attr-key matrix.
      - ``xrspatial/geotiff/tests/test_release_gate_codec_round_trip_2341.py``
    * - Codec ``lerc`` / ``jpeg2000`` / ``j2k`` / ``lz4``
      - experimental
--- a/xrspatial/geotiff/tests/test_release_gate_codec_round_trip_2341.py
+++ b/xrspatial/geotiff/tests/test_release_gate_codec_round_trip_2341.py
@@ -1,0 +1,339 @@
+"""Release gate: stable-codec read/write/read round-trip (epic #2341).
+
+PR 4 of 5 of epic #2341. The release contract names a specific set of
+codecs as ``stable`` in :data:`xrspatial.geotiff.SUPPORTED_FEATURES`:
+``none``, ``deflate``, ``lzw``, ``zstd``, ``packbits``. The release
+notes promise that on any of these codecs, a round-trip preserves both
+bit-exact pixels AND every canonical release attr key, on every dtype
+the library promises to round-trip.
+
+Existing tests split the contract:
+
+* ``test_compression.py`` covers codec internals (LZW dictionary edge
+  cases, PackBits boundary cases, deflate stream framing).
+* ``test_supported_features_tiers_2137.py`` pins the
+  ``SUPPORTED_FEATURES`` tier table.
+* ``test_release_gate_codecs.py`` pins lossless pixel round-trip for
+  two dtypes (``uint16``, ``float32``).
+
+This file is the joint gate: the cartesian product of every stable
+codec with every promised dtype, asserting both pixel equality AND
+release-attr equality through a full read/write/read cycle.
+
+Out of scope:
+
+* Experimental codecs (``lerc``, ``jpeg2000``, ``j2k``, ``lz4``) --
+  release tier is ``experimental``; covered by
+  ``test_supported_features_tiers_2137.py``.
+* Internal-only ``jpeg`` -- not part of the public surface.
+* COG layout (``test_release_gate_cog.py``).
+* Backend parity (``test_backend_parity_matrix.py``).
+"""
+from __future__ import annotations
+
+import uuid
+
+import numpy as np
+import pytest
+import xarray as xr
+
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Add stable-codec read/write/read round-trip release gate (#2360)

This is a test-only PR. No production code changed. The review focuses on test correctness, attr coverage, and the docs row.

(none)

- [ ] `test_release_gate_codec_round_trip_2341.py:268-270` -- Pass 2 writes the baseline DataArray with `nodata=baseline.attrs.get("nodata")`, but the float-dtype reader stamps `attrs["nodata"]` as float NaN even when the dtype is float32/float64. That happens to round-trip fine with the writer, but if a future change makes the writer reject `nodata=NaN` (since NaN is already implicit for float), this test starts failing for reasons unrelated to the codec. Consider only passing the `nodata=` kwarg explicitly for the integer branch.
- [ ] `test_release_gate_codec_round_trip_2341.py:113-117` -- the int sentinel sits at `arr[0, 0]` and the regular ramp covers the rest of the array. With a 128*128 array, the ramp climbs to 16383, well inside `int16` range, so no overflow. Fine for int16/int32, but if the dtype set grows to include `int8` later, the ramp would wrap. Worth a comment that the ramp magnitude is dtype-aware (currently it relies on the smallest tested dtype having range >= 16384).

- [ ] `test_release_gate_codec_round_trip_2341.py:194-196` -- pytest parametrize id order is `[dtype-codec]` (e.g. `[int16-none]`) because the outer decorator parametrizes `codec` first. The docs row and test docstring describe the order as `codec * dtype`. Either reorder the decorators or note the id order explicitly in the docstring.
- [ ] `test_release_gate_codec_round_trip_2341.py:283-285` -- `_assert_pixels_equal` reports a boolean mismatch. On failure it would be helpful to surface the first divergent pixel index so someone debugging a codec regression can locate the bad tile quickly. Optional.
- [ ] `docs/source/reference/release_gate_geotiff.rst:105` -- the new row says "On every stable codec ... crossed with every promised dtype ... a write / read / write / read cycle". Pretty long for a table cell; consider trimming the dtype list and pointing readers to the test file for the full matrix.

### What looks good

- The drift guard against `SUPPORTED_FEATURES` is the right shape: if a codec gets promoted into or out of `stable` and this file is not updated, the build breaks loudly with a clear message.
- The unique tmp filenames (uuid suffix + dtype + codec) sidestep collisions across parallel pytest workers and parallel worktrees.
- The on-disk TIFF Compression tag check (via `parse_ifd`) is a good answer to the "or whatever the canonical attr key is" wording in the issue body. No `attrs['compression']` exists; pinning the TIFF tag pins the actual codec choice.
- The two-pass write / read / write / read shape genuinely exercises the "round-trip" promise. Single-pass tests already exist in `test_release_gate_codecs.py`.
- The `mask_nodata=False` branch for integer dtypes correctly keeps the sentinel as a real pixel so the byte-exact comparison is meaningful (otherwise the reader would promote int to float64 with NaN, defeating the dtype assertion).
- Out-of-scope sections of the module docstring are explicit about what this gate does not cover.

### Checklist

- [x] Algorithm matches reference -- N/A (test-only)
- [x] All implemented backends produce consistent results -- N/A (the test exercises the default eager numpy backend; backend parity is already covered by `test_backend_parity_matrix.py`)
- [x] NaN handling is correct (`equal_nan=True` for float, strict equality for int)
- [x] Edge cases covered (NaN sentinel at (0, 0); int sentinel at (0, 0); cartesian product across 5 codecs * 4 dtypes)
- [x] Benchmark not applicable (release-gate test)
- [x] README feature matrix not applicable (no new public function)

brendancol (commented):
## Follow-up PR Review: Add stable-codec read/write/read round-trip release gate (#2360)

After the c8667d4 commit addressing the first review.

### Disposition of the first review

- Suggestion 1 (drop `nodata=` kwarg on the float branch in both passes): fixed in c8667d4.
- Suggestion 2 (comment the int-ramp magnitude assumption): fixed in c8667d4.
- Nit 1 (parametrize id order): fixed in c8667d4. Ids now read `[codec-dtype]`.
- Nit 2 (first divergent pixel index in `_assert_pixels_equal`): fixed in c8667d4.
- Nit 3 (trim the docs row): fixed in c8667d4.

### Blockers

(none)

### Suggestions

(none)

### Nits

- [ ] `test_release_gate_codec_round_trip_2341.py:261-263` and `:270` -- `np.issubdtype(np.dtype(dtype_name), np.floating)` is computed twice (once for `mask_kwargs`, once into `is_float`). Hoist `is_float` to the top of the test body and reuse it for `mask_kwargs`. Tiny win.

### What looks good

- Removing the `nodata=NaN` kwarg from the float path is the right call. The writer infers NaN from the array itself, so passing it again was redundant.
- The first-divergent-index branch in `_assert_pixels_equal` is what you want on a CI failure -- "byte 14732 of tile 3" beats a generic boolean.
- Docs row trimmed to a one-line description plus a single test path. Easier to scan in the table.

### Checklist

- [x] Algorithm matches reference -- N/A (test-only)
- [x] All implemented backends produce consistent results -- N/A
- [x] NaN handling is correct
- [x] Edge cases covered

brendancol:
## CI status: 4 pre-existing failures, unrelated to this PR

The `pytest` job (macOS / Ubuntu 3.14 full lane) reports 4 failures. All 4 are on test files this PR does not touch, and all 4 reproduce locally against the current `main` tip:

- `xrspatial/geotiff/tests/test_unsupported_features_2349.py::test_vrt_with_skewed_geotransform_rejected` -- `pytest.raises(match="rotated affine")` no longer matches the writer's "non-zero rotation/shear terms" message.
- `xrspatial/geotiff/tests/test_vrt_metadata_parity_2321.py::test_unsupported_resample_alg_raises` -- expects `ValueError`, the validator now raises `VRTUnsupportedError`.
- `xrspatial/geotiff/tests/test_vrt_metadata_parity_2321.py::test_negative_srcrect_size_rejected` -- regex `SrcRect.*negative size` no longer matches the SimpleSource validator message.
- `xrspatial/geotiff/tests/test_vrt_metadata_parity_2321.py::test_negative_dstrect_size_rejected` -- regex `DstRect.*negative size`, same kind of message drift.

These look like a message-format / exception-type drift between the validators and the regex assertions in the tests. The PR's own test file (`test_release_gate_codec_round_trip_2341.py`) passes 21/21 locally, and the four red tests do not touch any code path this PR introduces.

Stopping CI iteration here per `/rockout` Step 12: real CI defect, out of scope for this PR, flagged for separate follow-up.

Issue #2342, opened 2026-05-24
Epic: Conservative VRT support contract for GeoTIFF release


brendancol:
## Goal

Keep VRT support for real user workflows, but make it a conservative advanced feature with a documented, enforced subset.

VRT is valuable for mosaics and interoperability, but full GDAL VRT parity is not a realistic release promise for this module. The release should support simple mosaics well and fail closed outside that subset.

## Supported For This Release

- Simple GDAL VRT mosaics backed by GeoTIFF sources.
- Compatible source CRS, dtype, transform orientation, pixel size, band count, and band layout.
- Windowed reads where source and destination windows map cleanly.
- Lazy/dask reads over the same supported subset.
- Explicit nodata handling.
- Mixed-band nodata rejection by default, with only documented opt-ins.
- `missing_sources='raise'` as the default behavior.

## Not Promised For This Release

- Full GDAL VRT compatibility.
- Warped VRTs / reprojection VRTs.
- Nested VRTs.
- Arbitrary resampling semantics beyond the implemented and tested subset.
- Mixed CRS, mixed dtype, mixed resolution, or mixed band metadata unless explicitly supported.
- Complex mask/alpha/source semantics not represented in the GeoTIFF attrs contract.

## Work Items

- Add a VRT capability validator that runs after VRT parse and before read execution where feasible.
- Centralize validation so `read_vrt` and `open_geotiff(... .vrt ...)` fail consistently.
- Validate CRS compatibility, dtype compatibility, source/destination rect sanity, band count, nodata policy, transform compatibility, and resampling support.
- Add tests for supported simple mosaics.
- Add negative tests for unsupported VRT features.
- Ensure missing-source behavior is explicit and tested for `raise` and `warn`.
- Add docs and docstrings that describe VRT as advanced simple-mosaic support, not GDAL replacement.

## Acceptance Criteria

- Supported simple VRT mosaics work through eager and dask paths.
- Unsupported VRT features fail with actionable errors.
- Mixed or ambiguous metadata cannot silently flatten into one output raster.
- VRT tests assert values, coords, and attrs.
- Documentation plainly states the VRT subset and non-goals.

brendancol:
Closing as complete. All five work-item sub-issues are merged:

- #2368 — VRT docs: conservative simple-mosaic subset and non-goals
- #2369 — Positive tests for simple GeoTIFF VRT mosaics (eager + dask)
- #2370 — Negative tests for unsupported VRT features
[...]

PR #2366, opened 2026-05-24
Release-gate negative test for ambiguous GeoTIFF metadata (#2361)

### LINKED ISSUE #2361: Release-gate: ambiguous metadata fails closed (PR 5 of 5 of epic #2341)
## Reason or Problem

Epic #2341's acceptance criteria include: \"Unsupported or ambiguous metadata fails loudly instead of flattening or guessing.\" Today, several near-ambiguous cases are tested in scattered files (e.g. `test_ambiguous_metadata_hooks_1987.py`, `test_allow_rotated_geotiff_2115.py`, `test_allow_rotated_no_crs_2122.py`), but no single regression test pins the release-gate promise that the entire family of ambiguous-metadata cases fails closed with an actionable message.

## Proposal

Add `xrspatial/geotiff/tests/test_release_gate_negative_2341.py`.

Parametrized negative cases:

1. **Conflicting CRS between header and `.aux.xml` sidecar:** open should raise with a message naming both sources and the configured resolution policy (or a way to override).
2. **Integer nodata on a float-promoted raster:** open should raise (or refuse to coerce) rather than silently masking with the wrong sentinel.
3. **Rotated transform without `allow_rotated=True`:** open raises with an actionable message naming the flag — already covered for some paths; this test ensures every promised read entry point (eager, dask, windowed) raises uniformly.
4. **Mixed-tier VRT children when stable-only is requested:** if a VRT references a child whose codec is experimental, the open call refuses with a message naming the offending child and the flag to opt in.

Each case asserts:

- The raised exception type matches the project's existing convention for input-rejection (e.g. `ValueError` or a project-specific subclass).
- The exception message contains the name of the option/flag that would unlock the call, the feature tier, and a pointer to the release-contract doc.

Assertions inlined per-file. Add a row to ``docs/source/reference/release_gate_geotiff.rst`` under \"Cross-cutting CI gates\" citing the new file.

## Stakeholders and Impacts

Touches one new test file. If any of the four cases is currently NOT failing closed, the corresponding fix is in scope — the test pins the release promise, and a regression here is a release blocker.

Out of scope: VRT subset definition (epic #2342), the experimental-codec opt-in policy itself (#2352).

## Drawbacks

Case 4 (mixed-tier VRT children) depends on VRT epic #2342 having defined the \"stable-only\" knob. If that knob does not yet exist, this PR ships cases 1-3 and leaves case 4 as a follow-up TODO with an `xfail` marked against #2342.

## Alternatives

- Split each case into a separate PR. Rejected: the negative tests are small, the assertions are uniform, and bundling them keeps the release-gate citation row pointing at one file.

Parent epic: #2341. PR 5 of 5. Sibling issues: #2357 (PR 1), #2358 (PR 2), #2359 (PR 3), and PR 4 (to be filed).

brendancol:
Closes #2361.

## Summary

Adds `xrspatial/geotiff/tests/test_release_gate_negative_2341.py`, the
fifth and final PR of epic #2341. The file pins the release promise
that ambiguous GeoTIFF metadata fails closed at every promised read
entry point, with an actionable typed error.

Four parametrized cases:

1. Conflicting CRS between the GeoTIFF header and a sibling
   `.aux.xml` PAM sidecar. PAM sidecar reads are not implemented
   today, so this case is `xfail(strict=False)` and flips to a pass
   the moment the production fix lands.
2. Integer nodata sentinel that cannot be honoured on a float-promoted
   raster. Issue #1774 currently treats this as a silent no-op; the
   release promise upgrades the no-op to a typed rejection. Tracked
   as `xfail(strict=False)` against the #1774 follow-up.
3. Rotated affine transform without `allow_rotated=True`. Actively
   passes across the eager numpy, dask, and windowed read entry
   points. The production-side message in `_geotags.py` and
   `_validation.py` was updated to name the `reader.allow_rotated`
   experimental tier opt-in and point the caller at
   `docs/source/reference/release_gate_geotiff.rst`.
4. Mixed-tier VRT children when stable-only is requested. The
   stable-only knob is owned by epic #2342; tracked as
   `xfail(strict=False)` until that knob lands.

Adds a row to `docs/source/reference/release_gate_geotiff.rst` under
"Cross-cutting CI gates" citing the new file.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_release_gate_negative_2341.py -v`
  (3 pass, 3 xfail).
- [x] `pytest xrspatial/geotiff/tests/test_allow_rotated_geotiff_2115.py
  xrspatial/geotiff/tests/test_allow_rotated_crs_drop_2126.py
  xrspatial/geotiff/tests/test_allow_rotated_no_crs_2122.py
  xrspatial/geotiff/tests/test_ambiguous_metadata_hooks_1987.py` --
  confirms the error-message edits did not regress existing rotated
  or ambiguous-metadata coverage.
- [x] `pytest xrspatial/geotiff/tests/test_release_gate_2321.py` --
  meta-gate that checks every cited file in the checklist exists.

Assertions inlined per case; no shared helper module so the four
sibling PRs of epic #2341 cannot collide on a cross-PR symbol.

### DIFF
--- a/docs/source/reference/release_gate_geotiff.rst
+++ b/docs/source/reference/release_gate_geotiff.rst
@@ -504,6 +504,15 @@ These gates are not tier rows but they back the rest of the checklist.
 * ``test_release_gate_2321.py`` -- meta-gate that asserts every promised
   VRT behaviour in this checklist resolves to a real test file and a real
   ``SUPPORTED_FEATURES`` entry.
+* ``xrspatial/geotiff/tests/test_release_gate_negative_2341.py`` --
+  negative cross-cutting gate from epic #2341 PR 5. Pins that
+  ambiguous metadata fails closed at every promised read entry point:
+  conflicting CRS between header and ``.aux.xml`` PAM sidecar
+  (xfail until PAM sidecar support lands), integer nodata sentinel
+  that cannot be honoured on a float-promoted raster (xfail against
+  ``#1774`` follow-up), rotated transform without ``allow_rotated=True``
+  uniformly across eager / dask / windowed paths, and mixed-tier VRT
+  children when stable-only is requested (xfail against epic ``#2342``).
 
 Placeholder PR cross-references
 ===============================
--- a/xrspatial/geotiff/_geotags.py
+++ b/xrspatial/geotiff/_geotags.py
@@ -683,7 +683,11 @@ def _extract_transform(ifd: IFD,
                         f"M[6]={m[6] if len(m) > 6 else 0.0!r}). "
                         "Only axis-aligned affine transforms are supported. "
                         "Pass ``allow_rotated=True`` to read the pixel grid "
-                        "without the geospatial assumption (issue #2115)."
+                        "without the geospatial assumption "
+                        "(``reader.allow_rotated`` is the experimental tier "
+                        "opt-in; see "
+                        "``docs/source/reference/release_gate_geotiff.rst`` "
+                        "for the release-contract row; issue #2115)."
                     )
                 # Opt-in: drop georef, stash the rotated matrix on the
                 # GeoTransform so the validator + attrs-roundtrip code
--- a/xrspatial/geotiff/_validation.py
+++ b/xrspatial/geotiff/_validation.py
@@ -1117,8 +1117,9 @@ def _check_read_rotated_transform(context: Mapping[str, Any]) -> None:
         f"zonal) assume axis-aligned rasters and would silently produce "
         f"wrong results on a rotated grid. Pass ``allow_rotated=True`` "
         f"to read the pixel grid without the geospatial assumption "
-        f"(useful when you only want the array, not the geo-aware "
-        f"downstream ops). See issue #1987."
+        f"(``reader.allow_rotated`` is the experimental tier opt-in; "
+        f"see ``docs/source/reference/release_gate_geotiff.rst`` for "
+        f"the release-contract row; issue #1987)."
     )
 
 
--- a/xrspatial/geotiff/tests/test_release_gate_negative_2341.py
+++ b/xrspatial/geotiff/tests/test_release_gate_negative_2341.py
@@ -1,0 +1,459 @@
+"""Release-gate negative cases: ambiguous metadata fails closed (issue #2341 PR 5).
+
+Epic #2341's acceptance criteria include the promise that "unsupported
+or ambiguous metadata fails loudly instead of flattening or guessing".
+The positive release-gate suites in ``test_release_gate_*.py`` lock the
+"works" path. This file pins the negative side: when metadata is
+ambiguous and the caller did NOT opt in via the documented flag, every
+promised read entry point raises a typed error whose message names the
+unlocking flag and points at the release-contract docs.
+
+Four parametrized cases:
+
+1. Conflicting CRS between the GeoTIFF header and a sibling ``.aux.xml``
+   PAM sidecar. The reader must not silently prefer one over the other.
+2. Integer nodata sentinel that cannot be honoured on a float-promoted
+   raster. The reader must not silently mask with the wrong value.
+3. Rotated affine transform without ``allow_rotated=True``. Eager,
+   dask, and windowed entry points must each raise uniformly.
+4. Mixed-tier VRT children when the caller asked for stable-only
+   sources. The reader must name the offending child and the opt-in.
+
+Assertions inlined per case so a single failing row is locatable
+without cross-file helpers. Sibling PRs of epic #2341 are running in
+parallel; this file does NOT introduce a shared helper module.
+
+xfail policy
+------------
+Cases 1, 2, and 4 are tagged ``strict=False`` xfail because the
+production-side rejection is tracked under a separate epic or a
+follow-up issue:
+
+* Case 1 -- ``.aux.xml`` PAM sidecar read is not a supported feature
+  today (no entry in :data:`xrspatial.geotiff.SUPPORTED_FEATURES`).
+  The release promise is that *if* the reader gains PAM sidecar
+  support, it must fail closed on a CRS conflict; the xfail flips
+  to a pass the moment that production fix lands.
+* Case 2 -- the current behaviour on a non-finite or fractional
+  integer nodata sentinel is a no-op (issue #1774). The release
+  promise is to upgrade that no-op to a typed error so the caller
+  sees the silent-coercion risk; the xfail flips when the upgrade
+  lands.
+* Case 4 -- the VRT stable-only knob is owned by epic #2342 and
+  has not landed. The xfail flips when #2342 ships the knob.
+
+Case 3 (rotated) actively passes today on every cited entry point.
+"""
+from __future__ import annotations
+
+import struct
+import uuid
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+from xrspatial.geotiff import open_geotiff, read_geotiff_dask
+from xrspatial.geotiff._errors import (
+    GeoTIFFAmbiguousMetadataError,
+    RotatedTransformError,
+)
+
+# --------------------------------------------------------------------------- #
+# Shared release-contract pointers.                                           #
+# --------------------------------------------------------------------------- #
+#
+# The release-contract doc lives next to the release-gate checklist. The
+# negative-case error messages should mention either the contract page,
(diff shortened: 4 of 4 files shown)


brendancol (commented):
## PR Review: Release-gate negative test for ambiguous GeoTIFF metadata (#2361)

None.

- `xrspatial/geotiff/tests/test_release_gate_negative_2341.py:386-389` -- the comment above the tier check claims "The tier word ``advanced`` is the live tier for this opt-in in SUPPORTED_FEATURES", but `_attrs.py:298` tags `reader.allow_rotated` as `experimental`, not `advanced`. The runtime check itself is fine because it accepts any of `("advanced", "experimental", "stable")`, but the comment is inaccurate and will mislead the next person who tries to tighten the check. Update the comment to match the live tier.

- `xrspatial/geotiff/tests/test_release_gate_negative_2341.py:451` -- the case-4 trick `_tmp(...).replace(".tif", ".vrt")` is brittle if the `tmp_path` ever happens to embed `.tif` in a directory segment. Construct the `.vrt` path directly via `tmp_path / f"...uuid....vrt"` rather than string-replacing the `.tif` suffix. Same outcome, less surface area for surprise.

- The case-3 release promise per the issue spec also names "every promised read entry point" -- GPU read is `experimental` but it is a promised entry point in `SUPPORTED_FEATURES`. Adding a fourth sub-case `test_release_gate_negative_rotated_gpu` (skipped on hosts without cupy + CUDA via the existing `requires_gpu` fixture from `conftest.py`) would close the matrix without changing scope. Optional; the issue explicitly listed eager / dask / windowed, so leaving GPU out is defensible.

- `xrspatial/geotiff/tests/test_release_gate_negative_2341.py:107` -- the `_TAG_MODEL_TRANSFORMATION` and rotation constants are duplicated from `test_allow_rotated_geotiff_2115.py`. The PR description says this is intentional (no shared helper module across the four sibling PRs). Keep the inline copy; consider adding a one-line comment pointing back at the source file so a future grep finds the canonical location.

- `xrspatial/geotiff/_geotags.py:684-690` and `xrspatial/geotiff/_validation.py:1117-1122` -- the two rewritten messages share the trailing pointer prose. A short module-level constant in `_errors.py` (e.g. `_RELEASE_CONTRACT_POINTER`) would keep both raise sites in lockstep if the docs path changes again. Not in this PR's scope; flag for follow-up if the message gets edited a third time.

- The case-4 xfail body calls `open_geotiff(path, stable_only=True)` and relies on `TypeError: unexpected keyword argument` to satisfy the xfail. The assertions on the message inside the `with pytest.raises` block never run today. When epic #2342 lands and `stable_only=` becomes a real kwarg, the test will flip to XPASS even if the new code raises a different exception type or with a message that does not mention the opt-in. Either tighten the xfail to `raises=GeoTIFFAmbiguousMetadataError` (so it stays xfail until the correct shape lands) or add a short note in the docstring that the xfail-to-pass transition needs a manual review of the message contract.

### What looks good

- Cases 1, 2, 4 follow the spec's xfail policy: `strict=False`, distinct reason strings, each pointing at the production-side blocker that flips the gate.
- Case 3 actually exercises three entry points (eager, dask, windowed) with the shared fixture, matching the issue spec wording.
- Production-side message edits in `_geotags.py` and `_validation.py` are minimal and the existing `test_allow_rotated_geotiff_2115.py` regex `match="rotation"` keeps matching, so the edit does not regress sibling tests.
- The `_msg_cites_release_contract` helper is intentionally loose (multiple acceptable pointers) so the gate does not break when the docs file changes extension.
- The new row in `release_gate_geotiff.rst` lands under "Cross-cutting CI gates" exactly where the issue spec asked.
- Unique tmp file naming via the per-call UUID matches the project memo about parallel sibling PRs.

### Checklist

- [x] Algorithm matches reference/paper (n/a -- this is a test-only / message-edit PR)
- [x] All implemented backends produce consistent results (eager / dask / windowed locked; GPU optional, see Suggestion above)
- [x] NaN handling is correct (case 2 covers the non-finite nodata sentinel as xfail)
- [x] Benchmark exists or is not needed (not needed for a release-gate test)
- [x] README feature matrix updated (n/a -- no new public function)

brendancol (commented):
## PR Review (follow-up): Release-gate negative test for ambiguous GeoTIFF metadata (#2361)

Re-reviewed after commit ``1f3c49ca``.

None.

None remaining.

- ``xrspatial/geotiff/tests/test_release_gate_negative_2341.py:485-487`` -- the inline comment above the case-4 ``_tmp`` call still says "this call fails for an unrelated reason (unknown kwarg), which the strict=False xfail swallows". The same point now lives in the docstring's "XFAIL-to-PASS transition note" block. Either trim the inline comment or remove it -- the longer docstring is the canonical home and the duplication can drift. Pure tidy-up, not blocking.

### Disposition of the prior review

- Suggestion 1 (tier-string comment drift): fixed in ``1f3c49ca``. The comment now matches ``_attrs.py:298`` (``reader.allow_rotated`` is ``experimental``).
- Suggestion 2 (brittle ``.replace(".tif", ".vrt")``): fixed in ``1f3c49ca`` by adding a ``suffix=`` kwarg to ``_tmp``.
- Suggestion 3 (GPU sub-case): fixed in ``1f3c49ca``. ``test_release_gate_negative_rotated_gpu`` was added, gated by the existing ``requires_gpu`` marker from ``conftest.py``. The case 3 release promise now covers all four read entry points: eager, dask, windowed, and GPU.
- Nit 1 (point-back comment to canonical source): fixed in ``1f3c49ca``.
- Nit 2 (shared release-contract pointer constant in ``_errors.py``): deferred. The PR description was explicit that this is a follow-up if the message ever gets edited a third time; the two raise sites stayed in sync in this PR.
- Nit 3 (case-4 xfail tightening): fixed via a docstring "XFAIL-to-PASS transition note" rather than ``raises=``. Tightening to ``raises=GeoTIFFAmbiguousMetadataError`` would break the test today because the current call site raises ``TypeError`` (unknown kwarg). The docstring documents the manual verification step required before the xfail marker comes off, which preserves the gate value without flipping the test red right now.

### What looks good

- All four case-3 entry points (eager / dask / windowed / GPU) now pass locally and share the ``_assert_rotated_message`` helper, so message drift in any one path fails the others uniformly.
- The ``_tmp`` helper signature is cleaner with the explicit ``suffix=`` kwarg; the case-4 path no longer depends on string replacement.
- The xfail markers all stay ``strict=False`` and carry distinct reason strings pointing at the production-side blockers (``#1774``, PAM sidecar epic, ``#2342``).
- The production-side message edits in ``_geotags.py`` and ``_validation.py`` did not regress sibling tests; ``test_allow_rotated_geotiff_2115.py``, ``test_allow_rotated_crs_drop_2126.py``, ``test_allow_rotated_no_crs_2122.py``, and ``test_ambiguous_metadata_hooks_1987.py`` all pass.

### Checklist

- [x] Algorithm matches reference/paper (n/a -- test-only / message edit)
- [x] All implemented backends produce consistent results (eager / dask / windowed / GPU locked for case 3)
- [x] NaN handling is correct
- [x] README feature matrix updated (n/a -- no new public function)

brendancol:
## CI status

The new test file passes on every platform in CI:

```
xrspatial/geotiff/tests/test_release_gate_negative_2341.py xx...sx
```

(``xx`` = cases 1 and 2 xfail, ``...`` = case 3 eager / dask / windowed pass, ``s`` = GPU sub-case skipped without a CUDA host, ``x`` = case 4 xfail.)

The 3.14 ``pytest`` job is red, but the four failing tests are pre-existing failures on ``main`` and are not introduced by this PR:

* ``test_unsupported_features_2349.py::test_vrt_with_skewed_geotransform_rejected``
* ``test_vrt_metadata_parity_2321.py::test_negative_dstrect_size_rejected``
* ``test_vrt_metadata_parity_2321.py::test_negative_srcrect_size_rejected``
* ``test_vrt_metadata_parity_2321.py::test_unsupported_resample_alg_raises``

Verification: checking out ``origin/main``'s ``_validation.py`` and ``_geotags.py`` and running the same four tests locally reproduces the same four failures. The most recent ``main`` CI run (``gh run view 26363918141``, head ``9c40df49``) reports the same four failures. They are out of scope for this PR (epic #2341 PR 5 is a release-gate test); the breakage was introduced upstream by the merge of #2355 against an earlier message contract in ``test_vrt_metadata_parity_2321.py``.

PR #2372, opened 2026-05-25
Add consolidated VRT missing_sources policy matrix tests (#2367)

### LINKED ISSUE #2367: VRT tests: missing_sources policy coverage for raise and warn (#2342 work item)
Sub-task of #2342.

The release docs say `missing_sources='raise'` is the default and `'warn'` is opt-in. These tests lock that contract in for both eager and dask reads.

Cases to cover:
- Default (no `missing_sources` kwarg): a VRT pointing at a non-existent source raises. Assert the exception type and that the message names the missing file.
- `missing_sources='raise'` explicit: same as above.
- `missing_sources='warn'`: emits the warning class used in the code and names the missing source. The read returns an array with nodata in the corresponding region. Assert both the warning message and the output values.
- `missing_sources='invalid'`: raises `ValueError` naming the bad value.
- Both eager (`read_vrt`) and dask (`open_geotiff(... .vrt ..., chunks=...)`) paths get exercised for each case.

Done when:
- All cases pass under both read paths.
- Assertions cover exception/warning types, messages, and output array values.

brendancol:
Closes #2367. Sub-task of epic #2342.

Adds a single test file that pins the four-by-two policy matrix for VRT `missing_sources`:
- default / explicit `'raise'` / `'warn'` / invalid value
- each exercised against eager `read_vrt` and dask `open_geotiff(.vrt, chunks=...)`

Assertions cover the exception type and message (raise / invalid), the `GeoTIFFFallbackWarning` class plus message text (warn), and the actual output array values (NaN on the missing half, fill value on the present half).

The existing tests at `test_vrt_missing_sources_policy_1799.py`, `test_vrt_chunked_missing_sources_1799.py`, `test_vrt_missing_sources_default_raise_1843.py`, and `test_read_vrt_default_missing_sources_1860.py` each pin a slice of this matrix. They stay in place; this file keeps the full matrix together so a kwarg refactor that drops eager/dask parity regresses one focused file.

## Test plan
- [x] `pytest xrspatial/geotiff/tests/test_vrt_missing_sources_policy_2367.py` (18 passed locally)
- [x] Re-ran the four sibling test files; all still pass.

### DIFF
--- a/xrspatial/geotiff/tests/test_vrt_missing_sources_policy_2367.py
+++ b/xrspatial/geotiff/tests/test_vrt_missing_sources_policy_2367.py
@@ -1,0 +1,270 @@
+"""Consolidated VRT ``missing_sources`` policy matrix (#2367, work item of #2342).
+
+This file complements ``test_vrt_missing_sources_policy_1799.py`` and
+``test_vrt_chunked_missing_sources_1799.py`` by covering the full
+release contract in one place: every policy value (default,
+``'raise'``, ``'warn'``, invalid) is exercised against both read paths
+(eager ``read_vrt`` and dask ``open_geotiff(..., chunks=...)``), with
+assertions on the exception or warning type, the message text, and the
+actual output array values where applicable.
+
+The existing 1799 / 1843 / 2265 tests pin individual cases. This file
+keeps the four-by-two matrix together so a future kwarg refactor that
+silently drops parity between the eager and chunked paths regresses a
+single, focused test file.
+
+Release contract (see ``_backends/vrt.py:206`` docstring):
+
+* ``'raise'`` is the default since #1860.
+* ``'raise'`` fails fast with ``FileNotFoundError`` naming the missing
+  source path. The chunked path raises at build time (#2265) so a
+  partial mosaic never surfaces silently from a delayed compute.
+* ``'warn'`` is the explicit opt-in. It emits
+  ``GeoTIFFFallbackWarning`` naming the missing source and returns the
+  mosaic with NaN (or the band's nodata sentinel) in the corresponding
+  region. ``attrs['vrt_holes']`` records the affected source(s).
+* Any other value raises ``ValueError`` naming the bad kwarg.
+"""
+from __future__ import annotations
+
+import os
+import warnings
+
+import numpy as np
+import pytest
+import xarray as xr
+
+from xrspatial.geotiff import (
+    GeoTIFFFallbackWarning,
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: VRT missing_sources policy matrix tests (#2367)

### Blockers
None.

### Suggestions
None.

### Nits
- `test_vrt_missing_sources_policy_2367.py:113-117`: the eager warn test uses `pytest.warns(GeoTIFFFallbackWarning) as record` and re-checks the message in the body instead of passing `match=`. Either style works; flagging only because the sibling 1799 test goes with `match=`.
- `test_vrt_missing_sources_policy_2367.py:262-269`: nice parametrize set on the bad values (`'ignore'`, `'RAISE'`, `'raises'`, empty, trailing space, `'1'`). The value-validation block at `vrt.py:399-402` runs before chunked dispatch, so the chunked invocations hit the same error path, which is the cross-mode parity this matrix is meant to pin.

### What looks good
- Four-by-two matrix in one file: default / explicit-raise / warn / invalid against both eager and dask readers.
- Output value assertions on the warn path (NaN on the missing half, 7.0 on the present half) under both readers, not just "did it warn".
- Warning class is checked as `GeoTIFFFallbackWarning`, not bare `UserWarning`.
- Build-time `vrt_holes` is asserted on the dask warn path so callers can branch on partial mosaics without forcing a compute.
- Temp filenames all carry `_2367` so parallel runs don't collide.
- The parametrize-by-reader pattern keeps eager and dask symmetric for the raise / invalid cases without duplicating the test bodies.

### Checklist
- [x] Test-only change; algorithm correctness not applicable
- [x] Both supported read paths exercised (eager `read_vrt`, dask `open_geotiff(.vrt, chunks=)`)
- [x] NaN handling on the missing region asserted
- [x] Edge cases on the bad-value side (empty string, trailing space, wrong case) covered
- [x] Dask path's parse-time `vrt_holes` and compute-time warning both pinned
- [x] No performance-critical paths touched
- [x] No new public API; README / docs untouched on purpose
- [x] Docstrings on every test class explain what they pin and why

brendancol:
Followed up on the Nit 1 from the prior review (sibling test parity with #1799): the eager warn assertion now uses `pytest.warns(GeoTIFFFallbackWarning, match="missing_2367.tif")` in 52328fac, matching the 1799 style. The reviewer flagged this as "either style works" but consistency across the missing-sources test pair is worth the change. Nit 2 was a positive observation, no action.

PR #2373, opened 2026-05-25
Positive VRT mosaic tests over GeoTIFF sources (#2369)

### LINKED ISSUE #2369: VRT tests: positive coverage for simple GeoTIFF mosaics (#2342 work item)
Sub-task of #2342.

Simple GDAL VRT mosaics over GeoTIFF sources are listed as supported for both eager and dask reads. Right now that subset has implicit coverage through other VRT tests but no dedicated positive cases. This issue tracks adding them so the supported shape is pinned down by tests.

Test cases:
- 2x1 horizontal mosaic of two compatible GeoTIFF tiles. Assert values match the expected concatenated array, plus coords and attrs (CRS, transform, nodata).
- 2x2 mosaic of four compatible tiles. Same assertions.
- Windowed read where the request window maps cleanly into source windows.
- Dask read of the same mosaics with multiple chunks.
- At least one multi-band mosaic case.

Acceptance:
- Tests pass on numpy and dask+numpy paths.
- Tests assert values, coords, and attrs (not just shapes).

brendancol:
Closes #2369. Sub-task of #2342.

Positive coverage for the simple VRT mosaic subset that the release contract lists as supported. The other VRT tests touch this subset while exercising codecs, dtypes, or chunk policies; nothing pinned the contract by itself.

Six tests, all driving the public `read_vrt`:

- eager 2x1 horizontal mosaic
- eager 2x2 mosaic with one distinct value per tile (catches placement swaps loudly)
- windowed read aligned with the source seam
- dask read of both mosaics with multiple chunks
- multi-band 2x1 mosaic

Each test asserts values, coords, and key attrs (crs, transform, nodata).

## Backend coverage

- numpy: yes
- dask + numpy: yes
- cupy / dask + cupy: not in scope (the release contract only promises eager + dask for VRT)

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_vrt_simple_mosaic_2369.py` passes locally
- [ ] CI runs green on the same file


brendancol (commented):
## PR Review: Positive VRT mosaic tests over GeoTIFF sources (#2369)

### Blockers
None.

### Suggestions
- [ ] `xrspatial/geotiff/tests/test_vrt_simple_mosaic_2369.py:118,146,184`: the three fixtures use `tempfile.mkdtemp(prefix='tmp_2369_...')`, which leaks directories between runs. The implementation notes called for `tmp_path`, and most newer tests in this directory use the pytest fixture. Switching gets auto-cleanup and keeps the prefix-based collision protection.
- [ ] `xrspatial/geotiff/tests/test_vrt_simple_mosaic_2369.py:240-254`: `_assert_attrs_ok` checks the pixel size on the transform but not `origin_x` / `origin_y`. A transform with the wrong origin would still pass. Asserting `transform[2]` and `transform[5]` against the fixture origin would catch that case.

### Nits
- [ ] `xrspatial/geotiff/tests/test_vrt_simple_mosaic_2369.py:333`: `h, w_total = expected.shape` -- `w_total` is unused. Drop it or reference it in the comment.
- [ ] `xrspatial/geotiff/tests/test_vrt_simple_mosaic_2369.py:410-431`: no dask variant of the multi-band test. The eager case covers the contract; a chunked multi-band read would also pin down per-block band handling.

### What looks good
- Six tests, narrow scope, exactly the supported subset from the issue.
- 2x2 fixture uses one distinct constant per tile so placement bugs show up in the value diff instead of as numeric drift.
- Each test asserts values, coords, and key attrs (crs, transform, nodata).
- Tile builders are short and shared cleanly across fixtures.

### Checklist
- [x] Algorithm matches reference: synthesises mosaics from `to_geotiff` + internal `write_vrt`, mirroring `test_vrt_lazy_chunks_1814.py`.
- [x] All implemented backends produce consistent results: dask compute compared against eager.
- [x] NaN handling: not exercised in this PR; mosaics use finite data plus declared nodata sentinel.
- [x] Edge cases: windowed read on a 2x1 seam covers the documented boundary case.
- [x] Dask chunk boundaries handled correctly: 2x1 and 2x2 mosaics with sub-tile chunks both match eager.
- [x] No premature materialization: compute() only fires after the multi-block dask graph is built.
- [n/a] Benchmark exists or is not needed: test-only PR.
- [n/a] README feature matrix updated: no public API change.

brendancol (commented):
## Follow-up review (post-fixes)

Disposition for the four findings from the previous review:

- [x] Fixture `tempfile.mkdtemp` -> `tmp_path` subdirs. Fixed.
- [x] `_assert_attrs_ok` now checks `transform[2]` / `transform[5]`. Every call site passes the fixture origin; the windowed call passes the shifted origin (`ox + _PIXEL_W*c0`, `oy + _PIXEL_H*r0`). Fixed.
- [x] Unused `w_total` removed; the windowed test now binds only `h`. Fixed.
- [x] Dask multi-band test added (`test_dask_multiband_2x1_mosaic_matches_eager`) so per-block band handling is covered. Fixed.

Total: 7 tests pass locally. No deferrals, no dismissals.

Issue #1783, opened 2026-05-13
geotiff: VRT unknown dataType silently maps to float32


brendancol:
**Describe the bug**

`parse_vrt()` in `xrspatial/geotiff/_vrt.py` resolves the per-band data type with:

```python
dtype_name = band_elem.get('dataType', 'Float32')
dtype = np.dtype(_DTYPE_MAP.get(dtype_name, np.float32))
```

at line 280-281. `_DTYPE_MAP` (line 119) only covers eight GDAL dtype names: `Byte`, `UInt16`, `Int16`, `UInt32`, `Int32`, `Float32`, `Float64`, `Int8`. Any other valid GDAL dataType -- `UInt64`, `Int64`, `CInt16`, `CInt32`, `CFloat32`, `CFloat64` -- silently falls back to `float32`.

For `UInt64` / `Int64` this is precision loss above ~2^24 (a 64-bit integer raster reads as float32 and quietly loses low-order bits). For the complex types it discards the imaginary component entirely. The XML claims one thing; the array holds another.

The `.get(..., np.float32)` default also dates from when only `Float32` was the documented default for missing `dataType`. The default-for-missing-attribute and default-for-unknown-attribute cases should not collapse together.

**Expected behavior**

Unknown `dataType` values should raise a typed `ValueError` from `parse_vrt()` so the malformed/unsupported VRT is surfaced rather than silently re-typed.

If a GDAL dataType can be represented natively (e.g. `UInt64`, `Int64`), it should be added to `_DTYPE_MAP` deliberately. Complex types remain unsupported and should raise with a clear message.

**Categories**

- Cat 2 (data corruption): silent precision loss for 64-bit integer / complex VRT files
- Cat 3 (input validation): unknown XML attribute value is accepted instead of rejected

PR #2375, opened 2026-05-25
VRT tests: negative coverage for unsupported features (#2370)

### LINKED ISSUE #2370: VRT tests: negative coverage for unsupported features (#2342 work item)
Sub-task of #2342.

The release does not promise full GDAL VRT parity. Unsupported VRT features should fail with actionable errors instead of silently producing wrong data. Add negative test coverage that pins the rejection contract.

Cases that must raise:
- Warped VRT (`<VRTRasterBand subClass="VRTWarpedRasterBand">` or a VRT containing `<GDALWarpOptions>`).
- Nested VRT (a `.vrt` referenced as `<SourceFilename>` inside another `.vrt`).
- Mixed source CRS across band sources.
- Mixed source dtype across band sources where the output dtype is ambiguous.
- Mixed band count across sources.
- Complex mask source semantics that the attrs contract cannot represent.
- Unsupported resample algorithm (anything outside the implemented subset).

Each test asserts the exception type and checks that the error message names the unsupported feature.

Acceptance:
- One dedicated test per rejection path.
- Both `read_vrt` and `open_geotiff(... .vrt ...)` are exercised.

Coordination notes:
- Sibling PR 1 adds a centralized `validate_vrt_capability`. Where current code already rejects a case, lock that in directly. Where rejection depends on PR 1, mark with `pytest.xfail`.
- Sibling PRs in parallel: PR 2 positive mosaics, PR 4 missing-source policy, PR 5 docs.

brendancol:
## Summary

- Negative tests that pin the rejection contract for VRT inputs the release does not promise to support.
- Covers warped VRT, nested VRT, mixed source CRS, mixed dtype, mixed band count, complex mask source, and unsupported resample algorithm.
- Each test checks the exception type and that the message names the unsupported feature, so users can fix the input.
- Both `read_vrt` and `open_geotiff(... .vrt ...)` entry points are exercised.

## Backend coverage

Test-only PR. The VRT read path is numpy-eager; backend dispatch is not in scope.

## Coordination

PR 3 of epic #2342. Sibling PR 1 (#2329) is the centralized `validate_vrt_capability`. Where the current code already rejects a case (complex dtype #1783, non-nearest resample alg #1751, bad source band number), the test pins the existing behaviour. Where the rejection depends on #2329, the test is wrapped with `pytest.xfail(reason="depends on centralized validate_vrt_capability (#2329)")` so this PR can land on its own. Those tests will flip to passing once #2329 merges.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_vrt_unsupported_2370.py`: 9 passed, 8 xfailed
- [x] Adjacent VRT tests still pass (`test_vrt_resample_alg_1751`, `test_vrt_dtype_1783`)
- [x] All tmp filenames include `2370` to avoid collisions with parallel rockout worktrees

Closes #2370.


brendancol (commented):
## PR Review: VRT tests: negative coverage for unsupported features (#2375)

Test-only PR. I read the new file `xrspatial/geotiff/tests/test_vrt_unsupported_2370.py` and ran the suite locally (9 passed, 8 xfailed, as advertised).

None.

- `xrspatial/geotiff/tests/test_vrt_unsupported_2370.py:121`: `_assert_raises_or_xfail` is defined under the "Group 1 -- Warped VRT" header but it is shared by every group. Move it above the Group 1 header (or put it under a dedicated "Helpers" header) so the section layout matches the code. Cosmetic, but it trips readers scanning the headers.
- `xrspatial/geotiff/tests/test_vrt_unsupported_2370.py:141`: `except BaseException` is broader than needed and would swallow `KeyboardInterrupt` / `SystemExit` if one ever leaked out of XML setup. `except Exception` keeps the diagnostic xfail for genuine errors without masking interruption signals.

- `xrspatial/geotiff/tests/test_vrt_unsupported_2370.py:79`: `_simple_source_xml` accepts `x_size`, `y_size`, `dst_x_size`, `dst_y_size`, `extra_inner` parameters that no caller in this PR uses. Either drop them or note in the docstring that they are forward-compat for follow-up work.
- The resample-alg tests overlap by design with `test_vrt_resample_alg_1751.py`. A one-line note in the module docstring explaining the split (this file is the rejection-contract anchor; the #1751 file is the regression anchor) would save a future reader from de-duping them.

### What looks good
- Every test checks both the exception type and that the message names the unsupported feature, matching the issue's acceptance criteria.
- `_assert_raises_or_xfail` produces one consistent xfail reason naming PR #2329, so a reader who sees an xfail does not have to dig into the test to learn why.
- Coverage spans `read_vrt` and `open_geotiff` for the cases where a regression at either entry point would leak past the contract (warped, nested, resample alg).
- The supported-VRT anchor test prevents the file from quietly degrading if `.vrt` extension dispatch breaks.
- All tmp filenames carry `2370`, matching the project's collision-avoidance convention.
- The two cases the current code already rejects (complex dtype `CFloat32`, non-nearest resample alg) lock in real messages with `pytest.raises(..., match=...)` rather than the looser xfail-friendly helper.

### Checklist
- [x] Test-only change; correctness checks N/A for production code
- [x] Four-backend coverage N/A (VRT read path is numpy-eager today)
- [x] NaN handling N/A
- [x] Edge cases captured via crafted VRT XML
- [x] Dask chunk boundaries N/A
- [x] No performance-sensitive code changed
- [x] No README or docs update needed (test-only PR)
- [x] Docstrings present on every test
- [x] `Closes #2370` in the PR body

brendancol (commented):
## PR Review (follow-up): commit 6f92559a

Re-reviewed after the follow-up commit. All four findings from the first pass were addressed:

- Helper moved above the Group 1 header into a dedicated Helpers section.
- `except BaseException` narrowed to `except Exception` on the diagnostic branch, with a docstring note explaining why.
- Unused parameters dropped from `_simple_source_xml`; the helper is now pinned to the 4x4 geometry every caller uses.
- Module docstring now explains the intentional overlap with `test_vrt_resample_alg_1751.py`.

Re-ran the suite locally: 9 passed, 8 xfailed -- no change in disposition, which is the expected result for these refactors.

### Blockers
None.

### Suggestions
None.

### Nits
None.

The PR is ready as far as the test contract goes. CI is the next gate.

PR #2376, opened 2026-05-25
Centralise VRT capability validator at both read entry points (#2371)

### LINKED ISSUE #2371: VRT capability validator: centralize post-parse pre-read gate (#2342 work item)
Sub-task of #2342.

Add a centralized `validate_vrt_capability(parsed)` that runs after VRT parse and before any read executes. Both `read_vrt` in `_vrt.py` and the `open_geotiff(..., .vrt)` path in `_backends/vrt.py` need to call it so unsupported VRT features fail identically at both entry points.

`validate_parsed_vrt` already exists in `xrspatial/geotiff/_vrt_validation.py` and covers transform orientation, dtype, pixel size, mixed-nodata, CRS, SrcRect/DstRect sanity, and part of the resample algorithm case. This work extends it to cover the remaining gaps and wires the internal `_vrt.read_vrt` entry point through the same gate. Today only `_backends/vrt.read_vrt` calls the validator, so a direct call into `_vrt.read_vrt` skips it.

Detection scope:

- Nested VRTs: a `.vrt` referenced as a `SourceFilename` inside another VRT. Reject.
- Warped VRTs: `<VRTRasterBand subClass="VRTWarpedRasterBand">` or a `<GDALWarpOptions>` block anywhere in the tree. The current parse step catches the subClass attribute via its generic subClass rejection but does not look for a `<GDALWarpOptions>` element. Reject both.
- Resample algorithm beyond nearest: fire the rejection whenever SrcRect and DstRect sizes differ and the algorithm is not nearest, regardless of which source tag declared it.
- Complex mask / alpha source semantics that the GeoTIFF attrs contract cannot carry: `<UseMaskBand>` flags and per-source mask references. Reject.

Each rejection raises `VRTUnsupportedError` (already defined in `_errors.py`) and names the offending source path and field, so the caller can find the bad source without re-parsing the XML.

Acceptance:
- Both entry points route through the validator.
- Unit tests cover every new rejection path.
- Existing supported VRT tests still pass.

brendancol:
Closes #2371.

## Summary

- Extend `validate_parsed_vrt` to reject nested VRTs, per-source `<UseMaskBand>` flags, and per-source `<MaskBand>` children with `VRTUnsupportedError`.
- Extend `parse_vrt` to reject dataset-level and band-level `<GDALWarpOptions>` blocks with `UnsupportedGeoTIFFFeatureError`.
- Wire the internal `_vrt.read_vrt` entry point through `validate_parsed_vrt` so direct callers and the chunked dask path get the same capability rejections as the public `_backends/vrt.read_vrt` entry point.
- Expose `validate_vrt_capability` as a public alias on `_vrt_validation` matching the epic naming.

## Backend coverage

The validator is a metadata gate that runs before any backend dispatch, so numpy, cupy, dask+numpy, and dask+cupy all see the same rejections at graph-build / eager-read setup time.

## Test plan

- [x] New file `xrspatial/geotiff/tests/test_vrt_capability_validator_2371.py` (22 tests) covers every new rejection path at four entry points: validator-direct, internal `_vrt.read_vrt`, public `_backends/vrt.read_vrt`, and `open_geotiff`.
- [x] `test_vrt_resample_alg_1751.py` accepts `VRTUnsupportedError` alongside the historical `NotImplementedError`. Both encode the same contract.
- [x] Full `xrspatial/geotiff/tests/` suite passes (5661 passed).

Sub-task of epic #2342. Sibling PRs in flight: positive simple-mosaic tests, negative tests for unsupported VRT features, missing-source policy tests, and VRT contract docs.


brendancol (commented):
## PR Review: Centralise VRT capability validator at both read entry points (#2371)

None.

- [ ] `xrspatial/geotiff/_vrt.py:713`: the `<UseMaskBand>` truthy set is `('1', 'true', 'yes')`. GDAL writes lowercase `true`, so `'1'` and `'yes'` are spellings the parser will never see from GDAL itself. Tightening to `('1', 'true')` keeps the surface narrow and matches what real VRTs contain. Not blocking; the wider set is harmless.

- [ ] `xrspatial/geotiff/_vrt_validation.py:303`: the nested-VRT message ends with "are not a supported feature in this release". The `parse_vrt` rejections point readers at `xrspatial.geotiff.SUPPORTED_FEATURES`. Adding the same pointer here keeps the rejection messages aligned on one discovery anchor.
- [ ] `xrspatial/geotiff/_vrt.py:1247`: the validator import is lazy inside the function body. That matches the existing pattern elsewhere in the file, but a one-line comment ("lazy to avoid circular import") would help the next reader.
- [ ] `xrspatial/geotiff/tests/test_vrt_capability_validator_2371.py:339`: the `test_use_mask_band_truthy_spellings_rejected` parametrize list includes `'yes'`. If you tighten the truthy set per the suggestion above, drop that row or move it to a "must be accepted as not-truthy" test.

### What looks good
- The validator extends cleanly. Every new rejection names the offending source path and field, so the error message tells the caller where to look.
- `validate_vrt_capability` as an alias keeps every existing call site working.
- Test coverage is strong: every new rejection path runs at four entry points (validator-direct, internal `_vrt.read_vrt`, public `_backends/vrt.read_vrt`, `open_geotiff`).
- The `test_vrt_resample_alg_1751.py` fix accepts both `NotImplementedError` and `VRTUnsupportedError`, so the legacy contract intent is preserved without breaking the new routing.
- The `parsed is None` short-circuit in `_vrt.read_vrt` avoids double-validation when the chunked path threads a pre-validated instance in.
- `<GDALWarpOptions>` rejection covers dataset-level, band-level, and the existing `subClass="VRTWarpedRasterBand"` marker. All three locations a warp config can appear are caught.

### Checklist
- [x] Algorithm matches the VRT spec
- [x] Implemented backends produce consistent results (validator runs before backend dispatch)
- [x] NaN handling not applicable (metadata-only gate)
- [x] Edge cases covered: uppercase `.VRT`, `false` flag accepted, truthy spellings, both source tag types
- [x] Dask chunk boundaries handled correctly (chunked path threads a pre-validated VRT)
- [x] No premature materialization (metadata gate runs before any decode)
- [x] Benchmark not needed (cheap XML-only check)

brendancol (commented):
## Follow-up review after addressing prior comments

- **Suggestion** (`_vrt.py:713` UseMaskBand truthy set): **Fixed.** Narrowed to `('1', 'true')`. Added `test_use_mask_band_non_canonical_truthy_accepted` to lock in the contract for non-canonical tokens.
- **Nit 1** (`_vrt_validation.py:303` SUPPORTED_FEATURES anchor): **Fixed.** Nested-VRT message now references `xrspatial.geotiff.SUPPORTED_FEATURES`.
- **Nit 2** (lazy import comment): **Fixed.** Added a one-line note explaining the circular-import avoidance.
- **Nit 3** (drop `'yes'` parametrize row): **Fixed.** Row removed; non-canonical tokens are covered by the new explicit test instead.

### Verification
- Full geotiff test suite: 5663 passed, 68 skipped, 4 xfailed, 1 xpassed.
- New test count: 24 (was 22). The two new tests cover the narrowed truthy set.

No further findings.

PR #2377, opened 2026-05-25
Make corpus determinism check toolchain-agnostic (#2299)

### LINKED ISSUE #2299: geotiff corpus: determinism test is toolchain-coupled (conda-forge GDAL differs from fixture-build env)
Surfaced by #2290 / #2289 while wiring up the conda-forge rasterio CI workflow.

## Symptom
`xrspatial/geotiff/tests/golden_corpus/test_corpus_determinism.py::test_fixture_bytes_are_deterministic` fails for two fixtures when run against the GDAL / libjpeg shipped by conda-forge today:

- `cog_internal_overview_uint16`: committed md5 `ad1c9adf5958ba10f19319c59e99e1dc`, regenerated md5 `3f5111931f0c6e24c90f7139376bc72b`.
- `compression_jpeg_uint8_ycbcr`: committed md5 `ccf1387a7799cc30065882874de5d709`, regenerated md5 `d3702bde1e887945740796257ff80dc2`.

Both are sensitive to GDAL internals (overview pyramid encoding) and libjpeg version (JPEG encoder output), so the byte mismatch is expected when the toolchain differs from the one that originally produced the fixtures.

## Workaround in #2290
The new `pytest-geotiff-corpus` workflow `--ignore`s `test_corpus_determinism.py` so the parity oracle still runs in CI. The oracle and nodata tests compare semantic output via rasterio, which is what we actually want from the conda-forge lane.

## What needs to happen here
Decide how the determinism test should behave when the toolchain differs. Options worth weighing:

1. Pin GDAL / libjpeg versions in `setup.cfg` and in the corpus generator's docstring so anyone regenerating gets a deterministic result. The CI workflow then matches those pins. Brittle long-term but explicit.
2. Loosen the determinism test: compare semantic content (rasterio read of the file) instead of file-byte md5 for fixtures known to depend on the encoder (COG overviews, JPEG). Keep md5 checks for the simple uncompressed fixtures.
3. Regenerate the two drifting fixtures against the conda-forge toolchain and commit the new bytes. Trades the current local-machine baseline for a conda-forge baseline; better for CI but worse for developers who don't have conda-forge installed.

(2) is probably the right shape, since the test's stated purpose is "the generator is reproducible," not "the encoder bytes are stable across versions."

Tracking this so the new conda-forge workflow can drop its `--ignore` once the determinism test is toolchain-agnostic.

brendancol:
## Summary

- `test_corpus_determinism.py` picks the comparison method per fixture. Byte-md5 for fixtures that are stable across GDAL / libjpeg builds; a semantic rasterio-read comparison (dtype, shape, transform, CRS, nodata, overview chain, pixels) for `cog: true` and `compression: jpeg` cells where the encoder output is toolchain-coupled.
- Lossy cells (`tolerance.lossy: true` in the manifest, currently the JPEG-YCbCr fixture) skip the pixel-equality check.
- The `pytest-geotiff-corpus` workflow drops its `--ignore=test_corpus_determinism.py` flag now that the test is no longer toolchain-coupled.

Closes #2299.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/golden_corpus/test_corpus_determinism.py`: 33 passed, 1 skipped.
- [x] `pytest xrspatial/geotiff/tests/golden_corpus/`: 94 passed, 1 skipped.
- [x] Manual negative test: doctored copies that flip a pixel are rejected by `_assert_semantic_equal`.
- [ ] conda-forge `pytest-geotiff-corpus` workflow goes green on PR with the COG and JPEG fixtures in scope.


brendancol (commented):
## PR Review: Make corpus determinism check toolchain-agnostic

### Blockers
None.

### Suggestions
- `_assert_semantic_equal` only compares base-IFD pixels even when both files have an overview chain (test_corpus_determinism.py:143-154). For the COG fixture, an overview-pixel regression in the generator would slip past: the decimation factor list matches but the overview pixels never get read. Consider iterating the overview levels for lossless cells with `rasterio.open(path, OVERVIEW_LEVEL=lvl)` and comparing each level's pixel array. `_oracle.py` already uses that pattern in `compare_to_oracle`.
- Lossy JPEG cells skip pixel comparison entirely (line 147-148). A pixel-content regression in the generator's JPEG branch (wrong input, wrong band order before encode) would not be caught here. A coarse band-wise mean check within a loose tolerance would close the gap without re-introducing libjpeg coupling.

### Nits
- `_is_encoder_sensitive`, `_assert_semantic_equal`, and `_nodata_equal` have no direct unit tests. The drift-detection paths were verified manually in the PR; a tiny test that builds a doctored copy with a flipped pixel and asserts `_assert_semantic_equal` raises would lock the negative path in.
- The tmp-dir name in `regenerated_dir` is still `"regen_corpus_1930"` (line 178). Harmless but misleading now that the test's scope was extended in #2299.

### What looks good
- Dispatch is driven from manifest fields (`cog`, `compression`, `tolerance.lossy`), not hard-coded fixture ids, so new encoder-sensitive cells inherit the right comparison mode automatically.
- The semantic check covers dtype, shape, transform, CRS, nodata (NaN-aware via `_nodata_equal`), overview decimation chain, and lossless pixels.
- The workflow comment explains why the `--ignore` is gone instead of leaving a silent loosening.

### Checklist
- [x] Algorithm matches reference: no algorithm; test-infra change
- [x] All implemented backends produce consistent results: rasterio-only
- [x] NaN handling is correct: `_nodata_equal` + `equal_nan=True`
- [x] Edge cases: missing-fixture skip preserved
- [ ] Overview-level pixels exercised: not yet (see suggestion)
- [x] Dask chunk boundaries: n/a
- [x] No premature materialization: n/a
- [x] Benchmark: n/a
- [x] README feature matrix: n/a

brendancol (commented):
## Follow-up review

All four findings from the prior review pass were addressed in `f5aa2141`:

- Suggestion 1 (overview-level pixels): `_assert_semantic_equal` now iterates `rasterio.open(..., OVERVIEW_LEVEL=lvl)` for every level the file declares and compares each level's pixel array (lossless cells). See `_assert_overview_pixels_exact` at test_corpus_determinism.py:194.
- Suggestion 2 (lossy mean tolerance): `_assert_pixels_close_lossy` at test_corpus_determinism.py:209 computes per-band means and rejects drift above `_LOSSY_PIXEL_MEAN_TOL = 4.0` (0-255 scale).
- Nit 1 (drift unit tests): two new tests cover the negative path: `test_semantic_equal_rejects_lossless_pixel_drift` and `test_semantic_equal_rejects_lossy_mean_drift` (test_corpus_determinism.py:357 and :371).
- Nit 2 (tmp-dir rename): `tmp_path_factory.mktemp("regen_corpus_determinism")` at test_corpus_determinism.py:256.

The branch was rebased onto `origin/main` after the follow-up so the PR diff is now scoped to the two #2299 files; force-push was clean.

Local run: `pytest xrspatial/geotiff/tests/golden_corpus/`: 96 passed, 1 skipped.

No new blockers, suggestions, or nits from this pass.

PR #2419, opened 2026-05-26
Tolerate unreadable .ovr sidecar on base read (#2416)

### LINKED ISSUE #2416: Corrupt .ovr sidecar breaks stable base read in eager paths
**Describe the bug**

The release contract puts `reader.local_file` at the stable tier and `reader.sidecar_ovr` at advanced (`docs/source/reference/geotiff_release_contract.md:82`). The eager reader does not honour that split. `_read_to_array` in `xrspatial/geotiff/_reader.py:211` discovers and parses a sibling `.ovr` before IFD selection, even when the caller did not ask for an external overview. The GPU eager path in `xrspatial/geotiff/__init__.py` calls `find_sidecar` / `load_sidecar` the same way.

A stale, truncated, or malformed `.ovr` next to a healthy base TIFF makes the base read fail. `.ovr` is an external file the caller often does not control: gdaladdo droppings, third-party tooling, a partial download, leftovers from a failed build. The sidecar fault is advanced-tier, but the failure surfaces on the stable-tier base read.

**Expected behavior**

A request for the base image (`overview_level=None` or `overview_level=0`) should succeed against a valid base TIFF regardless of whether the sibling `.ovr` parses. Two reasonable fixes:

1. Defer sidecar discovery / parse until an external overview level is actually requested.
2. Catch sidecar parse failures during full-resolution reads, warn, and fall through to base-file-only behaviour.

The dask metadata path uses `discover_remote_sidecar`, which already swallows sidecar parse failures (`xrspatial/geotiff/_sidecar.py:356-364`). The eager paths should match that contract.

**Reproduction**

```python
from pathlib import Path
from xrspatial.geotiff import open_geotiff, write_geotiff
# ... write a valid base.tif via write_geotiff ...
Path("base.tif.ovr").write_bytes(b"not a tiff")
open_geotiff("base.tif")
# ValueError: Invalid TIFF byte order marker: b'no'
```

**Additional context**

Tests should cover both `open_geotiff` (CPU eager path) and `read_geotiff_gpu` (GPU eager path) since both call `find_sidecar` / `load_sidecar` before IFD selection.

brendancol:
Closes #2416.

## Summary

- The release contract puts `reader.local_file` at the stable tier and `reader.sidecar_ovr` at advanced. Before this change a stale or malformed sibling `.ovr` would take the stable base read down on all three eager paths (CPU, GPU, metadata-only helper).
- Catch sidecar load failures in `_read_to_array`, `_read_geo_info`, and the GPU backend; emit a `RuntimeWarning` and fall through to base-only. `CloudSizeLimitError` still propagates because the byte budget is a caller-set contract.
- Requesting a specific external overview level (`overview_level >= 1`) still surfaces the underlying parse error. Silent fallback only applies when the caller did not ask for the sidecar surface.
- Pre-existing `discover_remote_sidecar` (dask metadata path) already followed this rule. The eager paths now match.

## Backend coverage

- numpy: covered by `_read_to_array` change.
- cupy: covered by `_backends/gpu.py` change.
- dask+numpy and dask+cupy: route through `_read_geo_info` (covered) and `discover_remote_sidecar` (already tolerant since #2239).

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_sidecar_bad_does_not_break_base_2416.py` (12 new tests including 5 parametrized payloads, GPU eager path, CloudSizeLimitError pass-through, explicit-level error surfacing).
- [x] `pytest xrspatial/geotiff/tests/test_sidecar_*.py xrspatial/geotiff/tests/test_remote_sidecar_*.py` (70 existing sidecar tests still pass).
- [x] Full geotiff suite: `pytest xrspatial/geotiff/tests/` -> 5731 passed, 68 skipped, 6 xfailed.

### DIFF
--- a/docs/source/reference/geotiff_release_contract.md
+++ b/docs/source/reference/geotiff_release_contract.md
@@ -89,7 +89,7 @@ category. The `Key` column matches the runtime key.
 | `reader.http` | advanced | Plain HTTP/HTTPS reads; SSRF and private-host filters apply. |
 | `reader.http_cog` | advanced | HTTP COG with range-request fetching. The transport surface (redirects, retries) is not yet contracted at the stable bar. |
 | `reader.vrt` | advanced | Simple VRT mosaics. Full GDAL VRT parity is out of scope. |
-| `reader.sidecar_ovr` | advanced | External `.tif.ovr` sidecar overviews. |
+| `reader.sidecar_ovr` | advanced | External `.tif.ovr` sidecar overviews. A stale or malformed sidecar does not break a base read: the eager CPU, eager GPU, and dask metadata paths warn and fall through to base-file-only behaviour for `overview_level=None`/`0`. Requesting a specific external level surfaces the underlying parse error. Caller-set `max_cloud_bytes` breaches still raise `CloudSizeLimitError` either way. See issue #2416. |
 | `reader.allow_rotated` | experimental | Opt-in `allow_rotated=True`; drops the axis-aligned `transform` attr in favour of `rotated_affine`. |
 | `reader.allow_unparseable_crs` | experimental | Opt-in escape hatch for CRS strings pyproj cannot parse. |
 | `reader.gpu` | experimental | GPU read path; no cross-backend numerical parity claim. |
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -275,22 +275,42 @@ def _read_geo_info(source, *, overview_level: int | None = None,
         # Append sibling `.tif.ovr` sidecar IFDs onto the pyramid list
         # so ``overview_level`` indexes both internal and external
         # overviews (issue #2112). Local file paths only.
+        #
+        # A broken sidecar must not break the base read. The release
+        # contract puts ``reader.local_file`` at the stable tier and
+        # ``reader.sidecar_ovr`` at advanced; a stale or corrupt
+        # ``.ovr`` written by an external tool falls back to base-only
+        # behaviour with a warning. Mirrors the eager CPU path in
+        # ``_reader._read_to_array`` and the dask metadata helper
+        # ``_sidecar.discover_remote_sidecar``. Issue #2416.
         from ._sidecar import attach_sidecar_origin, find_sidecar, load_sidecar
         sidecar_origin: dict[int, tuple] = {}
         sidecar_path = find_sidecar(source)
         if sidecar_path is not None:
-            sidecar = load_sidecar(sidecar_path)
-            # The origin mapping is consumed below for georef extraction
-            # only -- strip/tile bytes are sliced by ``read_to_array`` on
-            # the actual read. A sidecar IFD that carries its own
-            # GeoKeyDirectory / ModelPixelScale / ModelTiepoint /
-            # ModelTransformation needs the sidecar's byte order to
-            # parse cleanly; without the mapping the helper falls back
-            # to the base file's bytes (today's default, correct under
-            # the usual GDAL convention). See issue #2315.
-            sidecar_origin = attach_sidecar_origin(
-                sidecar.ifds, sidecar.data, sidecar.header)
-            ifds = ifds + sidecar.ifds
+            try:
+                sidecar = load_sidecar(sidecar_path)
+            except Exception as exc:
+                warnings.warn(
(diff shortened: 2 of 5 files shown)


brendancol (commented):
## Review

The change is tight and the test surface is solid. A few items below; none are blockers.

### Blockers
None.

### Suggestions
- `xrspatial/geotiff/__init__.py:290-301` and `xrspatial/geotiff/_backends/gpu.py:419-435`: the `_read_geo_info` and GPU eager paths catch a bare `Exception` and do not re-raise `CloudSizeLimitError` the way `_reader._read_to_array` does. In practice both paths are local-file-only at this call site, so the budget error cannot fire today, but the asymmetry is easy to miss the next time someone routes a cloud source through here. Either re-raise `CloudSizeLimitError` for symmetry with `_reader.py:244` or add a one-line comment at each site noting "local-file-only here; budget breach cannot fire" so the next reader does not wonder.
- `xrspatial/geotiff/_reader.py:248` (and the two parallel sites): `stacklevel=2` points at the immediate caller of `_read_to_array`, not at the user's `open_geotiff` / `read_geotiff_dask` call. The warning location in the user's traceback may end up inside the library rather than at the user's call line. Consider bumping to `stacklevel=3` or higher after walking the actual call chain.

### Nits
- The warning text says "Request a specific external overview level to surface the error instead." That is accurate but a bit indirect. A short hint like "Delete the `.ovr` file or pass `overview_level=N >= 1` to surface the parse error" is more actionable. Optional.
- `xrspatial/geotiff/tests/test_sidecar_bad_does_not_break_base_2416.py` has a `_gpu_available()` helper that duplicates the same pattern used in `test_sidecar_ovr_2112.py`. Not worth a refactor in this PR, but a candidate for a shared `conftest.py` fixture if anyone consolidates later.

### What looks good
- Three call sites covered: `_reader._read_to_array`, `__init__._read_geo_info`, `_backends.gpu.read_geotiff_gpu`. The contract is consistent across CPU eager, GPU eager, and dask metadata paths.
- `CloudSizeLimitError` re-raise on the CPU eager path keeps the caller-set byte budget visible.
- Tests pin the level-1-still-raises behaviour, the `overview_level=0` path, the `CloudSizeLimitError` pass-through (via monkeypatch on the local-file path), and five corrupt-payload shapes.
- Module-level `import warnings` added to `_reader.py` rather than function-scoped, avoiding the `UnboundLocalError` trap that bit the GPU file before the fix.
- Release-contract row updated so the tier promise is documented, not just implemented.

### Checklist
- [x] Algorithm matches reference: contract aligns with `_sidecar.discover_remote_sidecar`'s existing dask-path behaviour.
- [x] Backends consistent: CPU eager, GPU eager, metadata-only, dask metadata helper all follow the same rule.
- [x] NaN handling: not applicable (read-path control flow).
- [x] Edge cases tested: empty file, short file, gzip magic, PNG magic, plain text; plus `overview_level=0` and `overview_level=1`.
- [x] Benchmark: not applicable (no perf-sensitive change).
- [x] README feature matrix: not applicable (no new function).
- [x] Docstrings present.

brendancol (commented):
## Follow-up review after 2ddafa06

All Suggestions and Nits from the prior review applied:

### Fixed
- Suggestion 1: Re-raise `CloudSizeLimitError` on the GPU eager path (`_backends/gpu.py:421-429`) and the metadata-only helper (`__init__.py:292-300`). Both sites now match `_reader._read_to_array:244` for symmetry. `CloudSizeLimitError` added to module-level imports in `__init__.py` and `_backends/gpu.py`.
- Suggestion 2: `warnings.warn` `stacklevel` bumped from 2 to 3 at all three call sites so the warning resolves at the user's `open_geotiff` / `read_geotiff_dask` call rather than inside the library.
- Nit 1: Warning text rewritten to give an actionable next step ("Delete the .ovr file or pass overview_level>=1 to surface the parse error.").

### Dismissed
- Nit 2 (shared `_gpu_available()` fixture): out of scope for this PR. The duplication is in a parallel test file (`test_sidecar_ovr_2112.py`); consolidating it would touch unrelated tests.

### New test
- `test_read_geo_info_cloud_size_limit_error_is_not_silenced` pins the re-raise on the metadata-only path so the new symmetry stays covered.

### Verification
- `pytest xrspatial/geotiff/tests/`: 5732 passed, 68 skipped, 6 xfailed.

PR #2420, opened 2026-05-26
Reject compound EPSG codes on the stable-EPSG writer path (#2418)

### LINKED ISSUE #2418: Writer emits structurally wrong CRS metadata for compound / vertical / geocentric EPSG codes
**Describe the bug**

The stable-path EPSG writer in `to_geotiff` writes structurally wrong CRS metadata for compound, vertical, and geocentric CRSs. The public docstring at `xrspatial/geotiff/_writers/eager.py:120` says integer EPSG codes are stable. Validation in `xrspatial/geotiff/_crs.py:86` accepts any code pyproj can resolve. But `_model_type_from_epsg` in `xrspatial/geotiff/_geotags.py:1418` only branches on `is_geographic` and treats everything else as projected, and `build_geo_tags` at `xrspatial/geotiff/_geotags.py:1555` only knows how to emit `GeographicTypeGeoKey` (2048) or `ProjectedCSTypeGeoKey` (3072).

So a valid integer EPSG goes in, passes validation, and gets written into the wrong GeoKey slot. The xrspatial reader hides it on round-trip by re-resolving the integer through pyproj. rasterio / GDAL cannot recover the CRS.

**Reproduction**

EPSG:6349 is "NAD83(2011) + NAVD88 height", a compound horizontal + vertical CRS. pyproj reports `is_geographic=True` because pyproj surfaces the horizontal sub-CRS through that property, so `_model_type_from_epsg` returns `MODEL_TYPE_GEOGRAPHIC` and the writer stores 6349 in `GeographicTypeGeoKey`.

```python
import numpy as np, xarray as xr, tempfile, rasterio

arr = xr.DataArray(np.zeros((4, 4), dtype=np.float32))
with tempfile.NamedTemporaryFile(suffix=".tif", delete=False) as tmp:
    path = tmp.name

to_geotiff(arr, path, crs=6349)

with rasterio.open(path) as src:
    print("rasterio CRS:", src.crs)
    print("rasterio to_epsg:", src.crs.to_epsg())

da = open_geotiff(path)
print("xrspatial crs attr:", da.attrs.get("crs"))
```

Output:

```
rasterio CRS: GEOGCS["unknown",DATUM["unnamed",SPHEROID["unretrievable - using WGS84",6378137,298.257223563]],PRIMEM["Greenwich",0],UNIT["unknown",0.0174532925199433],AXIS["Latitude",NORTH],AXIS["Longitude",EAST],AUTHORITY["EPSG","6349"]]
rasterio to_epsg: None
xrspatial crs attr: 6349
```

rasterio sees a GeoTIFF that claims AUTHORITY=EPSG:6349, but the GeoKeys describe an unknown geographic CRS with the WGS84 ellipsoid, so `to_epsg()` returns `None`. xrspatial's own reader masks the corruption by resolving the integer back through pyproj.

The same failure mode applies to other non-horizontal CRSs the writer currently accepts:

- Geocentric CRS (e.g. EPSG:4978): `is_geographic` and `is_projected` are both False, so the writer falls into the `else` branch and tags it as a projected 2D CRS.
- Vertical-only CRS (e.g. EPSG:5773): same as above.
- Other compound horizontal+vertical CRSs (e.g. EPSG:5498).

**Expected behavior**

Two options:

1. The writer recognises non-representable CRSs (compound, vertical-only, geocentric) and rejects them with a clear error pointing callers at the WKT-only fallback. The stable-path contract then guarantees that any integer EPSG accepted on the input side produces a GeoTIFF whose CRS round-trips through rasterio / GDAL with `to_epsg()` returning the input code.

2. The writer implements proper GeoKey emission for those families (`VerticalCSTypeGeoKey`, geocentric ModelType, compound CRS via citation + sub-CRS keys).

Option (1) is the smaller fix and matches the existing posture of "EPSG codes are strongly preferred for interop; the WKT-only path emits 32767".

**Additional context**

Any test added for this fix has to validate output through rasterio (or another external GeoTIFF reader), not only through xrspatial's own reader. The current reader masks the bug by re-resolving the stored integer through pyproj.

Related: issue #1768 (WKT-only path), issue #2277 (model type classification heuristic).

brendancol:
Closes #2418.

## Summary

- The integer-EPSG writer path emits only `GeographicTypeGeoKey` (2048) and `ProjectedCSTypeGeoKey` (3072). Both are spec'd to hold a 2D horizontal CRS code. Pass a compound CRS like EPSG:6349 and pyproj reports `is_geographic=True` (because the horizontal sub-CRS is geographic), so the writer stores the compound code in the horizontal slot. rasterio / GDAL reads the file back as the wrong CRS or as no CRS at all.
- Add `NonRepresentableEPSGCRSError` and reject compound EPSG codes from `_validate_crs_arg`. Defense in depth in `_model_type_from_epsg` catches a direct call into `build_geo_tags`.
- `_wkt_to_epsg` now returns None for compound CRSs so a caller who passes the full compound WKT routes through the user-defined CRS / citation path (#1768) instead of being downgraded into the broken integer path.

## Backend coverage

Pure metadata fix in `_crs.py` / `_geotags.py` / `_errors.py`. Numpy, cupy, dask+numpy, and dask+cupy writers all flow through the same `_validate_crs_arg` gate, so the fix applies to every backend.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_compound_crs_reject_2418.py` (22 new tests, including rasterio round-trip assertions for EPSG:6349, 5498, 7415, 8360 and regression coverage for plain horizontal CRSs).
- [x] `pytest xrspatial/geotiff/tests/` (4863 passed, 25 skipped, 3 xfailed; no regressions).
- [x] Manual repro from the issue body now raises `NonRepresentableEPSGCRSError` at the `to_geotiff(..., crs=6349)` call instead of silently producing a corrupt GeoTIFF.

### DIFF
--- a/docs/source/user_guide/geotiff_safe_io.rst
+++ b/docs/source/user_guide/geotiff_safe_io.rst
@@ -235,6 +235,13 @@ the ambiguous-metadata family at once.
      - The writer cannot classify an EPSG code as geographic or
        projected.
      - Pass a code pyproj can resolve, or install pyproj.
+   * - :class:`~xrspatial.geotiff.NonRepresentableEPSGCRSError`
+     - The integer EPSG code resolves to a compound (horizontal +
+       vertical) CRS, which the writer cannot represent in a single
+       ``GeographicTypeGeoKey`` or ``ProjectedCSTypeGeoKey`` slot.
+     - Pass the full compound CRS as WKT to take the user-defined CRS
+       fallback path, or pass the horizontal sub-CRS EPSG directly if
+       the vertical component is not needed.
    * - :class:`~xrspatial.geotiff.UnsupportedGeoTIFFFeatureError`
      - The input declares a feature the GeoTIFF module does not
        implement (warped / reprojection VRTs, pansharpened or derived
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -72,9 +72,9 @@ from ._coords import \
     transform_tuple_from_pixel_geometry as _transform_tuple_from_pixel_geometry  # noqa: F401
 from ._crs import _resolve_crs_to_wkt, _wkt_to_epsg  # noqa: F401
 from ._errors import (ConflictingCRSError, ConflictingNodataError, GeoTIFFAmbiguousMetadataError,
-                      InvalidCRSCodeError, MixedBandMetadataError, NonUniformCoordsError,
-                      RotatedTransformError, UnknownCRSModelTypeError, UnparseableCRSError,
-                      UnsupportedGeoTIFFFeatureError)
+                      InvalidCRSCodeError, MixedBandMetadataError, NonRepresentableEPSGCRSError,
+                      NonUniformCoordsError, RotatedTransformError, UnknownCRSModelTypeError,
+                      UnparseableCRSError, UnsupportedGeoTIFFFeatureError)
 from ._geotags import RASTER_PIXEL_IS_AREA, RASTER_PIXEL_IS_POINT, GeoTransform  # noqa: F401
 from ._reader import _MAX_CLOUD_BYTES_SENTINEL, UnsafeURLError
 from ._reader import read_to_array as _read_to_array
@@ -110,6 +110,7 @@ __all__ = [
     'GEOREF_STATUS_VALUES',
     'InvalidCRSCodeError',
     'MixedBandMetadataError',
+    'NonRepresentableEPSGCRSError',
     'NonUniformCoordsError',
     'RotatedTransformError',
     'SUPPORTED_FEATURES',
--- a/xrspatial/geotiff/_crs.py
+++ b/xrspatial/geotiff/_crs.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 import numbers
 import warnings
 
-from ._errors import UnparseableCRSError
+from ._errors import NonRepresentableEPSGCRSError, UnparseableCRSError
 from ._runtime import GeoTIFFFallbackWarning, _geotiff_strict_mode
 
 #: WKT root keywords. A string that starts with one of these (after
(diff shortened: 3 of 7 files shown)


brendancol (commented):
## PR Review: Reject compound EPSG codes on the stable-EPSG writer path (#2418)

None. The fix is correct, and the reproduction from the issue body now raises at the call site instead of corrupting the file on disk.

- `xrspatial/geotiff/_geotags.py:1463`: lazy `from ._errors import NonRepresentableEPSGCRSError` inside `_model_type_from_epsg`. `_errors` is already imported at the top of the module (line 7 imports `UnknownCRSModelTypeError` from the same module), so there is no circular-import reason for the deferred import. Move it to the top alongside the existing `_errors` import.

- `xrspatial/geotiff/tests/test_compound_crs_reject_2418.py:154`: `test_compound_epsg_does_not_round_trip_via_integer_path` is a tautology. The docstring says it demonstrates corruption by bypassing the validator, but the body only asserts pyproj structural properties (`crs.is_compound is True`, `crs.is_geographic is True`). It does not exercise the write path. Either remove it, or call into `build_geo_tags` with the validator monkey-patched and assert the rasterio readback is broken. The latter would be a useful regression in case a future refactor drops either guard.

- `xrspatial/geotiff/_crs.py:48`: the helper name `_reject_non_representable_epsg` is a mouthful. The current name is fine for clarity at the call site; no change needed unless someone has a shorter idea.

- `xrspatial/geotiff/__init__.py:74`: the multi-line `from ._errors import (...)` is getting wide. Not introduced by this PR; an isort pass would tidy it.

- GPU writer at `xrspatial/geotiff/_writers/gpu.py:472` calls `_validate_crs_arg`, so the fix flows through. No explicit GPU-path test for compound EPSG rejection. Low value to add given the validator is shared, but worth noting that GPU coverage is structural only.

### What looks good

- The fix is properly scoped. Only compound CRSs are rejected. Non-compound non-2D codes (EPSG:4979 3D-geographic, EPSG:4978 geocentric, EPSG:5773 vertical-only) are left alone because they round-trip through rasterio with the input code intact. `test_geographic_3d_crs_still_works` pins that scoping.
- The validator gate plus the defense-in-depth check in `_model_type_from_epsg` means a future caller that bypasses `_validate_crs_arg` still cannot write a compound EPSG to disk.
- `_wkt_to_epsg` returning None for compound CRSs is the right call. It routes the compound-WKT path to the user-defined CRS / citation fallback instead of silently downgrading to the broken integer path.
- The negative-case tests use rasterio for round-trip assertions, which is what the issue body asked for. The xrspatial reader would mask the bug.
- The error message names the failing EPSG, names the CRS, and offers both the WKT fallback and the horizontal sub-CRS workaround.

### Checklist
- [x] Algorithm matches the spec (GeoTIFF GeoKey emission rules, Annex A)
- [x] All implemented backends produce consistent results (fix lives in shared validation)
- [x] NaN handling not applicable (metadata-only change)
- [x] Edge cases covered (compound geographic, compound projected, attrs path, kwarg path)
- [x] Dask chunk boundaries not relevant
- [x] No benchmark needed (validation-only change)

brendancol (commented):
## Follow-up Review (post-fixes)

Re-reviewed after commit ffcaee42.

### Original findings -- disposition

- **Suggestion: hoist `NonRepresentableEPSGCRSError` import** -- fixed. `_geotags.py:7` now imports it alongside `RotatedTransformError` and `UnknownCRSModelTypeError`; the in-function import in `_model_type_from_epsg` is gone.
- **Suggestion: tautology test** -- fixed. Replaced with `test_compound_epsg_corruption_surfaces_when_validator_bypassed`, which monkey-patches both gates and asserts `rasterio_epsg != 6349` from the result. The new test pins the actual corruption the fix prevents.
- **Nit: helper name `_reject_non_representable_epsg`** -- dismissed. Name is descriptive at the call site, no shorter alternative offered, rename would be pure churn.
- **Nit: wide `__init__.py` import line** -- dismissed. Pre-existing, explicitly out of scope per the original finding.
- **Nit: no explicit GPU test** -- dismissed. The reviewer themselves flagged it as low-value. `_validate_crs_arg` is the gate and has direct unit tests; the GPU writer calls it at `_writers/gpu.py:472`.

### Current state

No new findings. The 23 tests in the new file pass. The full `xrspatial/geotiff/tests/` suite (4863 tests) still passes locally.

Ready for human review.

PR #2447, opened 2026-05-26
Reject VRT sources under stable_only=True (#2443)

### LINKED ISSUE #2443: Reject mixed-tier VRT child sources under stable_only=True (epic #2342)
## Summary

The release-gate test `test_release_gate_negative_mixed_tier_vrt_children` pins a promise from epic #2342: when a caller asks for stable-only sources and a VRT child uses an experimental codec, the reader must name the offending child and the opt-in flag. The knob does not exist yet, so the test is `xfail(strict=False)` today.

## Proposed change

Add `stable_only: bool = False` to `open_geotiff`, `read_geotiff_dask`, `read_geotiff_gpu`, and `read_vrt`. When `True`, the VRT capability validator rejects any source whose codec tier in `SUPPORTED_FEATURES` is not `stable`. The rejection raises a `GeoTIFFAmbiguousMetadataError` subclass that names the offending source path and the `allow_experimental_codecs` opt-in, and cites the release-contract docs.

## Scope

- Fires at graph-build / eager-read setup time so the failure surfaces before any pixel decode.
- Non-VRT call sites with `stable_only=True` are a no-op; the per-source codec gate already covers them via `_validate_read_codec_optin`.
- Default behaviour (`stable_only=False`) is unchanged.

## Acceptance

- `test_release_gate_negative_mixed_tier_vrt_children` flips from xfail to a regular pass.
- A `GeoTIFFAmbiguousMetadataError` subclass names the source path and the `allow_experimental_codecs` opt-in.
- The kwarg threads through every public read entry point in the canonical position from `_CANONICAL_ORDER`.

## Related

- Epic #2342.
- Precedent: PR #2442 (typed error + opt-in kwarg + canonical-order entry for `allow_invalid_nodata`).

brendancol:
## Summary

- Add `stable_only: bool = False` to `open_geotiff`, `read_geotiff_dask`, `read_geotiff_gpu`, and `read_vrt`. When True on a VRT source, the read raises `VRTStableSourcesOnlyError` (a `GeoTIFFAmbiguousMetadataError` subclass) at graph-build / eager-read setup time, before any pixel decode.
- The rejection message names the offending VRT path, the `stable_only` flag, and the `allow_experimental_codecs` unlock, and cites `docs/source/reference/release_gate_geotiff.rst` plus epic #2342.
- Flips `test_release_gate_negative_mixed_tier_vrt_children` from xfail to a regular pass.

## Backend coverage

Numpy, cupy, dask+numpy, dask+cupy. VRT routing happens in `open_geotiff` and `read_geotiff_dask`; both entry points forward the kwarg to `read_vrt`, which carries the validator call. The GPU eager path does not open VRT sources directly, so the flag is a no-op there (kept on the signature for cross-backend symmetry, per the canonical-order test).

## Test plan

- [x] `pytest xrspatial/geotiff/tests/test_vrt_stable_only_2443.py` (7 new tests: default rejection on each entry point, default-false no-op, `allow_experimental_codecs` unlock, error subclass invariant)
- [x] `pytest xrspatial/geotiff/tests/release_gates/test_stable_features.py::test_release_gate_negative_mixed_tier_vrt_children` (xfail flipped to pass)
- [x] `pytest xrspatial/geotiff -n auto` (5811 passed, 4 unrelated xfailed)
- [x] `pytest xrspatial -n auto` (10675 passed; two `*_dask_temp_cleanup` failures are pre-existing flakes unrelated to this change)

Closes #2443.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -76,7 +76,8 @@ from ._errors import (ConflictingCRSError, ConflictingNodataError, GeoTIFFAmbigu
                       MixedBandMetadataError,
                       NonRepresentableEPSGCRSError, NonUniformCoordsError, RotatedTransformError,
                       UnknownCRSModelTypeError,
-                      UnparseableCRSError, UnsupportedGeoTIFFFeatureError)
+                      UnparseableCRSError, UnsupportedGeoTIFFFeatureError,
+                      VRTStableSourcesOnlyError)
 from ._geotags import RASTER_PIXEL_IS_AREA, RASTER_PIXEL_IS_POINT, GeoTransform  # noqa: F401
 from ._reader import _MAX_CLOUD_BYTES_SENTINEL, CloudSizeLimitError, UnsafeURLError
 from ._reader import read_to_array as _read_to_array
@@ -122,6 +123,7 @@ __all__ = [
     'UnparseableCRSError',
     'UnsafeURLError',
     'UnsupportedGeoTIFFFeatureError',
+    'VRTStableSourcesOnlyError',
     'open_geotiff',
     'read_geotiff_gpu',
     'read_geotiff_dask',
@@ -389,6 +391,7 @@ def open_geotiff(source: str | BinaryIO, *,
                  allow_unparseable_crs: bool = False,
                  allow_inconsistent_geokeys: bool = False,
                  allow_invalid_nodata: bool = False,
+                 stable_only: bool = False,
                  allow_experimental_codecs: bool = False,
                  allow_internal_only_jpeg: bool = False,
                  band_nodata: str | None = None,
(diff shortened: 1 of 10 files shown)


brendancol (commented):
## PR Review: Reject VRT sources under stable_only=True (#2443)

None.

None substantive. The PR mirrors the #2442 precedent closely.

- `xrspatial/geotiff/_validation.py:715-737`: the validator does not verify the source is actually a VRT. `read_vrt` is the only caller and is contractually VRT-only, so this is defensive at worst -- but the message says `"VRT source '{source}'"` unconditionally, which would mislabel a non-VRT path if someone called `_validate_stable_only_vrt` directly from a future code path. Optional: tighten the docstring to say "the caller must have already classified source as a VRT," or sniff `source.lower().endswith('.vrt')` and pass through silently otherwise.
- `xrspatial/geotiff/tests/test_vrt_stable_only_2443.py:65-79`: the "unlock" test asserts that with `allow_experimental_codecs=True` we do not get `VRTStableSourcesOnlyError`, but does not pin the downstream error (still `VRTUnsupportedError` for the no-bands case). Slight risk that a future refactor silently broadens the unlock past intent. Optional: assert `pytest.raises(VRTUnsupportedError)` instead of `Exception`.

### What looks good
- Matches the typed-error-plus-opt-in-kwarg shape from PR #2442 / #2441: subclass of `GeoTIFFAmbiguousMetadataError`, kwarg threaded through every public reader, canonical-order test updated alongside.
- Error message satisfies both release-gate assertions (names the opt-in flag and cites `release_gate_geotiff` plus `#2342`).
- Gate fires at the top of `read_vrt`, before `_validate_dispatch_kwargs`, so the typed error is the first raise the caller sees.
- New test file covers all four entry points (eager VRT, direct VRT, dask VRT, the unlock path) plus the subclass invariant.

### Checklist
- [x] Algorithm matches reference: matches the release-gate test contract.
- [x] All implemented backends produce consistent results: validator runs in `read_vrt`; both routing paths (`open_geotiff` and `read_geotiff_dask`) forward the kwarg.
- [x] NaN handling is correct: N/A (validation-only).
- [x] Edge cases are covered by tests: default-false no-op, unlock path, subclass invariant.
- [x] Dask chunk boundaries handled correctly: N/A.
- [x] No premature materialization or unnecessary copies: validator raises before any decode.
- [x] Benchmark exists or is not needed: not needed.
- [x] README feature matrix updated (if applicable): not applicable.
- [x] Docstrings present and accurate: every entry point's docstring carries the new kwarg.

brendancol (commented):
## Follow-up Review: nits addressed (commit 2a642dd2)

- **Nit 1** (validator does not check for VRT extension): **Fixed.** ``_validate_stable_only_vrt`` now passes through silently when ``source`` is not a string ending in ``.vrt`` (case-insensitive). Docstring updated to spell out the pass-through.
- **Nit 2** (unlock test asserts only ``Exception``): **Fixed.** Both the default-false and ``allow_experimental_codecs`` unlock tests now assert ``pytest.raises(VRTUnsupportedError)`` so a future refactor cannot silently broaden either branch.

### Verification

- ``pytest xrspatial/geotiff/tests/test_vrt_stable_only_2443.py xrspatial/geotiff/tests/release_gates/test_stable_features.py::test_release_gate_negative_mixed_tier_vrt_children xrspatial/geotiff/tests/test_features.py::TestPublicAPI xrspatial/geotiff/tests/test_reader_kwarg_order_1935.py`` -- 16 passed, 0 failed.

No further blockers, suggestions, or nits.

PR #2448, opened 2026-05-26
Reject mixed-CRS VRTs at validator (#2444)

### LINKED ISSUE #2444: geotiff: VRT read silently flattens mixed-CRS mosaic (#2321 gap)
## Reason or Problem

`read_vrt` silently flattens a mixed-CRS mosaic to the VRT-declared SRS. If an underlying source TIFF carries a different CRS than the VRT's `<SRS>`, the read still succeeds, `attrs['crs']` reports the VRT-declared CRS, and the pixel content from the disagreeing source has been written into the mosaic regardless. No error, no warning. The output array claims one projection while some of its pixels came from another.

Two test pins already exist for this gap:

- `test_mixed_crs_vrt_does_not_silently_flatten` in `xrspatial/geotiff/tests/vrt/test_metadata.py` (strict xfail).
- `test_mixed_source_crs_rejected` in `xrspatial/geotiff/tests/vrt/test_validation.py` (non-strict xfail).

Both should flip to pass once the validator opens each source TIFF and checks its CRS.

## Proposal

Add a per-source CRS check to `validate_parsed_vrt` in `xrspatial/geotiff/_vrt_validation.py`. The check opens each `SimpleSource`, pulls its CRS via `extract_geo_info`, and compares to the VRT-declared SRS. On disagreement, raise `VRTUnsupportedError` naming the source path and both CRSes.

**Design:**

- Walk every `_Source` in every `_VRTBand` on the parsed `VRTDataset`.
- For each unique source path, open the file via `_FileSource`, parse the header and first IFD, and pull `crs_wkt` / `crs_epsg` via `extract_geo_info`.
- Compare each source CRS to `parsed.crs_wkt` using `pyproj.CRS.equals` after canonicalisation. If the VRT carries no `<SRS>`, all sources must agree among themselves.
- On disagreement raise `VRTUnsupportedError` with both disagreeing CRSes named.
- Runs at graph-build (chunked path) and eager-read setup (eager path), matching the existing validator wiring in `_backends/vrt.py`.
- If pyproj is not installed, skip the check (same disposition as `_check_write_conflicting_crs`). A string compare would false-positive on equivalent representations.
- If a source is unreadable, skip silently. The existing `missing_sources` path fires on the real read.

**Usage:**

No new public API. The check fires automatically from both `read_vrt` entry points. No opt-out flag: the mosaic reader has no reprojection step, so a mixed-CRS VRT cannot be served correctly.

**Value:** Closes the silent-correctness gap pinned by both xfails. After this change they pass with `VRTUnsupportedError` raised at validation time.

## Stakeholders and Impacts

Affects callers feeding mixed-CRS VRTs to `read_vrt`. They get a typed error instead of a mis-georeferenced mosaic. Single-CRS VRTs (the common case) see no change beyond one extra metadata read per unique source at validation time. Implementation is local to `_vrt_validation.py`; the call sites in `_backends/vrt.py` pick up the new check automatically.

## Drawbacks

One extra mmap + IFD parse per unique source at validation time. Cheap; no pixel decode.

## Alternatives

Defer the check to a per-chunk task. Not used because the documented validator contract is graph-build / eager-read setup, and a chunked-path failure deep in `compute()` is the opaque user experience the centralised validator was added to avoid.

## Unresolved Questions

None.

Test pins: `test_mixed_crs_vrt_does_not_silently_flatten` and `test_mixed_source_crs_rejected`. Threading pattern follows PR #2442.

brendancol:
Closes #2444.

## Summary

- Adds Rule 10 to `validate_parsed_vrt`: walk every `SimpleSource`, open it for metadata-only IFD parsing, and compare its CRS to the VRT-declared `<SRS>` via pyproj canonicalisation. Disagreement raises `VRTUnsupportedError` naming the source path and both CRSes.
- When the VRT carries no `<SRS>`, the first parseable source becomes the reference and the rest must agree with it.
- Flips both xfail pins that tracked this gap (the same validator naturally satisfies both):
  - `test_mixed_crs_vrt_does_not_silently_flatten` in `xrspatial/geotiff/tests/vrt/test_metadata.py` (was strict xfail).
  - `test_mixed_source_crs_rejected` in `xrspatial/geotiff/tests/vrt/test_validation.py` (was non-strict xfail).
- Skip conditions for the new check (no raise): pyproj missing (mirrors `_check_write_conflicting_crs`), source unreadable (existing `missing_sources` path fires on the real read), source has no CRS at all (reader has always inherited the VRT-declared SRS in that case), source CRS pyproj cannot canonicalise (cannot prove inequality).

## Backend coverage

Pure metadata-side validator. Runs once at graph build (chunked VRT path) and once at eager-read setup (eager VRT path). No backend-specific work; numpy / cupy / dask+numpy / dask+cupy all route through the same `_backends/vrt.py` call sites that already invoke `validate_parsed_vrt`.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/vrt/test_metadata.py::test_mixed_crs_vrt_does_not_silently_flatten` passes.
- [x] `pytest xrspatial/geotiff/tests/vrt/test_validation.py::test_mixed_source_crs_rejected` passes.
- [x] Four new positive coverage tests pass: source-path-in-message, chunked-path rejection, homogeneous-CRS round trip, no-`<SRS>` sources-disagree rejection.
- [x] Full `xrspatial/geotiff` suite: 5808 passed, 68 skipped, 3 xfailed (the remaining three xfails are unrelated and tracked elsewhere).
- [x] Non-geotiff suite passes apart from the two pre-existing hydro dask-cleanup flakes called out in the rockout context.

### DIFF
--- a/xrspatial/geotiff/_vrt_validation.py
+++ b/xrspatial/geotiff/_vrt_validation.py
@@ -463,13 +463,17 @@ def validate_parsed_vrt(
     # path) via the existing call sites in ``_backends/vrt.py``, so a
     # mixed-CRS VRT is refused at the same boundary as the other
     # capability checks above.
-    _check_mixed_source_crs(parsed, _unique_source_paths, source=source)
+    _check_mixed_source_crs(
+        _unique_source_paths,
+        vrt_crs_wkt=parsed.crs_wkt,
+        source=source,
+    )
 
 
 def _check_mixed_source_crs(
-    parsed: 'VRTDataset',
     source_paths: list[str],
     *,
+    vrt_crs_wkt: str | None,
     source: str,
 ) -> None:
     """Rule 10: reject a VRT whose sources disagree on CRS.
@@ -477,8 +481,8 @@ def _check_mixed_source_crs(
     Walks ``source_paths`` (already deduplicated by the caller), opens
     each one for metadata-only IFD parsing, and pulls
     ``crs_wkt`` / ``crs_epsg`` via :func:`extract_geo_info`. Compares
-    each source CRS to the VRT-declared ``parsed.crs_wkt`` after pyproj
-    canonicalisation. When ``parsed.crs_wkt`` is empty / None, the
+    each source CRS to the VRT-declared ``vrt_crs_wkt`` after pyproj
+    canonicalisation. When ``vrt_crs_wkt`` is empty / None, the
     sources are required to agree with each other (the first parseable
     source becomes the reference).
 
@@ -526,7 +530,7 @@ def _check_mixed_source_crs(
     # then becomes the first parseable source.
     vrt_crs = None
     vrt_crs_display: str | None = None
-    raw_vrt_srs = parsed.crs_wkt
+    raw_vrt_srs = vrt_crs_wkt
     if raw_vrt_srs:
         try:
             vrt_crs = _PyProjCRS.from_user_input(raw_vrt_srs)
@@ -641,11 +645,17 @@ def _format_crs_for_message(raw: str, crs) -> str:
     EPSG code (the common case for a well-formed source TIFF). Falls
     back to a truncated excerpt of the raw text otherwise so the
     message stays readable for hand-built or non-EPSG WKT strings.
+
+    ``CRS.to_epsg()`` is documented to return ``None`` on failure
+    rather than raise, so the call is unguarded; we still defend
+    against a ``crs`` that does not expose ``to_epsg`` at all (the
+    helper accepts an untyped ``crs`` because tests and future callers
+    may pass a mock or a parsed-but-not-pyproj CRS shim).
     """
-    try:
-        epsg = crs.to_epsg()
-    except Exception:
-        epsg = None
+    epsg = None
+    to_epsg = getattr(crs, 'to_epsg', None)
+    if callable(to_epsg):
+        epsg = to_epsg()
     if epsg is not None:
         return f"EPSG:{epsg}"
     raw = raw.strip()
--- a/xrspatial/geotiff/_vrt_validation.py
+++ b/xrspatial/geotiff/_vrt_validation.py
@@ -25,6 +25,7 @@ with the offending source path and field embedded in the message.
 """
 from __future__ import annotations
 
+import struct
 from typing import TYPE_CHECKING
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Reject mixed-CRS VRTs at validator (#2444)

None.

- `xrspatial/geotiff/_vrt_validation.py:469` - `_check_mixed_source_crs` takes `parsed` as its first positional argument but never reads from it. The caller only needs the deduplicated path list and the outer-VRT path. Drop the unused parameter (and the matching argument at line 466) so the helper's signature does not lie about what it depends on.

- `xrspatial/geotiff/_vrt_validation.py:632` - the `try`/`except`/`finally`/`continue` shape around `_FileSource` works, but it is a little hard to scan. A `with` wrapper would be cleaner, except `_FileSource` is not a context manager today, so that is a `_sources.py` change rather than a PR-2448 fix. File a follow-up if anyone wants to clean up the call sites broadly.

- `xrspatial/geotiff/_vrt_validation.py:637` - `_format_crs_for_message` catches bare `Exception` around `crs.to_epsg()`. `to_epsg()` is documented to return `None` on failure rather than raise, so the catch defends against a pyproj quirk that probably does not exist. Tightening to `(AttributeError, TypeError)` (or dropping the try with a comment) would be marginally clearer. Safe to leave.

### What looks good

- The dedup step (`_unique_source_paths` / `_seen_source_paths`) opens each source file once for metadata even when a multi-band VRT references the same TIFF on every band.
- The CRS check is placed after the cheaper rect / mask / resample rejects, so a malformed VRT that fails one of those still raises before any source file gets opened.
- Skip-on-missing-pyproj matches `_check_write_conflicting_crs`.
- The two flipped xfails plus four new positive tests (homogeneous round trip, chunked-path rejection, source-name-in-message, no-`<SRS>` sources-disagree) cover the rule from several angles.
- Error messages name the offending source path, both CRSes, and a fix to try (`gdalwarp -t_srs` or per-CRS VRTs). VRT mosaics are hard enough to debug already.

### Checklist

- [x] Algorithm matches the contract (pyproj canonicalisation, fail-closed on inequality).
- [x] All backends route through the same validator; chunked path covered by `test_mixed_source_crs_chunked_path_also_rejects`.
- [x] No NaN handling concerns (metadata-only check).
- [x] Edge cases covered: no-`<SRS>`, homogeneous, unreadable source skip, no-CRS source skip.
- [x] Dask chunked path tested.
- [x] No premature materialization (metadata-only IFD reads).
- [N/A] No benchmark needed; the validator is metadata-only and runs once per read.
- [N/A] README feature matrix unchanged; no new public API.
- [x] Docstrings present on the new helper and on the format helper.

brendancol (commented):
## Follow-up review (commit 825d474f)

The two actionable findings from the previous pass have been addressed:

- `_check_mixed_source_crs` now takes `vrt_crs_wkt` directly instead of the full parsed dataset. Helper signature now matches what the function actually reads.
- `_format_crs_for_message` swaps the bare `Exception` catch for a `getattr` probe on `to_epsg`. Real exceptions from a future pyproj refactor would now propagate instead of being swallowed.

The remaining nit (wrap `_FileSource` with `with`) was deferred because `_FileSource` does not implement the context-manager protocol today; pulling that in would expand the diff to `_sources.py` and every caller. Worth a follow-up issue when someone wants to do the broader cleanup.

No new findings. Tests still green: `pytest xrspatial/geotiff -n auto` reports 5809 passed, 68 skipped, 3 xfailed (the three xfailures are unrelated and tracked elsewhere).

PR #2469, opened 2026-05-27
geotiff tests: consolidate nodata read-side cluster (Sub-PR A)

brendancol:
Closes part of #2434 (Sub-PR A). Cluster 10 of long-tail epic #2424.

## Summary

Fold 8 read-side nodata files into the existing `xrspatial/geotiff/tests/read/test_nodata.py` alongside the GPU helper coverage:

- `test_masked_nodata_attr_2092.py`
- `test_nodata_attr_aliases_1582.py`
- `test_nodata_lifecycle_attrs_2135.py`
- `test_nodata_lifecycle_parity_2211.py`
- `test_nodata_nan_int_1774.py`
- `test_nodata_no_extra_copy_1553.py`
- `test_nodata_semantics_split_1988.py`
- `test_helper_band_nodata_2210.py`

Tests-only. The shared `requires_gpu` marker from `_helpers/markers.py` (aliased `_gpu_only`) replaces the per-file `_gpu_available` / `_gpu_only` helpers. Module-level helper functions are suffixed with the source issue number to keep sibling sections collision-free. The rasterio `importorskip` from `test_nodata_semantics_split_1988.py` is localised to per-helper / per-test scope so the rest of the module still collects when rasterio is absent.

185 tests collected (9 pre-existing GPU + 176 newly folded), matching the baseline sum.

## Cross-cutting follow-ups

- `test_invalid_int_nodata_rejection_2441.py` (sibling cluster, not in this PR) imported `_build_uint16_tiff` from the now-deleted #1774 file; the import points at the renamed helper `_build_uint16_tiff_1774` in `read/test_nodata.py`.
- `docs/source/reference/release_gate_geotiff.rst`: the nodata round-trip row cites `read/test_nodata.py` so the release-gate checklist-parity test still passes.
- `docs/source/reference/geotiff.rst`: prose pointers updated.
- `test_roundtrip_properties.py`: stale docstring pointer updated.

## Audit

`xrspatial/geotiff/tests/CLUSTER_AUDIT_NODATA.md` maps every old `file::test` to its new `file::test_id`. The audit file is deleted on a final commit on this branch before merge.

## Verification

- `pytest xrspatial/geotiff/tests/read/test_nodata.py`: 185 passed.
- `pytest xrspatial/geotiff/tests/test_invalid_int_nodata_rejection_2441.py`: 21 passed.
- `pytest xrspatial/geotiff/tests/release_gates/test_stable_features.py::test_release_gate_cites_only_existing_test_files`: 1 passed.

[...]

Issue #2434, opened 2026-05-26
Consolidation cluster 10: nodata (long-tail epic #2424)


brendancol:
Follow-up to epic #2424 / #2390. Cluster 10 of 16. **Parallel-safe.**

Likely 1-2 PRs. The 13 files have a clean read/write split, so a 1 PR vs 2 PR call comes down to review surface.

## Scope

Consolidate nodata semantics, attrs, lifecycle, validation, and round-trip into one parametrised home each on the read and write sides.

### Sub-PR A: read-side -> extend `read/test_nodata.py`

- `test_masked_nodata_attr_2092.py`
- `test_nodata_attr_aliases_1582.py`
- `test_nodata_lifecycle_attrs_2135.py`
- `test_nodata_lifecycle_parity_2211.py`
- `test_nodata_nan_int_1774.py`
- `test_nodata_no_extra_copy_1553.py`
- `test_nodata_semantics_split_1988.py`
- `test_helper_band_nodata_2210.py`

### Sub-PR B: write-side / validation -> extend `write/test_basic.py` or new `write/test_nodata.py`

- `test_mask_nodata_kwarg_2052.py`
- `test_nodata_bool_rejection_1911.py`
- `test_nodata_int64_precision_1847.py`
- `test_nodata_out_of_range_1581.py`
- `test_nodata_validation_1973.py`

## Boundary

GPU nodata (`test_gpu_nodata_1542.py`, `test_mask_nodata_gpu_vrt_2052.py`, `test_gpu_writer_nan_sentinel_1599.py`) belongs to the GPU cluster.

## Constraints

- Tests-only.
- Markers from `_helpers/markers.py`; `make_minimal_tiff` from `_helpers/tiff_builders.py`.
- Descriptive IDs (e.g. `id="alias[GDAL_NODATA->nodata]"`, `id="lifecycle[masked-then-write]"`).
- `CLUSTER_AUDIT_NODATA.md` deleted before merge.

## Verification

```bash
pytest xrspatial/geotiff/tests/read/test_nodata.py xrspatial/geotiff/tests/write/ -v
pytest xrspatial/geotiff/tests/ -x -q
```

File count drops by 12 (13 deleted, 1 added if write-side becomes new file).

Issue #2424, opened 2026-05-26
Epic: Consolidate the GeoTIFF test long tail (follow-up to #2390)


brendancol:
# Epic: Consolidate the GeoTIFF test long tail

Follow-up to #2390. That epic reduced the GeoTIFF test suite from 354 files to 254 files across 11 PRs. The new directory structure (`_helpers/`, `attrs/`, `integration/`, `parity/`, `read/`, `release_gates/`, `unit/`, `vrt/`, `write/`) is in place. The remaining 225 top-level files are the long tail: per-issue regression tests that did not fit cleanly into any of the clusters the original epic carved out.

This issue inventories what is left, groups files by natural concern, and proposes a second consolidation pass.

## What is left

```
xrspatial/geotiff/tests/
├── conftest.py
├── _helpers/                   (3 files)
├── attrs/                      (1)
├── integration/                (3)
├── parity/                     (2)
├── read/                       (9)
├── release_gates/              (1)
├── unit/                       (4)
├── vrt/                        (5)
├── write/                      (4)
├── golden_corpus/              (4 + helpers, intentionally left structured)
└── 225 top-level test_*.py     ← the long tail
```

Total: 254 test files (excluding `golden_corpus/` internals).

## Inventory by candidate cluster

Files grouped by what they actually exercise. Counts are the raw file counts; each row would collapse to one or two consolidated modules.

| Cluster | Files | Proposed home |
|---|---|---|
| GPU | 43 | new `gpu/` directory split by concern (reader / writer / codec / kernel-kwargs / orientation) |
| Nodata semantics | 13 | extend `read/test_nodata.py` or new `nodata/test_contract.py` |
| Georef / rotation | 12 | extend `read/test_crs.py` plus new `read/test_georef.py` |
| VRT tail | 12 | fold into existing `vrt/test_*.py` files |
| Remote / sidecar | 11 | extend `integration/test_http_sources.py` plus new `integration/test_sidecar.py` |
| Golden-corpus variants | 11 | move under `golden_corpus/` (they share the corpus oracle anyway) |
| Codec (CPU) | 11 | extend `unit/test_compression.py` |
| Overview | 10 | extend `write/test_overview.py`; reader-side overview into `read/test_overview.py` |
| Input validation | 9 | new `unit/test_input_validation.py` |
| Kwarg / signature | 9 | new `unit/test_signatures.py` |
| Writer tail | 9 | fold into `write/test_basic.py` and new `write/test_streaming.py` |
[...]

PR #2470, opened 2026-05-27
geotiff tests: consolidate nodata write-side cluster (Sub-PR B)

brendancol:
Closes part of #2434 (Sub-PR B). Cluster 10 of long-tail epic #2424.

## Summary

Fold 5 write-side nodata files into a new `xrspatial/geotiff/tests/write/test_nodata.py`:

- `test_nodata_validation_1973.py` (non-numeric / bool rejection at every writer entry point)
- `test_nodata_bool_rejection_1911.py` (bool rejection at `to_geotiff` + `build_geo_tags`)
- `test_nodata_int64_precision_1847.py` (full-precision 64-bit integer sentinels via `_parse_nodata_str`; eager / dask / VRT / GPU round-trip)
- `test_nodata_out_of_range_1581.py` (out-of-range integer sentinel is a value-match no-op rather than `OverflowError`)
- `test_mask_nodata_kwarg_2052.py` (`mask_nodata=False` keeps the source integer dtype so the `dtype=` cast contract is reachable)

Tests-only. The shared `requires_gpu` marker from `_helpers/markers.py` replaces the per-file `_gpu_available` / `_gpu_only` helpers. 87 tests collected, matching the baseline sum.

The release-gate row for nodata round-trip and the geotiff.rst prose are updated to cite `write/test_nodata.py` alongside the existing read-side files.

## Verification

- `pytest xrspatial/geotiff/tests/write/test_nodata.py`: 87 passed.
- `pytest xrspatial/geotiff/tests/release_gates/test_stable_features.py::test_release_gate_cites_only_existing_test_files`: 1 passed.

## Test plan

- [x] New consolidated `write/test_nodata.py` passes on this checkout.
- [x] Release-gate checklist-parity test passes.

## Companion PR

The read-side companion (Sub-PR A) is #2469. The two are independent; merging in either order works.


brendancol (commented):
## PR Review: nodata write-side consolidation (Sub-PR B)

Tests-only. Five source files folded into a new `write/test_nodata.py`. Bodies preserved verbatim.

**Blockers:** none.

**Suggestions:** none.

**Nits:** the file already has `# ===` banners at the top of each of the five sections; the parametrised groups inside each section could carry one-line subheaders for skim-readability, but it is fine as-is.

**Verified locally**
- 87 tests collected and pass. Matches the baseline sum (27 + 17 + 25 + 8 + 10).
- No duplicate test IDs.
- One shared `requires_gpu` marker replaces five per-file cupy probes.
- Release-gate checklist-parity passes; the rst row adds `write/test_nodata.py` alongside the still-present read-side files.
- flake8 clean.

Independent of #2469 (the read-side companion). Either merge order works.

PR #2479, opened 2026-05-27
geotiff: make _FileSource a context manager (#2449)

### LINKED ISSUE #2449: geotiff: make _FileSource a context manager (cleanup follow-up to #2444)
## Reason or Problem

`xrspatial.geotiff._sources._FileSource` exposes `close()` but no `__enter__` / `__exit__`. Every call site has to wrap it in `try` / `finally` to guarantee mmap release. PR #2448 added another such call site in `_vrt_validation._check_mixed_source_crs`, and a code-review nit flagged the resulting `try` / `except` / `finally` / `continue` shape as harder to scan than a `with` block.

## Proposal

Add `__enter__` / `__exit__` to `_FileSource` so callers can write:

```python
with _FileSource(path) as src:
    data = src.read_all()
    ...
```

Then sweep the existing call sites (search for `_FileSource(` in `xrspatial/geotiff/`) and replace each `try` / `finally` with a `with` block. The mmap-cache release semantics stay the same; the helper change is mechanical.

## Drawbacks

Mechanical sweep across several files. Worth verifying that no call site relies on the handle outliving a function scope before swapping wholesale.

Surfaced during review of PR #2448. Not blocking that PR; tracked here so the cleanup is not forgotten.

brendancol:
## Summary

Adds `__enter__` / `__exit__` to `_FileSource` so callers can write `with _FileSource(path) as src:` instead of wrapping every open in a `try` / `finally`.

Sweeps the mechanical call sites in `_vrt_validation.py`, `_vrt.py`, `_backends/gpu.py`, and the four test modules that exercised the helper directly. The mmap-cache refcount semantics are unchanged; `__exit__` just calls the existing `close()`.

Two production sites are left as-is. `_reader._read_to_array` and `_backends/gpu._read_geotiff_gpu_eager` both wrap a several-hundred-line function body in a `try` / `finally` and mix `_FileSource` with `_CloudSource`. Converting those would mean either teaching `_CloudSource` the protocol or restructuring large blocks, which is outside the mechanical scope flagged in #2449.

## Test plan

- [x] New tests `xrspatial/geotiff/tests/test_file_source_context_2449.py` cover `__enter__` returning self, refcount release on normal exit, release on exception, double-close safety, and nested-`with` refcount accounting.
- [x] `pytest xrspatial/geotiff/tests/` (non-gpu / non-integration): 5096 passed, 65 skipped, 2 xfailed.
- [x] `pytest xrspatial/geotiff/tests/gpu/`: 384 passed, 3 skipped.
- [x] `pytest xrspatial/geotiff/tests/integration/`: 474 passed.

Closes #2449.

### DIFF
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -766,11 +766,8 @@ def read_geotiff_gpu(source: str, *,
 
             def _read_once():
                 if not _shared_data_cache:
-                    src2 = _FileSource(source)
-                    try:
+                    with _FileSource(source) as src2:
                         _shared_data_cache.append(src2.read_all())
-                    finally:
-                        src2.close()
                 return _shared_data_cache[0]
 
             band_arrays = []
@@ -880,15 +877,12 @@ def read_geotiff_gpu(source: str, *,
                     for i in range(len(offsets))
                 ]
             else:
-                src2 = _FileSource(source)
-                data2 = src2.read_all()
-                try:
+                with _FileSource(source) as src2:
+                    data2 = src2.read_all()
                     compressed_tiles = [
                         bytes(data2[offsets[i]:offsets[i] + byte_counts[i]])
                         for i in range(len(offsets))
                     ]
-                finally:
-                    src2.close()
 
         if arr_gpu is None:
             try:
@@ -1284,15 +1278,12 @@ def _decode_window_gpu_direct(file_path, all_offsets, all_byte_counts,
         # usable on the host. Open the file via mmap, slice out just the
         # bytes for these tiles, and run the GPU decoder on those.
         from .._reader import _FileSource
-        src = _FileSource(file_path)
-        try:
+        with _FileSource(file_path) as src:
             data = src.read_all()
             compressed_tiles = [
                 bytes(data[sub_offsets[i]:sub_offsets[i] + sub_byte_counts[i]])
                 for i in range(len(sub_offsets))
             ]
-        finally:
-            src.close()
         arr_gpu = gpu_decode_tiles(
             compressed_tiles, tw, th, sub_w, sub_h,
             compression, predictor, file_dtype, samples,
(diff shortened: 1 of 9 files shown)


brendancol (commented):
## PR Review: geotiff: make _FileSource a context manager (#2449)

### Blockers
None.

### Suggestions
None. The refactor is mechanical and the behaviour is preserved at every call site.

### Nits
- `xrspatial/geotiff/tests/test_file_source_context_2449.py:21` uses `__import__('os').path.realpath(path)` instead of a top-level `import os`. The inline import keeps the helper self-contained but is unusual; a top-level `import os` reads more naturally.
- `xrspatial/geotiff/tests/test_file_source_context_2449.py:50` has a `del src` line annotated "unreachable" after a call that always raises. Dead code; can drop the line and the trailing comment.
- `xrspatial/geotiff/_sources.py:332` returns `False` from `__exit__` explicitly. That matches the implicit `None` return; either is fine, but dropping the `return` keeps the body to a single line.

### What looks good
- `__enter__` returns `self`, `__exit__` calls the existing `close()`. Idiomatic, and the mmap-cache refcount semantics are preserved.
- `_vrt_validation.py:559` flattens the nested `try` / `except` / `finally` into `try` / `with` / `except`. The outer `except` catches OSError both from the constructor (matching the old inner try) and from the read/parse calls, so the observable behaviour is the same with fewer indentation levels.
- `_vrt.py:1958` is a latent fix in addition to the cleanup. The old loop only called `src.close()` on the success path, so a `parse_header` exception would leak a cache refcount. The `with` block closes on both paths.
- New tests cover `__enter__` identity, refcount release on normal exit, refcount release on exception, double-close safety, and nested-`with` refcount accounting.
- The PR body is explicit about the two intentionally-skipped production sites (`_reader._read_to_array`, `_backends/gpu._read_geotiff_gpu_eager`). Both wrap several hundred lines and mix `_FileSource` with `_CloudSource`, so converting them is outside the mechanical scope #2449 was scoped to.

### Checklist
- [x] Algorithm matches reference/paper: N/A (helper refactor)
- [x] All implemented backends produce consistent results: no backend-specific behaviour changes
- [x] NaN handling is correct: N/A
- [x] Edge cases are covered by tests: exception path, double-close, nesting all covered
- [x] Benchmark exists or is not needed: not needed
- [x] README feature matrix updated (if applicable): not applicable
- [x] Docstrings present and accurate: `__enter__` / `__exit__` are self-evident protocol methods; the class docstring already covers the source

brendancol (commented):
## Follow-up review: nits addressed

Original nits, all fixed in a44d5326:

- [x] `_sources.py` `__exit__` no longer returns `False`; falls through to implicit `None`.
- [x] `test_file_source_context_2449.py` now uses a top-level `import os` instead of `__import__('os')`.
- [x] `test_exit_releases_on_exception` drops the unreachable `del src` line and the trailing `# always raises` comment.

Re-ran `pytest xrspatial/geotiff/tests/test_file_source_context_2449.py`: 5 passed.

No further findings.

PR #2488, opened 2026-05-27
geotiff tests: gate TestSparseTilesGPU on usable-CUDA marker (#2487)

### LINKED ISSUE #2487: GPU test gated on _HAS_CUPY instead of usable-CUDA capability marker (test_ifd.py)
**Describe the bug**

`TestSparseTilesGPU` at `xrspatial/geotiff/tests/unit/test_ifd.py:857` gates on `_HAS_CUPY`, which only checks whether the cupy module imports. On a host where cupy imports but CUDA is unusable (insufficient driver, no device), `test_sparse_tile_gpu_round_trip` runs and fails instead of skipping.

The repo already has the right probe at `xrspatial/geotiff/tests/_helpers/markers.py:17`: `requires_gpu` / `gpu_available()` checks both the import and `cupy.cuda.is_available()`. Other GPU tests in this suite use a local `_gpu_available()` with the same shape.

**Expected behavior**

On a host with cupy installed but no working CUDA device, `TestSparseTilesGPU::test_sparse_tile_gpu_round_trip` should skip with a "cupy + CUDA required" reason, like the rest of the GPU-gated tests.

**Fix**

Swap the decorator at `test_ifd.py:858` from `@pytest.mark.skipif(not _HAS_CUPY, ...)` to `@requires_gpu` imported from `xrspatial.geotiff.tests._helpers.markers`. The rest of the geotiff tests already use a `cupy.cuda.is_available()` probe, so the sweep is one line.

**Additional context**

`_HAS_CUPY` stays defined in `test_ifd.py` for any non-GPU rasterio fixture branches that reference it; the swap only touches the GPU-class decorator.

brendancol:
## Summary

- Swap the gate on `TestSparseTilesGPU` in `xrspatial/geotiff/tests/unit/test_ifd.py` from a bare `_HAS_CUPY = importlib.util.find_spec("cupy") is not None` probe to `@requires_gpu` from `xrspatial.geotiff.tests._helpers.markers`. The shared marker also probes `cupy.cuda.is_available()`, so on hosts where cupy imports but the CUDA runtime is unusable, the test now skips with reason `"cupy + CUDA required"` instead of failing at device-call time.
- Swept the rest of `xrspatial/geotiff/tests/` for the same defect. Every other GPU-gated test in the suite already uses a local `_gpu_available()` helper that probes the same way, so this is a one-file fix. The sweep is intentionally tight; only obviously-wrong gates were in scope.
- Add a regression meta-test, `test_sparse_tiles_gpu_uses_capability_marker_2487`, that pins the gate so a future "simplification" back to an import-only probe trips the suite.

## Backend coverage

Test plumbing only. No spatial-op behaviour changes. Numpy / cupy / dask+numpy / dask+cupy paths are unaffected.

## Test plan

- [x] `TestSparseTilesGPU::test_sparse_tile_gpu_round_trip` still passes on a host with a working CUDA device.
- [x] On a simulated no-CUDA host (in-process monkey-patch of `_helpers.markers.requires_gpu` to `skipif(True, reason="cupy + CUDA required")`), the test skips with that reason instead of failing inside the GPU pipeline.
- [x] New meta-test `test_sparse_tiles_gpu_uses_capability_marker_2487` passes and would fail if the gate is regressed to a bare import probe.

Closes #2487

### DIFF
--- a/xrspatial/geotiff/tests/unit/test_ifd.py
+++ b/xrspatial/geotiff/tests/unit/test_ifd.py
@@ -60,10 +60,16 @@ from xrspatial.geotiff._header import (
 )
 from xrspatial.geotiff._reader import read_to_array
 
-from .._helpers.markers import requires_loopback
+from .._helpers.markers import requires_gpu, requires_loopback
 
 _HAS_RASTERIO = importlib.util.find_spec("rasterio") is not None
-_HAS_CUPY = importlib.util.find_spec("cupy") is not None
+# Note: ``_HAS_CUPY`` (bare import probe) is intentionally NOT used to
+# gate GPU tests below. A bare ``import cupy`` succeeds on hosts where
+# the CUDA runtime is missing or unusable, which made
+# ``TestSparseTilesGPU`` fail at device-call time instead of skipping.
+# Use ``requires_gpu`` from ``_helpers.markers`` (which also probes
+# ``cupy.cuda.is_available()``) for any test that needs a working GPU
+# device. See issue #2487.
 
 requires_rasterio = pytest.mark.skipif(
     not _HAS_RASTERIO, reason="rasterio required to write sparse fixtures"
@@ -855,7 +861,7 @@ class TestSparseStrips:
 
 
 @requires_rasterio
-@pytest.mark.skipif(not _HAS_CUPY, reason="cupy required")
+@requires_gpu
 class TestSparseTilesGPU:
 
     def test_sparse_tile_gpu_round_trip(self, tmp_path):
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: geotiff tests: gate TestSparseTilesGPU on usable-CUDA marker (#2487)

None.

None.

- [ ] Commit body has a typo: "A swept the rest" should be "Swept the rest" (line 11). Not worth amending on its own.

### What looks good
- Diff is surgical. One decorator swap, one removed module global (replaced with a rationale comment), one pinned regression meta-test. Scope matches the issue.
- `requires_gpu` from `_helpers.markers` is the right gate. It probes both the cupy import and `cupy.cuda.is_available()`, which is what hosts with broken CUDA runtimes need.
- The meta-test compares against `_markers_mod.requires_gpu.mark` rather than the `MarkDecorator` wrapper. Pytest stores the underlying `Mark` in `pytestmark`, not the decorator, so checking the wrapper directly would fail. The inline comment explains this.
- The "no other geotiff test has this defect" claim in the PR body checks out. I spot-checked `test_bytesio_source.py`, `test_strict_mode_1662.py`, `test_predictor.py`, `test_codec_roundtrip.py`, and `test_invalid_int_nodata_rejection_2441.py` -- each rolls its own `_gpu_available()` that calls `cupy.cuda.is_available()`. A one-file fix is enough.
- Reason string `"cupy + CUDA required"` matches what `markers.requires_gpu` emits, so the meta-test's `any('cupy + CUDA required' in r ...)` assertion is coupled to the marker's actual output.
- `_HAS_CUPY` global is gone. The two remaining hits in the file are a rationale comment and a docstring, not code.

### Checklist
- [x] Algorithm matches reference/paper -- N/A (test gate)
- [x] All implemented backends produce consistent results -- N/A (test gate)
- [x] NaN handling is correct -- N/A (test gate)
- [x] Edge cases are covered -- the meta-test covers both directions (skip vs. run) via the marker's own reason string and identity
- [x] Dask chunk boundaries handled correctly -- N/A (test gate)
- [x] No premature materialization or unnecessary copies -- N/A (test gate)
- [x] Docstrings present and accurate -- meta-test docstring explains the regression it pins

PR #2489, opened 2026-05-27
geotiff: surface sidecar parse error on explicit overview_level>=1

### LINKED ISSUE #2484: geotiff: explicit overview_level>=1 against corrupt sidecar hides the parse error
## Summary

When a caller passes `overview_level=1` against a corrupt `.tif.ovr` sidecar, the reader does not surface the parse error. Both eager-read paths catch the sidecar exception and drop it, so the caller gets a generic "out of range" `ValueError` from `select_overview_ifd` instead of the real failure.

The release contract for `reader.sidecar_ovr` at `docs/source/reference/geotiff_release_contract.md:92` says:

> Requesting a specific external level surfaces the underlying parse error.

The warning emitted on the silent-fallback path also tells the user to pass `overview_level>=1` to see the parse error. Neither promise holds today.

## Where

- `xrspatial/geotiff/_reader.py:243` (eager CPU read path)
- `xrspatial/geotiff/__init__.py:302` (`open_geotiff` metadata path)
- `xrspatial/geotiff/tests/integration/test_sidecar.py:1045` (test pins the broken behaviour)

## Expected vs actual

Expected: `open_geotiff(path, overview_level=1)` against a corrupt sidecar raises with the sidecar parse failure visible (as the raised exception, or chained via `__cause__`).

Actual: `ValueError("overview_level=1 out of range")`. The real cause is gone.

## Proposal

For `overview_level >= 1` (an explicit external/sidecar request), do not swallow the sidecar parse error. Re-raise it, or chain it as `__cause__` of a clearer outer error so the traceback shows what actually went wrong.

`overview_level=None` and `overview_level=0` keep the current silent-fallback behaviour, which the contract documents.

## Acceptance

- Eager CPU and metadata paths re-raise (or chain) the sidecar parse exception when `overview_level >= 1`.
- `None` / `0` still warn and fall back silently.
- `CloudSizeLimitError` still propagates either way.
- Test at `test_sidecar.py:1045` updated to assert the new behaviour. Add a contrast test for the auto-overview fallback.

brendancol:
Closes #2484.

## Summary

- The release contract for `reader.sidecar_ovr` says "Requesting a specific external level surfaces the underlying parse error." The eager CPU, eager GPU, and metadata paths were silently swallowing the sidecar exception, so a caller asking for `overview_level=1` against a corrupt `.tif.ovr` got a misleading `ValueError("overview_level=1 out of range")` from `select_overview_ifd` instead of the real parse failure.
- For `overview_level >= 1`, chain the parse exception via `raise ... from exc` so the traceback shows what actually went wrong.
- `overview_level=None` / `0` keep the documented silent-fallback behaviour. `CloudSizeLimitError` still propagates either way.

## Backend coverage

- numpy (eager CPU): fixed in `_reader._read_to_array`.
- cupy (eager GPU): fixed in `_backends/gpu` for symmetry with the contract.
- metadata path used by the dask graph builder: fixed in `__init__._read_geo_info`.
- HTTP / fsspec dask path goes through `discover_remote_sidecar`, which has its own carve-out and is out of scope for this finding.

## Test plan

- [x] `test_open_geotiff_requesting_sidecar_level_surfaces_parse_error` (rewritten from the test that pinned the broken behaviour)
- [x] `test_open_geotiff_overview_level_two_surfaces_parse_error`
- [x] `test_open_geotiff_overview_level_none_silently_falls_back`
- [x] `test_open_geotiff_overview_level_zero_silently_falls_back`
- [x] `test_read_geo_info_requesting_sidecar_level_surfaces_parse_error`
- [x] `test_open_geotiff_gpu_requesting_sidecar_level_surfaces_parse_error` (gated on GPU)
- [x] All 88 tests in `xrspatial/geotiff/tests/integration/test_sidecar.py` pass locally.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -296,7 +296,8 @@ def _read_geo_info(source, *, overview_level: int | None = None,
         # behaviour with a warning. Mirrors the eager CPU path in
         # ``_reader._read_to_array`` and the dask metadata helper
         # ``_sidecar.discover_remote_sidecar``. Issue #2416.
-        from ._sidecar import attach_sidecar_origin, find_sidecar, load_sidecar
+        from ._sidecar import (attach_sidecar_origin, find_sidecar,
+                                handle_sidecar_parse_failure, load_sidecar)
         sidecar_origin: dict[int, tuple] = {}
         sidecar_path = find_sidecar(source)
         if sidecar_path is not None:
@@ -313,31 +314,14 @@ def _read_geo_info(source, *, overview_level: int | None = None,
                 # patch routes a cloud-source path through here.
                 raise
             except Exception as exc:
-                # Explicit ``overview_level >= 1`` is a request for the
-                # external sidecar surface. The release contract
-                # (``geotiff_release_contract.md`` row
-                # ``reader.sidecar_ovr``) and the fallback warning
-                # text below both promise that this case surfaces the
-                # underlying parse error rather than silently degrading
-                # to base-only and then raising a misleading "out of
-                # range" from ``select_overview_ifd``. Chain via
-                # ``raise ... from exc`` so the traceback carries the
-                # concrete parse failure. Mirrors the eager CPU path
-                # in ``_reader._read_to_array``. Issue #2484.
-                if overview_level is not None and overview_level >= 1:
-                    raise ValueError(
-                        f"Cannot read overview_level={overview_level} "
-                        f"from external sidecar {sidecar_path!r}: "
-                        f"sidecar parse failed with "
-                        f"{type(exc).__name__}: {exc}"
-                    ) from exc
-                warnings.warn(
-                    f"Ignoring unreadable sidecar {sidecar_path!r}: "
-                    f"{type(exc).__name__}: {exc}. Falling back to "
-                    f"base-file-only read. Delete the .ovr file or pass "
-                    f"overview_level>=1 to surface the parse error.",
-                    RuntimeWarning,
-                    stacklevel=3,
+                # Shared policy: surface the parse error when the
+                # caller asked for a level the base file alone cannot
+                # serve; warn and fall back otherwise. See
+                # ``_sidecar.handle_sidecar_parse_failure`` for the
+                # rationale. Issue #2484.
+                handle_sidecar_parse_failure(
+                    exc, sidecar_path, overview_level,
+                    base_ifd_count=len(ifds),
                 )
                 sidecar = None
             if sidecar is not None:
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -313,6 +313,24 @@ def _read_geo_info(source, *, overview_level: int | None = None,
                 # patch routes a cloud-source path through here.
                 raise
             except Exception as exc:
+                # Explicit ``overview_level >= 1`` is a request for the
+                # external sidecar surface. The release contract
+                # (``geotiff_release_contract.md`` row
+                # ``reader.sidecar_ovr``) and the fallback warning
+                # text below both promise that this case surfaces the
+                # underlying parse error rather than silently degrading
+                # to base-only and then raising a misleading "out of
+                # range" from ``select_overview_ifd``. Chain via
+                # ``raise ... from exc`` so the traceback carries the
+                # concrete parse failure. Mirrors the eager CPU path
+                # in ``_reader._read_to_array``. Issue #2484.
+                if overview_level is not None and overview_level >= 1:
+                    raise ValueError(
+                        f"Cannot read overview_level={overview_level} "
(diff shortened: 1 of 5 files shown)


brendancol (commented):
## PR Review: surface sidecar parse error on explicit overview_level>=1

### Blockers
None.

### Suggestions

- `xrspatial/geotiff/_reader.py:248-266` and the parallel sites in `__init__.py` and `_backends/gpu.py`: the new guard fires for any `overview_level >= 1`, but level 1 can also be a legitimate internal-overview IFD when the base TIFF carries its own pyramids. With internal overviews AND a corrupt sidecar, the caller's request is reachable from the base file alone (`select_overview_ifd` would resolve level 1 against the base IFD list, and the sidecar IFDs are appended after). The current fix raises anyway. GDAL's convention is pyramids inside (COG / internal overviews) OR in a sidecar, not both, so the collision is rare. Two ways to handle it: short comment acknowledging the trade-off, or gate the raise on "the requested level actually falls in the sidecar range" (`overview_level >= len(base_ifds)`). The latter matches what the warning text literally says.

- The three sites (eager CPU, `_read_geo_info`, eager GPU) now carry the same ~10-line raise block. Worth factoring into a `_handle_sidecar_parse_failure(exc, sidecar_path, overview_level)` helper in `_sidecar.py` so future contract changes touch one place. The near-identical comments are a hint that this should live somewhere shared.

### Nits

- The outer `ValueError` message embeds `f"{type(exc).__name__}: {exc}"` AND chains the same exception via `__cause__`. Python's traceback formatting already prints chained causes; the in-message embed is belt-and-braces but does make `str(exc)` verbose. Either keep both (users who print only `str(exc)` get the full story) or drop the embed and lean on the chain. Both defensible -- flagging because the test asserts both, which locks the choice in.

- The `level=2` test comment could mention that any level outside the base IFD count triggers the sidecar load attempt and therefore the new branch. Right now the test reads as "arbitrary level" when really it's "any level the base alone cannot serve."

### What looks good

- Fix is surgical. Silent-fallback for `None` / `0` is unchanged, matching the contract.
- `CloudSizeLimitError` still re-raises ahead of the new branch in all three sites, so the byte-budget contract is preserved.
- Tests pin both halves of the contract (raise on explicit level, fall back silently on `None`/`0`), plus metadata-only and GPU symmetry.
- `raise ... from exc` is the right idiom here.
- Comments in each new branch reference the contract file and issue number.

### Checklist

- [x] Algorithm matches reference: contract at `geotiff_release_contract.md:92` matches.
- [x] All implemented paths covered (HTTP via `discover_remote_sidecar` is out of scope per the original finding).
- [x] NaN handling: n/a.
- [x] Edge cases: `0`, `None`, `>=1` all covered.
- [x] Dask: n/a.
- [x] Benchmark: not needed.
- [x] README feature matrix: not needed.
- [x] Docstrings: signatures unchanged.

brendancol (commented):
## Follow-up review after refactor commit

### Disposition of the original review

- **Suggestion 1 (internal-overview collision at level 1):** Fixed. The guard now gates on `overview_level >= base_ifd_count` rather than `>= 1`, so a base TIFF with internal overviews satisfies the request without the sidecar even when a sibling `.ovr` is corrupt. New test `test_open_geotiff_internal_overview_level_silent_on_bad_sidecar` pins it: level 1 warns and reads the internal overview, level 2 still surfaces the sidecar parse error because that level is reachable only via the sidecar.
- **Suggestion 2 (factor into shared helper):** Fixed. Added `_sidecar.handle_sidecar_parse_failure(exc, sidecar_path, overview_level, base_ifd_count)` and updated the three callsites (eager CPU, `_read_geo_info`, eager GPU) to dispatch through it. Policy and rationale live in the helper docstring; callers carry only a short pointer comment.
- **Nit 1 (message duplication of `__cause__`):** Dismissed. The duplication is intentional: callers that print only `str(exc)` (logs, simple error handlers) still see the underlying parse type/message; callers that walk `__cause__` see the chained exception. The test asserts both, which is the contract I want to lock in.
- **Nit 2 (level=2 test comment):** Fixed. The comment now spells out that the base IFD chain has length 1, so any level > 0 is sidecar-only in that fixture.

### Quick read of the refactor

- The helper signature passes `base_ifd_count` rather than the IFD list so the helper does not need to know anything about IFD shape. Keeps the coupling narrow.
- All three callsites read `base_ifd_count=len(ifds)` at the point where the base IFDs have just been parsed but the sidecar has not yet been merged. That is the correct count for the gate.
- `CloudSizeLimitError` is still caught and re-raised separately in each callsite; the helper only sees the swallowable-by-default exceptions, which is the right separation of concerns.

### Tests

- 89 tests in `xrspatial/geotiff/tests/integration/test_sidecar.py` pass locally.
- 480 tests in `xrspatial/geotiff/tests/integration/` pass.
- 902 tests in `xrspatial/geotiff/tests/read/` pass.

No new findings.

PR #2490, opened 2026-05-27
geotiff: reject duplicate IFD tags with DuplicateIFDTagError (#2483)

### LINKED ISSUE #2483: GeoTIFF parser silently accepts malformed TIFFs with duplicate IFD tags
**Describe the bug**

The TIFF/BigTIFF IFD parser in `xrspatial/geotiff/_header.py` silently accepts files with duplicate IFD tags. Both entry loops write into the same `entries` dict via `entries[tag] = ...`, so the last duplicate wins. The dimension pre-scan (`dims[tag] = ...`) has the same behavior.

Reproduction: an IFD with two conflicting `ImageWidth` entries (tag 256), values 4 and 999, parses without error. `parse_ifd()` returns width 999.

The same last-value-wins behavior applies to every tag, including ones that control the read contract: ImageLength, BitsPerSample, Compression, StripOffsets / StripByteCounts, TileOffsets / TileByteCounts, GDAL_NODATA, and the Geo* tags. At the stable read boundary a malformed or adversarial file can change interpretation silently instead of failing closed.

Affected lines:

- `xrspatial/geotiff/_header.py:645` (`entries = {}` -- initial dict)
- `xrspatial/geotiff/_header.py:746` (`entries[tag] = IFDEntry(...)`)
- `xrspatial/geotiff/_header.py:653` (`dims: dict[int, Any] = {}` -- pre-scan)

**Expected behavior**

Reject duplicate IFD tags at parse time with a typed exception that names the duplicated tag and the offsets of the conflicting entries. Apply the same check in the dimension pre-scan so the early-exit code path can't silently accept a malformed dimension tag. The failure should surface at the public read boundary.

**Additional context**

TIFF 6.0 section 2 ("The TIFF Structure") says IFD entries "must be sorted in ascending order by Tag" with no duplicates. Accepting duplicates is a spec violation; silent acceptance is the worst failure mode.

Found by a security sweep of the GeoTIFF parser.

brendancol:
Closes #2483.

## Summary

- The TIFF/BigTIFF IFD parser stored entries in a dict keyed by tag id and let the last write win. A file with two ``ImageWidth`` (or any other tag) entries silently parsed to whichever value came second. The same ambiguity applied to every tag, including ones that control the read contract: ``ImageLength``, ``BitsPerSample``, ``Compression``, ``StripOffsets`` / ``StripByteCounts``, ``TileOffsets`` / ``TileByteCounts``, ``GDAL_NODATA``, and the Geo* keys.
- New typed exception ``DuplicateIFDTagError`` (a ``ValueError`` subclass) fires at parse time and names the duplicated tag id plus the byte offsets of the two conflicting entries. Re-exported from ``xrspatial.geotiff``.
- The check runs in the dimension pre-scan (which already walks the full entry table), so the early-exit code path that bounds pixel-array sizes cannot bypass it either.

## Backend coverage

Parse-time fix in the core TIFF header parser. The change is dispatch-free; every backend that reads through ``open_geotiff`` (numpy, cupy, dask+numpy, dask+cupy) inherits the fail-closed behavior automatically.

## Test plan

- [x] ``xrspatial/geotiff/tests/unit/test_ifd.py`` Section 2b adds 8 new tests, including the exact reproduction case from #2483 (two ``ImageWidth`` entries with values 4 and 999) and parametrised coverage across ``ImageWidth``, ``ImageLength``, ``BitsPerSample``, and ``GeoKeyDirectory``.
- [x] BigTIFF entry shape (20-byte entries) covered.
- [x] Error message asserts the prior and duplicate byte offsets surface in the text.
- [x] ``DuplicateIFDTagError`` subclasses ``ValueError`` so existing ``except ValueError`` callers keep catching the case.
- [x] ``open_geotiff`` surfaces the error at the public read boundary.
- [x] Regression: a normal IFD with many distinct tag ids still parses cleanly.
- [x] Full ``xrspatial/geotiff/tests/unit/`` suite: 831 passed, 1 skipped.

## Docs

- Added a row for ``DuplicateIFDTagError`` to the "Failure-closed contract" table in ``docs/source/user_guide/geotiff_safe_io.rst``.
- Added a row to the "Known unsupported combinations" table in ``docs/source/reference/geotiff.rst`` so the doc and the regression test stay linked.


brendancol (commented):
## PR Review: geotiff: reject duplicate IFD tags with DuplicateIFDTagError (#2483)

None.

- [ ] `xrspatial/geotiff/_errors.py:282` -- `DuplicateIFDTagError` is appended to `__all__` after `UnsupportedGeoTIFFFeatureError`, breaking the otherwise roughly-alphabetical ordering. Move it up to sort between `ConflictingNodataError` and `GeoTIFFAmbiguousMetadataError`. The `__init__.py` `__all__` is already alphabetised, which makes the inconsistency in `_errors.py` more visible.

- [ ] `xrspatial/geotiff/_header.py:660-685` -- the duplicate-detection block runs after the BigTIFF / classic-TIFF `if/else` that reads tag/type/count. Moving the check up to right after `tag` is read would make the failure path tighter: today we read `type_id`, `count`, and `value_area_offset` only to ignore them on a duplicate. Tiny; mostly a readability nit.

- [ ] `xrspatial/geotiff/tests/unit/test_ifd.py:561-575` -- the parametrise `id`s use a `dup-<name>` prefix while the rest of the test file uses unprefixed ids (`chain_cap[over-limit-le]`, etc.). Inconsistent, not wrong.

- [ ] `xrspatial/geotiff/_errors.py:226-249` -- the docstring does not name the public read entry (`open_geotiff`). The other typed errors in this module do (e.g. `RotatedTransformError` mentions "the read entry points"). Worth a sentence.

### What looks good

- The reproduction case from #2483 (two `ImageWidth` entries, values 4 and 999) is locked by `test_duplicate_image_width_rejected_reproduction_2483`.
- BigTIFF gets its own test rather than relying on the classic-TIFF case transferring.
- The error message names both byte offsets and there is a test that asserts on the exact offsets.
- `DuplicateIFDTagError` subclasses `ValueError` (with a test that pins it), so existing `except ValueError` callers still catch the case.
- A public-boundary test goes through `open_geotiff` so the error surfaces at the user-facing entry, not only via `parse_ifd`.
- The negative-case row in `geotiff.rst` is paired with the test file path, matching the "row + locking test" pattern the rest of that table follows.
- Pre-scan placement is correct: the check fires before any value bytes are read, so a duplicate tag fails before the parser would otherwise allocate or follow a pointer.

### Checklist

- [x] Algorithm matches the cited spec (TIFF 6.0 section 2, no duplicate tag ids per IFD).
- [x] All backends inherit the fix (dispatch-free parse-time check; numpy / cupy / dask+numpy / dask+cupy all read through `parse_ifd`).
- [x] NaN handling not relevant (parser-level, not pixel math).
- [x] Edge cases covered (classic TIFF, BigTIFF, four safety-relevant tag ids, error message shape, public-boundary surfacing).
- [x] No premature materialisation; O(num_entries) dict lookup.
- [x] Benchmark not needed (parse-time check on a path that is already O(num_entries)).
- [x] README feature matrix not affected (no new function, no backend support change).

brendancol (commented):
## Follow-up review pass

All findings from the first review pass were applied in 5ed809a7:

- **Suggestion 1** (`_errors.py:__all__` ordering): Fixed. The list is now alphabetised, so `DuplicateIFDTagError` sorts between `ConflictingNodataError` and `GeoTIFFAmbiguousMetadataError`, matching `__init__.__all__`.
- **Nit 1** (move duplicate check up): Fixed. The tag id is a 2-byte unsigned at byte offset 0 of every entry in both classic TIFF and BigTIFF, so the check fires before the type / count / value_area_offset reads on the rejection path.
- **Nit 2** (parametrise id prefix): Fixed. Renamed to `duplicate_tag[<variant>]` to match `chain_cap[<variant>]` and `entry_table[<variant>]` elsewhere in `test_ifd.py`.
- **Nit 3** (docstring naming `open_geotiff`): Fixed. The `DuplicateIFDTagError` docstring now points at the public read entry points (eager, chunked, GPU, VRT), matching the wording used by `RotatedTransformError`.

No new findings. The 64 tests in `test_ifd.py` (including the 11 new duplicate-tag tests) still pass.

PR #2491, opened 2026-05-27
geotiff: enforce mixed-band dtype rejection in VRT reader

### LINKED ISSUE #2485: VRT reader silently widens mixed-band dtypes instead of raising
**Describe the bug**

`_backends/vrt.py:158` documents the VRT contract: sources must agree on dtype, and mismatches raise a typed error instead of being flattened. The code does not match the docs. `_effective_dtype_for_bands` in `_vrt.py:969` calls `np.result_type(*effective_dtypes)` on the band dtypes and returns the widened common dtype. A VRT with a `UInt16` band and a `Float32` band reads into a `Float32` buffer with no error.

A non-strict `xfail` at `tests/vrt/test_validation.py:1039` hides this. Remove the marker and the read returns a DataArray where it should raise.

The fallout is silent dtype changes for downstream code: categorical band values get widened, integer extremes lose precision, and nodata sentinel handling no longer matches the declared dtype.

**Expected behavior**

When two `<VRTRasterBand>` elements declare different `dataType` values (and no `ScaleRatio` / `ScaleOffset` widening is in play), `_effective_dtype_for_bands` should raise `MixedBandMetadataError` naming the conflicting bands and their dtypes. No `np.result_type` papering-over.

**Reproduction**

`test_mixed_source_dtype_ambiguous_widening_rejected` at `tests/vrt/test_validation.py:1039` already reproduces it. Drop the `@pytest.mark.xfail` and the test fails.

**Fix**

- In `_effective_dtype_for_bands`, check the collected `effective_dtypes` for disagreement before computing a result dtype. On mismatch, raise `MixedBandMetadataError` with each band's index and dtype.
- Skip the `np.result_type` call on the mismatch path.
- Convert the xfail into a real passing test and add a second dtype combo (e.g. Int32 + Float64) so the rule covers more than the original case.

brendancol:
## Summary

- Tightens `_effective_dtype_for_bands` in `xrspatial/geotiff/_vrt.py` to honour the contract at `_backends/vrt.py:158`: per-band declared dtype mismatch raises `MixedBandMetadataError` instead of silently widening through `np.result_type`.
- ComplexSource `ScaleRatio` / `ScaleOffset` widening still works when every band declares the same `dataType` (matching-dtype intra-band widening stays supported).
- Closes #2485.

## Backend coverage

Pure parse-time / reader logic, no kernel changes. Behaviour applies on every read path (numpy, dask).

## Test plan

- `xrspatial/geotiff/tests/vrt/test_validation.py::test_mixed_source_dtype_ambiguous_widening_rejected` is now a strict, parametrised test (`UInt16+Float32`, `Int32+Float64`) -- passes.
- `test_dtype_conversion.py::test_mixed_byte_and_float32_bands_raise` and `test_mixed_byte_and_int16_bands_raise` flipped from widening-expectation to raise-expectation -- pass.
- ScaleRatio / ScaleOffset tests (same declared dtype, intra-band widening) still pass.
- Full geotiff suite: 5956 passed, 68 skipped, 1 xfailed.

### DIFF
--- a/xrspatial/geotiff/_backends/vrt.py
+++ b/xrspatial/geotiff/_backends/vrt.py
@@ -1018,7 +1018,8 @@ def _read_vrt_chunked(source, *, window, band, name, chunks, gpu, dtype,
     #     source integer dtype (handled by the ``promotes is False``
     #     fall-through below).
     # See also Copilot review on PR #1822.
-    declared_dtype = _effective_dtype_for_bands(selected_bands)
+    declared_dtype = _effective_dtype_for_bands(
+        selected_bands, source=source)
 
     if mask_nodata and declared_dtype.kind in ('u', 'i'):
         promotes = False
--- a/xrspatial/geotiff/_vrt.py
+++ b/xrspatial/geotiff/_vrt.py
@@ -966,7 +966,7 @@ def _sentinel_for_dtype(nodata_val, dtype):
     return dtype.type(int(nodata_f))
 
 
-def _effective_dtype_for_bands(selected_bands) -> np.dtype:
+def _effective_dtype_for_bands(selected_bands, *, source=None) -> np.dtype:
     """Return the output buffer dtype shared by every selected band.
 
     Each band's effective dtype is the declared ``<VRTRasterBand>``
@@ -990,6 +990,16 @@ def _effective_dtype_for_bands(selected_bands) -> np.dtype:
     truncation. When every selected band declares the same ``dataType``,
     a ``ComplexSource``-driven widening on any band promotes the shared
     output buffer to ``float64``.
+
+    Parameters
+    ----------
+    selected_bands
+        The ``VRTRasterBand`` entries to validate.
+    source
+        Optional path / identifier of the VRT being read. When provided,
+        the ``MixedBandMetadataError`` message includes it so the caller
+        can locate the file from the traceback without re-parsing the
+        VRT XML.
     """
     if not selected_bands:
         raise ValueError(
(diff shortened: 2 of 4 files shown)


brendancol (commented):
## PR Review: geotiff: enforce mixed-band dtype rejection in VRT reader

None.

- [ ] `xrspatial/geotiff/tests/vrt/test_validation.py:1039` -- the new parametrised mixed-dtype test only exercises the eager `read_vrt` path. The chunked dask path (`read_vrt(..., chunks=...)`) also calls `_effective_dtype_for_bands` via `_backends/vrt.py:1021`, so it should raise too, but there is no test that pins that behaviour. Add a `chunks=2` variant so a future change that routes around `_effective_dtype_for_bands` in the chunked path cannot regress this silently.

- [ ] `xrspatial/geotiff/_vrt.py:1006` -- the message says `"VRT bands declare conflicting dtypes; expected a single declared dataType..."`. It does not name the source VRT file. From a deeply nested pipeline the user has to grep the VRT path out of the surrounding traceback. Other parse-time errors in this module include the VRT path in the message; the same trick here would be nicer to debug.

- [ ] `xrspatial/geotiff/tests/vrt/test_validation.py:1039` -- the parametrise `ids` default renders as `[uint16-float32-UInt16-Float32]`, so each name appears twice in the test id. Drop the numpy dtype params and derive them from the dataType string, or pass `ids=` for cleaner test names like `UInt16+Float32`.

### What looks good

- The split between declared-dtype check (raise) and per-band `ComplexSource` widening (still allowed) is the right call. `test_complex_source_scale_promotes_buffer_to_float` still passes, confirming the `ScaleRatio` use case survives.
- The error message names both band indices and dtypes, matching the issue request.
- Repurposing the two existing widening tests as raise tests (rather than deleting them) keeps the documented behaviour visible at the same call sites. `git blame` will surface the contract flip cleanly.
- 5956 geotiff tests still pass.

### Checklist
- [x] Algorithm matches contract at `_backends/vrt.py:158`
- [x] Eager and chunked paths both call `_effective_dtype_for_bands`
- [x] NaN handling unchanged
- [x] Edge cases covered (UInt16+Float32, Int32+Float64, ScaleRatio on matching dtypes still widens)
- [N/A] Dask chunk boundaries -- not relevant to this parse-time check
- [N/A] Benchmark -- parse-time validation, not a hot path
- [N/A] README feature matrix -- bug fix, no new function
- [x] Docstrings updated to describe the new behaviour

brendancol (commented):
## Follow-up review after commit 4e45ec6e

All three findings from the previous review are addressed in this commit.

### Findings disposition

- [x] **Suggestion**: chunked-path coverage. `test_mixed_source_dtype_ambiguous_widening_rejected` is now parametrised over `chunks=None` (eager) and `chunks=2` (dask). Both paths raise `MixedBandMetadataError`. Four test ids: `[eager-UInt16-Float32]`, `[eager-Int32-Float64]`, `[chunked-UInt16-Float32]`, `[chunked-Int32-Float64]`.
- [x] **Nit**: source-path in error. `_effective_dtype_for_bands` accepts a `source=` keyword and the message now reads `VRT '<path>': VRT bands declare conflicting dtypes; ...`. Both call sites (`_vrt.py` eager and `_backends/vrt.py` chunked) thread the source through. A new assertion in the test pins the VRT path in the exception text.
- [x] **Nit**: cleaner test ids. The numpy dtype params dropped; a small `_VRT_TO_NP_DTYPE` lookup table maps the dataType strings to numpy dtypes. Ids fall out as the clean `UInt16-Float32` rather than the duplicated `uint16-float32-UInt16-Float32`.

### Verification

`pytest xrspatial/geotiff/` reports 5958 passed, 68 skipped, 1 xfailed (the unrelated long-standing xfail).

PR #2492, opened 2026-05-27
geotiff: close cached mmap handles at interpreter shutdown (#2486)

### LINKED ISSUE #2486: GeoTIFF: _MmapCache retains file handles through interpreter shutdown
**Describe the bug**

The module-level `_MmapCache` singleton in `xrspatial/geotiff/_sources.py` keeps mmap-backed file handles open until LRU eviction or a manual `clear()`. Nothing closes idle entries at interpreter shutdown, so processes (and pytest runs) finish with open file handles. A targeted sidecar test emits `ResourceWarning: unclosed file .../base_2416.tif`; the full suite emits many of these on the geotiff path.

The retention is bounded by the cache size, not an unbounded leak. Production processes and test runners still should not exit with GeoTIFF file handles open.

**Expected behavior**

At interpreter shutdown, the mmap cache should close every entry it still holds. Idle entries always; non-zero-refcount entries too, since the process is going away. The cleanup must be idempotent and tolerate partial runtime teardown (already-closed file / mmap, missing modules during finalization).

**Reproduce**

```python
import warnings
from xrspatial.geotiff._sources import _FileSource

with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter('always')
    src = _FileSource('xrspatial/tests/data/base_2416.tif')
    src.close()  # returns the entry to the idle pool; does not close it.
    # Entry sits idle in the cache; on interpreter exit, no close runs.
```

**Fix sketch**

Register an `atexit` (or `weakref.finalize`) handler that closes both idle and in-use entries at interpreter shutdown. Handler must be idempotent and ordering-tolerant. `release()` semantics for live processes do not change; the only behavioural difference is that no handles leak past shutdown.

**Additional context**

- File: `xrspatial/geotiff/_sources.py:238` (release), `:288` (singleton)
- The `_http_pool` module-level urllib3 PoolManager has the same shape and is worth treating the same way.

brendancol:
Closes #2486.

## Summary

- Add an ``atexit`` hook that closes every entry in the module-level ``_MmapCache`` (idle and in-use) at interpreter shutdown, so GeoTIFF file handles do not leak past exit.
- The hook is idempotent and tolerates partial runtime teardown (already-closed mmap or file objects). ``release()`` semantics for live processes are unchanged.
- Also clears the module-level urllib3 ``_http_pool`` in the same hook so HTTP source sockets don't leak past exit.

## Backend coverage

Pure I/O path; no compute backends touched. ``_FileSource`` is the only consumer of ``_MmapCache`` and it is backend-agnostic.

## Test plan

- [x] New ``test_shutdown_cleanup_2486.py``: 8 cases covering idle / in-use entries, idempotency, late ``release()`` after ``close_all``, already-closed resources, and a subprocess that asserts no ``ResourceWarning`` is emitted at exit.
- [x] Full ``xrspatial/geotiff/tests/`` suite passes locally (5962 passed, 68 skipped, 2 xfailed).
- [x] ``test_file_source_context_2449.py`` and ``test_polish_1488.py`` cache tests still pass with the new shutdown hook in place.

### DIFF
--- a/xrspatial/geotiff/_sources.py
+++ b/xrspatial/geotiff/_sources.py
@@ -303,11 +303,15 @@ class _MmapCache:
         file objects are swallowed so a partially torn-down interpreter
         (where ``mmap.mmap.close`` or the file object's ``close`` raises
         because the underlying resource is already gone) does not crash
-        finalization. After this returns, ``_entries`` is empty and any
-        ``_FileSource`` that still holds an ``entry`` token will become a
-        no-op on its eventual ``release()`` -- the refcount decrement
-        still runs, but the close path is a no-op because the file
-        handle is already closed.
+        finalization. After this returns, ``_entries`` is empty.
+
+        If a stale ``_FileSource`` later calls ``release()`` on an entry
+        token that was torn down here, the release routes through the
+        orphan branch and calls ``_close_entry_locked`` a second time.
+        That second close is rare at interpreter shutdown (it requires a
+        live ``_FileSource`` to survive past ``atexit``), and it stays
+        safe because ``mmap.close()`` and the standard file object's
+        ``close()`` are both idempotent in CPython.
         """
         with self._lock:
             entries = list(self._entries.values())
@@ -375,8 +379,8 @@ def _shutdown_cleanup():
         pass
     # urllib3 PoolManager holds onto sockets until garbage collected.
     # Close it explicitly so the test runner's resource-warning filter
-    # does not catch leaked sockets at exit.
-    global _http_pool
+    # does not catch leaked sockets at exit. Read-only access of the
+    # module-level name, so no ``global`` is needed.
     pool = _http_pool
     if pool is not None:
         try:
--- a/xrspatial/geotiff/_sources.py
+++ b/xrspatial/geotiff/_sources.py
@@ -26,6 +26,7 @@ epic (issue #2228). Behaviour-preserving move; no public API change.
 """
 from __future__ import annotations
 
+import atexit
 import mmap
 import os as _os_module
 import threading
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: geotiff: close cached mmap handles at interpreter shutdown (#2486)

None.

- [ ] `xrspatial/geotiff/tests/test_shutdown_cleanup_2486.py:6` and `:12` — `import warnings` and `from xrspatial.geotiff import _sources as _src_module` are unused. Drop them.
- [ ] `xrspatial/geotiff/tests/test_shutdown_cleanup_2486.py:158-165` — `test_atexit_hook_is_registered` only checks that `_shutdown_cleanup` is callable; it does not verify the function is actually wired up with `atexit`. The neighbouring `test_subprocess_read_emits_no_resource_warning` is much closer to a real registration check. Either drop the weaker test or replace it with one that calls `atexit.unregister(_shutdown_cleanup)` and asserts the return is `True` (then re-registers).

- [ ] `xrspatial/geotiff/_sources.py:368` — `global _http_pool` is declared but `_http_pool` is only read, never reassigned. The `global` keyword is redundant here.
- [ ] `xrspatial/geotiff/_sources.py:312-336` — the docstring claim that the orphan branch calls `_close_entry_locked` again is correct, but only triggers if a stale `_FileSource` still holds the entry token after `close_all`. Worth one extra line noting that's a rare case at interpreter shutdown and that safety relies on mmap / file `close()` idempotency.

### What looks good
- The atexit hook is the right shape: idempotent, swallows per-entry errors, leaves `release()` semantics unchanged for live processes.
- `close_all` marks entries as orphaned before tearing them down so a late `release()` doesn't touch the cleared dict.
- Tests cover idle entries, in-use entries, idempotency, post-`close_all` release safety, already-closed resources, and a subprocess assertion against `ResourceWarning`.
- Cleanup of the `_http_pool` urllib3 PoolManager in the same hook stops HTTP socket handles leaking past exit.

### Checklist
- [x] Algorithm matches reference/paper — N/A (resource cleanup, not numeric)
- [x] All implemented backends produce consistent results — N/A (I/O layer)
- [x] Edge cases are covered by tests — yes (idle, in-use, idempotent, already-closed)
- [x] No premature materialization or unnecessary copies — yes
- [x] README feature matrix updated (if applicable) — not applicable (private I/O fix)
- [x] Docstrings present and accurate — yes

brendancol (commented):
## Follow-up review on commit 99d8dc5a

All four findings from the previous review have been addressed:

- [x] Unused imports (``warnings``, ``_sources as _src_module``) dropped from the test module.
- [x] ``test_atexit_hook_is_registered`` reworked to call ``atexit.unregister`` + re-register, instead of only asserting callability. The subprocess test remains the strong end-to-end check.
- [x] ``global _http_pool`` removed from ``_shutdown_cleanup`` (read-only access).
- [x] ``close_all`` docstring expanded with a note that late ``release()`` on an orphaned entry is rare at shutdown and stays safe because ``mmap.close()`` and file ``close()`` are idempotent in CPython.

13/13 cleanup-related tests pass locally. No new findings.

PR #2495, opened 2026-05-27
Make matplotlib an optional [plot] extra (#2494)

### LINKED ISSUE #2494: Move matplotlib to an optional [plot] extra to slim the default install
## Reason or Problem

A plain `pip install xarray-spatial` installs matplotlib and everything it
depends on: pillow, fonttools, kiwisolver, contourpy, cycler, pyparsing. None
of these are needed to run a spatial operation. matplotlib is listed in
`install_requires` in setup.cfg, but nothing in the package imports it at module
top level. Every use is a lazy import inside a plotting helper (`.xrs.plot` in
`accessor.py`, the plotting paths in `zonal.py`, and `bump.py`). So a user who
only wants the compute functions still pays for the whole plotting stack.

## Proposal

Move matplotlib out of `install_requires` and into an optional `plot` extra. It
then installs only when asked for, via `pip install xarray-spatial[plot]`.

**Design:**
- Remove `matplotlib` from `[options] install_requires` in setup.cfg.
- Add a `plot = matplotlib` entry under `[options.extras_require]`.
- Add a small `_require_matplotlib()` helper that raises a clear ImportError
  pointing at `pip install xarray-spatial[plot]` when matplotlib is missing.
- Call that helper at the top of each plotting entry point. The
  `_listed_colormap_from_attrs` helper in accessor.py already returns None on
  ImportError and runs on the read path rather than a plot request, so it stays
  as it is.

**Usage:** `pip install xarray-spatial` for compute only;
`pip install xarray-spatial[plot]` when you want `.xrs.plot()` and the other
plotting helpers.

**Value:** Smaller default install and fewer transitive dependencies, which
helps compute-only and CI environments in particular.

## Drawbacks

This changes behavior. Anyone who relied on matplotlib showing up implicitly
will need the `[plot]` extra or their own matplotlib install.

## Alternatives

Keep matplotlib as a hard dependency. That forces the plotting stack on every
user, which is the thing this issue is trying to fix.

Note: pandas can't be dropped the same way. xarray requires pandas>=2.2, and a
few core modules import it at top level.

brendancol:
Closes #2494

matplotlib was listed in `install_requires`, but nothing in the package imports
it at module top level. Every use is a lazy import inside the `.xrs.plot`
accessor helpers. So a plain `pip install xarray-spatial` was dragging in
matplotlib plus its whole chain (pillow, fonttools, kiwisolver, contourpy,
cycler, pyparsing) for users who only run the compute functions.

- Move `matplotlib` out of `install_requires` into a new `plot` extra.
- Add a `_require_matplotlib()` guard at the two plot entry points in
  `accessor.py`. Without matplotlib, `.xrs.plot()` now raises a clear
  ImportError pointing at `pip install xarray-spatial[plot]` instead of a bare
  `ModuleNotFoundError`.
- Add `matplotlib` to the `tests` extra so the existing plot tests keep running
  in CI.
- Update README, install docs, and CHANGELOG.

pandas is intentionally left alone: xarray requires it and several core modules
import it at top level, so it can't move to an extra.

This is a compute-only change, not backend-specific. No spatial operation
imports matplotlib on any backend.

### Test plan
- [x] `pytest xrspatial/tests/test_accessor.py` (29 passed) — includes new tests
  that block matplotlib in `sys.modules` and assert core modules still import
  and the plot helpers raise the friendly error.
- [x] `pytest xrspatial/geotiff/tests/test_features.py -k "plot or palette or colormap"` (8 passed)
- [x] Bare-import check: `import xrspatial` and the focal/zonal/dasymetric
  modules import with matplotlib blocked.


brendancol (commented):
## PR Review: Make matplotlib an optional [plot] extra (#2494)

Scope is packaging + two guard calls, so the geospatial-specific checks
(backends, dask, CUDA, numerics) don't apply here. I focused on whether the
dependency move is complete and the guard behaves.

### Blockers
None.

### Suggestions
- `_require_matplotlib()` (accessor.py:24) checks `import matplotlib` but the
  call sites then do `import matplotlib.pyplot as plt`. matplotlib can import
  while pyplot fails (no usable backend, broken install). That's an uncommon
  case and the error would still be an ImportError, just not the friendly one.
  Checking `import matplotlib.pyplot` in the guard would close the gap. Not
  blocking.

### Nits
- `test_compute_imports_without_matplotlib` calls `importlib.reload` on
  `xrspatial.focal`, `zonal`, and `dasymetric`. The reload re-executes those
  modules in place, so other tests in the same session keep the reloaded
  objects. It passes today, but reloading numba/ngjit modules for an
  import-side-effect check is heavier than needed. Importing the modules in a
  subprocess with matplotlib blocked, or just asserting the import doesn't
  raise, would isolate the check better.

### What looks good
- The core finding holds: no compute path imports matplotlib. The only runtime
  uses are the two `.xrs.plot` accessor methods; the matches in `zonal.py` and
  `bump.py` are inside `.. plot::` docstring examples, not executed code.
- matplotlib added to the `tests` extra, so the existing plot/palette tests keep
  running in CI rather than silently skipping.
- pandas correctly left as required, with the reason recorded.
- `_listed_colormap_from_attrs` kept as-is; it already returns None on
  ImportError and sits on the read path, so guarding it would have broken
  non-plotting GeoTIFF reads.
- New tests assert both the friendly-error path and that core modules import
  with matplotlib blocked.

### Checklist
- [x] No compute path imports matplotlib (verified across the package)
- [x] Friendly ImportError points at the `plot` extra
- [x] tests extra keeps plot tests running in CI
- [x] Docs (README, installation.rst) and CHANGELOG updated
- [x] pandas correctly identified as non-removable
- [ ] Guard covers `matplotlib.pyplot`, not just `matplotlib` (see suggestion)

brendancol (commented):
## Follow-up review (after 1e1157dc)

Both findings from the first pass are resolved.

- Suggestion (guard pyplot): `_require_matplotlib()` now does
  `import matplotlib.pyplot`, so an installed-but-unusable matplotlib also
  produces the friendly error. Resolved.
- Nit (reload-based test): `test_compute_imports_without_matplotlib` now runs
  in a subprocess against a clean module cache instead of reloading numba
  modules in-session. Resolved.

No new issues. `pytest xrspatial/tests/test_accessor.py` is 29 passed and
flake8 is clean on the changed files. No remaining items beyond CI.

PR #2497, opened 2026-05-27
Make shapely an optional [vector] extra (#2496)

### LINKED ISSUE #2496: Move shapely to an optional [vector] extra so it isn't loaded on import
## Reason or Problem

shapely is in `install_requires`, so a plain `pip install xarray-spatial`
installs it (a compiled package built on GEOS) and loads it on every
`import xrspatial`. But most of the library never touches shapely. It is only
used by the vector-to-raster paths: `rasterize` and `polygonize`. Users who run
terrain, focal, hydrology, or multispectral functions still pay for shapely on
import.

The eager load happens because `xrspatial/rasterize.py` imports shapely at
module top level (line 19) and `xrspatial/__init__.py` imports `rasterize` and
`polygonize` at top level. `polygonize` already imports shapely lazily inside
its functions; `rasterize` is the one module pulling it in at import time.

## Proposal

Move shapely out of `install_requires` into an optional `vector` extra, and make
`rasterize.py` import it lazily. Then `import xrspatial` no longer loads shapely,
and `pip install xarray-spatial[vector]` enables the rasterize/polygonize paths.

**Design:**
- Remove `shapely>=2.0` from `[options] install_requires` in setup.cfg.
- Add a `vector = shapely>=2.0` entry under `[options.extras_require]`.
- Replace the top-level `import shapely` in `rasterize.py` with a
  `_require_shapely()` helper that returns the module or raises a clear
  ImportError pointing at `pip install xarray-spatial[vector]`. Call it at the
  entry points that use shapely.
- Keep shapely in the `tests` extra so CI still exercises rasterize/polygonize.

**Usage:** `pip install xarray-spatial` for everything except the vector paths;
`pip install xarray-spatial[vector]` to use `rasterize` and `polygonize`.

**Value:** shapely (and GEOS) no longer load on `import xrspatial`, and they drop
off the default install for users who don't rasterize vector geometry.

## Drawbacks

This changes behavior. Code that relied on shapely arriving with xarray-spatial
will need the `[vector]` extra. `rasterize` and `polygonize` raise a clear error
instead of working out of the box.

## Alternatives

Keep shapely required. Rejected for the same reason as the matplotlib change
(#2494): it forces a compiled dependency on users who never call the paths that
need it.

This follows the pattern set by #2494 (matplotlib moved to a `plot` extra).
urllib3 is intentionally left as a hard dependency of the geotiff module and is
out of scope here.

brendancol:
Closes #2496

shapely (and GEOS) loaded on every `import xrspatial` because `rasterize.py`
imported it at module top level, even though only `rasterize` and `polygonize`
use it. This moves it to an optional `vector` extra, matching the matplotlib
change in #2494.

- Replace the top-level `import shapely` in `rasterize.py` with a cached
  `_require_shapely()` helper. Each function that uses the shapely array API
  binds the module locally, so dask tile workers (which call the tile helpers
  directly, not through `rasterize`) also get the clear error. `rasterize()`
  calls it up front so a missing install fails early with a message pointing at
  `pip install xarray-spatial[vector]`.
- Remove `shapely>=2.0` from `install_requires`; add a `vector` extra and keep
  shapely in the `tests` extra so CI still exercises the vector paths.
- `polygonize` already imports shapely lazily, and only on its `geopandas`
  return path (which needs geopandas anyway), so it needs no change.
- Update README, install docs, and CHANGELOG.

This is a packaging change, not backend-specific. The rasterize backends
(numpy / cupy / dask+numpy / dask+cupy) are unchanged; shapely is only used for
geometry classification and coordinate extraction, which feed all backends.

### Test plan
- [x] New `test_optional_shapely.py` (3 passed): `import xrspatial` works in a
  subprocess with shapely blocked; `_require_shapely()` and `rasterize()` raise
  the friendly error pointing at the `vector` extra.
- [x] `pytest xrspatial/tests/test_rasterize.py xrspatial/tests/test_polygonize.py`
  (356 passed, 15 skipped) with shapely installed.
- [x] Confirmed `import xrspatial` no longer loads shapely (`shapely` absent
  from `sys.modules`) and a real rasterize still produces correct output.


brendancol (commented):
## PR Review: Make shapely an optional [vector] extra (#2496)

Packaging change plus lazy-import plumbing. The rasterize algorithm and its
four backends are untouched, so the numeric/backend checklist mostly doesn't
apply here. I focused on whether every shapely use path is actually guarded.

### Blockers
None.

### Suggestions
None blocking.

### Nits
- `polygonize.py` still does a bare `import shapely` inside `_to_geopandas`
  (around line 609). Without shapely that raises a plain ImportError rather than
  the `[vector]`-pointing message. In practice geopandas is imported first on
  that same path and geopandas requires shapely, so the user hits geopandas's
  own error first -- but the messaging is slightly inconsistent with the new
  rasterize behavior. Optional to route it through `_require_shapely` for a
  uniform message.
- `test_optional_shapely.py::test_import_xrspatial_without_shapely` imports
  `xrspatial.rasterize` in the subprocess but not `xrspatial.polygonize`. Adding
  polygonize to that import list would lock in that it also imports clean
  without shapely.

### What looks good
- Every bare `shapely.` reference (13 of them) sits in a function that binds
  `shapely = _require_shapely()` first; `rasterize()` calls the guard up front
  for an early, clear failure. Verified each call site against its enclosing
  function.
- The dask path is handled correctly: tile workers call `_polys_from_wkb` /
  `_classify_geometries` directly, not through `rasterize()`, and those helpers
  bind via `_require_shapely()` so the worker either imports shapely or raises
  the friendly error. The cache lives in a module global, so each worker
  process resolves it once.
- Confirmed `import xrspatial` no longer loads shapely (absent from
  `sys.modules`), and a real rasterize still produces correct output.
- shapely kept in the `tests` extra, so CI still exercises the vector paths;
  356 rasterize/polygonize tests pass.
- New tests cover the subprocess import, the helper message, and the up-front
  `rasterize()` error, and they reset the `_shapely` cache so the in-process
  checks are real.

### Checklist
- [x] Every shapely use path guarded (rasterize + dask helpers)
- [x] `import xrspatial` no longer loads shapely
- [x] Friendly ImportError points at the `vector` extra
- [x] tests extra keeps shapely so CI exercises vector paths
- [x] Docs (README, installation.rst) and CHANGELOG updated
- [ ] polygonize geopandas path uses the same `[vector]` message (nit)
- [x] No backend/algorithm change; existing rasterize tests pass

brendancol (commented):
## Follow-up review (after 7047f03e)

Disposition of the two nits from the first pass:

- Nit 2 (subprocess import test): fixed. `test_import_xrspatial_without_shapely`
  now also imports `xrspatial.polygonize` with shapely blocked, locking in that
  it imports clean too. 3 passed.
- Nit 1 (route polygonize's geopandas path through `_require_shapely`):
  dismissed. On that path `import geopandas` runs first, and geopandas
  hard-requires shapely, so a "geopandas installed but shapely missing" state is
  not reachable. Adding a cross-module import for a message that can't surface
  isn't worth it.

No remaining items beyond CI.

PR #2509, opened 2026-05-27
reproject: preserve integer dtype in dask+cupy fast path (#2505)

### LINKED ISSUE #2505: reproject: dask+cupy fast path drops integer dtype to float64
## Summary

`xrspatial.reproject.reproject()` silently changes the output dtype to `float64` when the input is a dask-backed cupy array with an integer dtype. The other three backends (numpy, cupy, dask+numpy) and the chunked dask+cupy fallback all preserve the source integer dtype.

## Reproducer

```python
import numpy as np
import xarray as xr
import cupy as cp
import dask.array as da

data = np.arange(64, dtype=np.int16).reshape(8, 8)
raster = xr.DataArray(
    da.from_array(cp.asarray(data), chunks=(4, 4)),
    dims=['y', 'x'],
    coords={'y': np.linspace(5, -5, 8), 'x': np.linspace(-5, 5, 8)},
    attrs={'crs': 'EPSG:4326', 'nodata': -32768},
)
result = reproject(raster, 'EPSG:4326', resolution=1.0)
print(result.dtype)  # float64  <-- expected int16
```

Comparison across backends with the same int16 input:

| Backend       | Output dtype |
|---------------|--------------|
| numpy         | int16        |
| dask+numpy    | int16        |
| cupy (eager)  | int16        |
| dask+cupy     | **float64**  |

## Root cause

`_reproject_dask_cupy` in `xrspatial/reproject/__init__.py` enters an eager fast path when the output fits in GPU memory. That path allocates the output with `cp.full(out_shape, nodata, dtype=cp.float64)` and writes `_resample_cupy_native` / `_resample_cupy` chunks into it without ever casting back to the source dtype.

The chunked fallback (`_reproject_dask(is_cupy=True)`) already gets this right because `_reproject_dask` computes `out_dtype` from `raster.dtype` and `_reproject_chunk_cupy` clamps each chunk back to the source integer dtype.

## Proposed fix

In `_reproject_dask_cupy`:

1. Compute `out_dtype` the same way `_reproject_dask` does — integer source keeps its dtype, float source stays `float64`.
2. Allocate `result = cp.full(out_shape, nodata, dtype=out_dtype)`.
3. After each chunk resample, clamp and cast back to `out_dtype` when the source was integer (mirroring `_reproject_chunk_cupy`).

## Why this matters

A user reading `result.attrs['nodata'] == -32768` while the array is float64 has metadata that disagrees with the data on cross-backend round-trips (Cat 5: backend-inconsistent metadata). Memory blowup for large integer rasters on GPU is a secondary concern.

## Test coverage

Add a `TestDaskCupyDtypeParity` class (gated on CUDA + cupy + dask) mirroring `TestDaskDtypeParity`:

- `test_dask_cupy_reproject_int8_preserves_dtype`
- `test_dask_cupy_reproject_uint16_preserves_dtype`
- `test_dask_cupy_reproject_float32_stays_float64`

brendancol:
Closes #2505.

## Summary

- `_reproject_dask_cupy` eager fast path force-allocated the output buffer as `cp.float64` regardless of source dtype, so int16/uint8/etc. inputs were silently promoted to float64 while every other backend kept the source dtype.
- Compute `out_dtype` the same way `_reproject_dask` does (integer source keeps its dtype, float source stays float64), allocate the result with `out_dtype`, and clamp+cast each chunk before placement -- mirrors the per-band cast in `_reproject_chunk_cupy`.
- New `TestDaskCupyDtypeParity` class with six tests covering int8 / int16 / uint8 / uint16 / float32 plus a direct dask+numpy vs dask+cupy cross-backend dtype check.

## Backend coverage

| Backend       | Before        | After        |
|---------------|---------------|--------------|
| numpy         | int16 -> int16    | unchanged    |
| dask+numpy    | int16 -> int16    | unchanged    |
| cupy (eager)  | int16 -> int16    | unchanged    |
| dask+cupy     | int16 -> **float64** | int16 -> int16 |

## Test plan

- [x] `pytest xrspatial/tests/test_reproject.py::TestDaskCupyDtypeParity` -- 6 passing
- [x] `pytest xrspatial/tests/test_reproject.py` -- 351 passing (345 baseline + 6 new)
- [x] New tests fail against pre-fix code, pass against fixed code


brendancol (commented):
## PR Review: reproject: preserve integer dtype in dask+cupy fast path (#2505)

### Blockers
None.

### Suggestions
None.

### Nits
- The new `TestDaskCupyDtypeParity` tests check `result.dtype` only. The sibling `TestDaskDtypeParity` checks both `result.data.dtype` and `result.compute().dtype`. Since `_reproject_dask_cupy` returns an eager cupy ndarray (not a dask array), `result.compute()` is a no-op so the two forms are equivalent here. Fine as-is; add `result.data.dtype` if you want exact symmetry with the sibling class.

### What looks good
- Same pattern as `_reproject_dask` (lines 1839-1843) and `_reproject_chunk_cupy` (lines 609-611, 633). No new patterns introduced.
- Clamp+cast on `chunk_data` runs at the per-chunk write level, so chunks skipped via `continue` retain the pre-filled `out_dtype` nodata.
- The `nodata` value reaching `_reproject_dask_cupy` is already integer-compatible when the source is integer (resolved by `_detect_nodata(..., dtype=raster.dtype)` at line 851), so the `cp.full(out_shape, nodata, dtype=int_dtype)` allocation can't hit a NaN→int conversion.
- New tests fail on pre-fix code and pass on fixed code (verified by temporarily reverting).
- Six tests covering int8 / int16 / uint8 / uint16 / float32 plus a direct dask+numpy vs dask+cupy parity check.

### Checklist
- [x] Pattern matches `_reproject_dask` and `_reproject_chunk_cupy`
- [x] All backends now produce consistent dtypes
- [x] NaN handling unchanged
- [x] Edge cases: signed, unsigned, float, skipped chunks
- [x] Benchmark not needed (correctness fix)
- [x] README feature matrix unaffected
- [x] Inline comments reference #2505

brendancol (commented):
## Follow-up review (#2505)

Addressed the nit from the previous review: `TestDaskCupyDtypeParity` now asserts both `result.dtype` and `result.data.dtype`, matching the symmetry used in `TestDaskDtypeParity`.

No outstanding findings. All 6 new tests still pass.

Disposition of original review:
- Nit (assertion symmetry): fixed in bd45b820.

PR #2510, opened 2026-05-27
rasterize: hoist poly_props/poly_global cupy.asarray above all_touched (#2506)

### LINKED ISSUE #2506: rasterize: duplicate cupy.asarray(poly_props/poly_global) when all_touched=True
## Summary

In `_run_cupy` (`xrspatial/rasterize.py`) and `_rasterize_tile_cupy`, when `all_touched=True` and the input contains polygons, `cupy.asarray(poly_props)` and `cupy.asarray(poly_global)` are called twice: once when staging the scanline `poly_launch` tuple, and again when staging the supercover boundary `boundary_launch` tuple.

For the dask+cupy path this duplicate transfer fires for every tile, doubling the host-to-device traffic on the per-tile props/global tables. The two launches operate on the same tile, so there is no benefit to allocating separate device copies.

## Locations

- `xrspatial/rasterize.py:2065` -- `poly_launch` line in `_run_cupy`
- `xrspatial/rasterize.py:2083` -- `boundary_launch` line in `_run_cupy` (same poly_props/poly_global)
- `xrspatial/rasterize.py:2541` -- `poly_launch` line in `_rasterize_tile_cupy`
- `xrspatial/rasterize.py:2555` -- `boundary_launch` line in `_rasterize_tile_cupy` (same poly_props_2d/poly_global_2d)

## Impact

Microbenched on the current host:

- 1000 polygons x 4 cols: duplicate 0.051 ms/iter vs single 0.024 ms/iter (2.1x)
- 10000 polygons x 8 cols: duplicate 0.218 ms/iter vs single 0.092 ms/iter (2.4x, saves 720 KB/tile)

For a 100-tile dask+cupy raster over 10k polygons with `all_touched=True`, that is ~13 ms and 72 MB of redundant PCIe traffic eliminated per call.

## Proposed fix

Stage `cupy.asarray(poly_props)` and `cupy.asarray(poly_global)` once before the scanline / boundary conditional, and have both launch tuples reference the same device arrays.

## Backend / severity

- Backends affected: `cupy`, `dask+cupy`
- Severity: MEDIUM
- Category: Cat-3 GPU transfer (redundant cupy.asarray)
- OOM verdict: SAFE/graph-bound (unchanged)

## Discovered via

`/deep-sweep` performance pass on `rasterize` (2026-05-27).

brendancol:
## Summary

- `_run_cupy` and `_rasterize_tile_cupy` previously transferred `poly_props` and `poly_global` to the GPU twice when `all_touched=True` (once for the scanline launch, once for the supercover boundary launch). Hoist the `cupy.asarray()` calls above the scanline / boundary conditional so both launches share the same device buffer.
- For 10k polygons / 8 columns, the per-tile props transfer drops from 0.218 ms to 0.092 ms (2.4x), saving 720 KB / tile of redundant PCIe traffic. On a 100-tile dask+cupy raster that is ~13 ms and 72 MB saved per call.
- Closes #2506.

## Test plan

- [x] 4 new AST-level assertions in `test_rasterize_props_hoist_2506.py` confirm each function calls `cupy.asarray(poly_props/poly_global)` exactly once.
- [x] 5 new cupy vs numpy parity tests covering `last`/`first`/`max`/`min`/`sum` merges under `all_touched=True`.
- [x] 3 new dask+cupy smoke tests that exercise the hoisted upload through every per-tile launch.
- [x] All 470 existing rasterize tests pass alongside the 12 new ones.

## Notes

The dask+cupy + `all_touched` pixel-level parity gap (boundary segments crossing tile borders behave differently than the eager numpy path) predates this fix and is not addressed here. The smoke tests assert the hoisted launches still produce a populated raster rather than asserting full numpy parity.

Discovered via `/deep-sweep` performance pass on `rasterize` (2026-05-27).


brendancol:
Self-review pass via /review-pr. 0 blockers, 1 suggestion, 1 nit, both addressed on this branch.

**S1: gate the hoist on actual usage**

As originally landed, the hoist uploads `poly_props` / `poly_global` to the GPU whenever `poly_geoms` (or `poly_wkb`) is non-empty. If all polygons fall outside the raster bounds (so `edge_y_min` is empty) and `all_touched` is False, neither launch fires, so the upload becomes work the pre-hoist code did not do. Tightened the guard to `len(edge_y_min) > 0 or all_touched` in both `_run_cupy` and `_rasterize_tile_cupy`. The `all_touched` fast path still gets the single shared upload that #2506 was about. Commit 958a5576.

**N1: strip agent prompt residue from the test**

The `_to_numpy` helper carried a comment referencing "Skill instructions" about `.compute()`. That is prompt context that leaked into committed code. Removed. Commit 2b439815.

Tests: 464 passed, 2 skipped (full `xrspatial/tests/test_rasterize*.py` suite) after both fixes.

PR #2511, opened 2026-05-27
test(rasterize): close eager-cupy merge-mode coverage gap (deep-sweep pass 3)

brendancol:
## Summary

Deep-sweep test-coverage pass 3 on `xrspatial.rasterize`.  Adds 23 tests in `test_rasterize_coverage_2026_05_27.py`, all passing on a CUDA host.  No source changes.

- **Cat 1 HIGH** -- eager cupy (`use_cuda=True`, no `chunks=`) had no parametrised six-mode merge parity test against eager numpy.  `TestDaskNumpy.test_merge_mode_parity` and `TestDaskCupy.test_merge_mode_parity` carry it for the tiled paths, and pass 1 only pinned `merge='last'` on a single non-overlapping polygon.  A routing regression in any of the six GPU atomic kernels (`_ensure_gpu_kernels`, rasterize.py:1308-1556) would slip past the dask+cupy parity tests because the tiled finalize path is different.  Pin polygon and point overlap scenes across `last` / `first` / `max` / `min` / `sum` / `count`, with sanity checks (`first != last`, `min < max`) that lock the fixture as non-degenerate.
- **Cat 1 MEDIUM** -- `_run_cupy` with zero geometries had no direct test (eager numpy and both dask backends did).  Pin that `use_cuda=True` with `[]` keeps the cupy backend instead of short-circuiting to numpy, under both an explicit fill and the default NaN fill.
- **Cat 2 MEDIUM** -- `merge='count'` with all-equal property values: four overlapping rectangles burning the same `1.0` must still count overlaps as > 1 on all four backends.  A future GPU atomic optimisation that deduplicated identical-value writes would silently divide counts by the number of unique values.
- **Cat 4 MEDIUM** -- `name=` kwarg thread-through on `dask+numpy` / eager cupy / `dask+cupy` (eager numpy was the only path covered).

State CSV updated: `rasterize` row -> `last_inspected=2026-05-27`, `severity_max=HIGH`, `categories_found=1;2;4`.

## Test plan

- [x] `pytest xrspatial/tests/test_rasterize_coverage_2026_05_27.py` -- 23 passed locally on CUDA host
- [x] `pytest xrspatial/tests/test_rasterize*.py` -- 325 passed, 2 skipped (no regressions in passes 1+2 or the main test file)
- [ ] CI green on numpy-only and dask-only runners (no GPU markers must be hit there)


brendancol:
One blocker, two nits. Fix in 99b9b3e6.

**Blocker -- module-level `box(...)` crashes import without shapely**

`_EAGER_CUPY_MERGE_PAIRS`, `_ALL_EQUAL_PAIRS`, and `TestNameKwargBackends._SIMPLE_PAIR` all called `box(...)` at module scope. Since #2496 made shapely optional, a numpy-only runner hits `NameError: name 'box' is not defined` during collection, before `pytestmark` skip has a chance to fire.

Reproduced by injecting `ImportError` on `import shapely.*` and running `pytest --collect-only`: prior tip failed, HEAD collects all 23 tests cleanly.

Fix: wrap the two list constants in `if has_shapely:` and turn the class attribute into a `_simple_pair()` method so `box` is only touched from test bodies.

**Nits (not pushed)**

1. `test_eager_cupy_first_differs_from_last` and `test_eager_cupy_min_differs_from_max` sit under `@skip_no_cuda` but only call the numpy backend. The non-degeneracy property they pin doesn't depend on the backend, so they'd still be useful on numpy-only runners. Not worth re-shuffling.

2. `test_count_numpy` asserts `max >= 3.0`; observed max is 4.0 (four-way overlap at world (5.5, 5.5)). `== 4.0` would catch a regression that dropped one of the four writes, but the `>= 3.0` bound is what the test was actually built to defend against (dedup-to-1), so leaving it.

**Test status**

- new file: 23 passed (CUDA host)
- new file + `test_rasterize.py`: 238 passed, 2 skipped
- `--collect-only` with shapely suppressed: 23 collected, no NameError

Issue #1903, opened 2026-05-15
geotiff: preflight CUDA in read_geotiff_gpu so broken-driver errors are clean


brendancol:
**Describe the bug**

`open_geotiff(path, gpu=True)` (and `read_geotiff_gpu` directly) raises `cudaErrorInsufficientDriver` deep inside `cupy.asarray(...)` at `xrspatial/geotiff/_backends/gpu.py:520` when CuPy is installed but the CUDA driver is unusable. The docstring (lines 124-127) acknowledges this propagates even in `on_gpu_failure='auto'` mode.

The combination "CuPy installed, driver broken" is common in two situations:

- containers where the host driver is older than the CuPy build expects
- machines that had a GPU but lost driver state (suspended VMs, removed devices)

In both, the user gets a low-level CUDA error from the middle of a function call rather than a clear "CUDA is not usable" message.

**Expected behavior**

`read_geotiff_gpu` should preflight CUDA availability after the `import cupy` step and raise a clean `RuntimeError` if the driver does not respond. The user then knows immediately to either fix the driver, install a matching CuPy build, or pass `gpu=False`.

**Reproduction sketch**

```python
# install cupy-cuda12x on a machine with a broken or missing driver
from xrspatial.geotiff import open_geotiff
open_geotiff("any.tif", gpu=True)
# -> cupy.cuda.runtime.CUDARuntimeError: cudaErrorInsufficientDriver
```

**Additional context**

Cheapest preflight is `cupy.cuda.runtime.getDeviceCount()`, which raises `CUDARuntimeError` when the driver isn't usable and returns 0 when no device is visible. Wrap the result in a `RuntimeError` that names the original CUDA error so the message stays diagnostic.

Found in external code review.

PR #2518, opened 2026-05-27
Decouple GPU alias test from real CUDA runtime (#2515)

### LINKED ISSUE #2515: GeoTIFF GPU alias test fails on CuPy-installed but unusable-CUDA environments
## Summary

`test_gpu_alias_accepts_old_values_without_validation_error_1560` (around line 232 of `xrspatial/geotiff/tests/gpu/test_kernels_and_kwargs.py`) fails on hosts where CuPy imports but the CUDA runtime can't actually run. The test does not accept the `RuntimeError` that `xrspatial/geotiff/_backends/gpu.py:38` raises by design in that case.

## Reproducer

On a host with CuPy installed but no usable CUDA driver:

```
pytest -q xrspatial/geotiff/tests -m 'not slow' -x -vv
1 failed, 323 passed, 74 skipped, 54 deselected
```

Failure trace:

```
RuntimeError: read_geotiff_gpu: CUDA runtime is not usable
(cudaErrorInsufficientDriver: CUDA driver version is insufficient ...)
```

## Why this matters

The preflight `RuntimeError` is the documented behaviour from issue #1903. The bug is in the test, not the preflight. The geotiff suite is supposed to be a release gate, and right now it can't run cleanly on the common developer / CI setup where CuPy is importable but CUDA is not.

## Required correction

Either:

1. Accept the documented CUDA preflight `RuntimeError` alongside the exceptions the test already tolerates, or
2. Stop touching CUDA at all in the alias test (preferred). Reuse the existing `_install_cupy_stub_1903` helper so the test verifies the alias value passes validation without ever calling into a real CUDA runtime.

Either way the full geotiff suite should pass on CuPy-installed-but-CUDA-unusable hosts.

## Out of scope

Don't change the preflight `RuntimeError`. It's the documented contract for issue #1903.

brendancol:
Closes #2515.

## Summary

- `test_gpu_alias_accepts_old_values_without_validation_error_1560` was failing on hosts where CuPy imports but the CUDA runtime can't run. The test sends `gpu='strict'` through `read_geotiff_gpu` to confirm the deprecated alias passes value validation. On a broken-CUDA host the call hits `_preflight_cuda_runtime` (the documented behaviour from issue #1903) and the resulting `RuntimeError` is not in the test's accepted-exception tuple.
- Fix: reuse the existing `_install_cupy_stub_1903` helper to stub cupy with a preflight that reports one device. The alias call then reaches the file-read stage on every host and fails with `FileNotFoundError`, which is what the alias check needs to assert.
- The preflight `RuntimeError` itself is unchanged; it's the correct contract for #1903.

## Backend coverage

Test-only change. No public API surface, no backend dispatch, no GPU code paths touched. Alias-validation coverage now runs uniformly on CPU-only, broken-CUDA, and GPU hosts.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/gpu/test_kernels_and_kwargs.py::test_gpu_alias_accepts_old_values_without_validation_error_1560` passes on a CuPy-installed-but-CUDA-unusable host.
- [x] `pytest -q xrspatial/geotiff/tests -m 'not slow'` is fully green on that same host (5913 passed, 0 failed).
- [ ] CI validates the same fix on CPU-only and (where available) GPU runners.


brendancol (commented):
## PR Review: Decouple GPU alias test from real CUDA runtime (#2515)

### Blockers
None.

### Suggestions
None.

### Nits
- [ ] `xrspatial/geotiff/tests/gpu/test_kernels_and_kwargs.py:76-81` — the `_install_cupy_stub_1903` docstring says "Used on machines without cupy installed; lets us exercise the preflight failure path on CPU-only CI." This PR also uses it on broken-CUDA hosts for the preflight success path, which is a wider role than the docstring describes. A one-line tweak to the helper's docstring would keep its contract aligned with its callers.

### What looks good
- The fix targets the test, not the preflight contract. The PR body says so directly.
- Reuses `_install_cupy_stub_1903` instead of inventing a new pattern.
- `monkeypatch.setitem(sys.modules, ...)` is function-scoped and auto-restored, so sibling tests aren't affected.
- Drops `ImportError` from the accepted-exception tuple now that the stub guarantees cupy resolves. An unexpected `ImportError` will now fail loudly rather than be silently tolerated.
- Test-only. No public API, no GPU code paths, no backend dispatch touched.
- Verified locally: 5913 passed, 0 failed on the CuPy-installed-but-CUDA-unusable host that reproduced the original bug.

### Checklist
- [x] Algorithm matches reference/paper — N/A (test-only)
- [x] All implemented backends produce consistent results — N/A (alias coverage now uniform across hosts)
- [x] Edge cases covered — alias validation passes on CPU-only, broken-CUDA, and GPU hosts
- [x] Benchmark exists or is not needed — not needed (test-only)
- [x] README feature matrix updated — N/A
- [x] Docstrings present and accurate — yes

brendancol (commented):
## PR Review: Decouple GPU alias test from real CUDA runtime (#2515) — follow-up pass

### Blockers
None.

### Suggestions
None.

### Nits
None. The earlier nit on helper docstring scope was handled in 3164b236; the docstring now covers both the failure-path and success-path callers.

### What looks good
- The docstring update names the alias-validation use case directly (mentions #2515), so anyone reading later can see why the success-path mode exists.
- Diff is still scoped to one test file. No drift into unrelated code.
- All 163 tests in the file pass after the docstring tweak.

### Checklist
- [x] Algorithm matches reference/paper — N/A (test-only)
- [x] All implemented backends produce consistent results — N/A
- [x] Edge cases covered — alias validation runs uniformly across hosts
- [x] Dask chunk boundaries — N/A
- [x] No premature materialization — N/A
- [x] Benchmark — not needed
- [x] README feature matrix — N/A
- [x] Docstrings present and accurate — yes (helper docstring is broader after the follow-up commit)

PR #2519, opened 2026-05-27
Reject distinct per-band nodatavals on GeoTIFF write (#2514)

### LINKED ISSUE #2514: GeoTIFF writer silently flattens distinct per-band nodatavals
**Describe the bug**

The GeoTIFF writer silently flattens distinct per-band `nodatavals` into a single scalar, corrupting missing-data semantics.

`_resolve_nodata_attr` in `xrspatial/geotiff/_attrs.py:1246` picks the first usable value from `attrs["nodatavals"]`. The conflict check in `xrspatial/geotiff/_validation.py:1000` allows differing per-band values when one band happens to match the selected scalar, and skips the check entirely when only `nodatavals` is present (no scalar `nodata` key).

**Reproduction (public API)**

```python
import numpy as np
import xarray as xr

band1 = np.array([[1.0, -9999.0], [3.0, 4.0]], dtype=np.float32)
band2 = np.array([[2.0, -8888.0], [5.0, 6.0]], dtype=np.float32)
data = np.stack([band1, band2])
da = xr.DataArray(
    data,
    dims=("band", "y", "x"),
    coords={"band": [1, 2], "y": [1.0, 0.0], "x": [0.0, 1.0]},
    attrs={"nodatavals": (-9999.0, -8888.0)},
)

to_geotiff(da, "out.tif")
out = open_geotiff("out.tif")
# written nodata: -9999.0
# band 0: NaN at the -9999 cell  (correctly masked)
# band 1: -8888.0 preserved as real data  (sentinel became real data)
```

Same corruption with `attrs={"nodata": -9999, "nodatavals": (-9999, -8888)}` because one band's sentinel happens to match the canonical scalar, so the conflict check returns without raising.

**Expected behavior**

The writer should reject distinct per-band sentinels with a clear error rather than silently dropping one. The repo already documents the correct limitation in `xrspatial/geotiff/tests/parity/test_reference.py:627`: a TIFF `GDAL_NODATA` value is file-wide, so distinct per-band sentinels cannot safely round-trip in this representation.

**Affected paths**

The faulty validation feeds CPU writes in `xrspatial/geotiff/_writers/eager.py:472`, GPU writes in `xrspatial/geotiff/_writers/gpu.py:404`, and tiled/VRT-driven writing through the same eager validation path.

**Coverage gap**

- `xrspatial/geotiff/tests/unit/test_metadata.py:1766` rejects only the case where every tuple member disagrees with the canonical value.
- `xrspatial/geotiff/tests/read/test_nodata.py:248` covers single-valued `nodatavals`, not distinct per-band sentinels.

**Proposed fix**

- Tighten `_validation.py` so any time `nodatavals` contains 2+ distinct concrete (non-NaN, non-None) sentinels, validation raises a clear error on write, regardless of whether `nodata` is also provided or whether one of them happens to equal the flattened scalar.
- Update `_attrs.py:1246` flattening so it does not silently pick a winner on disagreement; the error must surface from validation before the writer sees a flattened scalar.
- Cover both CPU (`_writers/eager.py`) and GPU (`_writers/gpu.py`) paths since they share validation.
- Add tests for the dangerous case (distinct per-band sentinels, with and without an accompanying scalar `nodata` key) and confirm the error fires before any file write.
- Allow the safe cases: a single distinct concrete value (possibly repeated), all-NaN, or `None`s mixed with a single concrete value.

brendancol:
Closes #2514.

## Summary

- A TIFF stores one `GDAL_NODATA` tag per file. The legacy resolver flattened a multi-entry `attrs['nodatavals']` tuple to its first usable value and silently dropped the rest, so any band declaring a different sentinel had its missing-data cells written as real data on the output file.
- New write-side check `_check_write_distinct_per_band_nodatavals` in `_validation.py` raises `ConflictingNodataError` whenever `nodatavals` holds 2+ distinct concrete values, with or without an accompanying scalar `attrs['nodata']`. The earlier check covered only the case where every tuple entry disagreed with the scalar; the dangerous case (one entry happens to match) slipped through.
- `_resolve_nodata_attr` in `_attrs.py` carries the same distinct-value check as defense-in-depth, so any internal bypass of validation surfaces the error rather than silently picking a winner.
- Safe cases still write: a single distinct concrete value (possibly repeated), all-NaN, `None` mixed with a single concrete value, and explicit `to_geotiff(..., nodata=X)` override.

## Backend coverage

The validation is shared by all write paths: CPU eager (`_writers/eager.py`), GPU (`_writers/gpu.py`), and tiled / VRT-driven writing through the same eager validation. No backend-specific code was changed.

## Test plan

- [x] `pytest -q xrspatial/geotiff/tests/unit/test_metadata.py -k "distinct_per_band or repeated_concrete or none_and_single or all_nan_nodatavals or distinct_nodatavals_explicit"` (7 new tests)
- [x] `pytest -q xrspatial/geotiff/tests/unit/test_metadata.py -k "nodata"` (31 tests, all pass)
- [x] `pytest -q xrspatial/geotiff/tests -m 'not slow'` (5920 passed, 74 skipped)
- [x] Manual reproduction confirms the original bug case now raises before the file is written


brendancol (commented):
## PR Review: Reject distinct per-band nodatavals on GeoTIFF write (#2514)

None.

- The defensive raise in `_resolve_nodata_attr` (`xrspatial/geotiff/_attrs.py:1310-1320`) has no direct unit test. Current tests all hit the writer-side `_check_write_distinct_per_band_nodatavals` first, so the resolver's defensive branch is never exercised on its own. Add one test that calls `_resolve_nodata_attr({'nodatavals': (-9999.0, -8888.0)})` directly and asserts `ConflictingNodataError`. Cheap to add, keeps the two layers honest if they drift.

- The error message in `_validation.py:1166-1175` and the one in `_attrs.py:1311-1319` are nearly identical but not byte-identical. Either share a single message builder or accept the duplication intentionally; if the latter, a comment cross-referencing the other site would help future readers.
- In `_attrs.py:1283-1291` the `usable: list = []` annotation lacks a parameter, e.g. `usable: list[tuple[float, object]] = []`. Minor type-hint precision.

### What looks good
- Validator is registered alongside the existing `_check_write_conflicting_nodata`, so all three write entry points (CPU eager, GPU, streaming) pick it up via the shared context-dict pattern.
- Tests assert the error fires before any file is written (`assert not os.path.exists(out_path)`), which is the right contract.
- Safe cases (repeated value, `None` + single concrete, all-NaN, explicit kwarg override) are all covered.
- Float equality dedup is correct here: the TIFF GDAL_NODATA tag stores one scalar, so any unequal pair is a real collision.

### Checklist
- [x] Algorithm matches reference/paper (n/a, validation tightening)
- [x] All implemented backends produce consistent results (shared validator)
- [x] NaN handling is correct (NaN entries skipped, all-NaN tuple still allowed)
- [x] Edge cases are covered by tests (3-band, repeated, all-NaN, `None` + single)
- [x] Dask chunk boundaries handled correctly (n/a, write metadata check, no compute)
- [x] Benchmark exists or is not needed (not needed, validation only)
- [x] README feature matrix updated if applicable (n/a, no new function)

brendancol (commented):
## PR Review (follow-up): Reject distinct per-band nodatavals on GeoTIFF write (#2514)

Reviewing commit `eb88dace` against the previous round.

None.

None.

None.

### What looks good
- The shared message builder `_distinct_per_band_nodatavals_msg` lives next to `ConflictingNodataError` in `_errors.py:107-127`, and both `_validation._check_write_distinct_per_band_nodatavals` and `_attrs._resolve_nodata_attr` now call it. Wording stays byte-identical at both sites.
- Type hint tightened to `usable: list[tuple[float, Any]]` in `_attrs.py:1296` — reuses the existing `Any` import.
- `test_resolve_nodata_attr_raises_on_distinct_per_band` (`tests/unit/test_metadata.py:1899`) hits the resolver directly, and the same test asserts the `attrs['nodata']` short-circuit case so the precedence contract stays documented.
- Full geotiff test suite (5921 passed, 74 skipped, 1 xfailed) stays green.

### Checklist
- [x] Algorithm matches reference/paper (n/a, validation tightening)
- [x] All implemented backends produce consistent results (shared validator)
- [x] NaN handling is correct
- [x] README feature matrix updated if applicable (n/a)

All three findings from the first round are resolved. No new findings.

PR #2529, opened 2026-05-27
Fix hypsometric_integral dask+cupy backend (#2525)

### LINKED ISSUE #2525: `hypsometric_integral` dask+cupy backend silently returns dask+numpy result
## Description

`_hi_dask_cupy` converts cupy chunks to numpy via `map_blocks(.get())` and then calls `_hi_dask_numpy`. The returned dask graph yields numpy chunks, so the wrapping `xr.DataArray` is dask+numpy backed even though the caller passed dask+cupy.

`is_dask_cupy(result)` returns False. Downstream code that dispatches on backend type (another zonal function using `ArrayTypeFunctionMapping`) will route through the numpy path. A GPU-resident pipeline loses GPU residency at this hop and the caller never finds out.

`_regions_dask_cupy` in the same module does this correctly: it materializes on the GPU and returns a dask-cupy result.

## Reproducer

```python
import numpy as np
import cupy as cp
import dask.array as da
import xarray as xr
from xrspatial.zonal import hypsometric_integral
from xrspatial.utils import is_dask_cupy

zones_np = np.array([[1, 1, 2, 2], [1, 1, 2, 2]], dtype=np.int32)
values_np = np.array([[10.0, 20.0, 30.0, 40.0], [15.0, 25.0, 35.0, 45.0]])

zones = xr.DataArray(
    da.from_array(cp.asarray(zones_np), chunks=(2, 2)), dims=['y', 'x'],
)
values = xr.DataArray(
    da.from_array(cp.asarray(values_np), chunks=(2, 2)), dims=['y', 'x'],
)

assert is_dask_cupy(zones)
assert is_dask_cupy(values)

result = hypsometric_integral(zones, values)
assert is_dask_cupy(result)  # FAILS: result is dask+numpy
```

## Expected

A dask+cupy input should produce a dask+cupy output. `is_dask_cupy(result)` should be True and `result.data.blocks[0, 0].compute()` should return a `cupy.ndarray`.

## Suggested fix

After `_hi_dask_cupy` calls `_hi_dask_numpy`, wrap the result through `map_blocks(cupy.asarray, ...)` so every chunk lands back as a cupy array. The dask graph stays lazy. The numpy reduce in the middle stays as-is (the per-zone HI lookup is a Python dict), but the painted output matches the input backend.

## Audit trail

Surfaced by `/sweep-metadata` on the `zonal` module, 2026-05-27. Cat 5 (backend-inconsistent metadata), severity MEDIUM.

brendancol:
Closes #2525.

## Summary

- `_hi_dask_cupy` previously converted cupy chunks to numpy and then called `_hi_dask_numpy`, returning a dask+numpy result for dask+cupy input. Downstream `ArrayTypeFunctionMapping` dispatch silently routed through the numpy path.
- After this change, the painted output is re-wrapped through `map_blocks(cupy.asarray, ...)` so each chunk lands back as a cupy array. The dask graph stays lazy and `is_dask_cupy(result)` is True for dask+cupy inputs.
- The numpy reduce in the middle is kept as-is. The per-zone HI lookup is a Python dict so the reduce must pass through host memory; only the painted output needs to match the input backend.

## Backend coverage

- numpy: unchanged, returns numpy ndarray
- cupy: unchanged, returns cupy ndarray
- dask+numpy: unchanged, returns dask Array with numpy chunks
- dask+cupy: now returns dask Array with cupy chunks (was: numpy chunks)

## Test plan

- [x] New parametrized regression test `test_hi_preserves_backend_2525` asserts the output backend matches the input across all four backends.
- [x] Existing `test_hypsometric_integral.py` suite still passes (33 tests).
- [x] Existing `test_zonal.py` suite still passes (125 tests).
- [x] Verified the new test fails with the fix reverted (confirms it catches the regression).

## Audit trail

Surfaced by `/sweep-metadata` on the zonal module, 2026-05-27. Cat 5 (backend-inconsistent metadata), severity MEDIUM.

### DIFF
--- a/xrspatial/tests/test_hypsometric_integral.py
+++ b/xrspatial/tests/test_hypsometric_integral.py
@@ -254,3 +254,66 @@ def test_hypsometric_integral_list_of_pairs_zones():
     result = hypsometric_integral(zones_pairs, values, nodata=0)
     assert isinstance(result, xr.DataArray)
     assert result.shape == values.shape
+
+
+# --- backend-consistency regression (#2525) ---------------------------------
+
+def _has_cuda_and_cupy():
+    try:
+        from numba import cuda
+        import cupy  # noqa: F401
+        return cuda.is_available()
+    except Exception:
+        return False
+
+
+@pytest.mark.parametrize("backend", ['numpy', 'cupy', 'dask+numpy', 'dask+cupy'])
+def test_hi_preserves_backend_2525(backend):
+    """Regression: output backend must match input backend.
+
+    Before the fix, `_hi_dask_cupy` converted chunks to numpy and called
+    `_hi_dask_numpy`, returning a dask+numpy array even when the input was
+    dask+cupy. Downstream dispatch via ArrayTypeFunctionMapping then routed
+    through the numpy path silently. See issue #2525.
+    """
+    if 'cupy' in backend and not _has_cuda_and_cupy():
+        pytest.skip("Requires CUDA and CuPy")
+    if 'dask' in backend and da is None:
+        pytest.skip("Requires Dask")
+
+    from xrspatial.zonal import hypsometric_integral
+    from xrspatial.utils import (
+        is_cupy_array, is_dask_cupy, has_dask_array,
+    )
+
+    zones_np = np.array([[1, 1, 2, 2],
+                         [1, 1, 2, 2]], dtype=np.int32)
+    values_np = np.array([[10.0, 20.0, 30.0, 40.0],
+                          [15.0, 25.0, 35.0, 45.0]])
+
+    zones = create_test_raster(zones_np, backend, dims=['y', 'x'],
+                               chunks=(2, 2))
+    values = create_test_raster(values_np, backend, dims=['y', 'x'],
+                                chunks=(2, 2))
+
+    result = hypsometric_integral(zones, values)
+    assert isinstance(result, xr.DataArray)
+    assert result.shape == values.shape
+
+    if backend == 'numpy':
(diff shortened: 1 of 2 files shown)


brendancol:
## Review summary

The fix is correct and minimal. `_hi_dask_cupy` now wraps the painted output via `map_blocks(cupy.asarray, ..., meta=cupy.array(()))`, mirroring the pattern in `_apply_dask_cupy` (zonal.py:1720, 1731). All four backends remain wired in `ArrayTypeFunctionMapping`, and the new parametrized test verifies both DataArray-level and chunk-level backend identity.

### Blockers
None.

### Suggestions
None blocking. The extra `map_blocks(cupy.asarray, ...)` adds one host-to-device copy per chunk, but this is unavoidable given that the per-zone lookup is a Python dict and must go through host memory anyway.

### Nits applied in follow-up commit
- Replaced the locally redeclared `_has_cuda_and_cupy()` with the shared `has_cuda_and_cupy` from `xrspatial.utils`, and swapped `da is None` for `has_dask_array()` to match the rest of the test suite.
- Dropped the em-dash in the `_hi_dask_cupy` docstring.

### What looks good
- Test asserts chunk-level type via `.blocks[0, 0].compute()`, not just the DataArray wrapper.
- PR body confirms the test fails when the fix is reverted.
- Docstring updated to explain why the round-trip through host memory exists.

### Pre-existing items (not from this PR)
- `xrspatial/zonal.py:39` F401 `not_implemented_func` unused and `xrspatial/zonal.py:455` E501 both exist on the merge base.

PR #2531, opened 2026-05-27
Raise on 'majority' in zonal_stats dask path instead of silent drop (#2528)

### LINKED ISSUE #2528: zonal_stats dask backend silently drops 'majority' from default stats
## Bug Description

Calling `zonal_stats()` on dask-backed inputs without specifying `stats_funcs` returns only seven columns (mean, max, min, sum, std, var, count) instead of the eight in the documented default list. The default list includes `majority`, but the dask path drops it without warning the caller. The numpy and cupy backends honour the full default list.

The same silent drop happens whenever a caller passes a list that includes `majority`. There is no error, no warning, and the output dataframe contains fewer columns than requested.

## Reproduction

```python
import numpy as np
import xarray as xr
import dask.array as da

zones_data = np.array([[1, 1, 2, 2], [1, 1, 2, 2]], dtype=float)
values_data = np.array([[1.0, 1, 2, 2], [3, 3, 5, 5]])

zones = xr.DataArray(da.from_array(zones_data, chunks=(2, 2)))
values = xr.DataArray(da.from_array(values_data, chunks=(2, 2)))

df = stats(zones=zones, values=values, stats_funcs=['mean', 'majority'])
print(df.columns.tolist())
# ['zone', 'mean']  -- 'majority' is silently dropped
```

## Expected

Either raise a `ValueError` pointing the caller at the supported dask stats list, or document that `majority` will be filtered out when running on dask. The current behaviour breaks the contract printed in the validation message at `xrspatial/zonal.py:849-853`, which says `stats_funcs` "must be a subset of default supported stats" -- yet the actual default list includes `majority`.

## Suggested Fix

Raise `ValueError` listing the supported dask stats whenever a caller explicitly passes `majority` on a dask backend. For the bare `stats(zones, values)` call with no `stats_funcs`, filter `majority` from the default before validation so the call still succeeds but the documented contract holds.

## Backends Affected

- `dask+numpy`: silently drops
- `dask+cupy`: silently drops (delegates to dask+numpy)
- `numpy`: returns majority correctly
- `cupy`: returns majority correctly

## Categories

- Cat 5: backend inconsistency between numpy/cupy and dask paths.

brendancol:
## Summary

- zonal_stats() on dask-backed input silently dropped 'majority' from the requested stats list because the default included it but the dask block-by-block path could not compute it.
- Replaced the mutable list default with ``stats_funcs=None`` and resolved the default per backend; numpy/cupy keep the full eight-stat default, dask resolves to the seven-stat subset.
- Explicit 'majority' on a dask input now raises ``ValueError`` instead of being silently filtered.

Closes #2528.

## Test Plan

- [x] ``pytest xrspatial/tests/test_zonal.py`` -- 129 pass (125 pre-existing + 4 new regression tests).
- [x] ``pytest xrspatial/tests/test_dasymetric.py`` -- 61 pass (dasymetric uses zonal.stats internally).
- [x] Cross-backend smoke test confirms numpy/cupy keep 'majority', dask backends omit it, explicit majority on dask raises with a clear message.


brendancol:
Review pass (worktree, fresh test run):

- (a) ValueError vs deprecation: ValueError is appropriate. There is no documented contract that majority worked on dask; the prior behaviour silently dropped it. Callers who explicitly passed 'majority' on dask were already broken in a worse way (silent data loss). No deprecation needed.
- (b) Empty list stats_funcs=[]: behaves consistently across backends; returns a frame with only the 'zone' column. The new validation does not flag it (empty unsupported list). Not adding a dedicated test, the existing subset tests cover the surrounding code path.
- (c) Docstring stats table: the per-backend default lists are now documented and the dask majority rejection is called out.
- (d) Other callers: `dasymetric.py` references zonal.stats only in docstrings; `accessor.zonal_stats` passes kwargs through; the Dataset branch recurses with the same stats_funcs (None resolves per-backend on each per-variable call). No other module passes stats_funcs explicitly.

Applied one suggestion: the dict-on-dask error message hard-coded the supported list. Sourced it from _DEFAULT_STATS_DASK to prevent drift (1 file, 2 lines).

Tests: 129/129 test_zonal.py + 150/150 dasymetric/dataset_support/validation pass.

PR #2532, opened 2026-05-27
zonal: rechunk apply 3D dask output along stacked axis (#2526)

### LINKED ISSUE #2526: zonal.apply 3D dask path leaves axis-2 chunked at size 1 after da.stack
## Summary

`xrspatial.zonal.apply` on a 3D dask-backed `values` array calls
`da.stack(layers, axis=2)` to assemble the per-layer `map_blocks`
results. `da.stack` produces unit chunks along the new axis, and the
function never `rechunk()`s afterward, so the output has `chunks[2] =
(1, 1, ..., 1)` regardless of the input chunking.

Confirmed by graph probe (256x256 raster, chunks=(64,64), 3 bands):

```
"apply_dask_3d": {
  "task_count": 176,
  "tasks_per_chunk": 11.0,
  "output_chunks_axis2": [1, 1, 1]
}
```

This is the documented anti-pattern from CLAUDE.md:

> `da.stack` creates chunks of size 1 along the new axis — use
> `.rechunk({0: N})` to merge

## Impact

- Downstream operations that touch all bands together pay scheduling
  overhead proportional to the band count rather than seeing one chunk.
- Task count inflates linearly with band count.
- Both `_apply_dask_numpy` (`zonal.py:1691`) and `_apply_dask_cupy`
  (`zonal.py:1731`) have the same pattern.

## Fix

Rechunk the stacked output to merge unit chunks along the new axis.

## How to reproduce

```python
import dask.array as da
import numpy as np
import xarray as xr
from xrspatial.zonal import apply

shape = (256, 256, 3)
z = xr.DataArray(
    da.from_array(np.zeros(shape[:2], dtype=np.int64), chunks=(64, 64)),
    dims=['y', 'x'],
)
v = xr.DataArray(
    da.from_array(np.zeros(shape), chunks=(64, 64, 3)),
    dims=['y', 'x', 'band'],
)
out = apply(z, v, lambda x: x * 2)
print(out.data.chunks[2])  # (1, 1, 1) -- bug
```

brendancol:
## Summary

- `zonal.apply` on a 3D dask `values` array left output chunks at size 1 along the stacked axis because `da.stack(layers, axis=2)` was not followed by `rechunk()`.
- Rechunk the stacked output back to `values_data.chunks[2]` in both `_apply_dask_numpy` and `_apply_dask_cupy`.
- Adds regression `test_apply_dask_3d_axis2_rechunked_2526` and updates the sweep-performance state row for `zonal`.

Fixes #2526.

## Background

Graph probe on a 256x256 raster with `chunks=(64, 64, 3)` and 3 bands previously yielded:

```
"apply_dask_3d": {
  "task_count": 176,
  "tasks_per_chunk": 11.0,
  "output_chunks_axis2": [1, 1, 1]
}
```

The documented anti-pattern from CLAUDE.md is `da.stack` without a following `rechunk` — see existing zonal `hypsometric_integral` audit notes that called out the same shape.

## Test plan

- [x] `pytest xrspatial/tests/test_zonal.py` (126 passed)
- [x] New regression test asserts `result.data.chunks[2] == values.data.chunks[2]`
- [x] Existing `test_apply_3d[dask+numpy]` still passes


brendancol:
## Review

Targeted fix, low risk. The `da.stack` -> `rechunk({2: ...})` pattern is the documented remediation for unit-chunk-along-stack-axis (also referenced in the in-tree `hypsometric_integral` audit).

### Verified
- `rechunk({2: values_data.chunks[2]})` only touches axis 2; axes 0/1 keep `zones_data.chunks`.
- `meta=np.array(())` / `meta=cupy.array(())` from `map_blocks` propagates through `da.stack` and `rechunk`, so dtype/backend metadata is preserved.
- N=1 band edge case: rechunk target becomes `(1,)`, which matches what `da.stack` already produces - no-op, harmless.
- Both `_apply_dask_numpy` (line 1691-1694) and `_apply_dask_cupy` (line 1734-1737) get the fix.
- `pytest xrspatial/tests/test_zonal.py` -> 126 passed locally.

### Notes (not blocking)
- Regression test exercises `_apply_dask_numpy` only. The cupy path uses the same construction so the fix is symmetric, and this matches the existing `test_apply_3d` parametrization (`['numpy', 'dask+numpy']`). cupy parity test can land separately when the cupy CI runner exercises it.
- No benchmark added; existing zonal benchmark already covers `apply`.

Looks ready to merge.

Issue #1296, opened 2026-04-28
sieve(): numpy and cupy backends have no memory guard


brendancol:
## Description

`sieve()` on the numpy and cupy backends has no memory guard. `_label_connected` allocates parent, rank, and region_map_flat as int32 arrays plus a float64 result copy — about 20 bytes/pixel of working memory — before any check runs.

The dask paths guard via `_available_memory_bytes()` (lines 343-355, 366-381). The public `sieve()` API at line 489 dispatches numpy DataArrays straight into `_sieve_numpy` with no check. `_sieve_cupy` at line 308 calls `data.get()` then `_sieve_numpy`, inheriting the gap.

A 50000x50000 numpy raster asks for ~50 GB of host memory before anything errors out.

Same asymmetric-guard pattern already fixed in cost_distance #1262, mahalanobis #1288, multispectral #1291, and kde #1287.

## Expected behavior

`sieve()` raises `MemoryError` with a clear message on every backend when the projected working set exceeds available memory, matching the existing `_sieve_dask` behavior.

## Proposed fix

Add `_check_memory(rows, cols)` and `_check_gpu_memory(rows, cols)` helpers (28 bytes/pixel, 50% threshold) and call them from `_sieve_numpy` and `_sieve_cupy` before the union-find allocations run.

## Followup (separate)

The int32 indices in `_label_connected` silently truncate when n > 2^31 (rasters above ~46340x46340). The docstring notes the limit but nothing enforces it at runtime. The memory guard rejects rasters that large before the int32 issue triggers, so this is a documentation/clarity follow-up rather than an exploitable bug.

Issue #1291, opened 2026-04-27
true_color allocates ~17 bytes/pixel of working memory with no guard


brendancol:
## Description

`true_color()` materializes a (H, W, 4) uint8 output cube plus three intermediate (H, W) float32 normalize buffers and an alpha array on the eager (numpy and cupy) backends with no memory check. Peak working memory is roughly 17 bytes per pixel.

A 50000x50000 input raster asks for about 42 GB of RAM (or device VRAM on the GPU path), which surfaces as an opaque allocator error rather than a clean `MemoryError` that points at the input shape.

## File and lines

- `xrspatial/multispectral.py`
  - `_true_color_numpy` line 1602: `out = np.zeros((h, w, 4), dtype=np.uint8)` plus three `_normalize_data` calls that each allocate a full-shape float32 buffer.
  - `_true_color_cupy` line 1631: same pattern via `cupy.empty`/`cupy.full`.
- The dask paths (`_true_color_dask`, `_true_color_dask_cupy`) build the result lazily through `da.stack` and `map_blocks`, so they are not affected.

## Severity

HIGH (Cat 1, unbounded stacked-output allocation). Found during the security sweep of the `multispectral` module.

## Proposed fix

Same shape as the guards already in `dasymetric.py`, `cost_distance.py`, and `diffusion.py`:

- Module-local `_available_memory_bytes()` helper (try `/proc/meminfo`, then `psutil`, then a 2 GB default).
- `_check_memory(h, w)` raises `MemoryError` before the first allocation when the 17 bytes/pixel footprint exceeds 50% of available RAM.
- `_check_gpu_memory(h, w)` uses `cupy.cuda.runtime.memGetInfo()` and applies the same 50% budget on free VRAM.
- Wire `_check_memory` into `_true_color_numpy` and `_check_gpu_memory` into `_true_color_cupy` before the first allocation. Dask paths are skipped because they build the cube lazily.

The error message should hint at reducing the raster size or rechunking a dask-backed input.

## Tests

Monkeypatch the new `_available_memory_bytes` helper to return a small value and assert that `MemoryError` is raised before `np.zeros`/`cupy.empty` runs, on a small input. Don't actually allocate large rasters.

## Out of scope (follow-up)

[...]

Issue #1288, opened 2026-04-27
mahalanobis: unbounded float64 allocations on numpy and cupy backends


brendancol:
## Describe the bug

`mahalanobis()` in `xrspatial/mahalanobis.py` has no memory guard on its working-set allocations. Every eager backend stacks `n_bands` rasters into a float64 cube before any further work:

- `_compute_stats_numpy` (line 45): `np.stack([b.astype(np.float64) for b in bands_data], axis=0)`
- `_compute_stats_cupy` (line 80): same on GPU
- `_mahalanobis_pixel_numpy` (line 184): `flat = stacked.reshape(n_bands, -1).T` followed by `diff @ inv_cov`, both of which materialise float64 copies of the same size

The covariance and inverse-covariance matrices are tiny (`n_bands` by `n_bands`), so the dominant cost is `n_bands * H * W * 8` bytes, repeated 3 to 4 times across the working set.

## Reproducer

```python
import numpy as np
import xarray as xr
from xrspatial.mahalanobis import mahalanobis

# 100k x 100k raster, 5 bands -> ~400 GB just for the np.stack
shape = (100_000, 100_000)
bands = [
    xr.DataArray(np.zeros(shape, dtype=np.float32), dims=['y', 'x'])
    for _ in range(5)
]
mahalanobis(bands)  # OOM the process
```

## Expected behavior

The function should reject inputs whose projected working-set exceeds available memory with a `MemoryError` that names the offending dimensions, the same way `bump`, `cost_distance`, `dasymetric`, and `glcm_texture` do after the earlier security-sweep PRs.

## Severity

HIGH (Cat 1, unbounded allocation / DoS)

## Proposed fix

1. Add a local `_available_memory_bytes()` helper at the top of `xrspatial/mahalanobis.py`, mirroring `classify.py:38-52` (try `/proc/meminfo`, fall back to `psutil`, fall back to 2 GB).
2. Add `_check_memory(bands_data)` that estimates `~4 * n_bands * H * W * 8` bytes for the stack, reshape-transpose, centered, and diff-transformed working set, and raises `MemoryError` when the estimate exceeds 50% of available memory.
3. For the cupy path, query free GPU memory via `cupy.cuda.runtime.memGetInfo()` (mirroring `cost_distance.py:261`) and apply the same 50% budget.
4. Skip the guard on the dask paths since chunks are bounded by the user.

## Out of scope (separate follow-up)

[...]

PR #2533, opened 2026-05-27
zonal: guard unbounded allocation in stats(return_type='xarray.DataArray') (#2523)

### LINKED ISSUE #2523: zonal.stats: unbounded allocation with return_type='xarray.DataArray'
## Security finding (HIGH, Cat 1 — Unbounded Allocation / DoS)

`xrspatial.zonal._stats_numpy` allocates an N-stat result array sized by user-controlled dimensions without a memory check:

https://github.com/makepath/xarray-spatial-contrib/blob/main/xrspatial/zonal.py#L505

```python
else:
    result = np.full((len(stats_funcs), values.size), np.nan)
```

This branch runs whenever the public `stats(..., return_type='xarray.DataArray')` is called. Memory required is `n_stats * H * W * 8` bytes (float64), with `n_stats` user-controlled via the `stats_funcs` dict. For the documented default 8 stats on a 20000x20000 input the working buffer is ~25.6 GB; with a larger custom `stats_funcs` dict and/or input it scales linearly with both.

This mirrors the unbounded-allocation pattern already fixed in cost_distance (#1262), mahalanobis (#1288), kde (#1287), multispectral (#1291), sieve (#1296), resample (#1295), and several other modules.

## Reproducer

```python
import numpy as np
import xarray as xr

# Allocates n_stats * H * W * 8 bytes with no guard
zones = xr.DataArray(np.zeros((20000, 20000), dtype=np.int32))
values = xr.DataArray(np.zeros((20000, 20000), dtype=np.float32))
out = stats(zones=zones, values=values, return_type='xarray.DataArray')
# Working buffer >= 25 GB on top of the input rasters
```

## Proposed fix

Add an `_available_memory_bytes()` / `_check_memory(n_stats, h, w)` helper to `zonal.py` and call it from `_stats_numpy` before the `np.full` allocation when `return_type == 'xarray.DataArray'`. Mirrors the established pattern (raise `MemoryError` with a clear message when `n_stats * H * W * 8` exceeds 50% of available RAM). Dask / cupy paths return a DataFrame and are not affected; the cupy path already raises `TypeError` for 3D input.

Found by deep-sweep-security-zonal-2026-05-27 (carry-over MEDIUM from 2026-04-22 audit, re-classified HIGH given the same severity bar used for the related per-module unbounded-allocation fixes).

brendancol:
## Summary

- `_stats_numpy` xarray.DataArray return path allocated `np.full((n_stats, values.size), nan)` (float64) with no memory check; `n_stats` is user-controlled via `stats_funcs` and `values.size` scales with the input raster.
- Added `_check_stats_dataarray_memory(n_stats, values_shape)` that reuses the existing `_available_memory_bytes()` helper and raises `MemoryError` before the allocation when the buffer would exceed 50% of available RAM.
- Follows the same pattern as the unbounded-allocation guards in cost_distance (#1262), kde (#1287), mahalanobis (#1288), multispectral (#1291), resample (#1295), sieve (#1296), etc.

Closes #2523. Found by deep-sweep-security-zonal-2026-05-27.

## Test plan

- [x] New unit test `test_stats_dataarray_return_type_memory_guard_2523` mocks `_available_memory_bytes` and asserts both the guard fires and the normal path still produces a correct `xr.DataArray`
- [x] Full `xrspatial/tests/test_zonal.py` suite passes (126 tests)


brendancol:
## PR Review: zonal: guard unbounded allocation in stats(return_type='xarray.DataArray')

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/zonal.py:2025-2027`: `n_cells = 1; for s in values_shape: n_cells *= int(s)` works fine, but `int(np.prod(values_shape))` would match the style of `_regions_dask` (line 2047). Cosmetic.

### What looks good
- Guard runs before the `np.full(...)` call on line 506, so it fires before the allocation, not after.
- Reuses the existing `_available_memory_bytes()` helper and the `required > 0.5 * available` threshold. Matches sieve.py, mahalanobis.py, dasymetric.py, and `_regions_dask`.
- Error message names the call site (`stats(return_type='xarray.DataArray')`), reports required vs available GB, and lists three concrete mitigations: reduce `stats_funcs`, use a smaller raster, or switch to `return_type='pandas.DataFrame'`.
- Regression test mocks `_available_memory_bytes` to a 100-byte budget, asserts the guard fires, and verifies the normal path still produces a correct `xr.DataArray`.
- Other backends do not hit this code path. `_stats_cupy` and the dask backends always return a DataFrame, and the top-level `stats()` only assembles the xarray result after the numpy backend has returned the `(n_stats, H, W)` ndarray, so the guard sits in the right place.
- Sweep state CSV updated.

### Checklist
- [x] Guard fires before the allocation
- [x] Error message is specific and actionable
- [x] Threshold pattern matches sibling modules (cost_distance, kde, mahalanobis, dasymetric, sieve)
- [x] Regression test with synthetic limit
- [x] `pytest xrspatial/tests/test_zonal.py -x -q` passes (126/126)

PR #2548, opened 2026-05-28
resample: close test-coverage gaps surfaced by deep-sweep audit

brendancol:
## Summary

Test-only PR closing test-coverage gaps in `xrspatial.resample`
surfaced by `/deep-sweep test-coverage` on 2026-05-27. Adds
`test_resample_coverage_2026_05_27.py` with 70 tests (68 passing,
2 skipped on the documented dask cubic Nx1 limitation tracked in
#2547).

Gaps closed:

- **Cat 3 HIGH:** Nx1 single-column input across all four backends
  and all eight methods. Existing tests covered 1x1 and 1xN strips
  but never width=1 with multiple rows.
- **Cat 2 MEDIUM:** NaN parity for `cupy` and `dask+cupy`. Existing
  parity tests used random data without NaNs, so the weight-mask
  gate and spline-prepad GPU paths had no NaN coverage.
- **Cat 3 MEDIUM:** all-equal-value input across methods (zero
  variance / zero gradient).
- **Cat 5 MEDIUM:** non-default dim names (`lat`/`lon`,
  `latitude`/`longitude`, `channel`/`lat`/`lon` 3D) propagate
  without being renamed to `y`/`x`; per-dim `units` attrs
  round-trip.
- **Cat 3 MEDIUM:** empty raster (0 rows / 0 cols) behaviour pin.

Source untouched. Filed #2547 for the cubic-on-dask Nx1 source bug
the audit surfaced; the two skipped tests gate on that fix.

## Test plan

- [x] `pytest xrspatial/tests/test_resample_coverage_2026_05_27.py`
  -> 68 passed, 2 skipped
- [x] `pytest xrspatial/tests/test_resample.py
  xrspatial/tests/test_resample_coverage_2026_05_27.py` -> 237
  passed, 2 skipped (no regressions in existing suite)

PR #2549, opened 2026-05-28
resample: preserve scalar non-dim coords and refresh stale nodata attr (#2542)

### LINKED ISSUE #2542: resample: drops non-dim coords (spatial_ref) and leaves stale attrs['nodata'] on 2D path
## Summary

`xrspatial.resample()` has two metadata-propagation gaps on the 2D non-identity code path that combine to corrupt chained pipelines built on rioxarray conventions.

### 1. Non-dimension coords are silently dropped (Cat 2)

The 2D path rebuilds the output `xr.DataArray` with only the spatial dim-coords:

```python
result = xr.DataArray(
    result_data,
    name=name,
    dims=agg.dims,
    coords={ydim: new_y, xdim: new_x},
    attrs=new_attrs,
)
```

Any extra non-dim coord on the input -- most notably rioxarray's `spatial_ref`
scalar coord, but also scalar `time` / `band` selectors -- is gone from the output.

A user pipeline like:

```python
agg = rioxarray.open_rasterio('dem.tif').squeeze()
out = resample(agg, scale_factor=0.5)
out.rio.crs  # -> None (was EPSG:32633)
```

silently loses the CRS even though `attrs['crs']` round-trips fine.

The identity path (`scale_factor=1.0`) uses `agg.copy()` and preserves the
coord, and the 3D path (per-band recursion + `xr.concat`) also happens to
preserve it. Only the 2D non-identity path is broken, so the bug is also
path-inconsistent (Cat 5).

### 2. `attrs['nodata']` is read but never refreshed (Cat 4)

`_resolve_nodata()` reads the nodata sentinel from three sources:

```python
for key in ('_FillValue', 'nodata'):
    v = agg.attrs.get(key)
    if v is not None:
        nodata = v
        break
```

After resampling, sentinel pixels are replaced with NaN. The output post-processing
refreshes `_FillValue` and `nodatavals` to NaN but leaves `attrs['nodata']` at its
old finite value:

```
Input  attrs: {'nodata': -9999, 'res': (1.0, 1.0)}
Output attrs: {'nodata': -9999, '_FillValue': nan, 'res': (2.0, 2.0)}
```

A downstream consumer that trusts `attrs['nodata']` will mask the wrong cells.

## Repro

```python
import numpy as np
import xarray as xr

# Bug 1: spatial_ref dropped
agg = xr.DataArray(
    np.arange(16, dtype=np.float32).reshape(4, 4),
    dims=('y', 'x'),
    coords={'y': np.linspace(3, 0, 4),
            'x': np.linspace(0, 3, 4),
            'spatial_ref': 0},
    attrs={'res': (1.0, 1.0), 'crs': 'EPSG:4326'},
)
out = resample(agg, scale_factor=0.5)
assert 'spatial_ref' not in out.coords  # bug

# Bug 2: stale nodata attr
data = np.array([[-9999, -9999, 10, 10],
                 [-9999, -9999, 10, 10],
                 [20, 20, 30, 30],
                 [20, 20, 30, 30]], dtype=np.float32)
agg = xr.DataArray(
    data, dims=('y', 'x'),
    coords={'y': np.linspace(3, 0, 4), 'x': np.linspace(0, 3, 4)},
    attrs={'res': (1.0, 1.0), 'nodata': -9999},
)
out = resample(agg, scale_factor=0.5, method='nearest')
assert out.attrs['nodata'] == -9999  # bug; data now has NaN there
```

## Proposed fix

In the 2D non-identity path in `resample()`:

1. After building the new dim-coord dict, fold in input non-dim coords whose
   dims are a subset of `(ydim, xdim)` or empty (scalar coords) so things
   like `spatial_ref`, scalar `time`, and scalar `band` round-trip.
2. When `has_nodata` is true, also refresh `attrs['nodata']` to NaN (matching
   the existing `_FillValue` / `nodatavals` refresh).

This brings the 2D path in line with the identity path's coord preservation
and closes the `attrs['nodata']` half of the sentinel refresh.

## Backend coverage

Fix lives in the single `xr.DataArray` constructor at the end of `resample()`
so it applies uniformly to numpy / cupy / dask+numpy / dask+cupy.

## Discovered via

Deep-sweep metadata propagation audit (sweep-metadata, 2026-05-27).

brendancol:
## Summary

- The 2D non-identity path in `resample()` rebuilt its output `xr.DataArray` with only the spatial dim-coords, silently dropping any scalar non-dim coord on the input (e.g. rioxarray's `spatial_ref`, or a squeezed `time` / `band` selector). The identity path (`scale_factor=1.0`) and the 3D per-band path both kept these coords, so the 2D path was the inconsistent outlier (Cat 5).
- `_resolve_nodata()` reads `attrs['nodata']` as a sentinel fallback, but the output post-processing only refreshed `_FillValue` and `nodatavals`. Inputs declaring `attrs['nodata'] = -9999` came back with that same finite value alongside data that was now NaN (Cat 4).
- Fixed both gaps in one place at the final `xr.DataArray` constructor. Spatially-shaped non-dim coords intentionally do not carry across; their length is tied to the input resolution.

## Backend coverage

numpy / cupy / dask+numpy / dask+cupy all route through the same constructor. Verified end-to-end for the `spatial_ref` carry across all four backends. Nodata-attr refresh verified on numpy / cupy / dask+numpy; dask+cupy with a non-NaN sentinel hits a pre-existing `xr.where` + `cupy.astype` quirk unrelated to this audit.

## Test plan

- [x] 175 tests pass in `xrspatial/tests/test_resample.py` (168 existing + 7 new)
- [x] New `TestMetadataPropagation` cases: nodata-attr refresh, spatial_ref carry, scalar band/time carry, identity-vs-downsample coord parity, intentional drop of spatially-shaped extras
- [x] 4-backend spatial_ref parity confirmed manually
- [x] Discovered via deep-sweep metadata audit (sweep-metadata, 2026-05-27)

Closes #2542


brendancol:
## PR Review: preserve scalar non-dim coords and refresh stale nodata attr

Read source (`resample.py:1300-1424`) and tests in full from the worktree at the PR head commit. The fix lives at the single 2D output constructor, so all 4 backends inherit it without per-backend changes.

### Blockers
None.

### Suggestions
- `xrspatial/tests/test_resample.py:203-308` — the PR body claims 4-backend parity for `spatial_ref` carry was verified manually, but no automated test pins this. Since the fix lives at the shared `xr.DataArray` constructor, a single `@cuda_and_cupy_available` test asserting `spatial_ref` survives a cupy resample would catch a backend-divergent regression. The companion PR (#2548) already has the cross-backend scaffolding to copy from.
- `xrspatial/resample.py:1404-1412` — the 3D path (lines 1290-1322) hands each band to the 2D `resample()` call, so this fix transitively propagates `spatial_ref` through `xr.concat`. Worth one explicit test confirming the 3D `spatial_ref` case so the contract is pinned, not just inferred.

### Nits
- `xrspatial/resample.py:1418` — `coords={ydim: new_y, xdim: new_x, **extra_coords}` relies on the `coord_name in (ydim, xdim)` skip at line 1409 to keep `extra_coords` from shadowing the rebuilt spatial coords. The current filter prevents collision; the safety is one-sided though (if the filter ever loosens, `**extra_coords` would silently override).
- `xrspatial/tests/test_resample.py:228,247` — `int(out.coords['spatial_ref'].values) == 0` and `int(out.coords['band'].values) == 1` cast the scalar coord to int. Works for the values used; `== 0` / `== 1` against the np scalar is one less coercion to read.

### What looks good
- Fix is at the single shared 2D output constructor, so all four backends pick it up without per-backend code.
- Comments on lines 1387-1393 and 1396-1402 explain the *why*: which input attrs feed `_resolve_nodata`, why the identity path was the consistent outlier, why spatially-shaped coords are intentionally dropped.
- Test names mirror the audit categories (`test_nodata_attr_refreshed_to_nan`, `test_identity_and_downsample_coords_consistent`), so a failure maps straight back to the issue.
- `test_nodata_attr_refreshed_to_nan` pins both the attr refresh AND that the corresponding data pixel is NaN — catches the exact "metadata says NaN but data still has -9999" failure mode that motivated the fix.
- `test_2d_spatial_extra_coord_not_carried` explicitly documents the chosen scope: scalar coords carry, spatially-shaped coords don't.

### Checklist
- [x] Behavior matches issue #2542
- [x] All implemented backends share the fix at the constructor seam
- [x] NaN handling correct (nodata attr now matches data state)
- [x] Edge cases covered (nodata absent stays absent, 2D spatial coord drops, identity vs downsample parity)

PR #2556, opened 2026-05-28
geotiff: add bbox= to open_geotiff for geographic-space windowed reads

### LINKED ISSUE #2555: geotiff: add bbox= parameter to open_geotiff for geographic-space window reads
## Reason or Problem

`open_geotiff(window=...)` only accepts a pixel window `(row_start, col_start, row_stop, col_stop)`. Most geospatial users think in data-space coordinates: a bounding box in the file's CRS (`x_min, y_min, x_max, y_max`). Doing the conversion by hand means reading the file's transform, running the affine math, and clamping to file bounds. We already do exactly that internally via `_extent_to_window` in `xrspatial/geotiff/_attrs.py:912`, but it is not exposed on the public API.

A related gap: the `PixelSafetyLimitError` recovery hint that #2553 added mentions `window=` and `chunks=` but not a geographic bbox option. Once `bbox=` is on `open_geotiff`, the hint should mention it too so users with a known area of interest reach for the right kwarg.

## Proposal

Add `bbox=(x_min, y_min, x_max, y_max)` to `open_geotiff`:

- Geographic coordinates in the file's CRS.
- Mutually exclusive with `window=`. Passing both raises ValueError.
- Resolved internally via `_read_geo_info(source, overview_level=...)` + `_extent_to_window(geo_info.transform, h, w, y_min, y_max, x_min, x_max)`. The result is forwarded to the existing backend dispatch as `window=`.
- Requires the source to be georeferenced. A file without georef, or with a rotated affine that has not been cleared via `allow_rotated`, raises a clear ValueError naming the limitation.

Then update the `PixelSafetyLimitError` recovery hint in `xrspatial/geotiff/_layout.py:_recovery_hint` to add a fourth bullet:

```
  * Read a geographic sub-region with bbox=(x_min, y_min, x_max, y_max).
```

**Design:**

`open_geotiff` runs the bbox->window conversion before fanning out to backends. The dispatcher is the right place since it already coerces the source path once and dispatches once. Backends (`read_geotiff_dask`, `read_geotiff_gpu`, `read_vrt`, `read_to_array`) keep their existing `window=` surface; no new plumbing.

The conversion uses `_read_geo_info`, which already supports local files, BytesIO, HTTP, and fsspec URIs via a header-only read. For HTTP and cloud sources this is a small range request, not a full download, and the bytes are typically cached for the subsequent main read.

**Usage:**

```python
arr = open_geotiff(
    "s3://bucket/large-dem.tif",
    bbox=(-122.5, 37.6, -122.3, 37.8),  # San Francisco area, WGS84
)
```

**Value:** Removes the most common workaround in user code (manual affine math). Mirrors how rasterio's `windows.from_bounds` is used.

## Stakeholders and Impacts

Anyone reading sub-regions of large GeoTIFFs by geographic coordinates. The new parameter is opt-in; existing `window=` callers see no behaviour change. The error hint added in #2554 expands by one line.

## Drawbacks

The bbox->window resolution requires a metadata read before the main read. `_read_geo_info` is O(1) memory and fast for local files; for HTTP it adds one range request that is usually cached for the subsequent read.

Rotated transforms are rejected unless `allow_rotated=True` has cleared the rotation, since `_extent_to_window` assumes an axis-aligned grid. Same restriction the existing read path applies.

## Alternatives

Exposing `_extent_to_window` as public API would work but pushes the pixel math onto every caller. The whole point of adding `bbox=` is to remove that step.

brendancol:
## Summary

- `open_geotiff(..., bbox=(x_min, y_min, x_max, y_max))` reads a sub-region by geographic coordinates in the file's CRS.
- Resolves to a pixel `window=` via the existing `_read_geo_info` + `_extent_to_window` helpers. That makes the conversion one header-only metadata read, forwarded through the existing backend dispatch with no new plumbing.
- Mutually exclusive with `window=`. Requires a georeferenced, axis-aligned source.
- The `PixelSafetyLimitError` recovery hint (added in #2554) gains a fourth bullet: `Read a geographic sub-region with bbox=(x_min, y_min, x_max, y_max) in the file's CRS.`

Closes #2555.

## Backend coverage

CPU dispatcher change. numpy / cupy / dask+numpy / dask+cupy / VRT all reuse the existing `window=` surface since `bbox` is resolved at the `open_geotiff` entry. No backend-specific code paths added.

## Test plan

- [x] New `test_bbox_2555.py` covers: bbox vs window equivalence, full-bbox round-trip, extent clamping, mutual exclusion with `window=`, malformed `bbox` (wrong arity, non-sequence, inverted x/y), out-of-extent rejection, and missing-georef rejection.
- [x] The safety-error recovery hint test asserts the new `bbox=` line appears alongside the existing `window=` and `chunks=` lines.
- [x] `bbox` slots into `_CANONICAL_ORDER` immediately after `window`; the canonical-order signature test passes.
- [x] Full `xrspatial/geotiff/tests/` suite: 6005 passed, 81 skipped, 1 xfailed.


brendancol (commented):
## PR Review: geotiff: add bbox= to open_geotiff for geographic-space windowed reads

None.

- [ ] `xrspatial/geotiff/__init__.py:399`. NaN coordinates in `bbox` slip past the `x_min >= x_max` and `y_min >= y_max` guards because `NaN >= NaN` is False. They then land in `_extent_to_window`, where the affine math produces NaN row/col values and the integer cast fails with a confusing `ValueError`. Add an `np.isfinite` check after unpacking and raise a clear bbox-malformed error.

- [ ] Missing coverage: `overview_level` together with `bbox=`. The helper takes `overview_level` and forwards it to `_read_geo_info`, which selects the overview IFD whose transform has different `pixel_width` / `pixel_height` than the base. A test confirming that `open_geotiff(path, bbox=..., overview_level=1)` produces a pixel window appropriate for the overview's transform (not the base IFD's) would catch a future refactor that drops the plumbing.

- [ ] `xrspatial/geotiff/__init__.py:418`. The rotated-affine rejection branch (`geo_info.transform.rotated_affine is not None`) is untested. Constructing a rotated GeoTIFF takes a few lines via `write(..., geo_transform=GeoTransform(rotated_affine=...))`. Optional because the branch is straightforward and the message is clear.

- [ ] `xrspatial/geotiff/tests/read/test_bbox_2555.py:39`. The fixture uses `np.arange(...).reshape(100, 100)` for pixel values, which is fine. Consider asserting the bbox slice's *values* match the equivalent slice of the full read (you already do this for shape and coords; pixel-value parity catches row/col swaps too).

### What looks good

- Resolution happens once at the dispatcher (`open_geotiff`), so all four backends (eager / dask / GPU / VRT) reuse the existing `window=` path without learning a new kwarg.
- `_read_geo_info` is the right primitive: O(1) memory, handles local / BytesIO / HTTP / fsspec uniformly.
- Mutual exclusion with `window=` is explicit and raises before any I/O.
- The rotated-affine path routes users toward the existing `allow_rotated=True` recovery rather than silently producing wrong pixel windows.
- `_CANONICAL_ORDER` is updated in the same PR with a comment explaining why `bbox` sits next to `window`.
- The safety hint's bullet list stays parallel and readable; the `chunks=` line still appears last.

### Checklist
- [x] Algorithm matches reference (rasterio `windows.from_bounds` semantics). Clamp-to-extent matches.
- [x] All implemented backends produce consistent results. Single dispatcher resolution.
- [x] NaN handling. See Suggestion #1.
- [x] Edge cases covered for arity / type / inverted axes / non-overlap; not for NaN or overview_level.
- [x] README feature matrix updated. N/A.

brendancol (commented):
## Follow-up review pass (commit 9c2933b)

### Disposition

- **Suggestion #1 (NaN check)**: Fixed. `_bbox_to_window` now rejects non-finite bbox coordinates with a "must contain finite coordinates" error before the ordering check.
- **Suggestion #2 (overview_level coverage)**: Fixed. New `test_bbox_with_overview_level_uses_overview_transform` builds a 256x256 COG with 2x/4x reductions and confirms each successive level shrinks the bbox slice monotonically. Pins the overview-aware resolution against a future refactor that drops the plumbing.
- **Nit #1 (rotated-affine coverage)**: Fixed. New `test_bbox_rejects_rotated_affine` uses the existing `test_crs._write_rotated_tiff` helper to build a rotated source and asserts the targeted "rotated affine" error.
- **Nit #2 (pixel-value parity)**: Fixed. `test_bbox_matches_equivalent_pixel_window` now also asserts `arr.values == full.values[r0:r1, c0:c1]`, so a row/col swap in the resolver would surface.

### Incidental fix uncovered while writing the rotated test

`allow_rotated=True` drops the rotation AND sets `has_georef=False` on the resulting `GeoInfo`. The original guard order (`has_georef` first) meant rotated-dropped files raised the generic "no GeoTIFF tags" error instead of the rotated-specific message. Reordered so the `rotated_affine is not None` check runs first; the user now gets guidance to fall back to `window=` for pixel-space windowing.

### Tests

- `xrspatial/geotiff/tests/read/test_bbox_2555.py`: 15 passed.
- Full `xrspatial/geotiff/tests/` suite: 6009 passed, 81 skipped, 1 xfailed.

PR #2577, opened 2026-05-28
zonal: fix crosstab cat_ids overcount when earlier categories are filtered out

### LINKED ISSUE #2560: crosstab(cat_ids=[...]) overcounts when earlier categories are filtered out
**Describe the bug**

`crosstab(zones, values, cat_ids=[...])` returns inflated counts when the `cat_ids` filter excludes one or more categories that actually appear in the values raster.

The cause is in `_single_zone_crosstab_2d` in `xrspatial/zonal.py` (around line 959). The function builds cumulative break indices for every unique category via `_strides`, but `cat_start` is only advanced inside the `if cat in cat_ids:` branch. When an earlier category is filtered out, `cat_start` stays at 0, so the next selected category's count is computed as `zone_cat_breaks[j] - 0` instead of `zone_cat_breaks[j] - zone_cat_breaks[j-1]`.

The helper is shared across all four backends (numpy, dask+numpy, cupy, dask+cupy), so every backend produces the same wrong counts.

**Expected behavior**

The count for category `c` should equal the number of cells in the zone whose value is exactly `c`, regardless of which other categories are listed in `cat_ids`.

**Reproduction**

```python
from xrspatial import zonal

zones = xr.DataArray(np.ones((2, 2), dtype=int))
values = xr.DataArray(np.array([[1, 2], [2, 3]]))

print(zonal.crosstab(zones=zones, values=values, cat_ids=[2]))
# Returns count 3 for cat 2; correct answer is 2.

print(zonal.crosstab(zones=zones, values=values, cat_ids=[3]))
# Returns count 4 for cat 3; correct answer is 1.
```

Confirmed on the numpy backend; the same code path runs for dask, cupy, and dask+cupy.

**Fix direction**

Advance `cat_start` for every category in `unique_cats`, not only for selected ones. Equivalently, compute each category's count as `zone_cat_breaks[j] - (zone_cat_breaks[j-1] if j > 0 else 0)`.

Add cross-backend tests covering `cat_ids` that omit an earlier category.

brendancol:
## Summary

Closes #2560.

- Fixes the `_single_zone_crosstab_2d` helper in `xrspatial/zonal.py` so `cat_start` advances for every unique category, not only those selected via `cat_ids`. Previously, when `cat_ids` skipped a category that was present in the values raster, the cumulative cursor stayed behind and the next selected category's count absorbed the skipped cells.
- The helper is shared by all four backends (numpy, dask+numpy, cupy, dask+cupy), so the fix lands for every backend.
- Adds a cross-backend regression test `test_crosstab_2d_cat_ids_skips_earlier_category` covering both a single-cat selection that skips a leading category and a gapped selection that skips a middle category.

## Reproduction (before the fix)

```python
from xrspatial import zonal
zones = xr.DataArray(np.ones((2, 2), dtype=int))
values = xr.DataArray(np.array([[1, 2], [2, 3]]))
zonal.crosstab(zones=zones, values=values, cat_ids=[3])  # returned 4, should be 1
```

## Backend coverage

numpy / cupy / dask+numpy / dask+cupy. All four use the same `_single_zone_crosstab_2d` helper.

## Test plan

- [x] New regression test passes on all four backends locally
- [x] Existing crosstab tests still pass (16 selected, 16 passed)
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 
 #### Fixed
 
+- `crosstab(cat_ids=[...])` no longer overcounts when `cat_ids` skips a
+  category that is present in the values raster. The 2D helper now
+  advances its cumulative cursor for every unique category instead of
+  only the selected ones, so each selected category's count starts from
+  the correct offset. All four backends (numpy, dask+numpy, cupy,
+  dask+cupy) share the helper and are fixed together. (#2560)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/tests/test_zonal.py
+++ b/xrspatial/tests/test_zonal.py
@@ -1156,6 +1156,56 @@ def test_percentage_crosstab_2d(backend, data_zones, data_values_2d, result_perc
     assert_input_data_unmodified(data_values_2d, copied_data_values_2d)
 
 
+@pytest.mark.parametrize("backend", ['numpy', 'dask+numpy', 'cupy', 'dask+cupy'])
+def test_crosstab_2d_cat_ids_skips_earlier_category(backend, data_zones, data_values_2d):
+    """Regression test for issue #2560.
+
+    When cat_ids filters out a category that appears in the values raster,
+    the count for later selected categories must not be inflated by cells
+    that belong to the skipped category.
+
+    Zone 3 in the test fixture contains values [3, 3, 0, 3, 3] after nodata
+    filtering. Asking for cat_ids=[3] alone (skipping 0) should report 4
+    threes, not 5.
+    """
+    if 'cupy' in backend and not has_cuda_and_cupy():
+        pytest.skip("Requires CUDA and CuPy")
+    if 'dask' in backend and not dask_array_available():
+        pytest.skip("Requires Dask")
+
+    copied_data_zones = copy.deepcopy(data_zones)
+    copied_data_values_2d = copy.deepcopy(data_values_2d)
+
+    # cat_ids=[3] alone, with zone 3 containing 0s too. unique_cats is
+    # [0, 1, 2, 3]; the buggy code would let cat_start stay at the start
+    # of the sorted zone array and report 5 threes for zone 3.
+    df_result = crosstab(
+        zones=data_zones, values=data_values_2d,
+        zone_ids=[0, 1, 2, 3], cat_ids=[3],
+    )
+    expected = {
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: zonal: fix crosstab cat_ids overcount when earlier categories are filtered out

- None.

- None.

- None.

### What looks good
- Root cause is correctly identified. In `xrspatial/zonal.py:959`, `_strides` returns cumulative end-indices per unique category, so the count for category `j` must be `zone_cat_breaks[j] - zone_cat_breaks[j-1]`. The pre-fix code only advanced `cat_start` inside the `if cat in cat_ids:` branch, so when a leading or middle category was filtered out, the cursor stayed behind and the next selected category's count picked up the skipped cells.
- The fix is minimal and exactly targeted: one indentation change so `cat_start = zone_cat_breaks[j]` runs on every iteration. The comment block above the loop explains the invariant and references issue #2560, which is helpful for the next reader.
- Backend scope is correct. `_single_zone_crosstab_2d` is shared by `_crosstab_numpy`, `_crosstab_cupy` (which delegates to numpy), `_crosstab_dask_numpy` via `_single_chunk_crosstab`, and `_crosstab_dask_cupy` (which delegates to dask+numpy). One fix lands for all four. `_single_zone_crosstab_3d` uses a different per-cat slicing pattern and does not have the bug.
- Test parametrizes over all four backends with `has_cuda_and_cupy()` and `dask_array_available()` skips, uses the existing `data_zones` and `data_values_2d` fixtures, and goes through `check_results` for dask compute and tolerance handling. Coverage hits two distinct shapes: `cat_ids=[3]` (skip leading category 0, present in zone 3), and `cat_ids=[1, 3]` (skip middle category 2). Both cases would fail under the pre-fix code in zone 3 and zone 1 respectively.
- Input data immutability assertions remain (`assert_input_data_unmodified`), consistent with the other crosstab tests.
- CHANGELOG entry is clear and points at the right issue number.

### Checklist
- [x] Algorithm matches reference/paper: n/a, cumulative-stride accounting is internal.
- [x] All implemented backends produce consistent results: confirmed via parametrized test.
- [x] NaN handling is correct: unchanged; mask + isfinite applied before sort.
- [x] Edge cases are covered by tests: leading-skip and middle-skip both included.
- [x] Dask chunk boundaries handled correctly: no chunk-boundary semantics involved; per-zone reduction.
- [x] No premature materialization or unnecessary copies: diff does not add any.
- [x] Benchmark exists or is not needed: not needed for a one-line correctness fix.
- [x] README feature matrix updated (if applicable): not applicable; no new function or backend change.
- [x] Docstrings present and accurate: n/a for private helper; comment block added.

PR #2578, opened 2026-05-28
Validate return_type at entry to stats() (#2558)

### LINKED ISSUE #2558: stats(return_type=...) is not validated; bad values silently return a raw numpy array
## Description

`stats(return_type=...)` in `xrspatial/zonal.py` silently accepts any string and falls through to an undocumented path. The dispatch is:

- `return_type == 'pandas.DataFrame'` returns a `pandas.DataFrame`.
- `return_type == 'xarray.DataArray'` wraps the buffer into an `xr.DataArray`.
- Anything else returns the raw `numpy.ndarray` stats buffer from `_stats_numpy`.

So `stats(zones, values, return_type='bogus')` returns a numpy array of shape `(n_stats, *values.shape)` instead of raising. Typos in `return_type` are hidden, and callers get back a type that is not documented.

## Expected behavior

`return_type` should be validated at entry. Only the documented values (`'pandas.DataFrame'`, `'xarray.DataArray'`) should be accepted. Anything else should raise `ValueError` listing the allowed values.

## Reproducer

```python

zones = xr.DataArray(np.array([[0, 0], [1, 1]]))
values = xr.DataArray(np.array([[1.0, 2.0], [3.0, 4.0]]))
result = stats(zones, values, return_type='bogus')
print(type(result))  # numpy.ndarray -- should raise ValueError
```

## Proposed fix

1. Validate `return_type` at entry to `stats()` against an allowed set.
2. Raise a clear `ValueError` listing the allowed values.
3. Expand the docstring to enumerate what each allowed value returns.
4. Add tests covering both valid values, the default, and the invalid case.

brendancol:
Closes #2558.

## Summary

- `stats(return_type=...)` now validates against `('pandas.DataFrame', 'xarray.DataArray')` at entry and raises `ValueError` listing the allowed values for anything else. Before this change, an unrecognised string fell through to the numpy buffer branch and silently returned a raw `numpy.ndarray`.
- `return_type='xarray.DataArray'` on dask-backed input now raises a clear `ValueError` (the docstring already said it was unsupported, but the code path silently produced the wrong shape).
- Docstring expanded to enumerate what each allowed value returns.

## Backend coverage

The validator runs before backend dispatch, so the new error is consistent across numpy, cupy, dask+numpy, and dask+cupy. The dask-specific guard rejects the `'xarray.DataArray'` return type on any dask-backed input (numpy or cupy chunks).

## Test plan

- [x] Default `return_type` still returns `pandas.DataFrame`.
- [x] Explicit `'pandas.DataFrame'` returns `pandas.DataFrame`.
- [x] Explicit `'xarray.DataArray'` returns `xarray.DataArray` with the right dims and shape.
- [x] Invalid values (`'bogus'`, `'numpy.ndarray'`, `''`, `'DataFrame'`, `'xarray.dataarray'`) raise `ValueError`.
- [x] The error message names both allowed values.
- [x] Dask input rejects `'xarray.DataArray'`.
- [x] Dataset input with an invalid `return_type` raises `ValueError`.
- [x] Full existing zonal test suite still passes (147 tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 #### Fixed
 
+- `zonal_stats` now validates `return_type` at entry. Previously any
+  string other than `'pandas.DataFrame'` or `'xarray.DataArray'` fell
+  through to an internal branch that returned a raw `numpy.ndarray`,
+  hiding typos. Allowed values are now enforced and a clear
+  `ValueError` is raised otherwise. `return_type='xarray.DataArray'`
+  on dask-backed input also raises instead of silently returning
+  the wrong shape. (#2558)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/tests/test_zonal.py
+++ b/xrspatial/tests/test_zonal.py
@@ -2195,7 +2195,8 @@ class TestVectorZones:
 # ---------------------------------------------------------------------------
 
 
-def _small_zones_values_2558():
+@pytest.fixture
+def small_zones_values_2558():
     zones = xr.DataArray(np.array([[0, 0, 1, 1],
                                    [0, 0, 1, 1]], dtype=np.int32))
     values = xr.DataArray(np.array([[1.0, 2.0, 3.0, 4.0],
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Validate return_type at entry to stats() (#2558)

None.

- `xrspatial/zonal.py` around the new dask guard: `return_type='xarray.DataArray'` is also unsupported for the cupy backend. `_stats_cupy` returns a DataFrame and the downstream `xr.DataArray(result, coords=..., dims=...)` wrapper would crash or produce garbage. The current code only guards the dask case. Consider adding a parallel guard for cupy, or restructure as "only the pure-numpy path supports xarray.DataArray". This is not a regression introduced by this PR (the bug existed before), but the PR is already in the right area.

- `xrspatial/zonal.py` line ~833: the comment ends with "produce a numpy array instead of one of the documented types". Minor word choice, but "raw numpy buffer" or "internal stats buffer" might be clearer about *why* this is bad (it is not just the wrong type, it is an undocumented intermediate).
- `xrspatial/tests/test_zonal.py` `_small_zones_values_2558`: the helper is private to this test group and only used by tests in this file. Consider a `@pytest.fixture` for symmetry with the other fixtures in this module. Pure style.

### What looks good
- Early validation against an explicit allowed-set is the right shape for this fix.
- Error message includes the bad value (`return_type={return_type!r}`) and the allowed list, so callers can diagnose typos quickly.
- Test matrix covers the default, both valid values, a parameterized set of invalid strings, the error message contents, the dask path, and the Dataset path.
- Docstring rewrite spells out what each value returns and notes which backends each is valid for.
- CHANGELOG entry under Unreleased / Fixed is in the right section and references the issue.

### Checklist
- [x] Algorithm matches reference/paper: N/A, validation-only change
- [x] All implemented backends produce consistent results: the validator runs before dispatch
- [x] NaN handling is correct: not touched
- [x] Edge cases are covered by tests: empty string, wrong case, near-match
- [x] Dask chunk boundaries handled correctly: not touched
- [x] Benchmark exists or is not needed: not needed for a validation fix
- [x] README feature matrix updated: N/A, no new function

brendancol (commented):
## Follow-up review

All findings from the previous review pass have been addressed.

### Disposition

- Suggestion (cupy guard): fixed in 0a068d67. The non-numpy guard now covers both dask and cupy backends and reports which one it tripped on.
- Nit (validator comment wording): fixed. Rephrased to "leak that undocumented intermediate to the caller" so it is clear the old behaviour exposed an internal buffer, not just a wrong type.
- Nit (helper to fixture): fixed. `_small_zones_values_2558` is now a `@pytest.fixture` named `small_zones_values_2558`. A cupy rejection test was added in the same section, gated on `has_cuda_and_cupy()`.

### Verification

- All 12 issue-2558 tests pass locally (cupy test also runs and passes on this machine).
- Full `xrspatial/tests/test_zonal.py` suite: 148 passed.

No remaining findings from my side.

PR #2579, opened 2026-05-28
rasterize: reject partial width/height overrides instead of silent ignore

### LINKED ISSUE #2569: rasterize: partial width/height overrides with like= are silently ignored
## Description

`rasterize(..., like=template, width=W)` silently ignores `width` when `height` is not also provided, returning the template's width with no warning. The docstring says values from `like` "can still be overridden explicitly", but the implementation at xrspatial/rasterize.py:3209 only honors explicit dimensions when both width and height are supplied:

    if width is not None and height is not None:
        final_width, final_height = int(width), int(height)
    elif resolution is not None:
        ...
    elif like_width is not None:
        final_width, final_height = like_width, like_height

A partial override (only `width` or only `height`) falls through to the `like` branch and the explicit dimension is discarded without warning. Silent ignore is a bad contract for a foundational module.

## Reproducer

    import xarray as xr, numpy as np

    like = xr.DataArray(
        np.zeros((10, 10), dtype=np.float64),
        dims=("y", "x"),
        coords={"x": np.arange(10) + 0.5, "y": np.arange(10) + 0.5},
    )
    out = rasterize([(box(2, 2, 8, 8), 1.0)], like=like, width=2, fill=0)
    print(out.sizes)  # {'y': 10, 'x': 10} -- width=2 ignored, no warning

## Expected behavior

Raise a `ValueError` that names which dimension was ignored and tells the caller to provide both or neither. The other option (deriving the missing dimension from `like`'s aspect/resolution) is worse here because the bounds still come from `like`, so the x and y pixel resolutions would diverge and the output coords would no longer match the template.

## Fix plan

Raise `ValueError` when exactly one of `width` or `height` is provided together with `like`, naming the missing dimension.

Add tests covering:
- `like` + both `width` and `height` (honored, same as today)
- `like` + only `width` (raises)
- `like` + only `height` (raises)
- `like` + neither (uses `like` grid, same as today)

brendancol:
Closes #2569.

## Summary

- `rasterize(..., like=template, width=W)` used to silently drop `width` and return the template's size. Now it raises `ValueError` naming the missing dimension.
- The same check applies without `like`: passing only one of `width`/`height` (with or without `resolution`) now raises with a clear message instead of falling through to the next branch.
- Docstring updated so the both-or-neither contract is part of the documented API.

## Why ValueError rather than deriving the missing dim

When `like` is given, the bounds also come from the template. Filling in the missing dimension from aspect/resolution would force the x and y pixel sizes to disagree, and the output coords would no longer match `like`. A loud error is the less surprising behaviour.

## Backend coverage

This is in the front-end argument resolver, before any backend dispatch. Same behaviour for numpy / cupy / dask+numpy / dask+cupy.

## Test plan

- [x] `like` + both `width` and `height` honored (size matches the explicit args)
- [x] `like` + only `width` raises with "height was not"
- [x] `like` + only `height` raises with "width was not"
- [x] `like` + neither uses the template grid bit-exactly
- [x] No `like`: only-width-plus-resolution raises
- [x] No `like`: only-height raises
- [x] Full `xrspatial/tests/test_rasterize.py` still passes (215 passed, 2 skipped)

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -3002,10 +3002,14 @@ def rasterize(
         ``(shapely.geometry, numeric_value)`` pair.
     width : int, optional
         Number of columns in the output raster.  Required unless
-        ``resolution`` or ``like`` is given.
+        ``resolution`` or ``like`` is given.  Must be passed together
+        with ``height``: a partial override (only one of the two) is
+        rejected with ``ValueError`` rather than silently filling the
+        missing dimension from ``like`` or ``resolution``.
     height : int, optional
         Number of rows in the output raster.  Required unless
-        ``resolution`` or ``like`` is given.
+        ``resolution`` or ``like`` is given.  Must be passed together
+        with ``width`` (see above).
     bounds : tuple of (xmin, ymin, xmax, ymax), optional
         Geographic extent of the output raster.  Inferred from the
         geometries (or ``like``) if omitted.
@@ -3048,7 +3052,9 @@ def rasterize(
         A single float uses the same resolution for both axes.
     like : xr.DataArray, optional
         Template raster.  Width, height, bounds, and dtype are copied
-        from this array (any can still be overridden explicitly).
+        from this array.  Bounds and dtype can be overridden one at a
+        time; width and height must be overridden together (passing
+        only one raises ``ValueError``).
         Must have uniformly spaced ``x`` and ``y`` dim coords -- the
         rasterizer only writes to a regular grid, so a non-uniform
         ``like`` is rejected with ``ValueError`` rather than silently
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: rasterize: reject partial width/height overrides instead of silent ignore

### Blockers
None.

### Suggestions
None.

### Nits (optional)
- `xrspatial/rasterize.py:3214-3228`: the partial-dims check runs after `_parse_input` and the inferred-bounds computation. A caller passing only `width` with a large geometry set pays the bbox-inference cost before the error fires. Moving the check up to right after argument unpacking would fail faster on the unhappy path. Strictly cosmetic; the practical impact is small.
- `xrspatial/rasterize.py:3228`: the error message ends with "use resolution or like to size the output". Resolution alone still needs `bounds` (explicit or inferable from geometries), so the phrasing slightly oversells what resolution covers. Something like "pass both width and height together, or omit both and supply resolution or like" would avoid the implication.

### What looks good
- `(width is None) != (height is None)` is the clean XOR form and reads well.
- The error message names the missing dimension, which is exactly what the issue asked for.
- Tests cover all four cases from the issue (both / only-width / only-height / neither) plus a no-`like` variant that catches a related silent fall-through.
- Docstring updates on `width`, `height`, and `like` keep the documented contract aligned with the runtime check.
- Full `test_rasterize.py` still passes (215 passed, 2 skipped). No regression.

### Checklist
- [x] Algorithm matches reference/paper (N/A, argument validation only)
- [x] All backends behave consistently (check runs before backend dispatch)
- [x] NaN handling correct (N/A)
- [x] Benchmark exists or not needed (validation-only)
- [x] README feature matrix (no change required)

brendancol (commented):
## Follow-up review (after ce0acc1c)

Both nits from the previous review have been addressed:

- The partial-dims check now sits right after merge validation, before `_extract_grid_from_like` and `_parse_input`. The error fires before any geometry or template work, so a caller passing only `width` no longer pays the bbox-inference cost.
- The error message now reads "pass both width and height together, or omit both and supply resolution or like to size the output", so the phrasing no longer suggests resolution alone is sufficient.

### Status
- Blockers: none
- Suggestions: none
- Nits: none

Tests still green (221 passed, 2 skipped). Ready from the review side.

PR #2580, opened 2026-05-28
zonal: make crop() return (0, 0) on all backends when no zone matches

### LINKED ISSUE #2561: crop() returns different results by backend when requested zones are absent
**Describe the bug**

`xrspatial.zonal.crop()` returns different shapes from the numpy/cupy and dask backends when none of the requested `zone_ids` are present in the `zones` raster.

In `xrspatial/zonal.py`:

- The numpy/cupy `_crop` helper (around line 2498) leaves `top`, `bottom`, `left`, `right` in an inverted state when no requested zone is found, so the slice `[top:bottom+1, left:right+1]` collapses to `(0, 0)`.
- The dask `_crop_bounds_dask` helper (around line 2614) explicitly returns `0, rows-1, 0, cols-1` when no matches exist, which slices back the full input raster.

So the same call with the same data silently gives a caller either an empty result or the full untouched raster, depending only on whether the input happened to be chunked. That is a downstream-poison condition: callers can't tell which one happened without re-checking shapes against the input.

**Expected behavior**

All four backends (numpy, cupy, dask+numpy, dask+cupy) should agree on one defined behavior when the requested zones are absent. The recommended choice is to return an empty `(0, 0)` raster instead of silently returning the full input. The full-raster fallback hides the fact that the crop did nothing.

The chosen behavior should be:

1. Documented in the `crop()` docstring.
2. Covered by cross-backend tests using the helpers in `xrspatial/tests/general_checks.py` for:
   - Zone IDs that don't exist at all.
   - Zone IDs partially present.
   - Standard happy path.

**Reproduction**

```python
import numpy as np
import xarray as xr
import dask.array as da
from xrspatial.zonal import crop

arr = np.array([[0, 4, 0, 3],
                [0, 4, 4, 3],
                [0, 1, 1, 3]], dtype=np.int64)

zones_np = xr.DataArray(arr, dims=['y', 'x'])
zones_dask = xr.DataArray(da.from_array(arr, chunks=(2, 2)), dims=['y', 'x'])

# zone_ids=(99,) does not exist anywhere
print("numpy:", crop(zones_np, zones_np, zone_ids=(99,)).shape)
print("dask: ", crop(zones_dask, zones_dask, zone_ids=(99,)).shape)
```

Observed:

```
numpy: (0, 0)
dask:  (3, 4)
```

Expected: both backends return the same shape (preferred: `(0, 0)`).

**Additional context**

Surfaced by a code-review sweep of `zonal.py`. No screenshots / desktop / smartphone info needed; this is a Python library bug.

brendancol:
## Summary

- Fix backend divergence in `xrspatial.zonal.crop()`. When the requested `zone_ids` are not present in the `zones` raster, numpy/cupy used to return shape `(0, 0)` while dask silently returned the full input raster unchanged. Now all four backends return `(0, 0)`.
- Reworked `_crop` (CPU) and `_crop_bounds_dask` to return an explicit `found` flag (1/0) so the call site picks the empty slice deterministically instead of relying on inverted bounds.
- Documented the no-match behaviour in the `crop()` docstring Notes section.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy (parametrized regression tests).

Closes #2561

## Test plan

- [x] `pytest xrspatial/tests/test_zonal.py -k crop`, 20/20 pass (existing 11 + 9 new parametrized cases)
- [x] `pytest xrspatial/tests/test_zonal.py`, full file 148/148 pass
- [x] Manual repro from the issue body: numpy and dask both return `(0, 0)` for absent zones
- [ ] CI (numpy, cupy where available, dask+numpy, dask+cupy)

### DIFF
--- a/xrspatial/tests/test_zonal.py
+++ b/xrspatial/tests/test_zonal.py
@@ -1901,6 +1901,95 @@ def test_crop_missing_zone_ids_raises():
 
 
 # ---------------------------------------------------------------------------
+# Regression tests for #2561: crop() must return the same shape across all
+# backends when the requested zone_ids are absent (previously the dask path
+# returned the full input raster while numpy/cupy returned an empty (0, 0)).
+# ---------------------------------------------------------------------------
+
+_CROP_ARR_2561 = np.array([[0, 4, 0, 3],
+                           [0, 4, 4, 3],
+                           [0, 1, 1, 3],
+                           [0, 1, 1, 3],
+                           [0, 0, 0, 0]], dtype=np.int64)
+
+
+def _crop_backends_2561():
+    backends = ['numpy']
+    if has_dask_array():
+        backends.append('dask+numpy')
+    if has_cuda_and_cupy():
+        backends.append('cupy')
+        if has_dask_array():
+            backends.append('dask+cupy')
+    return backends
+
+
+@pytest.mark.parametrize("backend", _crop_backends_2561())
+def test_crop_absent_zone_ids_returns_empty_2561(backend):
+    """All backends must return shape (0, 0) when no requested zone exists."""
+    zones = create_test_raster(_CROP_ARR_2561, backend=backend)
+    values = create_test_raster(_CROP_ARR_2561, backend=backend)
+
+    # zone_id 99 is not present anywhere in the raster.
+    result = crop(zones, values, zone_ids=(99,))
+
+    # Output array type must match input type.
+    assert isinstance(result.data, type(zones.data))
+    assert result.shape == (0, 0)
+
+
+@pytest.mark.parametrize("backend", _crop_backends_2561())
+def test_crop_partial_zone_ids_2561(backend):
+    """Mix of present and absent zone_ids crops to the present ones only."""
+    zones = create_test_raster(_CROP_ARR_2561, backend=backend)
+    values = create_test_raster(_CROP_ARR_2561, backend=backend)
+
+    # Only zone_id 1 is actually present here. 77 and 88 are absent.
+    result_partial = crop(zones, values, zone_ids=(77, 1, 88))
+    result_present_only = crop(zones, values, zone_ids=(1,))
+
+    assert result_partial.shape == result_present_only.shape
+
+    if backend.startswith('dask'):
+        partial_np = result_partial.data.compute()
+        present_np = result_present_only.data.compute()
+    else:
+        partial_np = result_partial.data
+        present_np = result_present_only.data
+
+    if 'cupy' in backend:
+        partial_np = partial_np.get()
+        present_np = present_np.get()
+
+    np.testing.assert_array_equal(partial_np, present_np)
+
+
+@pytest.mark.parametrize("backend", _crop_backends_2561())
+def test_crop_happy_path_matches_numpy_2561(backend):
+    """All backends agree on the standard crop result."""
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: zonal: make crop() return (0, 0) on all backends when no zone matches

None.

- `xrspatial/tests/test_zonal.py:1959` — `test_crop_partial_zone_ids_2561` only checks that partial matches behave the same as the "present-only" subset. It does not pin an expected shape, so if both code paths drifted together the test would still pass. Adding `assert result_partial.shape == (2, 2)` (the bounding box of `zone_id=1`) makes the regression bar explicit.

- `xrspatial/zonal.py:2599` — the final return uses a literal `1` for the `found` flag while the docstring describes it as the variable `found`. Using `return top, bottom, left, right, found` is a touch more self-documenting; the local `found` is already 1 by the time we reach this line. Cosmetic only.
- `xrspatial/tests/test_zonal.py:1916` — `_crop_backends_2561()` runs at module import time inside `@pytest.mark.parametrize(...)`. Other tests in this file build backend lists the same way, so this is consistent. Flagging it only as a heads-up for future readers.

### What looks good

- The `found` flag convention is symmetric across `_crop` and `_crop_bounds_dask`, so the call site is uniform.
- Slicing with `values[0:0, 0:0]` preserves underlying array type (numpy / cupy / dask) and dim names. Verified by manual repro; no special-casing per backend needed.
- The numpy branch short-circuits with `if found == 0: return 0, 0, 0, 0, 0` right after the top-pass, so it does not waste work scanning for bottom / left / right.
- The `crop()` docstring Notes section spells out the new `(0, 0)` contract for all four backends.
- Cross-backend tests use `create_test_raster` and gate on `has_dask_array` / `has_cuda_and_cupy`, matching the existing pattern in this file.
- Existing `test_crop_nothing_to_crop` (where `zone_id=0` is actually present in the raster) still passes; the behavior change only affects the genuinely-absent case.

### Checklist
- [x] Algorithm matches reference/paper — N/A, this is a behaviour contract change
- [x] All implemented backends produce consistent results — `(0, 0)` everywhere
- [x] NaN handling is correct — unchanged from prior behaviour
- [x] Edge cases are covered by tests — absent, partial, happy path
- [x] Dask chunk boundaries handled correctly — uses the existing reduction approach
- [x] No premature materialization or unnecessary copies — `dask.compute` was already there
- [x] Benchmark exists or is not needed — bug fix, no perf impact
- [x] README feature matrix updated (if applicable) — no API surface change
- [x] Docstrings present and accurate — Notes section updated

brendancol (commented):
## Follow-up review (after ca1cae22)

Re-reviewed the diff after the follow-up commit. Both addressable findings from the first pass are resolved.

### Disposition

- Suggestion: pin shape in `test_crop_partial_zone_ids_2561` — fixed. `assert result_partial.shape == (2, 2)` now anchors the bounding box of `zone_id=1`, so the test can no longer pass by accidental co-drift of both code paths. Tests still green (20/20).
- Nit: `_crop` final return uses `found` instead of literal `1` — fixed. Self-consistent with the docstring now.
- Nit: `_crop_backends_2561()` called at module import — dismissed. Matches the existing convention in `test_zonal.py` and elsewhere; changing it would diverge from the project pattern.

No new findings on this pass. Ready for CI.

PR #2581, opened 2026-05-28
test_reproject: skip pyproj geoid cross-check on non-finite probes (#2567)

### LINKED ISSUE #2567: Reproject test: pyproj geoid cross-check skip misses non-finite probe values
**Describe the bug**

`test_geoid_height_matches_pyproj_within_cm` in `xrspatial/tests/test_reproject.py` skips only when the pyproj probe is near zero. When the EGM96 grid is missing and PROJ network access is off, pyproj's vertical transform can return `-inf`, `+inf`, or `nan` instead of `0.0`. The guard at line 4308-4313 misses that, so the assert at line 4326 fires against the local geoid lookup even though the local code returned a correct, finite value.

CI then flags a real bug in `geoid_height` when the actual problem is that pyproj has nothing to compare against.

**Reproduction**

```python
_, _, h_probe = transformer.transform(-74.0, 40.7, 0.0)
# If the grid is missing, h_probe can be -inf rather than 0.0.
if abs(h_probe) < 1.0:        # skip is missed for non-finite values
    pytest.skip(...)
```

**Expected behavior**

Skip the cross-check whenever the probe value is non-finite (`-inf`, `+inf`, `nan`), in addition to the near-zero case.

**Proposed fix**

- Replace `abs(h_probe) < 1.0` with a check that also handles `np.isfinite(h_probe)`. Pull the logic into a small helper so it can be reused if other pyproj-based geoid cross-checks land later.
- Add a regression test that feeds `-inf` into the helper (or monkeypatches the transformer) and confirms the cross-check skips instead of asserting.

**Scope**

Tests-only. No public API changes, no notebook or README updates.

**Severity**

Medium. The test produces false negatives on runners without the EGM96 grid, which hides real test signal.

brendancol:
Closes #2567.

## Summary

- The pyproj geoid cross-check in `test_geoid_height_matches_pyproj_within_cm` only skipped on the near-zero pyproj fallback. When PROJ returns `-inf`, `+inf`, or `nan` instead, the skip was missed and the assert fired against the local lookup, falsely flagging a bug in `geoid_height`.
- Pull the skip logic into a small `_pyproj_geoid_probe_is_usable` helper that classifies non-finite or near-zero probes as "grid unavailable". The failing test now goes through the helper.
- Direct unit coverage for the helper: finite, near-zero (0.0, +/-0.5), `-inf`, `+inf`, `nan`, and the configurable `zero_tol`.

## Backends

Tests-only. No backend code paths touched.

## Test plan

- [x] `pytest xrspatial/tests/test_reproject.py::TestPyprojGeoidProbeUsable -v`, 6 passed
- [x] `pytest xrspatial/tests/test_reproject.py::TestGeoidPixelCenterIndexing -v`, 3 passed (grid available locally; on grid-unavailable runners the cross-check now skips instead of asserting)
- [x] `pytest xrspatial/tests/test_reproject.py`, 360 passed

### DIFF
--- a/xrspatial/tests/test_reproject.py
+++ b/xrspatial/tests/test_reproject.py
@@ -4296,7 +4296,7 @@ class TestPyprojGeoidProbeUsable:
 
     def test_negative_inf_probe_is_not_usable(self):
         # Regression for the original bug: -inf used to slip past the
-        # near-zero guard and fire the assert below.
+        # near-zero guard and fire the assert in the pyproj cross-check.
         assert not _pyproj_geoid_probe_is_usable(float('-inf'))
 
     def test_positive_inf_probe_is_not_usable(self):
--- a/xrspatial/tests/test_reproject.py
+++ b/xrspatial/tests/test_reproject.py
@@ -61,6 +61,28 @@ def _gradient_raster(h=64, w=64, crs='EPSG:4326',
     return _make_raster(data, crs=crs, x_range=x_range, y_range=y_range)
 
 
+def _pyproj_geoid_probe_is_usable(probe, zero_tol=1.0):
+    """Return True if a pyproj vertical-transform probe value indicates the
+    geoid grid is actually installed and usable for a cross-check.
+
+    pyproj has two failure modes when the EGM96 grid is missing and PROJ
+    network access is disabled:
+
+    - it silently returns the input ellipsoidal height unchanged (so the
+      probe comes back as ~0.0 at a well-known sample point where the
+      true geoid undulation is tens of metres), or
+    - it returns a non-finite sentinel (``-inf``, ``+inf``, ``nan``).
+
+    Treat both as "grid unavailable" so the caller skips the cross-check
+    instead of asserting against the local lookup.
+    """
+    if not np.isfinite(probe):
+        return False
+    if abs(probe) < zero_tol:
+        return False
+    return True
+
+
 # ---------------------------------------------------------------------------
 # CRS utils
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: test_reproject: skip pyproj geoid cross-check on non-finite probes (#2567)

None.

None.

- `xrspatial/tests/test_reproject.py:4298-4299`: the comment "fire the assert below" reads as if there is an assert in this test. The assert it refers to lives in `test_geoid_height_matches_pyproj_within_cm`, in the next class. Reword to "the assert in the pyproj cross-check below" or drop "below".

### What looks good
- Helper is small, focused, and documented.
- Both pyproj failure modes (no-op ~0 and non-finite sentinel) are covered.
- The `zero_tol` knob keeps the helper reusable without baking the 1.0 m magic number into future callers.
- Unit tests cover the matrix: typical finite, near-zero positive/negative/zero, `-inf`, `+inf`, `nan`, and the configurable tolerance.
- Skip message now includes the actual probe value, which is useful when CI logs are the only signal.

### Checklist
- [x] Algorithm matches reference/paper (helper logic is straightforward)
- [x] All implemented backends produce consistent results (tests-only)
- [x] NaN handling is correct (`np.isfinite` catches nan and +/-inf together)
- [x] Benchmark exists or is not needed (N/A, tests-only)
- [x] README feature matrix updated (N/A, no new public API)

brendancol (commented):
## Follow-up review after nit fix

The single nit (misleading "fire the assert below" comment) is addressed in 13c2b365. The comment now references "the assert in the pyproj cross-check" instead of "below", which removes the false implication that the assert lives in the same test.

No outstanding findings. Helper tests still pass locally (6/6).

PR #2584, opened 2026-05-28
polygonize: reject non-finite simplify_tolerance (nan, inf)

### LINKED ISSUE #2575: polygonize: reject non-finite simplify_tolerance (nan, inf)
**Describe the bug**
`polygonize`'s `simplify_tolerance` only rejects negative values. Non-finite values slip through:

- `nan`: silently disables simplification. The gating check is `simplify_tolerance > 0`, which is `False` for `nan`, so the call returns unsimplified polygons even though the caller asked for simplification.
- `inf` (positive): collapses every polygon to nothing and returns empty output.
- `-inf`: caught by the existing `< 0` check, but only by accident of the sign.

The same module already validates `atol` and `rtol` as finite + non-negative. `simplify_tolerance` should follow the same pattern.

**Expected behavior**
Passing `simplify_tolerance=nan`, `simplify_tolerance=inf`, or `simplify_tolerance=-inf` should raise `ValueError` with a clear message, matching the existing `atol`/`rtol` validation.

**Reproducer**
```python
import numpy as np
import xarray as xr

raster = xr.DataArray(np.array([[0, 1], [1, 0]], dtype=np.float32))
polygonize(raster, simplify_tolerance=float('nan'))  # silently no-op, should raise
polygonize(raster, simplify_tolerance=float('inf'))  # empty output, should raise
```

**Fix**
Mirror the `atol`/`rtol` validation block at `xrspatial/polygonize.py:2013` and apply it to `simplify_tolerance` at `xrspatial/polygonize.py:1999`.

**Additional context**
Surfaced during an input-validation audit of `polygonize`.

brendancol:
## Summary

- Reject non-finite values (`nan`, `+inf`, `-inf`) for `polygonize`'s `simplify_tolerance` argument with a clear `ValueError`. Previously `nan` silently disabled simplification (because `nan > 0` is False) and `+inf` collapsed every polygon to empty output.
- Mirrors the existing `atol` / `rtol` validation contract in the same function: finite and non-negative.

Closes #2575

## Backend coverage

Validation lives in the public `polygonize` entry point, before backend dispatch, so the new check applies uniformly to numpy, cupy, dask+numpy, and dask+cupy.

## Test plan

- [x] `simplify_tolerance=nan` raises `ValueError`
- [x] `simplify_tolerance=+inf` raises `ValueError`
- [x] `simplify_tolerance=-inf` raises `ValueError` (was rejected by sign, now by finite check too)
- [x] `simplify_tolerance=-1.0` still raises (regression)
- [x] `simplify_tolerance=0.0` and `simplify_tolerance=1e-9` still succeed (guard against over-eager fix)
- [x] Full `TestPolygonizeSimplify` class passes locally (18 passed, 1 skipped)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 
 #### Fixed
 
+- `polygonize` now rejects non-finite values for `simplify_tolerance`
+  (`nan`, `+inf`, `-inf`) with a clear `ValueError`. Previously `nan`
+  silently disabled simplification (because `nan > 0` is False) and
+  `+inf` collapsed every polygon to empty output. Matches the existing
+  `atol` / `rtol` validation contract: finite and non-negative. (#2575)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -1995,10 +1995,15 @@ def polygonize(
             raise ValueError(
                 f"Incorrect transform length of {len(transform)} instead of 6")
 
-    # Check simplification parameters.
-    if simplify_tolerance is not None and simplify_tolerance < 0:
+    # Check simplification parameters.  Mirror the atol/rtol validation
+    # below: reject NaN and +/-inf along with negative values.  NaN would
+    # silently disable simplification because ``nan > 0`` is False, and
+    # +inf would collapse every polygon to empty output -- neither is
+    # what a caller asking for simplification wants.
+    if simplify_tolerance is not None and (
+            not np.isfinite(simplify_tolerance) or simplify_tolerance < 0):
         raise ValueError(
-            "simplify_tolerance must be non-negative, "
+            "simplify_tolerance must be a non-negative finite number, "
             f"got {simplify_tolerance}")
     if simplify_method not in ("douglas-peucker", "visvalingam-whyatt"):
         raise ValueError(
--- a/xrspatial/tests/test_polygonize.py
+++ b/xrspatial/tests/test_polygonize.py
@@ -954,6 +954,34 @@ class TestPolygonizeSimplify:
         with pytest.raises(ValueError, match="simplify_tolerance"):
             polygonize(data, simplify_tolerance=-1.0)
 
+    @pytest.mark.parametrize("bad", [
+        float("nan"), float("inf"), float("-inf")])
+    def test_non_finite_tolerance_raises(self, bad):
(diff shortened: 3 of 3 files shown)


brendancol (commented):
## PR Review: polygonize: reject non-finite simplify_tolerance (nan, inf)

- None.

- None.

- None. The `simplify_tolerance` validation could be co-located with the `atol`/`rtol` validation block at `xrspatial/polygonize.py:2013` for tighter grouping, but the current placement keeps it next to `simplify_method` validation and matches the pre-existing structure. Moving it would be churn.

### What looks good
- The check `not np.isfinite(simplify_tolerance) or simplify_tolerance < 0` is correct: `np.isfinite` returns False for nan / +inf / -inf and True for normal finite floats including 0. Combined with the existing `< 0` guard, this allows only finite non-negative values (which includes 0.0). Python `or` short-circuits, so the `< 0` comparison never sees nan and we avoid the `RuntimeWarning: invalid value encountered in less` that bare numpy comparisons would emit.
- Validation lives in the public `polygonize` entry point before `ArrayTypeFunctionMapping` dispatch, so the new contract applies uniformly to numpy, cupy, dask+numpy, and dask+cupy without per-backend duplication.
- Error message matches the wording style of the existing `atol`/`rtol` errors ("must be a non-negative finite number, got ..."), and `match="simplify_tolerance"` in the tests stays stable across future message tweaks.
- Tests use a uniform 2x2 raster for the failure cases (no actual polygon edges) and a 2x4 two-label raster for the happy-path 0.0 / 1e-9 case so simplification has something to act on. The split is appropriate -- validation runs before geometry work, so the failure-case raster does not need real edges.
- The `test_zero_and_small_positive_tolerance_still_work` regression guard is a nice touch: it would catch a future "fix" that bans non-positive along with non-finite.
- CHANGELOG entry under Unreleased / Fixed is accurate and references the issue.

### Checklist
- [x] Algorithm matches reference/paper -- N/A, input validation only
- [x] All implemented backends produce consistent results -- validation runs pre-dispatch
- [x] NaN handling is correct -- nan rejected up front, no downstream NaN propagation concerns
- [x] Edge cases are covered by tests -- nan, +inf, -inf, -1.0, 0.0, 1e-9
- [x] Dask chunk boundaries handled correctly -- N/A, no chunk-level code touched
- [x] No premature materialization or unnecessary copies -- N/A, scalar validation only
- [x] Benchmark exists or is not needed -- not needed for a one-line validation check
- [x] README feature matrix updated (if applicable) -- N/A, no API surface change
- [x] Docstrings present and accurate -- existing `simplify_tolerance` docstring at line 1877 still applies; consistent with `atol`/`rtol` which also do not enumerate their validation in the docstring

PR #2585, opened 2026-05-28
rasterize: reject zig-zag duplicate-coord patterns in _check_uniform_axis

### LINKED ISSUE #2566: rasterize: _check_uniform_axis accepts zig-zag duplicate coords as uniform
**Describe the bug**
`_check_uniform_axis` in `xrspatial/rasterize.py` (line 2848) compares
`abs(np.diff(coords))` against `abs(expected_step)`. That misses zig-zag /
duplicate-coord patterns. With `x = [0.5, 1.5, 0.5, 1.5]`,
`abs(diff) == [1.0, 1.0, 1.0]`, so the validator accepts it as uniform. The
rasterizer then reuses those duplicate coords on the output, and
`.sel(x=0.5)` returns two columns whose data came from different burn cells.

The comment at line 2934 says non-monotonic / duplicate-valued coords have
already been rejected by `_check_uniform_axis`. They haven't.

**Reproducer**

```python
import numpy as np
import xarray as xr

x = np.array([0.5, 1.5, 0.5, 1.5])   # zig-zag duplicates
y = np.linspace(3.5, 0.5, 4)
like = xr.DataArray(
    np.zeros((4, 4)), dims=('y', 'x'), coords={'y': y, 'x': x}
)

# Passes validation today. Output has two columns labelled x=0.5 that
# came from different burn cells.
out = rasterize([(box(0, 0, 2, 4), 1.0)], like=like, fill=0)
print(out.sel(x=0.5))
```

**Expected behavior**
`_check_uniform_axis` should reject non-monotonic / duplicate-coord patterns
with a `ValueError` naming the offending axis. The check should require all
signed diffs to be equal, not just their magnitudes.

**Suggested fix**
Compare signed `np.diff(coords)` against the signed first interval
(`coords[1] - coords[0]`), not `abs(diff)` against `abs(first_interval)`.
Descending y still passes because its signed first interval is negative, so
all signed diffs are equal and negative.

**Additional context**
Same validator path as the earlier fix for non-uniform `like` grids
(#2168), which only handled the magnitude case.

brendancol:
## Summary
- Closes a validation gap in `_check_uniform_axis`. The previous check
  compared `abs(np.diff(coords))` against `abs(expected_step)`, so a
  zig-zag like `x = [0.5, 1.5, 0.5, 1.5]` (abs-diffs all 1.0) passed
  as "uniform" and produced output with duplicate coords that broke
  `.sel`.
- Now compares signed diffs against the signed first interval.
  Ascending and descending axes still pass; zig-zag and duplicate-coord
  patterns raise a `ValueError` naming the offending axis.

## Backend coverage
Validation lives in the shared `_extract_grid_from_like` path, so all
four backends (numpy, cupy, dask+numpy, dask+cupy) pick this up.

## Test plan
- [x] New `xrspatial/tests/test_rasterize_signed_step_2566.py`: ascending
      and descending uniform pass; zig-zag x and y rejected; strictly
      increasing non-uniform still rejected (the #2168 case); single-cell
      axis short-circuits; symptom-level check that the rasterize call
      raises before it can return a DataArray with duplicate coords.
- [x] Existing `TestLikeUniformGridValidation` (7 tests) still passes.
- [x] Full `test_rasterize.py`: 215 passed, 2 skipped.

Closes #2566

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -2816,12 +2816,14 @@ def _check_uniform_axis(axis_name, coords, expected_step):
     than three points cannot be non-uniform in a way this check would
     catch, so they pass trivially.
 
-    The comparison is on ``abs(diff)`` so the validation does not care
-    whether the axis is ascending or descending -- ascending-y ``like``
-    inputs are supported by the orientation flip in ``rasterize``, and
-    gating on the sign here would block that.  ``np.allclose`` is used
-    (rather than strict equality) because affine-transform-derived
-    coords drift by a few ulps in practice.
+    The comparison uses the *signed* ``np.diff`` against the signed
+    first interval so the validation does not care whether the axis
+    is ascending or descending -- the sign of the first interval
+    carries the direction.  This also rejects zig-zag /
+    duplicate-coord patterns like ``[0.5, 1.5, 0.5, 1.5]`` whose
+    ``abs(diff)`` is uniform but whose signed diffs alternate.
+    ``np.allclose`` is used (rather than strict equality) because
+    affine-transform-derived coords drift by a few ulps in practice.
     """
     if coords.size < 3:
         return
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: rasterize signed-diff uniformity check (#2585)

### Blockers
None.

### Suggestions
- `xrspatial/rasterize.py:2925-2929` -- the caller-side comment still reads "Compare `abs(diff)` against the first interval so the check stays agnostic to axis direction". After this PR the validator compares signed diffs against the signed first interval, so the comment now describes the old behaviour. Update it.

### Nits
- `xrspatial/rasterize.py:2810` -- `expected_step` is now only used for the `== 0` and `isfinite` short-circuits; the actual comparison computes `coords[1] - coords[0]` internally. The caller still passes `abs(x[1] - x[0])`, which works for both gates because only magnitude matters there. Worth a follow-up: either drop the parameter or rename it so the abs-magnitude role is explicit.
- `xrspatial/tests/test_rasterize_signed_step_2566.py` -- the file docstring lists six pinned cases but the class has seven test methods (the trailing "no duplicate output coords" symptom check). Update the count or drop the test (it overlaps with `test_zigzag_x_rejected`).

### What looks good
- Minimal surgical fix: signed diff against signed first interval, no other behaviour changes.
- Zero-step and non-finite short-circuits still work because the caller's `abs(...)` agrees with the new internal `signed_step` on both edge values.
- The comment at `rasterize.py:2944` ("rejected non-monotonic or duplicate-valued y coords") is now accurate -- it was aspirational before.
- Tests pin ascending uniform, descending uniform, zig-zag on x and y, strictly-increasing non-uniform (#2168), and the single-cell short-circuit.
- Validation lives upstream of all four backend paths, so numpy, cupy, dask+numpy, and dask+cupy all pick this up without per-backend changes.

### Checklist
- [x] Algorithm matches issue description
- [x] All backends produce consistent results (shared validation)
- [x] NaN handling unchanged (non-finite short-circuit preserved)
- [x] Edge cases covered (descending, single-cell, zig-zag)
- [ ] Dask chunk boundaries -- N/A
- [ ] Benchmark -- N/A
- [ ] README feature matrix -- N/A

brendancol (commented):
## Follow-up review pass

Addressed in `43fb6ff8`:

- Suggestion (stale caller comment at `rasterize.py:2925-2929`): **fixed**. The comment now describes the signed-diff comparison.
- Nit (test docstring listed 6 cases for 7 methods): **fixed**. Added the seventh entry covering the symptom-level zig-zag check.

Deferred to follow-up:

- Nit (`expected_step` parameter no longer drives the main comparison): not changed in this PR. The parameter is still consulted for the `expected_step == 0` and `np.isfinite(expected_step)` short-circuits, so it is not dead. Renaming or dropping it expands the surgical scope of the bug fix without changing observable behaviour; better as its own cleanup.

Test status after follow-up: all 14 in `TestSignedStepValidation_2566` + `TestLikeUniformGridValidation` pass.

PR #2586, opened 2026-05-28
rasterize: validate resolution= shape and element type (#2576)

### LINKED ISSUE #2576: rasterize: tighten resolution= shape and type validation
**Describe the bug**

`rasterize()` accepts a `resolution=` argument that should be either a scalar number or a length-2 `(x_res, y_res)` tuple. The code at `xrspatial/rasterize.py:3211-3215` does not check shape or element types, so several bad inputs either slip past or fail with confusing errors:

- `resolution=(1, 2, 3)` is silently accepted. The third element is dropped without a warning.
- `resolution=(1,)` raises `IndexError: tuple index out of range` from `resolution[1]`.
- `resolution='foo'` raises `ValueError: could not convert string to float: 'f'` because it iterates the string character-by-character.
- `resolution={'x': 1, 'y': 1}` raises `KeyError: 0` from `resolution[0]`.

None of these name the offending input or explain what was expected.

**Expected behavior**

A single readable `ValueError` for anything that is not a scalar number or a length-2 sequence of numbers, naming the offending value:

- Scalar `int` or `float`: accepted.
- Length-2 tuple/list of numbers: accepted.
- Length-0, length-1, length-3+ sequences: clean `ValueError`.
- Strings, dicts, non-numeric elements: clean `ValueError`.

**Repro**

```python
import geopandas as gpd

g = gpd.GeoDataFrame({'geometry': [box(0,0,10,10)], 'val': [1.0]})
rasterize(g, column='val', resolution=(1, 2, 3), bounds=(0,0,10,10))  # silent drop of third
rasterize(g, column='val', resolution=(1,), bounds=(0,0,10,10))       # IndexError
rasterize(g, column='val', resolution='foo', bounds=(0,0,10,10))      # raw float() error
```

**Additional context**

Surfaced in a code review of `xrspatial/rasterize.py`. Filed as a bug because the 3-tuple silently produces incorrect output, and the other cases produce noisy errors that should collapse into a single clean `ValueError`.

brendancol:
## Summary

- Tighten `resolution=` validation in `rasterize()` so the parameter only accepts a scalar number or a length-2 sequence of numbers.
- Anything else (length-0/1/3+ sequences, strings, dicts, bools, non-numeric elements) now raises a clean `ValueError` naming the offending input.
- Preserves the existing finite/positive checks.

Before this fix, `(1, 2, 3)` was silently truncated, `(1,)` crashed with `IndexError`, `'foo'` leaked a raw `float()` conversion error, and `{'x': 1, 'y': 1}` raised `KeyError: 0`.

Backend coverage: pure input validation that runs before any backend dispatch, so it applies uniformly across numpy, cupy, dask+numpy, dask+cupy.

## Test plan

- [x] New tests in `xrspatial/tests/test_rasterize_resolution_validation_2576.py` cover the happy paths (scalar int/float, length-2 tuple/list) and the rejection paths (3-tuple, 1-tuple, empty tuple, string, dict, bool, `(1.0, None)`, `(1.0, "bar")`, plus an error-message-content check).
- [x] Existing resolution coverage tests in `test_rasterize_coverage_2026_05_17.py` and `test_rasterize_coverage_2026_05_21.py` still pass.

Closes #2576

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@
 
 #### Fixed
 
+- `rasterize` now validates the `resolution=` argument shape and element
+  type before unpacking. A scalar number or a length-2 sequence of
+  numbers is accepted; anything else (length-0, length-1, length-3+
+  sequences, strings, dicts, non-numeric elements) raises a clean
+  `ValueError` naming the offending input. Previously a 3-tuple was
+  silently truncated to its first two elements, a 1-tuple crashed with
+  `IndexError`, strings leaked a raw `float()` conversion error, and
+  dicts raised `KeyError: 0`. (#2576)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -3209,10 +3209,32 @@ def rasterize(
     if width is not None and height is not None:
         final_width, final_height = int(width), int(height)
     elif resolution is not None:
+        # Validate shape and element type up front so bad inputs surface a
+        # single clean ValueError naming the offending value, instead of
+        # leaking IndexError (length-1 sequences), KeyError (dicts), or a
+        # raw float() conversion error (strings, since strings iterate
+        # character-by-character into resolution[0]/[1]).  A 3+-element
+        # sequence was previously silently truncated to the first two
+        # elements -- reject it here too.
+        if isinstance(resolution, bool) or not isinstance(
+                resolution, (int, float, tuple, list)):
+            raise ValueError(
+                f"resolution must be a number or a length-2 sequence of "
+                f"numbers (x_res, y_res), got {resolution!r}")
         if isinstance(resolution, (int, float)):
             x_res = y_res = float(resolution)
         else:
-            x_res, y_res = float(resolution[0]), float(resolution[1])
+            if len(resolution) != 2:
+                raise ValueError(
+                    f"resolution sequence must have length 2 (x_res, y_res), "
+                    f"got length {len(resolution)}: {resolution!r}")
+            try:
+                x_res = float(resolution[0])
+                y_res = float(resolution[1])
+            except (TypeError, ValueError):
+                raise ValueError(
+                    f"resolution sequence elements must be numbers, "
+                    f"got {resolution!r}")
         # Reject non-finite or non-positive resolution before dimension math.
         # Without this, inf/-1 quietly produce a 1x1 raster, 0 raises an
         # opaque ZeroDivisionError, and nan raises an int-conversion error.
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: rasterize: validate resolution= shape and element type (#2576)

None.

- [ ] `xrspatial/rasterize.py:3217-3221`: the `isinstance(resolution, (int, float, tuple, list))` whitelist rejects numpy scalars (`np.float32`, `np.int32`, `np.int64`) and 1-D numpy arrays. Before this PR, `resolution=np.array([2.0, 1.0])` and `resolution=np.int32(2)` worked via duck typing through `float()`. Numpy-typed scalars from geospatial pipelines are a realistic input. Consider widening the accepted types: add `np.number` to the scalar check and `np.ndarray` (with a 1-D / size-2 guard) to the sequence check, or fall back to a `numbers.Real`-based check.

- [ ] `xrspatial/tests/test_rasterize_resolution_validation_2576.py:15`: `import numpy as np` is unused. Drop the import. Once the numpy-scalar acceptance above is applied, add tests for `resolution=np.array([2.0, 1.0])` and `resolution=np.float32(2.0)` to pin that behaviour.

- [ ] `xrspatial/rasterize.py:3215`: the comment says "leaking IndexError (length-1 sequences)". The IndexError actually comes from `resolution[1]` on a length-1 input. Minor wording polish -- could read "length-1 sequences would crash at `resolution[1]`" to be precise.

### What looks good

- Validation runs after the explicit-`width`/`height` branch, so it does not slow that common path.
- Error messages name the offending input via `{resolution!r}`, which is what the finding asked for.
- `bool` is rejected before the `int` check, so `resolution=True` does not silently produce a 1x1 raster.
- The finite/positive check is preserved unchanged.
- Tests cover all six rejection cases from the finding plus happy paths and the error-message-content check.
- CHANGELOG entry is clear and references the issue.

### Checklist
- [x] Algorithm matches reference/paper (input-validation patch, n/a)
- [x] All implemented backends produce consistent results (validation runs before dispatch)
- [x] NaN handling is correct (preserved unchanged)

brendancol (commented):
## Follow-up review pass

All findings from the previous review have been addressed in 0d5e827f.

### Dispositions

- **Suggestion 1 (numpy scalar / ndarray support):** Fixed. The scalar check now accepts `(int, float, np.number)` (excluding `bool` and `np.bool_`), and the sequence check accepts `(tuple, list, np.ndarray)` with a 1-D guard so a 2-D array does not slip past as a length-2 sequence.
- **Suggestion 2 (unused numpy import + numpy-scalar tests):** Fixed. `numpy` is now used by the new tests covering `np.float32(2.0)`, `np.int32(2)`, `np.array([2.0, 1.0])`, and the 2-D rejection path.
- **Nit (comment wording):** Fixed. The comment now says "length-1 sequences would crash at `resolution[1]`".

### Verification

- 16/16 tests in `test_rasterize_resolution_validation_2576.py` pass.
- 7/7 existing `resolution=`-related tests in `test_rasterize_coverage_2026_05_17.py` and `test_rasterize_coverage_2026_05_21.py` still pass.

No new findings on this pass.

PR #2587, opened 2026-05-28
resample: refresh transform to match actual coord orientation (#2571)

### LINKED ISSUE #2571: resample: attrs['transform'] is wrong for ascending-y or descending-x inputs
**Describe the bug**

`resample()` always writes a north-up rasterio transform (positive x
scale, negative y scale, origin at the top-left edge), no matter what
orientation the input coords have. If the input array has ascending y
or descending x coords, `attrs['transform']` on the output no longer
matches the coords on the same DataArray. Downstream code that
georeferences via `attrs['transform']` will get a flipped raster.

The bug is in `xrspatial/resample.py` around line 1364. The current
code picks the right `top` / `left` corner for the coord direction,
but it still writes `(out_res_x, 0.0, left, 0.0, -out_res_y, top)`.
Rasterio's affine maps `(col=0, row=0) -> (x, y)` for the first pixel
in the array, not for the top-left geographic corner, so the scale
signs and origin corner have to follow the actual array layout.

**Expected behavior**

`attrs['transform']` should match the coords on the output array for
all four combinations of x ascending/descending and y
ascending/descending. `transform * (0, 0)` should land on the upper-
left corner of the first array pixel (the pixel at `[0, 0]`), and the
scale signs should follow the coord direction.

**Minimal repro**

\`\`\`python
import numpy as np
import xarray as xr

# Ascending-y input.
data = np.arange(16, dtype=np.float32).reshape(4, 4)
raster = xr.DataArray(
    data,
    dims=['y', 'x'],
    coords={'y': np.linspace(0, 3, 4), 'x': np.linspace(0, 3, 4)},
    attrs={'transform': (1.0, 0.0, -0.5, 0.0, 1.0, -0.5),
           'res': (1.0, 1.0)},
)

out = resample(raster, scale_factor=0.5)
print(out.attrs['transform'])
# Currently: (2.0, 0.0, -0.5, 0.0, -2.0, 3.5)  -- negative y scale,
# top-edge origin, even though the output y coords still ascend.
# Expected: (2.0, 0.0, -0.5, 0.0, 2.0, -0.5).
\`\`\`

**Additional context**

Surfaced by a code review of the resample module. Only the metadata
refresh block at the end of \`resample()\` is affected -- the resampled
pixel values are correct.

brendancol:
Closes #2571.

## Summary

- `resample()` was always writing a north-up rasterio transform
  (`+x`, `-y`, top-left origin), even when the input had ascending y
  or descending x coords. The refreshed `attrs['transform']` then no
  longer matched the coords on the same DataArray, so downstream
  consumers that georeference via `transform` would see a flipped
  raster.
- Use the signed `px` / `py` from `_new_coords` and anchor the
  origin at `*_edge_start` (the leading edge on the side of
  `vals[0]`). This matches rasterio's `(col=0, row=0) -> first array
  pixel` convention for every coord orientation.

## Backend coverage

The transform refresh runs in pure Python on the metadata after the
backend dispatch, so the fix applies to numpy, cupy, dask+numpy, and
dask+cupy identically. No backend-specific code changed.

## Test plan

- [x] `pytest xrspatial/tests/test_resample.py::TestMetadataPropagation`
  -- 19 passed, includes the 4 new orientation parametrizations.
- [x] `pytest xrspatial/tests/test_resample.py` -- 182 passed.

### DIFF
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -1361,19 +1361,18 @@ def resample(
     if has_nodata:
         new_attrs['_FillValue'] = float('nan')
 
-    # Refresh `transform` if the input had one. The rasterio 6-tuple is
-    # (res_x, 0.0, left, 0.0, -res_y, top). `top` is the upper edge of
-    # the first row, which is `y_edge_start` when y is descending and
-    # `y_edge_end` when y is ascending. `left` is the lower edge of the
-    # first column, which is `x_edge_start` when x is ascending and
-    # `x_edge_end` when x is descending.
+    # Refresh `transform` if the input had one. The rasterio 6-tuple
+    # `(a, b, c, d, e, f)` maps `(col, row) -> (x, y)` for the first
+    # array pixel at `(col=0, row=0)`, so the scale signs and the
+    # origin corner have to follow the actual array layout rather
+    # than assuming a north-up grid. `px` / `py` from `_new_coords`
+    # are already signed (positive when the coord ascends along the
+    # axis, negative when it descends), and `*_edge_start` is the
+    # leading edge of the first row / column on the side of
+    # `vals[0]` -- exactly what rasterio wants for `c` and `f`.
     if 'transform' in agg.attrs:
-        out_res_x = abs(px)
-        out_res_y = abs(py)
-        top = y_edge_start if y_vals[0] > y_vals[-1] else y_edge_end
-        left = x_edge_start if x_vals[0] < x_vals[-1] else x_edge_end
         new_attrs['transform'] = (
-            out_res_x, 0.0, left, 0.0, -out_res_y, top,
+            px, 0.0, x_edge_start, 0.0, py, y_edge_start,
         )
 
     # Resample replaces sentinel pixels with NaN regardless of input
--- a/xrspatial/tests/test_resample.py
+++ b/xrspatial/tests/test_resample.py
@@ -151,6 +151,54 @@ class TestMetadataPropagation:
             2.0, 0.0, 100.0, 0.0, -2.0, 200.0,
         )
 
+    @staticmethod
+    def _raster_with_orientation(x_asc, y_asc):
+        """4x4 raster with the chosen coord orientation and a matching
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: resample transform refresh (#2571)

### Blockers
None.

### Suggestions
None.

### Nits
- xrspatial/tests/test_resample.py:163 -- the `_raster_with_orientation`
  docstring says "upper-left edge of pixel [0, 0]", but for ascending
  y the first pixel sits at the geographic *lower*-left. The
  transform is still right; the wording just borrows rasterio's
  "upper-left" convention where "leading-edge corner of pixel [0, 0]"
  would be more honest.
- xrspatial/tests/test_resample.py:191 -- `assert b == 0.0 and d == 0.0`
  packs two facts into one assertion. Splitting them tells you which
  off-diagonal drifted when it fails.

### What looks good
- The math holds up. `px` / `py` out of `_new_coords` are already
  signed, and `*_edge_start` is the leading edge on the `vals[0]`
  side, so `transform * (col, row)` lands where the output coords
  say it should regardless of orientation.
- The pre-existing `test_transform_refreshed_on_downsample` is
  untouched and still passes, which is a useful regression anchor.
- Metadata-only change, so the fix covers numpy, cupy, dask+numpy,
  and dask+cupy without any per-backend code.
- The parametrized test sweeps all four x asc/desc x y asc/desc
  combos and round-trips `transform * (col+0.5, row+0.5)` against
  the output pixel centers.

### Checklist
- [x] Algorithm matches the rasterio Affine convention
- [x] All backends consistent (metadata-only)
- [x] NaN handling unchanged
- [x] Orientation edge cases covered
- [x] Dask chunks N/A
- [x] Benchmark not needed
- [x] Docstrings unchanged

brendancol (commented):
## Follow-up review (#2571)

Both nits from the previous review are fixed in 34dc110:

- `_raster_with_orientation` docstring now says "leading-edge corner"
  instead of "upper-left edge".
- The combined off-diagonal assert is split into two.

No new findings on the follow-up pass. The 19
`TestMetadataPropagation` tests still pass locally.

PR #2589, opened 2026-05-28
zonal: fix cupy backend dropping all-NaN zones and desyncing zone_ids (#2562)

### LINKED ISSUE #2562: zonal.stats cupy backend drops all-NaN zones and desyncs zone/index when zone_ids requests missing zones
**Describe the bug**

`_stats_cupy()` in `xrspatial/zonal.py` behaves differently from the numpy and dask paths in two ways.

First, the cupy path filters `values_by_zone` for finite / non-nodata entries *before* it computes `unique_zones`. If every value in a zone is NaN (or equals the nodata sentinel), the zone vanishes from the result. The numpy path keeps that zone and returns NaN stats for it.

Second, when the caller passes `zone_ids` containing IDs that aren't in the raster, the cupy path overwrites `unique_zones` with the full requested list but only accumulates counts/indices for zones it actually found. The downstream loop iterates over `unique_zones` (the requested list) while indexing into `unique_index` / `unique_counts` (the found-only list), so per-zone stats can shift onto the wrong zone ID. The numpy and dask paths filter `zone_ids` down to zones that do exist in the raster, so the same call returns a consistent (zone, stats) mapping.

**Expected behavior**

The cupy result should match the numpy result row-for-row for the same inputs:

- a zone whose values are all NaN or all nodata should appear in the output with NaN stats, not disappear
- `zone_ids=[...]` containing IDs not present in the zones raster should not desync the zone/stat alignment

**Steps to reproduce**

Build a zones array where zone 5 contains only NaN values in the corresponding values array. Call `zonal_stats` on the numpy and cupy backends with the same input. The numpy result lists zone 5 with NaN stats; the cupy result omits zone 5.

Separately, call `zonal_stats(zones, values, zone_ids=[1, 999])` where 999 doesn't appear in `zones`. The cupy path returns a row labeled 999 but populated with stats from whichever zone landed at that index. The numpy path returns just the zone 1 row.

**Fix direction**

1. Compute `unique_zones` from the raw zones array (filtered only by valid zone IDs, not by value validity), then accumulate stats per zone. All-NaN zones should appear with NaN stats.
2. When `zone_ids` includes zones not in the raster, either filter them out (matching numpy/dask) or return NaN at the correct positions. Either way, the zone column and stats columns must stay aligned.
3. Add cross-backend tests covering both scenarios.

brendancol:
## Summary

- Fix `_stats_cupy()` dropping zones whose values are entirely NaN or all equal to `nodata_values`. The cupy path now returns a row with NaN stats, matching numpy and dask.
- Fix `_stats_cupy()` desyncing the zone column from per-zone stats when `zone_ids` contains IDs that don't appear in the raster. Missing IDs are filtered out, matching numpy and dask.
- Add cross-backend regression tests for both cases plus a row-for-row numpy/cupy parity check. The pre-existing `test_stats_all_nan_zone` and `test_stats_nodata_wipes_zone` had per-backend branches that pinned the buggy cupy behavior; those branches are gone.

Closes #2562

## Backend coverage

- numpy: unchanged (already correct)
- cupy: fixed
- dask+numpy: unchanged (already correct)
- dask+cupy: unchanged (delegates to dask+numpy via `_stats_dask_cupy`)

## Test plan

- [x] `pytest xrspatial/tests/test_zonal.py` (141 passed)
- [x] New regression tests run on numpy and cupy backends
- [x] Updated `test_stats_all_nan_zone` and `test_stats_nodata_wipes_zone` no longer carry a cupy-specific branch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 #### Fixed
 
+- `zonal.stats` on the cupy backend now agrees with the numpy and dask
+  backends on two edge cases. A zone whose values are all NaN (or all
+  equal to `nodata_values`) is preserved in the output with NaN stats
+  instead of being dropped. When `zone_ids` contains IDs that do not
+  appear in the zones raster, the missing IDs are filtered out so the
+  zone column and stats columns stay aligned (previously the cupy path
+  could silently desynchronize the rows). (#2562)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/tests/test_zonal.py
+++ b/xrspatial/tests/test_zonal.py
@@ -620,6 +620,7 @@ def test_stats_all_nan_zone_preserved(backend):
         zones=zones, values=values, stats_funcs=['min', 'max', 'mean', 'count']
     )
 
+    assert list(df_result.columns) == ['zone', 'min', 'max', 'mean', 'count']
     expected_result = {
         'zone': [1, 2, 3, 5],
         'min': [1.0, 3.0, 9.0, np.nan],
@@ -658,6 +659,7 @@ def test_stats_zone_ids_missing_from_raster(backend):
         stats_funcs=['min', 'max', 'sum', 'count']
     )
 
+    assert list(df_result.columns) == ['zone', 'min', 'max', 'sum', 'count']
     expected_result = {
         'zone': [1],
         'min': [10.0],
@@ -666,6 +668,23 @@ def test_stats_zone_ids_missing_from_raster(backend):
         'count': [4],
     }
     check_results(backend, df_result, expected_result)
+
+    # Also verify that zone_ids passed in reverse order (or with duplicates)
+    # produces the same numpy-style ordering: np.unique sorts ascending and
+    # dedupes, so [3, 1, 1] becomes [1, 3]. This pins the cupy path's
+    # np.unique normalization that matches the numpy backend.
+    df_result_reversed = stats(
+        zones=zones, values=values, zone_ids=[3, 1, 1],
+        stats_funcs=['min', 'max', 'sum', 'count'],
+    )
+    expected_reversed = {
+        'zone': [1, 3],
+        'min': [10.0, 100.0],
+        'max': [40.0, 100.0],
+        'sum': [100.0, 400.0],
+        'count': [4, 4],
+    }
+    check_results(backend, df_result_reversed, expected_reversed)
 
 
 @pytest.mark.filterwarnings("ignore:All-NaN slice encountered:RuntimeWarning")
--- a/xrspatial/tests/test_zonal.py
+++ b/xrspatial/tests/test_zonal.py
@@ -590,6 +590,137 @@ def test_majority_with_ties(backend):
     check_results(backend, df_result, expected_result)
 
 
+# Regression tests for issue #2562: cupy backend used to drop all-NaN zones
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: zonal cupy fix for all-NaN zones and zone_ids desync (#2562)

None.

- `xrspatial/zonal.py:571-588`: the numpy path normalizes zone_ids with `np.unique(zone_ids)` (line 495), which deduplicates and sorts the requested IDs. The cupy path now preserves caller order and keeps duplicates. The PR explicitly claims row-for-row parity with numpy, so apply the same `np.unique` normalization on the cupy side before iterating. Without it, `stats(..., zone_ids=[3, 1])` returns rows in `[3, 1]` order on cupy versus `[1, 3]` on numpy.

- `xrspatial/zonal.py:617-619`: the `if zone_values.size == 0` check is re-evaluated inside the inner stats-funcs loop on every iteration, even though `zone_values` does not change across stats. Hoist it before the loop and append NaN to every `stats_dict[stats]` in a single pass. Negligible runtime impact on small zone counts; reads cleaner.
- `xrspatial/tests/test_zonal.py`: `check_results` allows different column orderings. Consider also asserting `list(df_result.columns) == ['zone', 'min', 'max', 'mean', 'count']` in the new regression tests to pin the column order returned by `_stats_cupy`. Not required for the bug fix.

### What looks good
- The fix correctly separates zone discovery (raw zones, finite-only) from per-zone value filtering (NaN + nodata), which matches the numpy/dask contract.
- The `kept_zones` list tracks the actual zones whose stats are populated, so the zone column and per-zone arrays cannot desynchronize.
- All-NaN zone returns `float('nan')` for every stat, matching numpy where `_calc_stats` leaves the slot at NaN.
- Two pre-existing tests (`test_stats_all_nan_zone`, `test_stats_nodata_wipes_zone`) carried per-backend branches that pinned the buggy behavior; collapsing them is the right call.
- The new `test_stats_cupy_matches_numpy_row_for_row` compares numpy vs cupy row-for-row with `equal_nan=True` over multiple stats, both with and without `zone_ids`.
- CHANGELOG entry is accurate and references the issue.

### Checklist
- [x] Algorithm matches reference (numpy path is the reference)
- [x] All implemented backends produce consistent results (verified by new parity test)
- [x] NaN handling is correct
- [x] Edge cases covered by tests (all-NaN zone, missing zone_ids, parity)
- [N/A] Dask chunk boundaries: `_stats_cupy` is pure-cupy; dask+cupy delegates to dask+numpy
- [N/A] Benchmark: pure correctness fix
- [N/A] README feature matrix: no new function, no backend support change
- [x] Docstrings present (no public API changed)

brendancol (commented):
## Follow-up review after addressing #2562 review findings

All three findings from the previous review have been applied in commit 0722de96.

### Dispositions

- Suggestion (`xrspatial/zonal.py:571-588` — np.unique normalization on zone_ids): **fixed**. The cupy path now runs `zone_ids = np.unique(zone_ids)` before iterating, matching the numpy path's dedupe + sort behavior. Added a reverse-order regression test (`zone_ids=[3, 1, 1]` returns rows in `[1, 3]` order).
- Nit (`xrspatial/zonal.py:617-619` — hoist size==0 check): **fixed**. The empty-zone branch now appends NaN for every stat in a single pass before the inner stats-funcs loop, instead of re-checking on every iteration.
- Nit (`xrspatial/tests/test_zonal.py` — assert column order): **fixed**. The new regression tests now assert `list(df_result.columns)` against the expected ordering, pinning the cupy column layout.

### Test status

`pytest xrspatial/tests/test_zonal.py` — 141 passed.

No remaining findings.

PR #2590, opened 2026-05-28
resample: validate target_resolution and scale_factor up front (#2574)

### LINKED ISSUE #2574: resample: target_resolution and scale_factor input validation is leaky
**Describe the bug**

`xrspatial.resample.resample` validates `method` and the sign of the resolved scale factors but does not validate the *shape* or *finiteness* of `target_resolution` / `scale_factor` up front. Four bad inputs slip past the guard at the top of the function and fail later with confusing errors (or, worse, silently use the wrong axis).

Around resample.py:1243 and resample.py:1249 the function indexes `target_resolution[0]`, `target_resolution[1]`, `scale_factor[0]`, `scale_factor[1]` without checking the sequence length, and divides `abs(res_y) / target_resolution[...]` without checking the value.

**Four cases**

1. **Overlong tuple** -- `target_resolution=(10, 20, 30)` or `scale_factor=(2, 2, 2)` is accepted silently. The extra element is dropped.
2. **Short tuple** -- `target_resolution=(10,)` or `scale_factor=(2,)` raises `IndexError: tuple index out of range` from deep inside the function, with no hint about which parameter is wrong.
3. **Zero** -- `target_resolution=0` or `target_resolution=(0, 5)` raises `ZeroDivisionError` from the `abs(res_y) / target_resolution` line. `scale_factor=0` is already caught later by the `scale_y <= 0` check, but only after the geometry math runs.
4. **NaN / inf** -- `scale_factor=float('nan')` or `target_resolution=float('nan')` propagates through the math and fails much later with a numpy `ValueError: cannot convert float NaN to integer` when `_output_shape` tries to round.

**Expected behavior**

Each of the four cases should raise a `ValueError` at the top of `resample()`, before any geometry work, naming the parameter and the offending value. Roughly:

- `target_resolution=(10, 20, 30)` -> `ValueError: target_resolution must have length 2, got length 3`
- `scale_factor=(2,)` -> `ValueError: scale_factor must have length 2, got length 1`
- `target_resolution=0` -> `ValueError: target_resolution must be > 0, got 0`
- `scale_factor=float('nan')` -> `ValueError: scale_factor must be finite and > 0, got nan`

**Scope**

Validation belongs at the public boundary (`resample()`), not sprinkled through the per-backend helpers. The existing `scale_y <= 0 or scale_x <= 0` check stays as a defence-in-depth catch but should no longer be the first line of feedback the user sees.

**Additional context**

Surfaced by a code review pass over the resample module. Sibling issues in the input-validation label (#2566, #2568, #2558) follow the same pattern: validate inputs at the entry point, fail with a message that names the parameter.

brendancol:
Closes #2574.

## Summary

- Add `_validate_resample_scalar_or_pair` and call it at the top of `resample()`, right after the `(scale_factor XOR target_resolution)` check.
- The four leaky cases now raise `ValueError` with a message that names the parameter and shows the offending value, before any geometry math runs:
  - overlong tuple -> `target_resolution must have length 2, got length 3`
  - short tuple -> `scale_factor must have length 2, got length 1`
  - zero or negative -> `target_resolution must be > 0, got 0`
  - NaN / inf -> `scale_factor must be finite and > 0, got nan`
- Existing defence-in-depth `scale_y <= 0 or scale_x <= 0` check stays, but it is no longer the first thing the user sees.

## Backend coverage

Pure CPython input validation at the public boundary. No backend code touched. The helper runs once before dispatch, so numpy / cupy / dask+numpy / dask+cupy all behave the same.

## Test plan

- [x] New `xrspatial/tests/test_resample_input_validation_2574.py` with 17 cases (overlong, short, zero scalar, zero component, negative, NaN, inf for each of the two params, plus a regression block confirming valid scalar and tuple inputs still produce the expected output shape).
- [x] Update one existing `test_negative_scale` to match the new wording.
- [x] Full `test_resample*.py` suite: 249 passed, 2 skipped (pre-existing skips).

### DIFF
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -165,7 +165,8 @@ def _validate_resample_scalar_or_pair(value, param_name):
     positive. Raises ``ValueError`` with a message naming the parameter
     and the offending value.
     """
-    if isinstance(value, (tuple, list)):
+    is_pair = isinstance(value, (tuple, list))
+    if is_pair:
         if len(value) != 2:
             raise ValueError(
                 f"{param_name} must have length 2, got length {len(value)}"
@@ -174,21 +175,25 @@ def _validate_resample_scalar_or_pair(value, param_name):
     else:
         components = (value,)
 
-    for comp in components:
+    for i, comp in enumerate(components):
+        # Suffix points at the bad slot when the input was a pair, so
+        # `(0.0, 1.0)` reports "got 0.0 at index 0 of (0.0, 1.0)"
+        # instead of dumping the whole tuple.
+        where = f"{comp!r} at index {i} of {value!r}" if is_pair else f"{value!r}"
         try:
             f = float(comp)
         except (TypeError, ValueError):
             raise ValueError(
                 f"{param_name} must be a finite positive number "
-                f"(or length-2 sequence of them), got {value!r}"
-            )
+                f"(or length-2 sequence of them), got {where}"
+            ) from None
         if not np.isfinite(f):
             raise ValueError(
-                f"{param_name} must be finite and > 0, got {value!r}"
+                f"{param_name} must be finite and > 0, got {where}"
             )
         if f <= 0:
             raise ValueError(
-                f"{param_name} must be > 0, got {value!r}"
+                f"{param_name} must be > 0, got {where}"
             )
 
 
@@ -1257,6 +1262,14 @@ def resample(
         Output dtype matches the input float dtype (float32 or float64);
         integer inputs return float32 since NaN-sentinel resampling
         requires a float type.
+
+    Raises
+    ------
+    ValueError
+        If neither or both of ``scale_factor`` and ``target_resolution``
+        are given; if either is a sequence whose length is not 2; if any
+        component is zero, negative, NaN, or infinite; or if ``method``
+        is not in :data:`ALL_METHODS`.
     """
     _validate_raster(agg, func_name='resample', name='agg', ndim=(2, 3))
 
@@ -1300,6 +1313,11 @@ def resample(
     else:
         scale_y = scale_x = float(scale_factor)
 
+    # Defence-in-depth: the public inputs were already validated above
+    # by ``_validate_resample_scalar_or_pair``, so on the scale_factor
+    # path this branch is unreachable. It still fires on the
+    # target_resolution path if ``calc_res(agg)`` returns zero from a
+    # degenerate coord array.
     if scale_y <= 0 or scale_x <= 0:
         raise ValueError(
             f"Scale factors must be positive, got ({scale_y}, {scale_x})"
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -155,6 +155,43 @@ def _check_resample_gpu_memory(out_h, out_w):
         )
 
 
+# -- Input-validation helpers ------------------------------------------------
+
+def _validate_resample_scalar_or_pair(value, param_name):
+    """Validate a scalar-or-2-tuple resolution / scale parameter.
+
+    Accepts either a real scalar or a length-2 tuple/list of scalars.
+    Each component must be finite (not NaN, not inf) and strictly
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: resample input validation (#2590)

### Blockers
None.

### Suggestions

1. `xrspatial/resample.py` (around the new validation block) -- once the helper succeeds, the existing `scale_y <= 0 or scale_x <= 0` check a few lines below is unreachable on the `scale_factor` path, and on the `target_resolution` path it can only fire if `calc_res(agg)` returns zero (degenerate coords). The PR body says this check stays as defence-in-depth, which is a fair call. A one-line comment at that check explaining why a "must be positive" guard still fires after explicit positivity validation would save the next reader a confused minute.

2. `_validate_resample_scalar_or_pair` -- for a bad component inside a tuple (e.g. `target_resolution=(0.0, 1.0)`), the message says `target_resolution must be > 0, got (0.0, 1.0)`. The full tuple is visible, but the bad index is not called out. Pointing at the failing index (e.g. `got 0.0 at index 0 of (0.0, 1.0)`) would be slightly kinder. Not a blocker.

3. `resample()` docstring -- no Raises section. With four new ValueError cases pinned by tests, a short Raises block listing `ValueError` for malformed `scale_factor` / `target_resolution` would help anyone reading the docs.

### Nits

1. `_validate_resample_scalar_or_pair`, the `float(comp)` block -- the bare `except (TypeError, ValueError)` re-raises a `ValueError` and quietly chains the original. `raise ValueError(...) from None` suppresses the chain explicitly; `from e` keeps the cause but makes it intentional. Either reads cleaner than the implicit chain.

### What looks good
- Validation lives at the public boundary, not sprinkled through internal helpers. That matches the scope of the issue.
- Every ValueError names the parameter and shows the offending value, which is what the issue asked for.
- One helper covers both `scale_factor` and `target_resolution`, so the wording stays consistent between them.
- The new test file locks all four cases from the issue (overlong, short, zero, NaN) plus the obvious siblings (negative scalar, inf, zero component). The regression block confirms valid scalar and tuple inputs still produce the expected output shape.
- The one existing test that hard-coded the old "must be positive" wording was updated in the same commit, so no stale assertions remain.
- Test file name includes the issue number, matching the project convention (e.g. `test_resample_signature_annot_2544.py`).

### Checklist
- [x] Validation correct for all four documented cases plus negatives.
- [x] numpy / cupy / dask+numpy / dask+cupy unaffected (validation runs once at the entry, before dispatch).
- [x] NaN explicitly rejected by `np.isfinite`.
- [x] Edge cases (zero scalar, zero component, inf, negative) covered by tests.
- [n/a] No dask chunk-boundary changes.
- [n/a] No materialization or copy concerns.
- [n/a] No new benchmark needed (pure input validation).
- [n/a] No README feature-matrix change.
- [x] Helper has a docstring; `resample()` docstring could optionally gain a Raises block.

brendancol (commented):
## Follow-up review pass (#2590)

All four findings from the previous review pass were applied in commit 87004722.

| # | Severity | Finding | Disposition |
|---|----------|---------|-------------|
| 1 | Suggestion | One-line comment on the residual `scale_y <= 0 or scale_x <= 0` check | Fixed -- comment added explaining the defence-in-depth role and the one path on which it still fires (degenerate `calc_res`). |
| 2 | Suggestion | Pair errors should point at the failing index | Fixed -- bad-component errors now say `got 0.0 at index 0 of (0.0, 1.0)`. Scalar errors keep their original wording. |
| 3 | Suggestion | Add a Raises section to `resample()` docstring | Fixed -- short Raises block listing the four ValueError cases plus the existing `method` check. |
| 4 | Nit | Use `raise ... from None` to suppress the chained `float()` exception | Fixed -- the user now sees one clean ValueError, not a stacked pair. |

Tests still green: `test_resample.py` + `test_resample_input_validation_2574.py` = 195 passed. The two `test_zero_component` cases were tightened to lock the new "at index N of" wording.

No remaining findings.

PR #2591, opened 2026-05-28
reproject: reject explicit out-of-range nodata for integer dtypes

### LINKED ISSUE #2572: reproject: out-of-range explicit nodata corrupts integer output and contradicts attrs
**Describe the bug**

`reproject(raster, target_crs, nodata=...)` does not check that the explicit nodata value fits the source/output integer dtype. When it does not fit, the cast-back step in the worker wraps or clamps it (usually to 0), so out-of-bounds output pixels look the same as valid zero pixels. `attrs["nodata"]` still records the original out-of-range value, so the array and the metadata disagree.

Relevant lines:

- `xrspatial/reproject/_crs_utils.py:146` -- `_detect_nodata` lets any finite explicit nodata pass through unchanged, even when it does not fit the dtype.
- `xrspatial/reproject/__init__.py:435` -- the per-chunk worker clips the float result to `iinfo.min..iinfo.max` and casts to the integer dtype, so `-9999` written into a uint8 array lands at 0.

**Reproduction**

```python
import numpy as np
import xarray as xr

arr = np.ones((4, 4), dtype=np.uint8) * 10
da = xr.DataArray(
    arr,
    dims=["y", "x"],
    coords={"y": np.linspace(40, 30, 4), "x": np.linspace(-5, 5, 4)},
    attrs={"crs": "EPSG:4326"},
)
out = reproject(da, "EPSG:4326", nodata=-9999, bounds=(-20, 20, 20, 50))

print(out.dtype)                  # uint8
print(out.attrs["nodata"])        # -9999.0
print(np.unique(out.values))      # [0 10]
print(out.values[0, 0])           # 0, supposed to be nodata
```

Both 0 (valid input) and 0 (corrupted nodata) appear in the output with no way to tell them apart, and `attrs["nodata"]` advertises a value that does not match anything in the array.

**Expected behavior**

When an explicit nodata cannot be represented in the resolved dtype, reproject should fail fast with a clear `ValueError` naming the dtype and the offending value. This matches what `xrspatial.rasterize` already does when `fill=NaN` is passed with an integer dtype.

**Suggested fix**

In `_detect_nodata` (or at the API boundary in `reproject`), reject explicit integer nodata that falls outside `np.iinfo(dtype).min..np.iinfo(dtype).max`. NaN passed for an integer dtype is already swapped for a representable sentinel; this is the symmetric guard for the finite-but-out-of-range case.

Tests should cover:

- `uint8` with `nodata=-9999` raises
- `uint16` with `nodata=70000` raises
- `uint8` with `nodata=255` still works (boundary case)
- `int16` with `nodata=-32768` still works (in-range)

brendancol:
## Summary

- `reproject(..., nodata=X)` used to let any finite `X` through, even when it could not fit the source/output integer dtype. The worker then cast-back wrapped the sentinel (e.g. `-9999` into `uint8` became `0`), so out-of-bounds pixels were the same as valid zero pixels while `attrs["nodata"]` still advertised `-9999`.
- Explicit out-of-range values now raise `ValueError` with a message naming the dtype and the offending value. Matches the precedent in `xrspatial.rasterize` for `fill=NaN + integer dtype`.
- Attrs/rioxarray-derived out-of-range values (legacy files like `uint16 + nodata=-9999`) emit a `UserWarning` and fall back to the dtype-appropriate sentinel (signed `dtype.min`, unsigned `dtype.max`). This keeps existing reads working.

Closes #2572.

## Backend coverage

The fix is in `_detect_nodata`, which runs once at the API boundary on the host and is backend-agnostic. The end-to-end test covers the numpy path. Cupy/dask paths go through the same `_detect_nodata` call, so the validation gate fires before any backend dispatch.

## Test plan

- [x] New `TestDetectNodataDtypeRange` covers the five cases the issue asked for: uint8 + -9999 raises, uint16 + 70000 raises, uint8 + 255 boundary passes, int16 + -32768 boundary passes, attrs-derived out-of-range warns and falls back.
- [x] New `TestReprojectIntegerNodataDtypeRange` runs the same matrix end-to-end through `reproject()` and confirms the happy path actually writes the sentinel into the edge pixels (not 0).
- [x] Full `test_reproject.py` + `test_reproject_coverage_2026_05_27.py` suite still passes (373 tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@
 
 #### Fixed
 
+- `reproject` now rejects an explicit `nodata=` value that does not
+  fit the source/output integer dtype range. Previously the worker's
+  cast-back step silently wrapped the sentinel (e.g. `-9999` in a
+  `uint8` array landed at `0`), so out-of-bounds output pixels were
+  the same as valid zero pixels while `attrs['nodata']` still
+  advertised `-9999`. Explicit out-of-range values now raise
+  `ValueError`; attrs-derived out-of-range values (legacy files
+  such as `uint16 + nodata=-9999`) emit a `UserWarning` and fall
+  back to the dtype-appropriate sentinel. (#2572)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -680,7 +680,12 @@ def reproject(
     resampling : str
         One of 'nearest', 'bilinear', 'cubic'.
     nodata : float or None
-        Nodata value. Auto-detected if None.
+        Nodata value. Auto-detected if None. For integer input dtypes,
+        an explicit value that does not fit the dtype range raises
+        ``ValueError`` (e.g. ``nodata=-9999`` with a ``uint8`` raster).
+        Attrs/rioxarray-derived out-of-range values emit a
+        ``UserWarning`` and fall back to ``dtype.min`` for signed or
+        ``dtype.max`` for unsigned so legacy files still load (#2572).
     transform_precision : int
         Control-grid subdivisions for the coordinate transform (default 16).
         Higher values increase accuracy at the cost of more pyproj calls.
--- a/xrspatial/reproject/_crs_utils.py
+++ b/xrspatial/reproject/_crs_utils.py
@@ -5,6 +5,8 @@ then fall back to pyproj for codes/formats not in the built-in table.
 """
 from __future__ import annotations
 
+import warnings
+
 import numpy as np
 
 from xrspatial.reproject._lite_crs import CRS as LiteCRS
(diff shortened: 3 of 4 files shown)


brendancol (commented):
## PR Review: reproject: reject explicit out-of-range nodata for integer dtypes

None. The fix is correct: the range check is properly placed after the NaN swap, the error message preserves the user's original value (not the floatified one), and the validation runs at the API boundary so all four backends share the gate.

- `xrspatial/reproject/_crs_utils.py:182` -- `stacklevel=2` on the `warnings.warn` call points at the `_detect_nodata` call site itself, not at the user's `reproject()` call. The chain is user -> `reproject()` -> `_detect_nodata()` -> `warnings.warn`, so `stacklevel=3` would surface the warning on the user's `reproject(...)` line, which is the actionable location. The whole point of the warning is to tell the user their attrs are inconsistent with the dtype, so pointing at the user code matters here.

- `xrspatial/reproject/__init__.py:651` (not modified by this PR) -- the public `reproject()` docstring still says only "Nodata value. Auto-detected if None." with no mention of the new dtype-range validation. Adding a sentence ("Raises `ValueError` if the value cannot be represented in the source dtype.") would give users discoverable docs for the new contract. Optional, but matches what other functions in the project do for similar guards.

- `xrspatial/tests/test_reproject.py:200` -- the `test_reproject_uint8_in_range_nodata_writes_correct_pixels` test asserts `int(out.values[0, 0]) == 255` to check the edge pixel. That works, but a tighter assertion would be to count the nodata pixels and confirm at least the border ring is filled. Not worth a rev on its own.

### What looks good

- The strict-vs-permissive split (explicit raises, attrs warns) matches the precedent set in `xrspatial.geotiff._decode._resolve_masked_fill` for legacy files with out-of-range sentinels and `xrspatial.rasterize`'s `ValueError` for `fill=NaN + integer dtype`.
- Error message is specific and actionable: it names the dtype, the offending value, the valid range, and the concrete failure mode (`-9999 into uint8 becomes 0`).
- The existing test `test_detect_nodata_explicit` (no dtype arg) still passes, confirming the new code only activates when dtype is provided.
- CHANGELOG entry is in the right section and references the issue.

### Checklist
- [x] Algorithm matches reference/paper -- this is a validation gate, no algorithm to match
- [x] All implemented backends produce consistent results -- gate runs at API boundary upstream of dispatch
- [x] NaN handling is correct -- NaN swap precedes the range check, so NaN never reaches the new branch
- [x] Edge cases are covered by tests -- boundary cases (dtype.max for uint8, dtype.min for int16) included
- [x] Dask chunk boundaries handled correctly -- gate runs on a scalar, no chunk concern
- [x] Benchmark exists or is not needed -- input validation, no benchmark needed
- [x] README feature matrix updated (if applicable) -- N/A for a bug fix
- [x] Docstrings present and accurate -- `_detect_nodata` docstring updated; public `reproject()` docstring not updated (see Suggestion 2)

brendancol (commented):
## Follow-up review after 53f33ee

All three previous-review items applied:

- Suggestion 1 (stacklevel=3): fixed in `_crs_utils.py:184`. The warning now surfaces at the user's `reproject()` call.
- Suggestion 2 (docstring): fixed in `__init__.py:683-688`. The `nodata` parameter doc now mentions the new validation behavior with both the raise and the fallback paths.
- Nit 1 (tighter test assertion): fixed in `test_reproject.py:205-211`. The happy-path test now asserts the full top row is the sentinel, not just one pixel.

Tests still pass: 9/9 new + 354/354 pre-existing.

No new findings.

PR #2592, opened 2026-05-28
resample: preserve integer precision in nodata mask comparison (#2570)

### LINKED ISSUE #2570: resample: int64 nodata mask loses precision via float cast
**Describe the bug**

`_resolve_nodata` in `xrspatial/resample.py` (line 1148) casts the nodata sentinel through `float()`, then `_apply_nodata_mask` (line 1168) compares the data array against that float. For int64 values above 2**53 the cast is lossy, so distinct integers collide after rounding and valid cells get masked to NaN.

**Repro**

```python
import numpy as np
import xarray as xr

# nodata = 2**60 + 1, valid cell at 2**60 are distinct in int64.
nodata = (1 << 60) + 1
data = np.full((4, 4), nodata, dtype=np.int64)
data[1, 1] = 1 << 60  # valid value, not nodata

agg = xr.DataArray(data, dims=('y', 'x'))
out = resample(agg, scale_factor=1.0, method='nearest', nodata=nodata)

# The valid 2**60 cell comes back as NaN because float(2**60+1) == 2**60.
print(out.values[1, 1])  # nan, even though the input was valid
```

**Expected behavior**

The cell holding `2**60` should survive masking. Only cells whose integer value equals the sentinel `2**60 + 1` should become NaN.

**Additional context**

The float round-trip drops the low bits whenever the magnitude exceeds 2**53 (the float64 mantissa width). Fix: keep the sentinel in the input's native dtype when the input is integer, and only cast through float for floating-point inputs (where NaN semantics matter). The masking helper feeds every backend (numpy / cupy / dask+numpy / dask+cupy), so the change lives in one place.

brendancol:
Closes #2570.

## Summary

- `_resolve_nodata` no longer routes integer sentinels through `float()`; it casts to the input dtype instead so the equality test in `_apply_nodata_mask` runs in integer space.
- For int64 sentinels above 2**53 (e.g. `nodata = 2**60 + 1`), valid cells that share the rounded float representation (e.g. `2**60`) are no longer masked out.
- Float inputs still route through `float()` so the existing NaN short-circuit keeps working.

## Backend coverage

The fix lives in the shared masking helper that every backend goes through (numpy / cupy / dask+numpy / dask+cupy). Verified on numpy and dask+numpy locally.

## Test plan

- [x] New `TestNodataIntegerPrecision` class in `xrspatial/tests/test_resample.py` covering:
  - int64 sentinel above 2**53 via explicit kwarg
  - int64 sentinel via `_FillValue` attr
  - int32 sentinel (negative min)
  - uint16 sentinel (max)
  - float NaN sentinel short-circuit still works
  - float finite sentinel still works
- [x] Pre-existing `TestNodata` cases still pass
- [x] Full `test_resample.py` suite (184 tests) green
- [x] `test_resample_coverage_2026_05_27.py` and `test_resample_signature_annot_2544.py` green

### DIFF
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -1158,11 +1158,15 @@ def _resolve_nodata(agg, nodata):
     # dtype so the equality test in _apply_nodata_mask compares
     # integer-to-integer. A NaN sentinel can never match an integer
     # value, so signal a no-op mask by returning NaN unchanged.
-    try:
-        if isinstance(nodata, float) and np.isnan(nodata):
-            return float('nan')
-    except TypeError:
-        pass
+    if isinstance(nodata, float) and np.isnan(nodata):
+        return float('nan')
+    # Reject fractional float sentinels for integer inputs -- silently
+    # truncating to int would mask cells the caller never asked to mask.
+    if isinstance(nodata, float) and not nodata.is_integer():
+        raise ValueError(
+            f"nodata={nodata!r} is not representable in integer dtype "
+            f"{agg.dtype}; pass an integer sentinel instead."
+        )
     return np.asarray(nodata).astype(agg.dtype).item()
 
 
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -1134,8 +1134,12 @@ def _resolve_nodata(agg, nodata):
     ``nodata`` in ``agg.attrs``. Returns ``None`` when no sentinel was
     found (the caller skips the masking step).
 
-    NaN sentinels are returned as NaN so the caller can branch on
-    ``np.isnan`` rather than ``==`` (which never matches NaN).
+    For floating-point inputs the sentinel is returned as a Python
+    ``float`` so the caller can branch on ``np.isnan`` rather than
+    ``==`` (which never matches NaN). For integer / bool inputs the
+    sentinel is cast to the input dtype so the comparison happens in
+    integer space -- routing it through ``float`` would lose precision
+    for int64 values above 2**53.
     """
     if nodata is None:
         for key in ('_FillValue', 'nodata'):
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: resample integer nodata precision fix

None.

- `xrspatial/resample.py:1162-1165` -- the `try/except TypeError` around the NaN check is dead code. `isinstance(nodata, float)` already gates the path, and `np.isnan` on a Python float will never raise TypeError. Drop the try wrapper and keep the `if isinstance(...) and np.isnan(...)` check on its own line.

- `xrspatial/resample.py:1166` -- `np.asarray(nodata).astype(agg.dtype).item()` will silently truncate a non-integer float sentinel (e.g. `nodata=-9999.5` on int32 data becomes -9999). Realistically nobody passes a fractional sentinel, but a one-line check like `if isinstance(nodata, float) and not nodata.is_integer(): raise ValueError(...)` would catch the obvious user error. Skip if you'd rather keep the helper minimal.
- `xrspatial/tests/test_resample.py:1456` -- `test_int64_sentinel_above_2pow53_preserves_distinct_values` is the headline regression test. Consider asserting `out.values[1, 1] == np.float32(1 << 60)` so a future regression that masks the cell to NaN AND a regression that swaps the surviving value both get caught. The current `np.isfinite` check is correct but a bit loose.

### What looks good
- The split between float-input and integer-input paths in `_resolve_nodata` is the right call. It keeps the float NaN short-circuit working while routing integer comparisons through the input dtype.
- `_apply_nodata_mask` computes the mask before promoting to float32, which is exactly the surgical change the bug needs. The mask carries over correctly to `agg.where(mask)` after the cast.
- Backend coverage is right: the masking helper sits upstream of every per-backend dispatch path, so numpy, cupy, dask+numpy, and dask+cupy all inherit the fix automatically. The PR body confirms numpy and dask+numpy were exercised locally.
- Test coverage hits the exact failure mode (int64 sentinel above 2**53), the explicit-kwarg and attr paths, and the unaffected float branches.

### Checklist
- [x] Algorithm matches reference / intent (integer equality in input dtype)
- [x] All implemented backends produce consistent results (single shared helper)
- [x] NaN handling is correct (float NaN short-circuit preserved; NaN sentinel on int data is a no-op)
- [x] Edge cases are covered by tests (int64 > 2**53, int32, uint16, float NaN, float finite)
- [x] No dask chunk boundary concerns (no neighborhood op in this change)
- [ ] Benchmark: not applicable for a correctness fix in a helper
- [ ] README feature matrix: not applicable, no new function or backend
- [x] Docstrings updated to reflect the new dispatch on input dtype

brendancol (commented):
## Follow-up review

Re-ran the review after the address-nits commit (1adea6f1). All previous findings are resolved:

- Suggestion (dead try/except): FIXED -- the wrapper is gone; the `isinstance` + `np.isnan` check stands alone.
- Nit (silent fractional truncation): FIXED -- `_resolve_nodata` now raises `ValueError("nodata=... is not representable in integer dtype ...")` when a non-integer float sentinel is passed for an integer input. Added `test_fractional_float_sentinel_on_int_input_raises` to lock it in.
- Nit (loose `isfinite` assertion): FIXED -- the int64-above-2**53 regression test now asserts the surviving cell equals `np.float32(1 << 60)`, so masking-to-NaN and value-swap regressions both get caught.

Full `test_resample.py` suite passes locally (185 tests). No remaining blockers, suggestions, or nits.

PR #2593, opened 2026-05-28
zonal: keep hypsometric_integral dask path fully lazy (#2563)

### LINKED ISSUE #2563: hypsometric_integral() triggers eager dask.compute() on dask inputs
**Describe the bug**

`hypsometric_integral()` in `xrspatial/zonal.py` calls `dask.compute()` inline (around line 1544) on its dask backend. Constructing the result on a dask-backed DataArray eagerly evaluates part of the graph at call time, instead of staying lazy until the caller asks for `.compute()` or `.persist()`.

The eager step is the per-zone reduce that produces the HI lookup dict. It lives inside `_hi_dask_numpy`, so it also affects the dask+cupy backend (which delegates through `_hi_dask_numpy`).

**Expected behavior**

Calling `hypsometric_integral()` on a dask-backed DataArray should return a DataArray whose `.data` is a `dask.array.Array`, with no part of the graph running until the caller materializes the result. That matches the convention used by other dask-backed functions in xrspatial.

**Reproduction**

```python
import dask.array as da
import numpy as np
import xarray as xr
from xrspatial.zonal import hypsometric_integral

zones = xr.DataArray(da.from_array(np.array([[1, 1, 2], [2, 2, 1]]), chunks=2))
values = xr.DataArray(da.from_array(np.array([[10.0, 20, 5], [30, 40, 15]]), chunks=2))

out = hypsometric_integral(zones, values, nodata=None)
# Today: a compute already fired inside the call.
# Expected: no compute fires until out.compute() is invoked.
```

**Scope**

Only the dask branch is affected. The numpy and cupy backends return eagerly because their inputs are eager. Commit `29b82839` fixed a separate dask+cupy correctness issue (#2525) on the same code path; that fix has to stay intact.

**Additional context**

The integral is painted back to a raster of the same shape as the input. The lookup dict that drives the paint is the part being computed eagerly. The dict can be kept as a `dask.delayed` and threaded into `map_blocks` so the whole graph stays lazy.

brendancol:
Closes #2563.

## Summary

- `_hi_dask_numpy` was calling `dask.compute()` inline on the per-zone reduce so it could pass the HI lookup dict to `map_blocks`. That broke the lazy convention for dask-backed results.
- Wrap the reduce in a 0-d dask object array (via `da.from_delayed`) and thread it through `map_blocks` as a positional dask arg. The lookup flows through the graph alongside the chunks; nothing computes until the caller asks for it.
- Drop the empty-lookup early return — painting over an empty dict already produces the all-NaN raster that case needs.
- The dask+cupy backend delegates to `_hi_dask_numpy`, so the same fix applies; the #2525 re-wrap step is preserved.

## Backend coverage

- numpy: unchanged
- cupy: unchanged
- dask+numpy: now returns a lazy dask graph
- dask+cupy: now returns a lazy dask graph (still cupy-typed chunks)

## Test plan

- [x] `pytest xrspatial/tests/test_hypsometric_integral.py` (36 passed locally, including 3 new laziness/parity tests)
- [x] `pytest xrspatial/tests/test_zonal.py xrspatial/tests/test_zonal_backend_coverage_2026_05_27.py xrspatial/tests/test_hypsometric_integral.py` (204 passed locally)
- [x] Manual: dask+cupy returns lazy graph and computes to the same values as numpy

### DIFF
--- a/xrspatial/tests/test_hypsometric_integral.py
+++ b/xrspatial/tests/test_hypsometric_integral.py
@@ -230,7 +230,7 @@ def test_hypsometric_integral_accessor(backend, hi_zones, hi_values):
 # --- laziness (issue #2563) --------------------------------------------------
 
 @pytest.mark.skipif(da is None, reason="dask not installed")
-def test_hypsometric_integral_dask_returns_lazy():
+def test_hypsometric_integral_dask_returns_lazy(monkeypatch):
     """The dask backend must return a lazy graph — no eager compute on call."""
     import dask
     from xrspatial.zonal import hypsometric_integral
@@ -251,11 +251,9 @@ def test_hypsometric_integral_dask_returns_lazy():
         calls['n'] += 1
         return orig_compute(*args, **kwargs)
 
-    dask.compute = counting_compute
-    try:
-        result = hypsometric_integral(zones, values, nodata=None)
-    finally:
-        dask.compute = orig_compute
+    monkeypatch.setattr(dask, "compute", counting_compute)
+
+    result = hypsometric_integral(zones, values, nodata=None)
 
     assert isinstance(result.data, da.Array), \
         f"expected dask.Array, got {type(result.data).__name__}"
@@ -293,7 +291,7 @@ def test_hypsometric_integral_dask_matches_numpy():
 
 
 @pytest.mark.skipif(da is None, reason="dask not installed")
-def test_hypsometric_integral_dask_empty_lookup_is_lazy():
+def test_hypsometric_integral_dask_empty_lookup_is_lazy(monkeypatch):
     """When every zone is nodata, the result is still a lazy dask array of NaN."""
     import dask
     from xrspatial.zonal import hypsometric_integral
@@ -314,11 +312,9 @@ def test_hypsometric_integral_dask_empty_lookup_is_lazy():
         calls['n'] += 1
         return orig_compute(*args, **kwargs)
 
-    dask.compute = counting_compute
-    try:
-        result = hypsometric_integral(zones, values, nodata=0)
-    finally:
-        dask.compute = orig_compute
+    monkeypatch.setattr(dask, "compute", counting_compute)
+
+    result = hypsometric_integral(zones, values, nodata=0)
 
     assert isinstance(result.data, da.Array)
     assert calls['n'] == 0
--- a/xrspatial/tests/test_hypsometric_integral.py
+++ b/xrspatial/tests/test_hypsometric_integral.py
@@ -227,6 +227,105 @@ def test_hypsometric_integral_accessor(backend, hi_zones, hi_values):
     np.testing.assert_allclose(out[z1_mask], 0.5, rtol=1e-10)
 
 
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review

### Blockers
None.

### Suggestions
- `xrspatial/tests/test_hypsometric_integral.py:243-248`, `:304-309`: the laziness tests monkey-patch `dask.compute` directly on the module. The counter is global state, so a parallel test that calls `dask.compute` would race. Using `monkeypatch.setattr("dask.compute", counting_compute)` would have pytest restore it deterministically per test, even on exception.

### Nits
- `xrspatial/zonal.py:1563-1564`: the `hi_map_arr[()] if hi_map_arr.shape == () else hi_map_arr.item()` branch is unreachable here — the lookup is always a 0-d dask object array, so `map_blocks` always hands `_paint` a 0-d numpy chunk. Could simplify to `hi_map = hi_map_arr[()]`.
- `xrspatial/zonal.py:1528-1538`: `_hi_lookup_as_object_array` is `@delayed` and then immediately wrapped in `da.from_delayed`. `_hi_reduce` already returns a `Delayed`, so this could be collapsed into one step (make the reduce return the object array directly). Minor — current form is more readable.

### What looks good
- The fix is minimal and targeted: one helper, one `map_blocks` rewiring, no shape or backend surface changes.
- The dask+cupy path is unchanged — it still delegates through `_hi_dask_numpy` and inherits the lazy behaviour for free, so the #2525 re-wrap fix stays intact.
- Tests cover both halves of the contract (lazy on construction; equal to numpy on compute) plus the empty-lookup edge case.
- Removing the empty-lookup early return is fine — painting over an empty dict produces the same all-NaN raster.

### Checklist
- [x] Algorithm matches reference (HI = (mean - min) / (max - min) per zone — unchanged)
- [x] NaN handling correct (flat zones, all-NaN zones, NaN cells)
- [x] Edge cases covered (empty lookup, flat zone, single cell, all-nan)
- [x] Dask chunk boundaries fine (no neighborhood op; per-chunk paint is independent)
- [x] No premature materialization (this PR removes one)
- [ ] Benchmark — N/A, laziness fix, no perf-critical hot path changed
- [ ] README feature matrix — N/A, no new function or backend change
- [x] Docstrings accurate (lazy guarantee documented)

brendancol (commented):
## Follow-up review (after 52682bc8)

- Suggestion (monkeypatch dask.compute): **fixed** — both laziness tests now use `monkeypatch.setattr` so pytest restores `dask.compute` per test.
- Nit (dead `.item()` branch in `_paint`): **fixed** — simplified to `hi_map = hi_map_arr[()]` with a short comment.
- Nit (collapse the two delayed nodes): **dismissed** — restructuring `_hi_reduce` to return the wrapped object array would either change its semantics for unclear gain or require a less readable `dask.delayed(...)(...)` call at the use site. Same node count, worse readability.

### New findings
None. All 36 hypsometric tests still pass locally. No new concerns introduced by the follow-up.

PR #2594, opened 2026-05-28
zonal: fix trim() default NaN sentinel on numpy and cupy backends

### LINKED ISSUE #2559: trim() ignores default values=(np.nan,) on numpy/cupy paths
**Describe the bug**

`xrspatial.zonal.trim()` defaults to `values=(np.nan,)`, but the numpy
and cupy paths cannot match a NaN sentinel. The numba inner loop in
`_trim` (zonal.py line 2283) uses `e == val`, and `NaN == NaN` is
always False, so a NaN-framed raster is returned unchanged on those
backends.

The dask branch `_trim_bounds_dask` (zonal.py line 2352) has a
dedicated `da.isnan` path, so the same input trims correctly under
dask+numpy and dask+cupy. The result: `trim()` returns different
shapes for the same input depending on backend when the default
`values` is used.

**Reproduction**

```python
import numpy as np
import xarray as xr
from xrspatial.zonal import trim

arr = xr.DataArray(np.array([
    [np.nan, np.nan, np.nan],
    [np.nan, 1.0,    np.nan],
    [np.nan, np.nan, np.nan],
]))

trim(arr).shape   # numpy: (3, 3) -- nothing trimmed
                  # expected: (1, 1)
```

The existing backend-coverage test `test_trim_dask_nan_values` in
`xrspatial/tests/test_zonal_backend_coverage_2026_05_27.py` already
pins this asymmetry as a known divergence.

**Expected behavior**

Calling `trim()` with the default `values=(np.nan,)` should trim a
NaN-framed raster on all four backends (numpy, cupy, dask+numpy,
dask+cupy) and produce the same output.

**Fix direction**

In `_trim` (numba CPU kernel) detect NaN sentinels with `np.isnan(val)`
rather than `e == val`. The cupy path routes through the same numba
kernel via `data.get()`, so one fix covers both. Behavior for non-NaN
sentinels (zero, integer zone ids, etc.) is unchanged.

Add cross-backend tests using the project's `_to_numpy` /
`general_checks` helpers that pass `np.nan` (both default and
explicit) and confirm all four backends agree.

brendancol:
Closes #2559.

## Summary

- Fix `zonal.trim()` so the default `values=(np.nan,)` actually trims
  a NaN-framed raster on the numpy and cupy backends. The numba kernel
  `_trim` matched sentinels with `e == val`, and `NaN == NaN` is
  False, so NaN borders were silently ignored. The dask branch
  already handled NaN via `da.isnan`, so the same input behaved
  differently across backends.
- Add `_trim_bounds_numpy` (a row/col `.all()` reduction on a
  NaN-aware mask, mirroring `_trim_bounds_dask`) and route NaN
  sentinels through it from the `trim()` wrapper. Finite sentinels
  still go through the existing numba kernel; `_trim` is unchanged.
- Update the pre-existing asymmetry-pin test in
  `test_zonal_backend_coverage_2026_05_27.py` and add cross-backend
  tests covering numpy / cupy / dask+numpy / dask+cupy with a NaN
  sentinel, plus mixed NaN+finite sentinels, integer-dtype regression,
  and all-NaN input.

## Backend coverage

- numpy: fixed, new tests
- cupy: fixed (routes through the same wrapper), new tests
  (skipped without CUDA)
- dask+numpy: unchanged behaviour, new parity test
- dask+cupy: unchanged behaviour, new parity test
  (skipped without CUDA)

## Test plan

- [x] `pytest xrspatial/tests/test_zonal.py -k trim`
- [x] `pytest xrspatial/tests/test_zonal_backend_coverage_2026_05_27.py -k trim`
- [x] `pytest xrspatial/tests/test_zonal.py xrspatial/tests/test_zonal_backend_coverage_2026_05_27.py xrspatial/tests/test_dask_cupy_gaps.py`
- [x] Repro snippet in the issue now returns shape (1, 1) on numpy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@
 
 #### Fixed
 
+- `zonal.trim()` with the default `values=(np.nan,)` (or any NaN
+  sentinel) now trims a NaN-framed raster on the numpy and cupy
+  backends, matching the existing dask behaviour. The numba kernel
+  `_trim` matched sentinels with `e == val` and `NaN == NaN` is False,
+  so NaN borders were silently ignored on those backends. NaN matching
+  is now handled in a numpy bounds helper that mirrors the dask
+  reduction. Behaviour for finite sentinels (zero, integer zone ids,
+  etc.) is unchanged. (#2559)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/tests/test_zonal_backend_coverage_2026_05_27.py
+++ b/xrspatial/tests/test_zonal_backend_coverage_2026_05_27.py
@@ -334,12 +334,12 @@ def test_trim_dask_cupy_matches_numpy(trim_input):
 
 @dask_required
 def test_trim_dask_nan_values(trim_input):
-    """Cat 2 HIGH: dask trim with default NaN sentinel.
+    """trim with a NaN sentinel agrees across numpy and dask backends.
 
-    The dask branch ``_trim_bounds_dask`` has a dedicated ``isnan`` path
-    (zonal.py:2287) that the numpy backend lacks: numpy's ``_trim`` uses
-    a plain equality check, which never matches NaN. This test pins the
-    backend asymmetry so any future change is visible as a diff.
+    Originally pinned a numpy/dask asymmetry: ``_trim``'s ``e == val``
+    check never matched NaN, so the numpy path left a NaN-framed
+    raster unchanged while dask trimmed it. Fixed in #2559 by routing
+    NaN sentinels through ``_trim_bounds_numpy`` in the wrapper.
     """
     arr_with_nan = np.where(trim_input == 0, np.nan, trim_input)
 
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: zonal: fix trim() default NaN sentinel on numpy and cupy backends

None.

None.

- `xrspatial/zonal.py:2377-2392` -- `_split_nan_excludes` uses
  `isinstance(v, float) and np.isnan(v)` to detect a NaN sentinel.
  This matches the existing convention in `_trim_bounds_dask`
  (zonal.py:2352), so the inconsistency is symmetric across the two
  helpers. Worth noting that `np.float32(np.nan)` would slip through
  because `np.float32` is not a subclass of `float`, but the dask
  path has the same gap, so the PR is not introducing new asymmetry.
  Not worth fixing in this PR.

- `xrspatial/zonal.py:2394-2414` -- `_trim_bounds_numpy` allocates an
  `O(rows*cols)` boolean mask, which is heavier than the existing
  numba kernel that only scans the edges. For typical sentinel
  counts (one or two) the bool mask cost is negligible against the
  cost of the data load; for very large rasters with a long list of
  sentinels this could become noticeable. Acceptable for the bug fix
  scope; a future PR could add an edge-walk numpy variant if
  measurements show it matters.

- `xrspatial/zonal.py:2530-2540` -- the `# NaN sentinels cannot be
  matched by the numba kernel` comment is helpful. Consider also
  mentioning that the same routing is what the dask path does via
  `_trim_bounds_dask`, so a reader can find the two NaN-aware paths
  by following one comment. Not necessary.

### What looks good

- Fix scope is tight. `_trim` numba kernel is unchanged, so finite
  sentinels still take the fast path. Only the NaN case goes through
  the numpy reduction.
- `_trim_bounds_numpy` mirrors the existing `_trim_bounds_dask`
  reduction, so numpy and dask now compute trim bounds from the same
  logical operation (NaN-aware mask -> row/col `.all()` -> nonzero
  indices). The dask asymmetry-pin test was updated to assert
  agreement instead of asymmetry.
- New tests cover all four backends (numpy / cupy / dask+numpy /
  dask+cupy) with the appropriate `cuda_and_cupy_available` and
  `dask_required` decorators. There are also tests for mixed
  NaN+finite sentinels, integer-dtype regression (the case that
  motivated routing through the wrapper rather than calling
  `np.isnan(val)` inside the numba kernel), and all-NaN input.
- CHANGELOG entry is concrete about what changed and why, and
  references the issue.
- Empty-slice semantics for all-NaN input (shape (0, 0)) match the
  pre-existing dask path's empty-slice return values, so the two
  paths agree at that edge too.

### Checklist

- [x] Algorithm matches reference (the reduction mirrors
  `_trim_bounds_dask`).
- [x] All implemented backends produce consistent results (new
  parity tests).
- [x] NaN handling is correct (`np.isnan` mask, guarded on float
  dtype).
- [x] Edge cases are covered by tests (all-NaN, integer dtype,
  mixed sentinels, default vs explicit NaN).
- [x] Dask chunk boundaries handled correctly (dask path unchanged).
- [x] No premature materialization or unnecessary copies (cupy path
  already does `data.get()` before reaching the wrapper; the new
  bool mask is the only added allocation and it is bounded by the
  input shape).
- [x] Benchmark not required for a sentinel-handling bug fix.
- [x] README feature matrix unchanged (no new function, no backend
  support change).
- [x] Docstrings are accurate (the existing `trim()` docstring
  already documented the default).

PR #2595, opened 2026-05-28
rasterize: support descending-x in like= templates (#2568)

### LINKED ISSUE #2568: rasterize: descending-x like silently mislabels output coords
**Describe the bug**

`rasterize(like=...)` silently produces spatially wrong output when the template `like` DataArray has a descending x-axis (largest x first, smallest last).

The comment at `xrspatial/rasterize.py:2936` says "descending-x templates would hit the same bug class and are not supported here," but the code neither rejects descending-x nor flips columns to compensate. `_extract_grid_from_like` notes the issue and moves on.

Ascending y is already handled at line 3349-3350: the burned array is flipped along axis 0 so `result.sel(y=...)` lines up with the geometry in world coordinates. There's no equivalent for x.

When a descending-x `like` is passed, the rasterizer writes pixels using ascending-x column order (column 0 = xmin), but `reuse_like_coords` at line 3340 assigns the descending-x coord array unchanged. A polygon at world x=0.5 ends up labelled under coord x=3.5.

**Expected behavior**

`result.sel(x=world_x)` should return the burned value at `world_x` regardless of whether `like.x` is ascending or descending. Either flip the output along axis 1 to match the coords, or reject descending-x with a clear `ValueError` (the existing comment already claims it is unsupported).

**Reproducer**

```python
import numpy as np
import xarray as xr

# Descending-x like grid
x_desc = np.linspace(3.5, 0.5, 4)  # descending
y = np.linspace(3.5, 0.5, 4)
like = xr.DataArray(
    np.zeros((4, 4)), dims=['y', 'x'], coords={'y': y, 'x': x_desc},
)

# Burn a 1x1 box at world (0.5, 0.5) -- the lower-left corner
result = rasterize([(box(0, 0, 1, 1), 1.0)], like=like, fill=0)

# Expected: result.sel(x=0.5, y=0.5).item() == 1.0
# Actual:   result.sel(x=0.5, y=0.5).item() == 0.0
#           result.sel(x=3.5, y=0.5).item() == 1.0  (mislabelled)
```

**Impact**

Any coord-aware downstream op (`xr.align`, `.sel`, slope/aspect, zonal stats keyed off coords) operates on the wrong cells. The output looks plausible but is spatially wrong. No exception is raised.

**Additional context**

- Same bug class as #2170 (ascending-y), which was fixed by the axis-0 flip.
- Fix should be small and local: detect descending-x in `_extract_grid_from_like`, carry it on `_LikeGrid`, and flip axis 1 in `rasterize` when reusing coords. Or reject with `ValueError`.
- Worth confirming descending-y is genuinely covered by the existing y-flip path while you're in there.

brendancol:
Closes #2568.

## Summary

- `rasterize(like=...)` silently mislabelled output coords when the template's x axis was descending. The burner writes column 0 = xmin (image convention), but `reuse_like_coords` assigned the descending x-coord unchanged, so a polygon at world x=0.5 ended up under coord x=3.5.
- Fix: detect descending x in `_extract_grid_from_like`, carry it on `_LikeGrid`, and flip the burned array along axis 1 in `rasterize` when reusing coords. Same shape as the existing ascending-y axis-0 flip.
- Backend coverage: numpy, cupy, dask+numpy, dask+cupy. All four use the same slicing semantics, so one flip line covers everything.

## Why "support" rather than "reject"

The y-axis already supports both ascending and descending transparently via the axis-0 flip (added for #2170). Treating x asymmetrically would be surprising for users who get a descending-x `like` from a pipeline that flipped a raster (e.g. negative-stride affine transforms). The fix is small and local, and it matches the existing pattern. Rejecting with `ValueError` was the other reasonable option; this PR picks the symmetric one.

Descending y was investigated while in the area and is already correctly handled (it is the default image convention -- the burner emits row 0 = ymax natively).

## Test plan

- [x] New tests at `xrspatial/tests/test_rasterize_descending_x_2568.py` cover polygons, points, lines, world-coord round-trip with `xr.align`, and both axes flipped together.
- [x] Tests run against numpy, dask+numpy, cupy, and dask+cupy.
- [x] Existing `xrspatial/tests/test_rasterize.py` (215 tests) still passes.
- [x] Related coverage suites (`test_rasterize_accuracy.py`, `test_rasterize_coverage_*.py`) still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 #### Fixed
 
+- `rasterize(like=...)` no longer silently mislabels output when the
+  template's x axis is descending. Previously the burned array was
+  written ascending-x (column 0 = xmin) but `reuse_like_coords`
+  assigned the descending x-coord unchanged, so a polygon at world
+  x=0.5 landed under coord x=3.5. The array is now flipped along
+  axis 1 so `result.sel(x=...)` agrees with the geometry's world
+  coordinates, mirroring the existing ascending-y flip on axis 0.
+  Works for numpy, cupy, dask+numpy, and dask+cupy. (#2568)
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -2935,7 +2935,9 @@ def _extract_grid_from_like(like):
     # up against the wrong cells.  Only the first and last samples are
     # inspected; the ``_check_uniform_axis`` call above has already
     # rejected non-monotonic or duplicate-valued coords on both axes,
-    # so this is safe.
+    # so this is safe.  Single-row / single-column templates have no
+    # direction to detect, so the ``> 1`` guard short-circuits to
+    # ``False`` (no flip) for those.
     y_ascending = height > 1 and float(y[-1]) > float(y[0])
     x_descending = width > 1 and float(x[-1]) < float(x[0])
 
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -2874,6 +2874,7 @@ class _LikeGrid(NamedTuple):
     extra_coords: dict
     attrs: dict
     y_ascending: bool
+    x_descending: bool
 
 
 def _extract_grid_from_like(like):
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: rasterize: support descending-x in like= templates (#2568)

None.

None.

- `xrspatial/tests/test_rasterize_descending_x_2568.py:42`: `_like_2568` has a `y_descending=True` parameter that only gets flipped in one test (`test_numpy_descending_x_and_ascending_y`). The asymmetry is harmless. If you want fewer one-off helpers later, this could fold into a parameterized test pair.
- `xrspatial/rasterize.py:2940`: `x_descending = width > 1 and float(x[-1]) < float(x[0])` mirrors the y case. Both correctly treat width/height == 1 as "no flip needed," but a comment noting that single-column templates short-circuit would help the next reader.

### What looks good
- The fix is symmetric with the existing y-axis treatment: same comment shape, same `_LikeGrid` field pattern, same plumbing through `rasterize`, same slicing flip.
- The `reuse_like_coords` gate is unchanged, so the flip only runs in the path where it matters (template grid reused bit-identically). When the caller passes explicit bounds/width/height/resolution, the linspace-rebuild path produces ascending coords as before. `test_numpy_explicit_bounds_skips_flip` covers that.
- Slicing (`out[:, ::-1]`) works the same way for numpy, dask, cupy, and dask+cupy, so no per-backend code is needed.
- Test matrix hits all four backends, plus polygon/point/line geometry types, plus xr.align round-trip, plus the combined ascending-y + descending-x case.

### Checklist
- [x] Algorithm matches reference/paper (n/a, bug fix to existing internal logic)
- [x] All implemented backends produce consistent results (covered by tests)
- [x] NaN handling is correct (unchanged path; fill semantics preserved)
- [x] Dask chunk boundaries handled correctly (flip is a view, no chunk impact)
- [x] No premature materialization or unnecessary copies (slice view, lazy under dask)
- [x] Benchmark exists or is not needed (n/a for a coord-orientation fix)
- [x] Docstrings present and accurate (no new public API; existing docstrings unaffected)

brendancol (commented):
## Follow-up review (after 2f5c8ce5)

### Disposition of the previous nits
- Nit 1 (parameterize `_like_2568` y_descending): dismissed. The current helper lets one test cover both-axes-flipped without bloating every other test signature. Folding it into a fixture would be more code for the same coverage.
- Nit 2 (single-column short-circuit comment): fixed in 2f5c8ce5. The detection comment now spells out that ``width/height > 1`` is the short-circuit guard.

### New findings
None. Follow-up is comment-only.

PR #2596, opened 2026-05-28
reproject: fix vertical-datum shift crash on integer DEMs

### LINKED ISSUE #2565: reproject: vertical datum shift crashes on integer DEMs
**Describe the bug**

`reproject(..., src_vertical_crs="EGM96", tgt_vertical_crs="ellipsoidal")` raises `UFuncTypeError` whenever the source raster has an integer dtype. The vertical-shift path copies the data at its original (integer) dtype on `xrspatial/reproject/__init__.py:1180`, then later does a float in-place add on `xrspatial/reproject/__init__.py:1228`:

```python
strip_data[is_valid] += N_total[is_valid]
```

NumPy refuses to assign a float result back into an integer array, so the call fails. DEMs in the wild are almost always `int16` (SRTM, ASTER GDEM, Copernicus DEM) or `uint16`, so this breaks the main use case for the vertical-datum feature.

Reproducer:

```python
import numpy as np
import xarray as xr

dem = xr.DataArray(
    np.full((8, 8), 100, dtype=np.int16),
    dims=("y", "x"),
    coords={"y": np.linspace(40, 39, 8), "x": np.linspace(-105, -104, 8)},
    attrs={"crs": "EPSG:4326", "nodata": -32768},
)
reproject(
    dem,
    target_crs="EPSG:4326",
    src_vertical_crs="EGM96",
    tgt_vertical_crs="ellipsoidal",
)
# UFuncTypeError: Cannot cast ufunc 'add' output from dtype('float64') to dtype('int16') ...
```

**Expected behavior**

Vertical-datum shifts should work on integer rasters. The geoid offset is a float, so the output has to be float. The fix is to promote the array to `float32` for integer or small inputs (and to `float64` only when the source is already `float64`) before the shift, and to update `attrs['nodata']`, `attrs['_FillValue']`, and `attrs['nodatavals']` so they stay consistent with the new dtype. NaN can serve as the nodata sentinel in the promoted output.

**Additional context**

- Affects numpy, cupy, and dask+numpy backends, since they all funnel through `_apply_vertical_shift_numpy`.
- Tests should cover at minimum `int16 -> float32`, `uint8 -> float32`, and `float32 -> float32` (no promotion needed). Verify output dtype, nodata propagation, and that the geoid offset matches a known reference value.

brendancol:
Closes #2565.

## Summary

- The vertical-datum shift path used to `result = data.copy()` at the input dtype and then do `strip_data[is_valid] += N_total[is_valid]`. For int16/uint16 DEMs (SRTM, Copernicus DEM, ASTER GDEM) this raised `UFuncTypeError` because a float offset cannot be cast back into an integer array.
- `_apply_vertical_shift` now picks an `out_dtype` once (float32 for integer or float32 inputs, float64 only when the source is float64) and forwards it through the numpy / cupy / dask backends. The numpy path casts up front and rewrites the integer nodata sentinel to NaN so the rest of the loop uses uniform NaN semantics.
- The caller picks up the promoted nodata and writes it back to `attrs['nodata']`, `attrs['_FillValue']`, and `attrs['nodatavals']` so the metadata stays consistent with the array.

## Backend coverage

- numpy: handled directly in `_apply_vertical_shift_numpy`.
- cupy: round-trips via the numpy path (existing behaviour), inherits the promotion.
- dask+numpy: `map_blocks` now advertises the promoted dtype in `meta` and forwards `out_dtype` into every per-block call.
- dask+cupy: collapses to numpy-backed dask upstream (pre-existing), so the same path applies.

## Test plan

- [x] `pytest xrspatial/tests/test_reproject.py::TestVerticalShiftIntegerDtype` -- 7 new tests covering int16 promotion, uint8 promotion, float32 / float64 no-extra-promotion, nodata to NaN propagation, `_FillValue` / `nodatavals` attr refresh, and the dask path.
- [x] `pytest xrspatial/tests/test_reproject.py::TestVerticalShift` -- 12 existing vertical-shift tests still pass.
- [x] Broader sweep: `pytest -k "vertical or geoid or Vertical"` -- 54 tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@
 
 #### Fixed
 
+- `reproject` no longer crashes when the source raster is an integer
+  dtype and a vertical-datum shift is requested
+  (`src_vertical_crs` / `tgt_vertical_crs`). The shift path now
+  promotes integer inputs to `float32` (or to `float64` when the
+  source is already `float64`) before applying the geoid offset, and
+  refreshes `attrs['nodata']`, `attrs['_FillValue']`, and
+  `attrs['nodatavals']` to NaN so the sentinel matches the new dtype.
+  Integer nodata pixels propagate as NaN in the promoted output.
+  Affects the numpy, cupy, and dask+numpy backends. (#2565)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -1065,19 +1065,15 @@ def _promoted_vertical_dtype(src_dtype):
     """Return the float dtype to use when applying a geoid shift.
 
     The geoid offset is fractional metres, so any integer input has to
-    be promoted before we can add it. We pick ``float32`` for integer
-    and small-float inputs and only upgrade to ``float64`` when the
-    source is already ``float64``. Returns ``None`` when the input is
-    already a float dtype that does not need promotion.
+    be promoted before we can add it. Returns ``None`` when the input
+    is already a float dtype (including ``float64``) -- we leave its
+    precision alone. Returns ``np.float32`` for any non-float input;
+    that gives ~0.1 mm resolution in metres at altitudes up to ~10 km,
+    which is well below the accuracy of the geoid grids themselves.
     """
     src_dtype = np.dtype(src_dtype)
-    if src_dtype == np.float64:
-        return None
     if np.issubdtype(src_dtype, np.floating):
         return None
-    # Integer (or other non-float) input -> promote to float32. That
-    # gives ~0.1 mm precision in metres at altitudes up to ~10 km,
-    # which is well below the accuracy of the geoid grids themselves.
     return np.float32
 
 
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -967,7 +967,7 @@ def reproject(
     # Vertical datum transformation (if requested)
     if src_vertical_crs is not None and tgt_vertical_crs is not None:
         if src_vertical_crs != tgt_vertical_crs:
-            result_data = _apply_vertical_shift(
+            result_data, nd = _apply_vertical_shift(
                 result_data, y_coords, x_coords,
                 src_vertical_crs, tgt_vertical_crs, nd,
                 tgt_crs_wkt=tgt_wkt,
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: reproject vertical-datum shift on integer DEMs (#2596)

### Blockers

None.

### Suggestions

- `xrspatial/reproject/__init__.py:1064-1083` -- `_promoted_vertical_dtype` has two consecutive checks (`if src_dtype == np.float64: return None` then `if np.issubdtype(src_dtype, np.floating): return None`) but the second check already covers `float64`. Collapse to a single `if np.issubdtype(src_dtype, np.floating): return None`. Functionally identical, one fewer line, and the intent reads cleaner.

- `xrspatial/reproject/__init__.py:1080` -- the comment says "small-float inputs" promote to float32, but the function actually returns `None` for every float dtype (including float16 and complex floats, both via the `np.floating` check). Either tighten the comment to "integer inputs" only, or add a branch that explicitly promotes `float16` to `float32` if that case matters. Recommend the wording fix unless float16 DEM support is on the roadmap.

### Nits

- `xrspatial/reproject/__init__.py:1234-1255` -- the new conditional block in `_apply_vertical_shift_numpy` reads a bit denser than the original two-line version. Consider extracting the dtype-promotion + nodata remap to a small helper (e.g. `_promote_for_shift(data, nodata, out_dtype)` returning `(result, is_nan_nodata)`) so the strip loop below it stays the focus. Optional.

- The CHANGELOG entry is solid but slightly long for a "Fixed" bullet. Could tighten to one paragraph if you want -- not blocking.

### What looks good

- The fix is at the right level: `_apply_vertical_shift` decides the output dtype once and threads it through every backend, so numpy / cupy / dask all converge on the same promoted dtype and the caller's `out_attrs` stays consistent.
- The integer-nodata-to-NaN remap is the right design choice; it avoids carrying an "integer sentinel in a float array" footgun forward.
- Test coverage hits all the cases the issue called out: int16 -> float32 (numerical check against `geoid_height`), uint8 -> float32, float32 / float64 no-extra-promotion, NaN propagation, `_FillValue` + `nodatavals` refresh, and a dask-backed int16 case.
- The dask path correctly updates the `map_blocks(meta=...)` template so downstream consumers see the promoted dtype before computing.
- `_apply_vertical_shift` returning `(data, nodata)` instead of just `data` is a clean way to thread the new sentinel back to the caller without adding mutable state.

### Checklist
- [x] Algorithm matches reference (geoid offset formula H + N = h)
- [x] All implemented backends produce consistent results (single dtype decision shared across them)
- [x] NaN handling is correct (integer nodata -> NaN before shift, NaN passes through is_valid mask)
- [x] Edge cases are covered by tests (nodata pixels, multiple dtypes)
- [x] Dask chunk boundaries handled correctly (meta dtype matches block return)
- [x] No premature materialization or unnecessary copies (one extra mask pass per call; acceptable)
- [x] Benchmark exists or is not needed (pure correctness fix, no benchmark needed)
- [x] README feature matrix updated (no new function, no backend coverage change)

brendancol (commented):
## Follow-up review on commit 361af8ff

### Dispositions for the prior review

- **Suggestion: collapse the duplicate `float64` check in `_promoted_vertical_dtype`** -- fixed in 361af8ff. The function now reads `if np.issubdtype(src_dtype, np.floating): return None`.
- **Suggestion: misleading "small-float" wording (and float16 handling)** -- fixed in 361af8ff by rewriting the docstring to match the actual contract (any non-float promotes; floats are left alone). Not chasing float16 promotion since float16 DEMs are not a real-world case for this library.
- **Nit: extract the dtype-promotion block to a helper** -- dismissed. The inline block is ~10 lines and lives next to the one place that uses it; extracting it would split the logic across two definitions without making the strip loop meaningfully clearer. Per CLAUDE.md "no abstractions for single-use code."
- **Nit: shorten the CHANGELOG entry** -- dismissed. The entry documents a user-visible behaviour change (dtype and nodata both shift), and the extra lines pay for themselves when someone is reading the changelog to decide whether the fix affects them.

### State after the follow-up

- All 19 existing + new vertical-shift tests still pass locally.
- No new findings on the follow-up commit.

PR #2597, opened 2026-05-28
rasterize: honor resolution= exactly when bounds don't divide evenly

### LINKED ISSUE #2573: rasterize: resolution= argument not honored when bounds don't divide evenly
**Describe the bug**

When `resolution=` is passed to `rasterize()`, width and height come from `ceil((xmax - xmin) / x_res)`, but output coordinates are then rebuilt over the *original* bounds. If the bounds don't divide evenly by the resolution, the actual cell size ends up smaller than what was asked for.

Example:

```python

r = rasterize([(box(0,0,1,1), 1.0)], resolution=0.3, bounds=(0,0,1,1), fill=0)
# shape (4, 4); actual cell size = (1 - 0) / 4 = 0.25, not 0.3
print(r.x.values[1] - r.x.values[0])  # 0.25
```

Caller asked for 0.3-unit pixels, got 0.25-unit pixels.

**Expected behavior**

Either honor `resolution=` exactly (expand the output bounds so each cell is genuinely 0.3 units; matches rasterio / GDAL `from_bounds` / `from_origin` semantics), or document the parameter as a maximum cell size with a note that the actual size may shrink to fit the input bounds.

**Code location**

- `xrspatial/rasterize.py:3223` -- width/height computed with `ceil(bounds / resolution)`
- `xrspatial/rasterize.py:3352` -- output coords rebuilt over original `xmin..xmax`

**Proposed fix**

Take the first path. When `resolution=` drives width/height, set `xmax_out = xmin + final_width * x_res` and `ymax_out = ymin + final_height * y_res` before building coords. Existing tests with bounds that divide evenly pass unchanged.

brendancol:
Closes #2573.

## Summary

- `rasterize(..., resolution=R)` used to compute width/height via `ceil(extent / R)` but rebuilt output coords over the original bounds, so the actual cell size shrank to fit. `bounds=(0,0,1,1), resolution=0.3` returned a 4x4 raster with cell size 0.25 instead of 0.3.
- After resolving width/height, `final_bounds` is rebuilt so the grid anchors at `(xmin, ymax)` and extends right / down by exactly `width * x_res` / `height * y_res`. Matches `rasterio.transform.from_origin(west, north, x_res, y_res)`.
- When the original bounds already divide evenly, the new values equal the originals to float precision, so existing callers see no change.

## Backend coverage

Fix lives upstream of backend dispatch, so all four backends pick it up: numpy, cupy, dask+numpy, dask+cupy. Cross-backend parity is pinned by the new test class.

## Test plan

- [x] New regression test for the headline 4x4 / 0.25 example
- [x] Divides-evenly case still passes (no behavior change)
- [x] Asymmetric `(x_res, y_res)` tuple with non-divisible bounds
- [x] Polygon burn within original bounds is preserved when the grid expands
- [x] Cross-backend parity (numpy / cupy / dask+numpy / dask+cupy)
- [x] Full existing rasterize test suite passes (505 tests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@
 
 #### Fixed
 
+- `rasterize(..., resolution=R)` now honors the requested cell size
+  exactly when the bounds don't divide evenly. Previously width and
+  height were computed via `ceil(extent / resolution)` but output coords
+  were rebuilt over the original bounds, so the actual cell size shrank
+  to fit (e.g. `bounds=(0,0,1,1), resolution=0.3` gave a 4x4 raster with
+  cell size 0.25 instead of 0.3). The grid now anchors at `(xmin, ymax)`
+  and extends right / down by `width * x_res` / `height * y_res`,
+  matching `rasterio.transform.from_origin(west, north, x_res, y_res)`.
+  When the bounds already divide evenly the new values equal the
+  originals so existing callers see no change. (#2573)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -3046,6 +3046,11 @@ def rasterize(
         Pixel size.  Must be finite and ``> 0``.  When given with
         ``bounds``, computes ``width`` and ``height`` automatically.
         A single float uses the same resolution for both axes.
+        The requested cell size is honored exactly.  When the bounds
+        don't divide evenly by the resolution, the output grid extends
+        past the original ``xmax`` / ``ymin`` (anchored at ``xmin`` /
+        ``ymax``) rather than shrinking the cell to fit.  Matches the
+        ``rasterio.transform.from_origin`` convention.
     like : xr.DataArray, optional
         Template raster.  Width, height, bounds, and dtype are copied
         from this array (any can still be overridden explicitly).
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: rasterize resolution= fix (PR #2597)

### Blockers
None.

### Suggestions
- [ ] `xrspatial/tests/test_rasterize_resolution_exact_2573.py`: no test combines `like=` with `resolution=` against a template whose bounds don't divide evenly. The existing `test_resolution_override_strips_stale_grid_attrs` at `test_rasterize.py:2096` uses `resolution=0.5` on a 10x10 template, so bounds divide evenly and the new overshoot path isn't exercised. A test with `like=` (10x10 template) plus `resolution=0.3`, asserting the output cell size is 0.3, would close that gap.
- [ ] `xrspatial/rasterize.py:3239-3241`: the inferred-bounds + resolution path also picks up the fix. When a GeoDataFrame's `total_bounds` supplies non-divisible bounds and the caller passes `resolution=`, the grid now extends past the geometry envelope to honor the cell size. No test pins this. A small case with an irregular geometry, no explicit `bounds=`, and `resolution=0.3` would lock in the inferred-bounds branch alongside the explicit-bounds branch.

### Nits
- [ ] `xrspatial/rasterize.py:3239-3241`: worth pinning the `get_dataarray_resolution(result)` side-effect. After the fix, that helper reports the cell size the caller actually asked for. A one-line assert against it in the existing test documents the downstream win for slope / aspect / proximity, which previously inherited the shrunk cellsize.

### What looks good
- Only the `resolution=` branch is touched. Explicit width/height and `like=`-only paths keep their behavior byte-for-byte.
- Anchoring at `(xmin, ymax)` matches `rasterio.transform.from_origin`, which is what GDAL / rasterio users will expect.
- `width * x_res == xmax - xmin` to float precision when bounds divide evenly is why the existing tests still pass without modification. The inline comment and CHANGELOG capture this.
- Cross-backend parity is pinned for numpy, cupy, dask+numpy, and dask+cupy, since the fix lives upstream of dispatch.
- Docstring update tells users what to expect when bounds don't divide evenly and points at the matching rasterio API. Better than silently changing behavior.
- 505 existing rasterize tests pass unchanged.

### Checklist
- [x] Algorithm matches the rasterio.transform.from_origin convention
- [x] NaN / fill semantics unchanged
- [x] Edge cases covered: divides-evenly, doesn't-divide-evenly, asymmetric tuple, geom inside original bounds with expanded grid
- [x] No dask chunk boundary concerns; the change is in bounds resolution, not in the rasterize kernels
- [n/a] No new public function, so no benchmark or README feature matrix entry needed
- [x] Docstring updated for the new contract

brendancol (commented):
## PR Review (follow-up): rasterize resolution= fix (PR #2597)

Re-reviewed after `c7eeacad`. All three prior findings are now addressed.

- Suggestion 1 (`like=` + non-divisible `resolution=`): fixed by `TestLikePlusResolutionOvershoot` (10x10 template + `resolution=0.3` -> 34x34 grid with exact cell size).
- Suggestion 2 (inferred-bounds + `resolution=`): fixed by `TestInferredBoundsPlusResolution` (no explicit `bounds=`, geometry envelope drives `final_bounds`).
- Nit (`get_dataarray_resolution` side-effect): fixed by `test_get_dataarray_resolution_matches_request`.

### Blockers
None.

### Suggestions
None.

### Nits
None.

12/12 regression tests pass. Existing 505-test rasterize suite still green. No further findings.

PR #2598, opened 2026-05-28
accessor: backend-aware .xrs.open_geotiff on DataArray and Dataset

### LINKED ISSUE #2557: accessor: add DataArray-side .xrs.open_geotiff with backend + CRS inference
## Reason or Problem

`ds.xrs.open_geotiff(source)` exists on the Dataset accessor (`xrspatial/accessor.py:1068`) and reads only the pixel window covered by the Dataset's `y`/`x` coords. Three gaps:

1. There is no DataArray counterpart. `da.xrs.to_geotiff(...)` is defined but `da.xrs.open_geotiff(...)` is not, which is inconsistent and forces a Dataset wrap to use the windowed-read shortcut.
2. Neither accessor adapts the read to the caller's backend. A dask+cupy DataArray that calls the accessor receives an eager numpy DataArray back, which then has to be rechunked and moved to GPU manually.
3. The windowing path silently assumes the caller's coords are in the file's CRS. If they are not, the computed window is wrong and the read returns the wrong region with no error.

## Proposal

Add `da.xrs.open_geotiff(source, **kwargs)` to the DataArray accessor and enhance both accessors to:

- Infer the caller's backend (numpy / cupy / dask+numpy / dask+cupy) via `_classify_backend` and pass matching `gpu=` / `chunks=` to `xrspatial.geotiff.open_geotiff` so the returned DataArray's backend matches `self`. Caller-supplied `gpu=` / `chunks=` always win.
- On CRS mismatch between caller and file, raise a clear `ValueError` by default (replacing today's silently-wrong window). With `auto_reproject=True`, project the caller's bbox to the file's CRS for the windowed read, then reproject the read result back to the caller's CRS via `xrspatial.reproject.reproject` so the returned DataArray lines up with `self`.

**Design:**

- A new private helper in `xrspatial/accessor.py` (e.g. `_open_geotiff_windowed(obj, source, *, auto_reproject, **kwargs)`) holds the shared body. Both accessor methods delegate to it.
- Backend inference uses `_classify_backend(obj)` from `xrspatial/utils.py:345`. Mapping:
  - `numpy` -> no override
  - `cupy` -> `gpu=True`
  - `dask+numpy` -> `chunks=<caller's y-chunk size>`
  - `dask+cupy` -> `gpu=True, chunks=<caller's y-chunk size>`
- CRS comes from `attrs['crs']` (int EPSG, the convention established by the geotiff reader). File CRS comes from `_read_geo_info(source).crs_epsg`.
- The reproject path uses `pyproj.Transformer` to convert the caller's bbox to the file CRS for windowing, then `xrspatial.reproject.reproject(read_da, target_crs=caller_crs, source_crs=file_crs)` for the final result. `reproject` already supports all four backends, so backend match is preserved.
- The new flag is `auto_reproject: bool = False`. Matches the `allow_*` / `auto_*` style already used in `open_geotiff`.

**Usage:**

```python
import xrspatial  # registers .xrs accessor
import xarray as xr

# DataArray-side, backend inferred from caller
da = some_dask_cupy_dataarray  # EPSG:4326
out = da.xrs.open_geotiff("aoi.tif")           # raises if file CRS != 4326
out = da.xrs.open_geotiff("aoi.tif", auto_reproject=True)  # returns dask+cupy in EPSG:4326

# Dataset-side keeps existing behavior; gains the same flag
out = ds.xrs.open_geotiff("aoi.tif", auto_reproject=True)
```

**Value:**

- Closes the symmetry gap between `da.xrs.to_geotiff` and `da.xrs.open_geotiff`.
- Makes the windowed-read shortcut backend-correct by default, removing manual rechunk plus `.map_blocks(cupy.asarray)` boilerplate at every call site.
- Turns today's silent CRS mismatch (wrong window, no error) into either a loud error or a correct, reprojected result, controlled by a single flag.

## Stakeholders and Impacts

Affects `xrspatial/accessor.py` only (new helper plus two accessor methods). Uses existing public utilities (`_classify_backend`, `_read_geo_info`, `_extent_to_window`, `xrspatial.reproject.reproject`). No changes to `xrspatial/geotiff/` or `xrspatial/reproject/`. I will implement.

## Drawbacks

- Default behavior on CRS mismatch changes from "silently wrong window" to "raises ValueError". This is a behavior change for any caller relying on the wrong-but-silent behavior; the upside is that the new error message points directly at `auto_reproject=True`.
- Adds an import-time dependency on `pyproj.Transformer` inside the accessor body. The rest of the repo already depends on pyproj via the reproject module, so this is a no-op for dependency scope.

## Alternatives

- Leave the DataArray accessor empty and document the Dataset wrap workaround. Rejected: keeps the asymmetry and the silent-CRS-bug.
- Always reproject on mismatch with no flag. Rejected: surprising for callers who explicitly want native-CRS data.

## Unresolved Questions

None.

brendancol:
Closes #2557.

## Summary

- Adds `da.xrs.open_geotiff(source, *, auto_reproject=False, **kwargs)` to the DataArray accessor, mirroring the existing Dataset method.
- Both accessors now infer the caller's backend (numpy / cupy / dask+numpy / dask+cupy) and pass matching `gpu=` / `chunks=` to `xrspatial.geotiff.open_geotiff` so the returned DataArray matches `self`. Caller-supplied `gpu=` / `chunks=` always override the inference.
- New `auto_reproject` flag turns CRS mismatch from "silently wrong window" into either a clear `ValueError` (default) or a correctly reprojected result that lines up with the caller's CRS (via `xrspatial.reproject.reproject`).

## Backend coverage

numpy / cupy / dask+numpy / dask+cupy. Backend inference reads the caller's y-chunk size via `_classify_backend`. The reproject path delegates to `xrspatial.reproject.reproject`, which already supports all four backends.

## Test plan

- [x] DataArray accessor windowed read returns the expected slice
- [x] DataArray accessor raises on missing y/x coords
- [x] kwargs (e.g. `name=`) are forwarded to `open_geotiff`
- [x] numpy caller -> numpy result
- [x] dask caller -> dask result with inferred y-chunk size from `self.chunks`
- [x] Explicit `chunks=` overrides inferred chunks
- [x] Dataset accessor infers backend from the first 2D y/x data variable
- [x] CRS mismatch raises `ValueError` by default
- [x] `auto_reproject=True` returns a DataArray in the caller's CRS
- [x] Caller without `attrs['crs']` skips the mismatch check (backward compatible)
- [x] Dataset CRS falls back from `ds.attrs['crs']` when the variable lacks one
- [x] Existing `xrspatial/geotiff/tests/integration/test_dask_pipeline.py` suite still passes (72 tests, no regressions)
- [x] Existing `xrspatial/tests/test_accessor.py` suite still passes (29 tests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@
 
 ### Unreleased
 
+#### Added
+
+- `da.xrs.open_geotiff(source, *, auto_reproject=False, **kwargs)`
+  DataArray accessor that mirrors the existing Dataset method, plus
+  backend-aware enhancements on both accessors. The accessor infers
+  the caller's backend (numpy / cupy / dask+numpy / dask+cupy) and
+  passes matching `gpu=` / `chunks=` to `open_geotiff` so the
+  returned DataArray matches the caller. Caller-supplied `gpu=` /
+  `chunks=` always override the inference. On CRS mismatch between
+  the caller and the file, the accessor raises a clear `ValueError`
+  by default (replacing the previous silently-wrong window) and
+  with `auto_reproject=True` it projects the caller's bbox into the
+  file's CRS for the windowed read and reprojects the result back
+  to the caller's CRS via `xrspatial.reproject.reproject`. (#2557)
+
 #### Fixed
 
 - `polygonize` now auto-detects the raster's affine transform from
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -104,6 +104,41 @@ def _infer_caller_y_chunk(obj):
     return int(y_chunks[0])
 
 
+def _to_pyproj_crs(crs):
+    """Normalize a CRS value (int EPSG, ``"EPSG:xxxx"``, WKT, PROJ
+    string, or ``pyproj.CRS``) into a ``pyproj.CRS`` instance, or
+    ``None`` when the value is missing or unparseable.
+    """
+    if crs is None:
+        return None
+    try:
+        from pyproj import CRS as _PyprojCRS
+        return _PyprojCRS(crs)
+    except Exception:
+        return None
+
+
+def _bbox_edge_samples(x_min, y_min, x_max, y_max, n_per_side=20):
+    """Return ``(xs, ys)`` arrays sampling each edge of the bbox.
+
+    Sampling the perimeter (instead of only the 4 corners) before
+    transforming into the source CRS keeps the envelope of the
+    transformed points close to the true projected bbox even when the
+    transform has curvature across the box (high latitudes, large
+    extents).
+    """
+    import numpy as np
+    a = np.linspace(x_min, x_max, n_per_side)
+    b = np.full(n_per_side, y_min)
+    c = np.full(n_per_side, y_max)
+    d = np.linspace(y_min, y_max, n_per_side)
+    e = np.full(n_per_side, x_min)
+    f = np.full(n_per_side, x_max)
+    xs = np.concatenate([a, a, e, f])
+    ys = np.concatenate([b, c, d, d])
+    return xs, ys
+
+
 def _open_geotiff_windowed(obj, source, *, auto_reproject=False, **kwargs):
     """Shared implementation for ``.xrs.open_geotiff`` on DataArray and
     Dataset.
@@ -111,8 +146,20 @@ def _open_geotiff_windowed(obj, source, *, auto_reproject=False, **kwargs):
     ``obj`` is a DataArray that carries the bbox (``y``/``x`` coords),
     the CRS (``attrs['crs']``), and the backend used to infer
     ``gpu=`` / ``chunks=`` kwargs for the underlying read.
+
+    The windowing extent is expanded by half a pixel on each side so
+    edge pixels of the caller's footprint are captured. CRS values on
+    both sides (``attrs['crs']`` and the file's georeference) are
+    normalized through ``pyproj.CRS`` for equality, so int EPSG codes,
+    ``"EPSG:xxxx"`` strings, WKT strings, and ``pyproj.CRS`` instances
+    all compare correctly.
+
+    When ``auto_reproject=True`` and the CRSs differ, the caller's
+    bbox is sampled along its perimeter (not just the four corners)
+    before transformation into the file CRS, so the windowed read
+    covers the full footprint even when the transform has curvature
+    across the bbox.
     """
-    import numpy as np
     from .geotiff import open_geotiff, _read_geo_info, _extent_to_window
     from .utils import _classify_backend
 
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: accessor: backend-aware .xrs.open_geotiff on DataArray and Dataset

None.

- [ ] `xrspatial/accessor.py:132-138` (and the matching `int(caller_crs)` / `int(file_crs)` calls at 142-143, 150-152, 186-187): the CRS comparison assumes `attrs['crs']` is int-convertible. Non-int values (`"EPSG:4326"`, raw WKT, `pyproj.CRS`) make `int(caller_crs)` raise with a confusing `ValueError` before the mismatch logic even runs. The chained case is the worst version: after `auto_reproject=True`, `xrspatial.reproject.reproject` sets `result.attrs['crs']` to a WKT string (visible in `test_auto_reproject_returns_caller_crs`), so `da.xrs.open_geotiff(...).xrs.open_geotiff(...)` breaks on the second call. Normalize both sides through `pyproj.CRS(...)` and compare with `.equals(...)`, or reuse `xrspatial.geotiff._crs._resolve_crs_to_wkt` for both.

- [ ] `xrspatial/accessor.py:153-159`: the bbox-to-file-CRS projection samples only the 4 corners. For projections with curvature across the bbox (high latitudes, large extents, anything pre/post-pole), the projected corner envelope is a strict subset of the true bbox and the windowed read misses data along the curved edges. At minimum add a docstring caveat; better, sample several points along each edge before taking min/max (e.g. 20 points per side via `np.linspace`) so the window covers the full footprint.

- [ ] `xrspatial/accessor.py:124-125`: `np.asarray(obj.coords['y'].values)` is a double conversion. `.values` already returns a numpy array. Use `obj.coords['y'].values` directly (matches the existing Dataset accessor style at line 1094 of the pre-change file).

- [ ] `xrspatial/geotiff/tests/integration/test_dask_pipeline.py:1166-1188` (`test_auto_reproject_returns_caller_crs`): `result.coords['y'].max() > 1e5` just confirms the y values look like mercator metres rather than degrees, not that the reprojection is numerically correct. Compare the caller's bbox to `(result.coords['x'].min/max, result.coords['y'].min/max)` within a small tolerance so a future regression in the reprojection direction gets caught.

- [ ] No GPU test path (gated on `has_cuda_and_cupy`) for backend inference. The existing accessor tests follow the same pattern, so this is consistent with the file, but the main new behaviour in this PR is backend matching, and the cupy / dask+cupy branches at `accessor.py:173-178` are currently unexercised. A gated cupy test (e.g. assert that a cupy caller gets back a cupy-backed result) would close the gap.

- [ ] The `_open_geotiff_windowed` docstring at `accessor.py:108-114` doesn't mention the half-pixel extent expansion or the EPSG-int assumption. Both surface as surprises (the mismatch error message names EPSG; a user with a WKT string in `attrs['crs']` will be confused).

### What looks good

- Helper extraction is clean. Dataset and DataArray paths share `_open_geotiff_windowed` so behaviour is identical in both directions.
- API hygiene is right: caller kwargs always win over inferred backend kwargs (`kwargs.setdefault` at 174 and 178), and the new flag has a default that preserves prior behaviour on the matching-CRS path.
- Default CRS-mismatch error message names the flag (`accessor.py:141-146`), so callers find the fix without re-reading the docs.
- Tests cover backend inference, the `chunks=` override, the no-`attrs['crs']` backward-compat path, the Dataset CRS-attr fallback, and the mismatch raise.
- CHANGELOG entry under `#### Added` calls out the behaviour change from silently-wrong-window to ValueError, which is the only user-visible regression risk.

### Checklist
- [x] Algorithm matches reference/paper (n/a -- no numerical algorithm here, just windowing + reproject delegation)
- [ ] All implemented backends produce consistent results (numpy/dask only; cupy/dask+cupy paths untested)
- [x] NaN handling is correct (n/a for this PR; coords are expected to be finite)
- [x] Edge cases are covered by tests (no-coords, no-CRS, kwargs override, mismatch raise, mismatch + reproject)
- [x] Dask chunk boundaries handled correctly (chunks inferred from caller's y-axis chunks)
- [x] Benchmark exists or is not needed (accessor-only, delegates to existing benchmarked function)
- [ ] README feature matrix updated (n/a -- accessor enhancement, not a new spatial op)

brendancol (commented):
## PR Review: follow-up after review fixes (#2598)

None.

- [ ] `xrspatial/accessor.py:107-118` (`_to_pyproj_crs`): the bare `except Exception` swallows every parse failure and returns `None`. The mismatch logic then treats `None` as "no CRS to compare" and skips the check, so a malformed but non-None `attrs['crs']` (e.g. `"EPSG:99999"`, a typo in a WKT) silently disables the safety net and reads with the caller's bbox assumed to be in the file CRS. Either narrow the except clause to the specific pyproj exceptions, or surface the parse failure as a `ValueError` at the same severity as a real mismatch. Today's behaviour is strictly less safe than the previous int() coercion, which would at least have raised on garbage.

- [ ] `xrspatial/accessor.py:121-139` (`_bbox_edge_samples`): `n_per_side=20` is a magic constant with no override path. Fine for now (the bbox transform is cheap), but worth a comment near the call site at `accessor.py:202` explaining the choice, since 20 will surprise the next reader.

- [ ] `xrspatial/geotiff/tests/integration/test_dask_pipeline.py:1264-1265` (`test_chained_open_geotiff_with_wkt_crs`): the final assertion `assert result.shape == (4, 4) or result.shape[0] >= 4` is permissive enough to pass on almost anything. The half-pixel expansion can grow the result by one row, so `result.shape[0] >= 4` alone is the real check; the `== (4, 4)` half is misleading. Drop the OR or assert a tight upper bound (e.g. `4 <= result.shape[0] <= 6`).

### What looks good

- All three Suggestions and all three Nits from the first review pass are addressed.
- `_to_pyproj_crs` + `pyproj.CRS.equals` is the right comparison primitive. It handles int / `"EPSG:xxxx"` / WKT / `pyproj.CRS` symmetrically, and `test_chained_open_geotiff_with_wkt_crs` exercises the chained WKT case directly.
- Perimeter sampling at `accessor.py:197-207` correctly drops the int() coercions and routes both CRSs through pyproj objects for `Transformer.from_crs`. The 20-point-per-side sampling closes the curvature gap for the worst-realistic case (a large bbox at high latitudes).
- The strengthened `test_auto_reproject_returns_caller_crs` now compares result bbox to template bbox within a one-pixel tolerance, which catches regressions in projection direction. Good replacement for the prior magnitude check.
- The new gated `TestOpenGeotiffGPUBackendInference_2557::test_cupy_caller_returns_cupy` covers the `gpu=True` branch (`accessor.py:221-222`) that was previously unexercised. Local runs skip cleanly when CUDA / cupy is absent.

### Checklist
- [x] Findings from the first review pass are addressed
- [x] CRS comparison handles int / EPSG-string / WKT / pyproj.CRS uniformly
- [x] bbox transform covers curved-edge cases
- [x] Tests strengthened with bbox tolerance and chained-WKT regression
- [x] GPU backend-inference test added (gated)
- [x] Existing tests still pass (74 in test_dask_pipeline.py)
- [ ] `_to_pyproj_crs` silently returns None on parse failure (see Suggestion)

PR #2599, opened 2026-05-28
resample: clamp dask interp overlap depth per axis (#2547)

### LINKED ISSUE #2547: resample: cubic on dask backends fails for arrays smaller than depth=16
## Summary

`resample(method='cubic')` on `dask+numpy` or `dask+cupy` raises
`ValueError: The overlapping depth 16 is larger than your array <N>`
when the input array is smaller than 16 cells along either axis.

The cubic prefilter depth was raised to 16 in #1464 to drive the IIR
boundary transient below float32 epsilon. `_ensure_min_chunksize`
enforces `min_size = max(2 * depth + 1, ...)`, so it grows individual
chunks to 33+ cells -- but dask's `overlap()` still rejects when the
*array total* is below `depth`, which happens for an Nx1 column or a
single tiny chunk.

## Reproducer

```python
import numpy as np
import dask.array as da

data = np.array([[i + 1.0] for i in range(8)], dtype=np.float32)
agg = create_test_raster(data, backend='dask+numpy', dims=['y', 'x'],
                         attrs={'res': (1.0, 1.0)}, chunks=(2, 1))
resample(agg, scale_factor=0.5, method='cubic')
# ValueError: The overlapping depth 16 is larger than your array 8.
```

The eager `numpy` and `cupy` paths handle this case fine; only the
dask backends fail.

## Expected

Either:

1. The cubic dask path falls back to a depth small enough for the
   array size when the array is too small to absorb depth=16.
2. The dask path materialises the small array (sum of `chunks[ax]` <
   `depth`) and runs the eager kernel before re-chunking the result.
3. The error message is rewritten to point at the documented
   minimum-size constraint so users can size their inputs.

## Coverage gate

A test in `test_resample_coverage_2026_05_27.py::TestSingleColumnRaster::test_nx1_downsample`
currently skips this combination. Remove the `pytest.skip` block once
the fix lands.

## Discovered via

`/deep-sweep test-coverage` on the resample module, 2026-05-27.

brendancol:
## Summary

- Clamp the cubic prefilter overlap depth per axis in `_run_dask_numpy` and `_run_dask_cupy` so dask's `overlap()` no longer rejects arrays smaller than depth=16.
- Thread per-axis depths (`depth_y`, `depth_x`) through `_interp_block_np` / `_interp_block_cupy` so local coordinate translation matches the actual padding added on each axis.
- Remove the `pytest.skip` that documented this hole and add `TestDaskCubicSmallInput` covering Nx1, 1xN, single-chunk 4x4, multi-chunk 4x4, and 8x8 inputs across both dask backends.

## Backend coverage

- numpy: unchanged (eager path)
- cupy: unchanged (eager path)
- dask+numpy: fixed (clamp + per-axis depth)
- dask+cupy: fixed (clamp + per-axis depth)

## Notes

On axes large enough to absorb the full depth=16, behaviour is byte-for-byte identical to before. On clamped axes the IIR boundary transient is larger than float32 epsilon, but this matches the eager kernels, which use no overlap padding at all.

## Test plan

- [x] `pytest xrspatial/tests/test_resample.py xrspatial/tests/test_resample_signature_annot_2544.py xrspatial/tests/test_resample_coverage_2026_05_27.py` (261 passed)
- [x] Original repro from issue #2547 returns the same values as the eager numpy backend
- [x] `TestDaskCubicSmallInput` covers Nx1, 1xN, single-chunk 4x4, multi-chunk 4x4, and 8x8 inputs

Closes #2547

### DIFF
--- a/xrspatial/resample.py
+++ b/xrspatial/resample.py
@@ -706,7 +706,8 @@ _AGG_BLOCK_FUNCS = {
 def _interp_block_np(block, global_in_h, global_in_w,
                      global_out_h, global_out_w,
                      cum_in_y, cum_in_x, cum_out_y, cum_out_x,
-                     depth, order, work_dtype, out_dtype, block_info=None):
+                     depth_y, depth_x, order, work_dtype, out_dtype,
+                     block_info=None):
     """Interpolate one (possibly overlapped) numpy block."""
     yi, xi = block_info[0]['chunk-location']
     target_h = int(cum_out_y[yi + 1] - cum_out_y[yi])
@@ -723,8 +724,8 @@ def _interp_block_np(block, global_in_h, global_in_w,
     ix = (ox + 0.5) * (global_in_w / global_out_w) - 0.5
 
     # Convert to local block coordinates (overlap shifts the origin)
-    iy_local = iy - (cum_in_y[yi] - depth)
-    ix_local = ix - (cum_in_x[xi] - depth)
+    iy_local = iy - (cum_in_y[yi] - depth_y)
+    ix_local = ix - (cum_in_x[xi] - depth_x)
 
     yy, xx = np.meshgrid(iy_local, ix_local, indexing='ij')
     coords = np.array([yy.ravel(), xx.ravel()])
@@ -769,7 +770,8 @@ def _interp_block_np(block, global_in_h, global_in_w,
 def _interp_block_cupy(block, global_in_h, global_in_w,
                        global_out_h, global_out_w,
                        cum_in_y, cum_in_x, cum_out_y, cum_out_x,
-                       depth, order, work_dtype, out_dtype, block_info=None):
+                       depth_y, depth_x, order, work_dtype, out_dtype,
+                       block_info=None):
     """CuPy variant of :func:`_interp_block_np`."""
     from cupyx.scipy.ndimage import map_coordinates as _cupy_map_coords
     from cupyx.scipy.ndimage import spline_filter as _cupy_spline_filter
@@ -790,8 +792,8 @@ def _interp_block_cupy(block, global_in_h, global_in_w,
     iy = (oy + 0.5) * (global_in_h / global_out_h) - 0.5
     ix = (ox + 0.5) * (global_in_w / global_out_w) - 0.5
 
-    iy_local = iy - float(cum_in_y[yi] - depth)
-    ix_local = ix - float(cum_in_x[xi] - depth)
+    iy_local = iy - float(cum_in_y[yi] - depth_y)
+    ix_local = ix - float(cum_in_x[xi] - depth_x)
 
     yy, xx = cupy.meshgrid(iy_local, ix_local, indexing='ij')
     coords = cupy.array([yy.ravel(), xx.ravel()])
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: resample: clamp dask interp overlap depth per axis (#2547)

- None.

- None.

- `xrspatial/resample.py:34-40`: the `_INTERP_DEPTH` block-comment still claims the cubic dask path uses depth=16 unconditionally ("Depth 16 puts the residual at ~7e-10"). After this PR the dask path may clamp depth below 16 on small inputs. A one-line addendum that points at the new clamp in `_run_dask_numpy` would keep the comment honest. The constant itself is unchanged so the comment is not strictly wrong.
- `xrspatial/tests/test_resample_coverage_2026_05_27.py` `((4, 4), (2, 2))` case: after `_ensure_min_chunksize` runs with `min_size=7`, a 4x4 input is rechunked into a single 4x4 chunk, so this parametrization effectively duplicates the `((4, 4), (4, 4))` single-chunk case rather than exercising the multi-chunk path. Not a defect; just noting the test does not cover what its name implies. A larger multi-chunk shape (for example `((16, 16), (4, 4))`) would actually keep multiple chunks after `_ensure_min_chunksize`.

### What looks good
- The clamp formula `min(depth, max(0, axis_total - 1))` is correct: it leaves depth unchanged whenever the axis is large enough and only reduces it for small inputs. Dask's `overlap()` accepts `depth <= sum(chunks)`, so staying one below total is conservative without being lossy in the common case.
- `min_size = 2 * max(depth_y, depth_x) + 1` matches the prior contract: at least one chunk wide enough to hold the kernel's reach on either side. Memory cost is the same or lower than before.
- The per-axis depth is threaded through `_interp_block_np` / `_interp_block_cupy` correctly. Local coordinate translation (`iy - (cum_in_y[yi] - depth_y)`) uses the same axis-specific depth that was passed to `dask.overlap`, so coordinates stay consistent.
- The `if depth_y > 0 or depth_x > 0` guard preserves the existing fast-path for `nearest`/`bilinear` (depth=1 always positive, so unchanged) and still skips the overlap call when both axes degenerate to depth 0.
- Test coverage is appropriate: Nx1, 1xN, single-chunk 4x4, and 8x8 each exercise a distinct edge case (one-axis clamp, both-axes clamp, no overlap needed on one axis).
- The skip removal at `test_resample_coverage_2026_05_27.py:75` correctly enables all 8 method x 4 backend cases for the Nx1 column, which now pass.

### Checklist
- [x] Algorithm matches reference (cubic prefilter behaviour unchanged on large inputs)
- [x] All implemented backends produce consistent results (numpy reference matches dask within float32 tolerance on small inputs; bit-for-bit on 8x1)
- [x] NaN handling is correct (unchanged code paths)
- [x] Edge cases are covered by tests (Nx1, 1xN, single-chunk, multi-chunk, depth<chunk)
- [x] Dask chunk boundaries handled correctly (per-axis depth matches overlap padding)
- [x] Benchmark exists or is not needed (pure bug fix)

brendancol (commented):
## Follow-up review

Both nits from the previous review are addressed in fc004239:

- `_INTERP_DEPTH` comment now mentions the per-axis clamp so readers do not assume depth=16 is the active value in every dask call.
- Dropped the redundant `((4, 4), (2, 2))` parametrization. `_ensure_min_chunksize` collapses any clamped axis to a single chunk (its `min_size` is `2 * (axis_total - 1) + 1 > axis_total`), so a multi-chunk user layout on a small axis is indistinguishable from the single-chunk case. Added a comment explaining this so the parametrization is honest about what it tests.

No new findings. The test suite still passes (259 tests).

PR #2600, opened 2026-05-28
reproject: unit-aware bounds_policy="auto" blow-up detection

### LINKED ISSUE #2582: reproject(bounds_policy="auto") crops ordinary geographic-to-projected reprojections
**Describe the bug**

`reproject(..., bounds_policy="auto")` crops valid edge data on ordinary geographic-to-projected reprojections. The blow-up heuristic in `xrspatial/reproject/_grid.py` (around line 255) compares the source span in source CRS units (e.g. degrees for EPSG:4326) against the output span in target CRS units (e.g. metres for EPSG:3857). When the units differ, the ratio is meaningless. The > 50x threshold trips on almost any geographic-to-projected case, forcing the 2/98 percentile fallback even when the reprojection is well-behaved.

**Reproduction**

```python
import numpy as np
import xarray as xr

data = np.random.RandomState(0).rand(64, 64).astype(np.float32)
r = xr.DataArray(
    data, dims=['y', 'x'],
    coords={'y': np.linspace(10, -10, 64), 'x': np.linspace(-10, 10, 64)},
    attrs={'crs': 'EPSG:4326'},
)
auto = reproject(r, 'EPSG:3857', bounds_policy='auto')
raw  = reproject(r, 'EPSG:3857', bounds_policy='raw')

# auto x span 2,120,376 m vs raw 2,226,395 m, 106 km cropped
# auto y span 2,155,716 m vs raw 2,226,395 m,  70 km cropped
```

The source bbox is (-10, -10, 10, 10) in EPSG:4326, nowhere near a singularity. Under `auto`, roughly 70-106 km per side gets trimmed compared to `raw`.

**Expected behavior**

`bounds_policy="auto"` should leave benign reprojections alone and only trip the percentile fallback when there is an actual projection singularity (antimeridian wrap, pole near a non-polar projection, etc.).

**Suggested fix**

Replace the unit-mismatched ratio comparison with a unit-agnostic signal. Two options:

1. Non-finite fraction of raw (unclamped) edge samples in the target CRS.
2. Ratio of max absolute coordinate to median absolute coordinate among projected samples. Benign cases stay near 1-2; real singularities produce ratios above 10.

Option 2 handles both inf-producing singularities (Mercator at the poles) and finite-but-astronomically-large singularities (polar stereographic at the opposite pole).

**Test plan**

- Regression test: EPSG:4326 -> EPSG:3857 on (-10,-10,10,10) under `bounds_policy="auto"` matches the `raw` output bounds to a small tolerance.
- Pathological case still trips: EPSG:4326 global -> polar stereographic still falls back to percentile bounds.
- Existing tests (`test_raw_skips_clamp_and_percentile`, `test_warns_when_percentile_fires_under_auto`, `test_no_warning_on_benign_input`) keep passing.

brendancol:
## Summary

Closes #2582. The `bounds_policy="auto"` heuristic compared source span (in source CRS units) to output span (in target CRS units), so the > 50x ratio fired on almost every geographic-to-projected reprojection and silently cropped valid edge data via the 2/98 percentile fallback. For a (-10, -10, 10, 10) EPSG:4326 bbox going to EPSG:3857, that was 70-106 km per side.

Replaced the unit-mismatched ratio with two unit-agnostic signals:

- Magnitude ratio: max absolute projected coordinate / median absolute projected coordinate. Benign reprojections stay near 1-2. Real singularities push past 10 (Mercator near the poles after clamp) or far higher (polar-stereographic on the opposite pole projects to ~1e23).
- Non-finite fraction: share of raw (unclamped) edge samples that produce inf/nan in the target CRS. Catches singularities that the auto clamp would otherwise hide.

## Verification

The reproduction case from the issue:

| | Before | After |
|--|--|--|
| auto vs raw x crop | 106 km | 0 m |
| auto vs raw y crop | 70 km | 0 m |

Global EPSG:4326 -> EPSG:3857 still trips the percentile fallback. The pathological EPSG:4326 -> EPSG:3413 case (south pole into a north polar stereographic) still trips.

## Backend coverage

- numpy: yes
- cupy: yes (heuristic runs on a small numpy probe, no GPU-specific path)
- dask + numpy: yes (regression test added)
- dask + cupy: yes (same as cupy)

## Test plan

- [x] Regression test: 4326 -> 3857 on (-10,-10,10,10) under auto matches raw within one pixel.
- [x] Benign 4326 -> 3857 emits no bounds_policy warning under auto.
- [x] Pathological 4326 -> 3413 still trips the auto fallback and warning.
- [x] Existing 13 TestBoundsPolicy tests pass.
- [x] Full reproject suite (368 tests) passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@
 
 #### Fixed
 
+- `reproject(..., bounds_policy="auto")` no longer crops valid edge data
+  on ordinary geographic-to-projected reprojections. The old blow-up
+  heuristic compared source span (e.g. degrees for EPSG:4326) against
+  target span (e.g. metres for EPSG:3857) and tripped on almost any
+  geographic-to-projected pair. The new heuristic is unit-agnostic:
+  it compares the max absolute projected coordinate to the median
+  and also checks the non-finite fraction of raw edge samples in the
+  target CRS. Benign reprojections stay untouched; real singularities
+  (Mercator at the poles, polar-stereographic on the opposite pole)
+  still trigger the percentile fallback. (#2582)
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/docs/source/reference/reproject.rst
+++ b/docs/source/reference/reproject.rst
@@ -26,9 +26,11 @@ The available options are:
 - ``"auto"`` (default): clamp geographic source bounds inward by
   0.01 deg from +/-180 longitude and +/-90 latitude, and fall back to
   the 2nd/98th percentile of a dense interior grid when the projected
-  extent is more than 50x the source extent. Matches the historical
-  behaviour. Emits a ``UserWarning`` when the clamp or percentile
-  fallback actually alters the bounds.
+  extent shows a real singularity blow-up. Blow-up detection is
+  unit-agnostic: it compares the max absolute projected coordinate
+  to the median and also flags non-finite raw edge projections.
+  Emits a ``UserWarning`` when the clamp or percentile fallback
+  actually alters the bounds.
 - ``"raw"``: use the true projected extent of the source corners and
   edges. No clamp, no percentile. Pass this when downstream tooling
   needs the exact projection of the input footprint and you are willing
--- a/xrspatial/reproject/_grid.py
+++ b/xrspatial/reproject/_grid.py
@@ -252,17 +252,51 @@ def _compute_output_grid(source_bounds, source_shape, source_crs, target_crs,
         # interior grid instead of just boundary points.
         raw_bounds = (float(tx.min()), float(ty.min()),
                       float(tx.max()), float(ty.max()))
-        src_w_crs = src_right - src_left
-        src_h_crs = src_top - src_bottom
-        out_w_crs = raw_bounds[2] - raw_bounds[0]
-        out_h_crs = raw_bounds[3] - raw_bounds[1]
 
-        # Heuristic: if output span is > 50x source span in either axis,
-        # boundary points likely wrapped around a singularity. Fall back
-        # to a dense interior grid to get a tighter bounding box.
-        ratio_x = out_w_crs / (abs(src_w_crs) + 1e-30)
-        ratio_y = out_h_crs / (abs(src_h_crs) + 1e-30)
-        auto_blowup = (ratio_x > 50 or ratio_y > 50)
+        # Unit-agnostic blow-up detection. Comparing source span (in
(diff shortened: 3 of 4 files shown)


brendancol (commented):
## PR Review

### Blockers
None.

### Suggestions

- [ ] `xrspatial/reproject/_grid.py:271-294`. The non-finite probe duplicates the edge-sampling pattern from lines 215-226, just with the unclamped source coordinates. A small helper would keep the two samplers in sync if `n_edge` is ever tuned. Optional given the surgical-changes rule, but flagging it.

### Nits

- [ ] `xrspatial/reproject/_grid.py:299`. The `nonfinite_frac > 0.05` threshold is reasonable but unexplained. A one-line comment ("5% means edges hit a singularity; benign cases yield 0%") would help future maintainers understand the choice.
- [ ] `xrspatial/reproject/_grid.py:295-298`. The magnitude-threshold comment mentions "ratio ~1e16" for the opposite-pole case. Worth also noting that the post-clamp Mercator-global case has a ratio around 5 (caught only by the non-finite probe), so the reader sees why both signals are needed.

### What looks good
- The bug is reproducible before the fix: 106 km x crop, 70 km y crop on (-10, -10, 10, 10) into EPSG:3857.
- Two-signal design cleanly separates the pathological cases. The magnitude ratio catches finite-but-huge singularities like polar-stereographic on the opposite pole. The non-finite probe catches Mercator at the poles after clamp would otherwise have hidden the issue.
- Tests cover the regression case, the silent-no-warning case, the polar-stereographic pathological case, and a dask-backed regression.
- CHANGELOG and docs both updated.
- All 13 existing TestBoundsPolicy tests pass; full reproject suite of 354 tests passes.
- The non-finite probe is gated on `bounds_policy == "auto" and source_crs.is_geographic` so the extra pyproj call only fires when needed.

### Checklist
- [x] Algorithm matches the issue's suggested approach
- [x] Backends consistent
- [x] NaN handling correct
- [x] Edge cases covered
- [x] Dask test added
- [x] Benchmark not required
- [x] README not applicable
- [x] Docstrings updated

brendancol (commented):
## Follow-up review

Addressed all three findings from the previous review.

### Fixed
- Suggestion (`_grid.py:271-294`): extracted `_edge_samples()` helper; both the main edge sampler and the non-finite probe now share the same edge construction.
- Nit (`_grid.py:299`): added a comment explaining why 5% is the non-finite threshold (benign cases yield 0%, real singularities yield 15%+, the floor absorbs numerical noise).
- Nit (`_grid.py:295-298`): expanded the magnitude-threshold comment so the reader sees why both signals are necessary. The post-clamp global Mercator case has ratio ~5, well under 10x; the non-finite probe is what catches that case.

All 17 TestBoundsPolicy tests still pass; full reproject suite of 358 tests passes.

PR #2601, opened 2026-05-28
polygonize: fix dask chunk-dependent merging and DN corruption

### LINKED ISSUE #2583: Dask polygonize: chunk-dependent merging and DN corruption for near-equal floats
**Describe the bug**

`polygonize()` on dask-backed float rasters produces different geometry counts and DN values than the numpy reference when chunks split chains of near-equal floats. The dask path buckets chunk-boundary polygons by a single representative float value before merging (`xrspatial/polygonize.py:295`, `xrspatial/polygonize.py:1739`). That bucketing ignores spatial topology, so two failure modes show up depending on whether the close-valued pixels are spatially connected or disconnected.

**Symptom A: geometry count differs from numpy**

```python
import numpy as np
import xarray as xr
import dask.array as da

arr = np.array([[1.0, 1.000009, 1.000018]])
v_np, _ = polygonize(xr.DataArray(arr))                         # len == 1
v_d,  _ = polygonize(xr.DataArray(da.from_array(arr, (1, 1))))  # len == 2
```

NumPy's CCL chains `1.0 -> 1.000009 -> 1.000018` because the middle pixel is within `atol+rtol*|val|` of both ends, so all three pixels land in one region. The dask cross-chunk merge compares each value pairwise against existing bucket keys, so `1.0` and `1.000018` (gap 18e-6, default rtol threshold 1e-5) never share a bucket and the chain breaks. Result: 2 polygons instead of 1.

**Symptom B: silent DN corruption on disconnected regions**

```python
arr = np.array([[1.0, 9.0, 1.000009]])
v_np, _ = polygonize(xr.DataArray(arr))                         # DN = [1.0, 9.0, 1.000009]
v_d,  _ = polygonize(xr.DataArray(da.from_array(arr, (1, 1))))  # DN = [1.0, 9.0, 1.0]
```

Here the `1.0` and `1.000009` pixels are spatially separated by the `9.0` pixel. NumPy reports three distinct regions with their original values. The dask path drops `1.000009` and `1.0` into the same float bucket (`abs(1.0 - 1.000009) <= 1e-8 + 1e-5*1.0`), and the bucket's representative key wins, so the third region's DN silently flips from `1.000009` to `1.0`. The geometry count is plausible, so this fails quietly and corrupts downstream attribute joins.

**Expected behavior**

Dask polygonize should match the numpy reference on the same input for both geometry count and DN values. The cross-chunk merge needs spatial topology: union chunk-boundary polygons that are both close in value (under `atol`/`rtol`) AND spatially adjacent (share a vertex at a chunk edge). Disconnected polygons that happen to be float-close must keep their original DN values.

**Existing test that pins the bug**

`xrspatial/tests/test_polygonize_atol_rtol_backend_coverage_2026_05_27.py:204` (`test_default_tolerance_still_merges_multi_chunk`) currently asserts `len(v_dc) == 2` for the symptom-A input and documents the pairwise-bucket behaviour as intentional. That test should be updated to assert numpy/dask parity instead.

**Reference locations**

- `xrspatial/polygonize.py:295` (`_bucket_key_for_value`): value-only bucketing
- `xrspatial/polygonize.py:1739` (`_merge_from_separated`): DN value emission for the (val, polys_list) pairs

**Additional context**

The integer path is unaffected because integer bucketing uses strict equality. The bug is float-specific and shows up on every dask backend (dask+numpy and dask+cupy share `_polygonize_dask`).

brendancol:
## Summary

Fixes the dask polygonize value-bucket bug where the cross-chunk merge produced different polygon counts and DN values than the numpy reference on float rasters. Closes #2583.

- Boundary polygons now carry a per-region `(val_min, val_max)` so the cross-chunk merge can match on the closest endpoint of the range instead of a single representative float.
- `_group_boundary_polygons` replaces the value-only bucket with a union-find over spatial adjacency AND value-range closeness, matching numpy CCL semantics across chunk boundaries.
- Spatial adjacency is connectivity-aware: unit-edge sharing under 4-connectivity, vertex sharing under 8-connectivity.

## Backend coverage

- dask+numpy: covered by new tests in `test_polygonize_issue_2583.py` and the updated default-tolerance pins.
- dask+cupy: same code path; covered by the `cuda_and_cupy_available` classes in `test_polygonize_issue_2583.py` and the updated `test_polygonize_atol_rtol_backend_coverage_2026_05_27.py` pin.
- numpy / cupy single-chunk paths: unchanged.

## Test plan

- [x] New `test_polygonize_issue_2583.py` covers both symptoms across dask+numpy and dask+cupy.
- [x] Updated existing pins that asserted the broken behaviour to assert numpy/dask parity instead.
- [x] Full polygonize test suite green locally (286 passed, 16 skipped).
- [x] Rest of the suite green locally (4008 passed, 4 skipped).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@
 
 #### Fixed
 
+- `polygonize` on dask-backed float rasters now matches the numpy
+  reference for both polygon count and DN values across every chunking
+  pattern. The cross-chunk merge previously bucketed chunk-boundary
+  polygons by a single representative float value, which broke
+  transitive value chains (`1.0 -> 1.000009 -> 1.000018` returned 2
+  polygons instead of the numpy reference's 1) and silently rewrote DN
+  values on disconnected near-equal regions (`[1.0, 9.0, 1.000009]`
+  reported DN `1.0` for the third region instead of `1.000009`). The
+  merge now runs a spatial-topology + value-closeness union-find over
+  chunk-boundary polygons: two polygons land in the same group only
+  when their pixel value ranges overlap within `atol`/`rtol` AND they
+  are spatially adjacent (sharing a unit edge for 4-connectivity, or
+  any vertex for 8-connectivity). Disconnected close-valued regions
+  keep their own DN values, and transitive value chains merge
+  correctly across chunk boundaries. (#2583)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -324,6 +324,203 @@ def _bucket_key_for_value(boundary_by_value, val,
     return val
 
 
+def _ranges_close(range_a, range_b, atol, rtol):
+    """Return True if any value pair across two ``(min, max)`` ranges
+    compares close under ``_values_close``.
+
+    Each per-polygon value range from ``_polygonize_chunk`` describes
+    the actual span of pixel values that fall inside a region (after
+    the within-chunk tolerance-chain CCL has run).  Two chunk polygons
+    should union when *some* pair of their pixel values lies within
+    tolerance, which is what numpy CCL would have done if the same
+    pixels lived in one chunk.  Probing the four endpoint combinations
+    is enough because each input range is a single interval.
+    """
+    a_min, a_max = range_a
+    b_min, b_max = range_b
+    if _values_close(a_min, b_min, atol, rtol):
+        return True
+    if _values_close(a_min, b_max, atol, rtol):
+        return True
+    if _values_close(a_max, b_min, atol, rtol):
+        return True
+    if _values_close(a_max, b_max, atol, rtol):
+        return True
+    # Overlapping intervals always have a close pair (zero distance).
+    if a_min <= b_max and b_min <= a_max:
+        return True
+    return False
+
+
+def _group_boundary_polygons(boundary_polys,
+                             atol: float = _DEFAULT_ATOL,
+                             rtol: float = _DEFAULT_RTOL,
+                             connectivity_8: bool = False):
+    """Group chunk-boundary polygons by spatial topology + value closeness.
+
+    Replaces the legacy value-only bucket in ``_merge_chunk_polygons`` and
+    ``_polygonize_dask`` so dask polygonize matches numpy CCL semantics
+    (#2583).  Two boundary polygons land in the same group when:
+
+    1. their value ranges have at least one close pair under ``atol`` /
+       ``rtol``, AND
+    2. their rings share enough spatial topology to be connected under
+       the user-selected connectivity:
+
+       * ``connectivity_8=False`` (4-conn): they must share at least one
+         opposing unit-length ring edge.  Two polygons that only touch
+         at a corner are not 4-connected and stay in separate groups
+         even when their values are close.
+
+       * ``connectivity_8=True`` (8-conn): sharing any integer-grid
+         vertex is enough, mirroring numpy 8-connectivity which merges
+         diagonal-only neighbours of the same value.
+
+    Connectivity is a union-find closure over those pairs.  Disconnected
+    close-valued polygons fall into separate groups so the merge step
+    cannot silently overwrite their distinct DN values.
+
+    The value-range check (rather than a single representative value)
+    fixes the chunk-chain bug where numpy CCL chains
+    ``1.0 -> 1.000009 -> 1.000018`` via the middle pixel: when a chunk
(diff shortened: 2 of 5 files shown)


brendancol (commented):
## PR Review: polygonize: fix dask chunk-dependent merging and DN corruption

None.

- [ ] `xrspatial/polygonize.py:295`: `_bucket_key_for_value` is now dead code. Both call sites (`_merge_chunk_polygons` and `_polygonize_dask`) were rewired to `_group_boundary_polygons` in this PR, so the function is orphaned. Per the project's "remove orphans my changes created" rule, this helper should be deleted in the same PR rather than left behind.

- [ ] `xrspatial/polygonize.py:1131`: `_compute_region_value_ranges` runs a second `_calculate_regions` pass per chunk. For integer dtypes the result is always `(val, val)` because integer CCL uses strict equality, so the second CCL pass produces no new information. Skipping the call when `block.dtype` is integer (return `[(c, c) for c in column]`) would halve per-chunk CCL work in the dominant categorical use case. Float dtypes still need the full pass.

- [ ] `xrspatial/polygonize.py:350`: the overlap test `a_min <= b_max and b_min <= a_max` in `_ranges_close` is correct but the comment "Overlapping intervals always have a close pair (zero distance)" reads as if the four endpoint checks would not catch overlap. They do not always: two wide intervals like `[0.0, 5.0]` and `[3.0, 10.0]` overlap but all four endpoint pairs sit outside default tolerance. Tightening the comment to explain why the overlap check is needed (not just redundant) would help the next reader.

- [ ] `xrspatial/polygonize.py:411`: the unit-edge / unit-vertex walk allocates intermediate Python lists `units` and `points` even when the corresponding key set is not used (e.g. `points` is allocated under `connectivity_8=False`). Building only the active key set inside the `if connectivity_8` branch would avoid the throwaway allocation on hot rasters with many chunk-boundary polygons. Optional micro-optimisation, not load-bearing.

- [ ] `xrspatial/tests/test_polygonize_issue_2583.py:142-152`: `TestNumpyDaskParity2D.test_parity` shapes the disconnected-corner array as `[[1.0, 9.0], [9.0, 1.000009]]` (2-pixel corner pair). Worth also pinning a larger raster (say 5x5) where the close-valued disconnected regions live multiple chunks apart, to guard against future regressions where adjacency probing becomes accidentally global.

### What looks good

- The bug analysis nails both symptoms in the issue and the fix lands at the right layer (cross-chunk merge, not per-chunk CCL).
- Connectivity-aware adjacency (unit-edge for 4-conn, vertex for 8-conn) is the right call. `test_polygonize_issue_2172.py::TestDask8ConnDoesNotRegress4Conn` continues to pass, confirming 4-conn diagonal pairs are not over-merged.
- Walking unit steps along ring edges (line 411 comment) is the right fix for the simplified-collinear-vertex issue; without this, the test_polygonize_issue_2172 suite would have broken on the 11x11 random raster.
- Existing pins that asserted the broken pairwise-bucket behaviour were updated to numpy-parity, not deleted. The audit trail of the change is preserved.
- `_ranges_close` overlap branch correctly handles the wide-interval case that endpoint probing misses.
- CHANGELOG entry under Unreleased covers both symptoms, the fix mechanism, and the connectivity branch.

### Checklist

- [x] Algorithm matches reference (numpy CCL semantics): verified by parity tests
- [x] All implemented backends produce consistent results (dask+numpy and dask+cupy share the path)
- [x] NaN handling unchanged: `_polygonize_numpy` masks NaN before CCL, val_min/val_max never see NaN pixels
- [x] Edge cases covered by tests (single-pixel chunks, chunk chains, disconnected near-equals, 8-conn diagonal)
- [x] Dask chunk boundaries handled correctly (the entire fix)
- [x] No premature materialization (boundary polys still accumulate one chunk at a time)
- [ ] Benchmark coverage: no benchmark exists for the dask polygonize merge path; not blocking
- [x] CHANGELOG updated under Unreleased / Fixed
- [x] Docstrings present and accurate on the new helpers

brendancol (commented):
## Follow-up review pass

All five findings from the earlier review have been addressed in commit f895c5aa.

### Disposition

- [x] Suggestion: removed dead `_bucket_key_for_value` (and updated three test-class docstrings that named it for context).
- [x] Suggestion: integer dtypes now skip the second `_calculate_regions` pass and reuse `column` directly; float dtypes keep the full per-region min/max scan.
- [x] Nit: `_ranges_close` overlap comment now explains the wide-interval case that endpoint probes miss.
- [x] Nit: `_group_boundary_polygons` builds only the active key set per connectivity branch; no more throwaway lists.
- [x] Nit: added `test_distant_close_valued_regions_5x5` pinning that two close-valued corners in a 5x5 raster do not collapse when separated by several chunks of a different value.

### Verification

- `xrspatial/tests/test_polygonize*.py` -- 287 passed, 16 skipped on the local run.
- Symptom A and Symptom B reproducers from the issue still report numpy-parity.
- No new blockers or suggestions surfaced on this pass.

PR #2603, opened 2026-05-28
geotiff: accept projected GeoTIFFs that carry a base geographic CRS

### LINKED ISSUE #2602: open_geotiff rejects valid projected GeoTIFFs that carry a base geographic CRS
**Describe the bug**

`open_geotiff` refuses to read standard projected GeoTIFFs. The
`_check_read_inconsistent_geokeys` validator (added in #2417) treats a
file that carries both a `ProjectedCSTypeGeoKey` and a
`GeographicTypeGeoKey` resolving to different EPSG codes as a
contradiction (its "Case 3"), and raises `InconsistentGeoKeysError`
unless the caller passes `allow_inconsistent_geokeys=True`.

That layout is not a contradiction. It is how a projected GeoTIFF is
normally encoded. A UTM file declares `ProjectedCSTypeGeoKey=32611`,
the projected CRS the pixel coordinates live in, alongside
`GeographicTypeGeoKey=4326`, the base geographic CRS the projection is
built on. The two codes are meant to differ.

The reader already resolves this correctly. `extract_geo_info` in
`_geotags.py` (around line 855) takes the projected EPSG first and only
falls back to the geographic code when no projected code is present.
The validator rejects the file before that resolution ever runs.

**Expected behavior**

When a GeoTIFF carries both a projected and a geographic coordinate
system, treat the geographic CRS as the base geographic CRS and read
the coordinates in the projected CRS. The file should open without any
flag.

Remove the `allow_inconsistent_geokeys` flag from `open_geotiff` and
the internal reader plumbing that threads it.

The two genuinely contradictory shapes keep raising
`InconsistentGeoKeysError`:

- Case 1: `ModelTypeGeoKey=projected` but only `GeographicTypeGeoKey`
  is populated, with no projected code.
- Case 2: `ModelTypeGeoKey=geographic` but `ProjectedCSTypeGeoKey` is
  populated.

With the flag gone there is no longer an opt-out for those two.

**Additional context**

Scope is the validator's Case 3 only. The flag is removed everywhere it
threads (`open_geotiff` in `__init__.py`, `_attrs.py`, and the
`_backends/{gpu,vrt,dask}.py` readers). The Case 1 and Case 2 error
messages and the safe-IO user guide need their references to the flag
dropped.

brendancol:
Closes #2602

## What changed

- `open_geotiff` now reads a projected GeoTIFF that also carries its base
  geographic CRS. The `_check_read_inconsistent_geokeys` validator used to
  reject "both `ProjectedCSTypeGeoKey` and `GeographicTypeGeoKey` resolve
  to different EPSG codes" (its Case 3), which is the normal UTM-over-WGS84
  shape. The reader already takes the projected code first, so dropping
  that case lands the projected EPSG on `attrs['crs']` and keeps the
  geographic key as the base geographic CRS.
- Removed the `allow_inconsistent_geokeys` read kwarg from `open_geotiff`,
  `read_geotiff_dask`, `read_vrt`, `read_geotiff_gpu`, and the internal
  reader plumbing (`_attrs.py`, `_validation.py`).
- The two genuine `ModelTypeGeoKey` contradictions still raise
  `InconsistentGeoKeysError` (Case 1: projected model, only
  `GeographicTypeGeoKey`; Case 2: geographic model carrying
  `ProjectedCSTypeGeoKey`). With the flag gone there is no opt-out for
  those.

## Backend coverage

The validator runs on the shared read-finalization path, so the change
applies to all read backends: numpy (eager), dask+numpy, cupy, and
dask+cupy. The flag was removed from every backend signature.

## Test plan

- [x] `test_validate_accepts_both_type_keys_with_different_epsg` (unit) and
      `test_open_geotiff_accepts_both_keys_with_different_epsg` (file-level)
      cover the fix; the projected EPSG surfaces on `attrs['crs']`.
- [x] Case 1 / Case 2 reject tests still pass.
- [x] Full geotiff suite: 6007 passed, 81 skipped, 1 xfailed.
- [x] Signature-contract and release-gate feature tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@
 
 #### Fixed
 
+- `open_geotiff` no longer rejects a projected GeoTIFF that also carries
+  its base geographic CRS. A file declaring both `ProjectedCSTypeGeoKey`
+  (e.g. UTM 32633) and `GeographicTypeGeoKey` (e.g. WGS84 4326) is the
+  normal projected shape: the geographic key names the base geographic
+  CRS and the coordinates live in the projected CRS. The reader already
+  resolves the projected code first, so it now reads cleanly and stamps
+  the projected EPSG on `attrs['crs']`. The `allow_inconsistent_geokeys`
+  read kwarg is removed; the two genuine `ModelTypeGeoKey` contradictions
+  (a projected model with only `GeographicTypeGeoKey`, or a geographic
+  model carrying `ProjectedCSTypeGeoKey`) still raise
+  `InconsistentGeoKeysError`. (#2602)
+
 - `polygonize` now auto-detects the raster's affine transform from
   `attrs['transform']` (xrspatial.geotiff convention) or
   `rio.transform()` (rioxarray) when the caller did not pass an
--- a/docs/source/user_guide/geotiff_safe_io.rst
+++ b/docs/source/user_guide/geotiff_safe_io.rst
@@ -225,10 +225,12 @@ the ambiguous-metadata family at once.
    * - :class:`~xrspatial.geotiff.InconsistentGeoKeysError`
      - The source's GeoKey directory is internally contradictory:
        ``ModelTypeGeoKey`` disagrees with the type-specific keys
-       actually populated, or ``ProjectedCSTypeGeoKey`` and
-       ``GeographicTypeGeoKey`` resolve to different EPSG codes.
-     - ``allow_inconsistent_geokeys=True`` to keep the legacy silent
-       acceptance for known-quirky historical files.
+       actually populated (a projected model with only
+       ``GeographicTypeGeoKey``, or a geographic model carrying
+       ``ProjectedCSTypeGeoKey``). A projected file that also names its
+       base geographic CRS is the normal shape and reads without error.
+     - No opt-in. Fix the source's GeoKey directory so the model type
+       matches the CRS codes it declares.
    * - :class:`~xrspatial.geotiff.ConflictingNodataError`
      - ``attrs['nodata']`` and ``attrs['nodatavals']`` disagree on
        write.
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -462,7 +462,6 @@ def open_geotiff(source: str | BinaryIO, *,
                  missing_sources: str = _MISSING_SOURCES_SENTINEL,
                  allow_rotated: bool = False,
                  allow_unparseable_crs: bool = False,
-                 allow_inconsistent_geokeys: bool = False,
                  allow_invalid_nodata: bool = False,
                  stable_only: bool = False,
                  allow_experimental_codecs: bool = False,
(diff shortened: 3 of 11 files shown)


brendancol (commented):
## PR Review: accept projected GeoTIFFs that carry a base geographic CRS

This is a focused validation change, not an algorithm or kernel change, so the numerical-accuracy, CUDA, and `map_overlap` sections of the checklist do not apply. I read the full diff and the surrounding code in `_validation.py`, `_attrs.py`, `__init__.py`, and the three backends.

None.

- `_validation.py` Case 3 removal means the reader now trusts `ProjectedCSTypeGeoKey` even when `GeographicTypeGeoKey` names a base CRS that disagrees with the projection's actual datum (e.g. projected 32633 over a declared 4267/NAD27). That is the intended behaviour for this issue (projected wins, geographic is the base), but it is also the one case the old validator caught that had a real-error flavour. The current test only exercises 32633 over 4326, where the declared base matches the projection. A short comment on the test, or a second case with a mismatched base, would record that the silent-accept is deliberate rather than untested. Not blocking.

- The CHANGELOG entry lives under `Fixed`. It also removes the `allow_inconsistent_geokeys` public kwarg that shipped in 0.10.0, which is a breaking change for anyone who adopted it. The Fixed note does mention the removal, so this is borderline, but a one-line `Changed`/`Removed` callout would be easier to spot in release notes.

### What looks good

- The fix leans on the resolver that already existed (`extract_geo_info` takes the projected code first), so it deletes a wrong rejection rather than adding new resolution logic. The diff is mostly subtraction (164 deletions, 63 insertions).
- Cases 1 and 2 stay rejected, and their messages were rewritten to drop the now-gone opt-out and point at fixing the source instead.
- The flag was removed cleanly from every backend signature; `test_signatures.py`'s canonical kwarg list and the parity signature-contract tests still pass, which is good evidence nothing drifted out of sync.
- The Case 3 unit and file-level tests were inverted to assert acceptance with `attrs['crs'] == 32633`, so the regression is locked in on both the validator and the public `open_geotiff` surface.

### Checklist

- [x] Algorithm matches reference/paper — N/A (validation change)
- [x] All implemented backends produce consistent results — validator runs on the shared finalize path; flag removed from all backends
- [x] Edge cases are covered by tests — both-keys, user-defined, same-EPSG, no-context all covered
- [ ] README feature matrix updated — N/A (no new function, no backend change)
- [x] Docstrings present and accurate — validator, error class, and safe-IO guide updated

brendancol (commented):
## Follow-up review (after `e7a54ac`)

- **Suggestion (record that silent-accept of a mismatched base is deliberate)** — fixed. Added `test_validate_accepts_projected_with_unexpected_base_geographic`, which feeds UTM 32633 over NAD27 4267 (a base the projection was not built on) and asserts the validator still passes. It states in the docstring that this is the one shape old Case 3 rejected and that the accept is by design, so the behaviour is now pinned rather than implicit.

- **Nit (separate CHANGELOG `Removed` callout for the kwarg)** — dismissed. The existing Fixed entry already says the `allow_inconsistent_geokeys` kwarg is removed, and the Unreleased block has no `Removed` section today. Adding one for a single advanced-tier kwarg that was both added and removed inside the 0.10.x dev cycle is more structure than the change warrants.

No new findings. The merge with `main` only touched `CHANGELOG.md` (kept both sides); the rest auto-merged and the geotiff suite is green.

PR #2617, opened 2026-05-29
zonal: fix int32 overflow in _strides() that silently corrupts stats/crosstab on huge rasters

### LINKED ISSUE #2612: stats()/crosstab() silently wrong on very large rasters: int32 overflow in _strides()
**Describe the bug**

`_strides()` in `xrspatial/zonal.py` allocates its cumulative zone-boundary counter as an `np.int32` array:

```python
strides = np.zeros(len(unique_zones), dtype=np.int32)
```

The `count` it accumulates runs up to the number of flattened input pixels. Since `_strides` is a `@ngjit` (numba nopython) kernel, storing a `count` above 2^31-1 into the int32 array wraps silently to a negative value instead of raising `OverflowError`. That threshold is about 2.1 billion pixels, which a single raster of roughly 46341 x 46341 hits.

`_strides` feeds the zone-boundary breaks that `_calc_stats` and the crosstab paths use as slice bounds (`values_by_zones[start:end]`). A wrapped negative bound produces wrong slices, so `stats()` and `crosstab()` return silently incorrect results on very large numpy-backed rasters, and on dask inputs whose individual chunks exceed the threshold.

**Expected behavior**

Zone-boundary counts should index the full input without overflowing at realistic raster sizes. Allocate `strides` as `int64` to match the natural array-index type.

**Additional context**

Found during a security sweep of the zonal module (integer overflow category). Severity: high. The failure mode is silent wrong results, not a crash.

brendancol:
Closes #2612

## What changed

- `_strides()` accumulated cumulative zone-boundary counts into an `np.int32` array inside a numba `@ngjit` kernel. For rasters above ~2.1 billion pixels (about 46341 x 46341) the count wraps silently to a negative int32, so the slice bounds `_calc_stats` and the crosstab paths derive from it go negative and `stats()`/`crosstab()` return silently wrong results.
- Allocate the counter as `int64`, matching the natural array-index type, so it cannot overflow at realistic raster sizes.
- Added a regression test asserting `_strides` returns an `int64`, non-negative, monotonic result.

## Backend coverage

`_strides` runs in the numpy and dask+numpy paths (and the cupy/dask+cupy paths, which route through the numpy implementation). The dtype fix applies to all of them.

## Test plan

- [x] `test_strides_int64_no_overflow_2612` passes
- [x] Existing stats and crosstab tests pass (111 passed)

### DIFF
--- a/xrspatial/tests/test_zonal.py
+++ b/xrspatial/tests/test_zonal.py
@@ -2566,6 +2566,9 @@ def test_strides_int64_no_overflow_2612():
     """
     from xrspatial.zonal import _strides
 
+    # A real >2**31-element input would need ~17 GB to allocate, so the
+    # dtype assertion below stands in for the actual overflow case; a
+    # small input is enough to lock in the int64 contract and correctness.
     flatten_zones = np.array([0, 0, 0, 1, 1, 2], dtype=np.int64)
     unique_zones = np.array([0, 1, 2], dtype=np.int64)
 
--- a/xrspatial/tests/test_zonal.py
+++ b/xrspatial/tests/test_zonal.py
@@ -2554,3 +2554,30 @@ def test_stats_return_type_invalid_rejected_for_dataset_2558(
     ds = xr.Dataset({'v': values})
     with pytest.raises(ValueError, match="return_type"):
         stats(zones=zones, values=ds, return_type='bogus')
+
+
+def test_strides_int64_no_overflow_2612():
+    """`_strides` must accumulate counts in int64, not int32.
+
+    The counter runs up to the flattened pixel count. In an int32 array
+    inside the numba kernel, a count above 2**31-1 wraps silently to a
+    negative value, corrupting the slice bounds that `_calc_stats` and
+    crosstab derive from it. See issue #2612.
+    """
+    from xrspatial.zonal import _strides
+
+    flatten_zones = np.array([0, 0, 0, 1, 1, 2], dtype=np.int64)
+    unique_zones = np.array([0, 1, 2], dtype=np.int64)
+
+    strides = _strides(flatten_zones, unique_zones)
+
+    # The result dtype is the load-bearing guarantee: int64 cannot wrap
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: fix int32 overflow in _strides()

### Blockers
None.

### Suggestions
None.

### Nits
- The test pins the dtype and checks correctness on a tiny input. It can't actually trigger the wrap without a ~17 GB array, so the dtype assertion is doing the real work here. A one-line comment to that effect would save the next reader a moment. Not blocking.

### What looks good
- Root cause is right. In a numba nopython kernel, storing a count past 2**31-1 into an int32 array wraps silently instead of raising, so int64 is the correct fix and it matches how the value is later used (as a slice index).
- Minimal change, no surrounding churn.
- `_strides` is the one chokepoint feeding `zone_breaks` into `_calc_stats` and the crosstab paths, so this covers `stats()` and `crosstab()` on numpy, dask+numpy, and the cupy/dask+cupy paths (which route through the numpy code). Nothing else to touch per-backend.
- Downstream uses these as plain Python slice bounds, and the index arrays come from `np.argsort` (already int64 on 64-bit), so there's no other int32 narrowing waiting to reintroduce this.
- Existing stats/crosstab tests still pass (111), so the wider dtype didn't move results at normal sizes.

### Checklist
- Algorithm vs reference: n/a (overflow fix)
- Backends consistent: yes, one shared codepath
- NaN handling: unchanged
- Edge cases: dtype + correctness covered; a real overflow isn't allocatable
- Dask boundaries: unchanged, per-block strides still correct
- Premature materialization: none
- Benchmark: not needed
- README matrix: n/a
- Docstrings: accurate

brendancol (commented):
Follow-up: addressed the one nit from the previous review. The test now carries a comment explaining that a real >2**31-element input isn't allocatable (~17 GB), so the int64 dtype assertion stands in for the overflow case. No other findings were open. Regression test still passes.

PR #2621, opened 2026-05-29
Add rasterize validation error-path test coverage (#2614)

### LINKED ISSUE #2614: Test coverage: rasterize() width/height and resolution input-validation error paths
## Gap

The test-coverage deep-sweep found two groups of validation error paths in `rasterize()` (`xrspatial/rasterize.py`) that no test exercises. These are pure-Python guards that run before any backend dispatch, so they need no CUDA.

### Finding A: partial width/height

When exactly one of `width` / `height` is passed, `rasterize()` raises `ValueError` ("width was provided but height was not. Pass both width and height together ..."). The docstring documents this as intended behavior, but no test covers either the width-only or height-only branch.

### Finding B: resolution input type and shape

`resolution=` has several input-validation branches that each raise a distinct `ValueError`:

- non-number / non-sequence (e.g. a string or dict) -> "must be a number or a length-2 sequence"
- numpy array with the wrong number of dimensions -> "array must be 1-D with length 2"
- sequence with the wrong length (1 or 3+) -> "sequence must have length 2"
- sequence whose elements are not numbers -> "sequence elements must be numbers"

The existing `test_invalid_resolution_scalar` / `test_invalid_resolution_tuple` tests only cover non-finite or non-positive *values*. The type and shape branches above have no coverage, so a regression that loosened or reordered them would ship silently.

## Proposed change

Add a test-only file `xrspatial/tests/test_rasterize_coverage_2026_05_29.py` that pins each branch to its expected `ValueError` (and message). No source changes.

brendancol:
Closes #2614

Test-only change. Pins two groups of pure-Python validation guards in `rasterize()` that no test exercised after the earlier coverage passes. No source changes.

- Partial width/height: width-only and height-only calls each raise the documented `ValueError`. Also pins that the guard fires before the resolution branch.
- `resolution=` input type and shape: non-number/non-sequence (string, dict), wrong-ndim numpy array, wrong-length sequence (1 or 3+), and non-numeric elements each raise their distinct `ValueError`. The existing value-guard tests only cover non-finite or non-positive values. A positive control confirms a 1-D length-2 numpy array is still accepted.

Backend coverage: these guards run before backend dispatch, so the tests are backend-independent (no numpy/cupy/dask split needed). They run without CUDA.

Test plan:
- [x] `pytest xrspatial/tests/test_rasterize_coverage_2026_05_29.py` -> 11 passed
- [x] Confirmed each test's expected message matches what the source raises


brendancol (commented):
Domain-aware review (test-coverage sweep, self-review).

Scope check: the diff is one new test file plus a one-line append to the sweep state CSV. No source files touched, which matches the stated test-only intent.

Blockers: none.

Suggestions: none. The error-path patterns match what `rasterize()` actually raises (verified independently against the live function), and each `pytest.raises` is anchored on a substring of the real message rather than the full string, so it stays robust to minor wording changes while still pinning the right branch. The positive control (`test_length_2_numpy_array_accepted`) guards against an over-eager shape check rejecting a valid 1-D length-2 array.

Nits:
- The `resolution=` shape branches are ordering-sensitive in the source (ndim check before length check). The tests pin each branch by its distinct message, so a reordering regression that changed which message fires would surface. Good.
- Tests are backend-independent because the guards run before backend dispatch; no cupy/dask variants are needed and none were added, which is correct rather than a gap.

Verification: `pytest xrspatial/tests/test_rasterize_coverage_2026_05_29.py` -> 11 passed.

PR #2622, opened 2026-05-29
zonal: make apply() output .name deterministic across backends

### LINKED ISSUE #2611: zonal.apply() output .name is inconsistent across backends
## Description

`xrspatial.zonal.apply()` returns a DataArray but never sets the result's `.name`, so the value of `.name` depends on which backend ran:

- numpy and cupy: `.name` is `None`
- dask+numpy and dask+cupy: `.name` is an internal dask task name such as `_chunk_fn-3e337fb640ced2398a1cced35ab359af`

The dask name is a graph-task hash, so it is non-deterministic between runs too.

Every other DataArray-returning function in `zonal.py` sets a deterministic name and exposes a `name=` parameter: `regions` -> `'regions'`, `hypsometric_integral` -> `'hypsometric_integral'`, `trim` -> `'trim'`, `crop` -> `'crop'`. `apply()` is the odd one out.

A pipeline that chains `apply(...)` into `.to_dataset()`, or otherwise relies on `.name`, gets a different (and on dask, meaningless) result depending on the backend.

## Expected behavior

`apply()` should set a stable `.name` on its output and accept a `name` parameter, matching the rest of the zonal functions.

## Proposed fix

Add a `name` parameter to `apply()` (default `None`) and pass it to the returned `xr.DataArray`. This makes `.name` deterministic and identical across all four backends.

## Backends affected

brendancol:
Closes #2611

## What changed

- `zonal.apply()` now sets a deterministic `.name` on its output and accepts a `name` parameter (default `None`), matching `regions`, `hypsometric_integral`, `trim`, and `crop`.
- The name is assigned after the DataArray is built. `xr.DataArray(dask_array, name=None)` otherwise inherits the dask graph's internal task name, which differs from the numpy/cupy backends and changes between runs.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy. The new test runs on all four; before this change the two dask backends returned a task-hash name while numpy/cupy returned `None`.

## Test plan

- [x] `test_apply_name_consistent_across_backends` (new): default name is `None` and an explicit `name=` is honored on every backend
- [x] Full zonal suite: 213 passed
- [x] flake8 clean on `xrspatial/zonal.py` and the added test lines


brendancol (commented):
## PR Review: zonal: make apply() output .name deterministic across backends

None.

None.

- The new `name` parameter goes at the end of the signature, after `rasterize_kw`. That keeps existing positional calls working and matches how `regions`/`trim`/`crop` expose `name`, so it is fine as-is. Noting the placement was a deliberate compatibility choice.

### What looks good
- Root cause is correct: `xr.DataArray(dask_array, name=None)` inherits the dask graph's task name, so assigning `.name` after construction is the right fix. Setting it at construction time does not override the dask name.
- The test runs on all four backends and checks both the default (`None`) and an explicit name, which is exactly the inconsistency being fixed.
- Attrs, dims, and coords handling is untouched; only `.name` changed.
- The code comment explains why the assignment happens after construction, which is the non-obvious part.

### Checklist
- Algorithm matches reference/paper: n/a (metadata fix)
- All implemented backends produce consistent results: yes, verified by the new parametrized test
- NaN handling is correct: unchanged by this PR
- Edge cases covered by tests: name default and explicit name both tested
- No premature materialization or unnecessary copies: confirmed, no compute added
- Docstrings present and accurate: the new `name` parameter is documented

PR #2624, opened 2026-05-29
resample: cover cupy non-integer-factor aggregate CPU fallback (#2615)

### LINKED ISSUE #2615: resample: cupy aggregate non-integer-factor CPU fallback path is untested
## Description

The eager cupy aggregate path in `_run_cupy` (xrspatial/resample.py, lines
957-973) has two branches:

1. A GPU reshape+reduce path that runs when the downsample factor is an
   integer and the method is `average`, `min`, or `max`.
2. A CPU fallback that copies the array to host, runs the numpy
   `_AGG_FUNCS` kernel, and copies the result back. It handles two cases:
   `median`/`mode` (always), and `average`/`min`/`max` when the downsample
   factor is not an integer.

`TestCuPyParity.test_aggregate_parity` uses a 12x12 raster at
`scale_factor=0.5`, so the factor is exactly 2. That hits the GPU path for
average/min/max and the CPU fallback for median/mode. The average/min/max
CPU fallback for a non-integer factor (e.g. a 10x10 raster at
`scale_factor=0.3`, which gives a 3x3 output and a factor of 3.33) is never
invoked by any test. A regression in the branch test (`fy == int(fy)`) or
in the host round-trip for that path would ship undetected.

## Proposed change

Add a cupy parity test that downsamples by a non-integer factor with
`average`/`min`/`max`, checked against the numpy reference. Test-only, no
source change.

## Backend coverage

cupy, checked against numpy. Needs a CUDA host.

brendancol:
Closes #2615.

Test-only coverage closure from the deep-sweep test-coverage pass on `resample`.

## What

- Add `test_resample_cupy_agg_fallback_2615.py`: 6 tests covering the
  `average`/`min`/`max` host fallback inside `_run_cupy` when the downsample
  factor is not an integer (10x10 at scale 0.3 gives a 3x3 output, factor
  3.33), checked against the numpy reference, plus a NaN-masked variant.
- The prior cupy parity test used scale 0.5 (factor 2, integer), which hits
  the GPU reshape path for average/min/max and only sends median/mode to the
  host fallback. The non-integer average/min/max fallback had no test.

No source change.

## Backend coverage

cupy, checked against numpy. The tests carry the project GPU-skip decorator
and were run on a CUDA host (6 passed).

## Test plan

- [x] `pytest xrspatial/tests/test_resample_cupy_agg_fallback_2615.py` (6 passed on GPU)
- [x] Full resample suite green (273 passed)
- [x] Verified the tests exercise the `fy == int(fy)` fallback branch


brendancol (commented):
## PR Review: resample cupy non-integer-factor aggregate CPU fallback (#2615)

None.

None.

- `xrspatial/tests/test_resample_cupy_agg_fallback_2615.py`: the NaN-variant test does not re-assert the fallback-branch premise (the `fy == int(fy)` guard) the way the first test does. It reuses the same 10x10 scale-0.3 geometry so it does land on the fallback, but a future edit to the shape would not trip an assertion. Low value, since the parity check itself would still fail loudly if the branch changed.

### What looks good
- The first test pins the premise with an explicit assertion that `10 / round(10 * 0.3)` is non-integer, so it cannot silently drift onto the GPU reshape path and stop covering the fallback.
- Tolerance and `equal_nan` match the existing `TestCuPyParity` convention, so the new tests read consistently with the file they extend.
- The NaN-masked variant feeds the host fallback through the NaN-aware numpy kernels, which is the part most likely to diverge.
- Test-only, source untouched, gated behind the GPU-skip decorator, and verified on a CUDA host (6 passed).

### Checklist
- Algorithm matches reference: N/A (test-only)
- Backends consistent: cupy checked against numpy reference
- NaN handling correct: covered by the NaN variant
- Edge cases: non-integer factor is the target; square 10x10 only
- Dask chunk boundaries: N/A (eager cupy path)
- README matrix: not needed (no API change)
- Docstrings present: module and class docstrings explain the gap

brendancol (commented):
## Follow-up review (#2615)

The single nit from the prior pass is addressed: the NaN-variant test now carries the same `10 / round(10 * 0.3)` non-integer premise assertion as the no-NaN test, so both stay pinned to the host fallback path. Re-ran the file on a CUDA host: 6 passed.

No remaining findings. Test-only, source untouched.

PR #2626, opened 2026-05-29
reproject: rename vertical CRS kwargs to source_/target_vertical_crs (#2613)

### LINKED ISSUE #2613: reproject(): vertical CRS params use src_/tgt_ while horizontal use source_/target_
## Summary

`reproject()` in `xrspatial/reproject` spells the source/target concept two ways in one signature.

Horizontal CRS parameters use full words: `source_crs`, `target_crs`.

Vertical datum parameters use the abbreviated forms: `src_vertical_crs`, `tgt_vertical_crs`.

A caller passing both a horizontal and a vertical CRS in the same call has to remember that one pair is spelled out and the other is abbreviated.

## Proposed change

Rename the vertical parameters to match the horizontal ones:
- `src_vertical_crs` -> `source_vertical_crs`
- `tgt_vertical_crs` -> `target_vertical_crs`

A rename breaks existing callers, so keep the old names working: accept both spellings, emit a `DeprecationWarning` when an old name is passed, and update the docstring to the new names.

## Scope

Only `reproject()` takes these vertical parameters. `merge()` does not expose them. `itrf_transform()` uses `src`/`tgt` for ITRF frame names, a separate function family that is out of scope here.

Found by the api-consistency sweep (Cat 1, parameter naming drift).

brendancol:
Closes #2613

- Renames `src_vertical_crs` / `tgt_vertical_crs` to `source_vertical_crs` / `target_vertical_crs` so the vertical-datum kwargs match the `source_crs` / `target_crs` spelling already used in the same signature.
- Keeps the old names working through a deprecation shim: passing one emits a `DeprecationWarning`, and passing both the old and new spelling for the same side raises `TypeError`.
- Updates the docstring and migrates the existing vertical-shift tests to the new names.

Backend coverage: the signature is shared across backends (numpy, cupy, dask+numpy, dask+cupy); verified the new and deprecated names on numpy and cupy entry points.

Test plan:
- [x] `pytest xrspatial/tests/test_reproject.py::TestVerticalShift` (14 passed)
- [x] cupy smoke test: new names and deprecated-alias warning both produce identical output


brendancol (commented):
## PR Review: reproject vertical CRS kwarg rename

None.

None.

- Passing the deprecated name with an explicit None (e.g. `src_vertical_crs=None`) still emits the DeprecationWarning, since the shim treats any non-sentinel value as "the old name was used" (`_resolve_deprecated_vertical_kwarg` in `reproject/__init__.py`). That is defensible: the caller did type the old name.

### What looks good
- The sentinel-based shim tells "not passed" apart from an explicit None, so the old and new defaults stay independent.
- Passing both spellings for one side raises TypeError, and the new test covers it.
- `stacklevel=3` points the warning at the user's call site rather than the internal helper.
- The new and old names were verified on both the numpy and cupy entry points and produce identical output.
- No other call sites in xrspatial, docs, or examples use the old names, so the rename is self-contained.

### Checklist
- [x] Algorithm unchanged (signature/shim only)
- [x] Implemented backends consistent (numpy + cupy verified)
- [x] NaN handling unchanged
- [x] Edge cases covered: deprecation warning, both-names conflict
- [x] Benchmark not needed (no compute change)
- [x] Docstrings updated for new and deprecated params

PR #2627, opened 2026-05-29
resample: fix dask nearest/bilinear chunk-seam values on downsampling (#2610)

### LINKED ISSUE #2610: resample: dask nearest/bilinear gives wrong chunk-seam values when downsampling with uneven chunks
**Describe the bug**

`resample()` produces wrong values at chunk boundaries when interpolating (method `nearest` or `bilinear`) a dask-backed raster that is being downsampled, when the input chunks do not divide the array evenly.

The interpolation dask path uses a fixed overlap depth of 1 input pixel for `nearest` and `bilinear` (`_INTERP_DEPTH`). That depth is the right stencil radius for upsampling, but it is too small for downsampling. With block-centered coordinate mapping, an output pixel maps to input coordinate `(o + 0.5) * (in/out) - 0.5`. For a downsample ratio `in/out > 1`, the source coordinate for an output pixel near a chunk seam can land up to about `(in/out)/2` input pixels past the chunk that `_output_chunks` assigned that pixel to. With depth 1 the overlapped block does not contain that source row or column, so `scipy.ndimage.map_coordinates(mode='nearest')` clamps the out-of-range local coordinate to the block edge and returns the wrong source value. Whole output rows or columns at the seam come back wrong.

The aggregate dask path already handles this: it sets `depth = ceil(global_in / global_out)`. The interpolation path does not.

**Reproduce**

```python
import numpy as np

rng = np.random.RandomState(7)
data = rng.rand(50, 50).astype(np.float32)

np_out = resample(create_test_raster(data, backend='numpy', attrs={'res': (1.0, 1.0)}),
                  scale_factor=0.33, method='nearest').values
dk = create_test_raster(data, backend='dask+numpy', attrs={'res': (1.0, 1.0)}, chunks=(11, 13))
dk_out = resample(dk, scale_factor=0.33, method='nearest').values

print(np.nanmax(np.abs(dk_out - np_out)))  # ~0.67, should be ~0
```

The same divergence shows up with `method='bilinear'` (~0.51). The whole bottom output row is pulled from the wrong input row.

**Expected behavior**

The dask interpolation result should match the eager numpy result for the same input, within float32 round-off, regardless of chunk layout. The existing dask parity tests pass only because they use small downsample ratios where depth 1 happens to be enough. Smooth gradients also hide the bug, because the clamped neighbor value sits close to the correct one.

**Fix**

In the interpolation branches of `_run_dask_numpy` and `_run_dask_cupy`, add the downsample stencil radius to the per-axis depth, clamped per axis as today:

```
depth_y = base_depth + ceil((global_in_h / global_out_h) / 2)
depth_x = base_depth + ceil((global_in_w / global_out_w) / 2)
```

Add a regression test that downsamples with a large ratio and chunks that do not divide the input evenly, comparing dask against eager numpy at tight tolerance.

**Affected backends:** dask+numpy and dask+cupy. Eager numpy and cupy are unaffected since they have no chunking.

**Severity:** high. Silent wrong values on a public API, the default method, for ordinary downsampling.

brendancol:
Closes #2610

## What

- The interp dask path used a fixed overlap depth of 1 input pixel for `nearest` and `bilinear`. Block-centered mapping sends an output pixel to input coordinate `(o + 0.5) * (in/out) - 0.5`, which on a downsample can land up to ~`(in/out)/2` pixels past the chunk `_output_chunks` assigned it to. With depth 1 the overlapped block did not contain that source row/column, so `map_coordinates(mode='nearest')` clamped to the block edge and returned wrong values for whole chunk-seam rows.
- Adds a per-axis downsample radius (`_downsample_radius`) to the interp overlap depth in `_run_dask_numpy` and `_run_dask_cupy`, clamped per axis exactly as before. The aggregate path already accounted for this; the interp path now matches.

## Backends

- dask+numpy and dask+cupy: fixed.
- numpy and cupy (eager): unaffected, no chunking involved.

## Test plan

- [x] New `TestInterpDaskDownsampleSeam` covers `nearest`/`bilinear`/`cubic` at ratios 0.33/0.2/0.1 with chunk layouts that do not divide the input evenly, plus an asymmetric-per-axis case. Fails on `main`, passes here.
- [x] Full `test_resample.py` suite: 217 passed (includes cupy and dask+cupy parity on a CUDA host).
- [x] Off-band sweep of 660 ratio/size/chunk/method combinations: dask matches eager numpy within float32 round-off.


brendancol (commented):
## PR Review: resample dask nearest/bilinear chunk-seam fix (#2610)

None.

None.

- `_downsample_radius` re-imports `math` at function scope. `_ensure_min_chunksize` and the aggregate branches already do the same local import, so this matches existing style. A module-level `import math` would be cleaner, but that pattern predates this PR, so leave it for a bug fix.

### What looks good
- Root cause is correctly identified, and the fix mirrors the depth logic the aggregate path already uses. `_downsample_radius` returns 0 for upsampling, so the upsample paths keep depth 1 and pay nothing.
- The augmented depth feeds the `min_size` chunk-sizing, the `_add_overlap` call, and the block function's origin offset from the same value, so they cannot drift apart.
- The per-axis clamp against `global_in - 1` is preserved, so tiny inputs and skinny axes still work. Checked on 2x2, single-chunk, and skinny-column cases.
- Both `_run_dask_numpy` and `_run_dask_cupy` are patched, so dask+numpy and dask+cupy stay in parity. Eager numpy/cupy are untouched.
- Test coverage holds up: the new class fails on main and passes here, uses random data (a gradient would hide the bug), and uses chunk sizes that do not divide the input evenly. An off-band sweep of 660 ratio/size/chunk/method combinations all matched eager numpy within float32 round-off.

### Checklist
- [x] Algorithm matches the block-centered mapping; depth now covers the downsample displacement
- [x] All four backends consistent (cupy and dask+cupy verified on a CUDA host)
- [x] NaN handling unchanged (fix only widens overlap)
- [x] Edge cases covered (small inputs, skinny axes, asymmetric per-axis ratios)
- [x] Dask chunk boundaries handled correctly (the point of the fix)
- [x] Benchmark not needed (bug fix, no new API)
- [x] README feature matrix already lists resample
- [x] Docstrings present; new helper documents the radius rationale

PR #2628, opened 2026-05-29
Add reproject regression tests for strip and constant-value rasters (#2618)

### LINKED ISSUE #2618: reproject(): no regression tests for single-row, single-column, or constant-value rasters
## Description

The reproject() tests cover 1x1 and 2x2 rasters but never run a single-row (1xN) or single-column (Nx1) strip, or a constant-value raster. Those shapes take different code paths:

- A strip has one spatial axis of size 1, which hits the `size < 2` early return in `_validate_regular_axis`. The pixel-resolution math and the resampling kernel then run on a degenerate axis, which is where boundary logic tends to break.
- A constant-value raster has zero gradient, so bilinear and cubic resampling run through the all-equal interpolation path.

I probed all four backends (numpy, dask+numpy, cupy, dask+cupy) by hand and they all return correct output today, so this is a coverage gap, not a live bug. The tests exist to catch a future kernel change that quietly breaks these degenerate inputs.

## Scope

Test-only, no source changes. If a test turns up a real defect, that goes in a separate accuracy issue linked here.

## Coverage to add

- 1xN single-row strip across numpy, dask+numpy, cupy, dask+cupy
- Nx1 single-column strip across the same four backends
- constant-value raster on numpy plus cross-backend parity

GPU tests gated on the existing HAS_CUPY runtime probe.

brendancol:
Closes #2618

Adds regression tests for reproject() input shapes the suite never covered:

- single-row (1xN) and single-column (Nx1) strip rasters, which hit the size<2 early return in `_validate_regular_axis` and run the resampling kernel on a degenerate axis
- constant-value (zero-gradient) raster, which exercises the all-equal interpolation path

New class `TestDegenerateShapeReproject`, 12 tests. The strip cases run across numpy, dask+numpy, cupy, and dask+cupy and assert parity against the numpy result. The constant-raster numpy test asserts every interpolated pixel equals the fill value (zero gradient must stay zero), plus cross-backend parity.

Backend coverage: numpy, dask+numpy, cupy, dask+cupy. GPU tests gated on the existing `HAS_CUPY` runtime probe.

Test-only change, no source modifications. All 12 tests executed and passed on a CUDA host.

Test plan:
- `pytest xrspatial/tests/test_reproject.py::TestDegenerateShapeReproject` -> 12 passed
- full reproject test file collects 392 tests


brendancol (commented):
## PR Review: Add reproject regression tests for strip and constant-value rasters (#2618)

Test-only PR. No source changes, so the review focuses on test correctness, backend parity, and quality.

None.

None.

- `test_single_row_strip_numpy` and `test_single_col_strip_numpy` (test_reproject.py) assert only that the output is 2-D and has at least one finite pixel. That is smoke-level, but it matches the existing `test_1x1_raster` / `test_2x2_raster` convention in the same file, and the cross-backend strip tests do assert numpy parity, so the degenerate axis is value-checked overall. Fine as-is.

### What looks good
- All four backends covered for both strip shapes (numpy, dask+numpy, cupy, dask+cupy), gated on the existing `HAS_DASK` / `HAS_CUPY` probes.
- The constant-raster numpy test asserts every interpolated pixel equals the fill value, a real reference assertion rather than a does-it-run check. Zero gradient must stay zero, and bilinear is forced explicitly.
- `_to_host` computes before `get` for the dask+cupy case, which is the correct order.
- Dask strip tests use ragged chunks along the long axis: (1,4) and (4,1).
- State CSV change is a single appended reproject row; other modules untouched.

### Checklist
- [x] Edge cases covered by tests (1xN, Nx1, constant)
- [x] All implemented backends produce consistent results (asserted via parity)
- [x] NaN handling correct (NaN/finite masks compared across backends)
- [x] Dask chunk boundaries handled (ragged chunks used)
- [x] No premature materialization (compute only at assertion time)
- [x] Docstrings present (class docstring explains the gap and code path)

Issue #2508, opened 2026-05-27
reproject: geoid undulation lookup has half-pixel offset error (up to 2 m)


brendancol:
## Summary

`xrspatial.reproject._vertical._interp_geoid_point` (and the parallel `_datum_grids._grid_interp_point`) compute the bilinear interpolation index as if the geoid grid stored values at pixel-edge anchors, while the grid is actually pixel-center anchored. This produces a systematic half-pixel offset on every lookup.

For the vendored EGM96 15-arcmin grid this corresponds to a roughly 14 km horizontal shift in the geoid sampling location. The resulting error in the geoid undulation N depends on the local gradient; sampled at every 50th pixel center it is:

```
mean abs error : 0.24 m
p95  abs error : 0.69 m
max  abs error : 2.13 m
```

The same offset is present in `_grid_interp_point`, which is used to apply NADCON / OSTN / NTv2 horizontal datum shift grids; for those grids the offset is hundreds of metres of horizontal misregistration (the actual error in degrees depends on grid resolution and local gradient).

## Reproducer

```python
from xrspatial.reproject._vertical import _interp_geoid_point, _load_geoid

data, left, top, res_x, res_y, h, w = _load_geoid("EGM96")
# Pixel (0,0) center: lon=-180, lat=90.
# data[0,0] = 13.606 (the geoid undulation actually stored at that point)
center_lon = left + 0.5 * res_x   # -180.0
center_lat = top  - 0.5 * res_y   #  90.0

N = _interp_geoid_point(center_lon, center_lat, data, left, top,
                        res_x, res_y, h, w)
print(N, "vs stored", data[0, 0])
# 13.5497 vs stored 13.6062 -- a half-pixel blend with the next column
```

Cross-checked against pyproj's `Transformer.from_crs('EPSG:4979', 'EPSG:5773')`:

```
N at (-180, 90)  : pyproj 13.6062,  ours 13.5497  (diff 5.6 cm)
N at ( -74, 40.7): pyproj -32.7736, ours -32.8596 (diff 8.6 cm)
```

## Root cause

`_interp_geoid_point` computes

```python
col_f = (lon - left) / res_x          # position in pixel-edge units
row_f = (top - lat) / res_y
```

[...]

PR #2629, opened 2026-05-29
reproject: fix cupy backend divergence from numpy on pyproj-fallback CRS pairs (#2620)

### LINKED ISSUE #2620: reproject cupy backend diverges from numpy on pyproj-fallback CRS pairs
## Describe the bug

The cupy backend of `reproject()` returns numerically different results from the numpy backend on identical valid input. The divergence comes from `_resample_cupy` in `xrspatial/reproject/_interpolate.py`, which resamples with `cupyx.scipy.ndimage.map_coordinates`. The other resamplers in that file (`_resample_numpy` via numba, and `_resample_cupy_native` via hand-written CUDA kernels) use a GDAL-matching renormalization scheme that `_resample_cupy` does not reproduce.

This path is reachable through the public API. `_reproject_chunk_cupy` only uses the native CUDA kernels when `try_cuda_transform` returns a result. For CRS pairs without a CUDA kernel (for example a projected-to-projected reprojection such as EPSG:32633 to EPSG:3857, where neither side is geographic WGS84/NAD83), it falls back to pyproj for coordinates and then to `_resample_cupy` for the resampling. In that case the numpy backend (Catmull-Rom cubic, renormalizing bilinear/nearest) and the cupy backend (map_coordinates) produce different numbers.

Two distinct facets:

1. Edge-band zero bleed (all resampling modes). `_resample_cupy` calls map_coordinates with `mode='constant', cval=0.0`. Sample points in the half-pixel boundary band (-1 < r < 0 and sh-1 < r < sh) are not flagged out-of-bounds, so the constant 0.0 bleeds into the result instead of the GDAL-style weight renormalization the numpy and native kernels apply. Per-pixel errors of hundreds of units show up along every chunk edge.

2. Cubic interpolant mismatch (everywhere). map_coordinates order=3 is a prefiltered B-spline, while the numpy and native kernels use Catmull-Rom. These are different interpolants, so they diverge across the whole array, not just at boundaries.

## Reproduce

Direct resampler comparison on a 5x5 source, querying a boundary point (r=4.6, c=2.0):
- `_resample_numpy` bilinear -> 23.0
- `_resample_cupy_native` bilinear -> 23.0
- `_resample_cupy` bilinear -> 0.0

End-to-end through `reproject()` with a UTM-33N source (EPSG:32633) reprojected to Web Mercator (EPSG:3857), comparing numpy vs cupy backends: max per-pixel difference ~534 for nearest/bilinear/cubic, mean ~14.

## Expected behavior

All four backends (numpy, cupy, dask+numpy, dask+cupy) should produce the same result on the same valid input, within floating-point tolerance. The existing four-backend parity tests assert this for the covered CRS pairs; the gap is the pyproj-fallback cupy path.

## Proposed fix

Route the cupy backend through the existing `_resample_cupy_native` CUDA kernels (which already match the numpy kernels exactly) instead of `map_coordinates`. The native kernels accept CPU coordinate arrays and transfer them, so they can serve the pyproj-fallback case too.

## Environment
- xarray-spatial: dev
- CUDA available, cupy 13.6.0

brendancol:
Closes #2620

## What changed

- `_reproject_chunk_cupy` and the eager `_reproject_dask_cupy` fast path now always resample through `_resample_cupy_native` (the hand-written CUDA kernels that match the numpy numba kernels exactly), instead of dropping to `cupyx.scipy.ndimage.map_coordinates` when the CUDA coordinate transform is unavailable.
- The native kernels accept CPU coordinate arrays and transfer them, so they cover the pyproj-fallback case too.
- Dropped the now-unused `_resample_cupy` import in `reproject/__init__.py`. The function itself stays in `_interpolate.py` (still referenced by existing unit tests).

## Why

On a projected-to-projected reprojection (e.g. EPSG:32633 -> EPSG:3857) neither CRS is geographic WGS84/NAD83, so no CUDA kernel applies and the cupy backend used `map_coordinates`. That path diverged from numpy two ways: it bled the `cval=0.0` constant into the half-pixel boundary band instead of renormalizing (per-edge-pixel errors ~534), and its order=3 cubic is a B-spline rather than the Catmull-Rom the numpy/native kernels use (~0.45 interior). After the change, numpy and cupy agree to ~3e-13.

## Backend coverage

- numpy: unchanged
- cupy: now matches numpy (nearest/bilinear/cubic)
- dask+numpy: unchanged
- dask+cupy: eager fast path now matches numpy

## Test plan

- [x] `TestCupyPyprojFallbackParity`: boundary-band renormalization + numpy/cupy/dask+cupy agreement for the three resampling modes on a projected-to-projected reprojection
- [x] Full `test_reproject.py` suite (385 passed)
- [x] `test_reproject_coverage_*`, `test_reproject_cupy_gate_2564`, `test_interpolation` (53 passed)


brendancol (commented):
## PR Review: reproject cupy backend divergence on pyproj-fallback CRS pairs (#2620)

### Blockers
None.

### Suggestions
None blocking. One thing worth noting for the record: this PR fixes the resampler divergence, but the prior inspection note (#2508) also flagged a separate deferred item. `_reproject_dask(is_cupy=True)` returns a dask array with cupy chunks but numpy meta, which could make `_apply_vertical_shift_dask` run the numpy shift on cupy blocks. That path is untouched here and stays deferred. It is out of scope for this PR; flagging only so it is not lost.

### Nits
- `_resample_cupy` is now unused by the reproject pipeline but stays in `_interpolate.py` because `TestCuPyResamplerClipping` still calls it directly. That is fine, but its docstring still says "CuPy equivalent of `_resample_numpy`", which is now misleading since the pipeline no longer treats it as equivalent. A one-line docstring note that it is kept only for the clipping unit tests would stop a future reader from wiring it back in.

### What looks good
- Root cause is right and the fix is the minimal one. The native CUDA kernels already match the numpy numba kernels (existing `test_cubic_nan_fallback_matches_cpu` proves it), so routing all cupy resampling through them removes the divergence without new kernel code.
- Both affected call sites are covered: the per-chunk `_reproject_chunk_cupy` (2D and the 3D band loop) and the eager `_reproject_dask_cupy` fast path.
- `_resample_cupy_native` accepts numpy coordinate arrays and transfers them, so the pyproj-fallback case works without forcing coordinates onto the GPU first.
- The double nodata-to-NaN conversion in the dask path (window converted at the call site, then again inside the native resampler) is harmless. NaN stays NaN.
- Tests assert both the finite/nodata mask equality and the value agreement, across nearest/bilinear/cubic plus the dask+cupy path. The standalone boundary-band test pins the specific failure mode (zero bleed at r=4.6 and r=-0.4).

### Checklist
- [x] Algorithm matches reference (native kernels match numpy numba kernels, verified by existing and new tests)
- [x] All implemented backends produce consistent results (numpy/cupy/dask+cupy now agree to ~3e-13)
- [x] NaN handling is correct (mask equality asserted; native kernel does its own nodata-to-NaN)
- [x] Edge cases covered (boundary band, descending y, three resampling modes)
- [x] Dask chunk boundaries handled (dask+cupy parity test passes with 32x32 chunks)
- [x] Benchmark not needed (no perf-critical change; native kernel was already the GPU fast path)
- [x] README feature matrix not applicable (no new function, backend support unchanged)
- [x] Docstrings/comments present and accurate (inline comments explain the routing choice)

brendancol (commented):
## Follow-up review (after 5311c66)

The one nit from the first pass is resolved: `_resample_cupy`'s docstring now states it is not numerically equivalent to `_resample_numpy` (B-spline cubic, cval boundary bleed) and that it is kept only for the clipping unit tests, with a pointer to #2620. The `TestCuPyResamplerClipping` tests still pass since the change is docstring-only.

The Suggestion item (the `_reproject_dask(is_cupy=True)` meta vs cupy-chunks mismatch noted in #2508) stays deferred. It is a distinct defect in a different code path and out of scope here.

No new findings. No blockers.

PR #2630, opened 2026-05-29
polygonize: derive transform from x/y coords so geopandas CRS matches geometry (#2607)

### LINKED ISSUE #2607: polygonize: derive transform from x/y coords so geopandas CRS does not lie about pixel-space geometries
**Describe the bug**

`polygonize()` auto-detects an affine transform from `attrs['transform']` and `rio.transform()` (added in #2536 so a `return_type='geopandas'` result would not carry a CRS while its geometries stayed in pixel space). It never derives a transform from the raster's own x/y coordinate values.

In xarray/xrspatial, the standard way to georeference a raster is through its x/y coordinate values; `calc_res` and `get_dataarray_resolution` already read resolution from there. A raster georeferenced through coords alone (no `attrs['transform']`, no rioxarray) plus `attrs['crs']` hits the same mismatch #2536 fixed, just through a different and more common channel:

- `return_type='geopandas'`: the GeoDataFrame's `.crs` is set (e.g. EPSG:3857) but the geometries sit in pixel-index space (0..N). The metadata disagrees with the coordinates.
- `return_type='numpy' / 'awkward' / 'geojson' / 'spatialpandas'`: coordinates come out in pixel space even though the input had georeferenced coords.

**Reproduction**

```python

data = np.array([[1, 1, 2], [1, 1, 2], [2, 2, 2]], dtype=np.int32)
raster = xr.DataArray(
    data, dims=('y', 'x'),
    coords={'y': [5_000_000.0, 4_999_990.0, 4_999_980.0],
            'x': [1_000_000.0, 1_000_010.0, 1_000_020.0]},
    attrs={'crs': 'EPSG:3857'},
)

col, polys = polygonize(raster, return_type='numpy')
# output x/y are 0..3 (pixel space) instead of ~1e6 (the coords)

df = polygonize(raster, return_type='geopandas')
# df.crs == EPSG:3857 but df.geometry.total_bounds == [0, 0, 3, 3]
```

**Expected behavior**

When no explicit `transform=`, `attrs['transform']`, or `rio.transform()` is available, `polygonize()` should derive the transform from the raster's x/y coords (when they are 1-D, length >= 2, and evenly spaced) so output coordinates land in CRS space. Irregular or too-short coords should fall back to pixel space rather than emit a wrong affine. The `_xrspatial_no_georef` marker should keep suppressing auto-detection, and explicit `transform=` / `attrs['transform']` should keep taking precedence.

**Affected backends**

All four (numpy, cupy, dask+numpy, dask+cupy). Transform detection runs in the public `polygonize()` before backend dispatch, so they all behave the same way.

**Additional context**

Found by the metadata-propagation sweep. Same class of silent CRS/coordinate misalignment as #2536.

brendancol:
## Summary

`polygonize()` auto-detected an affine transform from `attrs['transform']` and `rio.transform()` (added in #2536) but never from the raster's own x/y coordinate values, which is the usual way an xarray raster carries its georeferencing. A raster georeferenced through coords alone, plus an `attrs['crs']`, produced a `return_type='geopandas'` result whose `.crs` claimed projected space while the geometries stayed in pixel space. It's the same CRS/coordinate mismatch as #2536, just reached through coords instead of the transform attr.

- `_detect_raster_transform()` now falls back to deriving a rasterio-ordered transform from evenly-spaced 1-D x/y coords. The origin sits at the first pixel corner (`first_center - 0.5 * spacing`) so the pixel-corner indices from `_follow` map to cell corners.
- The `rio.transform()` identity-affine case now falls through to coords instead of returning None, so coords get used when rioxarray does not recognize the dims as spatial (for example `lat`/`lon`).
- Precedence is unchanged: explicit `transform=`, then `attrs['transform']`, then `rio.transform()`, then coords. The `_xrspatial_no_georef` marker still suppresses auto-detection.

## Backend coverage

Transform detection runs in the public `polygonize()` before backend dispatch, so all four backends (numpy, cupy, dask+numpy, dask+cupy) behave the same way. Tests cover all four.

## Test plan

- [x] Coords-derived transform applied for numpy and geopandas output
- [x] geopandas `.crs` agrees with geometry coordinates
- [x] `attrs['transform']` and explicit `transform=` still take precedence
- [x] Irregular / length-1 / no coords fall back safely
- [x] Non-`y`/`x` dim names (`lat`/`lon`) georeferenced from coords
- [x] `_xrspatial_no_georef` marker suppresses
- [x] numpy, cupy, dask+numpy, dask+cupy
- [x] Full existing polygonize suite passes (172 passed, 13 skipped)

Closes #2607

### DIFF
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -2298,11 +2298,11 @@ def polygonize(
     georeferencing convention; used when the coords are 1-D, length >= 2
     and evenly spaced).  An explicit ``transform=`` argument always
     overrides the auto-detected value.  Auto-detection is skipped when
-    the raster carries ``attrs['_xrspatial_no_georef']=True``.  This applies to all return
-    types -- the geometries themselves are transformed, so the
-    coordinates emitted in the "numpy", "awkward", "spatialpandas" and
-    "geojson" outputs are also in CRS coordinate space, not pixel
-    space, when the raster carries a transform.
+    the raster carries ``attrs['_xrspatial_no_georef']=True``.  This
+    applies to all return types -- the geometries themselves are
+    transformed, so the coordinates emitted in the "numpy", "awkward",
+    "spatialpandas" and "geojson" outputs are also in CRS coordinate
+    space, not pixel space, when the raster carries a transform.
     """
     _validate_raster(raster, func_name='polygonize', name='raster', ndim=2)
     if raster.shape[0] < 1 or raster.shape[1] < 1:
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -776,12 +776,20 @@ def _detect_raster_transform(raster: xr.DataArray):
     2. ``raster.rio.transform()`` (rioxarray, if installed). Returns an
        ``affine.Affine``; iterating it yields the rasterio-ordered 6
        coefficients ``(a, b, c, d, e, f)``.
-    3. ``None``.
+    3. The raster's own x/y coordinate values (the xarray /
+       xrspatial standard georeferencing convention -- the same one
+       ``get_dataarray_resolution`` / ``calc_res`` read).  This is the
+       common case: a raster loaded with georeferenced coords and a
+       ``crs`` attr but no separate ``transform`` attr would otherwise
+       emit pixel-space geometries while ``return_type='geopandas'``
+       attached a CRS, the exact mismatch #2536 set out to prevent.
+    4. ``None``.
 
     A raster carrying ``attrs['_xrspatial_no_georef']=True`` (the
     xrspatial.geotiff "no georeference" marker) is treated as having
-    no transform even if ``rio.transform()`` is present, because that
-    marker explicitly opts out of georeferencing.
+    no transform even if ``rio.transform()`` or georeferenced coords
+    are present, because that marker explicitly opts out of
+    georeferencing.
     """
     # Honour the xrspatial.geotiff "no georeference" marker if set.
     if raster.attrs.get('_xrspatial_no_georef'):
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: polygonize derive transform from x/y coords

None.

- The identity-affine fall-through changes behavior for rasters with integer-index coords that used to get pixel-space output. With rioxarray installed, a raster carrying `coords={'y': np.arange(3), 'x': np.arange(3)}` and no transform now resolves to a transform with origin -0.5 (the pixel-center to corner convention), so output spans -0.5..2.5 instead of 0..3. That is geometrically correct, but the existing `test_no_transform_info_yields_pixel_coords` (xrspatial/tests/test_polygonize.py:1458) now has a misleading name and docstring: arange coords are transform info after this change, and the test only passes because its assertion uses `<= 3.0` with slack. Worth a one-line note in that test's docstring, or tighten its bounds to document the new -0.5 offset.

- The `polygonize()` docstring line at xrspatial/polygonize.py:2301 wraps mid-sentence ("This applies to all return") at 88 chars. The wrapping is pre-existing and just shifted by the edit, so it is not a new defect, but it reads awkwardly.

### What looks good
- The corner-origin math is right: pixel centers map to cell corners via `first_center - 0.5*spacing`, matching the existing `attrs['transform']` convention.
- The `_coord_spacing` guard rejects irregular spacing, length-1 dims, and zero-step coords, returning None instead of emitting a wrong affine from the first two points.
- Precedence is preserved and tested: explicit `transform=`, then `attrs['transform']`, then `rio.transform()`, then coords.
- Uses `dims[-1]`/`dims[-2]` for x/y instead of hardcoding `'x'`/`'y'`, matching `get_dataarray_resolution`; a lat/lon test covers it.
- All four backends are tested (numpy, cupy, dask+numpy, dask+cupy). Detection runs before dispatch, so parity is structural.
- Descending y coords (negative spacing) are handled correctly.

### Checklist
- [x] Algorithm matches reference/convention (corner-origin transform)
- [x] NaN handling unchanged (not touched by this change)
- [x] Edge cases covered by tests (irregular, length-1, no coords, no-georef marker)
- [x] Dask chunk boundaries handled correctly (detection is pre-dispatch)
- [x] Benchmark not needed (detection is O(n) over a 1-D coord array)

brendancol (commented):

Both findings from the previous review are addressed.

- Suggestion (test name/docstring): `test_no_transform_info_yields_pixel_coords` now builds a coordless raster instead of one with `np.arange(3)` coords, so it actually exercises the no-transform path. The docstring explains why, since arange coords are a transform source after this change.
- Nit (docstring wrap): the `polygonize()` docstring paragraph is reflowed; no more mid-sentence break.

Full polygonize suite passes locally: 172 passed, 13 skipped (the 13 skips are optional-dependency / GPU-gated cases not present in CI's matrix slot). No remaining open items.

Issue #2623, opened 2026-05-29
polygonize: float-dtype mask raises TypeError on float rasters


brendancol:
The `polygonize()` docstring says the `mask` parameter takes bool, integer, or float values ("Pixels to include should have mask values of 1 or True, pixels to exclude should have 0 or False"). Integer masks work on every raster dtype, but a float-dtype mask on a float-dtype raster crashes with a TypeError instead of polygonizing.

## Reproduce

```python
import numpy as np
import xarray as xr
from xrspatial.polygonize import polygonize

raster = xr.DataArray(np.array([[1.0, 1.0, 2.0],
                                [1.0, 3.0, 2.0],
                                [1.0, 1.0, 2.0]], dtype=np.float64))
mask = xr.DataArray(np.array([[1, 1, 1],
                              [1, 0, 1],
                              [1, 1, 1]], dtype=np.float64))
polygonize(raster, mask=mask)
```

```
TypeError: ufunc 'bitwise_and' not supported for the input types, and the
inputs could not be safely coerced to any supported types according to the
casting rule 'safe'
```

## Cause

In `_polygonize_numpy` (`xrspatial/polygonize.py:918`) the float-raster branch runs `mask = mask & nan_mask`. `&` (bitwise_and) is undefined between a float array and the boolean `nan_mask`, so it raises. A bool or int mask coerces fine; a float one does not. The cupy and dask backends send float rasters through `_polygonize_numpy`, so they hit the same line.

A float mask on an integer raster is fine, because that branch never runs the `mask & nan_mask` line. Only the float-raster + float-mask combination breaks.

## Possible fix

Cast the mask to bool before the `&`, e.g. `mask = mask.astype(bool) & nan_mask`. That matches the truthiness already used in `_calculate_regions` (`mask[ij]`).

Found during the deep-sweep test-coverage sweep. The companion test PR pins the integer-mask paths (which work today) and xfails the float-mask case against this issue.

PR #2631, opened 2026-05-29
Add polygonize mask-dtype coverage; pin float-mask bug (#2623)

brendancol:
Adds test coverage for the `mask=` parameter of `polygonize`, which the
docstring documents as accepting bool, integer, or float values. Every
pre-existing test passed only a bool mask, so the integer- and float-mask
paths were untested.

## What this adds

- `test_polygonize_mask_dtype_coverage_2026_05_29.py`:
  - Integer masks (int32 / int64) pinned against the equivalent bool mask
    on numpy, cupy, dask+numpy, and dask+cupy, for both raster dtypes and
    connectivity 4 and 8. Each backend is compared to its own bool-mask
    output so the only variable is the mask dtype.
  - Float mask on an integer raster pinned (works today).
  - A sanity anchor confirming a non-bool mask with a 0 cell actually
    drops that pixel.

## Source bug found

A float-dtype mask on a float-dtype raster raises `TypeError` at
`polygonize.py:918` (`mask = mask & nan_mask` -- bitwise_and is undefined
between float and bool arrays). cupy and dask route float rasters through
the numpy path, so they crash the same way. Integer masks coerce fine.

Filed as #2623. The eight float-mask-on-float-raster cases are marked
`xfail(strict, raises=TypeError)` so they flip to passing once the source
casts the mask to bool. No source changed in this PR.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy -- all exercised (run on a CUDA host:
41 passed, 8 xfailed).

## Test plan

- [x] `pytest xrspatial/tests/test_polygonize_mask_dtype_coverage_2026_05_29.py` -> 41 passed, 8 xfailed
- [x] `pytest xrspatial/tests/test_polygonize.py` still green
- [x] Mutation check: dropping the mask exclusion in `_calculate_regions` flips 11 tests red
- [x] Source file md5 unchanged after mutation restore


brendancol (commented):
## PR Review: Add polygonize mask-dtype coverage; pin float-mask bug (#2623)

Test-only PR. One new test module plus a state-file update. No source changes, which matches the stated intent.

None.

- `_result_signature` (test file, ~line 60) reduces each polygon to (value, per-ring vertex counts). That catches a mask change that alters polygon count or ring structure, but it would not catch a mask change that swapped which pixels are included while keeping the same vertex-count shape (e.g. a mirrored mask). The `test_nonbool_mask_actually_excludes_pixels` anchor covers the "mask is honored at all" case, and the mutation check confirms the tests have teeth, so this is a soft spot rather than a hole. Comparing exact coordinates for at least the numpy case would tighten it.

- Float mask dtypes only cover float64. A float32 mask hits the same `mask & nan_mask` line, so it would land in the same #2623 xfail bucket; adding it is cheap but not essential.

### What looks good
- All four backends (numpy, cupy, dask+numpy, dask+cupy) are parametrized with the project's skip decorators, and the run on a CUDA host reports 41 passed / 8 xfailed.
- Each backend is compared against its own bool-mask output, which isolates the mask-dtype variable from the unrelated numpy-vs-dask polygon-representation difference. The module docstring explains this.
- The float-mask-on-float-raster crash is pinned with `xfail(strict, raises=TypeError)` referencing #2623, so the test flips to passing automatically once the source is fixed.
- The source bug was filed separately (#2623) rather than bundled into the test PR.

### Checklist
- Algorithm matches reference/paper: N/A (test-only)
- All implemented backends produce consistent results: yes, asserted
- NaN handling correct: the float-mask path interacts with the NaN mask; the crash is pinned
- Edge cases covered: connectivity 4/8, both raster dtypes, int/float masks
- Dask chunk boundaries handled: yes, (2, 2) chunks cross the mask
- README updated: not needed (no new function)
- Docstrings present and accurate: module docstring explains intent

brendancol (commented):
## Follow-up review (after e5cfa53)

- Suggestion (signature comparison could miss a same-shape pixel swap): added `test_integer_mask_exact_coords_numpy`, which compares full polygon coordinate arrays on the numpy backend where region order is deterministic. The signature comparison stays for the cross-backend cases where exact ordering is not guaranteed.
- Nit (float32 mask): the float-mask tests now parametrize over float32 and float64. float32 lands in the same #2623 xfail bucket on float rasters and passes on integer rasters.

Local run: 53 passed, 16 xfailed; flake8 clean. No remaining open items.

PR #2633, opened 2026-05-29
polygonize: keep diagonal-notch hole in dask 8-conn merge (#2606)

### LINKED ISSUE #2606: polygonize dask connectivity=8 over-fills diagonal notches at chunk boundaries
**Describe the bug**

The dask backend of `polygonize` returns wrong results for `connectivity=8` when two same-value regions touch only diagonally across a chunk boundary. The cross-chunk merge fills in the diagonal notch, so the merged polygon covers more pixels than it should. The total polygon area comes out larger than the raster, and the merged ring self-intersects (invalid per shapely).

numpy is fine. dask with `connectivity=4` is fine. Only dask `connectivity=8` is wrong, and only at chunk corners where the diagonal adjacency crosses a chunk boundary.

**Reproduce**

```python
import dask.array as da

data = np.array([[1,1,1,0],
                 [1,0,1,0],
                 [0,1,0,0],
                 [0,1,0,1]], dtype=np.int32)

rn = xr.DataArray(data.copy())
rd = xr.DataArray(da.from_array(data.copy(), chunks=(2,4)))

vn, pn = polygonize(rn, connectivity=8)
vd, pd = polygonize(rd, connectivity=8)

area = lambda v, p: sum(Polygon(r[0], r[1:]).area for r in p)
print(area(vn, pn))  # 16.0  (correct, equals raster area)
print(area(vd, pd))  # 17.0  (wrong, larger than raster)
```

**Expected behavior**

dask `connectivity=8` should match numpy: same per-value area sums, valid (non-self-intersecting) polygons, total area equal to the raster area.

**Root cause**

`_group_boundary_polygons` groups two diagonally adjacent same-value boundary polygons into one group under 8-connectivity. `_merge_polygon_rings` then edge-cancels and re-traces them into a single ring that fills the diagonal corner. numpy 8-connectivity keeps the notch, because the different-value cells on the opposite diagonal are also 8-connected and split the region the other way.

**Additional context**

Found by the deep-sweep accuracy audit (Cat 5, backend inconsistency). 4 of 300 random integer rasters with random chunkings showed an area mismatch for `connectivity=8`; 0 mismatches for `connectivity=4`.

brendancol:
Closes #2606

The dask `connectivity=8` cross-chunk merge filled in the diagonal notch where two same-value regions meet only at a corner across a chunk boundary. The merged polygon then covered one extra cell, so total polygon area came out larger than the raster.

- `_merge_polygon_rings` traced the notch as a separate negative (hole) ring, but `_group_rings_into_polygons` dropped it: the containment test used the hole's first vertex, which sits on the exterior boundary at the pinch point, so `_point_in_ring` returned False.
- Added `_ring_interior_point()` and used a strictly-interior point of the hole for the containment test. Works for the non-convex notch rings the merge produces.

After the fix, numpy and dask 8-conn report equal per-value areas. 4-conn was already correct.

Backend coverage: numpy (unaffected), dask+numpy (fixed), dask+cupy (fixed). The shared merge path is CPU-side, so the cupy and dask+cupy routes go through the same code.

Test plan:
- [x] `test_polygonize_issue_2606.py`: reproducer total-area and per-value-area parity, 20 random rasters/chunkings, dask+cupy parity
- [x] Full `test_polygonize*` suite green (291 passed)
- [x] cupy-marked polygonize tests green on a CUDA host (843 passed)
- [x] 0 area mismatches across 400 random integer rasters for both connectivities (was nonzero for 8-conn before)


brendancol (commented):
## PR Review: polygonize: keep diagonal-notch hole in dask 8-conn merge (#2606)

### Blockers
None.

### Suggestions
None.

### Nits
- xrspatial/polygonize.py:1503 — `_ring_interior_point`'s centroid fallback divides by `n`. Every ring that reaches `_group_rings_into_polygons` is closed with at least 3 unique vertices, so `n >= 3` and the divide is safe. A one-line note that the fallback assumes a non-empty ring would be nice, but it's not required.

### What looks good
- Root cause is pinned correctly. The hole got dropped because containment was tested with the hole's first vertex, which lands on the exterior at the pinch point. Switching to a strictly-interior point is the right fix and the minimal one.
- `_ring_interior_point` handles non-convex notch rings, not just convex ones, by nudging inward along the orientation-aware edge normal. I checked it on an L-shaped ring and on the CW hole ring.
- The `eps=1e-6` inward step is safe. `_group_rings_into_polygons` runs on integer pixel coordinates (the transform is applied later, in `_merge_chunk_polygons` / `_merge_from_separated`), so the offset stays well below the unit grid spacing.
- Area parity now holds for numpy, dask+numpy, and dask+cupy. The 367 cases where dask geometry is "invalid" per shapely match numpy exactly. 8-connectivity self-touching polygons are documented and expected, and the #2172 figure-8 tests still pass.
- Tests cover the reproducer, 20 random rasters with ragged chunks, and the dask+cupy path.

### Checklist
- Algorithm matches the numpy reference: yes (area parity)
- All implemented backends consistent: yes (numpy, dask+numpy, dask+cupy)
- NaN handling: unaffected by this change
- Edge cases covered: random ragged chunks, reproducer
- Dask chunk boundaries: this is the fix
- Benchmark: not needed (bug fix, no new function)
- README matrix: not applicable
- Docstrings present: yes, `_ring_interior_point` documented

brendancol (commented):
Follow-up: addressed the one nit from the previous review (commit 23aef0cf) — added a comment noting the centroid fallback assumes a non-empty ring (n >= 3). No code-path change; the 23 regression tests still pass. No remaining findings.

PR #2642, opened 2026-05-29
Raise clear ValueError for empty Dataset in stats()

### LINKED ISSUE #2637: stats() on an empty Dataset raises an opaque IndexError
**Describe the bug**

Calling `stats()` on an empty xarray Dataset (no data variables) raises an opaque `IndexError` instead of a clear error.

In the Dataset branch of `stats()` (xrspatial/zonal.py around line 871), the code builds a per-variable DataFrame list `dfs`, then does `result = dfs[0]`. If the Dataset has no data variables, `dfs` is empty and `dfs[0]` raises `IndexError: list index out of range`. The caller gets nothing useful about what actually went wrong.

**Expected behavior**

`stats()` should catch the empty-Dataset case up front and raise a `ValueError` saying the Dataset has no data variables to compute statistics over.

**Steps to reproduce**

    import xarray as xr
    import numpy as np

    zones = xr.DataArray(np.zeros((4, 4), dtype=int))
    empty = xr.Dataset()
    stats(zones=zones, values=empty)  # raises IndexError instead of ValueError

**Additional context**

The fix is to check for empty `data_vars` before the `dfs[0]` access and raise a clear `ValueError`. Needs a test that passes an empty Dataset and asserts `ValueError`.

brendancol:
Closes #2637

`stats()` accepts an xarray Dataset and runs per-variable, then merges the results. If the Dataset has no data variables, the merge hit `result = dfs[0]` on an empty list and raised an opaque `IndexError` that told the caller nothing.

This adds an early check in the Dataset branch: when `values.data_vars` is empty, raise a `ValueError` saying there is nothing to compute statistics over, before reaching the `dfs[0]` access.

- Guard the empty-Dataset case in the `stats()` Dataset branch with a clear `ValueError`.
- Add a regression test that passes an empty Dataset and asserts `ValueError`.

Backend coverage: the Dataset branch is shared by all backends and the guard runs before any backend dispatch, so numpy / cupy / dask+numpy / dask+cupy are all covered.

Test plan:
- [x] New test `test_stats_empty_dataset_raises_value_error_2637` passes
- [x] Existing Dataset return_type test still passes


brendancol (commented):
## PR Review: Raise clear ValueError for empty Dataset in stats()

None.

None.

- xrspatial/zonal.py:861 -- `len(values.data_vars) == 0` works and is clear. `if not values.data_vars:` would be slightly more idiomatic since the data_vars mapping is falsy when empty, but the explicit length check reads fine and matches the surrounding style. Take it or leave it.

### What looks good
- The guard is in the right place: after the return_type check and before the `dfs[0]` access, so the empty case is caught early with a message that names the actual problem (no data variables).
- The error message tells the caller what to do (pass a Dataset with at least one data variable), not just what went wrong.
- The regression test reuses the existing `small_zones_values_2558` fixture and asserts on the "no data variables" text, so the old IndexError would not satisfy it.
- Scope is tight: two files, no unrelated edits.

### Checklist
- Algorithm matches reference/paper: n/a (input validation fix)
- All implemented backends produce consistent results: yes, guard runs before backend dispatch
- Edge cases are covered by tests: yes, empty Dataset is the case in question
- No premature materialization or unnecessary copies: yes, the data_vars length check is cheap
- README feature matrix updated: not needed, no new public API
- Docstrings present and accurate: stats() docstring unchanged, still accurate

PR #2645, opened 2026-05-29
crop: validate zones/values shapes match instead of returning nonsense

### LINKED ISSUE #2638: crop() can silently return a nonsense result when zones and values shapes differ
**Describe the bug**

`crop()` in `xrspatial/zonal.py` can silently return a nonsensical result when `zones` and `values` have different shapes. The function checks that each input is 2D via `_validate_raster(...)`, but never calls `validate_arrays(zones, values)`. Pass a 3x3 zones raster and a 2x2 values raster and you get a tiny crop back instead of an error.

`stats()` and `crosstab()` both call `validate_arrays(zones, values)` to catch shape and backend mismatches. `crop()` doesn't.

**Expected behavior**

`crop()` should raise when `zones` and `values` have incompatible shapes or backends, the same way `stats()` and `crosstab()` do.

**Reproduce**

```python
from xrspatial.zonal import crop
zones = xr.DataArray(np.array([[1,1,2],[1,1,2],[3,3,3]]), dims=['y','x'])
values = xr.DataArray(np.array([[10,20],[30,40]]), dims=['y','x'])
crop(zones, values, zone_ids=(1,))  # returns a (2,2) crop; should raise
```

**Additional context**

Fix: add `validate_arrays(zones, values)` after the existing 2D checks, plus a regression test for the mismatched-shape case.

brendancol:
Closes #2638

`crop()` checked that `zones` and `values` were each 2D but never compared their shapes to each other, so a shape mismatch (e.g. 3x3 zones, 2x2 values) silently produced a meaningless crop instead of raising.

- Add `validate_arrays(zones, values)` after the existing 2D checks, matching `stats()` and `crosstab()`. It catches both shape and backend mismatches.
- Add a regression test for the 3x3 zones / 2x2 values case.

Backend coverage: the check sits before backend dispatch and applies to numpy, cupy, dask+numpy, and dask+cupy alike. `validate_arrays` is the same guard the other zonal functions use across all four backends.

Test plan:
- [x] `pytest xrspatial/tests/test_zonal.py -k crop` (21 passed)
- [x] new test fails without the fix, passes with it


brendancol (commented):
## PR Review: crop validates zones/values shapes match

None.

One behavior to confirm: `validate_arrays()` runs after `_maybe_rasterize_zones()`, so a GeoDataFrame or list-of-pairs zones gets rasterized against values first and the shapes line up. That rasterized zones is numpy-backed, though. If someone passes dask-backed values with a GeoDataFrame zones, `validate_arrays()` will now flag a backend mismatch (numpy zones vs dask values) and raise. That matches `stats()`/`crosstab()`, so it's defensible, but it's a slightly stricter contract for that one path than before. The existing GDF tests use numpy values, so they still pass. Flagging it, not blocking.

The `crop()` docstring doesn't mention that mismatched shapes now raise. Minor: the error message is self-explanatory, and `stats()`/`crosstab()` don't document it either.

### What looks good
- One-line guard placed where the other zonal functions put it, after the 2D checks and after rasterization. Order is right: non-2D inputs still hit the clearer `_validate_raster` ndim error first.
- `validate_arrays` is already imported, no new import needed.
- The regression test reproduces the reported 3x3 vs 2x2 case and asserts on the error message.
- Full `test_zonal.py` passes (170), including the GDF and dask crop paths.

### Checklist
- Algorithm matches reference: n/a (validation fix)
- All implemented backends consistent: yes, check runs before backend dispatch
- NaN handling correct: n/a
- Edge cases covered by tests: yes, mismatched-shape case added
- Dask chunk boundaries handled: n/a, validate_arrays rechunks to match
- Docstrings present and accurate: see nit above

brendancol (commented):

- Nit (docstring didn't mention the new raise): fixed in 57bd0c04. The crop() Notes section now states that zones and values must share shape and backend or a ValueError is raised, cross-referencing stats() and crosstab().
- Suggestion (dask values + GeoDataFrame zones now raises a backend mismatch): confirmed intended. The stricter contract matches stats()/crosstab(), and the rasterize path produces numpy-backed zones, so mixing it with dask values legitimately fails the backend check. No code change.

### Status
No remaining blockers, suggestions, or nits. crop tests pass (21) and full test_zonal.py passes (170).

PR #2648, opened 2026-05-29
Validate backend compatibility for 3D apply() inputs

### LINKED ISSUE #2639: zonal.apply() skips backend validation for 3D values, letting mixed dask/numpy inputs through
**Describe the bug**

`zonal.apply()` checks the shape of `zones` against `values` with `zones.shape != values.shape[:2]`, then dispatches to a backend based on the type of `values`. It only calls `validate_arrays(zones, values)` for the 2D case. For 3D `values`, backend compatibility between `zones` and `values` is never checked.

Mixed dask/numpy inputs slip through:

- `zones` is numpy and `values` is dask: the dask backend runs and tries `zones_data.chunks`, which raises `AttributeError: 'numpy.ndarray' object has no attribute 'chunks'`.
- `zones` is dask and `values` is numpy: the numpy backend runs and silently returns eager numpy output, ignoring that `zones` was chunked.

Neither outcome is a clear, early error.

**Expected behavior**

Mixed-backend `zones` and `values` should fail early with a clear error, the same way `validate_arrays` reports a backend mismatch elsewhere in the module. The 2D path already does this; the 3D path should too.

**Additional context**

The 2D path calls `validate_arrays(zones, values)`, which checks both shape and backend. The 3D path skips it because `validate_arrays` requires equal full shapes, and a 2D `zones` cannot equal a 3D `values`. The fix is to validate backend compatibility for the 3D case without requiring equal full shapes.

brendancol:
Closes #2639

`zonal.apply()` only called `validate_arrays(zones, values)` for 2D values. For 3D values it checked shape but never backend, so mixed dask/numpy inputs slipped through:

- numpy `zones` + dask `values`: the dask backend crashed with `AttributeError: 'numpy.ndarray' object has no attribute 'chunks'`.
- dask `zones` + numpy `values`: the numpy backend silently returned eager numpy output.

This adds an explicit backend check for the 3D path. `validate_arrays` can't be reused here because it requires equal full shapes (a 2D `zones` never equals a 3D `values`), so the check compares `_classify_backend` for each input and raises the same backend-mismatch error on a mismatch.

Backend coverage: no backend behavior changed for matched inputs. The change only adds an early error for mismatched-backend inputs across all four backends (numpy / cupy / dask+numpy / dask+cupy).

Test plan:
- [x] New `test_apply_3d_mixed_backend_raises` covers both mismatch directions and asserts a clear ValueError.
- [x] Verified the new test reproduces the original AttributeError without the fix.
- [x] `pytest xrspatial/tests/test_zonal.py -k apply` passes (14 passed).


brendancol (commented):
## PR Review: Validate backend compatibility for 3D apply() inputs

None.

None.

- `xrspatial/zonal.py:1929-1934`: the 3D backend check repeats the message string from `validate_arrays` (`utils.py:377-382`), but with "zones"/"values" labels instead of positional indices. That trade is fine here: the labels say more to the caller than "array 0"/"array 1", and `validate_arrays` can't be reused directly because it requires equal full shapes. A one-line note that the wording mirrors `validate_arrays` on purpose would keep the two greppable together. Not blocking.

### What looks good
- Minimal fix aimed at the exact gap: 3D values skipped backend validation while 2D went through `validate_arrays`.
- The comment explains the non-obvious part, that `validate_arrays` can't be reused because a 2D zones never equals a 3D values.
- Test covers both mismatch directions (numpy zones + dask values, dask zones + numpy values), and the PR verified it reproduces the original AttributeError without the fix.
- `_classify_backend` distinguishes cupy from numpy too, so the same check catches cupy/numpy mismatches for free.

### Checklist
- Algorithm matches reference/paper: N/A, validation fix
- Backends produce consistent results: unchanged for matched inputs; mismatched inputs now error early
- NaN handling: unchanged
- Edge cases covered by tests: both mismatch directions
- Dask chunk boundaries: 3D dask path already rechunks layers to zones.chunks internally
- No premature materialization or copies: none added
- Benchmark: not needed, validation-only
- README feature matrix: not applicable, no new function or backend change
- Docstrings: apply() docstring unchanged and still accurate

brendancol (commented):

Re-reviewed after commit 725dad92, which addresses the only finding from the first pass.

### Blockers
None.

### Suggestions
None.

### Nits
None remaining. The earlier nit (note that the backend error wording mirrors `validate_arrays`) is now addressed by the comment at `xrspatial/zonal.py:1933-1934`.

### Summary
The diff is the same focused validation fix plus a one-line comment. Tests still pass (14 apply tests). No new findings.

PR #2650, opened 2026-05-29
Remove dead no-op set_index in cupy zonal stats path

### LINKED ISSUE #2641: zonal.stats CuPy path has a dead no-op set_index("zone") call
## Describe the bug
In `xrspatial/zonal.py`, the CuPy stats path (`_stats_cupy`) builds the result DataFrame and then calls `stats_df.set_index("zone")` without assigning the return value or passing `inplace=True`. By default `set_index` returns a new DataFrame, so this call does nothing: the result is thrown away and the original `stats_df` is returned unchanged.

## Why it matters
The line is harmless at runtime, but it signals the path wasn't reviewed closely. It also raises a backend-parity question. The numpy path (`_stats_numpy`) and the dask path (`_stats_dask_numpy`) both return the stats DataFrame with `zone` as a plain column, not as the index. The CuPy path should match.

## Expected behavior
The CuPy stats DataFrame should match numpy and dask: `zone` as a regular column, not the index. Since the other backends keep `zone` as a column, the fix is to drop the dead line.

## Backend
CuPy / GPU only (`_stats_cupy`).

## Suggested fix
Remove the no-op `stats_df.set_index("zone")` line so the CuPy path matches the other backends.

brendancol:
Closes #2641

## What changed
- Removed the no-op `stats_df.set_index("zone")` line in `_stats_cupy` (`xrspatial/zonal.py`). The call discarded its return value, so it did nothing.
- Added a regression test asserting the cupy stats result keeps `zone` as a plain column with a default index, matching numpy.

## Why remove rather than apply the index
The numpy (`_stats_numpy`) and dask (`_stats_dask_numpy`) paths both return the stats DataFrame with `zone` as a regular column, not as the index. Removing the dead line keeps the cupy path's return shape consistent with the other backends. Turning it into `stats_df = stats_df.set_index("zone")` would have diverged from them.

## Backend coverage
CuPy-only code path. Behaviour for numpy / dask+numpy / dask+cupy is unchanged.

## Test plan
- [x] `pytest xrspatial/tests/test_zonal.py` (170 passed locally, GPU available, so the cupy path and the new test ran rather than skipping)


brendancol (commented):
## PR Review: Remove dead no-op set_index in cupy zonal stats path

None.

None.

- `xrspatial/tests/test_zonal.py`: the new test asserts both `df_cp.index.name is None` and `df_cp.index.name == df_np.index.name`. The second line covers the first once you trust numpy's index is unnamed, but keeping both is fine. One documents the expected value, the other pins parity. No change needed unless you want to trim.

### What looks good
- The fix is the right call. Both `_stats_numpy` and `_stats_dask_numpy` return the DataFrame with `zone` as a plain column, and `_stats_dask_cupy` delegates to `_stats_dask_numpy`, so dropping the discarded `set_index` call brings the cupy path in line instead of changing observable behaviour.
- The discarded `set_index("zone")` return value was a genuine no-op, so deleting it carries no runtime risk.
- The regression test checks column presence, an unnamed index, and column-order parity against numpy. It runs on GPU and skips cleanly without CUDA.
- Surgical diff: one deleted line plus one test.

### Checklist
- [x] Algorithm matches reference/paper (n/a, no algorithm change)
- [x] All implemented backends produce consistent results (cupy now matches numpy/dask)
- [x] Edge cases covered by tests (new parity test added)
- [x] Dask chunk boundaries handled correctly (n/a, unchanged)
- [x] Docstrings present and accurate (unchanged)

brendancol (commented):

The one nit from the first pass is addressed. The redundant `df_cp.index.name == df_np.index.name` assertion is now `df_cp.index.equals(df_np.index)`, which pins full index parity against numpy without overlapping the `index.name is None` check above it.

Re-checked the final diff: still one deleted no-op line in `_stats_cupy` plus the regression test. No new findings.

### Blockers
None.

### Suggestions
None.

### Nits
None remaining.

PR #2652, opened 2026-05-29
Filter reproject footprint coordinate pairs by joint finite mask

### LINKED ISSUE #2643: Dask footprint filtering builds invalid bboxes from unpaired x/y values
### Describe the bug

`_source_footprint_in_target()` in `xrspatial/reproject/__init__.py` computes the approximate bounding box of a source raster in the target CRS. The dask path uses this bbox to skip chunks that don't overlap the source. After transforming a set of probe points, it filters the transformed x and y coordinates independently:

```python
tx = [v for v in tx if np.isfinite(v)]
ty = [v for v in ty if np.isfinite(v)]
return (min(tx), min(ty), max(tx), max(ty))
```

When a transform sends some points to infinity or NaN (which happens near projection limits), the surviving x values and y values no longer line up with the same points. The bbox can then take its x from one point and its y from a different point, so it may not be the true footprint. That can drop chunks that actually overlap the source, or keep chunks that don't.

The fix is to filter coordinate pairs by a joint finite mask, `finite(tx) & finite(ty)`, before computing the bbox, so every x/y used comes from the same transformed point. The same independent-filter pattern shows up in both the fast-path branch and the pyproj fallback branch.

### Expected behavior

The footprint bbox is computed only from transformed points where both coordinates are finite.

### Affected backends

dask+numpy and dask+cupy. The chunk-skipping optimization lives in the dask reproject path.

brendancol:
Closes #2643

## What changed

- `_source_footprint_in_target()` filtered the transformed x and y coordinates with separate finite masks. When a transform sent some probe points to NaN or inf, the surviving x and y values no longer matched the same points, so the footprint bbox could mix coordinates from different points. That can throw off the dask chunk-skipping decision.
- Added `_finite_pair_bbox()`, which keeps only pairs finite in both coordinates before computing the box, and used it in both the fast-path and pyproj fallback branches.

## Backend coverage

The footprint helper feeds the dask reproject chunk-skipping path (dask+numpy and dask+cupy). The helper itself is backend-agnostic; the numpy/cupy non-dask paths don't use it.

## Test plan

- [x] `test_finite_pair_bbox_joint_mask`: unpaired NaN x/y collapse to the single jointly-finite point, not the wider independent-filter box.
- [x] `test_finite_pair_bbox_all_nan`: returns None when no pair is finite (NaN and inf).
- [x] `test_footprint_chunk_skip_with_unpaired_nan`: the joint footprint correctly skips a chunk that the spurious independent footprint would have kept.
- [x] Full `test_reproject.py` suite passes (383 tests).


brendancol (commented):
## PR Review: Filter reproject footprint coordinate pairs by joint finite mask

None.

None.

- The two call sites read `bbox = _finite_pair_bbox(...); if bbox is None: return None; return bbox`. This can collapse to `return _finite_pair_bbox(...)` since both branches return None on a None result.

### What looks good
- Root cause is fixed at the source. Pairs are masked jointly with `np.isfinite(tx) & np.isfinite(ty)` before the min/max, so a NaN or inf in either coordinate drops the whole point.
- Both the fast-path branch and the pyproj fallback branch were updated, so the bug cannot survive in one path.
- `_finite_pair_bbox` uses `np.asarray(..., dtype=float64)`, which handles both the numpy arrays from `transform_points` and the lists/tuples from `transformer.transform`.
- inf is covered too, not just NaN. `np.isfinite` rejects both, and there is a test for it.
- The regression test constructs tx/ty where independent filtering would yield (1, 2, 5, 6) and asserts the joint result is the single point (5, 6, 5, 6). A second test shows the spurious box would have kept a chunk the correct box skips.

### Checklist
- Algorithm matches intent: yes, joint finite mask
- Backends consistent: helper is backend-agnostic; feeds the dask chunk-skip path
- NaN handling correct: yes, NaN and inf both dropped
- Edge cases covered: all-NaN and inf return None; tested
- Dask chunk boundaries: chunk-skip decision tested against correct and spurious footprints
- No premature materialization: probe set is 12 fixed points
- Benchmark: not needed, internal helper on a tiny fixed-size array
- README feature matrix: not applicable, no public API change
- Docstrings: `_finite_pair_bbox` documents why pairs must be filtered together

brendancol (commented):

The nit from the first pass is addressed: both call sites now `return _finite_pair_bbox(...)` directly instead of the redundant `bbox = ...; if bbox is None: return None; return bbox`. This is behavior-preserving, since the prior code also returned None on a None result without falling through.

No new findings. No Blockers, Suggestions, or remaining Nits. Footprint tests still pass (5 passed).

Issue #2619, opened 2026-05-29
Add test coverage for 3D crosstab on cupy and dask+cupy backends


brendancol:
## Description

`crosstab()` in `xrspatial/zonal.py` accepts 3D categorical `values` arrays, where a category dimension is picked with `layer=`. The cupy (`_crosstab_cupy`) and dask+cupy (`_crosstab_dask_cupy`) backends handle 3D input by moving the data to host memory and calling the numpy implementation. Those 3D GPU paths run and return correct results, but no test covers them.

The 3D crosstab tests we have (`test_crosstab_3d_count`, `test_crosstab_3d_agg_method`, `test_nodata_values_crosstab_3d`) only parametrize `numpy` and `dask+numpy`. The 2D crosstab GPU paths got coverage in `test_zonal_backend_coverage_2026_05_27.py`, but the 3D GPU paths were left out.

## Why it matters

The test-coverage sweep flagged this as a backend-coverage gap. If the cupy or dask+cupy 3D handling broke (say, the host round-trip of a 3D cupy array, or the `is_cupy_array` check on the category coordinate), nothing would catch it, since no test runs that path on a GPU host.

## Proposed change

Add 3D crosstab tests for cupy and dask+cupy that check parity against the numpy backend with `agg='count'`. Tests only; no source change expected.

PR #2653, opened 2026-05-29
crosstab: validate backend compatibility on the 3D path

### LINKED ISSUE #2640: crosstab() skips backend/shape validation on the 3D path
**Describe the bug**

`crosstab()` only validates backend and shape compatibility on the 2D path. At `xrspatial/zonal.py:1414` the code reads `if values.ndim == 2: validate_arrays(zones, values)`, so a 3D call never checks that `zones` and `values` share a backend or that their spatial shapes line up.

When a 3D call mixes backends (numpy `zones` with a dask 3D `values`, say), nothing fails early. The call instead crashes deep inside the dask/numba machinery. With numpy zones and dask values you get `TypeError: 'NoneType' object is not subscriptable`, raised while building the expected dask chunk layout because numpy zones have `chunks=None`. Nothing in that error points back to the real problem, which is a backend mismatch.

**Expected behavior**

The 3D path should run the same early validation the 2D path already gets: `zones` must match `values` on the two spatial dimensions, and the `zones` and `values` backends must be compatible. A mismatch should raise a clear `ValueError` up front instead of crashing later.

**Reproduction**

```python
from xrspatial import zonal_crosstab as crosstab

zones = xr.DataArray(np.array([[0, 0, 1, 1], [2, 2, 3, 3]], dtype=float), dims=['lat', 'lon'])
vdata = da.from_array(np.ones(2 * 4 * 2).reshape(2, 4, 2), chunks=(2, 2, 1))
values = xr.DataArray(vdata, dims=['lat', 'lon', 'race'])
values['race'] = ['c1', 'c2']
crosstab(zones=zones, values=values, agg='count', layer=2)
# TypeError: 'NoneType' object is not subscriptable
```

**Additional context**

The 2D path already calls `validate_arrays(zones, values)`. The 3D path only checks shape much later (`zones.shape != values.shape[1:]`) and never checks backend compatibility at all.

brendancol:
Closes #2640

`crosstab()` only validated backend/shape compatibility on the 2D path. A 3D call with mismatched backends (numpy `zones` plus a dask 3D `values`, for example) sailed past validation and crashed deep inside the dask/numba machinery with `TypeError: 'NoneType' object is not subscriptable`, which gives the caller no hint that the real problem is a backend mismatch.

This adds an early backend check to the 3D path. `validate_arrays()` can't be reused here because it requires equal shapes and zones is 2D while values is 3D, so the check compares `_classify_backend()` of each input and raises a clear `ValueError` when they differ.

- Adds a backend-compatibility check for 3D `crosstab()` inputs, raising `ValueError` up front instead of crashing later.
- New test covers both directions of mismatch (numpy zones + dask values, and the reverse).

Backend coverage: the validation runs for all backends. Matching-backend 3D calls (numpy/numpy, dask/dask) are unchanged.

Test plan:
- [x] `pytest xrspatial/tests/test_zonal.py -k crosstab_3d`
- [x] `pytest xrspatial/tests/test_zonal.py` (full module, no regressions)


brendancol (commented):
## PR Review: crosstab: validate backend compatibility on the 3D path

None.

None.

- xrspatial/zonal.py:1416-1430 -- the new check covers the backend mismatch but not the "early shape" half of the original report. Spatial shape is still validated later (`zones.shape != values.shape[1:]`), which only fires after `layer` is resolved and the array is transposed. That's fine in practice: a real shape mismatch still raises a clear "Incompatible shapes" ValueError (verified). The one rough edge is that if `layer` is also invalid, the user sees "Invalid `layer`" first. Not worth reordering for this PR; flagging it so the call is on record.
- xrspatial/tests/test_zonal.py:1354 -- the test is gated on dask being available, which is right since both cases need a dask array. There's no numpy-vs-cupy mixed case, but that needs a GPU runner, so leaving it out is reasonable.

### What looks good
- The new check sits right next to the existing 2D `validate_arrays()` call, so the two paths read symmetrically.
- Reusing `_classify_backend()` keeps the backend names consistent with the rest of the codebase (numpy / cupy / dask+numpy / dask+cupy).
- The comment explains why `validate_arrays()` can't be reused here (it needs equal shapes), which is the non-obvious part.
- The test checks both directions of the mismatch and asserts the error message, not just the exception type.
- Matching-backend 3D calls (numpy/numpy, dask/dask) are unchanged; the full test_zonal.py passes (170 tests).

### Checklist
- Algorithm matches reference/paper: n/a (validation hardening, no algorithm change)
- All implemented backends produce consistent results: yes, the check is backend-agnostic
- NaN handling correct: unchanged
- Edge cases covered by tests: mixed-backend, both directions
- No premature materialization: yes, the check only inspects backend type, no compute
- Benchmark exists or not needed: not needed (validation path)
- README feature matrix updated: n/a (no new function, no backend support change)
- Docstrings present and accurate: unchanged public signature

brendancol (commented):
## PR Review (follow-up after merge with main)

Re-reviewed after merging `origin/main` (which brought in #2619's 3D crosstab cupy/dask+cupy test coverage). No code changes were needed in response to the first review; the two nits were both dismissed with reasons:

- Early-shape ordering nit: dismissed. Same-backend spatial mismatches already raise a clear "Incompatible shapes" ValueError. The "Invalid `layer`" ordering edge case is pre-existing and reordering it would expand this PR's scope past the backend gap it targets.
- No numpy-vs-cupy mixed test: dismissed. That case needs a GPU runner not present in CI.

The merge was conflict-free. The backend check survived intact and the full zonal suite passes (218 tests across test_zonal.py and the backend-coverage file, including the new #2619 3D cupy cases). No new findings.

PR #2655, opened 2026-05-29
reproject: account for datum shifts in output bounds estimation

### LINKED ISSUE #2649: reproject: output bounds estimation skips datum shifts, wrong bounds for NAD27 and high-res rasters
**Describe the bug**

`reproject()` computes the output raster grid in `_compute_output_grid()` (`xrspatial/reproject/_grid.py`), which transforms the source boundary into the target CRS through `_transform_boundary()`. That helper prefers the Numba fast path `transform_points()` (`xrspatial/reproject/_projections.py`) over pyproj.

`transform_points()` documents that it intentionally skips datum-shift wrapping, on the grounds that the metre-level error is sub-pixel for boundary estimation. But the per-pixel data path `try_numba_transform()` does apply a Helmert datum shift for the same CRS pairs. So when source and target differ by a datum shift, the output bounds are computed without the shift while the pixel data is reprojected with it. The bounds and the data disagree.

The error is not always sub-pixel. EPSG:4267 (NAD27) is treated as interchangeable with WGS84/NAD83 in `_is_supported_geographic()`, but NAD27 to WGS84/NAD83 differs by tens to over a hundred metres in the continental US. On a high-resolution raster that is many pixels of error in the computed grid.

**Reproduce**

```python
import numpy as np, pyproj
from xrspatial.reproject._projections import transform_points

src = pyproj.CRS.from_epsg(4267)  # NAD27
tgt = pyproj.CRS.from_epsg(3857)  # Web Mercator
xs, ys = np.array([-100.0]), np.array([40.0])  # CONUS

fast = transform_points(src, tgt, xs, ys)
ref = pyproj.Transformer.from_crs(src, tgt, always_xy=True).transform(xs, ys)
# fast x differs from pyproj by ~45 m because the NAD27 shift is skipped
```

**Expected behavior**

Output bounds estimation should account for the datum shift, or otherwise guarantee correct bounds for high-resolution and grid-shift-datum cases, instead of silently skipping it.

**Additional context**

Cited lines: `xrspatial/reproject/_projections.py:2115` (the "No datum-shift wrapping" note) and `xrspatial/reproject/_grid.py:261` (where `_compute_output_grid` uses it). Found during a code-review sweep.

brendancol:
Closes #2649

## What

- `transform_points()` ran its Numba projection kernels in WGS84 and skipped datum shifts. The per-pixel data path (`try_numba_transform`) applies a Helmert shift for the same CRS pairs, so output bounds estimated without the shift disagreed with the reprojected data by the shift magnitude. For NAD27 (EPSG:4267) that is tens to over a hundred metres in CONUS, which is many pixels on a high-resolution raster.
- Fix: detect datum-shift pairs via `_get_datum_params` and bail to pyproj, which applies the shift. The fast path is unchanged for WGS84/NAD83 pairs that need no shift.

## Backend coverage

`transform_points` is the shared CRS-pair dispatcher used by the bounds estimator across all backends (numpy / cupy / dask+numpy / dask+cupy). The change is backend-independent: it routes datum-shift pairs through pyproj before any per-backend array work, so all four paths get correct bounds.

## Test plan

- [x] `transform_points` returns `None` (bails to pyproj) for NAD27 <-> Web Mercator, both directions
- [x] Fast path is preserved for WGS84 and NAD83 (no shift), matching pyproj to 1e-3 m
- [x] `_transform_boundary` matches pyproj for NAD27 (was off by ~45 m in x before)
- [x] End-to-end `_compute_output_grid` bounds for a high-resolution NAD27 raster sit on the datum-shifted corners, not the unshifted ones
- [x] Full reproject suite green (500 passed)


brendancol (commented):
## PR Review: reproject: account for datum shifts in output bounds estimation

None.

None.

- `xrspatial/reproject/_projections.py:2128`: the bail also fires when source and target share the same datum-shift datum (e.g. NAD27 to a NAD27-based projected CRS), where the net shift cancels and the fast path would have been correct anyway. A few extra pairs go through pyproj for no accuracy gain. This is the safe default and the affected pairs are rare, so leaving it is fine.

### What looks good
- Root cause is correct. The fast path skipped the datum shift while the per-pixel data path (`try_numba_transform`) applies it, so the bounds and the data disagreed. The fix keys on the same `_get_datum_params` predicate the data path uses, so the two now agree by construction.
- The second bounds caller at `__init__.py:1796` already has a pyproj fallback right after the `transform_points` call, so it picks up the fix without changes.
- Tests cover the bail, the preserved no-shift fast path, boundary agreement with pyproj, and end-to-end grid bounds for a high-resolution NAD27 raster, with a guard against regressing to the unshifted bounds.

### Checklist
- [x] Algorithm matches reference (pyproj is the reference; fast path now defers to it for datum shifts)
- [x] Backend-independent change; no per-backend code paths
- [x] NaN handling unchanged (pyproj fallback already filters non-finite)
- [x] Edge cases covered (both directions, no-shift pairs, high-resolution grid)
- [x] No dask chunk boundary concerns (bounds estimation, not per-pixel)
- [x] Benchmark not needed (bounds estimation runs on a handful of points)
- [x] Docstring updated to explain the datum-shift bail

brendancol (commented):
## Follow-up review (after 89be27bb)

The one nit from the previous pass is addressed: added a comment at `xrspatial/reproject/_projections.py:2137` noting the bail is intentionally conservative for same-datum pairs (where the shift cancels), trading a rare fast-path miss for guaranteed correctness. Comment-only change, no logic touched.

No new findings. Datum-shift tests still pass (4/4). Nothing blocking.

PR #2656, opened 2026-05-29
zonal.stats: count returns 0 for empty zones (other stats stay NaN)

### LINKED ISSUE #2644: Empty-zone count semantics in zonal.stats: NaN vs 0
## Reason or Problem

In `xrspatial/zonal.py`, `stats()` computes per-zone summary statistics. For a
zone whose cells are all NaN or all `nodata_values` (an "empty" zone), the numpy
path's `_calc_stats` only calls the stat function when `len(zone_values) > 0`,
so `results[i]` stays NaN for every statistic, including `count`. The cupy and
dask paths match this: an empty zone reports `count` as NaN.

For `mean`, `sum`, `std`, and the others, NaN is a defensible answer to "what is
the mean of no values". For `count` it is awkward. A count is a cardinality (the
number of valid cells in the zone), and the natural value for an empty zone is
`0`, not NaN. Downstream code that filters or sums on counts (`df[df['count'] >
0]`, `df['count'].sum()`) breaks or silently drops rows when the count column
carries NaN.

The current behavior is pinned by tests (`test_stats_all_nan_zone`,
`test_stats_all_nan_zone_preserved`), so changing it is a behavior change, not a
bug fix. This needs an explicit decision and documentation either way.

## Proposal

Two options:

**Option A -- keep NaN for empty zones, document it.** Leave the behavior as is.
Add a paragraph to the `stats()` docstring stating that empty zones report NaN
for every statistic, including `count`. No code or test changes beyond the
docstring.

**Option B (recommended for `count` only) -- empty-zone `count` returns 0.**
Treat `count` as a cardinality. An empty zone reports `count = 0` while
`mean`, `min`, `max`, `sum`, `std`, and `var` stay NaN. Document the rule
explicitly and update the tests that pin `count = NaN` for empty zones to expect
`0`, with a comment and commit message explaining the deliberate change.

**Design (Option B):**
- numpy: `_calc_stats` already detects the empty-zone branch (`len == 0`); set
  the result to 0 when the statistic being computed is `count`.
- cupy: `_stats_cupy` has an explicit `zone_values.size == 0` branch that appends
  `float('nan')` per stat; append 0 for `count`.
- dask: the `count` reducer uses `_nanreduce_preserve_allnan(..., np.nansum)`,
  which forces NaN when all blocks are NaN. Count should instead use plain
  `np.nansum` so an all-empty zone sums to 0 across blocks.

**Usage:** No API change. `stats(zones, values, stats_funcs=['count', 'mean'])`
returns `count = 0` and `mean = NaN` for an empty zone.

**Value:** `count` becomes safe to use in numeric filters and aggregations
without special-casing NaN, and the empty-zone semantics are documented rather
than implicit.

## Stakeholders and Impacts

Users of `zonal.stats` who request `count`. Impact is limited to empty zones
(all-NaN or all-nodata). Only `count` changes; other statistics keep NaN.
`crosstab` and `apply` have their own count paths and are out of scope unless
they share the affected code.

## Drawbacks

It is a behavior change. Code that currently checks `isnan(count)` to detect
empty zones would need to check `count == 0` instead. The change is gated behind
a clear docstring note and a migration comment in the tests.

## Alternatives

Option A (document-only) avoids the behavior change but leaves `count` as NaN,
which is the awkward value the finding flags.

## Unresolved Questions

Whether to extend the 0-for-empty rule to `crosstab`/`apply` count paths. This
proposal scopes the change to `stats()` only.

If during implementation Option B turns out to cascade into `crosstab`/`apply`
or other shared code in a risky way, fall back to Option A (document-only) and
record why in the PR. Either way the docstring must state the empty-zone count
semantics precisely.

brendancol:
Closes #2644

## What this does

`zonal.stats()` used to report NaN for every statistic of an "empty" zone (one
that exists in the zones raster but has no valid values after filtering NaN and
`nodata_values`), including `count`. This makes `count` return `0` for empty
zones instead, while every other statistic stays NaN.

- `count` is a cardinality: zero valid cells means a count of `0`, not undefined.
- `mean`, `min`, `max`, `sum`, `std`, `var`, `majority`, and custom callables
  remain NaN for empty zones, since those are undefined over no values.
- The `stats()` docstring now documents the empty-zone rule explicitly.

This is a deliberate behavior change. Three tests pinned the old NaN-count
behavior; they now expect `0`, with comments and a commit message explaining
why. Issue #2644 frames the tradeoff (keep NaN vs return 0) and recommends 0
for `count` specifically.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy. The numpy path passes an `empty_zone_value`
into `_calc_stats`; the cupy path handles its `size == 0` branch; the dask path
uses a plain `nansum` count reducer so an all-empty zone totals 0. `crosstab`
and `apply` are untouched (they do not share the affected count code).

## Test plan

- [x] `test_stats_all_nan_zone` (all 4 backends): empty-zone count is 0
- [x] `test_stats_all_nan_zone_preserved` (numpy/cupy): count 0 for all-NaN zone
- [x] `test_stats_nodata_wipes_zone` (all 4 backends): count 0 for all-nodata zone
- [x] full `test_zonal.py` suite passes (169 passed locally; cupy/dask variants
      skip without CUDA)

## Skipped steps

No user-guide notebook and no README feature-matrix row: this refines the
documented behavior of an existing function and adds no new public API or
backend support.


brendancol (commented):
## PR Review: zonal.stats count returns 0 for empty zones

None.

None. The change is small and the three backend paths line up.

- `xrspatial/zonal.py` `_empty_zone_value` keys on the literal stat name
  `'count'`. A custom stats_funcs dict whose key happens to be `'count'` but
  whose callable is not the cardinality counter would also get 0 for empty
  zones. That is an unlikely corner and the current behavior is defensible
  (a column named count should behave like a count), so this is just worth a
  mention, not a change.

### What looks good
- The empty-zone rule is consistent across numpy (`_calc_stats` empty_zone_value),
  cupy (the `size == 0` branch), and dask (the dedicated `_count_reduce`).
  Verified locally: numpy and dask both return count 0 and mean/sum NaN for an
  all-NaN zone, including with ragged chunks that split a zone across blocks.
- Variance is unaffected: the dask var merge reads the raw per-block count
  stack, not the reduced count, so changing the reduced count to 0 does not
  perturb std/var.
- The dask mean stays NaN for an empty zone because sum is NaN and NaN/0 is NaN.
- The docstring documents the empty-zone semantics precisely, and the three
  tests that pinned NaN counts were updated to 0 with comments explaining the
  deliberate change.

### Checklist
- [x] Algorithm matches intent (count is a cardinality; 0 for empty)
- [x] NaN handling is correct (other stats stay NaN)
- [x] Edge cases covered by tests (all-NaN zone, all-nodata zone)
- [x] Dask chunk boundaries handled correctly (verified with ragged chunks)
- [x] Benchmark not needed (no new function, no hot-path change)

PR #2658, opened 2026-05-29
Fix per-band nodata masking for multi-band reproject

### LINKED ISSUE #2647: reproject: multi-band per-band nodata is flattened to one sentinel
**Describe the bug**

`reproject` collapses per-band nodata to a single sentinel for multi-band rasters. `_detect_nodata_raw()` in `xrspatial/reproject/_crs_utils.py` accepts the `nodatavals` attribute (the rasterio per-band convention) but reads only the first entry (`nv[0]`). The 3-D worker in `xrspatial/reproject/__init__.py` then applies that one sentinel to every band when converting source nodata to NaN, and the output `nodatavals` attr repeats the same value across all bands.

So for a raster whose bands declare different nodata values (say band 0 = -9999, band 1 = 255), only band 0's sentinel gets masked. Band 1's invalid pixels are treated as real data and resampled into the output.

**Expected behavior**

Each band's own nodata sentinel should be masked before resampling, so invalid pixels in every band become NaN (or the resolved output sentinel) instead of leaking through as valid data. The output `nodatavals` attr should keep the per-band values.

**To reproduce**

Build a 3-band DataArray with `attrs['nodatavals'] = (-9999, 255, 0)`, drop a distinct sentinel into each band, and reproject. The band 1 and band 2 sentinels survive into the output as if they were valid samples.

**Affected backends**

**Additional context**

Cited lines: `_crs_utils.py:232` (reads `nv[0]` only), `__init__.py:413` (per-band loop uses the single sentinel), `__init__.py:1005` (output attr repeats one value).

brendancol:
Closes #2647

## What changed

- `_detect_nodata_raw` read only `nodatavals[0]`, so the 3-D reproject worker masked every band with band 0's sentinel. A raster whose bands declared different nodata values leaked the other bands' invalid pixels into the output as valid data.
- Added `_detect_band_nodata`, which resolves a per-band source sentinel tuple from `nodatavals` (or `None` when one scalar covers every band: single-band rasters, an explicit `nodata=` arg, or a uniform tuple).
- Threaded the per-band sentinels through the chunk workers so each band is masked with its own value before resampling. The single resolved output sentinel still handles out-of-bounds fills and the integer cast-back, so output contents and the `nodatavals` band count stay consistent.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy. All four route through `_reproject_chunk_numpy` / `_reproject_chunk_cupy`, which now take an optional `band_nodata` sequence.

## Test plan

- [x] New `TestPerBandNodata` covers numpy, dask+numpy, cupy, and dask+cupy with a 3-band raster carrying distinct per-band sentinels (`-9999`, `255`, `0`), plus the `_detect_band_nodata` helper.
- [x] Confirmed the test fails on the pre-fix scalar path (band 1's `255` leaks) and passes with the fix.
- [x] Full `test_reproject.py` suite passes (387 tests). CuPy cases skip without a CUDA runtime.


brendancol (commented):
## PR Review: Fix per-band nodata masking for multi-band reproject

None.

None blocking. Worth flagging in passing: `_detect_band_nodata` assumes the `nodatavals` tuple is in the same order as the band axis (rasterio's positional convention). That's the standard and the rest of the module already assumes it, so this matches existing behavior. No change needed.

- `xrspatial/reproject/_crs_utils.py`: the per-band float equality mask (`band_data == src_nd`) mirrors the existing single-band path, so it stays consistent. Float sentinels that aren't bit-exact wouldn't match, but that's a pre-existing property of the module, not introduced here.

### What looks good
- Root cause is fixed at the source: each band is masked with its own sentinel before resampling, so band 1+ sentinels no longer leak through as valid data.
- The fix threads one optional `band_nodata` argument through every dispatch path (in-memory numpy/cupy, dask, dask+cupy, streaming) without changing the scalar output-sentinel contract. Out-of-bounds fills and the integer cast-back still use the single resolved `nd`, so output contents and the `nodatavals` band count stay consistent.
- `_detect_band_nodata` returns None for the cases that don't need per-band handling (single band, explicit nodata arg, uniform tuple, length mismatch, bare scalar), so the common path is unchanged.
- Test coverage hits all four backends plus the helper. The regression test was confirmed to fail on the pre-fix scalar path (band 1's 255 leaks) and pass with the fix.

### Checklist
- [x] Algorithm matches intended behavior (per-band source masking)
- [x] All four backends route through the same per-band logic
- [x] NaN handling correct (None entries become NaN, leaving that band unmasked)
- [x] Edge cases covered (single-entry tuple, length mismatch, None entry, bare scalar)
- [x] Dask chunk boundaries unchanged; band axis stays one chunk
- [x] No premature materialization introduced; lazy paths preserved
- [x] Docstrings present on the new helper

PR #2665, opened 2026-05-29
resample: reject empty rasters with a clear ValueError

### LINKED ISSUE #2661: resample(): empty rasters not cleanly rejected on scale_factor path
**Describe the bug**

`resample()` does not cleanly reject empty rasters on the `scale_factor` path. `_validate_raster()` checks dimensionality and numeric dtype, but not for zero-sized spatial axes. When a raster with a zero-length spatial axis (0 rows or 0 columns) is passed with `scale_factor`, the call reaches coordinate indexing and surfaces an `IndexError` instead of a clear `ValueError`.

The `target_resolution` path already guards against this with a check requiring at least 2 pixels per dimension, but the `scale_factor` path has no equivalent guard.

The existing coverage test (`TestEmptyRasterRejected` in `test_resample_coverage_2026_05_27.py`) currently accepts either `IndexError` or `ValueError`, so the contract is acknowledged but not watertight.

**Expected behavior**

A public function should raise a clear `ValueError` up front, before reaching coordinate indexing, when handed a raster with a zero-sized spatial axis.

**Fix**

Extend `_validate_raster()` (or the `scale_factor` path) to detect zero-sized spatial axes and raise a clear `ValueError`, then tighten the coverage test to expect `ValueError` specifically.

**Additional context**

Found during a code review of the resample input-validation path.

brendancol:
Closes #2661

`resample()` validated dimensionality and dtype but never checked for zero-sized spatial axes. On the `scale_factor` path, an empty raster (0 rows or 0 columns) reached the output-coordinate rebuild and surfaced an opaque `IndexError` instead of a clear error. The `target_resolution` path already had a minimum-size guard; the `scale_factor` path did not.

Changes:
- Add an explicit zero-sized spatial axis check in `resample()` right after `_validate_raster()`, raising a clear `ValueError` that names the parameter and reports the shape. The guard runs before any scale math, so it covers both the `scale_factor` and `target_resolution` paths.
- Tighten the existing `TestEmptyRasterRejected` coverage tests to expect `ValueError` specifically (with a message match) instead of accepting either `IndexError` or `ValueError`.

Backend coverage: the validation runs on the public entry point before backend dispatch, so it applies to all four backends (numpy, cupy, dask+numpy, dask+cupy).

Test plan:
- [x] `TestEmptyRasterRejected` passes (zero rows and zero cols)
- [x] Full resample suites pass (284 tests)


brendancol (commented):
## PR Review: resample: reject empty rasters with a clear ValueError

None.

None.

- The PR diff shows `.claude/sweep-test-coverage-state.csv` as changed, but that is a stale-base artifact: the branch forked from 0f8df572 and origin/main has since advanced past it. The fix commit only touches resample.py and the test file. Merging main into the branch in the conflict-resolution step clears it from the diff.
- The `resample` docstring's Raises section did not mention the empty-raster case. Documented it in a follow-up commit so the new ValueError condition is listed alongside the others.

### What looks good
- The guard sits right after `_validate_raster()` and before any scale math, so the error is raised up front as intended (resample.py:1329-1333).
- Negative indexing (`shape[-2]`, `shape[-1]`) targets the spatial axes correctly for both 2D and 3D input, so the 3D per-band path is covered too.
- The new check does not overlap with the existing `target_resolution` minimum-size guard at resample.py:1359; that path keeps its own stricter "at least 2 pixels" message for 1-row/1-col input.
- The error message names the parameter and reports the actual shape, matching the style of the surrounding validation messages.
- The coverage tests now assert ValueError specifically with a message match, closing the IndexError-or-ValueError loophole.

### Checklist
- Algorithm matches reference/paper: n/a (input validation)
- All implemented backends produce consistent results: validation runs on the public entry point before backend dispatch, so it applies to all four backends
- Edge cases are covered by tests: zero rows and zero cols both tested
- Dask chunk boundaries handled correctly: n/a (guard runs before dispatch)
- No premature materialization: the check reads only `agg.shape`
- Benchmark exists or is not needed: not needed (validation)
- Docstrings present and accurate: yes (Raises section now lists the empty-raster case)

brendancol (commented):
## Follow-up review (post-fix)

Both nits from the first pass are resolved.

### Disposition
- Stale CSV artifact: fixed. Merged origin/main, then realigned `.claude/sweep-test-coverage-state.csv` to main's version so the PR diff carries only the code change. The diff now lists `xrspatial/resample.py` and `xrspatial/tests/test_resample_coverage_2026_05_27.py` only.
- Docstring gap: fixed. The `resample` Raises section now lists the zero-length spatial dimension case.

### Verification
- flake8 clean on both changed files.
- 295 tests pass across `test_resample_coverage_2026_05_27.py` and `test_resample.py`, including the two `TestEmptyRasterRejected` cases that now assert ValueError with a message match.
- PR reports MERGEABLE with no conflicts against main.

No remaining blockers, suggestions, or open nits.

PR #2670, opened 2026-05-29
Refresh nodata/nodatavals on resample identity fast path (#2662)

### LINKED ISSUE #2662: resample identity fast path leaves stale nodata/nodatavals metadata
**Describe the bug**

`resample` masks input nodata sentinels to NaN before resampling. The non-identity path refreshes all three nodata-related attrs (`_FillValue`, `nodatavals`, `nodata`) to NaN so the metadata matches the masked data. The identity fast path (`scale_factor=1.0`) only refreshes `_FillValue` and leaves `nodata` and `nodatavals` pointing at the original integer sentinel.

The result advertises a finite sentinel that no longer appears in the data. Downstream consumers that trust `attrs['nodata']` or `attrs['nodatavals']` end up masking the wrong cells, or fail to mask the NaN cells.

**Repro**

```python

data = np.array([[-9999, -9999, 10, 10],
                 [-9999, -9999, 10, 10],
                 [20, 20, 30, 30],
                 [20, 20, 30, 30]], dtype=np.float32)
r = xr.DataArray(
    data, dims=('y', 'x'),
    coords={'y': np.linspace(3, 0, 4), 'x': np.linspace(0, 3, 4)},
    attrs={'res': (1.0, 1.0), 'nodata': -9999, 'nodatavals': (-9999,)},
)
out = resample(r, scale_factor=1.0)
print(out.values[0, 0])        # nan
print(out.attrs['nodata'])     # -9999  (stale)
print(out.attrs['nodatavals']) # (-9999,)  (stale)
print(out.attrs['_FillValue']) # nan  (correct)
```

**Expected behavior**

The identity fast path should refresh `attrs['nodata']` and `attrs['nodatavals']` to NaN, matching the non-identity path, whenever the nodata mask was applied. The 3D dispatch path has the same gap and should be fixed too.

**Location**

- Identity fast path: `xrspatial/resample.py` around line 1387
- Non-identity reference behavior: `xrspatial/resample.py` around line 1502

brendancol:
Closes #2662

## What changed

- The `resample` identity fast path (`scale_factor=1.0`) and the 3D dispatch path masked nodata sentinels to NaN but refreshed only `_FillValue`, leaving `attrs['nodata']` and `attrs['nodatavals']` pointing at the original finite sentinel. Output advertised `nodata=-9999` while the data was already NaN.
- Factored the non-identity path's refresh logic into a `_refresh_nodata_attrs` helper and called it from all three paths, so masked-to-NaN output never advertises a stale integer sentinel.

## Backend coverage

The fix is backend-independent attr handling; the mask itself routes through `xarray.where`. Regression tests cover numpy and dask+numpy. cupy and dask+cupy masking is exercised by the existing cross-backend nodata coverage.

## Test plan

- [x] New `TestIdentityNodataMetadata` class covers: identity path refreshes all three attrs, identity matches the non-identity path, absent attrs stay absent, and the 3D path is fixed too.
- [x] Full resample suite passes (289 tests).


brendancol (commented):
## PR Review: Refresh nodata/nodatavals on resample identity fast path (#2662)

None.

None.

- `xrspatial/resample.py` identity path (~L1392) and non-identity path (~L1503): the explicit `_FillValue = float('nan')` line is redundant whenever `_FillValue` is already in the input attrs, since `_refresh_nodata_attrs` sets it too. It is still needed for the explicit-`nodata=`-only case where no `_FillValue` attr exists, so leaving it is correct. A one-line comment noting that would help the next reader, but it is not required.

### What looks good
- The fix is symmetric across the three code paths: identity, 3D, and non-identity now all run `if has_nodata: _FillValue=nan` followed by `_refresh_nodata_attrs`, so they agree on what the output advertises.
- The shared helper reads from the original `src_attrs` and writes to `dst_attrs`. That avoids the order-dependence bug you would hit if it read the partially-updated output attrs instead.
- Tests hit the regression directly: masked pixel is NaN and no attr advertises the stale sentinel, identity matches the non-identity path, absent attrs stay absent, and the 3D path is covered.
- The cupy/dask+cupy parametrize was dropped with a documented reason. The attr fix is backend-independent, so numpy + dask+numpy is enough.

### Checklist
- [x] Algorithm matches intended behavior
- [x] Backends produce consistent results (attr fix is backend-independent)
- [x] NaN handling is correct
- [x] Edge cases covered (absent attrs, 3D path)
- [ ] Dask chunk boundaries: not applicable, no map_overlap change
- [ ] Benchmark: not needed, no perf-sensitive change
- [ ] README feature matrix: not applicable, no new function

brendancol (commented):
## Follow-up review (#2662)

The one nit from the first pass is addressed: the identity path now carries a comment explaining that the explicit `_FillValue = nan` line covers the explicit-`nodata=`-only case (where the input had no nodata attrs), while `_refresh_nodata_attrs` handles inputs that declared `nodata` / `nodatavals`. Logic is unchanged.

Also merged `origin/main` into the branch. The merge was clean (no conflicts); git auto-merged `resample.py` and `test_resample.py`. The full resample suite passes (222 tests) after the merge.

No new blockers, suggestions, or nits.

PR #2673, opened 2026-05-29
polygonize: batch dask per-chunk compute instead of serial loop

### LINKED ISSUE #2664: polygonize: Dask backend runs serial, one dask.compute per chunk
## Reason or Problem
The Dask backend for `polygonize` runs serially. In `_polygonize_dask` (xrspatial/polygonize.py), the nested chunk loop calls `dask.compute(...)` once per chunk:

```python
for iy in range(len(row_chunks)):
    for ix in range(len(col_chunks)):
        ...
        interior, boundary = dask.compute(
            dask.delayed(_polygonize_chunk)(block, ...))[0]
```

Each chunk is submitted and computed on its own, so Dask never sees more than one block of work at a time. That defeats the point of using Dask, and it can re-walk the upstream graph once per block.

My guess is the per-chunk compute was chosen to keep peak memory down: only boundary polygons accumulate, interior polygons get appended right away, so memory tracks the boundary-polygon count instead of total polygons times chunk count. Fair tradeoff, but right now we pay for it and get no parallelism back.

## Proposal
Batch the computation instead of computing one chunk at a time. Collect the `dask.delayed(_polygonize_chunk)(...)` tasks and submit them in bounded batches (one `dask.compute(*tasks)` call per batch). A default batch size keeps peak memory in check while letting Dask schedule many chunks at once.

The per-chunk offset bookkeeping (row_offsets, col_offsets, ny_total, nx_total) has to stay intact so results map back to the right (iy, ix) block in row-major order. The downstream boundary-merge stage depends on that ordering.

**Value:** Dask parallelism comes back for multi-chunk rasters, with a documented, bounded memory cost.

## Drawbacks
A bigger batch size means higher peak memory. The fix documents the memory/parallelism tradeoff and the batch bound.

Output has to stay identical to current behaviour. This is a performance change, not a semantics change.

brendancol:
Closes #2664

## Background

The dask backend for `polygonize` originally called `dask.compute` once per chunk in a nested loop, so it ran serially. #2608 already moved it to per-row batching (one compute per row of chunks). This PR generalizes that into a fixed `batch_size` so the memory bound no longer scales with the chunk-grid shape.

## What

- Build the per-chunk `dask.delayed(_polygonize_chunk)` tasks in row-major order, then compute them in bounded batches (one `dask.compute(*tasks)` call per batch).
- Default batch size is 32 chunks. Larger batches mean more parallelism and higher peak memory; smaller batches cap memory harder. Per-row batching bounded memory by one row of chunks, which grows with the number of column chunks; a fixed batch size bounds it independent of grid shape. The tradeoff is documented in the `_polygonize_dask` docstring, a code comment, and `docs/source/reference/dask_laziness.rst`.

## Why it is output-identical

Tasks are built and their results consumed in the same row-major `(iy, ix)` order the serial loop used. Interior and boundary polygons accumulate in that order, so the downstream boundary-merge stage sees the same input. `batch_size=1` reproduces the old per-chunk behaviour exactly.

## Backend coverage

Touches the dask path only (dask+numpy and dask+cupy both route through `_polygonize_dask`). numpy and cupy paths are unchanged.

## Test plan

- [x] `test_polygonize_dask_batch_size_invariant`: batched output is vertex-identical to `batch_size=1` across batch sizes 1, 2, 3, 100 on a 9-chunk raster.
- [x] `test_polygonize_dask_batched_matches_numpy_areas`: batched multi-chunk areas match numpy for batch sizes 1, 4, 100.
- [x] Updated the #2608 regression test (`test_compute_batches_chunks`) to pin the bounded-batch contract: ceil(n_chunks / batch_size) compute calls, neither per-chunk nor per-raster.
- [x] Existing `xrspatial/tests/test_polygonize*.py` pass.

### DIFF
--- a/docs/source/reference/dask_laziness.rst
+++ b/docs/source/reference/dask_laziness.rst
@@ -248,3 +248,26 @@ Pathfinding
    * - ``multi_stop_search``
      - Fully materialized
      - Iterative A*
+
+
+Vector conversion
+=================
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 20 50
+
+   * - Function
+     - Laziness
+     - Notes
+   * - ``polygonize``
+     - Fully materialized
+     - Polygonizes each chunk, then merges polygons across chunk edges
+
+``polygonize`` runs per chunk and stitches the results, so it never holds the
+whole raster in memory at once.  Chunks are polygonized in batches: each batch
+is one ``dask.compute`` call, so dask schedules the batch in parallel instead
+of computing one chunk at a time.  Peak memory is bounded by one batch worth of
+per-chunk polygons plus the boundary polygons that accumulate for the merge.
+Larger inputs with many small chunks parallelize well; rasters whose chunks each
+produce many polygons use more memory per batch.
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -2166,9 +2166,39 @@ def _merge_from_separated(all_interior, boundary_polys, transform,
     return column, polygon_points
 
 
+# Number of chunks polygonized per dask.compute call.  Caps how many
+# per-chunk results (interior + boundary polygons) are materialized at
+# once, bounding peak memory while still letting dask schedule the batch
+# in parallel.  See _polygonize_dask for the memory/parallelism tradeoff.
+_DASK_CHUNK_BATCH_SIZE = 32
+
+
 def _polygonize_dask(dask_data, mask_data, connectivity_8, transform,
-                     atol=_DEFAULT_ATOL, rtol=_DEFAULT_RTOL):
-    """Dask backend for polygonize: per-chunk polygonize + edge merge."""
+                     atol=_DEFAULT_ATOL, rtol=_DEFAULT_RTOL,
+                     batch_size=_DASK_CHUNK_BATCH_SIZE):
+    """Dask backend for polygonize: per-chunk polygonize + edge merge.
+
+    Chunks are polygonized in batches via :func:`dask.compute`.  Each
+    batch holds up to ``batch_size`` ``dask.delayed`` tasks computed in a
+    single call, so dask schedules them in parallel instead of running
+    one chunk at a time.
+
+    Memory / parallelism tradeoff: a single ``dask.compute`` over every
+    chunk would maximize parallelism but materialize all per-chunk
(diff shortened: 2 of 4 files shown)


brendancol (commented):
## PR Review: polygonize batch dask per-chunk compute

### Blockers
None.

### Suggestions
- `batch_size` defaults to 32 as a module constant, and there is no way to set it through the public `polygonize()` API. Fine for this PR since the task scoped it as internal, but worth a line in the docstring saying that tuning means editing the constant. Not blocking.

### Nits
- `xrspatial/polygonize.py`: `dask.compute(*tasks[start:start+batch_size])` returns a tuple even for a one-element slice, so the `for interior, boundary in results` loop is safe on the single-chunk path. No change needed, just confirming it.

### What looks good
- Output identity holds up. Tasks are built and consumed in the same row-major `(iy, ix)` order as the old serial loop, and `batch_size=1` reproduces the old per-chunk behaviour. `test_polygonize_dask_batch_size_invariant` pins vertex-exact equality across batch sizes, which is the right invariant here since numpy and dask vertex ordering already differ.
- The PR supersedes the per-row batching from #2608 and updates that regression test to the bounded-batch contract, rather than leaving a per-row assertion that is now false.
- The memory/parallelism tradeoff is documented in the docstring, a code comment, and `dask_laziness.rst`.
- Backend coverage checks out: only the dask path changed, and dask+numpy and dask+cupy both route through `_polygonize_dask`.

### Checklist
- Algorithm unchanged (performance-only): yes
- Backends consistent: yes, dask path only
- NaN handling: unchanged
- Edge cases tested: multi-chunk, masked, several batch sizes
- Dask chunk boundaries: ordering preserved for the merge stage
- No premature materialization: batching bounds peak memory
- Benchmark: none added, performance-only change to existing behaviour
- Docstrings: updated

brendancol (commented):
## Follow-up review

Applied the one suggestion from the previous pass: the `_polygonize_dask` docstring now states that `batch_size` is internal-only and that tuning it means editing the module-level `_DASK_CHUNK_BATCH_SIZE` constant.

The remaining nit needed no code change (it confirmed the single-chunk slice path is safe). No blockers. Tests pass: `test_polygonize.py` and `test_polygonize_dask_row_batch_2608.py`.

PR #2674, opened 2026-05-29
Make open_geotiff bbox= work for .vrt sources

### LINKED ISSUE #2668: open_geotiff bbox= does not work for .vrt sources
**Describe the bug**
`open_geotiff` advertises `bbox=` as a dispatcher-wide kwarg that works on every backend, including VRT. The code comment at the bbox resolution site even says bbox dispatch covers VRT sources. It doesn't work for `.vrt` files.

The dispatcher resolves `bbox` to a pixel `window` before it checks whether the source is a VRT. The resolver `_bbox_to_window` calls `_read_geo_info`, which only parses TIFF headers. For a `.vrt` source it tries to read VRT XML as a TIFF header and fails before the VRT branch is reached.

**Expected behavior**
`open_geotiff("mosaic.vrt", bbox=(x_min, y_min, x_max, y_max))` should resolve the bounding box against the VRT's own georeferencing and return the windowed mosaic, like it does for a `.tif`.

**Steps to reproduce**
1. Build a georeferenced VRT mosaic (e.g. `to_geotiff(raster, "mosaic.vrt", tile_size=128)`).
2. Call `open_geotiff("mosaic.vrt", bbox=(x_min, y_min, x_max, y_max))`.
3. The call raises while parsing the VRT XML as a TIFF header instead of returning the windowed mosaic.

**Additional context**
The VRT backend already parses VRT geo-info (GeoTransform, raster size, CRS) when it reads a mosaic, so the bbox-to-window math can reuse that metadata instead of the TIFF header parser.

brendancol:
Closes #2668

## What

`open_geotiff(..., bbox=...)` raised `Invalid TIFF byte order marker` on `.vrt` sources. The dispatcher resolved `bbox` to a pixel `window` via `_bbox_to_window`, which reads a TIFF header through `_read_geo_info` and cannot parse VRT XML. That happened before the VRT branch was reached, so the read died trying to read the XML as a TIFF.

## Fix path chosen: real VRT bbox support

I implemented real `bbox=` support for VRT rather than rejecting it, because the VRT backend already parses the metadata the resolution needs:

- Detect VRT sources before bbox resolution and route them through a new `_vrt_bbox_to_window`, which parses only the VRT XML (header-only, no tile decode) and reuses the synthesised `GeoInfo` the VRT read path already builds (`_vrt_to_synthetic_geo_info`).
- Factor the bbox validation and extent math out of `_bbox_to_window` into a shared `_geo_info_to_window`, so the TIFF and VRT resolvers run identical bbox validation, rotated-affine / no-georef rejection, and out-of-extent clamping.
- Update the dispatcher comment that wrongly claimed bbox already covered VRT.

The resolved window forwards to the existing backend dispatch, so the eager, dask, and GPU VRT paths all get bbox without any new kwarg.

## Backend coverage

`bbox=` resolves to a pixel `window=` at the dispatcher, so all VRT backends inherit it: eager (numpy), dask (chunks=), and GPU (gpu=). The numpy and dask paths are covered by tests here; GPU follows the same window forwarding.

## Test plan

- [x] `bbox=` on a VRT returns the correctly windowed mosaic, matching the equivalent explicit `window=` read and the full read sliced at the resolved offsets.
- [x] Full-extent bbox clamps to file bounds and matches a no-window read.
- [x] Chunked (`chunks=`) VRT bbox read matches the eager bbox read after compute.
- [x] Validation parity with the TIFF path: mutual-exclusion with `window=`, wrong arity, inverted bbox, NaN coords, out-of-extent, no-georef VRT, rotated VRT.
- [x] Existing TIFF bbox tests (`test_bbox_2555.py`) unchanged and passing.
- [x] Full geotiff read + vrt suites pass (1660 passed, 4 skipped).


brendancol (commented):
## PR Review: Make open_geotiff bbox= work for .vrt sources

None.

- `xrspatial/geotiff/__init__.py:467-471`: the no-georef error says "this file has no GeoTIFF tags," which is shared with the TIFF path but reads oddly for a `.vrt` (a VRT has no GeoTransform, not GeoTIFF tags). Both paths now share `_geo_info_to_window`, so it would be worth softening the wording to fit both source types. Low priority; the message still points at the right fix (`window=`).

- `xrspatial/geotiff/__init__.py:419-421`: the VRT XML is parsed here for bbox resolution and parsed again by `read_vrt` downstream. The parse is header-only and cheap, and the TIFF path reads its header twice the same way, so this is consistent rather than a regression. Noting it only for completeness.

### What looks good
- Real bbox support instead of a rejection. It reuses the VRT backend's own `_vrt_to_synthetic_geo_info`, so rotated and no-georef handling match the read path exactly.
- The shared `_geo_info_to_window` keeps TIFF and VRT validation identical, so there's no message or behavior drift between the two.
- Tests cover windowed-read parity (against an explicit window and the full slice), full-extent clamping, the chunked backend, and the validation matrix.
- Reuses `_read_vrt_xml`, so the XML size cap and path-containment guards still apply.

### Checklist
- [x] Algorithm matches reference (bbox-to-window affine reused from the TIFF path)
- [x] All implemented backends produce consistent results (window forwards to eager/dask/gpu)
- [x] NaN handling is correct (NaN bbox coords rejected upfront)
- [x] Dask chunk boundaries handled correctly (bbox resolves to a window before dispatch)
- [x] No premature materialization (header-only XML parse, no tile decode)
- [x] Benchmark not needed (bug fix, no perf-critical path)
- [x] README feature matrix update not needed (no new function)

brendancol (commented):
## PR Review (follow-up pass)

- Suggestion (no-georef error wording read oddly for `.vrt`): fixed in 1e53f23d. The message now says "this source has no georeferencing (no GeoTIFF tags or VRT GeoTransform)," which fits both the TIFF and VRT paths that share `_geo_info_to_window`. The two bbox tests that matched the old wording were updated to match "no georeferencing".
- Nit (VRT XML parsed once for bbox resolution and again by `read_vrt`): not changed. The parse is header-only and cheap, and the TIFF path reads its header twice the same way, so this is consistent with the existing pattern rather than a regression.

### New findings
None. The reword is a string change plus two test-regex updates; behavior is unchanged and the bbox suite (26 tests) passes.

PR #2675, opened 2026-05-29
Make dask float polygonize chunk-invariant for rtol

### LINKED ISSUE #2666: Dask float polygonize is not chunk-invariant for rtol
**Describe the bug**

For float rasters, `polygonize` on a dask-backed array can produce a different number of polygons than the same raster as a single numpy chunk, when `rtol` is non-trivial. The result depends on how the input is chunked, so it is not chunk-invariant.

The numpy connected-component predicate in `_calculate_regions` is direction-aware. For two adjacent pixels it tests `abs(value - reference) <= atol + rtol*abs(reference)`, where `reference` is the current (higher-`ij`) pixel in scan order and `value` is its West/South neighbour. The `rtol` term scales by the magnitude of the higher-`ij` pixel.

The dask cross-chunk merge compares unordered `(min, max)` value ranges via `_ranges_close` / `_values_close`, which does not preserve that scan-order orientation. Whichever boundary polygon happens to be iterated first becomes the `reference`, so the merge decision flips with chunk order instead of matching numpy.

**Reproduce**

```python

arr = np.array([[10.0, 11.05]])
polygonize(xr.DataArray(arr), atol=0.0, rtol=0.1)                               # 1 polygon
polygonize(xr.DataArray(da.from_array(arr, chunks=(1, 1))), atol=0.0, rtol=0.1) # 2 polygons
```

Reverse the values (`[[11.05, 10.0]]`) and the failure flips: numpy returns 2 polygons, dask returns 1.

**Expected behavior**

A chunked float raster produces the same polygon partition as the unchunked input for the same `atol` / `rtol`. The fix must be direction-aware: it has to replicate numpy's scan-order orientation (higher-`ij` pixel is the `reference`) rather than using a symmetric range comparison, so it does not over-merge the cases numpy deliberately splits.

**Additional context**

Found while auditing the #2583 dask-parity work. The `_ranges_close` interval-overlap fallback was added to handle transitive value chains across chunks; the orientation bug is separate and affects the per-edge close-value decision.

brendancol:
Closes #2666.

## What

- numpy CCL groups adjacent float pixels with `abs(value - reference) <= atol + rtol*abs(reference)`, where `reference` is the higher-`ij` pixel in scan order, so the test is asymmetric in `rtol`. The dask cross-chunk merge compared unordered `(min, max)` ranges, so its merge decision flipped with chunk order.
- Each chunk-boundary polygon now carries its internal-edge pixel values keyed by global `(col, row)`. The cross-chunk merge applies numpy's higher-`ij`-is-reference orientation at the exact pixel pair straddling a boundary, so a chunked float raster produces the same polygon partition as the unchunked input.
- Integer rasters keep strict equality (empty cell map, range fallback).

## Backends

Float path: numpy / cupy (float routes through numpy), dask+numpy, dask+cupy. The merge change lives in the shared dask path used by both dask+numpy and dask+cupy. Integer behaviour is unchanged.

## Known limitation

8-connectivity at large `rtol` still diverges between numpy and dask. That is a pre-existing ring-merge topology limitation (present on `main` before this change), not the `rtol` orientation bug, so it is left for a separate fix.

## Test plan

- [x] `pytest xrspatial/tests/test_polygonize_issue_2666.py` (29 passed)
- [x] Full polygonize suite still green (`test_polygonize.py`, `_issue_2583`, `_issue_2172`, coverage files): 291 passed, 16 skipped
- [x] Fuzz across shapes / chunkings / `rtol`: 4-connectivity now matches numpy in every case (was 26 failures on `main`)


brendancol (commented):
## PR Review: Make dask float polygonize chunk-invariant for rtol

### Blockers
None.

### Suggestions
- 8-connectivity at large rtol still diverges between numpy and dask (a dense 4x4 float array with rtol=0.1 gives different polygon counts). This is pre-existing on main, not introduced here. Tracing shows it is a ring-merge topology issue, not a close-value orientation issue: it fails identically across chunk shapes that do not even split the relevant pixels, so the cross-chunk boundary logic is not the cause. The PR scopes around it correctly. Worth a follow-up issue so the limitation is tracked outside the PR body.

### Nits
- `_ranges_close` (polygonize.py:295) and its interval-overlap fallback are now reached only for integer rasters and the defensive empty-cell-map case. For integers the check reduces to equality, so the overlap branch no longer fires in practice. Harmless and pre-existing, but a one-line comment marking it as a fallback path would help the next reader.

### What looks good
- The fix targets the root cause: numpy CCL uses the higher-ij pixel as the rtol reference, and the merge now replicates that orientation at the exact boundary pixel pair instead of comparing unordered ranges.
- Boundary-cell maps are O(chunk perimeter), keyed by global coords, so the dask path gains no full-array materialization.
- Integer rasters keep strict equality through the empty cell map and range fallback.
- NaN boundary cells are skipped during extraction, so they never merge, matching numpy.
- Tests cover the repro, its reverse, column/row orientation, disconnected close-valued pixels, several chunkings, the default tolerance path, integer invariance, and dask+cupy.

### Checklist
- [x] Algorithm matches numpy CCL scan-order orientation
- [x] Implemented backends consistent (4-conn float verified; integer unchanged)
- [x] NaN handling correct
- [x] README/docs unchanged (no public API change)
- [x] Docstrings updated for the new tuple field and helper

brendancol (commented):
## Follow-up review (after review commits)

Re-reviewed the two follow-up commits.

- Suggestion (8-conn divergence at large rtol): deferred to follow-up issue #2677. The fix is a separate ring-merge topology problem, pre-existing on main, and would expand this PR well beyond the rtol orientation bug.
- Nit (`_ranges_close` now a fallback path): fixed. Its docstring now states it is the fallback for integers and the empty-cell defensive case, with float merging routed through `_cells_close_directed`.

### New changes
- Added dask+cupy parity tests for the repro and its reverse, guarded by the cuda/cupy marker. They skip locally without a GPU.

No new blockers, suggestions, or nits. The 4-connectivity rtol parity is complete and the existing suite stays green (47 passed, 3 skipped on the polygonize parity files).

PR #2676, opened 2026-05-29
Forward allow_rotated/allow_invalid_nodata to VRT per-source reads (#2672)

### LINKED ISSUE #2672: VRT source-read opt-ins (allow_rotated, allow_invalid_nodata) are accepted but not forwarded to per-source reads
**Describe the bug**

The public `read_vrt` in `xrspatial/geotiff/_backends/vrt.py` accepts `allow_rotated`, `allow_unparseable_crs`, and `allow_invalid_nodata` and documents them as opt-ins. The VRT-level metadata validation honors them, but the per-source GeoTIFF reads don't. The eager internal call and the chunked per-chunk reader forward only the codec flags (`allow_experimental_codecs`, `allow_internal_only_jpeg`); they never pass `allow_rotated` or `allow_invalid_nodata` down to the per-source `read_to_array` call. So a caller who sets `allow_rotated=True` or `allow_invalid_nodata=True` still hits a failure when a source TIFF is rotated or carries invalid integer nodata.

Call chain:
- Eager: `read_vrt` -> `_read_vrt_internal` (which is `read_vrt` in `_vrt.py`) -> per-source `read_to_array`. The internal `read_vrt` doesn't accept the two opt-ins, so the per-source read uses defaults.
- Chunked: `read_vrt` -> `_read_vrt_chunked` -> `_vrt_chunk_read` -> internal `read_vrt` -> per-source `read_to_array`. Same gap.

**Expected behavior**

When a caller passes `allow_rotated=True` or `allow_invalid_nodata=True` to `read_vrt`, those opt-ins should reach the per-source GeoTIFF read, so a rotated source TIFF or a source with invalid integer nodata can be read on both the eager and chunked paths.

**Impact**

Under `missing_sources='warn'`, a source-metadata failure that the opt-in would have suppressed gets reclassified as a hole in the mosaic instead, which can silently drop a tile.

**Additional context**

`allow_unparseable_crs` is a validation-layer opt-in only. The per-source read helper (`read_to_array`) doesn't accept it, and the VRT path takes its CRS from the VRT XML, which is already validated with the opt-in. So only `allow_rotated` and `allow_invalid_nodata` need threading to the per-source read.

brendancol:
Closes #2672

The public `read_vrt` accepts `allow_rotated`, `allow_unparseable_crs`, and `allow_invalid_nodata`, and documents the first and third as opt-ins forwarded to the per-source GeoTIFF reads. They weren't actually forwarded. The eager internal call and the chunked per-chunk reader only passed the codec flags down, so a caller who set `allow_rotated=True` or `allow_invalid_nodata=True` still hit a typed rejection on a rotated source TIFF or a source with non-representable integer nodata. Under `missing_sources='warn'` that rejection got reclassified as a hole, silently dropping the tile.

What changed:
- Added `allow_rotated` and `allow_invalid_nodata` to the internal `read_vrt` in `_vrt.py` and forwarded them to the per-source `read_to_array` call.
- Forwarded both opt-ins from the eager backend call and from `_vrt_chunk_read` (the chunked per-chunk reader) in `_backends/vrt.py`.
- `allow_unparseable_crs` is a validation-layer opt-in only. `read_to_array` doesn't accept it, and the VRT takes its CRS from the already-validated VRT XML, so it doesn't need per-source threading.

Backend coverage: numpy (eager) and dask+numpy (chunked) VRT read paths. The change lives in the shared per-source read plumbing, so the cupy / dask+cupy VRT paths inherit it through the same internal `read_vrt`.

Test plan:
- New tests in `xrspatial/geotiff/tests/vrt/test_source_opt_ins_2672.py` build a single-source VRT over a rotated source TIFF and over a uint16 source with NaN GDAL_NODATA. Each case is rejected without the opt-in and reads cleanly with it, on both eager and chunked paths.
- A `missing_sources='warn'` test confirms the opt-in keeps a readable source out of the hole bucket instead of emitting a fallback warning and recording a false hole.
- Reverting the source-side forwarding makes the five "with opt-in" cases fail, which confirms the tests exercise the fix.
- Full VRT suite (491 tests) plus the read nodata/georef suites pass; flake8 clean on the changed files.

No CHANGELOG edit (updated separately at release time). No new public API: the opt-ins already existed and were documented; this makes them take effect.


brendancol (commented):
## PR Review: Forward allow_rotated/allow_invalid_nodata to VRT per-source reads (#2672)

None.

None. The two opt-ins now reach `read_to_array` on both the eager (`_vrt.py` per-source read) and chunked (`_vrt_chunk_read`) paths, and the GPU / dask+GPU VRT paths inherit it because the per-source decode is always the same CPU `read_to_array`.

- `_vrt_chunk_read` in `_backends/vrt.py` gained `allow_rotated` / `allow_invalid_nodata` params with no docstring lines, but the existing codec kwargs there are also undocumented, so it matches the surrounding style. A one-line note would mirror the public docstring; optional since the function is internal-only.

### What looks good
- `allow_unparseable_crs` is correctly left out of the per-source read: `read_to_array` doesn't accept it, and the VRT CRS comes from the already-validated VRT XML. This matches the dask backend, which also keeps `allow_unparseable_crs` at the validation layer and forwards only `allow_rotated` / `allow_invalid_nodata` to the reader.
- Both internal `read_vrt` call sites (eager at `_backends/vrt.py:540`, chunked at `:788`) are updated; nothing was missed.
- The tests prove the fix, not just exercise it. Reverting the source-side forwarding flips the five "with opt-in" cases to failures, and they pass with it. Both rejection types (`RotatedTransformError`, `InvalidIntegerNodataError`) are real `ValueError` subclasses, so the `missing_sources='warn'` hole-reclassification test hits the actual catch path described in the issue.
- Temp filenames carry the issue number (`rotated_src_2672.tif`, etc.), and the tests reuse the existing `_write_rotated_tiff` / `_build_uint16_tiff_1774` builders so the byte layout stays shared with the non-VRT reader tests.

### Backend coverage note
The per-source read is CPU-only (`read_to_array`) for all four VRT output backends, so numpy / dask+numpy are directly tested and cupy / dask+cupy inherit the same fix. A dedicated cupy test would need a GPU runner; the eager/chunked coverage is the right level here.

### Checklist
- Algorithm matches reference/paper: n/a (plumbing fix)
- All implemented backends consistent: yes (shared per-source read)
- NaN handling correct: unchanged; nodata masking path untouched
- Edge cases covered: yes (rotated source, NaN integer nodata, warn-mode hole)
- Dask chunk boundaries: yes (opt-ins forwarded per chunk task)
- No premature materialization or copies: yes
- Benchmark: not needed (no perf-sensitive change)
- README feature matrix: not needed (no new function, no backend change)
- Docstrings: public docstring already documented the forwarding; the fix makes it true

brendancol (commented):

The one nit from the previous pass is addressed: `_vrt_chunk_read` now documents that `allow_rotated` / `allow_invalid_nodata` are forwarded to the internal reader so each task's per-source GeoTIFF read honors the opt-ins, matching the eager path.

No new findings. No Blockers, no Suggestions. The 9 tests in `test_source_opt_ins_2672.py` still pass and flake8 is clean on the changed file. Nothing else to fix.

PR #2684, opened 2026-05-29
Enforce single-line records in sweep state CSV writers

### LINKED ISSUE #2680: Sweep state CSV writers emit multi-line records that merge=union corrupts
**Describe the bug**

The sweep state files (`.claude/sweep-*-state.csv`) carry `merge=union` in `.gitattributes` so parallel sweep agents can append rows without clobbering each other. Union merge is line-based: it unions and dedupes *physical lines*. The sweep writers use `csv.DictWriter` with `QUOTE_MINIMAL`, which is valid CSV but writes a newline-containing note as a quoted field spanning several physical lines.

When two agents append concurrently, union merge splits those multi-line records on their internal newlines. The file ends up with phantom rows, duplicate header lines, and notes torn apart mid-sentence. PR #2679 repaired the latest instance in `sweep-test-coverage-state.csv` (31 corrupted physical lines collapsed back to 7 clean records).

The `"notes": "<single-line notes or empty>"` placeholder in the writer is a hint, not enforcement. Nothing stops an agent from writing a note that contains a newline.

**Expected behavior**

Each record occupies exactly one physical line on disk regardless of note content, so union merges stay record-aligned and cannot corrupt the file.

**Fix**

In every sweep writer (`.claude/commands/sweep-*.md` and `deep-sweep.md`), sanitize each field to collapse embedded CR/LF to a `" | "` separator right before `writerow`, so a record is always one physical line. This enforces the invariant in code rather than relying on the unenforced placeholder hint.

**Additional context**

Affects all seven sweep commands plus the deep-sweep reset path. The OS/browser/device fields from the bug template do not apply; this is repository tooling, not a runtime code path.

brendancol:
Closes #2680

## What

Adds an `_oneline()` sanitizer to every sweep state-CSV writer so each record is always one physical line, even when a note contains newlines.

## Why

The `.claude/sweep-*-state.csv` files use `merge=union` so parallel sweep agents can append rows. Union merge is line-based, but the writers used `csv.DictWriter` with `QUOTE_MINIMAL`, which is valid CSV yet writes a newline-containing note as a quoted field spanning several physical lines. Concurrent appends then split those records on their internal newlines, producing phantom rows and duplicate headers. PR #2679 repaired one such corruption by hand; this stops it recurring at the source.

The previous `"single-line notes or empty"` placeholder was a hint, not enforcement, so nothing prevented a multi-line note.

## How

Each writer (six `sweep-*.md` commands plus the `deep-sweep` reset path) now routes every field through `_oneline()`, which replaces CR/LF with `" | "` right before `writerow`.

## Verification

Simulated the patched writer and a union merge:
- A note containing `\n`, `\r\n`, and `\r` writes to exactly one physical line and round-trips through `csv.DictReader`.
- Control: the old writer produced 4 physical lines for the same note (the bug).
- A union merge of two concurrent single-line writes yields clean records with no phantom rows.

No pytest test is added: these are instruction templates in `.claude/commands/`, not importable code, so there is no unit to import in `xrspatial/tests/`. Verification is the simulation above.

## Scope note

`.claude/commands/sweep-style.md` is untracked in the working tree and not part of this branch, so it is not patched here. It needs the same `_oneline()` change when it is committed.

### DIFF
--- a/.claude/commands/deep-sweep.md
+++ b/.claude/commands/deep-sweep.md
@@ -119,11 +119,17 @@ with path.open() as f:
     reader = csv.DictReader(f)
     header = reader.fieldnames
     rows = [r for r in reader if r["module"] != "{module}"]
+def _oneline(v):
+    # merge=union is line-based: a newline inside a quoted field splits
+    # the record on parallel-agent merges. Force one physical line per
+    # record by collapsing embedded newlines to " | ".
+    return "" if v is None else str(v).replace("\r\n", " | ").replace("\r", " | ").replace("\n", " | ")
+
 with path.open("w", newline="") as f:
     w = csv.DictWriter(f, fieldnames=header, quoting=csv.QUOTE_MINIMAL)
     w.writeheader()
     for r in rows:
-        w.writerow(r)
+        w.writerow({k: _oneline(v) for k, v in r.items()})
 ```
 
 This removes only the target module's row from each state file, leaving
--- a/.claude/commands/sweep-accuracy.md
+++ b/.claude/commands/sweep-accuracy.md
@@ -263,11 +263,17 @@ If CUDA_AVAILABLE is false:
        "notes": "<single-line notes (replace any newlines with spaces), or empty>",
    }
 
+   def _oneline(v):
+       # merge=union is line-based: a newline inside a quoted field splits
+       # the record on parallel-agent merges. Force one physical line per
+       # record by collapsing embedded newlines to " | ".
+       return "" if v is None else str(v).replace("\r\n", " | ").replace("\r", " | ").replace("\n", " | ")
+
    with path.open("w", newline="") as f:
        w = csv.DictWriter(f, fieldnames=header, quoting=csv.QUOTE_MINIMAL)
        w.writeheader()
        for m in sorted(rows):
-           w.writerow(rows[m])
+           w.writerow({k: _oneline(v) for k, v in rows[m].items()})
    ```
 
    Use empty strings (not `null`) for missing values. Set `issue` to the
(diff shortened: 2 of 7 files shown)


brendancol (commented):
## PR Review: Enforce single-line records in sweep state CSV writers

This is a tooling change to instruction templates in `.claude/commands/`, so the spatial-domain checks (algorithms, backend parity, dask/cupy, NaN, benchmarks, README matrix) do not apply. The review covers the sanitizer logic and the writer change.

### Blockers
None.

### Suggestions
None blocking.

### Nits
- The `"notes": "<single-line notes ...>"` placeholder hints are now redundant, since `_oneline()` enforces single-line output regardless of what the agent writes. Not worth a churny edit across files, but a future pass could simplify those hints to just describe the content and drop the "replace newlines" instruction.
- `.claude/commands/sweep-style.md` is untracked and not on this branch, so it keeps the old unguarded writer. The PR body already flags this. It needs the same `_oneline()` change when it is committed.

### What looks good
- The replace order (`\r\n` before `\r`/`\n`) avoids double-expanding a CRLF into `" |  | "`. Verified.
- Applying the sanitizer to every field via the dict comprehension, rather than just `notes`, is cheap defense in depth.
- Nice side effect: because existing rows are read back through `DictReader` and rewritten through `_oneline()`, the next write self-heals any rows that were already corrupted with embedded newlines.
- `deep-sweep`'s reset path gets the same guard, so a reset also normalizes.

### Verification performed
- A note containing `\n`, `\r\n`, and `\r` writes to one physical line and round-trips through `DictReader`.
- Control: the pre-fix writer produced 4 physical lines for the same note.
- A simulated `merge=union` of two concurrent single-line writes produced clean records with no phantom rows.

PR #2687, opened 2026-05-29
slope: widen name type hint to Optional[str] for terrain-family parity

### LINKED ISSUE #2681: slope() annotates name as str while aspect/curvature use Optional[str]
## Summary

In `xrspatial/slope.py`, the public `slope()` function annotates its `name` parameter as `name: str = 'slope'`. The other terrain functions all annotate the same parameter as `Optional[str]`:

- `aspect()` (aspect.py): `name: Optional[str] = 'aspect'`
- `northness()` (aspect.py): `name: Optional[str] = 'northness'`
- `eastness()` (aspect.py): `name: Optional[str] = 'eastness'`
- `curvature()` (curvature.py): `name: Optional[str] = 'curvature'`

## Why it matters

`slope()` hands `name` straight to `xr.DataArray(name=name)`, which accepts `None`. So `slope(agg, name=None)` runs fine, but the `str` annotation tells type checkers and IDE users that `None` is invalid. The hint contradicts the runtime behavior and breaks consistency with the rest of the family.

## Fix

Change the annotation to `name: Optional[str] = 'slope'` and import `Optional` in slope.py (the module currently imports only `Union` from `typing`). Widening a type hint accepts a superset of the old values, so this is non-breaking and needs no deprecation shim.

## Severity

MEDIUM. A type-hint inconsistency that misleads users and tooling. Found by the api-consistency sweep.

brendancol:
Closes #2681

`slope()` annotated its `name` parameter as `str`, but `aspect()`, `curvature()`, `northness()`, and `eastness()` all use `Optional[str]`. Since `name` flows into `xr.DataArray(name=name)` (which accepts `None`), `slope(agg, name=None)` already worked at runtime; the annotation was just wrong.

- Widen `name: str = 'slope'` to `name: Optional[str] = 'slope'` and import `Optional`.
- Add a test asserting the `name` annotation matches the rest of the terrain family.
- Add a test that `slope(agg, name=None)` returns a DataArray with `name is None`.

Widening a type hint is non-breaking, so no deprecation shim is needed.

Backend coverage: type-hint-only change on the public entry point; no backend logic touched. numpy / cupy / dask+numpy / dask+cupy paths are unchanged.

Test plan:
- [x] `pytest xrspatial/tests/test_slope.py` (43 passed locally)
- [x] new parity + `name=None` tests pass


brendancol (commented):
Domain-aware review (api-consistency sweep, slope module).

Scope: widen `slope()`'s `name` annotation from `str` to `Optional[str]` so it matches `aspect`, `curvature`, `northness`, and `eastness`, plus two tests.

Blockers: none.

Suggestions: none. The change is the minimal correct fix. `name` flows straight into `xr.DataArray(name=name)`, which accepts `None`, so widening the hint matches runtime behavior and the rest of the terrain family. No deprecation shim is needed since widening a type hint is non-breaking.

Nits: none. `Optional` import is placed correctly alongside the existing `Union` import; both are now used.

Verification:
- `get_type_hints(slope)['name']` resolves to `str | None`, matching `aspect`.
- `slope(agg, name=None)` returns a DataArray with `name is None`.
- Full `test_slope.py` passes (43 tests). No backend logic touched, so numpy/cupy/dask paths are unaffected.

PR #2694, opened 2026-05-29
slope: planar dask backends declare float32 meta dtype

### LINKED ISSUE #2686: Planar slope() dask backends report float64 dtype but compute float32
## Describe the bug

The planar `slope()` dask backends declare a `float64` lazy dtype, but the computed result is `float32`. So `result.dtype` disagrees with what `result.compute()` returns, and it disagrees with the numpy and cupy backends, which both report `float32`.

In `xrspatial/slope.py`, `_run_dask_numpy` calls `map_overlap(..., meta=np.array(()))` and `_run_dask_cupy` calls `map_overlap(..., meta=cupy.array(()))`. Both `meta` arrays default to `float64`, but the chunk functions (`_cpu`, `_run_cupy`) cast to and return `float32`, so the advertised lazy dtype is wrong.

The geodesic dask paths already pass `meta=np.array((), dtype=np.float32)`, so they stay consistent. The problem is isolated to the planar paths.

## To Reproduce

```python
import dask.array as da
from xrspatial import slope

data = np.random.default_rng(0).random((8, 10)) * 100
r = xr.DataArray(data, dims=['y', 'x'], attrs={'res': (30, 30)})
r['y'] = np.linspace(7*30, 0, 8); r['x'] = np.linspace(0, 9*30, 10)
r.data = da.from_array(r.data, chunks=(3, 4))

out = slope(r)
print(out.dtype)                  # float64  (lazy meta)
print(out.data.compute().dtype)   # float32  (actual)
```

## Expected behavior

The lazy dask dtype should match the computed dtype and the numpy/cupy backends: `float32`.

## Impact

Values are correct; only the declared dtype metadata is wrong. A caller that dispatches on or preallocates from the lazy `result.dtype` sees `float64`, then gets `float32` after compute.

## Fix

Pass `dtype=np.float32` / `dtype=cupy.float32` to the planar dask `meta` arrays, matching the geodesic paths.

Found by the deep-sweep metadata propagation sweep (Cat 4 dtype semantics / Cat 5 backend-inconsistent metadata).

brendancol:
Closes #2686

The planar `slope()` dask backends advertised a `float64` lazy dtype while the chunk functions returned `float32`, so `result.dtype` disagreed with `result.compute().dtype` and with the numpy/cupy backends.

- `_run_dask_numpy`: `meta=np.array(())` -> `meta=np.array((), dtype=np.float32)`
- `_run_dask_cupy`: `meta=cupy.array(())` -> `meta=cupy.array((), dtype=cupy.float32)`

Mirrors the geodesic dask paths, which already declared `float32` meta.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy all verified (CUDA host).

Test plan:
- [x] New regression tests assert lazy dtype == computed dtype == numpy dtype (float32) for dask+numpy, cupy, dask+cupy
- [x] `pytest xrspatial/tests/test_slope.py` — 44 passed


brendancol (commented):
## PR Review: slope planar dask float32 meta dtype

### Blockers
None.

### Suggestions
None.

### Nits
- test_slope.py: the three new tests repeat the same setup (data + create_test_raster). Could be parametrized, but the current form reads fine and matches the existing tests in this file, so leaving it is fine.

### What looks good
- The fix is minimal and correct. Both `_cpu` (line 64) and `_run_cupy` (cupy.empty dtype='f4', line 171) return float32, so a float32 meta matches what the chunks actually produce.
- Matches the geodesic dask paths, which already declared float32 meta. Consistent with the rest of the module.
- `meta` only changes the lazy dtype dask advertises, not the computed values, so there is no numerical risk. The QGIS and parity tests still pass.
- The new tests pin the broken invariant directly: lazy dtype == computed dtype == numpy dtype, across dask+numpy, cupy, and dask+cupy.

### Checklist
- Algorithm: unchanged
- Backend consistency: dtype now float32 across all four backends
- NaN handling: unchanged
- Edge cases: dtype invariant covered for every backend
- Dask boundaries: unchanged (depth still (1,1))
- Premature materialization: the .compute() calls live in tests only, not library code
- Benchmark: not needed (no perf change)
- README matrix: n/a (no API or backend change)
- Docstrings: unchanged

PR #2695, opened 2026-05-29
Style cleanup in viewshed.py: rename shadowed id, fix E127 + isort

### LINKED ISSUE #2690: Style cleanup in viewshed.py: shadowed id builtin, E127, isort
A style sweep over `xrspatial/viewshed.py` turned up three issues, all reported by the project's own tooling (flake8 `max-line-length=100`, isort `line_length=100`) or by grep.

First, a shadowed `id` builtin: lines 1409 and 1474 do `id = _pop(idle)` and pass `id` into `_insert_into_tree`. That local shadows the `id` builtin. The value lands in the callee's `node_id` parameter, so renaming the local to `node_id` drops the shadow and matches the signature. It isn't used anywhere else.

Second, E127 over-indented continuation lines at 2013 and 2014, in the `_viewshed_distance_sweep` signature.

Third, the `.utils` import block isn't wrapped at `line_length=100`; `isort --check-only --diff` reports the fix.

After the fix, `flake8 xrspatial/viewshed.py` should report nothing and `isort --check-only --diff xrspatial/viewshed.py` should be clean. None of this changes runtime behavior, the `id` rename is just a local variable rename.

brendancol:
Closes #2690.

Style-only cleanup of `xrspatial/viewshed.py`. Limited to what flake8 (`max-line-length=100`) and isort (`line_length=100`) flag, plus one manual rename. No formatter was run and no config was widened.

What changed:
- Renamed the local `id` to `node_id` at the two `_insert_into_tree` call sites (lines 1409 and 1474). It was shadowing the `id` builtin, and the value already maps to the callee's `node_id` parameter, so the name now matches. Pure local rename.
- Fixed E127 over-indented continuation lines in the `_viewshed_distance_sweep` signature.
- Reflowed the `.utils` import block for isort.

This covers the style sweep's Cat 1 (E127), Cat 4 (isort), and Cat 5 (shadowed `id` builtin). None of it changes runtime behavior; the `id` rename is just a local variable inside numba-jitted code, and the rest is formatting. Style fixes apply uniformly across the numpy/cupy/dask backends, so there's no per-backend verification to do.

Test plan:
- `flake8 xrspatial/viewshed.py` reports nothing
- `isort --check-only --diff xrspatial/viewshed.py` is clean
- the 25 tests in `xrspatial/tests/test_viewshed.py` pass
- a numpy viewshed smoke test exercises both renamed `_insert_into_tree` call sites


brendancol (commented):
## Review: style cleanup in viewshed.py

I read the full diff plus the surrounding code in `_viewshed_cpu_sweep` and `_viewshed_distance_sweep`, and re-ran the tooling against the head branch.

Blockers: none. Suggestions: none. Nits: none.

What looks good:
- The `id` to `node_id` rename is complete. No bare `id` tokens are left in the file, and `node_id` is already the parameter name in `_insert_into_tree` (line 321) and `_find_value_min_value` (line 75), so the rename follows existing module naming rather than inventing something new.
- No collision: `node_id` wasn't a local in `_viewshed_cpu_sweep` before, so nothing gets clobbered.
- The E127 fix lines the continuation up with the opening paren in the `_viewshed_distance_sweep` signature. flake8 reports zero now.
- The isort change just rewraps the `.utils` block; the imported names are identical.
- Checked locally: flake8 exits 0, `isort --check-only` exits 0.
- Behavior is unchanged. The renamed locals sit inside numba-jitted code and a local rename doesn't affect codegen. The 25 tests in `test_viewshed.py` pass and a numpy smoke test runs through both rename sites.

Checklist (most items not applicable for a static style change):
- Algorithm vs reference: n/a, no algorithm change
- Backend consistency: n/a, applies uniformly
- NaN handling: n/a, unchanged
- Edge-case tests: existing coverage unchanged and passing
- Dask chunk boundaries: n/a, unchanged
- Premature materialization: n/a, unchanged
- Benchmark: not needed
- README feature matrix: n/a, no API change
- Docstrings: unchanged

PR #2699, opened 2026-05-29
focal: align signatures across mean/apply/focal_stats/hotspots (#2689)

### LINKED ISSUE #2689: focal: align signatures across mean/apply/focal_stats/hotspots (agg arg, name=, defaults)
## Reason or Problem

The four public functions in `xrspatial/focal.py` have drifted apart in
their signatures. Users who switch between them hit avoidable surprises.

Current signatures:

- `mean(agg, passes=1, excludes=[np.nan], name='mean', boundary='nan')`
- `apply(raster, kernel, func=_calc_mean, name='focal_apply', boundary='nan')`
- `focal_stats(agg, kernel, stats_funcs=[...], boundary='nan')`
- `hotspots(raster, kernel, boundary='nan')`

The inconsistencies:

1. The first argument is `agg` in `mean`/`focal_stats` but `raster` in
   `apply`/`hotspots`. The rest of the library (`curvature`, `slope`,
   `aspect`, `hillshade`, `classify`, ...) uses `agg`.
2. `mean` and `apply` accept a `name=` argument; `focal_stats` and
   `hotspots` do not, so you cannot set the output DataArray name on half
   the functions.
3. `focal_stats` returns a DataArray named `focal_apply` on the numpy
   backend (a leaked internal name) but `None` on the cupy backend.
4. `mean`'s docstring omits the `excludes` parameter even though the
   signature has it.
5. `mean` and `focal_stats` use mutable list defaults (`excludes=[np.nan]`,
   `stats_funcs=[...]`).

## Proposal

Align the four signatures and fix the docstring and default-value issues.

- Rename the first argument of `apply` and `hotspots` from `raster` to
  `agg`. Keep `raster=` working as a keyword, emit a `DeprecationWarning`
  when it is used, and update the docstrings.
- Add a `name=` parameter to `focal_stats` (default `focal_stats`) and
  `hotspots` (default `hotspots`), and set the returned DataArray name
  the same way across all four backends.
- Add the missing `excludes` entry to `mean`'s docstring.
- Replace the mutable list defaults with `None` sentinels and substitute
  the real default inside the function body.

## Stakeholders and Impacts

This affects anyone calling `focal.apply` or `focal.hotspots` with the
first argument passed as the keyword `raster=`. Positional callers see no
change. The `raster=` keyword keeps working through the deprecation shim,
so nothing breaks immediately; callers get a warning and time to migrate.
The `name=` additions and the docstring and default-value fixes are
additive.

## Drawbacks

The deprecation shim adds a bit of code to two functions, and the warning
is something downstream users will eventually need to act on.

## Alternatives

Leave the names as they are. Rejected because the split first-argument
name is a visible inconsistency in a small, frequently used module.

_Surfaced by the api-consistency deep-sweep._

brendancol:
Closes #2689.

Makes the four public functions in `xrspatial/focal.py` line up on a
consistent signature.

- `apply` and `hotspots`: first argument renamed `raster` -> `agg` (matching
  `mean`, `focal_stats`, and the rest of the library). A deprecation shim
  keeps `raster=` working as a keyword and emits a `DeprecationWarning`;
  passing both `agg` and `raster` raises `TypeError`.
- `focal_stats` and `hotspots`: added a `name=` parameter and set the
  returned DataArray name the same way across backends. Before this,
  `focal_stats` returned a DataArray named `focal_apply` on numpy and `None`
  on cupy.
- `mean`: documented the existing `excludes` parameter, which the docstring
  had omitted.
- `mean` and `focal_stats`: replaced mutable list defaults with `None`
  sentinels.

Backend coverage: the `name=` and deprecation changes were verified on
numpy, cupy, dask+numpy, and dask+cupy entry points.

Test plan:
- [x] `pytest xrspatial/tests/test_focal.py` (124 passed, including new
  deprecation/name/default tests and their GPU variants)
- [x] `pytest test_validation.py test_dataset_support.py test_dask_laziness.py`
  (111 passed)
- [x] flake8 clean on the changed lines


brendancol (commented):
## PR Review: focal: align signatures across mean/apply/focal_stats/hotspots (#2689)

None.

- [ ] focal.py: `focal_stats` and `hotspots` gained a `name=` parameter. The
  `docs/source/reference/focal.rst` autosummary pulls params from the
  docstrings, so no rst change is needed here. Flagging it so the docs side
  doesn't get missed.

- [ ] focal.py `_resolve_raster_alias`: `stacklevel=3` is right for the current
  depth (user -> public func -> helper), but it's a magic number. If the call
  chain changes later, the warning points at the wrong frame. A one-line
  comment noting the assumed depth would save a future debugging session.

### What looks good
- The `raster=` shim is keyword-only (`*, raster=None`), so positional callers
  are untouched and only the deprecated keyword warns. It warns once even on
  the 3D per-band recursion, since the recursive calls pass `agg` positionally.
- Passing both `agg` and `raster` raises `TypeError` instead of silently
  picking one.
- `name=` is set the same way on numpy, cupy, dask+numpy, and dask+cupy. The
  old behaviour where `focal_stats` leaked the internal `focal_apply` name on
  numpy and returned `None` on cupy is gone.
- Mutable list defaults are now `None` sentinels, and a test checks that an
  explicit `excludes` doesn't bleed into the next default call.
- Tests cover the deprecation warnings, the both-args `TypeError`, name parity
  across backends including the GPU variants, and the default-value behaviour.

### Checklist
- [x] Algorithm matches reference/paper (no algorithm change; signature/docs only)
- [x] All implemented backends produce consistent results (name= verified on all four)
- [x] Edge cases are covered by tests (3D path, both-args conflict, default isolation)
- [x] Dask chunk boundaries handled correctly (unchanged; name= is a lazy metadata op)
- [x] No premature materialization or unnecessary copies (result.name= is metadata-only)
- [x] Benchmark exists or is not needed (not needed; no perf-relevant change)
- [x] README feature matrix updated (not applicable; no new function)
- [x] Docstrings present and accurate (excludes documented; name documented; raster deprecation noted)

brendancol (commented):
## Follow-up review (after 062160a4)

Addressed the one nit from the previous pass: `_resolve_raster_alias` now has
a comment explaining the `stacklevel=3` assumption (caller -> public func ->
helper).

The earlier Suggestion about docs was informational; the autosummary in
`focal.rst` picks up the new `name=` params from the docstrings, so no rst
change is needed.

No remaining blockers, suggestions, or open nits. Re-ran the focal API tests
(10 passed) and flake8 is clean on the changed lines.

PR #2702, opened 2026-05-29
viewshed: size max_distance window per axis on anisotropic rasters

### LINKED ISSUE #2691: viewshed max_distance clips cells on anisotropic rasters
## Description

`viewshed(..., max_distance=...)` drops cells that are within `max_distance` of the observer when the raster has anisotropic resolution (different cell spacing in x and y). Those cells come back INVISIBLE even though they should be evaluated.

In `_viewshed_windowed()` (xrspatial/viewshed.py) the analysis window is sized from a single cell radius:

```python
cell_size = max(abs(ew_res), abs(ns_res))
radius_cells = int(np.ceil(max_distance / cell_size))
r_lo = max(0, obs_r - radius_cells)
r_hi = min(height, obs_r + radius_cells + 1)
c_lo = max(0, obs_c - radius_cells)
c_hi = min(width, obs_c + radius_cells + 1)
```

`radius_cells` is derived from the coarser of the two resolutions and then used to clip both rows and columns. When `ew_res` and `ns_res` differ, the window ends up too small along the finer axis, and any cell beyond that truncated window never gets computed.

There is a circular distance mask applied afterward, but it can only remove cells, not bring back ones already cut from the window.

## Reproduce

Raster with `ew_res=1` (x) and `ns_res=10` (y). `max_distance=8` reaches 8 cells east of the observer, but `radius_cells = ceil(8/10) = 1`, so only one column on each side survives.

```python
import numpy as np, xarray as xa

ny, nx = 21, 21
t = np.zeros((ny, nx))
xs = np.arange(nx, dtype=float) * 1.0    # ew_res = 1
ys = np.arange(ny, dtype=float) * 10.0   # ns_res = 10
r = xa.DataArray(t, coords=dict(x=xs, y=ys), dims=['y', 'x'])

full = viewshed(r, x=xs[10], y=ys[10], observer_elev=50).values
win = viewshed(r.copy(), x=xs[10], y=ys[10], observer_elev=50, max_distance=8.0).values

# cell 5 columns east is 5 units away (< 8), should be visible
print(full[10, 15])  # 5.71...
print(win[10, 15])   # -1.0  (wrongly clipped)
```

## Expected

Cells within `max_distance` should be evaluated no matter how anisotropic the resolution is. The windowed result should match the full viewshed inside the radius, the same way `test_viewshed_max_distance_matches_full` already checks for isotropic rasters.

## Fix

Use a separate cell radius per axis: `radius_rows` from `ns_res`, `radius_cols` from `ew_res`.

## Impact

`_viewshed_windowed()` handles `max_distance` for every backend (numpy, cupy, dask+numpy, dask+cupy), so all of them are affected.

brendancol:
Closes #2691

`_viewshed_windowed()` sized its analysis window from a single cell radius derived from the coarser of `ew_res` / `ns_res`, then used that radius for both rows and columns. On anisotropic rasters (different x and y spacing) the window came out too small along the finer axis, so cells within `max_distance` there were dropped from the window and returned INVISIBLE.

- Compute `radius_rows` from `ns_res` and `radius_cols` from `ew_res` so each axis is windowed by its own resolution.
- Add a cross-backend regression test on an anisotropic raster (ew_res=1, ns_res=10) that checks in-range cells along the finer axis match the full viewshed instead of being clipped.

Backends: the fix lives in `_viewshed_windowed()`, the shared `max_distance` path, so numpy, cupy, dask+numpy, and dask+cupy are all covered. The new test runs numpy, cupy (skipped without rtxpy), and dask.

Test plan:
- [x] `pytest xrspatial/tests/test_viewshed.py` (28 passed, includes cupy/RTX and dask)
- [x] Reproducer from the issue now matches the full viewshed within the radius


brendancol (commented):
## PR Review: viewshed: size max_distance window per axis on anisotropic rasters

None.

None.

- xrspatial/tests/test_viewshed.py: the test only exercises one anisotropy direction (fine x, coarse y). A second case with the axes swapped (fine y, coarse x) would catch a future regression where someone wires radius_rows/radius_cols to the wrong resolution. Cheap to add via parametrize, but the current case already fails on the old code, so coverage of the bug itself is solid.

### What looks good
- The fix is the right shape: radius_rows from ns_res, radius_cols from ew_res, each axis windowed by its own spacing.
- abs() on both resolutions handles north-up rasters where ns_res is negative.
- The post-window circular mask still uses real coordinates, so the larger window does not leak extra visible cells outside max_distance.
- The regression test checks values against the full viewshed (a real reference, not just "runs without error") and confirms out-of-range cells stay INVISIBLE.
- Runs across numpy, cupy (RTX), and dask; all go through the shared _viewshed_windowed path that was fixed.

### Checklist
- [x] Algorithm matches intended behavior (per-axis radius)
- [x] NaN handling unchanged by this fix
- [x] Edge cases: out-of-range cells, anisotropic spacing covered
- [x] Dask path exercised (max_distance window)
- [x] Benchmark not needed (clip-bound fix)
- [x] README feature matrix unaffected (no API change)
- [x] Docstring already documents max_distance accurately

brendancol (commented):
## Follow-up review

The one nit from the previous pass is addressed: `test_viewshed_max_distance_anisotropic` is now parametrized over both anisotropy orientations (fine x / coarse y, and fine y / coarse x), so a swap between radius_rows and radius_cols would be caught. Full suite: 31 passed (numpy, cupy/RTX, dask).

No remaining Blockers, Suggestions, or Nits.

PR #2703, opened 2026-05-29
slope: add degenerate-shape test coverage (1x1, 1xN, Nx1)

### LINKED ISSUE #2697: slope: add test coverage for degenerate-shape rasters (1x1, 1xN, Nx1)
## Description

The slope tests (`test_slope.py`, `test_geodesic_slope.py`) cover all four backends, every boundary mode, NaN handling, constant surfaces, and several dtypes. One geometric edge case is missing: degenerate shapes where a dimension is too small for the 3x3 kernel to have an interior.

The smallest shape in the dask boundary test is (5, 5), and the cross-backend random-data test uses (2, 4). Nothing covers a 1x1 single pixel or a 1xN / Nx1 strip.

## Current behavior

I probed these shapes across numpy, cupy, dask+numpy, dask+cupy, and the geodesic method. They are all handled cleanly: the planar CPU kernel loops over `range(1, rows-1)`, which is empty when `rows <= 2`, so the output keeps the input shape and comes back all-NaN. Nothing raises, no shape mismatch. The geodesic path does the same.

There is no bug here. The problem is that no test pins this down, so a future change to the kernel bounds could start raising or returning the wrong shape on these inputs and no test would catch it.

## Proposed change

Add a parametrized test for shapes (1, 1), (1, 5), (5, 1): planar method on all four backends, plus geodesic on numpy. Assert each call returns a DataArray matching the input shape and does not raise. Test-only, no source changes.

Found by the test-coverage sweep. Severity MEDIUM: no live defect, but a kernel-boundary contract worth locking in.

brendancol:
Closes #2697

Adds tests for degenerate raster shapes where a dimension is too small for the 3x3 kernel to have an interior. These shapes already work (output keeps the input shape and comes back all-NaN), but nothing pinned that down. A future change to the kernel bounds could start raising or returning the wrong shape and no test would catch it.

- New `test_degenerate_shape_*` tests covering shapes (1, 1), (1, 5), (5, 1)
- Planar method on all four backends (numpy, cupy, dask+numpy, dask+cupy), geodesic on numpy
- Test-only change, no source modifications

Test plan:
- [x] `pytest xrspatial/tests/test_slope.py -k degenerate` -> 15 passed (cupy and dask+cupy ran on a CUDA host, not skipped)
- [x] `pytest xrspatial/tests/test_slope.py xrspatial/tests/test_geodesic_slope.py` -> 77 passed


brendancol (commented):
## PR Review: slope degenerate-shape test coverage

### Blockers
None.

### Suggestions
None.

### Nits
- test_slope.py:215 The geodesic test asserts shape and all-NaN directly but skips `general_output_checks`, which `test_degenerate_shape_numpy` (line 165) calls. The cross-backend helpers already cover attrs/coords on the dask and cupy paths, so this is minor. Adding `general_output_checks(raster, result)` to the geodesic test would make the attrs/coords/dims check uniform across all five tests.

### What looks good
- All four planar backends plus geodesic are covered through the existing helpers in general_checks.py.
- `nan_edges=False` is correct here. These outputs are entirely NaN, so `assert_nan_edges_effect` would have no non-NaN interior to compare against.
- The comment explains why these shapes come back all-NaN (the `range(1, rows-1)` loop is empty when rows<=2), so the test pins a real contract instead of freezing whatever the code happens to return today.
- Ran locally on a CUDA host: 15 degenerate tests passed, with cupy and dask+cupy actually executing rather than skipping.

### Checklist
- [x] Test-only, no algorithm change
- [x] NaN handling is correct
- [x] Edge cases covered
- [x] Benchmark not needed; README matrix and docstrings not applicable

brendancol (commented):
## Follow-up review

The one nit from the first pass is resolved: the geodesic degenerate-shape test now calls `general_output_checks(raster, result)`, so attrs/coords/dims preservation is asserted uniformly across all five degenerate tests (952062e6).

No remaining Blockers, Suggestions, or Nits. Re-ran `pytest xrspatial/tests/test_slope.py -k degenerate` -> all pass.

PR #2706, opened 2026-05-29
Fix flake8 F821 in contour.py: guard geopandas import with TYPE_CHECKING

### LINKED ISSUE #2698: flake8 F821 in contour.py: undefined name 'gpd' in contours() return annotation
**Describe the bug**

`flake8 xrspatial/contour.py` reports one violation:

```
xrspatial/contour.py:557:44: F821 undefined name 'gpd'
```

The `contours()` return type annotation uses the string forward reference `"gpd.GeoDataFrame"`, but `gpd` is only imported inside `_to_geopandas` (`import geopandas as gpd`). It is never bound at module scope, so pyflakes flags it as undefined when it parses the string annotation.

**Expected behavior**

flake8 reports zero violations for `xrspatial/contour.py`, matching the rest of the package. geopandas stays an optional runtime dependency.

**Fix**

Add a `TYPE_CHECKING`-guarded import, the same pattern already used in `xrspatial/polygonize.py`:

```python
from typing import TYPE_CHECKING, ...

if TYPE_CHECKING:
    import geopandas as gpd
```

`TYPE_CHECKING` is `False` at runtime, so geopandas stays optional and behaviour does not change. The string annotation resolves for type checkers, and pyflakes stops flagging F821.

**Additional context**

Found by the style sweep (flake8 F-codes).

brendancol:
Closes #2698

The `contours()` return annotation references `"gpd.GeoDataFrame"` as a string forward reference, but `gpd` was only imported inside `_to_geopandas`. pyflakes parses string annotations and flagged `gpd` as undefined (F821) at module scope.

- Added a `TYPE_CHECKING`-guarded `import geopandas as gpd`, the same pattern already used in `xrspatial/polygonize.py`.
- `TYPE_CHECKING` is `False` at runtime, so geopandas stays an optional dependency and there is no behaviour change. The forward reference resolves for type checkers and flake8 no longer reports F821.

Style sweep categories addressed: Cat 3 (F-codes, F821). No Cat 1/2/4/5 findings in this module. No behavioural change intended.

Backend coverage: not applicable. This is a static lint fix that applies uniformly across the numpy/cupy/dask backend paths.

Test plan:
- [x] `flake8 xrspatial/contour.py` now exits clean (previously one F821)
- [x] `isort --check-only` clean
- [x] `pytest xrspatial/tests/test_contour.py` -> 24 passed
- [x] Module imports with `gpd` not bound at runtime scope (geopandas stays optional)


brendancol (commented):
## PR Review: Fix flake8 F821 in contour.py

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The TYPE_CHECKING guard matches the existing pattern in polygonize.py, so contour.py now lines up with the rest of the package.
- TYPE_CHECKING is False at runtime, so the guarded `import geopandas as gpd` never runs in production. geopandas stays an optional dependency, and the import does not introduce a new F401 because the name is used by the string annotation.
- The "gpd.GeoDataFrame" forward reference still resolves for type checkers. Nothing about the runtime annotation changed.
- Verified locally: flake8 and isort both exit clean on contour.py, and the 24 tests in test_contour.py pass.

### Checklist
- [x] No algorithm change (import-only fix)
- [x] All backends unaffected (static fix applies uniformly)
- [x] NaN handling unchanged
- [x] Existing edge-case tests still pass
- [x] Dask paths unchanged
- [x] No materialization or copies introduced
- [x] Benchmark not needed
- [x] README not applicable (no new function)
- [x] Docstring unchanged and accurate

PR #2708, opened 2026-05-29
contours: keep CRS on empty geopandas result (#2700)

### LINKED ISSUE #2700: contours(return_type="geopandas") raises ValueError on empty result with a CRS
**Describe the bug**

`contours(agg, return_type="geopandas")` raises a `ValueError` when the input raster has a `crs` attribute but the requested levels produce no contour lines. This happens for a flat raster, for levels outside the data range, or any case where the result set is empty.

The error comes from `_to_geopandas()` in `xrspatial/contour.py`:

```python
gdf = gpd.GeoDataFrame(records, crs=crs)
```

When `records` is empty and `crs` is not `None`, geopandas raises:

```
ValueError: Assigning CRS to a GeoDataFrame without a geometry column is not supported
```

A related problem: the all-NaN early-return path calls `_to_geopandas([], None)`, which avoids the crash but drops the input CRS. So an empty result either crashes (when a CRS is present) or silently loses the CRS (on the all-NaN path).

**Reproduce**

```python
from xrspatial import contours

flat = xr.DataArray(
    np.ones((4, 4)), dims=["y", "x"],
    coords={"y": np.linspace(2, 0, 4), "x": np.linspace(0, 2, 4)},
    attrs={"crs": "EPSG:5070"},
)
contours(flat, levels=[5.0], return_type="geopandas")  # raises ValueError
```

**Expected behavior**

An empty result returns an empty GeoDataFrame that still carries the input CRS and has a `geometry` column, with no exception.

**Additional context**

Found by the deep-sweep metadata propagation sweep on the contour module. The existing geopandas test uses a raster with no `crs` attribute and a non-empty result, so this path is uncovered.

brendancol:
Closes #2700

`contours(agg, return_type="geopandas")` crashed when the input raster had a `crs` attribute but the requested levels produced no contour lines (flat raster, levels outside the data range, etc.). `_to_geopandas()` called `gpd.GeoDataFrame(records, crs=crs)` with an empty `records` list, and geopandas rejects a CRS on a frame that has no geometry column.

Changes:

- `_to_geopandas()` builds an empty frame with an explicit `geometry` column when there are no records, so the input CRS still attaches and the result is a well-formed (empty) GeoDataFrame.
- The all-NaN early-return path now passes the input CRS instead of `None`, so it no longer silently drops georeferencing.

Backend coverage: the geopandas conversion runs on the host after tracing, so this path is identical for numpy, cupy, dask+numpy, and dask+cupy. No backend-specific code changed.

Test plan:

- [x] New: empty result with a CRS returns an empty GeoDataFrame carrying the CRS (no crash)
- [x] New: all-NaN input with auto levels keeps the CRS
- [x] New: populated result still carries the CRS
- [x] New: empty result with no input CRS returns an empty frame without error
- [x] `pytest xrspatial/tests/test_contour.py` -> 28 passed


brendancol (commented):
## PR Review: contours: keep CRS on empty geopandas result (#2700)

None.

None.

- xrspatial/contour.py:551 - `crs` is passed twice in the empty branch: once to `gpd.GeoSeries([], crs=crs)` and once to the outer `GeoDataFrame(..., crs=crs)`. The outer one sets the frame CRS, so the inner one is redundant. It is harmless and guarantees the geometry column itself carries the CRS, so dropping it is optional.
- The empty-frame `level` column comes from `[]`, so its dtype is object/float rather than matching the populated path. This only matters if a caller concatenates an empty result with a populated one and checks dtypes. Low impact.

### What looks good
- The fix targets the exact failure: an empty records list plus a non-None CRS. Building the frame with an explicit geometry column is the right way to let geopandas attach the CRS.
- The all-NaN early-return now forwards the input CRS instead of None, which closes the silent-drop path.
- Four new tests cover the populated, empty-with-CRS, all-NaN, and empty-without-CRS cases. The empty-with-CRS test names the regression (#2700).
- The change is surgical, with no unrelated refactoring.

### Checklist
- Algorithm matches reference: n/a (CRS propagation fix)
- All implemented backends consistent: yes, conversion is host-side after tracing
- NaN handling correct: yes, all-NaN path verified by test
- Dask chunk boundaries: n/a
- Benchmark: not needed
- README feature matrix: not needed
- Docstrings present and accurate: yes, no API change

brendancol (commented):
## Follow-up review (after 5d46d52c)

Addressed the first nit: dropped the redundant `crs=crs` on the inner `gpd.GeoSeries([])`. The outer `GeoDataFrame(..., crs=crs)` still sets the frame CRS, and `test_geopandas_empty_result_keeps_crs` confirms the empty frame carries it. All 28 contour tests pass.

The second nit (empty-frame `level` column dtype) is left as is. Forcing a dtype on a zero-row column adds noise for an edge case that only matters under cross-frame concatenation with dtype checks, which the library does not do here. Recording it rather than fixing.

No remaining blockers or suggestions.

Issue #2704, opened 2026-05-29
contours() emits NaN coordinates for rasters containing +inf/-inf


brendancol:
## Description

`contours()` puts NaN coordinates into its output when the input raster has a `+inf` or `-inf` cell and you ask for a finite level near that cell.

The marching-squares kernel skips quads with a NaN corner using `if tl != tl or tr != tr or bl != bl or br != br: continue` (xrspatial/contour.py:67). That catches NaN but not infinity, since `inf != inf` is `False`. So a quad with an infinite corner is processed: the corner classifies as above any finite level via `>=`, and the edge interpolation `t = (level - tl) / (tr - tl)` either divides by an infinite denominator or computes `inf - inf`, which produces NaN in the emitted segment.

## Reproduce

```python
import numpy as np
from xrspatial.contour import contours
from xrspatial.tests.general_checks import create_test_raster

data = np.array([
    [0., 0., 0., 0., 0.],
    [0., 1., 1., 1., 0.],
    [0., 1., np.inf, 1., 0.],
    [0., 1., 1., 1., 0.],
    [0., 0., 0., 0., 0.],
], dtype=np.float64)
agg = create_test_raster(data, backend='numpy')
result = contours(agg, levels=[1.5])
for level, coords in result:
    print(coords)
```

The output contains NaN rows such as `[nan, 1.0]` and `[1.0, nan]`.

## Expected

A quad touching an infinite corner should be skipped the same way a NaN quad is, so the output never contains non-finite coordinates. The skip guard at line 67 needs to reject inf as well as NaN (e.g. a finiteness check on the four corners).

## Notes

Found during a test-coverage audit of the contour module. A separate test-only PR adds an xfail test documenting this behavior; this issue tracks the source fix.

Melissari1997:
Hi @brendancol, I came across xarray-spatial and wanted to see if I could contribute :)
I tried to fix this here: https://github.com/xarray-contrib/xarray-spatial/pull/2749

brendancol:
@Melissari1997 Hey thanks so much and excited for your contribution!

PR #2710, opened 2026-05-29
contour: tests for inf input, CRS propagation, and non-default dims

brendancol:
Test-only PR from a coverage audit of `xrspatial/contour.py`. No source changes.

## What this adds

- Inf handling: a test confirming the surrounding ring traces fine when an inf cell sits outside the crossing, plus two `xfail` tests (strict) documenting that a finite level near a `+inf`/`-inf` corner leaks NaN coordinates. The NaN-skip guard at contour.py:67 uses `x != x`, which does not catch infinity. Source fix tracked in #2704.
- CRS propagation: asserts `return_type="geopandas"` copies `agg.attrs['crs']` onto the output (resolves to EPSG:5070), and that a raster with no crs attr yields `gdf.crs is None`.
- Non-default dims: confirms `lat`/`lon` dim names go through the index-to-coordinate transform correctly, and that renaming `y`/`x` to `lat`/`lon` with identical coord values produces identical output.

## Backends

These cover the public `contours()` entry point on the numpy backend. Existing cross-backend equivalence tests (numpy vs cupy / dask+numpy / dask+cupy) already exist and were not changed.

## Test plan

- [x] `pytest xrspatial/tests/test_contour.py` -> 29 passed, 2 xfailed
- [x] The two xfail tests fail for the documented reason (strict xfail)

Audit note: #2704 is the separate accuracy bug surfaced by the inf tests.


brendancol (commented):
## PR Review: contour tests for inf input, CRS propagation, non-default dims

Test-only PR. Reviewed the test logic and assertions.

### Blockers
None.

### Suggestions
None.

### Nits
- The comment in `test_inf_far_level_no_crossing` first said the center quads were "all-finite", which is wrong since they contain the inf cell. They get skipped because every corner is >= 0.5 (idx 15), not because they're finite. Fixed in 3890e3ac.

### What looks good
- The two inf tests are `xfail(strict=True)` and link to #2704, so they document the NaN-leak bug without hiding a future regression. If the source gets fixed, strict xfail turns them into failures so the marker can't be left behind.
- `test_inf_far_level_no_crossing` covers the safe path (inf cell present but never interpolated), a good complement to the xfail cases.
- The CRS test checks `gdf.crs.to_epsg() == 5070` instead of string equality, so it survives GeoPandas normalization.
- Non-default-dim coverage includes an equivalence check against the y/x-named version, pinning the coordinate transform rather than only checking bounds.

### Checklist
- Backends consistent: existing cross-backend tests unchanged; the new tests are numpy-only by design (they target the public entry point and the coordinate transform, both backend-independent)
- NaN handling: the inf NaN-leak is tracked in #2704
- Edge cases: inf, missing crs attr, and non-default dims added
- Dask boundaries: n/a
- Benchmark: n/a (test-only)
- Docstrings: present on the new tests

brendancol (commented):
## Follow-up review

The one nit from the first pass (misleading "all-finite" comment in `test_inf_far_level_no_crossing`) is fixed in 3890e3ac. The comment now states the inf-touching quads are skipped via the all-above case (idx 15) before interpolation, which matches the kernel.

No remaining findings. Test-only change; source untouched. Local run: 29 passed, 2 xfailed (the two strict-xfail inf tests tracking #2704).

PR #2711, opened 2026-05-29
contours: validate input raster, reject complex dtype

### LINKED ISSUE #2701: contours() skips input validation: complex dtype silently coerced, non-DataArray errors late
**Describe the bug**

`contours()` in `xrspatial/contour.py` does not validate its input raster before processing. Unlike sibling functions (`curvature`, `morphology`, `sky_view_factor`), it skips the standard `_validate_raster()` check from `xrspatial.utils`. Two problems follow:

1. A complex-dtype DataArray is accepted silently. Inside `_contours_numpy`, `np.asarray(data, dtype=np.float64)` casts the complex array to float64 and drops the imaginary part, emitting only a `ComplexWarning`. The function then returns contour lines traced on the real part instead of rejecting the input. `_validate_raster()` already rejects complex dtypes.

2. A non-DataArray input (e.g. a plain numpy array) fails late with a confusing error. Instead of a clear "must be an xarray.DataArray" message, it reaches `ArrayTypeFunctionMapping` and raises `TypeError: Unsupported Array Type`.

**Expected behavior**

`contours()` should validate its input up front: a `ValueError` for complex/non-numeric dtypes and a `TypeError` for non-DataArray inputs, matching the rest of the codebase.

**Additional context**

Found during a security sweep (dtype confusion). The fix is to call `_validate_raster(agg, func_name='contours', name='agg', ndim=2)` at the top of `contours()`. The silent complex-dtype path is the higher-severity half: it produces wrong results with no error.

brendancol:
Closes #2701

## What

`contours()` did not validate its input before processing. This adds the standard `_validate_raster()` check the rest of the codebase uses.

- A complex-dtype DataArray was silently coerced to float64 (imaginary part dropped, only a ComplexWarning), so contours were traced on the real part instead of being rejected. Now raises ValueError.
- A non-DataArray input failed late inside `ArrayTypeFunctionMapping` with `TypeError: Unsupported Array Type`. Now raises a clear TypeError up front.

The inline `ndim == 2` check is removed because `_validate_raster(..., ndim=2)` covers it with the same "must be 2D" message.

## Backends

Validation runs before backend dispatch, so the single check covers numpy, cupy, dask+numpy, and dask+cupy. No backend-specific code changed.

## Test plan

- [x] Added test for complex dtype rejection (ValueError)
- [x] Added test for non-DataArray rejection (TypeError)
- [x] Full test_contour.py suite passes (26 tests, including cupy and dask+cupy equivalence on a CUDA host)

### DIFF
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -13,7 +13,7 @@ from typing import List, Optional, Sequence, Tuple, Union
 import numpy as np
 import xarray as xr
 
-from .utils import ArrayTypeFunctionMapping, ngjit
+from .utils import ArrayTypeFunctionMapping, _validate_raster, ngjit
 
 try:
     import dask
@@ -602,8 +602,7 @@ def contours(
     >>> # Each entry is (level_value, Nx2_coordinate_array)
     >>> level, coords = lines[0]
     """
-    if agg.ndim != 2:
-        raise ValueError("Input raster must be 2D")
+    _validate_raster(agg, func_name='contours', name='agg', ndim=2)
     if agg.shape[0] < 2 or agg.shape[1] < 2:
         raise ValueError(
             "Input raster must have at least 2 rows and 2 columns"
--- a/xrspatial/tests/test_contour.py
+++ b/xrspatial/tests/test_contour.py
@@ -180,6 +180,19 @@ class TestEdgeCases:
         with pytest.raises(ValueError, match="at least 2"):
             contours(agg, levels=[0.5])
 
+    def test_complex_dtype_rejected(self):
+        """Complex input raises instead of silently dropping the imaginary part."""
+        data = np.ones((3, 3), dtype=np.complex128)
+        agg = xr.DataArray(data, dims=['y', 'x'])
+        with pytest.raises(ValueError, match="real numeric"):
+            contours(agg, levels=[0.5])
+
+    def test_non_dataarray_rejected(self):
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: contours: validate input raster, reject complex dtype

### Blockers
None.

### Suggestions
None.

### Nits
- `test_non_dataarray_rejected` matches on "xarray.DataArray", which depends on the exact wording of `_validate_raster`'s TypeError message. Fine since both live in this repo, but the test breaks if that message ever changes. Low risk, no action needed.

### What looks good
- The fix matches the established pattern: curvature.py, morphology.py, and sky_view_factor.py call `_validate_raster` the same way.
- Validation runs before backend dispatch, so the single check covers all four backends without touching backend code.
- Removing the inline `ndim == 2` check is correct. `_validate_raster(..., ndim=2)` raises a ValueError containing "2D", so `test_invalid_ndim` still passes.
- Integer-dtype rasters still pass (numeric=True, integer_only=False), so no regression for the existing int input path.
- Both new tests assert the right exception types: ValueError for complex, TypeError for non-DataArray.

### Checklist
- Algorithm unchanged (validation-only change)
- All backends covered by the single pre-dispatch check
- NaN handling unchanged
- Edge cases (complex, non-DataArray, 3D, too-small) covered by tests
- No premature materialization
- No benchmark needed (validation guard)
- README unchanged (no API surface change)
- Docstring unchanged; the behavior change is input validation only

Issue #1378, opened 2026-04-30
mesh_utils: divide-by-zero on all-zero raster in create_triangulation


brendancol:
## Summary

`create_triangulation()` in `xrspatial/gpu_rtx/mesh_utils.py` (line 19) computes `scale = maxDim / maxH` where `maxH = float(cupy.amax(raster.data))`. On an all-zero raster `maxH` is `0.0` and `scale` becomes `inf`; on an all-NaN raster `maxH` is `nan` and `scale` becomes `nan`. Either value lands in every vertex z-coordinate via `verts[offset+2] = val * scale`, so the OptiX raytracer sees garbage geometry and `hillshade_rtx` / `viewshed_gpu` return garbage instead of an error.

## Context

Deferred Cat 3 finding from the gpu_rtx security audit. The HIGH-severity unbounded VRAM allocation in the same module was fixed in #1308 / PR #1310. The audit flagged this as MEDIUM and deferred to a follow-up:

> MEDIUM not fixed (Cat 3): mesh_utils.py:19 `scale = maxDim / maxH` divides by zero on an all-zero raster, propagating inf/NaN into mesh vertex z-coords; separate follow-up.

## Reproducer

```python
import cupy
import numpy as np
import xarray as xr
from xrspatial import hillshade

# All-zero raster (e.g. ocean tile, padding, mask result)
flat = xr.DataArray(cupy.zeros((16, 16), dtype=np.float32), dims=["y", "x"])
out = hillshade(flat, shadows=True)
# scale = 16 / 0.0 = inf, mesh z-coords become inf, garbage shadows
```

## Proposed fix

Check `maxH` before the divide and raise `ValueError` if it is non-positive or non-finite. An all-zero or all-NaN raster has no elevation to triangulate, so the mesh carries no useful information for hillshade/viewshed. A clear error beats silent garbage.

## Affected entry points

- `xrspatial.hillshade(..., shadows=True)` on a cupy raster (routes through `hillshade_rtx`)
- `xrspatial.viewshed()` on a cupy raster (routes through `viewshed_gpu`)

PR #2715, opened 2026-05-29
Add test coverage for viewshed: dask+cupy, metadata, and NaN input

### LINKED ISSUE #2693: viewshed: ValueError 'node not found' on NaN input (CPU sweep)
## Description

`viewshed` on a numpy raster crashes with `ValueError: node not found` when a NaN (nodata) cell sits at certain positions relative to the observer. The CPU sweep has explicit NaN handling in `_calc_event_elev` (it falls back to the center elevation when any of the four bilinear-interpolation neighbours is NaN), but the handling is incomplete: a NaN elevation can still produce a NaN event angle/key, and when the sweep later tries to remove that event from the red-black status tree, `_delete_from_tree` searches for a node that was never inserted and raises.

## How to reproduce

```python
import numpy as np
import xarray as xr

arr = np.full((5, 5), 2.0)
arr[2, 4] = np.nan          # NaN cell on the observer's row, to the right
r = xr.DataArray(arr, coords=dict(x=np.arange(5.0), y=np.arange(5.0)), dims=["y", "x"])
viewshed(r, x=2.0, y=2.0, observer_elev=5)   # ValueError: node not found
```

The failure is position-dependent. With the same 5x5 raster and observer at (2, 2):

- NaN at (1, 1), (0, 0), (4, 4): runs fine, no NaN propagated into the output.
- NaN at (2, 4): raises `ValueError: node not found`.

## Traceback

```
  File ".../xrspatial/viewshed.py", line 1715, in viewshed
    return _viewshed_cpu(raster, x, y, observer_elev, target_elev)
  File ".../xrspatial/viewshed.py", line 1595, in _viewshed_cpu
    viewshed_img = _viewshed_cpu_sweep(...)
  File ".../xrspatial/viewshed.py", line 562, in _delete_from_tree
    raise ValueError("node not found")
ValueError: node not found
```

## Expected behavior

A NaN/nodata cell should be treated as nodata. The GRASS `r.viewshed` algorithm this is ported from skips nodata cells when building events. The function should either skip NaN cells when generating events or mark them INVISIBLE/NaN in the output, not crash. Behaviour should not depend on where the NaN sits.

## Notes

Found during the viewshed test-coverage sweep. A test that documents the working NaN positions plus an xfail marking the crashing position is being added in a separate test-only PR; this issue tracks the underlying source fix.

## Environment

- xarray-spatial: current main
- numpy backend (CPU sweep path)

brendancol:
Test-only coverage additions for the `viewshed` module, from the test-coverage sweep. No source changes to `viewshed.py`.

## What this adds

- **dask+cupy backend.** The dask+cupy dispatch path in `_viewshed_dask` (Tier B) and `_viewshed_windowed` (max_distance) was a registered path that no test ever invoked. Adds `test_viewshed_dask_cupy_flat` (analytical-angle parity, atol 0.03) and `test_viewshed_dask_cupy_max_distance` (windowed GPU run). Both use non-zero flat terrain because the RTX mesh builder rejects an all-zero raster (#1378).
- **Metadata preservation.** Only the numpy `test_viewshed` asserted attr/coord/dim preservation via `general_output_checks`. Adds a parametrised `test_viewshed_metadata_preserved` over {numpy, cupy, dask+numpy, dask+cupy} x {full, max_distance=2.0}. It asserts attrs, dims, shape, and x/y coords are preserved, and runs the full type-parity check for every backend except dask+cupy (see below).
- **NaN input.** No test passed NaN terrain before. Adds `test_viewshed_nan_input_supported_positions` for the positions that work, plus an xfail `test_viewshed_nan_input_crashing_position` documenting the crash tracked in #2693.

## Surfaced bug (filed separately, not fixed here)

`viewshed` on a numpy raster raises `ValueError: node not found` from `_delete_from_tree` when a NaN cell sits at certain positions (e.g. (2,4) in a 5x5 with the observer at (2,2)), while NaN at other positions runs fine. Filed as #2693; the xfail test pins it.

Also noted, not fixed (a source change, out of scope for a test-only sweep): the dask+cupy path does not preserve the cupy backing. `_viewshed_dask` rewraps the computed result with `da.from_array(result_np)`, so a dask+cupy input produces a dask array that computes to numpy. That is why `general_output_checks` (which checks array-type parity) is skipped for dask+cupy in the metadata test. Candidate for the metadata/backend-parity sweep.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy. All new tests ran on a CUDA + RTX host.

## Test plan

- [x] `pytest xrspatial/tests/test_viewshed.py` — 38 passed, 1 xfailed
- [x] New tests run (not skipped) on a CUDA + RTX host
- [x] `viewshed.py` unchanged

Closes #2693 is intentionally NOT used — #2693 is the source bug, fixed elsewhere. This PR only adds the documenting tests.


brendancol (commented):
## PR Review: Add test coverage for viewshed (dask+cupy, metadata, NaN)

Test-only PR. Reviewed the three new test groups against the viewshed source and the existing test patterns. Ran the file locally: 38 passed, 1 xfailed on a CUDA + RTX host.

### Blockers
None.

### Suggestions
- `test_viewshed_dask_cupy_flat` (test_viewshed.py:479) tests a single elevation pair (observer_elev=5, target_elev=0). The numpy/cupy `test_viewshed_flat` parametrizes over a couple of pairs; matching that here would catch a target_elev regression on the dask+cupy path. Not blocking, since the eager cupy path already covers the parametrized cases.

### Nits
- The xfail reason string on test_viewshed.py:624 uses an em dash. Minor, but the rest of the file uses plain ASCII.

### What looks good
- The dask+cupy backend was genuinely untested; both new tests run the real GPU path (not skipped) on a CUDA host. Good catch on the all-zero-raster mesh-builder rejection (#1378), using 1.3 instead of 0 is the right call.
- The metadata test documents that dask+cupy does not preserve the cupy backing and skips the type-parity check for that backend only, with a comment explaining why. This is the right scope for a test-only sweep: it pins the behaviour without papering over the gap.
- The NaN crash is split into a passing test for the working positions and a strict xfail for the crashing one, both linked to #2693. The strict xfail means the test fails loudly once the source bug is fixed, which prompts removal of the marker.
- No source changes, matching the test-coverage sweep contract.

### Checklist
- All implemented backends produce consistent results: covered by the new metadata test across all four backends
- NaN handling is correct: documented, not changed (bug filed as #2693)
- Edge cases covered by tests: NaN positions added; geometry edges noted as already covered
- Dask chunk boundaries handled correctly: dask+cupy uses chunks (5,4) and (10,10)
- No premature materialization: n/a (test-only)
- Docstrings present and accurate: new test functions have docstrings

brendancol (commented):
## Follow-up review (after 6f056dd)

- Suggestion (parametrize the dask+cupy flat test): fixed. `test_viewshed_dask_cupy_flat` now parametrizes observer_elev in {5, 2} and target_elev in {0, 1}, matching the eager flat tests. All four cases run and pass on a CUDA + RTX host.
- Nit (em dash in the xfail reason): fixed. Replaced with ASCII.

No new issues. Re-ran the file: 41 passed, 1 xfailed locally.

PR #2716, opened 2026-05-29
stream_order_d8: accept `method` alias; basins_d8: emit DeprecationWarning

### LINKED ISSUE #2709: stream_order_d8 uses `ordering` kwarg while dinf/mfd siblings use `method`
## Summary

The D8 stream order function names its strahler/shreve selector differently from its sibling routing variants:

- `stream_order_d8(flow_dir, flow_accum, threshold=100, ordering='strahler', ...)`
- `stream_order_dinf(flow_dir_dinf, ..., method='strahler', ...)`
- `stream_order_mfd(..., method='strahler', ...)`

All three are publicly exported from `xrspatial.hydro`. A user who learns the dinf or mfd API and then switches to D8 (or copies code between them) hits a `TypeError` because the keyword changed name.

The unified `stream_order(..., routing=...)` dispatcher in `xrspatial/hydro/__init__.py` already papers over this with a special `_StreamOrderDispatch` class that translates `ordering` to `method` for the non-D8 paths. The comment there names the drift outright: "the ordering param (strahler/shreve) is called `ordering` in d8 and `method` in dinf/mfd."

## Proposed fix (D8 scope)

Make `stream_order_d8` accept `method` as an alias for `ordering`. Keep `ordering` working so existing callers and the dispatcher (which passes `ordering=`) are unaffected. Passing both raises a `ValueError`.

This is the smallest change that closes the user-facing inconsistency without a breaking rename. A full rename of `ordering` to `method` is out of reach here: the dispatcher in `__init__.py` (out of scope for this D8 sweep) passes `ordering=` straight to the D8 path, so deprecating `ordering` now would fire a warning on every `stream_order(routing='d8')` call.

## Second item: basins_d8 deprecated alias has no warning

`basins_d8` in `watershed_d8.py` is a backward-compatible wrapper whose docstring says "use basin instead", but it emits no `DeprecationWarning`. Add one so callers get a runtime signal.

## Scope

Only the D8 hydro files are touched. The dinf/mfd variants and the `__init__.py` dispatcher are out of scope.

brendancol:
Closes #2709

## What changed

- `stream_order_d8` now accepts a `method` keyword as an alias for `ordering`. This matches `stream_order_dinf` and `stream_order_mfd`, which already name the strahler/shreve selector `method`. `ordering` keeps working, so existing callers and the `stream_order(routing=...)` dispatcher (which passes `ordering=`) are unaffected. Passing both `ordering` and `method` with conflicting values raises `ValueError`.
- `basins_d8` (the backward-compatible wrapper for `basin_d8`) now emits a `DeprecationWarning`, which its docstring already pointed users toward.

A full rename of `ordering` to `method` was avoided on purpose: the dispatcher in `xrspatial/hydro/__init__.py` (out of scope for this D8 sweep) passes `ordering=` straight to the D8 path, so deprecating `ordering` now would warn on every unified `stream_order(routing='d8')` call.

## Backend coverage

The selector is a plain string passed to every backend, so the alias behaves the same on numpy, cupy, dask+numpy, and dask+cupy. Verified `method`/`ordering` parity on numpy and on a cupy DataArray.

## Test plan

- [x] `method` alias produces the same result as `ordering` (numpy)
- [x] `method` parity on a cupy DataArray
- [x] invalid `method` value raises `ValueError`
- [x] conflicting `ordering` + `method` raises `ValueError`
- [x] `basins_d8` emits `DeprecationWarning` and still returns the basin result
- [x] existing stream_order_d8 and watershed_d8 suites pass (70 passed)


brendancol:
## PR Review: stream_order_d8 `method` alias; basins_d8 DeprecationWarning

### Blockers
None.

### Suggestions
None. The conflict check and precedence logic hold up, and the tests cover alias parity, precedence, invalid value, conflict, and the case-insensitive non-conflict path.

### Nits
- The pre-existing `test_basins_backward_compat` (test_watershed_d8.py:153) now emits an unasserted DeprecationWarning, since the unified `basins` path in hydro/__init__.py routes through `basins_d8`. The test still passes (warnings are not errors here). If you want a clean run, wrap that call in `pytest.warns`, or just leave it since `basins` is itself the deprecated alias. Not blocking.
- `import warnings` sits inside the `basins_d8` body (watershed_d8.py:1095). It matches the local-import style the function already uses for `basin_d8`, so it's fine.

### What looks good
- `method` as a case-insensitive alias for `ordering` lines up with the `stream_order_dinf`/`stream_order_mfd` signatures, and it doesn't break the `stream_order(routing='d8')` dispatcher, which still passes `ordering=`.
- The `method: str | None = None` sentinel plus a conflict check that only fires when `ordering` is set to a non-default value is the right call here. Python can't tell an explicit default from an unset arg, so this is about as good as it gets.
- The `method: str | None` hint matches the None default and the docstring.
- basins_d8 now emits the DeprecationWarning its docstring already implied, with stacklevel=2 pointing the warning at the caller.
- The selector is a plain string handed to every backend, so the alias behaves the same on numpy, cupy, dask+numpy, and dask+cupy. 71 D8 tests pass, including 16 cross-backend cupy/dask tests on this CUDA host.

### Checklist
- [x] Algorithm unchanged (alias routes to the existing strahler/shreve paths)
- [x] Backends consistent (string selector, backend-agnostic)
- [x] NaN handling unchanged
- [x] Edge cases covered (invalid value, conflict, case-insensitivity)
- [x] README feature matrix unaffected (no new public function)

PR #2718, opened 2026-05-29
flow_path: cap downstream walk to break cycles (#2714)

### LINKED ISSUE #2714: flow_path hangs forever on a cyclic D8 flow direction grid
**Describe the bug**

`flow_path` traces downstream paths through a D8 flow direction grid by following direction codes from each start cell. The numpy kernel (`_flow_path_cpu`) and the dask path (`_flow_path_dask`) both walk downstream in a `while True` loop with no cycle guard. The loop only stops on NaN, a pit (code 0), or a grid-edge exit.

If the supplied flow direction grid contains a cycle (say, two cells pointing at each other: code 1 East and code 16 West), the walk never hits a terminating condition and loops forever. `_flow_path_cpu` runs under numba `nopython` mode, so the loop doesn't check Python signals: you can't Ctrl-C out of it, and the process has to be killed.

`flow_path` is public and accepts any user-supplied D8 grid, including grids built by hand or produced by other tools. `flow_direction()` returns acyclic grids, but nothing forces the input to come from `flow_direction()`. The test suite already notes that "randomly assigned D8 codes can create cycles" (see `test_flow_accumulation_d8.py`), so cyclic input is a known possibility.

The other hydro D8 kernels survive cycles. The ones built on Kahn's BFS topological sort (flow_accumulation, stream_order, stream_link, flow_length, hand) leave cycle cells unprocessed because they never reach in-degree zero. watershed uses an in-trace state marker that breaks on re-entry. flow_path is the one that doesn't protect itself.

**Reproduction**

```python
import numpy as np
from xrspatial.hydro import flow_path

# 2-cycle: (0,0) flows East to (0,1); (0,1) flows West to (0,0)
fd = create_test_raster(np.array([[1.0, 16.0]]), backend='numpy')
sp = create_test_raster(np.array([[5.0, np.nan]]), backend='numpy')
flow_path(fd, sp)  # hangs forever
```

**Expected behavior**

`flow_path` should terminate on any input. A downstream path can visit each cell at most once, so the walk can be capped at H*W steps. Past that, the path is in a cycle and should stop.

**Affected backends**

numpy and dask+numpy directly. cupy and dask+cupy route through `_flow_path_cpu` / `_flow_path_dask`, so they inherit the same hang.

**Additional context**

Found during a security sweep of the hydro D8 modules (denial of service via logic error). The fix is scoped to `xrspatial/hydro/flow_path_d8.py`.

brendancol:
Closes #2714

`flow_path` traced downstream paths with a `while True` loop that had no cycle guard, in both the numpy kernel (`_flow_path_cpu`) and the dask path (`_flow_path_dask`). A user-supplied D8 grid with a cycle made the walk loop forever. The numba nopython kernel can't be interrupted by a signal, so the process had to be killed; the dask path also grew its path buffers without bound.

A valid downstream path visits each cell at most once, so this caps the walk at H*W steps in both paths. The break conditions (NaN, pit, grid-edge exit) are unchanged, so acyclic tracing gives the same results as before.

- Cap the per-start-point walk at H*W steps in `_flow_path_cpu` and `_flow_path_dask`
- Add tests for a 2-cell cycle, a square 4-cell loop (numpy and dask), and an acyclic regression check

Backend coverage: numpy and dask+numpy fixed directly. cupy and dask+cupy route through the same two functions, so they inherit the fix.

Test plan:
- [x] `pytest xrspatial/hydro/tests/test_flow_path_d8.py` passes (26 tests, 4 new)
- [x] Cyclic input now returns instead of hanging
- [x] Acyclic paths unchanged

### DIFF
--- a/xrspatial/hydro/flow_path_d8.py
+++ b/xrspatial/hydro/flow_path_d8.py
@@ -138,6 +138,12 @@ def _flow_path_cpu(flow_dir, start_points, H, W):
     out = np.empty((H, W), dtype=np.float64)
     out[:] = np.nan
 
+    # A valid path visits each cell at most once, so it can take at most
+    # H*W steps.  A cyclic flow_dir grid (e.g. two cells pointing at each
+    # other) would otherwise loop forever, and numba nopython loops can't
+    # be interrupted by a signal.  Cap the walk to break out of cycles.
+    max_steps = H * W
+
     for r in range(H):
         for c in range(W):
             v = start_points[r, c]
@@ -145,7 +151,7 @@ def _flow_path_cpu(flow_dir, start_points, H, W):
                 continue
             label = v
             cr, cc = r, c
-            while True:
+            for _ in range(max_steps):
                 out[cr, cc] = label
                 code = flow_dir[cr, cc]
                 if code != code:  # NaN
@@ -336,9 +342,14 @@ def _flow_path_dask(flow_dir_data, start_points_data):
     _buf_labels = np.empty(_init_cap, dtype=np.float64)
     _buf_len = 0
 
+    # A valid path visits each cell at most once (at most H*W steps).
+    # Cap the walk so a cyclic flow_dir grid can't loop forever and grow
+    # the path buffers without bound.
+    max_steps = H * W
+
     for r, c, label in points:
         cr, cc = r, c
-        while True:
+        for _ in range(max_steps):
             # Grow buffers if needed (doubling strategy)
             if _buf_len >= len(_buf_rows):
                 new_cap = len(_buf_rows) * 2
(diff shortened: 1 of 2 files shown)


brendancol (commented):
Domain-aware review of the flow_path cycle-cap fix.

Blockers: none.

Suggestions: none.

Nits: none.

Notes:
- The cap `max_steps = H * W` is exactly tight. A valid acyclic path visits each cell at most once, so it needs at most H*W write-and-move iterations; the loop never truncates a legitimate path. Verified on the 1x2 and 2x2 cases.
- Both walk sites are converted (`_flow_path_cpu` and `_flow_path_dask`). cupy and dask+cupy reach these same functions, so all four backends are covered without separate edits.
- The dask path also accumulates every visited cell into growable buffers, so the cap fixes an unbounded-memory growth path in addition to the hang.
- Break conditions (NaN, pit, grid-edge exit) are untouched, so acyclic output is unchanged. The new acyclic regression test confirms this.

PR #2719, opened 2026-05-29
proximity: add return annotations to allocation and direction

### LINKED ISSUE #2717: proximity: add missing return annotations to allocation and direction
## Description

In `xrspatial/proximity.py`, the three public functions share the same signature and return the same type, but their return annotations don't match:

- `proximity` is annotated `-> xr.DataArray`
- `allocation` has no return annotation
- `direction` has no return annotation

All three build their result with `xr.DataArray(...)` and carry `@supports_dataset`, so they return a DataArray (or a Dataset when handed a Dataset). The docstrings already document the return type, so the annotation on `allocation` and `direction` is just missing.

## Proposed change

Add `-> xr.DataArray` to `allocation` and `direction` to match `proximity`. Annotation only: no behavior, signature, or default changes, so nothing breaks and no deprecation shim is needed.

## Found by

API-consistency sweep over the proximity module (Cat 3: type hints).

brendancol:
Closes #2717

`allocation` and `direction` now carry the same `-> xr.DataArray` return annotation that `proximity` already has. The three functions share a signature and return type, so their annotations should match.

- Annotation only: no behavior, signature, or default changes. Not a breaking change.
- Adds a parametrized test asserting the three siblings report the same return annotation via `inspect.signature`.

Backend coverage: unaffected (no runtime code changed). numpy / cupy / dask+numpy / dask+cupy paths are untouched.

Test plan:
- [x] `pytest xrspatial/tests/test_proximity.py::test_return_annotation_consistency` (3 passed)
- [x] proximity/allocation/direction default-behavior slice (17 passed)
- [x] flake8 on changed files (no new findings)


brendancol (commented):
Domain-aware review (API-consistency sweep, proximity module).

**Blockers:** none.

**Suggestions:** none blocking. The annotation widens to match `proximity`'s existing `-> xr.DataArray`, which is the right consistency target. Note that all three are wrapped by `@supports_dataset` and can return a Dataset, so a stricter `xr.DataArray | xr.Dataset` would be more precise. Matching the existing `proximity` annotation is the correct scope here; broadening all three is a separate call and out of scope for this PR.

**Nits:** none.

Verified:
- `inspect.signature` reports `xr.DataArray` for all three functions (the `@supports_dataset` wrapper preserves the signature via functools.wraps), so the new test exercises the real public surface.
- Annotation-only change; no runtime path touched. numpy / cupy / dask+numpy / dask+cupy behavior unchanged.
- New test and the proximity/allocation/direction default-behavior slice pass; flake8 reports no new findings on the changed files.

PR #2720, opened 2026-05-29
Add degenerate-shape test coverage for D8 hydro functions (#2713)

### LINKED ISSUE #2713: Add degenerate-shape (1x1 / 1xN / Nx1) test coverage for D8 hydro functions
## Summary

Four D8 hydrology functions have no tests for degenerate raster shapes: 1x1 single-pixel, 1xN single-row, and Nx1 single-column inputs. The functions are `flow_accumulation_d8`, `flow_direction_d8`, `flow_length_d8`, and `hand_d8`. On these shapes every cell sits on the kernel boundary, so they exercise a different path than interior cells.

The other D8 variants already cover single-pixel or strip inputs: sink, stream_link, stream_order, snap_pour_point, watershed, basin, fill, flow_path, and twi. These four are the gap.

## Current behavior

None of the four raise on these shapes:

- `flow_direction_d8` returns an all-NaN array of the same shape, since every cell is an edge cell.
- `flow_accumulation_d8`, `flow_length_d8`, and `hand_d8` return finite values of the matching shape.

This is not a bug. The functions work. The behavior is just untested, so a regression in the boundary handling would slip through.

## Proposed change

Add tests only, no source changes. Lock in the current behavior for these four functions on the numpy backend, parametrized over (1,1), (1,N), and (N,1). Check that the output shape matches the input and that the edge behavior holds: all-NaN for `flow_direction_d8`, finite for the other three.

Found during a test-coverage sweep of the hydro-d8 module.

brendancol:
Closes #2713.

Adds tests for degenerate raster shapes (1x1, 1xN, Nx1) to four D8 hydrology functions that had no single-pixel or strip coverage: `flow_accumulation_d8`, `flow_direction_d8`, `flow_length_d8`, and `hand_d8`.

- All four already handle these shapes without raising. The tests pin the current behavior so a regression in the kernel-boundary path gets caught.
- `flow_direction_d8`: output is all-NaN at the input shape, since every cell is an edge cell.
- `flow_accumulation_d8`, `flow_length_d8`, `hand_d8`: output is finite at the input shape. The flow_length test covers both `direction` modes.

Tests only, no source changes.

Backends: numpy. The degenerate-shape question is about kernel-boundary handling, and the numpy path is the reference. cupy, dask+numpy, and dask+cupy parity is already covered by existing tests in each of these files.

Test plan:
- [x] `pytest` on the four touched files passes on a CUDA host: 164 passed, including the 15 new cases. cupy and dask+cupy tests ran rather than skipped.


brendancol (commented):
## PR Review: Add degenerate-shape test coverage for D8 hydro functions (#2713)

Test-only change. Adds 1x1, 1xN, and Nx1 cases to four D8 test files. I read all four diffs against the source functions.

### Blockers
None.

### Suggestions
- `test_hand_d8.py` `test_degenerate_shape`: this one only checks output shape. With fa=200 and threshold=100, every cell is a stream, so HAND should be 0 everywhere. Asserting that value would catch a regression that returns the wrong magnitude, not just the wrong shape. The other three tests check a value-level property (all-NaN, min >= 1, no NaN), so this is the odd one out.

### Nits
- All four tests use the same name (`test_degenerate_shape`) with near-identical bodies. Fine across separate files. If they ever get pulled into a shared parametrized helper, the per-function edge semantics (all-NaN vs finite) would need to stay distinct.

### What looks good
- Assertions match the current behavior for each function. I verified: flow_direction returns all-NaN because every cell is an edge cell, and the other three stay finite.
- flow_length covers both direction modes.
- numpy-only scope makes sense for a kernel-boundary check, and the PR body says so. Cross-backend parity is already covered by existing tests in each file.
- Reuses the existing builders (`_make_flow_dir_raster`, `_make_hand_rasters`, `create_test_raster`), no new helpers.

### Checklist
- [x] Assertions match verified current behavior
- [x] NaN handling correct (flow_direction all-NaN edge case)
- [x] Edge cases covered (this PR is the edge-case coverage)
- [n/a] Dask chunk boundaries (numpy-only by design)
- [n/a] Benchmark (test-only)
- [n/a] README matrix (no API change)
- [x] Docstrings present on each test

brendancol (commented):
## Follow-up review (#2713)

Addressed the one Suggestion from the first pass.

- Fixed: `test_hand_d8.py` `test_degenerate_shape` now asserts `HAND == 0` for the all-stream degenerate input, not just the output shape. Verified the value before adding the assert; the strengthened test passes.
- Dismissed (with reason): the Nit about the shared `test_degenerate_shape` name. The four tests live in separate files and keep distinct per-function semantics (all-NaN for flow_direction, finite/zero for the others). No shared helper is planned, so the duplicate name is not a problem here.

No other findings remain.

brendancol:
## PR Review: Add degenerate-shape test coverage for D8 hydro functions (#2713)

Test-only PR adding 1x1, 1xN, and Nx1 regression tests to four D8 hydrology functions. I read all four changed test files in full and ran the 15 new cases on a CUDA host: all 15 passed. The hand and flow_direction tests assert the exact output. The other two assert weaker conditions than the value they document.

None.

- test_flow_accumulation_d8.py:448 -- asserts `result.data.min() >= 1.0`, but for all-pit input every cell drains only itself, so the exact value is 1.0 everywhere (np.unique returns [1.]). The `>=` check still passes if a kernel-boundary regression made a cell accumulate spurious upstream flow. Assert `result.data == 1.0` to pin the documented behavior.
- test_flow_length_d8.py:388 -- asserts `not np.isnan(result.data).any()`, but the docstring says the length is zero, and the output is 0.0 for both directions on all three shapes. A finiteness check passes even if a regression returns a wrong nonzero length. Assert `result.data == 0.0`.

None.

### What looks good
- hand_d8 asserts the exact value (`assert_allclose(result.data, 0.0)`) and flow_direction asserts all-NaN.
- All four parametrize over (1,1), (1,4), (4,1); flow_length also covers both direction modes.
- No temp files; reuses the existing create_test_raster / _make_* helpers.
- Restricting to the numpy reference backend is reasonable for a kernel-boundary question, and the PR body says why. Cross-backend parity is already covered elsewhere in each file.

### Checklist
- [x] All implemented backends produce consistent results (numpy reference, justified)
- [x] NaN handling is correct
- [x] Dask chunk boundaries handled correctly (n/a, numpy-only)
- [x] Benchmark exists or is not needed (n/a, test-only)

brendancol:
Followup da3af03: applied both suggestions. flow_accumulation now asserts the result equals 1.0 exactly and flow_length asserts 0.0 exactly, instead of the looser min >= 1.0 / not-NaN checks. All four files pass: 164 passed.

Issue #879, opened 2026-02-24
proximity KDTree: sequential `.compute()` loop + unbounded target accumulation


brendancol:
## Summary

The KDTree path in `proximity()` iterates over every chunk calling `.compute()` sequentially, accumulating all target coordinates in memory. For dense target data this can consume unbounded RAM, and the subsequent `cKDTree` construction requires all coordinates in RAM simultaneously.

## OOM Code Paths

| Line | Code | Issue |
|------|------|-------|
| `proximity.py:441` | `chunk_data = raster.data.blocks[iy, ix].compute()` | sequential `.compute()` in loop |
| `proximity.py:452` | `target_list.append(coords)` | unbounded accumulation of target coords |

For a 30 TB raster with dense target data, `target_list` could accumulate billions of `(y, x)` coordinate pairs, and `cKDTree` construction requires all of them in RAM.

**Estimated peak RAM: unbounded** — depends on target density.

## Severity

**CRITICAL** — OOM for datasets with dense target pixels.

## Suggested Fix

- Stream target coordinates to disk or use an out-of-core spatial index
- Limit KDTree construction to per-chunk or tiled regions with overlap
- Consider using a chunked nearest-neighbor approach that doesn't require a global spatial index

PR #2726, opened 2026-05-29
proximity: parallelize per-chunk-row compute in dask KDTree count pass

### LINKED ISSUE #2724: proximity: serialized per-chunk .compute() in dask KDTree target-count pass
## Summary

The unbounded dask KDTree path in `proximity()` (and `allocation`/`direction`) runs a Phase-0 streaming pass that counts target pixels chunk by chunk. The pass is a nested Python loop that calls `.compute()` on one chunk at a time:

```python
# xrspatial/proximity.py, _stream_target_counts (~L688)
for iy in range(n_tile_y):
    for ix in range(n_tile_x):
        chunk_data = raster.data.blocks[iy, ix].compute()
        ...
```

Each `.compute()` is a separate synchronous scheduler round-trip, so the chunk reads run strictly one after another. There is no parallelism, even when the scheduler is threaded or distributed and could read several chunks at once. This pass runs on every unbounded dask proximity/allocation/direction call.

## Why it is serialized

`blocks[iy, ix].compute()` builds and runs a one-chunk graph, blocks until it returns, then moves to the next chunk. The scheduler never sees more than one chunk of work at a time, so reads can't overlap.

## Measurement

On an 8x8 (64-chunk) grid with 3 ms of simulated per-chunk read latency:

| Pattern | Wall time |
|---------|----------:|
| Sequential `.compute()` per chunk (current) | ~350 ms |
| Row-batched `dask.compute(*row_blocks)` (one call per chunk-row) | ~57 ms |

So roughly 6x on this grid, and the gap grows with chunk count and read latency. Target counts come out identical between the two.

## Constraint

The streaming design exists to satisfy #879: it must not pull the whole array into memory at once. So the fix should batch per chunk-row, not over all blocks. One `dask.compute()` per chunk-row keeps peak driver memory bounded to a single row of chunks (sqrt(N) chunks) while letting the scheduler read that row in parallel. Same shape as the polygonize row-batching fix in #2608.

## Scope

Change only the loop structure in `_stream_target_counts`. The per-chunk target counting, the 25% memory cache budget, and the cache eviction behavior stay the same.

## Severity

MEDIUM. Performance, not correctness. The memory guards from #879 stay in place; this is about read parallelism on the count pass.

brendancol:
Closes #2724.

The unbounded dask KDTree path in `proximity`/`allocation`/`direction` runs a Phase-0 pass that counts target pixels chunk by chunk. That pass called `raster.data.blocks[iy, ix].compute()` one chunk at a time inside a nested Python loop, so the reads ran strictly serially even when the scheduler could have read several chunks at once.

## Change

- `_stream_target_counts` now issues one `da.compute()` per chunk-row instead of one blocking `.compute()` per chunk. The scheduler reads each row in parallel, and peak driver memory stays bounded to a single row of chunks, so the streaming guarantee from #879 still holds. Same shape as the polygonize row-batch in #2608.
- Per-chunk counting, the 25% memory cache budget, and cache eviction are unchanged.

## Measurement

64-chunk grid, 3 ms/chunk simulated read latency: ~350 ms sequential vs ~57 ms row-batched (~6x). Target counts and caches come out byte-identical.

## Backends

Affects the dask+numpy and dask+cupy unbounded KDTree path (dask+cupy routes through this after `.get()`). numpy and cupy single-array paths are untouched. Verified the dask+cupy result still matches the numpy baseline end-to-end on this CUDA host.

## Test plan

- [x] `xrspatial/tests/test_proximity.py` passes (71 tests)
- [x] New: structural test that `.compute()` is not called in the inner chunk loop
- [x] New: behavioural test that `target_counts`/`total`/caches match a sequential baseline
- [x] dask+cupy unbounded run matches numpy baseline


brendancol (commented):
## PR Review: proximity: parallelize per-chunk-row compute in dask KDTree count pass

None.

None.

- `test_stream_target_counts_matches_sequential_baseline` recomputes every chunk three times (once inside `_stream_target_counts`, then twice more in the two assertion loops). Cheap for a 24x30 raster, so fine as-is; flagging it in case the fixture grows.

### What looks good
- The fix is the right shape. `da.compute(*(blocks for ix in row))` returns results in argument order, so `row_chunks[ix]` lines up with chunk `(iy, ix)` and everything downstream stays untouched. Counting, the 25% cache budget, and eviction are unchanged.
- Memory stays bounded. Peak goes from one chunk to one chunk-row (sqrt(N) chunks), which is the bound the #879 streaming design needs. The code comment calls this out.
- Good test pairing: the AST check that `.compute()` no longer sits in the inner loop catches a regression to the old pattern, and the baseline test confirms counts and caches match the sequential version exactly.
- dask+cupy was checked end-to-end against the numpy baseline (it reaches this pass after `.get()`), so both dask backends are covered.

### Checklist
- [x] Algorithm matches reference/paper (no algorithm change; read scheduling only)
- [x] All implemented backends produce consistent results (dask+numpy and dask+cupy verified)
- [x] Edge cases are covered by tests (no-targets and all-targets paths already covered; baseline test added)
- [x] Dask chunk boundaries handled correctly (per-row batching preserves the memory bound)
- [x] No premature materialization or unnecessary copies (peak bounded to one chunk-row)
- [x] Benchmark exists or is not needed (benchmarks/benchmarks/proximity.py exists; internal change)
- [x] README feature matrix updated (not applicable; no API change)

PR #2728, opened 2026-05-29
proximity/allocation/direction: consistent output dtype and .name across backends

### LINKED ISSUE #2723: proximity/allocation/direction: output dtype and .name differ across backends
**Describe the bug**

`proximity`, `allocation`, and `direction` in `xrspatial/proximity.py` return different output metadata depending on the array backend. Two mismatches:

1. Output dtype depends on the backend. With a finite `max_distance`, the dask+numpy path returns a DataArray whose declared dtype is `float64`, while numpy, cupy, dask+cupy, and the unbounded dask+numpy KDTree path all return `float32`. The dask chunk function actually produces `float32` data, so the `float64` label is wrong both against the real data and against every other backend. All three functions document `dtype=float32` in their examples. The cause is in `_process_dask`: it calls `da.map_overlap(..., meta=np.array(()))`, and that empty array defaults to float64.

2. `.name` leaks an internal dask token. On dask backends the returned DataArray's `.name` comes out as a dask op name like `_trim-<hash>`, `_kdtree_chunk_fn-<hash>`, or `asarray-<hash>`. On numpy and cupy it's `None`. Downstream code that keys on `.name` or serializes it gets backend-dependent junk.

**Expected behavior**

For a given input, all four backends (numpy, cupy, dask+numpy, dask+cupy) should return the same dtype (`float32`) and the same `.name`, for both bounded and unbounded `max_distance`. attrs, coords, and dims are already preserved consistently and should stay that way.

**Additional context**

Found during a metadata-propagation audit of the proximity module. The computed values are correct; only the declared metadata is wrong, so this only bites downstream code that inspects dtype or name rather than the array values.

brendancol:
Closes #2723

`proximity`, `allocation`, and `direction` returned backend-dependent output metadata. This makes all four backends agree.

- dtype: the bounded dask+numpy path declared `float64` output meta via `da.map_overlap(..., meta=np.array(()))`, even though the chunk function returns `float32` and every other backend/path returns `float32`. Set the meta dtype to `float32`.
- `.name`: dask backends leaked an internal dask op name (`_trim-<hash>`, `_kdtree_chunk_fn-<hash>`, `asarray-<hash>`) into the result `.name`, while numpy/cupy return `None`. Reset `.name` to `None` after building the result so all backends match. (xarray ignores a `name=None` kwarg when the data is a named dask array, so the reset has to happen after construction.)

attrs, coords, and dims were already preserved and stay that way.

Verified end-to-end on all four backends, both bounded and unbounded `max_distance`, for all three functions.

Test plan:
- [x] New parametrized regression test asserts declared dtype `float32` and `.name is None` across all four backends, both bounded and unbounded, for proximity/allocation/direction.
- [x] Full `test_proximity.py` suite passes (93 tests).


brendancol (commented):
## PR Review: consistent output dtype and .name across backends

None.

None.

- test_proximity.py:299 - the new test asserts `result.attrs == test_raster.attrs`, but the `test_raster` fixture has empty attrs, so that line only checks empty == empty. It still guards against attrs being dropped, and the existing checks already cover attr preservation with the same fixture. Fine as-is; noting it only.

### What looks good
- The dtype fix hits the exact root cause. The bounded dask+numpy `map_overlap` declared float64 meta while the chunk function returns float32. `meta=np.array((), dtype=np.float32)` lines the declared dtype up with the real data and with the other three backends.
- The `.name = None` reset is the right call. xarray ignores a `name=None` kwarg when the data is a named dask array, so resetting after construction is the only thing that clears the leaked dask op name. The inline comment says why.
- Changes are surgical: three identical two-line resets plus one meta dtype, nothing else touched.
- The regression test covers the bounded path (max_distance=10) and the unbounded/KDTree path (np.inf), across all four backends and all three functions. It checks the declared (pre-compute) dtype rather than the post-compute dtype, which is what general_output_checks already covers and what let this bug through.

### Checklist
- [x] Output dtype consistent across backends (float32)
- [x] .name consistent across backends (None)
- [x] attrs / coords / dims preserved (unchanged by this PR)
- [x] Bounded and unbounded dask paths both covered by tests
- [x] No benchmark needed (metadata-only fix)
- [x] No README/docstring change needed (no API change)

PR #2729, opened 2026-05-29
proximity: style cleanup (flake8 F841/E128, isort, mutable default)

### LINKED ISSUE #2725: Style cleanup in xrspatial/proximity.py (flake8 F841/E128, isort, mutable default)
A style audit of `xrspatial/proximity.py` using the project's configured flake8 (`max-line-length=100`) and isort (`line_length=100`) settings turned up four issues. None of them break correctness on their own, but two touch runtime semantics, so it makes sense to fix them in one pass.

### flake8 F841: dead local (line 1274)

`original_chunks = raster.data.chunks` is assigned in the unbounded dask+cupy branch and never read. It is a refactor leftover from when chunk geometry was restored after the cupy-to-numpy conversion. Remove it.

### Mutable default argument (lines 1347, 1490, 1630)

The three public functions `proximity`, `allocation`, and `direction` declare `target_values: list = []`. A mutable default is shared across all calls. The current code never mutates it, so there is no live bug, but the pattern is fragile. Switch to a `None` sentinel and normalize to `[]` inside the body so behaviour stays identical.

### flake8 E128: continuation line under-indent (line 291)

The nested `np.where` in `_vectorized_calc_direction` is under-indented for the visual indent. Reflow it.

### isort: import ordering (lines 27-32)

The three `from xrspatial...` imports are not in the order isort expects under `line_length=100`. Re-sort.

### Scope

- No behavioural change for the E128 reflow or the isort reordering.
- The F841 removal deletes dead code only.
- The mutable-default fix preserves behaviour (None normalizes to `[]`).
- No flake8 or isort config changes, no autoformatter.

After the fix, `flake8 xrspatial/proximity.py` and `isort --check-only xrspatial/proximity.py` should both pass clean.

brendancol:
Closes #2725

Style sweep cleanup for `xrspatial/proximity.py`. Four findings from flake8 (`max-line-length=100`) and isort (`line_length=100`), bundled into one pass.

### What changed

- **Cat 3 (F841):** removed the dead local `original_chunks` in the unbounded dask+cupy branch. It was assigned and never read, a leftover from when chunk geometry used to be restored after the cupy-to-numpy conversion.
- **Cat 5 (mutable default):** `proximity`, `allocation`, and `direction` declared `target_values: list = []`. Switched to a `None` sentinel and normalize to `[]` at the top of each function body. The old default was never mutated, so this preserves behaviour while dropping the shared-mutable-default anti-pattern.
- **Cat 1 (E128):** reflowed the under-indented `np.where` continuation in `_vectorized_calc_direction`.
- **Cat 4 (isort):** re-sorted the `xrspatial` import block and added the blank line isort wants after the inline `import cupy as cp`.

After the change, `flake8 xrspatial/proximity.py` and `isort --check-only xrspatial/proximity.py` both pass clean.

### Behaviour

No behavioural change intended. The E128 reflow and isort reordering are formatting only, the F841 removal deletes dead code, and the mutable-default fix is behaviour-preserving (None normalizes to `[]`).

### Backends

Static source-level changes, uniform across numpy / cupy / dask+numpy / dask+cupy. No backend-specific logic touched.

### Test plan

- [x] `flake8 xrspatial/proximity.py` clean
- [x] `isort --check-only xrspatial/proximity.py` clean
- [x] Existing suite: `pytest xrspatial/tests/test_proximity.py` (69 passed)
- [x] Added regression test: `None` default matches explicit `[]` for all three public functions

Note: `test_proximity.py` carries two pre-existing E127 warnings (lines 726, 752) and isort import drift that predate this PR. Left untouched to keep the diff scoped to the audited module.


brendancol (commented):
## PR Review: proximity style cleanup (flake8 F841/E128, isort, mutable default)

None.

None.

- The annotation is now `target_values: list = None`, which really describes an `Optional[list]`. The strictly correct form would be `Optional[list]`, but `list = None` is common shorthand in this codebase and the body normalizes immediately, so it is fine as-is. Not worth changing unless the project decides to standardize on `Optional`.

### What looks good
- The F841 removal is safe: `original_chunks` had a single occurrence and was never read.
- The mutable-default fix preserves behaviour. The old `[]` default was never mutated, and `None` normalizes to `[]` at the top of each public function before any use.
- The E128 reflow and isort reordering are formatting only, confirmed clean by both linters.
- The regression test covers all three public functions via parametrize and asserts the None default matches an explicit empty list.
- The diff stays scoped to the audited module. Pre-existing lint in the test file was correctly left alone.

### Checklist
- Algorithm unchanged (no logic touched)
- All backends: static source change, uniform across numpy/cupy/dask
- NaN handling unchanged
- Edge cases: existing 69 tests pass, new regression test added
- Dask chunk boundaries not touched
- Benchmark not needed (no perf change)
- README feature matrix not applicable (no new function)
- Docstrings present and unchanged

PR #2736, opened 2026-05-30
focal: honour boundary on the cupy backend (#2730)

### LINKED ISSUE #2730: focal mean/apply/focal_stats ignore boundary on the cupy backend
## Description

`mean()`, `apply()`, and `focal_stats()` ignore the `boundary` parameter on
the single-GPU CuPy backend. The numpy and dask backends honor
`boundary='nearest'`, `'reflect'`, and `'wrap'`, but the CuPy paths always
behave as `boundary='nan'` (edge clamping).

The CuPy dispatch functions in `xrspatial/focal.py` never receive the
`boundary` argument:

- `mean()` -> `cupy_func=_mean_cupy`
- `apply()` -> `cupy_func=_apply_cupy`
- `focal_stats()` -> `cupy_func=_focal_stats_cupy`

Each is registered in the `ArrayTypeFunctionMapping` without a `boundary`
partial, unlike the matching numpy and dask entries. The GPU kernels clamp
the neighbourhood window to the raster edge, which matches `boundary='nan'`
semantics no matter what the caller passed.

## Reproduce

```python
from xrspatial import mean

data = np.random.default_rng(0).random((6, 6))
np_agg = xr.DataArray(data, dims=['y', 'x'])
cp_agg = xr.DataArray(cupy.asarray(data), dims=['y', 'x'])

for b in ['nan', 'nearest', 'reflect', 'wrap']:
    n = mean(np_agg, boundary=b).data
    c = mean(cp_agg, boundary=b).data.get()
    print(b, np.allclose(n, c, equal_nan=True, rtol=1e-4))
```

Output:

```
nan True
nearest False
reflect False
wrap False
```

The same divergence occurs for `apply()` and `focal_stats()`.

## Expected

For a given `boundary` value, all four backends (numpy, cupy, dask+numpy,
dask+cupy) should produce matching results on identical input.

## Fix

Pad the CuPy input per the boundary mode (reuse `_pad_array`) before the GPU
kernel and trim the result, mirroring the numpy boundary path in
`_mean_numpy_boundary` and `_apply_numpy_boundary`.

brendancol:
Closes #2730

## What

The single-GPU cupy backend of `mean()`, `apply()`, and `focal_stats()`
ignored the `boundary` parameter. It always behaved as `boundary='nan'`
(edge clamping), while numpy and dask honoured `'nearest'`, `'reflect'`,
and `'wrap'`. The cupy dispatch functions were registered in the
`ArrayTypeFunctionMapping` without a `boundary` partial.

## Change

- Add `_mean_cupy_boundary`, `_apply_cupy_boundary`, and
  `_focal_stats_cupy_boundary` wrappers that pad the input per the
  boundary mode with `_pad_array` and trim the result, mirroring the
  existing numpy boundary path. The `'nan'` mode is unchanged.
- The `focal_stats` wrapper rebuilds the output DataArray so the
  original coords, dims, and attrs are preserved after trimming.

## Backend coverage

numpy and dask were already correct. This fixes the cupy path. dask+cupy
already routed boundary through `map_overlap`. All four backends now match
for a given `boundary` value.

## Test plan

- [x] `mean`/`apply`/`focal_stats` cupy results match numpy across
      `nan`/`nearest`/`reflect`/`wrap` (new parametrized tests)
- [x] cupy `focal_stats` preserves coords/dims/attrs under non-nan boundary
- [x] full `test_focal.py` suite: 129 passed (CUDA host)


brendancol (commented):
## PR Review: focal honours boundary on the cupy backend (#2730)

None.

None.

- `xrspatial/focal.py`: `_apply_cupy_boundary` and `_focal_stats_cupy_boundary` repeat the `r0/r1/c0/c1` trim slicing that already lives in `_apply_numpy_boundary`. A small shared `_trim(result, pad_h, pad_w)` helper would drop the duplication. I'd leave it as-is though, since the current form matches the existing numpy code and a one-off helper for three call sites is a wash.

### What looks good
- The three cupy wrappers follow the existing `_mean_numpy_boundary` / `_apply_numpy_boundary` shape, so the boundary semantics line up with numpy without a separate code path to keep in sync.
- `boundary='nan'` short-circuits to the old code, so the default path is untouched.
- `_focal_stats_cupy_boundary` rebuilds the output DataArray and reattaches the original coords and attrs after trimming. The new coord-preservation test covers that.
- Tests check cupy against numpy across all four boundary modes for mean, apply, and focal_stats, run on a real CUDA host.

### Checklist
- [x] Algorithm matches the numpy boundary reference (pad then trim)
- [x] All implemented backends produce consistent results (numpy vs cupy verified across modes)
- [x] NaN handling correct ('nan' mode unchanged)
- [x] Edge cases covered (1x1 kernel pads 0, trims to full array via the `if pad else None` guard)
- [x] Dask chunk boundaries unchanged (dask+cupy already forwarded boundary)
- [x] Docstrings accurate (boundary already documented on each public function)

PR #2737, opened 2026-05-30
Style cleanup in focal.py: unused import, isort, mutable default (#2731)

### LINKED ISSUE #2731: Style cleanup in focal.py: unused import, isort drift, mutable default arg
## What

Three style issues in `xrspatial/focal.py`, all flagged by the project's own `setup.cfg` config (flake8, isort) plus a grep for mutable default arguments. None of them change behaviour.

**1. Unused import (flake8 F401)**

Line 36 imports `not_implemented_func` from `xrspatial.utils` and never uses it. There's no `__all__` and nothing imports it from `xrspatial.focal`, so it's not a re-export -- just a leftover from an earlier refactor.

**2. Import ordering (isort)**

The import block at the top of the file doesn't match the project's isort config (`line_length=100`). The stdlib imports are interleaved (`import copy` / `from functools import partial` / `from math import isnan` / `import math`), there's a stray blank line inside a grouped block, and the `from`-imports for `xrspatial.convolution` and `xrspatial.utils` aren't alphabetised or wrapped to the configured width. The `xrspatial.dataset_support` import is also out of order.

**3. Mutable default argument**

`mean(agg, passes=1, excludes=[np.nan], ...)` uses a list literal as a default. The list is never mutated here -- it's only iterated and passed through `tuple()` -- so behaviour is correct today. But a shared mutable default is a footgun, and the style sweep flags it. Fix is the usual sentinel: default to `None`, assign `[np.nan]` in the body. Same behaviour, plus a regression test that the default doesn't leak across calls.

## Scope

`xrspatial/focal.py` and a test in `xrspatial/tests/test_focal.py`. No config widening, no autoformatter.

brendancol:
Closes #2731

Style-only cleanup of `xrspatial/focal.py`, flagged by the project's own `setup.cfg` (flake8, isort) and a mutable-default grep. From the deep-sweep style pass on the focal module.

- Cat 3 (F401): removed the unused `not_implemented_func` import. No `__all__`, nothing imports it from `xrspatial.focal`, so it wasn't a re-export.
- Cat 4 (isort): reordered and rewrapped the top-of-file import block to `line_length=100`. Mechanical, no functional change.
- Cat 5 (mutable default): `mean(..., excludes=[np.nan], ...)` becomes `excludes=None`, resolved to `[np.nan]` in the body. The list was never mutated, so behaviour is identical. Added `test_mean_default_excludes_does_not_leak`.

No behavioural change is intended for the Cat 3/4 fixes. The Cat 5 fix is behaviour-preserving. No config widening, no autoformatter.

Backends: static source-level changes, uniform across numpy / cupy / dask+numpy / dask+cupy. `mean` dispatches through `ArrayTypeFunctionMapping`; no path-specific logic touched.

Test plan:
- [x] `flake8 xrspatial/focal.py` clean
- [x] `isort --check-only xrspatial/focal.py` clean
- [x] 115 focal tests pass (incl. GPU/dask mean tests, CUDA available)
- [x] new regression test passes


brendancol (commented):
Domain-aware review (deep-sweep style worker)

Scope check: the diff is style-only and matches the three findings in #2731.

Correctness:
- The `excludes is None -> [np.nan]` resolution sits at the top of `mean()`, ahead of both the 3D `_apply_per_band` branch and the 2D loop, so both paths get the intended default. Behaviour is unchanged from the old `excludes=[np.nan]` default.
- isort output wraps the `xrspatial.convolution` and `xrspatial.utils` from-imports onto continuation lines at the 100-char limit; flake8 is clean on the result (no E128/E501), so the wrapping is consistent with the project config.
- `not_implemented_func` removal is safe: no `__all__`, and nothing imports it from `xrspatial.focal`.

Backends: changes are static source edits and apply uniformly across numpy / cupy / dask+numpy / dask+cupy. `mean` dispatches through `ArrayTypeFunctionMapping`; no per-backend path was touched. The full focal suite (115 tests, including the GPU and dask mean tests) passes with CUDA available.

Nit (non-blocking): `test_mean_default_excludes_does_not_leak` runs on `data_random`, which has no NaNs, so it proves `None` == `[np.nan]` equivalence and no cross-call leak but does not itself exercise the NaN-exclusion path. That path is already covered by the existing `test_mean_*` transfer-function tests, so I left it as is.

No blockers, no required changes.

PR #2738, opened 2026-05-30
Add dask+cupy test coverage for focal.hotspots()

### LINKED ISSUE #2732: hotspots() dask+cupy backend has no test coverage
focal.hotspots() registers a dask_cupy_func (`_hotspots_dask_cupy`) in its dispatch table, but nothing in test_focal.py calls it. mean, apply, and focal_stats each have a dedicated dask+cupy test. hotspots is the only one without one, so a bug in its fused convolve/z-score/classify path on dask+cupy could ship unnoticed.

This is a test-only gap. I ran the path on a CUDA host and it currently matches the numpy reference exactly, so no source change is needed.

Found by the test-coverage sweep.

brendancol:
Adds `test_hotspots_dask_cupy`. The `_hotspots_dask_cupy` dispatch path was registered but had no test; this closes that gap.

The test compares the dask+cupy result against the numpy reference on the chunk interior, the same approach as `test_apply_dask_cupy` and `test_focal_stats_dask_cupy`. I checked on a CUDA host that it passes and actually routes through `_hotspots_dask_cupy`. Test-only, no source changes.

Closes #2732


brendancol:
Closing in favor of a fresh PR from the same branch. This PR's head ref detached after the branch was deleted and recreated, and GitHub would not re-point it to the resolved commit (7dc49b6a), so pushes/force-pushes here no longer sync. The conflict-resolved branch is unchanged; a replacement PR tracks it.

PR #2739, opened 2026-05-30
Keep hotspots dask+cupy classification on the GPU

### LINKED ISSUE #2734: hotspots dask+cupy backend round-trips every chunk through host memory
**Describe the bug**

The dask+cupy backend of `hotspots` (in `xrspatial/focal.py`) copies every chunk off the GPU and back. The per-chunk function `_chunk_fn` inside `_hotspots_dask_cupy` runs:

```python
def _chunk_fn(chunk):
    convolved = _convolve_2d_cupy(chunk, norm_kernel)
    z = (convolved - global_mean) / global_std
    return cupy.asarray(_calc_hotspots_numpy(cupy.asnumpy(z)))
```

`cupy.asnumpy(z)` pulls the chunk's float32 z-score array off the GPU, `_calc_hotspots_numpy` classifies it on the CPU, then `cupy.asarray(...)` pushes the int8 result back. So every chunk pays a full PCIe round trip and the classification serializes on the transfer instead of running on the GPU.

The single-GPU cupy path (`_hotspots_cupy`) does not have this problem. It uses the existing GPU kernel `_run_gpu_hotspots`, which classifies each z-score cell into its int8 hotspot code on the device.

**Expected behavior**

The dask+cupy path should classify on the GPU, reusing `_run_gpu_hotspots`, so no chunk leaves the device. Results should match the single-GPU cupy backend on the chunk interior.

**Additional context**

Found during a performance sweep of the focal module. The OOM verdict for the dask paths is SAFE -- `map_overlap` keeps memory bounded to a chunk plus its halo. This is a throughput problem specific to the dask+cupy backend, not a memory one. The fix is contained to `_hotspots_dask_cupy`.

brendancol:
Closes #2734

The dask+cupy path of `hotspots` classified each chunk on the CPU. `_chunk_fn` inside `_hotspots_dask_cupy` ran `cupy.asnumpy(z)` to pull the z-score array off the GPU, classified it with `_calc_hotspots_numpy`, then pushed the int8 result back with `cupy.asarray`. Every chunk paid a host round trip.

This reuses `_run_gpu_hotspots`, the kernel the single-GPU cupy path already uses, so classification stays on the device. Convolution, z-score, and classification now all run on the GPU within the chunk, and the per-chunk `cupy.asnumpy` round trip is gone.

Behavior is unchanged: it reuses the same kernel the single-GPU path runs, so results match.

Backends: only the dask+cupy path of `hotspots` changed. The numpy, cupy single-GPU, and dask+numpy paths are untouched.

Test plan:
- [x] Existing hotspots tests pass (numpy, dask+numpy, cupy, zero-std, NaN handling)
- [x] New `test_hotspots_dask_cupy`: result stays a cupy-backed dask array and matches the single-GPU cupy backend on the chunk interior
- [x] Full `test_focal.py` suite passes (115 tests)


brendancol (commented):
## PR Review: Keep hotspots dask+cupy classification on the GPU

None.

None.

- `_hotspots_dask_cupy` (~line 1267): `cupy.zeros_like(z, dtype=cupy.int8)` could be `cupy.empty` since `_run_gpu_hotspots` writes every in-bounds cell. It mirrors the single-GPU path (`_hotspots_cupy` line 1338 uses the same `zeros_like`), so staying consistent is the better call. Leave it.

### What looks good
- The fix reuses `_run_gpu_hotspots` the same way the single-GPU `_hotspots_cupy` path does: same `cuda_args`, same int8 `zeros_like`, same kernel launch. So the dask+cupy interior matches the single-GPU result rather than approximating it. The removed `cupy.asnumpy`/`cupy.asarray` pair was the only host hop in the chunk function.
- Classification is elementwise (each z cell maps to its own int8 code), so classifying per chunk-with-halo and trimming via `map_overlap` gives the same answer as classifying the whole array at once.
- `test_hotspots_dask_cupy` checks both that the result stays a cupy-backed dask array (`_meta` is a cupy ndarray) and that the interior matches the single-GPU backend exactly, using the same interior-trim pattern as `test_apply_dask_cupy` and `test_focal_stats_dask_cupy`.
- `_calc_hotspots_numpy` is still used by the numpy and dask+numpy paths, so nothing is left orphaned.

### Checklist
- [x] Algorithm matches reference (reuses existing GPU kernel; behavior unchanged)
- [x] All implemented backends produce consistent results (new test asserts dask+cupy == cupy interior)
- [x] NaN handling is correct (unchanged; `_run_gpu_hotspots` classifies z-scores, NaN z stays 0)
- [x] Edge cases covered (existing zero-std and NaN hotspots tests pass)
- [x] Dask chunk boundaries handled correctly (depth = kernel//2, boundary forwarded, elementwise classify)
- [x] No premature materialization or unnecessary copies (removes the per-chunk host round trip)
- [x] Benchmark exists (benchmarks/benchmarks/focal.py)
- [ ] README feature matrix - not applicable (no new function, no backend support change)
- [x] Docstrings present and accurate (no API change)

PR #2741, opened 2026-05-30
aspect: planar dask backends report float32 dtype matching computed data

### LINKED ISSUE #2682: aspect() planar dask backends report float64 dtype but compute float32
## Description

The planar dask backends of `aspect()` (and by extension `northness()` /
`eastness()`) advertise a lazy dtype of `float64`, but the array they
actually compute is `float32`. The numpy and cupy backends correctly
report `float32`.

This is a backend-inconsistent metadata bug. Code that reads `agg.dtype`
to make a decision (buffer allocation, storage class, a dtype assertion)
gets a different answer on the dask backends than on numpy/cupy, and the
dtype silently flips from `float64` to `float32` once `.compute()` runs.

## Root cause

In `xrspatial/aspect.py`, the planar dask functions call `map_overlap`
with a default-dtype meta:

- `_run_dask_numpy`: `meta=np.array(())` (defaults to float64)
- `_run_dask_cupy`: `meta=cupy.array(())` (defaults to float64)

But the chunk functions `_cpu` / `_run_cupy` cast to and return float32.
Dask trusts `meta` for the advertised dtype, so the lazy DataArray reports
float64 while the realized data is float32. The geodesic dask paths already
pass `meta=np.array((), dtype=np.float32)`, so they are fine. Only the
planar paths are affected.

## Reproduction

```python
from xrspatial import aspect

data = np.random.rand(8, 10).astype(np.float64) * 100
arr = xr.DataArray(da.from_array(data, chunks=(4, 5)),
                   dims=['y', 'x'],
                   coords={'y': np.arange(8), 'x': np.arange(10)})
out = aspect(arr)
print(out.dtype)            # float64  (advertised)
print(out.compute().dtype)  # float32  (realized)
```

Expected: both print `float32`, matching the numpy and cupy backends.

## Fix

Pass an explicit `dtype` to the planar dask `meta`: `np.float32` for the
numpy path and `cupy.float32` for the cupy path, mirroring what the
geodesic dask paths already do.

## Note

`slope.py` and `curvature.py` share the same default-dtype meta pattern on
their planar dask paths. They are out of scope here but likely have the
same inconsistency.

Found via metadata-propagation sweep.

brendancol:
Closes #2682

## What this fixes

The planar dask backends of `aspect()` advertised a lazy dtype of
`float64`, but the array they actually computed was `float32`. The numpy
and cupy backends already reported `float32`, so the dtype you saw
depended on which backend was active, and it silently changed from
`float64` to `float32` once `.compute()` ran.

Root cause: `_run_dask_numpy` and `_run_dask_cupy` called `map_overlap`
with a default-dtype meta (`np.array(())` / `cupy.array(())`), which
defaults to float64. The chunk functions `_cpu` / `_run_cupy` return
float32. The fix passes an explicit float32 dtype to the meta, matching
what the geodesic dask paths already do.

`northness()` and `eastness()` derive from `aspect()`, so they inherit
the corrected dtype.

## Backend coverage

- numpy: already float32, unchanged
- cupy: already float32, unchanged
- dask+numpy: was lazy-float64 / computed-float32, now float32 both
- dask+cupy: was lazy-float64 / computed-float32, now float32 both

## Test plan

- [x] New `test_dask_numpy_advertised_dtype_matches_computed` asserts lazy
      and computed dtype are both float32
- [x] New `test_dask_cupy_advertised_dtype_matches_computed` covers the
      GPU dask path
- [x] Full `test_aspect.py` suite passes (66 passed, cupy + dask+cupy
      included on a CUDA host)

## Note

`slope.py` and `curvature.py` use the same default-dtype meta pattern on
their planar dask paths and likely have the same inconsistency. Out of
scope for this PR (metadata sweep was scoped to aspect).


brendancol (commented):
## PR Review: aspect planar dask backends report float32 dtype matching computed data

None.

None. One optional coverage gap below.

- The two new tests only cover the default `boundary='nan'` path. The
  other boundary modes run through `_pad_array` and then hit the same
  chunk functions, so they return float32 too, but a parametrize over
  boundary would pin that down. Low value, skip if you prefer.

### What looks good
- The fix is minimal and correct. `_cpu` / `_run_cupy` already cast to
  float32 and return it, so declaring float32 in the meta is the right
  call, and it matches what the geodesic dask paths in this same file
  already do.
- numpy and cupy already reported float32. This brings the two planar
  dask paths in line, so all four backends now agree on both the lazy
  and the computed dtype.
- Both dask paths get the same treatment.
- The tests check both the lazy advertised dtype and the dtype after
  compute, which is the actual bug in #2682: the two disagreed.
- The computation itself is untouched. No change to NaN handling, depth,
  or boundary forwarding.

### Checklist
- Algorithm: unchanged, dtype-only fix
- Backend consistency: float32 verified on numpy, cupy, dask+numpy, dask+cupy
- NaN handling: unchanged
- Edge cases: dtype consistency covered; boundary-mode dtype not parametrized (low risk)
- Dask chunk boundaries: unchanged, depth=(1,1)
- No premature materialization or copies: confirmed
- Benchmark: not needed, no perf change
- Docstrings: unchanged

brendancol (commented):
## Follow-up review (after 7c9017a)

The one nit from the previous pass is addressed: `test_dask_numpy_advertised_dtype_matches_computed` is now parametrized over all four boundary modes (`nan`, `nearest`, `reflect`, `wrap`), so the `_pad_array` path is covered for dtype too. All five dtype tests pass.

PR #2745, opened 2026-05-30
Guard degenerate-axis resolution in viewshed CPU sweep

### LINKED ISSUE #2744: viewshed() returns wrong results for single-row or single-column rasters (divide-by-zero in resolution)
**Describe the bug**

`viewshed()` on a single-row or single-column raster silently returns wrong results. The CPU sweep helper `_viewshed_cpu` computes the grid resolution as:

```python
ew_res = (x_range[1] - x_range[0]) / (width - 1)
ns_res = (y_range[1] - y_range[0]) / (height - 1)
```

When `width == 1` or `height == 1`, the denominator is zero. With float coordinates this gives NaN (a divide-by-zero RuntimeWarning, not an exception). That NaN resolution then flows through every distance and gradient calculation, so every non-observer cell comes back INVISIBLE, even on flat terrain where they should all be visible. Nothing raises, so the caller gets a plausible-looking but wrong array.

The sibling helpers `_viewshed_windowed` and `_viewshed_dask` already guard this with `if width > 1 else 1.0` / `if height > 1 else 1.0`. `_viewshed_cpu` is missing the same guard. And since `_viewshed_windowed` routes a degenerate window back through `_viewshed_cpu`, the `max_distance` path is affected too.

**Reproduce**

```python
arr = np.zeros((1, 5))
r = xr.DataArray(arr, coords=dict(x=np.arange(5.0), y=np.arange(1.0)), dims=["y", "x"])
print(viewshed(r, x=2.0, y=0.0, observer_elev=5).values)
# [[ -1.  -1. 180.  -1.  -1.]]  -- every cell but the observer is INVISIBLE
```

**Expected behavior**

On flat terrain with an elevated observer, all cells should be visible. The fix is to fall back to unit resolution when an axis has length 1, matching what the windowed and dask helpers already do.

**Additional context**

Found during a robustness sweep of the viewshed module. Silent wrong output in an analysis function, so worth treating as high priority.

brendancol:
Closes #2744

`_viewshed_cpu` computed grid resolution as `range / (dim - 1)`. For a single-row or single-column raster that is `0 / 0 = NaN`, which then flowed through every distance and gradient calculation and silently marked every non-observer cell INVISIBLE on flat terrain (no error raised).

- Fall back to unit resolution when an axis has length 1, matching the guard already present in `_viewshed_windowed` and `_viewshed_dask`.
- The `max_distance` path routes a degenerate window back through `_viewshed_cpu`, so it is covered by the same fix.

Backend coverage: numpy (direct fix). The dask and cupy/RTX entry points already guarded this, so their behavior is unchanged; the windowed path that delegates to the CPU helper is now correct on degenerate inputs.

Test plan:
- [x] `test_viewshed_single_row_or_column` (1x5 and 5x1) — observer cell is 180, all cells visible, no NaN
- [x] `test_viewshed_single_row_or_column_max_distance` (1x7 and 7x1) — windowed path correct
- [x] Full `test_viewshed.py` suite passes (29 passed)


brendancol (commented):
## PR Review: Guard degenerate-axis resolution in viewshed CPU sweep

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The fix matches the guard already used in `_viewshed_windowed` (viewshed.py:2078-2079) and `_viewshed_dask` (viewshed.py:2180-2181), so the three CPU-resolution computations now agree.
- Falling back to `1.0` for a length-1 axis is the right call: a single row has no defined y-resolution, and unit resolution keeps the distance/gradient geometry well-defined.
- Tests cover both orientations (1xN and Nx1) and both code paths (full sweep and `max_distance` windowing, which delegates back to `_viewshed_cpu`). The assertions check the real failure mode: observer cell at 180, all other cells visible, and no NaN in the output.
- I checked the 1x1 case by hand (both axes degenerate): it returns `[180.]` without error, so the two guards compose correctly.

### Checklist
- [x] Algorithm matches reference/convention (mirrors sibling helpers)
- [x] NaN handling is correct (the fix removes the NaN source)
- [x] Edge cases covered by tests (1xN, Nx1; 1x1 verified manually)
- [x] README feature matrix not applicable (no API change)

PR #2747, opened 2026-05-30
viewshed: consistent output name and dtype across backends

### LINKED ISSUE #2743: viewshed() output .name and dtype differ across backends
**Describe the bug**

`viewshed()` returns a DataArray whose `.name` and dtype depend on which backend ran, even when the input is identical. For an input DataArray named `elevation`:

- numpy: `.name` is `None`, dtype `float64`
- cupy (RTX): `.name` is hardcoded to `"viewshed"`, dtype `float32`
- cupy + `max_distance`: `.name` is `None`, dtype `float64`
- dask+numpy / dask+cupy (in-memory tier): `.name` is the dask graph token (e.g. `array-ffa3...`, `concatenate-be6...`), and dtype is `float32` when the GPU tier ran, else `float64`

There are two distinct problems:

1. The output name is not set consistently. The CPU path leaves it `None`, the RTX path hardcodes `"viewshed"`, and the dask paths leak the internal dask graph token because the output DataArray is built without an explicit `name=`.
2. The output dtype is `float32` on the GPU/RTX exact path but `float64` everywhere else.

Sibling modules (slope, curvature, hillshade) already take a `name=` parameter, pass it through to the output, and keep dtype consistent across numpy and cupy. `viewshed()` does neither.

**Expected behavior**

For the same input, `viewshed()` should return a DataArray with the same `.name` and the same dtype no matter which backend ran. Matching the sibling modules: add a `name=` parameter (default `None`), set the output name from it on every backend, and return `float64` on all backends.

**Additional context**

Found by the /deep-sweep metadata propagation sweep on the viewshed module. Severity MEDIUM (Cat 5, backend-inconsistent metadata). attrs, coords, and dims are preserved correctly on every backend; only `.name` and output dtype drift.

brendancol:
Closes #2743

`viewshed()` returned a DataArray whose `.name` and dtype depended on which backend ran:

- `.name`: numpy left it `None`, the RTX path hardcoded `"viewshed"`, and the dask paths leaked the internal dask graph token (built without an explicit `name=`).
- dtype: `float32` on the GPU/RTX exact path, `float64` everywhere else.

Changes:

- Add a `name=` parameter (default `'viewshed'`), threaded through every backend helper and set on the output. Matches the convention in slope, curvature, and hillshade.
- Force `float64` output on the GPU and dask Tier B paths so dtype no longer depends on the backend (the RT kernel still works in float32 internally).

attrs, coords, and dims were already preserved on every backend; this PR does not change them.

Backend coverage: numpy, cupy (RTX), dask+numpy, dask+cupy. Verified live on a host with CUDA + RTX.

Test plan:

- [x] `test_viewshed_output_name_and_dtype_consistent` — name == 'viewshed' and dtype == float64 across all four backends, with and without `max_distance`
- [x] `test_viewshed_custom_name` — a user-supplied name is honoured on every backend
- [x] Full `test_viewshed.py` suite (37 passed)
- [x] `test_visibility.py`, `test_min_observable_height.py`, `test_gpu_rtx_memory.py`, `test_validation.py` (120 passed)


brendancol (commented):
## PR Review: viewshed consistent output name and dtype across backends

None.

None.

- The docstring lists the `name` type as `str`, but the signature is `Optional[str]` and `None` is a valid value (it produces an unnamed DataArray). slope.py has the same `str` vs `Optional[str]` docstring wording, so this matches the rest of the codebase. Not worth changing on its own.

### What looks good
- The `name=` default of `'viewshed'` is the same string the RTX path used to hardcode, and it follows the sibling-module convention (slope, curvature, hillshade all default to their own function name).
- `name=` reaches every backend helper (`_viewshed_cpu`, `_viewshed_windowed`, `_viewshed_dask`, `viewshed_gpu`, `_viewshed_rt`) and is set at DataArray construction. Verified live: numpy, cupy/RTX, dask+numpy, and dask+cupy all come back with `name='viewshed'` (or the custom name). Passing `name=` at construction does override the dask graph token here, so the dask paths no longer leak it.
- The float64 cast widens from float32, so values do not change. The numpy/cupy value differences are the pre-existing algorithm difference already noted in the docstring, not something this PR introduced.
- Tests cover all four backends with and without `max_distance`, plus a custom-name case. cupy and dask+cupy are gated on `has_rtx()`.

### Checklist
- [x] Algorithm unchanged (metadata-only change)
- [x] All four backends produce consistent name and dtype
- [x] NaN handling unchanged
- [x] Edge cases: existing suite still passes (37 tests)
- [x] Dask chunk handling unchanged
- [x] Docstring present for the new parameter

PR #2749, opened 2026-05-31
contours: handle infinite nan coords

### LINKED ISSUE #2704: contours() emits NaN coordinates for rasters containing +inf/-inf

Melissari1997:
## Summary
Closes #2704 
This PR fixes the contour generation to properly handle input data containing infinite values. The marching squares algorithm now skips quadrilaterals with non-finite corners instead of producing invalid output or raising errors.

**Key change:**
- **contour**: Skip quads with any non-finite (NaN, +inf, -inf) corner values during contour line generation

## AI-assisted review

- [x] I ran the xarray-spatial AI-assisted review workflow for this PR.
- [x] I reviewed the AI output myself and made my own judgment about what to accept.
- [x] I understand this change and can maintain it.
- [x] This PR does not attribute authorship to an AI tool or service.

## AI-assisted review notes

The xarray-spatial review workflow was run against this change.

**Issues found and addressed:**
- Added regression tests for contour handling of infinite values (`np.inf`, `-np.inf`)
- Verified no change to public API behavior
- Confirmed existing NaN handling remains correct

**Open concerns:**
- None identified

## Tests

Added tests in `test_contour.py` covering:
- Contour generation with `np.inf` values in input arrays
- Contour generation with `-np.inf` values in input arrays
- Mixed finite and infinite value scenarios

### DIFF
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -66,8 +66,9 @@ def _marching_squares_kernel(data, level, seg_rows, seg_cols, seg_count):
             bl = data[r + 1, c]
             br = data[r + 1, c + 1]
 
-            # Skip quads with any NaN corner.
-            if tl != tl or tr != tr or bl != bl or br != br:
+            # Skip quads with any non-finite corner (NaN or +/-inf).
+            if not (np.isfinite(tl) and np.isfinite(tr) and
+                    np.isfinite(bl) and np.isfinite(br)):
                 continue
 
             # Build 4-bit case index.
--- a/xrspatial/tests/test_contour.py
+++ b/xrspatial/tests/test_contour.py
@@ -133,6 +133,65 @@ class TestNaNHandling:
 
 
 # ---------------------------------------------------------------------------
+# Inf handling
+# ---------------------------------------------------------------------------
+
+class TestInfHandling:
+
+    def test_pos_inf_center(self):
+        """A quad touching +inf must not produce NaN coordinates."""
+        data = np.array([
+            [0., 0., 0., 0., 0.],
+            [0., 1., 1., 1., 0.],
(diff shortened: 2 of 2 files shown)


brendancol (approved):

Melissari1997:
Apparently my fork wasn't in sync with the official repo

PR #2773, opened 2026-06-01
Propagate NaN center cell in planar slope

### LINKED ISSUE #2761: Planar slope returns a valid value for a NoData (NaN) center cell
**Describe the bug**

Planar slope can return a finite slope for an input NoData (NaN) cell. The `_cpu()` kernel in `xrspatial/slope.py` reads the 8 surrounding neighbours but never looks at the center cell `data[y, x]`. So when the center is NaN but its 8 neighbours are valid, the slope formula returns a real number instead of NaN. The GPU planar path (`_gpu` / `_run_gpu`) has the same gap.

This bites downstream hydrology: a hole in a DEM silently turns into "valid flat terrain", which throws off flow routing and fill.

The geodesic method doesn't have this problem because it includes the center cell in its 9-cell plane fit, so a NaN center comes back as NaN.

**Steps to reproduce**

```python
from xrspatial import slope

data = np.zeros((5, 5), dtype=np.float64)
data[2, 2] = np.nan          # single hole in an otherwise flat raster
agg = xr.DataArray(data, attrs={'res': (1, 1)})
result = slope(agg)
print(result.data[2, 2])     # prints 0.0, expected nan
```

**Expected behavior**

A NaN center cell should yield NaN at that cell, matching the geodesic method and normal NoData semantics.

**Additional context**

The fix needs to cover both the planar CPU path (`_cpu` / `@ngjit`) and the planar GPU path (cupy/cuda), pinned with regression tests across the supported backends.

brendancol:
Closes #2761

## What

- The planar `_cpu` and `_gpu` slope kernels read the 8 surrounding cells but never the center, so a NaN hole with valid neighbours returned a finite slope (a flat 5x5 raster with only the center NaN gave 0.0 at the center).
- Add a center-cell NaN check to both kernels so a NaN center comes back as NaN, matching the geodesic method and normal NoData semantics.

## Backend coverage

The check lives in the shared `_cpu` (@ngjit) and `_gpu` (cuda device) kernels, so it covers all four planar paths: numpy, cupy, dask+numpy, dask+cupy.

## Test plan

- [x] `test_center_nan_propagates_numpy` - NaN center stays NaN
- [x] dask+numpy / cupy / dask+cupy variants compare against numpy with `equal_nan=True`
- [x] existing `test_slope.py` suite still green (65 passed locally, GPU included)

### DIFF
--- a/xrspatial/slope.py
+++ b/xrspatial/slope.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 # std lib
 from functools import partial
-from math import atan
+from math import atan, isnan, nan
 from typing import Optional, Union
 
 # 3rd-party
@@ -52,6 +52,8 @@ def _cpu(data, cellsize_x, cellsize_y):
     rows, cols = data.shape
     for y in range(1, rows - 1):
         for x in range(1, cols - 1):
+            if np.isnan(data[y, x]):
+                continue
             a = data[y + 1, x - 1]
             b = data[y + 1, x]
             c = data[y + 1, x + 1]
@@ -112,6 +114,8 @@ def _run_dask_cupy(data: da.Array,
 
 @cuda.jit(device=True)
 def _gpu(arr, cellsize_x, cellsize_y):
+    if isnan(arr[1, 1]):
+        return nan
     a = arr[2, 0]
     b = arr[2, 1]
     c = arr[2, 2]
--- a/xrspatial/tests/test_slope.py
+++ b/xrspatial/tests/test_slope.py
@@ -274,3 +274,48 @@ def test_degenerate_shape_geodesic(shape):
     general_output_checks(raster, result)
     assert result.shape == shape
     assert np.all(np.isnan(result.data))
+
+
+# A NoData (NaN) center cell must stay NaN in the slope output, even when all 8
+# of its neighbours are valid. The planar kernels read the center cell, so a
+# hole in the DEM can't masquerade as valid flat terrain. Regression for #2761.
+def _center_nan_data():
+    data = np.zeros((5, 5), dtype=np.float64)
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Propagate NaN center cell in planar slope

None.

None.

- `xrspatial/slope.py:116` -- the CPU path checks the center with `np.isnan` and the GPU path uses `math.isnan`. Each is the right call for its kernel, so nothing to change here. Just flagging it so the split doesn't trip up a future reader.

### What looks good
- The fix lives in the two shared kernels (`_cpu`, `_gpu`), so all four planar backends pick up the behaviour without touching the dispatch wiring. numpy and cupy were verified locally on a real GPU; the dask paths run the same kernels through `map_overlap`.
- `continue` (CPU) and `return nan` (GPU) both land on a NaN that `out` is already pre-filled with, so the center stays NaN and its neighbours go NaN on their own (they read the hole as one of their 8 inputs). Same result the geodesic method already gives.
- Tests hit all four backends and check numpy-vs-backend parity with `equal_nan=True`, plus a direct assertion on the center cell. The regression comment names the issue.
- Scope is tight. Geodesic paths and boundary handling are left alone.

### Checklist
- Algorithm matches reference: n/a (NoData propagation, not an algorithm change)
- All implemented backends produce consistent results: yes
- NaN handling is correct: yes
- Edge cases covered by tests: yes (center NaN with valid neighbours, all four backends)
- Dask chunk boundaries handled correctly: yes (depth=1 unchanged, kernel handles the NaN)
- Benchmark exists or not needed: not needed (no perf change)
- README feature matrix updated: n/a (no new function, no backend change)
- Docstrings present and accurate: yes (public docstring unchanged and still valid)

PR #2774, opened 2026-06-01
Skip full-raster geodesic memory guard for dask aspect

### LINKED ISSUE #2763: Chunked geodesic dask aspect blocked by full-raster memory guard
## Describe the bug
The full-raster memory guard for geodesic aspect rejects chunked dask inputs that it should accept.

`aspect(method='geodesic')` calls `_check_geodesic_memory(rows, cols, ...)` once, before backend dispatch, sizing the check against the entire raster (`xrspatial/aspect.py:426-427`). That guard estimates the memory needed to materialize a `(3, H, W)` float64 stack plus padded copies and a float32 output all at once.

For numpy and cupy backends that estimate is correct, since the whole raster is allocated in one shot. But the dask geodesic path never does that. It processes the raster chunk-by-chunk through `map_overlap` (`xrspatial/aspect.py:260` for dask+numpy and `xrspatial/aspect.py:286` for dask+cupy), so peak memory tracks chunk size, not full-raster size. The guard still measures the full raster and raises `MemoryError`, even though the dask path could have run fine.

The error message tells the user to "process it in chunks," but a chunked dask raster is exactly what gets rejected.

## To Reproduce
Build a 200x200 dask raster with small chunks, monkeypatch `xrspatial.geodesic._available_memory_bytes` to return a tiny value, and call `aspect(raster, method='geodesic')`. It raises `MemoryError` even though each chunk easily fits.

## Expected behavior
The dask geodesic path (dask+numpy and dask+cupy) should not be blocked by the full-raster guard. The guard should apply only to in-memory numpy/cupy arrays, where the whole raster is materialized at once. For dask arrays the guard should be skipped (peak memory is bounded by chunk size).

## Severity
High. This makes the chunked geodesic path unusable under memory pressure, which is the exact case dask is meant to handle.

brendancol:
Closes #2763

## What changed
- `aspect(method='geodesic')` no longer runs the full-raster memory guard for dask-backed inputs. The dask backends process the raster chunk-by-chunk via `map_overlap`, so peak memory is bounded by chunk size, not full-raster size. The guard now applies only to in-memory numpy/cupy arrays, where the whole `(3, H, W)` stack is materialized at once.

## Backend coverage
- numpy: unchanged, guard still runs
- cupy: unchanged, guard still runs
- dask+numpy: guard skipped (was incorrectly raising MemoryError)
- dask+cupy: guard skipped (same path)

## Test plan
- [x] New test: a 200x200 dask raster with (40, 40) chunks runs under geodesic aspect with available memory monkeypatched to 1 MB, and computes to the right shape. Verified this fails on unpatched code and passes after the fix.
- [x] Existing in-memory guard test still raises MemoryError when the full raster genuinely will not fit.
- [x] Full test_geodesic_aspect.py suite passes (36 tests).

### DIFF
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -19,7 +19,7 @@ from xrspatial.geodesic import (INV_2R, WGS84_A2, WGS84_B2, _check_geodesic_memo
                                 _cpu_geodesic_aspect, _run_gpu_geodesic_aspect)
 from xrspatial.utils import (Z_UNITS, ArrayTypeFunctionMapping, _boundary_to_dask,
                              _extract_latlon_coords, _pad_array, _validate_boundary,
-                             _validate_raster, cuda_args, ngjit)
+                             _validate_raster, cuda_args, has_dask_array, ngjit)
 
 
 def _geodesic_cuda_dims(shape):
@@ -423,8 +423,14 @@ def aspect(agg: xr.DataArray,
             )
         z_factor = Z_UNITS[z_unit]
 
-        rows, cols = agg.shape[-2], agg.shape[-1]
-        _check_geodesic_memory(rows, cols, func_name='aspect')
+        # The full-raster memory guard only applies to in-memory (numpy/cupy)
+        # arrays, which materialize the whole (3, H, W) stack at once. Dask
+        # backends process the raster chunk-by-chunk via map_overlap, so peak
+        # memory is bounded by chunk size, not full-raster size.
+        is_dask = has_dask_array() and isinstance(agg.data, da.Array)
+        if not is_dask:
+            rows, cols = agg.shape[-2], agg.shape[-1]
+            _check_geodesic_memory(rows, cols, func_name='aspect')
 
         lat_2d, lon_2d = _extract_latlon_coords(agg)
 
--- a/xrspatial/tests/test_geodesic_aspect.py
+++ b/xrspatial/tests/test_geodesic_aspect.py
@@ -257,6 +257,24 @@ class TestGeodesicAspectMemoryGuard:
         result = aspect(raster, method='planar')
         assert result.shape == (8, 8)
 
+    @dask_array_available
+    def test_dask_chunked_skips_full_raster_guard(self, monkeypatch):
+        """A chunked dask raster is processed chunk-by-chunk via map_overlap,
+        so the full-raster guard must not reject it even when available memory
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Skip full-raster geodesic memory guard for dask aspect

None.

- The new test covers dask+numpy only. The dask+cupy path goes through the same `is_dask` branch in aspect.py, so the fix applies to it too, but there is no test asserting dask+cupy skips the guard under low memory. A small `@cuda_and_cupy_available` test mirroring the dask+numpy one would lock that path down. Not blocking, since the code path is shared and the guard skip is backend-agnostic.

- The same full-raster guard pattern exists in slope.py:397 with the identical dask-rejection bug. Out of scope for this PR (the issue is scoped to aspect), but worth a follow-up so slope's dask geodesic path isn't blocked the same way.

### What looks good
- The `is_dask` check reuses the project's standard detection (`has_dask_array() and isinstance(agg.data, da.Array)`), matching ArrayTypeFunctionMapping and validate_arrays.
- The guard still runs for numpy/cupy, so the in-memory OOM protection is intact. The existing oversized-raster test confirms it.
- The comment explains why dask is exempt (chunk-by-chunk via map_overlap), so the next reader won't re-add the guard.
- The new test forces `.compute()`, so it actually exercises the chunked path rather than just building a lazy graph.

### Checklist
- Algorithm matches reference/paper: n/a, no algorithm change
- All implemented backends produce consistent results: unchanged
- Edge cases covered by tests: guard-skip for dask and guard-raise for in-memory both tested
- Dask chunk boundaries handled correctly: unchanged map_overlap path
- No premature materialization or unnecessary copies: confirmed, the fix removes a false rejection only
- README feature matrix updated: not applicable, no API or backend change
- Docstrings present and accurate: unchanged public docstring still accurate

brendancol (commented):
## Follow-up review (after 204a20c0)

Disposition of the prior findings:

- Suggestion (dask+cupy guard-skip test): fixed. Added `test_dask_cupy_chunked_skips_full_raster_guard`, gated on `cuda_and_cupy_available`, mirroring the dask+numpy test.
- Nit (slope.py:397 has the same dask-rejection bug): deferred to #2778. It's a separate module outside this issue's scope and warrants its own change with its own slope tests.

No new findings on the added test. The memory-guard tests pass (the dask+cupy one skips here without a GPU, as expected). Nothing else changed.

PR #2776, opened 2026-06-01
Keep dask+cupy geodesic slope lat/lon lazy to avoid GPU OOM

### LINKED ISSUE #2762: dask+cupy geodesic slope eagerly materializes full lat/lon grids on GPU
**Describe the bug**

`_run_dask_cupy_geodesic` in `xrspatial/slope.py` calls `cupy.asarray(lat_2d)` and `cupy.asarray(lon_2d)` while building the dask graph (around line 274). That densifies the full (H, W) lat/lon grids onto a single GPU at graph-construction time, before any compute runs. On large rasters it allocates roughly `2 * H * W * 8` bytes of GPU memory up front and can OOM before the computation starts.

Aspect already fixed this. In `xrspatial/aspect.py` (around line 286), `_run_dask_cupy_geodesic` keeps lat/lon as dask-of-numpy on the broadcast views and converts each block to cupy lazily with `map_blocks(_to_cupy_f64)`. A regression test guards it at `xrspatial/tests/test_geodesic_aspect.py:309`.

**Expected behavior**

The dask+cupy geodesic slope path should keep lat/lon chunked and convert blocks lazily, the same way aspect does, so graph construction does not allocate the full lat/lon grids on the GPU.

**Additional context**

Port the lazy-block pattern from `aspect.py` into `slope.py`, and add the matching regression test that asserts graph construction stays well under one full lat/lon grid of GPU memory.

brendancol:
## Summary

- `_run_dask_cupy_geodesic` in `slope.py` no longer builds the full lat/lon grids on the GPU during graph construction. It keeps lat/lon as dask-of-numpy on the broadcast views and converts each block to cupy lazily with `map_blocks(_to_cupy_f64)`, the same way `aspect.py` already does it.
- Before this, `cupy.asarray(lat_2d)` and `cupy.asarray(lon_2d)` densified both full (H, W) grids onto a single GPU before any compute ran, which could OOM on large rasters.

## Backend coverage

Only the dask+cupy geodesic path changed. The numpy, cupy, and dask+numpy paths are untouched. Cross-backend parity stays covered by the existing `test_numpy_equals_dask_cupy`.

## Test plan

- [x] New regression test `test_latlon_not_materialized_on_gpu_at_graph_build` checks that graph construction allocates well under one full lat/lon grid of GPU memory.
- [x] Confirmed the test fails on the old eager code (allocated 2x one grid) and passes after the fix.
- [x] Full `test_geodesic_slope.py` suite passes (22 passed) on a CUDA GPU.

Closes #2762

### DIFF
--- a/xrspatial/slope.py
+++ b/xrspatial/slope.py
@@ -271,11 +271,21 @@ def _run_dask_numpy_geodesic(data, lat_2d, lon_2d, a2, b2, z_factor, boundary='n
     return out[0]
 
 
+def _to_cupy_f64(block):
+    # Only reached from the dask+cupy path, so `cupy` is the real module here,
+    # never the import-time fallback class.
+    return cupy.asarray(block, dtype=cupy.float64)
+
+
 def _run_dask_cupy_geodesic(data, lat_2d, lon_2d, a2, b2, z_factor, boundary='nan'):
-    lat_dask = da.from_array(cupy.asarray(lat_2d, dtype=cupy.float64),
-                             chunks=data.chunksize)
-    lon_dask = da.from_array(cupy.asarray(lon_2d, dtype=cupy.float64),
-                             chunks=data.chunksize)
+    # Keep lat/lon as dask-of-numpy on the (zero-stride) broadcast views, then
+    # convert each block to cupy lazily. Converting up front with
+    # cupy.asarray(lat_2d) would densify the full (H, W) grid onto a single GPU
+    # at graph-construction time and OOM on large rasters.
+    lat_dask = da.from_array(lat_2d, chunks=data.chunksize).map_blocks(
+        _to_cupy_f64, dtype=np.float64)
+    lon_dask = da.from_array(lon_2d, chunks=data.chunksize).map_blocks(
+        _to_cupy_f64, dtype=np.float64)
     stacked = da.stack([
         data.astype(cupy.float64),
         lat_dask,
--- a/xrspatial/tests/test_geodesic_slope.py
+++ b/xrspatial/tests/test_geodesic_slope.py
@@ -296,7 +296,6 @@ class TestGeodesicSlopeDask:
 class TestGeodesicSlopeCupy:
 
     def test_numpy_equals_cupy(self):
-        import cupy
         elev = _east_tilted_surface(H=8, W=10, grade=100.0)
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Keep dask+cupy geodesic slope lat/lon lazy to avoid GPU OOM

None.

None.

- `xrspatial/tests/test_geodesic_slope.py:299` has a pre-existing F401 (`import cupy` unused) in `test_numpy_equals_cupy`, just above the touched class. Not introduced by this PR, but it is a one-line cleanup if you are already in the file.

### What looks good
- The fix is a direct port of the aspect.py pattern (`aspect.py:286`), down to the `_to_cupy_f64` helper and the explanatory comment. Both modules now handle the dask+cupy geodesic path the same way.
- `lat_2d`/`lon_2d` come from `_extract_latlon_coords` as `np.broadcast_to` views, so `da.from_array(lat_2d)` keeps a small zero-stride numpy view in the graph and `map_blocks(_to_cupy_f64)` defers the cupy conversion to compute time. No full grid lands on the GPU at graph-build.
- The regression test measures real GPU pool usage and asserts it stays under one full lat/lon grid. I confirmed it fails on the old eager code (allocated 2x one grid, 67MB) and passes after the fix.
- Only the dask+cupy path changed. numpy, cupy, and dask+numpy are untouched, and the existing `test_numpy_equals_dask_cupy` still guards cross-backend parity.

### Checklist
- [x] Algorithm matches reference (port of aspect.py)
- [x] All implemented backends produce consistent results (parity test unchanged)
- [x] Edge cases covered (new graph-build memory test; large-raster path)
- [x] Dask chunk boundaries handled correctly (map_overlap depth unchanged)
- [x] No premature materialization (the point of the change)
- [x] Benchmark exists or is not needed (internal bug fix, no new API)

brendancol (commented):

The one nit from the first pass is resolved: the unused `import cupy` in `test_numpy_equals_cupy` is gone, and flake8 now runs clean on both changed files. The `import cupy` inside the new graph-build test is used (`cupy.zeros`, `cupy.get_default_memory_pool`), so it stays.

No new findings. No blockers, no suggestions, no remaining nits. The fix still matches the aspect.py pattern and the regression test still passes on a CUDA GPU.

PR #2777, opened 2026-06-01
Warn at runtime on degree coords + meter elevation in slope planar path

### LINKED ISSUE #2764: slope() planar path does not warn on degree coordinates with meter elevation
**Describe the bug**

`slope()` with the default `method='planar'` uses the coordinate spacing directly as the cellsize (xrspatial/slope.py:378). When the coordinates are in degrees (lat/lon) but the elevation values are in meters, the rise-over-run computation mixes incompatible units and the returned slope is wrong by orders of magnitude. Nothing tells the caller this happened, so downstream code trusts the numbers.

A warning helper already exists for this case, `warn_if_unit_mismatch` in xrspatial/utils.py:781. It is tested in isolation but no module calls it. The docs mention the degree/meter pitfall, but a doc note does not help someone who never reads it and feeds the output into a model.

**Expected behavior**

When the planar path detects degree-like horizontal coordinates alongside meter-like elevation values, `slope()` should emit a `UserWarning` at runtime pointing the caller to reproject to a projected CRS. Projected/meter coordinates should produce no warning.

**Additional context**

Fix is to wire the planar branch of `slope()` to call the existing `warn_if_unit_mismatch` helper. Add a test asserting the warning fires for degree coordinates and stays silent for projected coordinates.

brendancol:
Closes #2764

The planar slope path uses coordinate spacing directly as the cell size. When coordinates are in degrees (lat/lon) but elevation is in meters, the rise-over-run mixes units and the output is wrong by orders of magnitude, silently. A warning helper for this case already existed (`warn_if_unit_mismatch` in utils.py) but nothing called it.

- Wire `warn_if_unit_mismatch` into the planar branch of `slope()` so a `UserWarning` fires when degree-like coordinates meet meter-like elevation.
- Document the runtime warning in the `slope` docstring.

No new computation, only a runtime check on the existing planar path, so all four backends (numpy / cupy / dask+numpy / dask+cupy) are unaffected. The warning runs on the eager DataArray metadata before dispatch.

Test plan:
- [x] New test asserts the warning fires for degree coordinates with meter elevation.
- [x] New test asserts no warning for projected/meter coordinates.
- [x] Full `test_slope.py` suite passes (63 tests).


brendancol (commented):
## PR Review: Warn at runtime on degree coords + meter elevation in slope planar path

None.

- aspect() has the same planar/degree footgun and the same helper would apply there. Out of scope for this PR, but flagging it so it does not get forgotten.

- xrspatial/slope.py: the warning only fires in the planar branch, which is correct since geodesic handles lat/lon properly. Noting it so the asymmetry is on the record as intentional.

### What looks good
- Reuses the existing warn_if_unit_mismatch helper instead of reinventing the check. The helper was tested in isolation but unused; this wires it to a real call site.
- The warning runs on eager DataArray metadata before backend dispatch, so all four backends behave identically and no array is materialized.
- Tests cover both directions: the warning fires for degree coords with meter elevation, and stays silent for projected/meter coords. The no-warn test uses simplefilter("error", UserWarning), so any stray warning would fail it too.
- Docstring Notes section explains the failure mode and points to the fix (reproject or use geodesic).

### Checklist
- Algorithm matches reference: n/a, no algorithm change
- All implemented backends consistent: yes, check runs pre-dispatch
- NaN handling: unchanged
- Edge cases covered by tests: degree and projected cases both covered
- Dask chunk boundaries: unchanged, warning is metadata-only
- No premature materialization: confirmed, helper samples metadata only
- Benchmark: not needed
- README feature matrix: not needed, no new function

brendancol (commented):
## PR Review: Warn on degree coords + meter elevation in slope planar path

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- Reuses the existing `warn_if_unit_mismatch` helper (utils.py:833) instead of writing fresh detection logic, called at the top of the planar branch (slope.py:402) before the resolution gets used.
- Scoped to the planar path only. Geodesic is left alone, which is right since it handles lat/lon itself.
- The check reads coordinate metadata, so it fires once and never triggers a dask compute. That also makes it backend-agnostic.
- Tests cover both directions: degree coords + meter elevation warns, projected meter coords stay silent under `simplefilter("error", UserWarning)`.
- The merge with main is clean. The test rasters use evenly spaced coords, so the separately-merged irregular-spacing warning (#2784) doesn't fire and trip the no-warn test. Full test_slope.py passes (67 tests).
- A Notes section in the slope docstring spells out the failure mode.

### Checklist
- [x] No algorithm change (adds a runtime warning)
- [x] Backend-agnostic (warning runs on metadata)
- [x] NaN handling unchanged by this PR
- [x] Edge cases tested (degree warn, projected no-warn)
- [x] No neighborhood/dask change
- [x] Metadata-only check, no materialization
- [x] No benchmark needed (no compute change)
- [x] README matrix N/A (no new function, no backend change)
- [x] Docstring Notes section added

PR #2779, opened 2026-06-01
Make geodesic slope memory guard backend-aware for dask

### LINKED ISSUE #2765: Geodesic memory guard rejects chunked Dask rasters using full-raster size
**Describe the bug**

`slope(method='geodesic')` calls `_check_geodesic_memory(rows, cols)` (xrspatial/slope.py:397) before it dispatches to a backend. The guard sizes the working buffers against the full raster's `(rows, cols)`. That's right for the eager NumPy and CuPy paths, which allocate a single `(3, H, W)` float64 stacked array up front. It's wrong for Dask. The dask geodesic backends work chunk by chunk, so peak memory tracks the largest chunk, not the whole raster. A big raster that's chunked small enough to run fine can still get rejected with a MemoryError before any chunking happens.

**Expected behavior**

The guard should be backend-aware. Keep the full-raster check for eager NumPy/CuPy. For Dask, check the largest chunk's `(h, w)` instead, so a raster that's too big to run eagerly but chunked small enough to run lazily is allowed through.

**Additional context**

The dask geodesic paths (`_run_dask_numpy_geodesic`, `_run_dask_cupy_geodesic`) stack three channels, rechunk `{0: 3}`, then `map_overlap` with depth `(0, 1, 1)`. Peak per-chunk working set is roughly the largest spatial chunk times the per-cell byte budget, not the full raster size.

brendancol:
Closes #2765

## What changed

- `slope(method='geodesic')` sized its memory guard against the full raster before dispatch. That's correct for eager NumPy/CuPy (they build a `(3, H, W)` float64 stack up front), but wrong for dask, which streams that stack chunk by chunk via `map_overlap`. The full-raster check rejected large-but-chunked rasters that would run fine.
- Added `_check_geodesic_memory_backend_aware`: eager backends keep the full-raster check; dask backends check the largest spatial chunk plus its 1-cell `map_overlap` halo (+2 per spatial dim). A single whole-raster chunk has no memory advantage and is still rejected.

## Backend coverage

The guard runs before dispatch, so the new path covers all four backends: numpy and cupy use the eager full-raster check; dask+numpy and dask+cupy use the per-chunk check. `aspect(method='geodesic')` has the same eager-only guard and is left for a separate change to keep this PR scoped to the reported finding.

## Test plan

- [x] Eager raster too large for available RAM still raises `MemoryError` (existing test).
- [x] Same raster chunked small enough is allowed through (new dask test).
- [x] Dask array with a single whole-raster chunk is still rejected (new dask test).
- [x] `test_geodesic_slope.py` passes (23 tests).
- [x] `test_geodesic_aspect.py` still passes (shared `geodesic.py` untouched in behaviour).

### DIFF
--- a/xrspatial/geodesic.py
+++ b/xrspatial/geodesic.py
@@ -82,6 +82,35 @@ def _check_geodesic_memory(rows, cols, func_name):
         )
 
 
+def _check_geodesic_memory_backend_aware(agg, func_name):
+    """Backend-aware front end for :func:`_check_geodesic_memory`.
+
+    Eager NumPy / CuPy backends build a ``(3, H, W)`` float64 stack for the
+    whole raster up front, so the guard sizes against the full ``(rows, cols)``.
+
+    Dask backends run the same stack chunk by chunk via ``map_overlap``, so
+    peak working memory is bounded by the largest chunk plus its 1-cell halo,
+    not the full raster. Sizing the guard against the full raster would reject
+    large-but-chunked rasters that would otherwise stream through fine, so for
+    Dask we check the largest chunk instead.
+    """
+    data = getattr(agg, "data", agg)
+
+    chunks = getattr(data, "chunks", None)
+    if chunks is None:
+        # Eager (NumPy / CuPy): size against the full raster.
+        rows, cols = agg.shape[-2], agg.shape[-1]
+        _check_geodesic_memory(rows, cols, func_name=func_name)
+        return
+
+    # Dask: size against the largest spatial chunk, plus the 1-cell halo that
+    # ``map_overlap`` adds on each side (depth (0, 1, 1) -> +2 per spatial dim).
+    row_chunks, col_chunks = chunks[-2], chunks[-1]
+    rows = (max(row_chunks) if row_chunks else 0) + 2
+    cols = (max(col_chunks) if col_chunks else 0) + 2
+    _check_geodesic_memory(rows, cols, func_name=func_name)
+
+
 # =====================================================================
 # CPU (Numba ngjit) primitives
 # =====================================================================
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Make geodesic slope memory guard backend-aware for dask

None.

- `aspect.py:427` has the same eager-only guard (`_check_geodesic_memory(rows, cols)`) and so has the same bug. The PR scopes itself to slope per the issue and says so in the body, which is fair. File a follow-up so the aspect path doesn't quietly keep the wrong behavior.

- `geodesic.py`: the guard uses `max(row_chunks)` and `max(col_chunks)` independently, so the two maxima can come from different chunks and the bound can slightly overestimate the largest real chunk. That over-rejects rather than under-rejects, which is the safe direction for a memory guard, so it's fine as-is. Noting it so the conservatism is on record.

### What looks good
- The `+2` per spatial dim matches the depth-1 `map_overlap` halo. The chunk function allocates against the haloed chunk shape, so the byte budget tracks the real peak.
- `getattr(data, "chunks", None)` splits eager (no `chunks` attribute) from dask without importing dask in the guard.
- Tests pin all three behaviors: eager rejected, chunked allowed, single whole-raster chunk still rejected. The chunked test first checks that the same raster is rejected eagerly, which makes the contrast explicit.

### Checklist
- Contract matches the issue: yes
- Both backend classes covered: yes (eager full-raster, dask per-chunk)
- NaN handling: unchanged, n/a
- Edge cases tested: empty chunk tuple guarded with `... else 0`
- Dask chunk boundaries: yes, `+2` halo accounted for
- No premature materialization: guard reads `.chunks` metadata only, no compute
- Benchmark: not needed (cheap metadata check)
- README matrix: not needed (no API change)
- Docstrings: present on the new helper

brendancol:
Review follow-up: the aspect.py suggestion is tracked as a separate issue (#2783) to keep this PR scoped to slope. The `max() * max()` nit is intended (conservative over-rejection is the safe direction for a memory guard); no change. No code changes resulted, so the existing review stands.

PR #2780, opened 2026-06-01
Apply cell-size correction to planar aspect (#2760)

### LINKED ISSUE #2760: Planar aspect() ignores cell size; wrong for non-square cells
**Describe the bug**

Planar `aspect()` ignores cell size, so it returns wrong values for rasters with non-square cells.

The planar Horn kernels in `xrspatial/aspect.py` divide both gradients by 8 without dividing by `cellsize_x` and `cellsize_y`:

- `xrspatial/aspect.py:67` (CPU kernel `_cpu`): `dz_dx = ((c + 2*f + i) - (a + 2*d + g)) / 8` and `dz_dy = (...) / 8`
- `xrspatial/aspect.py:106` (GPU device kernel `_gpu`): same divide-by-8

`slope()` already divides `dz_dx` by `8 * cellsize_x` and `dz_dy` by `8 * cellsize_y` (see `xrspatial/slope.py:63-64`), and its public function extracts resolution via `get_dataarray_resolution`. `aspect()` does neither. For square cells the answer is right because the cell size cancels in the atan2 ratio, but for rectangular cells the East/North gradient ratio is wrong and the aspect direction is off.

**Expected behavior**

For a constant-gradient raster, planar aspect should equal the analytically correct Horn value regardless of cell aspect ratio.

Confirmed with a constant-gradient raster where xres=10, yres=1: current planar aspect returns 275.7106, while the cell-size-correct Horn calculation gives 315.0.

Because `northness()` returns `cos(aspect)` and `eastness()` returns `sin(aspect)`, both are corrupted by the same bug.

**Scope**

- Affects all four backends (numpy, cupy, dask+numpy, dask+cupy) since they share the `_cpu`/`_gpu` kernels.
- Geodesic method is unaffected (it does not use these kernels).

**Fix**

Thread `cellsize_x`/`cellsize_y` through the planar aspect path the same way `slope()` does: extract resolution in the public `aspect()`, pass it into the numpy/cupy/dask backend wrappers, and divide `dz_dx` by `8 * cellsize_x` and `dz_dy` by `8 * cellsize_y` in both the CPU and GPU kernels.

brendancol:
Closes #2760.

## What

Planar `aspect()` divided both Horn gradients by 8 without dividing by `cellsize_x` / `cellsize_y`, so it returned wrong directions for non-square cells. `slope()` already applied this correction; aspect did not.

- Thread `cellsize_x` / `cellsize_y` through the planar aspect path the same way `slope()` does: extract resolution in the public `aspect()`, pass it into the numpy/cupy/dask backend wrappers, and divide `dz_dx` by `8 * cellsize_x` and `dz_dy` by `8 * cellsize_y` in both the CPU (`_cpu`) and GPU (`_gpu`/`_run_gpu`) kernels.
- Square cells are unchanged because the factor cancels in the atan2 ratio, so the existing QGIS reference tests still pass.
- `northness()` and `eastness()` are fixed for free since they derive from `aspect()`.

## Backend coverage

The fix lives in the shared `_cpu` and `_gpu` kernels, so all four backends benefit: numpy, cupy, dask+numpy, dask+cupy.

## Test plan

- [x] New regression test: constant-gradient raster with xres=10, yres=1 now yields aspect 315.0 (was 275.7106), with matching northness (cos 315) and eastness (sin 315). Covers numpy and dask+numpy directly; cupy and dask+cupy variants run on GPU CI.
- [x] Verified the new test fails on the unpatched kernel and passes after the fix.
- [x] Existing aspect, slope, and geodesic-aspect suites pass (231 tests), confirming square-cell behavior is unchanged.


brendancol (commented):
## PR Review: Apply cell-size correction to planar aspect (#2760)

None.

None.

- The aspect docstring still says the planar method "uses the classic Horn algorithm with uniform cell size". After this fix it handles non-uniform cells too. slope() carries the same stale wording, so leaving it keeps the two in sync; updating both would be a separate cleanup. Optional.

### What looks good
- The fix mirrors the existing, tested slope() implementation: same 8*cellsize_x / 8*cellsize_y scaling, same get_dataarray_resolution call in the public function, same 1-element cupy arrays passed to the GPU kernel. Low risk because it copies a known-good pattern.
- The kernel orientation and compass conversion are untouched, so the QGIS reference tests still pass. Square cells are unchanged since the factor cancels in the atan2 ratio.
- All four backends are covered through the shared _cpu / _gpu kernels and the four wrapper functions; ArrayTypeFunctionMapping already dispatches to them.
- The regression test asserts the analytical Horn value (315.0) for a rectangular-cell raster rather than backend parity, and also checks northness/eastness. It was verified to fail on the unpatched kernel.

### Notes checked, no action needed
- cellsize_y sign: calc_res derives yres from max-min of the y coords, so it is positive regardless of coordinate ordering, and the res attr is positive in practice. atan2 sign behavior matches slope's existing handling, so no new flip risk.
- Dask depth stays (1, 1) for the 3x3 kernel; boundary is forwarded unchanged. The cellsize values are bound via functools.partial, which is fine for map_overlap.
- Benchmarks already exist in benchmarks/benchmarks/aspect.py; no new function added, so no benchmark or README matrix change needed.

### Checklist
- [x] Algorithm matches the Horn cell-size correction used by slope()
- [x] All implemented backends share the corrected kernels
- [x] NaN handling unchanged (flat -> -1, edges -> NaN)
- [x] Edge cases covered (existing degenerate-shape tests still pass)
- [x] Dask chunk boundaries unchanged (depth 1, boundary forwarded)
- [x] Benchmark exists
- [x] Docstrings present (minor stale wording noted above)

brendancol (commented):

The one nit from the first pass is fixed: the planar aspect docstring now says the gradients are scaled by the x and y cell sizes, so it no longer claims uniform cell size.

Nothing else changed in the code path. All 135 aspect tests still pass locally. No remaining findings.

PR #2781, opened 2026-06-01
aspect: add independent rectangular-cell oracle for planar aspect

### LINKED ISSUE #2768: aspect tests: add an independent rectangular-cell oracle (parity-only coverage gap)
## Reason or Problem

The aspect tests mostly check that the backends agree with each other, not that any of them is correct. The planar tests in xrspatial/tests/test_aspect.py compare NumPy, dask, CuPy, and dask+CuPy, but all four run the same `_cpu`/`_gpu` kernels. A bug shared by every backend slips through, because the backends still match.

The missing-cellsize bug in #2760 is exactly this case. Planar aspect divides the Horn gradients by 8 but never by `cellsize_x`/`cellsize_y`, so it gives wrong answers on non-square cells, yet every backend agrees with every other one.

The QGIS fixture (`qgis_aspect`) is a real external reference, but its raster has square cells. It never touches rectangular cells and never checks that cell size is read from the coordinates.

## Proposal

Add an analytic oracle for planar aspect that does not depend on any other backend.

Take a surface with a known constant gradient, `z = gx*X + gy*Y` (X east, Y north). The true aspect is a closed-form compass conversion of `atan2(dz_dy, -dz_dx)`, with the gradient components scaled by the real cell sizes read from the x/y coordinate spacing. Compare xrspatial's output against that value instead of against another backend.

The oracle should use rectangular cells (xres != yres) so it checks that cell size comes from the coordinates, which is what the missing-cellsize bug gets wrong. Run it across whatever backends are available so it still works as a parity check, but the thing being asserted is the analytic value.

## Coupling with #2760

The rectangular-cell oracle fails on current main because the missing-cellsize bug (#2760) is still there. This issue is only the test, not the fix. Mark the rectangular-cell assertions xfail with strict=False and point at #2760, so they start passing on their own once that fix lands. The square-cell oracle passes now and is not xfail.

## Value

This gives the module a test that checks aspect correctness directly, including the rectangular-cell case the QGIS fixture skips. The xfail doubles as a tripwire: when #2760 merges, the rectangular oracle goes green and tells you the marker can come out.

brendancol:
Closes #2768

## What this does

The planar aspect tests only prove the four backends agree with each other. They share the same `_cpu`/`_gpu` kernels, so a defect common to every backend slips through. The missing-cellsize bug (#2760) is exactly that: planar aspect divides the Horn gradients by 8 but never by `cellsize_x`/`cellsize_y`, so it is wrong on non-square cells while every backend still matches. The QGIS fixture is a real external reference but uses square cells, so it never touches rectangular cells or checks that cell size comes from the coordinates.

This adds an analytic oracle for planar aspect:

- Builds a plane `z = gx*X + gy*Y` on a coordinate grid with no `res` attr, so resolution must be recovered from the x/y coordinate spacing.
- Asserts against the closed-form Horn compass value, not against another backend.
- Square-cell oracle passes unconditionally.
- Rectangular-cell oracle is `xfail(strict=False)` referencing #2760. It flips to XPASS automatically once that fix lands, which flags the marker for removal.

This PR is test-only. It does not touch `aspect.py`; the source fix is #2760's scope.

## Backend coverage

numpy, dask+numpy, cupy, dask+cupy. The cupy/dask+cupy oracle variants are gated behind the existing CUDA+CuPy skip marker.

## Test plan

- [x] `pytest xrspatial/tests/test_aspect.py` -- 165 passed, 16 xfailed (numpy + dask; cupy paths skipped, no GPU here)
- [x] Square-cell oracle passes on numpy and dask
- [x] Rectangular-cell oracle xfails against current main as expected (confirmed it reproduces the #2760 buggy value 275.71 vs oracle 315.0 for xres=10, yres=1)
- [x] flake8 clean on the added lines

When #2760 merges, the rectangular oracle should start XPASSing; remove the `xfail` markers at that point.

### DIFF
--- a/xrspatial/tests/test_aspect.py
+++ b/xrspatial/tests/test_aspect.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pytest
+import xarray as xr
 
 try:
     import dask.array as da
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: aspect: add independent rectangular-cell oracle for planar aspect

None.

- `_aspect_oracle` hand-copies the compass conversion from `aspect._cpu` (lines 74-81). That duplication is the whole point of an independent oracle, but it means the two can drift apart if the source convention ever changes. The square-cell tests re-check the oracle against the real function on every run, so this is fine as written. A one-line comment saying the conversion is duplicated on purpose would save the next reader some confusion. (xrspatial/tests/test_aspect.py:265)
- The flat case `gx == gy == 0` has a branch in `_aspect_oracle` but no test feeds it, so the -1 path is dead code in this PR. Either drop the branch or add a `(0.0, 0.0)` parametrize case. A flat-surface oracle case is cheap and worth having. (xrspatial/tests/test_aspect.py:271)

- In the cupy oracle tests, the inner `if backend == 'dask+cupy' and not has_dask_array(): pytest.skip(...)` only matters on a machine with CuPy but no dask, which is rare. Harmless, just slightly defensive. (xrspatial/tests/test_aspect.py:325)

### What looks good
- The oracle asserts against an analytic value, not another backend. That is the gap the issue called out.
- Rectangular cells with no `res` attr force resolution to come from coordinate spacing, which is the exact path the missing-cellsize bug breaks.
- The `xfail(strict=False)` pointing at #2760 is the right call: green CI now, automatic XPASS when the fix lands.
- Reuses `_to_numpy` and the existing skip markers; coverage spans numpy, dask, cupy, dask+cupy.

### Checklist
- [x] Oracle matches the Horn compass convention in aspect._cpu
- [x] Square-cell oracle passes on all available backends
- [x] Rectangular-cell oracle xfails pending #2760
- [x] Resolution is taken from coordinates (no res attr)
- [x] Test-only change; aspect.py untouched
- [x] flake8 clean on added lines
- [ ] Flat-surface (-1) oracle case not exercised

brendancol (commented):
## Follow-up review

The two suggestions from the first pass are addressed in 12be9683:

- The flat-surface case `(0.0, 0.0)` is now in the square oracle params, so the -1 branch in `_aspect_oracle` is exercised. xrspatial returns -1 for a constant plane and the oracle agrees.
- The docstring now says the arctan2/compass conversion is duplicated from `aspect._cpu` on purpose, with the square tests guarding against drift.

The one remaining nit (the defensive `has_dask_array()` skip inside the cupy tests) is left as-is on purpose. It only matters on a CuPy-without-dask machine, and keeping it costs nothing while avoiding a confusing failure in that rare setup.

Oracle tests: 42 passed, 16 xfailed (numpy + dask; cupy paths skipped, no GPU here). flake8 clean on the added lines. No new findings.

PR #2784, opened 2026-06-01
Warn when planar resolution averages irregular coordinates (#2766)

### LINKED ISSUE #2766: Planar slope silently averages resolution on irregular coordinates
### Description

Planar `slope` assumes uniform cell spacing even when the input coordinates are irregular, and it does so silently. `get_dataarray_resolution` (in `xrspatial/utils.py`) falls back to `calc_res`, which derives a single resolution from the full coordinate span divided by `n - 1`:

```python
xres = (xrange[-1] - xrange[0]) / (w - 1)
yres = (yrange[-1] - yrange[0]) / (h - 1)
```

On an irregular grid that collapses all the varying cell widths into one average value, and the user gets no indication it happened.

### Reproduce

Take `x = [0, 1, 2, 4, 8]` with `z = x` along that axis. The true physical slope is constant, but `slope` returns varying values because every cell is treated as if it had the averaged width.

### Expected

Either the resolution reflects the actual per-cell spacing, or the user is told the grid is irregular and an averaged resolution is being used. The `res` attribute already gives an explicit override path, which `get_dataarray_resolution` honors before it ever reaches `calc_res`.

### Proposed fix

When coordinates are present and not evenly spaced (within a tolerance), warn from `calc_res` so the averaging is visible, and point users at `attrs["res"]` for an explicit resolution. Regular grids keep their current behavior with no warning. This mirrors the existing `resample` validation for irregular coordinates (#2663) and the unit-mismatch warning already in `utils.py`.

brendancol:
Closes #2766

## What changed
- `calc_res` now checks whether each 1-D spatial coordinate is evenly spaced. When it isn't, it emits a `UserWarning` instead of quietly reducing the irregular grid to one averaged cell size, and points at `attrs["res"]` for an explicit override.
- The averaged resolution value is unchanged, so regular grids behave exactly as before and no warning fires.
- A descending (north-up) axis is regular even though its steps are negative, so the check compares step magnitudes against the averaged step. Without that, a normal north-up raster would warn falsely.

## Why
Planar `slope` (and other distance-based tools that read `calc_res`) treated `x=[0,1,2,4,8]` with `z=x` as if every cell had the averaged width, so it returned a varying slope where the true physical slope is constant, with no signal to the user. `attrs["res"]` already gives an explicit override that `get_dataarray_resolution` honors before reaching `calc_res`, so the warning only fires when the grid is irregular and no `res` attr is set.

## Backends
This is pure coordinate inspection in `calc_res`, so it's backend-agnostic. Coordinates are read as numpy regardless of array backend, so numpy / cupy / dask+numpy / dask+cupy all behave the same.

## Test plan
- [x] Regular grid: no warning, resolution unchanged
- [x] Irregular x and irregular y: warns with the offending dim name
- [x] Descending (north-up) regular axis: no warning
- [x] Float regular grid (lat/lon): no warning
- [x] `attrs["res"]` set on an irregular grid: no warning
- [x] No spatial coords: no warning
- [x] End-to-end through `slope` on irregular coords: warns
- [x] Existing slope and rasterize/resample consumer suites pass


brendancol (commented):
## PR Review: Warn when planar resolution averages irregular coordinates (#2766)

None.

- The `coord.size < 3` guard in `_warn_if_irregular_spacing` (utils.py) means a 2-cell axis never warns. That's correct, since a 2-point axis has a single diff that always equals the averaged step, but a one-line comment saying so would save the next reader a moment of squinting.

- `np.asarray(coord.values)` would raise on a cupy-backed coordinate. In practice xrspatial coords stay numpy even for cupy/dask data, and the existing `get_xy_range` already calls `.item()` on them, so this matches the surrounding code and isn't a regression. Worth keeping in mind if coords ever move to the GPU.

### What looks good
- The descending-axis fix (comparing `abs(diffs)` to `abs(res)`) is the right call and has a regression test. Without it every north-up raster would warn falsely; the consumer suites confirm rasterize/resample no longer trip on it.
- The warning only fires when there's no `attrs['res']`, because `get_dataarray_resolution` short-circuits on the attr before reaching `calc_res`. That path has a direct test.
- Test coverage is thorough: regular, irregular x, irregular y, descending, float lat/lon, no-coords, res-attr override, and an end-to-end slope case.
- The averaged resolution value is unchanged, so regular grids keep identical numeric behavior.

### Checklist
- [x] Resolution value unchanged for regular grids
- [x] Descending/north-up axis handled (no false positive)
- [x] attrs['res'] override path tested
- [x] Edge cases covered (2-cell axis, no coords, float jitter)
- [x] Backend-agnostic (coordinate inspection only)
- [x] No new public API, so no README/docs change needed
- [x] Existing consumer suites (slope, rasterize, resample) pass

brendancol (commented):
## Follow-up review (after commit 112aaa7e)

Re-reviewed after the follow-up commit that comments the `coord.size < 3` guard in `_warn_if_irregular_spacing`.

### Blockers
None.

### Suggestions
None left. The one accepted suggestion from the prior pass (comment the `>= 3` point guard, utils.py:476-478) is in.

### Nits
None left. The earlier `np.asarray` on cupy coords nit stays dismissed: xarray coordinate values are numpy-backed even for cupy/dask DataArrays, so `np.asarray(coord.values)` at utils.py:481 never touches device memory.

### What looks good
- Descending-axis handling compares `abs(diffs)` to `abs(res)`, so north-up rasters don't warn falsely. Covered by `test_calc_res_descending_axis_no_warning`.
- The `rtol=1e-5` tolerance keeps float jitter from tripping the warning; `test_calc_res_float_regular_grid_no_warning` exercises it with realistic lat/lon spacing.
- Missing coords, non-numeric dtype, non-finite diffs, and the <3 point case each return early.
- The averaged resolution value is unchanged, so regular grids behave exactly as before. Only the warning is new.

### Checklist
- [x] Algorithm matches intent (averaged-resolution detection)
- [x] Backend-agnostic (coords read as numpy regardless of array backend)
- [x] NaN/non-finite handling correct
- [x] Edge cases covered by tests (descending, float, no-coords, res-attr, end-to-end slope)
- [x] No premature materialization (pure coordinate inspection)
- [x] No README/feature-matrix change needed (internal helper, no new public API)

Nothing actionable remains. Only the dismissed cupy nit stands.

PR #2798, opened 2026-06-01
Validate focal_stats() stats_funcs before dispatch

### LINKED ISSUE #2770: focal_stats() does not validate stats_funcs before dispatch
## Description

`focal_stats()` doesn't validate `stats_funcs` before it dispatches to a backend, so two bad inputs fail in confusing ways:

1. An unknown stat name (e.g. `stats_funcs=['mena']`) falls through to a backend mapper dict lookup and surfaces as a bare `KeyError`. Nothing tells you which names are valid.
2. A bare string instead of a list (e.g. `stats_funcs='mean'`) gets iterated character by character, so the first lookup is for `'m'` and it dies there.

The dispatch happens around `xrspatial/focal.py:1115` in `_focal_stats_cpu` and the other backend helpers.

## Expected behavior

Validate `stats_funcs` up front, before dispatch:
- An unknown stat name raises a `ValueError` that lists the valid options.
- A bare string is accepted as a single stat name (wrapped into a one-element list).

## Steps to reproduce

```python
from xrspatial.convolution import circle_kernel
from xrspatial.focal import focal_stats

agg = xr.DataArray(np.ones((5, 5), dtype=float))
kernel = circle_kernel(1, 1, 1)

focal_stats(agg, kernel, stats_funcs=['mena'])  # raw KeyError
focal_stats(agg, kernel, stats_funcs='mean')    # KeyError on 'm'
```

brendancol:
Closes #2770

## Summary
`focal_stats()` now checks its `stats_funcs` argument before dispatching to a backend. Before this, an unknown name fell through as a raw `KeyError`, and passing a bare string like `'mean'` got iterated character by character and blew up on `'m'`.

A new `_validate_stats_funcs` helper wraps a bare string into a one-element list and raises a `ValueError` that lists the valid options when it sees an unknown name.

## Backend coverage
Validation happens before backend dispatch, so it applies to numpy, cupy, dask+numpy, and dask+cupy alike.

## Test plan
- [x] Unknown name raises a clear `ValueError`
- [x] Bare string is treated as a single stat name
- [x] Valid list still works
- [x] Existing focal_stats suite still passes (30 passed)

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -1121,6 +1121,28 @@ def _focal_stats_cpu(agg, kernel, stats_funcs, boundary='nan'):
     return stats
 
 
+_VALID_STATS_FUNCS = ('mean', 'max', 'min', 'range', 'std', 'var',
+                      'sum', 'variety')
+
+
+def _validate_stats_funcs(stats_funcs):
+    """Normalise and validate the ``stats_funcs`` argument of focal_stats.
+
+    A bare string is wrapped into a one-element list so it is not iterated
+    character by character. Unknown names raise a ValueError listing the
+    valid options.
+    """
+    if isinstance(stats_funcs, str):
+        stats_funcs = [stats_funcs]
+    unknown = [s for s in stats_funcs if s not in _VALID_STATS_FUNCS]
+    if unknown:
+        raise ValueError(
+            f"Invalid stats_funcs {unknown}, "
+            f"must be one of {list(_VALID_STATS_FUNCS)}"
+        )
+    return stats_funcs
+
+
 def focal_stats(agg,
                 kernel,
                 stats_funcs=None,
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Validate focal_stats() stats_funcs before dispatch

None.

None.

- xrspatial/focal.py (the new `_validate_stats_funcs` helper): an empty list (`stats_funcs=[]`) passes validation and then fails later inside the backend at `xr.concat` of an empty sequence. This is pre-existing behavior the PR does not change, so it is not a blocker. Since you are already adding a validation gate here, rejecting an empty list with the same clear ValueError would be cheap. Optional.

### What looks good
- `_VALID_STATS_FUNCS` matches the keys in `_focal_stats_cpu` and the cuda/cupy mappers exactly, so validation accepts precisely the names the backends implement.
- The bare string is normalized before the membership check, so the character-iteration bug is fixed at the source.
- Validation runs before backend dispatch, so numpy, cupy, dask+numpy, and dask+cupy all get the same behavior through one code path.
- The 3D per-band recursion re-validates the already-normalized list, which is idempotent and harmless.
- Three focused regression tests cover the unknown-name, bare-string, and valid-list cases. The full focal_stats suite still passes (30 tests).

### Checklist
- Algorithm matches reference: n/a (validation only)
- All implemented backends produce consistent results: yes, validation is backend-agnostic
- Edge cases covered by tests: unknown name, bare string, valid list covered; empty list not covered

brendancol (commented):
## Follow-up review (after empty-list fix)

The earlier nit is resolved: `_validate_stats_funcs` now rejects an empty `stats_funcs` up front with a clear ValueError instead of letting it fail later at `xr.concat`. A regression test covers it.

No further findings.

### Status
- Blockers: none
- Suggestions: none
- Nits: none (empty-list nit fixed)

Full focal_stats suite passes locally (31 tests).

PR #2799, opened 2026-06-01
Validate hotspots() kernel like apply()/focal_stats()

### LINKED ISSUE #2771: hotspots() skips kernel validation: bad kernels raise AttributeError or divide by zero
**Describe the bug**

`hotspots()` in `xrspatial/focal.py` skips the kernel validation that `apply()` and `focal_stats()` both run. Those two call `custom_kernel(kernel)` before touching the kernel. `hotspots()` doesn't. It goes straight to `_check_kernel_vs_raster_memory(kernel, ...)` near line 1509, which reads `kernel.shape` directly.

So an invalid kernel either crashes with a confusing `AttributeError` or silently produces wrong output, depending on what you pass.

**Failure modes (numpy backend)**

- `hotspots(agg, kernel=None)` raises `AttributeError: 'NoneType' object has no attribute 'shape'` instead of a clear validation error.
- `hotspots(agg, kernel=[[1, 1, 1]])` raises `AttributeError: 'list' object has no attribute 'shape'`. A list should get rejected with the same "not a Numpy array" message `custom_kernel` raises.
- `hotspots(agg, kernel=np.ones((2, 2)))` succeeds, even though even-dimensioned kernels are rejected by `custom_kernel` everywhere else.
- A zero-sum kernel (e.g. a Laplacian) divides by zero at line 1269: `convolve_2d(data, kernel / kernel.sum(), boundary)`. It emits RuntimeWarnings and returns all-zero output instead of raising.

**Expected behavior**

`hotspots()` should validate its kernel the way `apply()` and `focal_stats()` do: route through `custom_kernel()` so `None`, lists, and even-dimensioned kernels raise clear, consistent `ValueError`s. Zero-sum kernels should raise an explicit error rather than dividing by zero.

**Additional context**

`custom_kernel()` lives in `xrspatial/convolution.py`. It checks ndarray type and odd dimensions, but not the kernel sum. The zero-sum case is specific to `hotspots()` since it's the only focal function that divides by `kernel.sum()`.

brendancol:
## Summary

hotspots() skipped the kernel validation that apply() and focal_stats() run via custom_kernel(), so bad kernels produced confusing errors or wrong behavior:

- Run custom_kernel(kernel) before the memory check. None and list-of-list kernels now raise a clear ValueError instead of AttributeError on kernel.shape, and even-dimensioned kernels are rejected for improper shape (previously they silently succeeded).
- Add a zero-sum kernel guard. hotspots() is the only focal function that normalizes by kernel.sum(), so a zero-sum kernel previously divided by zero.

## Backend coverage

The validation runs before backend dispatch, so it applies to numpy, cupy, dask+numpy, and dask+cupy.

## Test plan

- [x] New regression tests for None, list-of-list, even-dim, and zero-sum kernels
- [x] Valid-kernel happy path still produces the expected classification
- [x] Full test_focal.py suite passes (155 passed)

Closes #2771

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -1505,6 +1505,14 @@ def hotspots(agg=None, kernel=None, name='hotspots', boundary='nan', *,
 
     _validate_boundary(boundary)
 
+    kernel = custom_kernel(kernel)
+    if kernel.sum() == 0:
+        raise ValueError(
+            "hotspots(): kernel sums to zero. The kernel is normalized by "
+            "its sum, so a zero-sum kernel divides by zero. Supply a kernel "
+            "with at least one non-zero cell."
+        )
+
     rows, cols = agg.shape[-2], agg.shape[-1]
     _check_kernel_vs_raster_memory(kernel, rows, cols, func_name='hotspots')
 
--- a/xrspatial/tests/test_focal.py
+++ b/xrspatial/tests/test_focal.py
@@ -776,6 +776,47 @@ def test_hotspots_zero_global_std():
         hotspots(agg, kernel)
 
 
+def test_hotspots_kernel_none_2771():
+    # Regression for #2771: hotspots skipped custom_kernel validation, so a
+    # None kernel raised AttributeError on kernel.shape instead of ValueError.
+    agg = create_test_raster(np.ones((10, 10), dtype=np.float32))
+    with pytest.raises(ValueError):
+        hotspots(agg, None)
+
+
+def test_hotspots_kernel_list_of_list_2771():
+    # Regression for #2771: a list-of-list kernel reached kernel.shape and
+    # raised AttributeError; it should be rejected as a non-ndarray.
+    agg = create_test_raster(np.ones((10, 10), dtype=np.float32))
+    with pytest.raises(ValueError):
+        hotspots(agg, [[1, 1, 1]])
+
+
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Validate hotspots() kernel like apply()/focal_stats()

None.

None.

- focal.py:1509 -- the guard catches exactly-zero sums. A mixed-sign kernel that sums to roughly zero but not exactly zero (floating-point near-cancellation) still slips through and produces a near-infinite normalization. This matches focal_stats() and is within the issue scope, so leaving it is fine.

### What looks good
- Validation runs before backend dispatch, so numpy, cupy, dask+numpy, and dask+cupy all reject bad kernels the same way. The 3D path recurses through hotspots() per band, so each band validates too.
- custom_kernel() runs before kernel.sum(), so the sum call always has an ndarray. No AttributeError risk.
- Error messages name the function and explain the failure.
- Tests cover None, list-of-list, even-dim, and zero-sum kernels plus a valid happy path. Full test_focal.py passes (155).

### Checklist
- [x] Algorithm unchanged; only input validation added
- [x] All backends get consistent validation (runs before dispatch)
- [x] NaN handling unchanged
- [x] Dask path unaffected (validation is pre-dispatch)
- [x] Docstrings present (existing hotspots docstring unchanged)

PR #2800, opened 2026-06-01
Fix GPU variety undercount for kernels larger than 5x5

### LINKED ISSUE #2775: GPU variety focal statistic silently undercounts unique values for kernels larger than 5x5
**Describe the bug**

The GPU `variety` focal statistic undercounts unique values for kernels larger than 5x5, and it does so silently. `_focal_variety_cuda` in `xrspatial/focal.py` (around line 908) tracks seen values in a per-thread local array sized for 25 unique values (`MAX_UNIQ = 25`). Once that buffer fills, any further distinct values are dropped. A 7x7 window where every cell is unique returns 25 on the GPU but 49 on the CPU path.

A code comment calls the cap an acceptable trade-off for GPU register pressure. For a statistic whose only job is counting distinct values, a wrong count with no warning is a correctness bug.

**Expected behavior**

GPU `variety` should return the same count as the CPU and dask paths for any kernel size. A 7x7 all-unique window should return 49 on all four backends (numpy, cupy, dask+numpy, dask+cupy).

**Steps to reproduce**

Run `focal_stats(agg, np.ones((7, 7)), stats_funcs=['variety'])` on a raster where every value in a 7x7 window is distinct, then compare the cupy backend against numpy. The cupy result caps at 25.

**Additional context**

The CPU implementation `_calc_variety` has no cap. The fix should make the GPU path match CPU for realistic kernel sizes without bringing back the register-pressure problem the cap was meant to avoid.

brendancol:
Closes #2775

## What changed

The CUDA variety kernel used a fixed 25-element `cuda.local.array`, so it silently capped unique-value counts at 25. A 7x7 all-unique window returned 25 on GPU versus 49 on CPU. For a correctness module that is wrong.

- Rewrote `_focal_variety_cuda` to count distinct values without a scratch buffer. For each valid non-NaN cell it scans only the earlier cells in the same window and increments the count when no earlier cell matches. O(window^2) per pixel, no `cuda.local.array`, so it works for arbitrary kernel sizes and matches the CPU `_calc_variety` exactly.
- The rewrite removes both the 25-value cap and the register-pressure concern the old buffer was sized for.

## Backend coverage

GPU only (cupy / dask+cupy go through this kernel). numpy and dask+numpy were already correct and are unchanged.

## Test plan

- [x] `test_variety_gpu_large_kernel_parity[7]` / `[9]`: cupy matches numpy on 7x7 (49) and 9x9 (81) all-unique windows. Verified on a real GPU.
- [x] `test_variety_large_kernel_numpy[7]` / `[9]`: numpy reference returns 49 and 81; runs without a GPU.
- [x] Full `test_focal.py` suite: 154 passed.

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -929,11 +929,14 @@ def _focal_variety_cuda(data, kernel, out):
             if v != v:  # NaN check (NaN != NaN)
                 continue
 
-            # Scan earlier kernel cells (flattened index < k * kcols + h).
+            # Scan earlier kernel cells (flattened index < target).
+            target = k * kcols + h
             first = True
             for pk in range(krows):
+                if pk * kcols >= target:
+                    break
                 for ph in range(kcols):
-                    if pk * kcols + ph >= k * kcols + h:
+                    if pk * kcols + ph >= target:
                         break
                     if kernel[pk, ph] == 0:
                         continue
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -905,34 +905,53 @@ def _focal_variety_cuda(data, kernel, out):
     dr = kernel.shape[0] // 2
     dc = kernel.shape[1] // 2
 
-    # Local buffer for up to 25 unique values (covers kernels up to 5x5).
-    # For larger kernels the buffer simply fills and stops counting,
-    # which is an acceptable trade-off for GPU register pressure.
-    MAX_UNIQ = 25
-    buf = cuda.local.array(MAX_UNIQ, nb.float32)
+    krows = kernel.shape[0]
+    kcols = kernel.shape[1]
+
+    # Count distinct non-NaN values without a scratch buffer: for each valid
+    # cell, scan only the earlier cells in the same window (row-major order).
+    # If none of them equals the current value, this is its first occurrence.
+    # This is O(window^2) per pixel but needs no cuda.local.array, so it works
+    # for arbitrary kernel sizes and matches the CPU implementation exactly.
     count = 0
 
-    for k in range(kernel.shape[0]):
-        for h in range(kernel.shape[1]):
+    for k in range(krows):
+        for h in range(kcols):
             if kernel[k, h] == 0:
                 continue
 
             ii = i + k - dr
             jj = j + h - dc
 
-            if 0 <= ii < rows and 0 <= jj < cols:
-                v = data[ii, jj]
-                if v != v:  # NaN check (NaN != NaN)
-                    continue
-                # check if already in buffer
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Fix GPU variety undercount for kernels larger than 5x5

None.

None.

- focal.py:934-951: the buffer-free scan is correct, but the inner double-break does a few no-op passes. Once the running flat index `pk*kcols+ph` reaches `k*kcols+h`, the `ph` loop breaks out of the current row, then every later `pk` row breaks again on its first iteration. No correctness impact. Comparing against a precomputed `target = k*kcols+h` would exit cleanly, but the current form is fine.
- benchmarks/benchmarks/focal.py:36: the focal_stats benchmark runs with default stats, so it never exercises the variety path. Not required for this fix, but a variety case would catch future regressions in this kernel's cost.

### What looks good
- The GPU count now matches the CPU `_calc_variety` exactly: each distinct value is counted once, at its earliest occurrence in the window. Verified on a real GPU (cupy == numpy for 7x7 -> 49 and 9x9 -> 81).
- NaN handling matches the CPU path (`v != v` skip, all-NaN window returns NaN).
- Dropping the cuda.local.array removes both the 25-value cap and the register-pressure reason it existed.
- Tests are split so the numpy reference is checked even without a GPU, and the cupy parity assertion is gated by the cuda_and_cupy_available marker.
- Full test_focal.py suite passes (154 tests).

### Checklist
- [x] Algorithm matches CPU reference
- [x] All implemented backends produce consistent results (cupy verified on GPU)
- [x] NaN handling is correct
- [x] Edge cases covered (existing all-NaN, single-cell tests still pass)
- [x] Dask chunk boundaries handled (dask+cupy focal test passes)
- [ ] Benchmark exists or is not needed (no variety-specific benchmark; not required)
- [x] Docstrings present and accurate (internal kernel, comment added)

brendancol (commented):
## Follow-up review (after 87100168)

The prior-cell scan now breaks the outer `pk` loop once `pk*kcols` reaches the target flat index (focal.py:935-937), so it stops at the row boundary instead of re-breaking on the first cell of each later row. This resolves the only actionable nit from the first pass. No behaviour change.

- Variety tests pass, including the on-GPU cupy/numpy parity cases for 7x7 (49) and 9x9 (81).
- flake8 clean on focal.py.

Remaining item, dismissed with reason:
- Variety-specific benchmark: out of scope for a correctness fix. A benchmark measures cost, not correctness, so it would not have caught the 25-value cap this PR removes. The focal_stats benchmark already exists for cost tracking.

No blockers or suggestions.

PR #2801, opened 2026-06-01
Fix contours() auto-levels silently dropping all contours with +/-inf

### LINKED ISSUE #2797: contours(): auto-levels silently drop all contours when the raster contains +/-inf
**Describe the bug**

When `contours()` picks levels automatically (`levels=None`), it computes the data range with `np.nanmin`/`np.nanmax` (and the dask/cupy equivalents), then guards only against NaN with `np.isnan`. If the raster mixes finite terrain values with `+inf` or `-inf` cells, `nanmin`/`nanmax` return `-inf`/`inf`. The NaN guard doesn't catch infinities, so `np.linspace(vmin, vmax, ...)` produces non-finite levels and the function returns no contours. Finite terrain that should have produced isolines just disappears, with no error.

**Expected behavior**

Auto-level generation should ignore non-finite extremes the way it already ignores NaN, so finite terrain still produces contours. Mask the infinities before computing the range (or fall back to the finite min/max).

**Reproduction**

```python
import numpy as np

data = np.tile(np.arange(10, dtype=np.float64), (5, 1))
data[0, 0] = np.inf
data[1, 0] = -np.inf
res = contours(agg, n_levels=5)
print(len(res))  # 0 -- expected several contour levels
```

**Additional context**

The affected code is the auto-level path in `xrspatial/contour.py` (around lines 624-644). The same nanmin/nanmax-then-isnan pattern runs in the numpy, cupy, and dask branches, so all backends are affected. Severity: high. This is silent data loss, not a crash.

brendancol:
## Summary

Fixes #2797. When `contours()` generates levels automatically (`levels=None`), a `+inf` or `-inf` cell mixed with finite terrain made `nanmin`/`nanmax` return infinity. The `np.isnan`-only guard let it through, and `np.linspace(vmin, inf, ...)` then produced non-finite levels, so the function returned no contours at all. Every contour for the finite terrain disappeared, with no error.

- Reduce the auto-level range over finite values only (treat `+/-inf` like NaN) in the numpy, cupy, and dask branches.
- Change the guard from `np.isnan` to `not np.isfinite` so an all-non-finite raster still returns an empty result cleanly.
- Silence the cosmetic "All-NaN slice" warning the all-non-finite case would otherwise raise.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy. The fix sits in the shared auto-level path before backend dispatch, so all four behave the same.

## Test plan

- [x] New `TestAutoLevelsInf` class: auto-levels ignore `+/-inf`, levels match the finite-only range, all-inf returns empty, explicit levels unaffected.
- [x] Cross-backend tests for numpy/dask/cupy/dask+cupy auto-levels with inf (cupy variants skip without a GPU).
- [x] Full `test_contour.py` suite passes (46 passed locally).

### DIFF
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -8,6 +8,7 @@
 # The algorithm is embarrassingly parallel across quads and across contour
 # levels, making it well suited to Dask chunking and GPU execution.
 
+import warnings
 from typing import TYPE_CHECKING, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Fix contours() auto-levels silently dropping all contours with +/-inf

None.

None.

- xrspatial/contour.py:649 -- the numpy branch builds a full-size copy with `np.where(np.isfinite(agg.values), agg.values, np.nan)`. Since the data is already in memory, the cost is one transient array of the same size that the old `np.nanmin(agg.values)` did not allocate. Fine to leave as-is: it keeps the three branches reading the same way, and the extra peak memory only matters for very large numpy rasters. Flagging so it is a conscious call rather than an oversight.
- xrspatial/tests/test_contour.py -- the file already has a `TestInfHandling` class from #2704 that exercises the kernel/explicit-level path. The new tests live in `TestAutoLevelsInf` to dodge the name clash, which is the right move. The two classes hit different code (auto-level range vs. per-quad interpolation), so the similar names are not a problem.

### What looks good
- The fix is in the shared pre-dispatch path, so one change covers numpy, cupy, dask+numpy, and dask+cupy identically.
- The dask branch stays lazy. `da.where` plus `da.nanmin`/`da.nanmax` under one `dask.compute`, same two-reduction shape as before, nothing newly materialized.
- Swapping the guard from `np.isnan` to `not np.isfinite` keeps the old all-NaN behavior (returns `[]`) and now covers all-inf too.
- The "All-NaN slice" RuntimeWarning from the all-non-finite case is suppressed in a tight scope around just that block, not globally.
- Tests cover the reported bug, a check that levels match the finite-only range, the all-inf empty case, explicit levels staying unaffected, and all four backends (cupy variants skip without a GPU).

### Checklist
- [x] NaN and inf handling is correct
- [x] All implemented backends produce consistent results (numpy + dask verified locally; cupy paths skip without a GPU)
- [x] Edge cases covered by tests (mixed inf, all-inf, explicit levels)
- [x] Dask path stays lazy, no premature materialization
- [x] Docstrings unchanged (no API change); contours already in reference docs
- [x] No README matrix change needed (no new function, no backend change)

PR #2803, opened 2026-06-01
Implement the Getis-Ord Gi* statistic in hotspots()

### LINKED ISSUE #2767: hotspots() does not compute the Getis-Ord Gi* statistic it advertises
## Describe the bug

`hotspots()` says it computes the Getis-Ord Gi* statistic ("statistically significant hot spots and cold spots"), but the implementation doesn't. What it actually does:

1. Convolves the raster with a normalized kernel (`kernel / kernel.sum()`), which gives a local weighted mean.
2. Z-scores that local mean against the global mean and global standard deviation of the raster.
3. Buckets the z-score into 90/95/99 confidence classes.

That's a local-mean z-score, not Gi*. The Gi* statistic for feature i is:

```
Gi* = (sum_j w_ij x_j - Xbar * sum_j w_ij) / (S * sqrt((n * sum_j w_ij^2 - (sum_j w_ij)^2) / (n - 1)))
```

Here Xbar is the global mean, S is the global population standard deviation, n is the count of valid features, and w_ij are the kernel weights (the star variant counts feature i itself). The current code normalizes the kernel and divides by the raw global std, so it drops the neighborhood weight sum, the squared weight sum, the sample count n, and the variance adjustment term entirely. Anything downstream that reads the output as Gi* gets wrong numbers.

The bad path lives at `xrspatial/focal.py` around line 1269 (the normalized convolution) and line 1278 (the global mean/std z-score), and the same formula is copied across the numpy, cupy, dask+numpy, and dask+cupy backends.

## Expected behavior

`hotspots()` should classify confidence bands from the actual Gi* z-score, not a local-mean z-score.

## Proposed fix

Compute the Gi* z-score per cell from the kernel weight sum, squared weight sum, valid feature count n, global mean, and global population std, on all four backends. Add tests that check the z-scores against a hand-computed reference for a small known window, not just the final classification.

brendancol:
Closes #2767

`hotspots()` claimed to compute the Getis-Ord Gi* statistic but computed a local-mean z-score instead: it convolved the raster with a normalized kernel, then z-scored that local mean against the global mean and std. That formula drops the neighborhood weight sum, the squared weight sum, the sample count n, and the variance adjustment term, so the classified output was statistically wrong.

What changed:

- Replace the local-mean z-score with the real Gi* z-score: `(sum_j w_ij x_j - Xbar * W_i) / (S * sqrt((n * W2_i - W_i^2) / (n - 1)))`.
- Compute `W_i` and `W2_i` by convolving the validity mask with the kernel and the squared kernel, so raster edges and interior NaN cells drop out of the neighborhood sums. `n`, `Xbar`, and `S` (population std) are global scalars over the valid cells.
- Guard `n < 2` (the variance term divides by `n - 1`).
- Update the docstring formula/example and the `focal.rst` NaN-handling note.

Backends: numpy, cupy, dask+numpy, dask+cupy all use the same Gi* math through the existing `ArrayTypeFunctionMapping` dispatch.

Test plan:

- [x] Gi* z-scores match a hand-computed reference on a small known 5x5 window (binary kernel).
- [x] Same against the reference for a non-binary (weighted) kernel, where `W_i != W2_i`.
- [x] NaN cells excluded from neighborhood sums and global stats; no NaN in the output.
- [x] `n < 2` raises.
- [x] numpy/dask parity (full interior; outer ring diverges by the documented boundary='nan' edge effect).
- [x] `pytest xrspatial/tests/test_focal.py` (full file), plus `test_emerging_hotspots.py`, `test_convolution.py`, `test_validation.py`, `test_dask_laziness.py` all pass.

Note: `emerging_hotspots()` carries the same local-mean z-score bug in its own code path. It is out of scope here (one fix per PR) and tracked in #2804.

### DIFF
--- a/docs/source/reference/focal.rst
+++ b/docs/source/reference/focal.rst
@@ -12,9 +12,9 @@ Focal
 
 .. note::
 
-   Focal functions output **float32**.  NaN handling varies: ``mean``
-   excludes NaN neighbours from the average, while ``hotspots`` propagates
-   NaN from any neighbour.
+   ``mean`` and ``focal_stats`` output **float32**; ``hotspots`` outputs
+   **int8** confidence bands. Both ``mean`` and ``hotspots`` exclude NaN
+   neighbours from their neighbourhood computation.
 
 Apply
 =====
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -1259,23 +1259,82 @@ def _calc_hotspots_numpy(z_array):
     return out
 
 
+def _gistar_global_stats(global_mean, global_std, n):
+    """Validate and return the global terms of the Gi* statistic.
+
+    ``global_mean`` is the mean of the valid (non-NaN) raster cells,
+    ``global_std`` their population standard deviation, and ``n`` the
+    count of valid cells. Gi* needs ``n > 1`` (the variance term divides
+    by ``n - 1``) and a non-zero spread.
+    """
+    if n < 2:
+        raise ValueError(
+            "hotspots() needs at least 2 valid (non-NaN) cells to "
+            "compute the Getis-Ord Gi* statistic."
+        )
+    if global_std == 0:
+        raise ZeroDivisionError(
+            "Standard deviation of the input raster values is 0."
+        )
+    return global_mean, global_std, n
+
+
+def _gistar_zscore(weighted_sum, weight_sum, sq_weight_sum,
+                   global_mean, global_std, n):
+    """Getis-Ord Gi* z-score from the per-cell convolution terms.
+
+    Gi* = (Sum_j w_ij x_j - Xbar * W_i)
+          / (S * sqrt((n * W2_i - W_i^2) / (n - 1)))
+
+    where ``weighted_sum`` is Sum_j w_ij x_j, ``weight_sum`` is W_i,
+    ``sq_weight_sum`` is W2_i, ``global_mean`` is Xbar, ``global_std`` is
+    the population std S, and ``n`` is the valid-cell count. Cells whose
+    neighborhood collapses to a single weight (variance term <= 0) get a
+    z-score of 0 (no significance).
+    """
+    numerator = weighted_sum - global_mean * weight_sum
+    variance_term = (n * sq_weight_sum - weight_sum * weight_sum) / (n - 1)
+    # Guard against tiny negatives from float rounding and the degenerate
+    # single-cell neighborhood (variance_term == 0) before the sqrt.
+    variance_term = np.where(variance_term > 0, variance_term, np.nan)
+    denominator = global_std * np.sqrt(variance_term)
+    z = numerator / denominator
+    return np.where(np.isfinite(z), z, 0.0).astype(np.float32)
+
+
+def _gistar_convolutions_numpy(data, kernel, boundary):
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Implement the Getis-Ord Gi* statistic in hotspots()

None.

- focal.py:1296 `_gistar_zscore` does its arithmetic in float32 (whatever the convolution returned) and casts at the end. For large `n` and large weights, `n * sq_weight_sum - weight_sum^2` is a difference of large numbers and can lose precision in float32. The hand-reference tests run in float64 and pass at rtol 1e-3 to 1e-4, so the tolerance hides it. Accumulating the variance term in float64 before the sqrt would be safer. Not blocking: the z thresholds (1.65/1.96/2.58) are coarse, so float32 rounding rarely flips a band.

- focal.py:1305 `_gistar_convolutions_numpy` is numpy-only and the tests import it directly, while the dask path inlines the same three convolutions in `_hotspots_chunk`. If those two spellings drift, the tests only catch the numpy one.
- The dask+numpy and dask+cupy paths each compute n/mean/std with three reductions in one `da.compute(...)`. It is a single pass, but the mask `.sum()` and `nanmean`/`nanstd` rebuild the NaN mask three times. Minor; not worth changing.

### What looks good
- Gi* z-scores checked against an independent hand-computed reference on a known window, for binary and weighted kernels, not just the classification banding.
- NaN cells drop out of the neighborhood sums and the global stats via a validity mask convolved with the kernel and the squared kernel. That is the correct treatment.
- All four backends share the same `_gistar_zscore` math through the existing dispatch. numpy, cupy, dask+numpy, and dask+cupy all pass on a real GPU here.
- `n < 2` and zero-std guards have clear errors.
- The numpy/dask edge divergence is the documented boundary='nan' artifact; the tests compare the interior, which matches the existing dask+cupy hotspots test.

### Checklist
- [x] Algorithm matches the reference formula
- [x] All backends agree on the interior
- [x] NaN handling is correct
- [x] Edge cases covered (NaN, constant surface, n<2, tiny raster)
- [x] Dask chunk boundaries handled (depth = kernel radius)
- [x] No premature materialization beyond the documented scalar pass
- [x] Benchmark: existing hotspots benchmark still applies; no new function
- [x] Docstrings present and accurate (formula added)

brendancol (commented):
## Follow-up review (after b36e43f8)

- Suggestion (float32 variance precision, focal.py:1296): fixed in b36e43f8. The weight sums and numerator are promoted to float64 before the sqrt, then the z-score is cast back to float32. `n * W2 - W^2` no longer loses precision for large rasters or weights. Banding is unchanged on the existing tests; interior z-scores now match the float64 reference to ~2e-7 (the residual is from the float32 convolution inputs, not the variance term).
- Nit (numpy-only `_gistar_convolutions_numpy` vs the inlined dask convolutions): dismissed. `test_hotspots_gistar_dask_matches_numpy_full` runs the dask inline path against the numpy reference on the full interior, so drift between the two spellings would fail a test.
- Nit (three reductions rebuild the NaN mask): dismissed, as flagged in the first pass. It is a single `da.compute` pass and the cost is negligible.

Re-ran test_focal, test_emerging_hotspots, test_validation, test_dask_laziness: 296 passed, including the GPU hotspots paths on a real device. No new findings.

PR #2805, opened 2026-06-01
Preserve input float dtype in apply() and focal_stats() (#2769)

### LINKED ISSUE #2769: apply() and focal_stats() silently downcast float64 input to float32
## Description

`apply()` and `focal_stats()` cast float64 input down to float32 on every backend. `convolve_2d()` already preserves the input floating dtype (via `_promote_float`) and has a test asserting this, but the focal functions don't, so a float64 raster silently comes back as float32.

The downcast happens in a few places in `xrspatial/focal.py`:

- `_apply_numpy` casts with `data.astype(np.float32)` (~line 462)
- `_apply_dask_numpy` casts with `data.astype(np.float32)` (~line 500)
- `_apply_cupy` casts with `data.astype(cupy.float32)` (~line 513)
- `_focal_stats_func_cupy` allocates the output as `dtype='f4'` (~line 985)
- `_focal_stats_cupy` and `_focal_stats_dask_cupy` cast with `data.astype(cupy.float32)` (~lines 1041, 1092)

## Expected behavior

`apply()` and `focal_stats()` should preserve the input floating dtype across all four backends (numpy, cupy, dask+numpy, dask+cupy): float64 in, float64 out, matching `convolve_2d()`. Integer input should still promote to at least float32.

## Why it matters

Silent precision loss is a correctness footgun for downstream numerical work, and it's inconsistent: `convolve_2d` already preserves float64 and has a test for it. Focal should behave the same way.

brendancol:
## Summary

`apply()` and `focal_stats()` cast every input to float32 internally, which silently downcast float64 rasters on all four backends (numpy, cupy, dask+numpy, dask+cupy). `convolve_2d()` already preserves the input floating dtype via the `_promote_float` helper, but focal did not.

- Route the focal backends through the same `_promote_float` helper so a float64 input yields a float64 output.
- Non-float inputs still promote to float32, matching the previous behavior.
- Covers `_apply_numpy`, `_apply_numpy_boundary`, `_apply_dask_numpy`, `_apply_cupy`, `_apply_dask_cupy`, `_focal_stats_func_cupy`, `_focal_stats_cupy`, and `_focal_stats_dask_cupy`. The CPU `focal_stats` path already routes through `apply`, so it inherits the fix.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy all preserve float64 end to end.

## Test plan

- [x] Added `test_apply_preserves_float64` and `test_focal_stats_preserves_float64`, parametrized over all four backends, mirroring the existing `convolve_2d` float64 coverage.
- [x] `pytest xrspatial/tests/test_focal.py` passes (158 tests, including the GPU paths on this machine).
- [x] flake8 clean on the changed lines.

Closes #2769

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -26,7 +26,7 @@ except ImportError:
         ndarray = False
 
 from xrspatial.convolution import (_available_memory_bytes, _convolve_2d_cupy, _convolve_2d_numpy,
-                                   convolve_2d, custom_kernel)
+                                   _promote_float, convolve_2d, custom_kernel)
 from xrspatial.dataset_support import supports_dataset
 from xrspatial.utils import (ArrayTypeFunctionMapping, _boundary_to_dask, _pad_array,
                              _validate_boundary, _validate_raster, _validate_scalar, cuda_args,
@@ -460,8 +460,7 @@ def _calc_variety(array):
 
 @ngjit
 def _apply_numpy(data, kernel, func):
-    data = data.astype(np.float32)
-
+    # Caller must promote ``data`` to a float dtype (see ``_promote_float``).
     out = np.zeros_like(data)
     rows, cols = data.shape
     krows, kcols = kernel.shape
@@ -483,6 +482,7 @@ def _apply_numpy(data, kernel, func):
 
 
 def _apply_numpy_boundary(data, kernel, func, boundary='nan'):
+    data = data.astype(_promote_float(data.dtype))
     if boundary == 'nan':
         return _apply_numpy(data, kernel, func)
     pad_h = kernel.shape[0] // 2
@@ -497,7 +497,7 @@ def _apply_numpy_boundary(data, kernel, func, boundary='nan'):
 
 
 def _apply_dask_numpy(data, kernel, func, boundary='nan'):
-    data = data.astype(np.float32)
+    data = data.astype(_promote_float(data.dtype))
     _func = partial(_apply_numpy, kernel=kernel, func=func)
 
     pad_h = kernel.shape[0] // 2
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Preserve input float dtype in apply() and focal_stats() (#2769)

None.

None.

- xrspatial/tests/test_focal.py: the new tests only assert float64 preservation. They don't cover the float32-stays-float32 case or the documented dask lazy-dtype quirk (a float32 dask input reports float64 lazily but computes to float32). This matches the existing convolve_2d coverage, so it's consistent, but a float32 case would pin down the full contract. Optional.

### What looks good
- The fix routes all six focal helpers through the same `_promote_float` helper that `convolve_2d` already uses, so `apply()` and `focal_stats()` now share one dtype contract instead of two.
- `_apply_numpy` dropped its in-kernel float32 cast and relies on the caller. Both callers (`_apply_numpy_boundary` and the `_apply_dask_numpy` partial) promote first, so the jitted function still sees a float input. `zonal.py` defines its own separate `_apply_numpy` with a different signature, so there's no collision.
- The CPU `focal_stats` path delegates to `apply()`, so it inherits the fix without a separate change.
- Tests run on all four backends, and the cupy / dask+cupy variants actually execute here (GPU present) rather than skip.

### Checklist
- Algorithm matches reference/paper: n/a, dtype-only change
- All implemented backends produce consistent results: yes, float64 preserved on all four
- Edge cases covered by tests: float64 covered on all backends
- Dask chunk boundaries handled correctly: unchanged, ragged (2,3) chunks used
- Benchmark exists or not needed: not needed, dtype-only fix
- README feature matrix updated: n/a, no new function or backend change
- Docstrings present and accurate: yes; the apply() docstring example already showed float64 output and now matches reality

brendancol (commented):
## PR Review (follow-up): Preserve input float dtype in apply() and focal_stats() (#2769)

Re-reviewed after the follow-up commit `c3b186b4`.

### Blockers
None.

### Suggestions
None.

### Nits
None remaining. The earlier nit (no float32 case) is addressed by `test_apply_keeps_float32`, which asserts a float32 input computes to float32 on all four backends and documents the dask lazy-dtype quirk in a comment.

### What looks good
- Both sides of the dtype contract are now covered: float64 stays float64, float32 stays float32, across numpy / cupy / dask+numpy / dask+cupy.
- Full `test_focal.py` suite passes locally (162 tests), with the GPU variants actually executing here.

This review is informational; rockout does not approve or block its own PRs.

PR #2811, opened 2026-06-01
Fix proximity Great Circle docstrings: meters, not kilometers

### LINKED ISSUE #2810: proximity docstrings say Great Circle max_distance is in kilometers, but implementation returns meters
## Description

The `proximity()`, `direction()`, and `allocation()` docstrings in `xrspatial/proximity.py` say the `max_distance` parameter is in kilometers when using the Great Circle metric:

> If using Great Circle distance metric, and thus all distances is in km, `max_distance` should also be provided in kilometer unit.

But the implementation returns meters. `great_circle_distance()` uses Earth radius `6378137` (the WGS84 equatorial radius, in meters) and returns `radius * 2 * arcsin(sqrt(a))`. The function's own docstring example confirms it: two Arctic points return `2378290.489801402`, which is meters, not kilometers.

So a user who trusts the kilometer wording and passes `max_distance` in kilometers will be off by a factor of 1000, and the result will be silently wrong.

## Scope

Docs only. The meters behavior is correct and established, so the computation stays as-is. The fix is the three docstring blocks. The reference page `docs/source/reference/proximity.rst` already says "returns metres" and needs no change.

## Expected

Docstrings say Great Circle `max_distance` is in meters.

## Actual

Docstrings say kilometers, contradicting the implementation.

brendancol:
Closes #2810

The `proximity()`, `direction()`, and `allocation()` docstrings said `max_distance` for the Great Circle metric is in kilometers. The implementation returns meters: `great_circle_distance()` uses Earth radius `6378137` (the WGS84 equatorial radius, in meters) and returns `radius * 2 * arcsin(sqrt(a))`. Trusting the kilometer wording put a user's `max_distance` off by 1000x.

Changes:
- Corrected the three docstring blocks to say meters instead of kilometers.
- Noted the unit on `great_circle_distance`'s `radius` parameter and return value.
- Added `test_great_circle_distance_returns_meters`: one degree of longitude at the equator is ~111319.49 m (it would be ~111 in km), which pins the unit down.

Docs only, no computation change. The reference page `docs/source/reference/proximity.rst` already says "returns metres", so it needed no edit.

Backends: not applicable. This touches docstrings plus one pure-Python unit test; no backend code paths changed.

Test plan:
- [x] `pytest xrspatial/tests/test_proximity.py -k great_circle_distance` passes (2 passed).


brendancol (commented):
## PR Review: Fix proximity Great Circle docstrings: meters, not kilometers

None.

None.

- xrspatial/tests/test_proximity.py: the first assertion (`dist == pytest.approx(expected_meters)`) re-derives the value from the same formula the implementation uses, so it mostly checks the code against itself. The literal `111319.49` assertion is the one that actually catches a unit regression. Both are fine to keep; the literal is the load-bearing one.

### What looks good
- The fix matches the verified behavior. `great_circle_distance()` uses radius 6378137 (WGS84 equatorial radius, meters) and returns `radius * 2 * arcsin(sqrt(a))`, which is meters. The existing docstring example output `2378290.489801402` confirms it.
- All three docstring blocks (proximity, allocation, direction) are corrected. A grep shows no kilometer claims left in the module.
- `docs/source/reference/proximity.rst` already said "returns metres", so it was correctly left alone.
- Test passes. One degree of longitude at the equator is ~111319.49 m, versus ~111 in km, so the unit is unambiguous.

### Checklist
- Algorithm matches reference: n/a (docs only), computation unchanged
- Backend parity: n/a, no backend code touched
- NaN handling: n/a
- Edge cases covered: unit pinned by the new test
- Dask chunk boundaries: n/a
- Benchmark: not needed (docs only)
- README feature matrix: not needed (no new function, no backend change)
- Docstrings present and accurate: yes, which is the point of this PR

brendancol (commented):
## Follow-up review

The nit from the first pass is addressed. The meters test now keeps only the fixed literal assertion (~111319.49 m for one degree of longitude at the equator), so it no longer checks the implementation against its own formula. Tests still pass (2 passed).

No new findings. Nothing else outstanding.

PR #2813, opened 2026-06-01
Validate negative max_distance in proximity/allocation/direction

### LINKED ISSUE #2808: Negative max_distance not validated, produces backend-inconsistent results in proximity/allocation/direction
**Describe the bug**

A negative `max_distance` passed to `proximity`, `allocation`, or `direction` is not validated. Instead of raising an error, it produces backend-dependent results.

On the NumPy path, `max_distance` is squared (`xrspatial/proximity.py:555`, `near_distance_square = max_distance ** 2 * 2.0`), so a negative value becomes positive and some neighbors are still returned. The CUDA kernel compares `best_dist <= max_distance` directly (`xrspatial/proximity.py:371`), and since `best_dist` is non-negative, a negative `max_distance` excludes every pixel. The same input gives different output depending on the backend.

**Expected behavior**

A negative `max_distance` should raise a clear `ValueError` at the public API boundary on every backend, before any backend-specific computation runs.

The edge values keep their current meaning: `max_distance=0` (only target cells qualify) and the `np.inf` / `None` default (no distance limit) stay unchanged. Only negatives are the bug.

**Additional context**

The three public functions (`proximity`, `allocation`, `direction`) all funnel through `_process` in `xrspatial/proximity.py`, which already normalizes `None` to `np.inf`. That is the natural spot for the validation so all backends share it.

brendancol:
## Summary

Closes #2808.

- Validate `max_distance` at the public API boundary in `_process` (shared by `proximity`, `allocation`, and `direction`) so a negative value raises a `ValueError` before any backend runs.
- Previously a negative `max_distance` gave backend-dependent output: the NumPy path squared it (`max_distance ** 2`) and still returned neighbors, while the CUDA path compared `best_dist <= max_distance` directly and excluded every pixel.
- Edge values keep their meaning: `max_distance=0` (only target cells qualify) and the `np.inf` / `None` default (no limit) are unchanged. Only negatives are rejected.

## Backend coverage

The check lives in `_process`, which runs before backend dispatch, so it covers numpy, cupy, dask+numpy, and dask+cupy.

## Test plan

- [x] New `test_negative_max_distance_raises`, parametrized over the three public functions, three negative values, and all four backends.
- [x] New `test_zero_max_distance_keeps_meaning` confirms `max_distance=0` still works on every backend.
- [x] Full `test_proximity.py` suite passes (217 tests).


brendancol (commented):
## PR Review: Validate negative max_distance in proximity/allocation/direction

None.

None.

- xrspatial/proximity.py:1058 -- The check sits right after the `None -> np.inf` normalization, the correct chokepoint. A `nan` max_distance would slip through (`nan < 0` is False). nan was never a supported input and the finding only covers negatives, so this is not required, but rejecting nan in the same guard is cheap and removes another silent-garbage path. Noting it for consideration.

### What looks good
- The validation lives in `_process`, the single function all three public entry points (`proximity`, `allocation`, `direction`) funnel through, and it runs before backend dispatch. One check covers numpy, cupy, dask+numpy, and dask+cupy.
- Edge values are preserved as the finding required: `max_distance=0` and the `np.inf` / `None` default are untouched. The new test pins the `0` case so it cannot regress.
- Test coverage is thorough: negative values parametrized across all three functions, three negative inputs (`-1`, `-0.5`, `-inf`), and all four backends. The 217-test suite still passes.
- The error message includes the offending value, which makes the misuse obvious.
- Docstrings for all three functions were updated to state the non-negative requirement.

### Checklist
- All implemented backends produce consistent results: yes, every backend now raises on negative input
- Edge cases are covered by tests: yes, negatives and the 0 boundary
- No premature materialization or unnecessary copies: yes, scalar comparison only

brendancol (commented):
## Follow-up review (after commit 909db76)

The one open nit from the previous pass is resolved.

- xrspatial/proximity.py:1058 now guards `np.isnan(max_distance) or max_distance < 0`, so a NaN `max_distance` raises the same ValueError instead of falling through to backend-specific garbage (numpy would yield all-NaN, the CUDA path would exclude every pixel). The test parametrization gained a `np.nan` case and the function was renamed to `test_invalid_max_distance_raises`. Docstrings updated to say non-negative and non-NaN.

No remaining Blockers, Suggestions, or Nits. The proximity suite passes (validation tests: 60 passed including the new NaN case).

PR #2814, opened 2026-06-01
Raise on invalid distance_metric in proximity/allocation/direction

### LINKED ISSUE #2807: Invalid distance_metric silently falls back to Euclidean in proximity/allocation/direction
**Describe the bug**

In `xrspatial/proximity.py`, `_process` normalizes the `distance_metric` argument with a dictionary `.get()` that falls back to Euclidean when the key is not found:

```python
distance_metric = DISTANCE_METRICS.get(distance_metric, None)
if distance_metric is None:
    distance_metric = DISTANCE_METRICS["EUCLIDEAN"]
```

So a typo like `distance_metric="euclidian"` or `"chebyshev"` is silently treated as `"EUCLIDEAN"`. The call succeeds and returns plausible-looking results, so the mistake goes unnoticed. This affects `proximity`, `allocation`, and `direction`, since all three route through `_process`.

**Expected behavior**

An unrecognized `distance_metric` should raise a `ValueError` that names the bad value and lists the valid options (`EUCLIDEAN`, `GREAT_CIRCLE`, `MANHATTAN`). Valid metrics should keep working exactly as before.

**Additional context**

The valid keys live in `DISTANCE_METRICS` (lines 41-52). The fix is localized to the normalization in `_process` (around line 1049), which covers all three public entry points and every backend.

brendancol:
Closes #2807

`_process` in `xrspatial/proximity.py` normalized `distance_metric` with `DISTANCE_METRICS.get(metric, None)` and fell back to EUCLIDEAN on a miss. A typo like `distance_metric="euclidian"` ran to completion with the wrong metric and gave no signal that anything was off.

- Replace the silent fallback with a `ValueError` that names the bad value and lists the valid options (EUCLIDEAN, GREAT_CIRCLE, MANHATTAN).
- The check sits at the top of `_process`, so it covers `proximity`, `allocation`, and `direction`. It runs before backend dispatch, so it raises eagerly rather than deferring into a dask graph.
- Updated the three public docstrings to note that an unrecognized value raises.

Behavior note: the old `.get()` fallback also caught `None`. The new check raises on `None` as well, since `None not in DISTANCE_METRICS`. All three public signatures default to the string `'EUCLIDEAN'`, so a `None` metric only reaches `_process` if a caller passes it explicitly.

Backends: the validation is backend-agnostic, running ahead of the numpy / cupy / dask+numpy / dask+cupy split.

Test plan:
- [x] `test_invalid_distance_metric_raises`: a bogus metric raises ValueError for proximity, allocation, and direction; the message names the value and lists valid options.
- [x] `test_valid_distance_metric_runs`: EUCLIDEAN, GREAT_CIRCLE, and MANHATTAN still run for all three entry points.
- [x] Confirmed the new tests fail on the pre-fix code (DID NOT RAISE) and pass after.
- [x] Full `test_proximity.py` suite passes (181 tests).

CHANGELOG left untouched on purpose.


brendancol (commented):
## PR Review: Raise on invalid distance_metric in proximity/allocation/direction

### Blockers
None.

### Suggestions
- `xrspatial/proximity.py:1049`: the old code fell back to EUCLIDEAN when `distance_metric` was `None`. The new code raises on `None` too, since `None not in DISTANCE_METRICS`. All three public signatures default to the string `'EUCLIDEAN'`, so `None` only reaches `_process` if a caller passes it explicitly. That is an acceptable behavior change for a validation fix, but worth a one-line note in the PR body so it is not a surprise.

### Nits
- `xrspatial/tests/test_proximity.py:55`: `test_valid_distance_metric_runs` asserts only `result.shape`. Per-metric value correctness is already covered by `test_manhattan_proximity` and `test_great_circle_proximity`, so this works as a "not rejected" smoke check. No change needed; noting it so the intent is clear.
- The new tests run on numpy only. The validation sits ahead of backend dispatch in `_process`, so numpy coverage is enough to prove the raise fires for every backend. Parametrizing over backends would be redundant here.

### What looks good
- The check sits at the top of `_process`, before the numpy/cupy/dask split, so it raises eagerly for all four backends and all three entry points without touching each path.
- The error message names the bad value and lists valid options via `sorted(DISTANCE_METRICS)`.
- TDD confirmed: tests fail (DID NOT RAISE) on the pre-fix code and pass after.
- Docstrings updated for all three entry points.

### Checklist
- [x] Validation runs before backend dispatch (covers numpy/cupy/dask/dask+cupy)
- [x] Valid-metric paths unchanged (existing value tests still pass)
- [x] New tests fail before fix, pass after
- [x] Docstrings updated
- [ ] README feature matrix: not applicable (no new function, no backend change)

brendancol (commented):
## Review follow-up

Disposition of the prior review:

- Suggestion (note the `None` behavior change): fixed. The PR body now documents that the new check raises on `None`, where the old `.get()` fallback would have returned EUCLIDEAN. No code change is needed since all three signatures default to the `'EUCLIDEAN'` string.
- Nit (shape-only assertion in `test_valid_distance_metric_runs`): dismissed. Per-metric value correctness is already covered by `test_manhattan_proximity` and `test_great_circle_proximity`; this test only needs to confirm valid metrics are not rejected.
- Nit (numpy-only tests): dismissed. Validation runs before backend dispatch, so numpy coverage proves the raise fires for every backend.

No code changes resulted from the follow-up, so the diff is unchanged.

PR #2815, opened 2026-06-01
Fix dask proximity halo depth on irregular coords and 1xN/Nx1 rasters

### LINKED ISSUE #2809: Dask finite-distance proximity: wrong results on irregular coords and crash on 1xN/Nx1 rasters
**Describe the bug**

Two related bugs live in the Dask finite-distance path of `proximity` (and the `allocation` / `direction` functions that share it). Both trace back to the halo-depth computation that sizes each chunk's overlap. The depth is taken from only the first row/column coordinate spacing, in `xrspatial/proximity.py` near lines 1251 and 437.

Bug 1: wrong results on irregular coordinates. When later coordinates sit closer together than the first pair, the halo is too thin, so chunks miss valid targets just past the chunk boundary. Cells the NumPy backend marks as in-range come back NaN under Dask.

Repro: `x = [0, 100, 101, 102, 103, 104, 105, 106]` (first gap 100, the rest 1), one target, chunks `(4, 2)`, `max_distance=2.5`. NumPy returns 21 in-range cells, Dask returns 6.

Bug 2: crash on single-row or single-column rasters. The same code indexes `y_coords[1]` and `x_coords[1]`. A `(1, N)` or `(N, 1)` raster has only one coordinate along that axis, so the index raises `IndexError`. NumPy handles these shapes fine.

**Expected behavior**

The Dask finite-distance path should match the NumPy backend on irregular coordinates and should not crash on `(1, N)` or `(N, 1)` rasters. The halo depth should come from the densest (smallest) spacing along each axis, and an axis with a single coordinate should add no halo along that axis.

**Additional context**

Both the dask+numpy path (`_process_dask`, near line 1251) and the dask+cupy path (`_process_dask_cupy`, near line 437) are affected. Same root cause in both.

brendancol:
Closes #2809

Two related bugs in the Dask finite-distance path of `proximity` / `allocation` / `direction`, both from the halo-depth computation that sizes each chunk's `map_overlap` padding.

- Bug 1: the depth was taken from only the first row/column coordinate spacing. On irregular coordinates where later cells are packed more tightly, the halo came out too thin and chunks dropped valid targets just past their boundary, so in-range cells turned NaN under Dask.
- Bug 2: the same code indexed `x_coords[1]` / `y_coords[1]`, which raises `IndexError` on `(1, N)` and `(N, 1)` rasters where one axis has a single coordinate. NumPy handled those shapes fine.

The fix pulls the depth math into one `_halo_depth` helper used by both the dask+numpy and dask+cupy paths. It sizes the padding from the densest (smallest positive) spacing along each axis, and treats a single-coordinate axis as zero halo along that axis.

Backends: dask+numpy and dask+cupy (the two paths that build the halo). numpy and cupy are unaffected.

Test plan:
- [x] New regression test: irregular coords (x = [0, 100, 101, ...]), chunks (4, 2), max_distance=2.5, dask matches numpy for proximity / allocation / direction.
- [x] New regression test: (1, N) and (N, 1) rasters on the bounded dask path no longer crash and match numpy.
- [x] Full test_proximity.py suite passes (178 tests).


brendancol (commented):
## PR Review: Fix dask proximity halo depth on irregular coords and 1xN/Nx1 rasters

None.

- GREAT_CIRCLE column spacing is still measured at a single latitude (`y_coords[0]`), in `_halo_depth` (xrspatial/proximity.py:283-285). East-west ground distance per degree of longitude shrinks toward the poles, so over a wide latitude range the densest column spacing in metres can land on a different row than the first. That can still under-pad at high latitudes. This is pre-existing behavior (the old code used `y_coords[0]` too) and the PR does not make it worse, but now that we size from the densest spacing it would be consistent to scan latitudes as well. Reasonable to leave for a follow-up given the bug report's scope.

- `_min_step_distance` builds a list of every step then calls `min` (xrspatial/proximity.py:276-290). For large axes that is an O(N) Python loop over the non-jitted `_distance`. It runs once per call rather than per chunk, so it is off the hot path, but a running minimum would skip the list allocation.

### What looks good
- The duplicated halo math in `_process_dask` and `_process_dask_cupy` collapses into one helper, so both dask paths get the fix.
- A single-coordinate axis returns depth 0, which is correct for 1xN / Nx1, and the regression tests pin both bugs on proximity, allocation, and direction.
- Sizing from the smallest positive step only ever grows the depth versus the old first-pair value, so no previously-correct case loses padding.

### Checklist
- [x] Algorithm matches intended behavior (densest-spacing halo)
- [x] Dask backends produce results consistent with numpy
- [x] NaN handling unchanged and correct
- [x] Edge cases (1xN, Nx1, irregular coords) covered by tests
- [x] No premature materialization (coords are 1D axis arrays, computed once)
- [x] README feature matrix not applicable (no new function or backend change)
- [x] Docstring present on the new helper

brendancol (commented):
## Follow-up review (after 3d820ac)

Both items from the first pass are addressed in the new commit:

- The great-circle column spacing now uses the worst-case latitude. `_min_step_distance` measures east-west steps at the highest-|latitude| row (`y_worst = max(y_coords, key=abs)`), so the halo no longer under-pads where meridians converge near the poles. North-south spacing stays at a fixed longitude since it does not vary with longitude (xrspatial/proximity.py:273-298).
- `_min_step_distance` tracks a running minimum instead of building a list (xrspatial/proximity.py:281-291).

Coords reaching `_halo_depth` are already computed to numpy at the call site (proximity.py:1107-1109), so `max(y_coords, key=abs)` iterates numpy scalars, not a lazy array.

Verification: full `test_proximity.py` passes (178 tests), and the irregular-coords and 1xN/Nx1 regressions still match the numpy backend.

No open findings remain.

PR #2816, opened 2026-06-01
Fix GREAT_CIRCLE proximity on the CPU line-sweep backend

### LINKED ISSUE #2812: GREAT_CIRCLE proximity is wrong on the CPU (numpy) line-sweep backend
**Describe the bug**

`proximity(..., distance_metric='GREAT_CIRCLE')` returns wrong distances on the CPU (numpy) backend. The numpy path uses the GDAL-style line-sweep in `_process_numpy` / `_process_proximity_line` (xrspatial/proximity.py:1073). That sweep carries a single nearest-target candidate from one pixel to its neighbors, which assumes the nearest-target distance changes smoothly between adjacent cells. That holds for Euclidean and Manhattan, but not for great-circle distance.

Great-circle distance is global. It wraps in longitude, and meridians converge as latitude increases, so the true nearest target to a pixel often can't be reached through the chain of neighbor-propagated candidates. The sweep then locks onto a farther target and reports the wrong distance.

This silently affects `proximity`, `allocation`, and `direction`, since allocation and direction both read out of the same nearest-target bookkeeping.

**Reproduction**

I ran a random search comparing the line-sweep against a brute-force nearest-target reference (the reference calls the same `great_circle_distance` per pair). Lots of configurations disagree, with errors past 1 million metres. The worst gap in a 300-trial sweep was about 5.9 million metres. The failures show up on rasters that span a wide longitude range, which is exactly where the longitude wrap and meridian convergence break the local-planarity assumption.

```python
import numpy as np
import xarray as xr
from xrspatial.proximity import great_circle_distance

W, H = 11, 6
lon = np.linspace(-127.2, 173.3, W)
lat = np.linspace(21.2, 19.9, H)
data = np.zeros((H, W))
data[0, 0] = 1.0
data[H - 1, W - 1] = 2.0
r = xr.DataArray(data, dims=['lat', 'lon'])
r['lon'] = lon
r['lat'] = lat
gc = proximity(r, x='lon', y='lat', distance_metric='GREAT_CIRCLE').data
# compare gc against a brute-force nearest-target reference -> off by ~1e6 m
```

**Expected behavior**

CPU GREAT_CIRCLE proximity, allocation, and direction should match a brute-force nearest-target reference within floating-point tolerance, no matter how wide the raster's longitude span is. The GPU (cupy) path already uses a brute-force nearest-target kernel and gets this right; the numpy path should agree with it.

**Additional context**

The line-sweep is fine for Euclidean and Manhattan, where the nearest-target field is locally monotonic. The fix is to drop the line-sweep for GREAT_CIRCLE on the numpy backend and do an exact nearest-target search instead.

brendancol:
Closes #2812

The numpy proximity path used the GDAL line-sweep, which carries a single nearest-target candidate between adjacent pixels. That assumption holds for EUCLIDEAN and MANHATTAN but breaks for GREAT_CIRCLE, where distance wraps in longitude and meridians converge with latitude. On wide-longitude rasters the sweep locked onto a farther target. A brute-force-vs-line-sweep sweep found errors up to ~5.9 million metres.

- Route GREAT_CIRCLE on the CPU through an exact brute-force nearest-target search, the same one the CUDA kernel already runs. EUCLIDEAN and MANHATTAN keep the fast line-sweep.
- The dispatch lives in plain Python, not inside the numba kernel, so the metric choice is never frozen into a cached closure compilation.

Backends: fixes numpy and dask+numpy. cupy and dask+cupy already used the brute-force kernel and were correct; the new tests cover all four.

Test plan:
- [x] New `test_great_circle_matches_brute_force` compares proximity/allocation/direction against a brute-force reference on a wide-longitude raster (numpy, dask+numpy; cupy/dask+cupy when CUDA is present).
- [x] New `test_great_circle_numpy_off_by_more_than_a_metre_is_fixed` fails on the old code (off by ~5.9e6 m) and passes with the fix.
- [x] Full `test_proximity.py` suite: 182 passed.


brendancol (commented):
## PR Review: Fix GREAT_CIRCLE proximity on the CPU line-sweep backend

### Blockers
None.

### Suggestions
- `_process_numpy_bruteforce` is O(H*W*T) per chunk, where T is the target count visible in the (padded) chunk. For the bounded dask path that stays small, but for unbounded GREAT_CIRCLE on a single chunk with many targets it can be slow. This matches what the CUDA kernel already does and is the price of correctness, so it is fine to merge, but worth a one-line note in the proximity docstring that GREAT_CIRCLE on the CPU is brute-force. (xrspatial/proximity.py:293)

### Nits
- The target-detection predicate is written out twice in `_process_numpy_bruteforce` (the counting pass at lines 311-321 and the fill pass at lines 330-345). Pulling it into a small `@ngjit` `_is_target(v, target_values)` helper would remove the duplication. Minor; the current form is readable. (xrspatial/proximity.py:311, 330)

### What looks good
- The dispatch lives in plain Python (`_process_numpy`) rather than inside the numba kernel, so the metric branch is never frozen into a cached closure. That avoids a subtle recompilation trap.
- NaN handling is correct: padded cells (`boundary=np.nan`) and NaN data are excluded from targets via `np.isfinite`, and NaN query pixels leave the output NaN.
- The regression test fails on the pre-fix code (off by ~5.9e6 m) and passes after, which pins the bug rather than just exercising it.
- EUCLIDEAN and MANHATTAN keep the fast line-sweep; the change is scoped to GREAT_CIRCLE.

### Checklist
- [x] Algorithm matches reference (brute-force nearest target, same as CUDA kernel)
- [x] All implemented backends produce consistent results (tests cover all four)
- [x] NaN handling is correct
- [x] Edge cases covered (degenerate-shape tests still pass; wide-longitude raster added)
- [x] Dask chunk boundaries handled (existing bounded-dask GC parity test passes)
- [x] Benchmark exists or not needed (existing proximity benchmarks cover it)
- [x] README feature matrix (n/a; no new function, no backend change)

brendancol (commented):
## Follow-up review (after 3175b497)

- Suggestion (docstring note that the NumPy GREAT_CIRCLE path is brute-force): fixed in the `proximity` docstring.
- Nit (duplicated target-detection predicate in `_process_numpy_bruteforce`): fixed by extracting an `_is_target_value` ngjit helper used by both the count and fill passes.

No new findings. Full `test_proximity.py` suite still passes (182 tests), `proximity.py` is flake8-clean.

PR #2817, opened 2026-06-02
polygonize: optimize visvalingam_whyatt with min-heap

### LINKED ISSUE #2539: polygonize: _visvalingam_whyatt is O(n^2) due to linear min-area scan
## What's happening

`_visvalingam_whyatt` in `xrspatial/polygonize.py` (L1217-1316) is O(n²)
in the number of input vertices. The main loop at L1262 does a linear
scan over all vertices to find the minimum-area one on every iteration:

```python
while remaining > 2:
    # Find vertex with minimum area.
    min_area = np.inf
    min_idx = -1
    for i in range(1, n - 1):
        if not removed[i] and areas[i] < min_area:
            min_area = areas[i]
            min_idx = i
    ...
```

For each removed vertex we walk all n entries again, so total work is
O(n²). On a ring with a few thousand vertices this dominates the
simplification step.

## Fix

Use a min-heap keyed on triangle area. Each removal is O(log n), and
the two affected neighbours get re-pushed with updated areas. Lazy
deletion via a `removed[]` flag (already present) handles stale heap
entries:

```python
import heapq
heap = [(areas[i], i) for i in range(1, n - 1)]
heapq.heapify(heap)
while remaining > 2:
    while heap:
        area, idx = heapq.heappop(heap)
        if not removed[idx] and area == areas[idx]:
            break
    else:
        break
    if area >= tolerance:
        break
    removed[idx] = True
    remaining -= 1
    # Recompute neighbour areas and push.
    ...
```

Numba supports `heapq` only with the `numba-scipy` extension, so the
heap version may need to drop `@ngjit` or move the simplification to
plain Python with a numba inner kernel for the area math. Worth
checking whether the constant-factor loss outweighs the asymptotic win
on typical inputs (rings with <100 vertices).

## Severity

MEDIUM. Only hits the visvalingam-whyatt code path
(`simplify_method="visvalingam-whyatt"`), not the default
douglas-peucker. Surfaces on long rings, e.g. the exterior of a large
connected region after polygonize.

Melissari1997:
Closes #2539

# Summary
Replace the O(n²) linear minimum-area scan in `_visvalingam_whyatt` with a binary min-heap keyed on triangle area, achieving **O(n log n)** scaling for line/polygon simplification.

## Why this solves the performance problem
The original `_visvalingam_whyatt` iteratively removes the vertex with the smallest triangle area
On each iteration it scans all remaining interior vertices to find the minimum — an O(n²) cost. 
For polygon rings with tens of thousands of vertices (common in high-resolution rasters), this quadratic scan dominates runtime.
This PR replaces that linear scan with a binary min-heap storing (area, index) pairs:

- Heap operations are O(log n): `_heap_push` and `_heap_pop` sift up/down in logarithmic time, so each of the n removals costs **O(log n)** instead of **O(n)**.
- Lazy deletion handles stale entries: After a vertex is removed, its neighbors' areas change. Rather than updating arbitrary heap elements (which would require an index map), the new area is simply `_heap_pushed` as a fresh entry. When the stale entry is popped later, the removed[idx] flag or a mismatched area tells us to skip it and pop the next one.
- Overall complexity drops from **O(n²)** to **O(n log n)**: Building the initial heap is **O(n)**, each removal is **O(log n)**, and each neighbor recomputation pushes **O(1)** entries — **totaling O(n log n).**

### DIFF
--- a/benchmarks/benchmarks/polygonize.py
+++ b/benchmarks/benchmarks/polygonize.py
@@ -3,7 +3,7 @@ import xarray as xr
 
 from xrspatial import polygonize
 from xrspatial.polygonize import (
-    _calculate_regions, _scan,
+    _calculate_regions, _scan, _simplify_polygons,
 )
 
 try:
@@ -122,3 +122,22 @@ class PolygonizeTracing:
     def time_scan(self, nx):
         _scan(self.regions, self.values, self.mask, False, None,
               self.nx, self.ny)
+
+
+class PolygonizeSimplify:
+    """Benchmark polygonize with simplification (DP and VW)."""
+    params = (
+        [100, 300, 1000],
+        ["douglas-peucker", "visvalingam-whyatt"],
+    )
+    param_names = ("nx", "method")
+
+    def setup(self, nx, method):
+        ny = nx // 2
+        rng = np.random.default_rng(9461713)
+        raster = rng.integers(low=0, high=4, size=(ny, nx), dtype=np.int32)
+        self.raster = xr.DataArray(raster)
+
+    def time_simplify(self, nx, method):
+        polygonize(self.raster, simplify_tolerance=1.0,
+                   simplify_method=method)
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -1809,7 +1809,8 @@ def _heap_push(heap_areas, heap_indices, heap_size, area, idx):
 def _heap_pop(heap_areas, heap_indices, heap_size):
     """Pop minimum from heap. Returns (area, idx, new_heap_size).
 
-    Returns (-1.0, -1, heap_size) if heap is empty.
+    Returns (-1.0, -1, heap_size) if heap is empty.  -1.0 is safe
+    as a sentinel because triangle areas are always non-negative.
     """
     if heap_size == 0:
         return -1.0, -1, heap_size
@@ -1870,16 +1871,19 @@ def _visvalingam_whyatt(coords, tolerance):
         areas[i] = abs((ax * (by - cy) + bx * (cy - ay) + cx * (ay - by)) / 2.0)
 
     removed = np.zeros(n, dtype=np.bool_)
-    remaining = n
 
     # Build min-heap of (area, index) for interior vertices.
+    # Each interior vertex is pushed once initially (n-2 entries), and at
+    # most 2 new entries are pushed per removal (one per affected neighbor),
+    # for a total of at most (n-2) + 2*(n-2) = 3n-6 entries.  n * 3
+    # provides a small safety margin.
     heap_areas = np.empty(n * 3, dtype=np.float64)
     heap_indices = np.empty(n * 3, dtype=np.int64)
     heap_size = 0
     for i in range(1, n - 1):
         heap_size = _heap_push(heap_areas, heap_indices, heap_size, areas[i], i)
 
-    while remaining > 2:
+    while True:
         # Pop minimum, skipping stale entries (lazy deletion).
         min_area = -1.0
         min_idx = -1
@@ -1897,7 +1901,6 @@ def _visvalingam_whyatt(coords, tolerance):
 
         # Remove vertex.
         removed[min_idx] = True
-        remaining -= 1
 
         # Update linked list.
         p = prev_idx[min_idx]
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -1756,12 +1756,84 @@ def _douglas_peucker(coords, tolerance):
 
 
 @ngjit
+def _heap_sift_down(heap_areas, heap_indices, heap_size, idx):
+    """Sift element at idx down to restore min-heap property.
+
+    Ties on area are broken by smallest index, matching the O(n^2)
+    linear scan which uses strict ``<`` and iterates in index order.
+    """
+    while True:
+        smallest = idx
+        left = 2 * idx + 1
+        right = 2 * idx + 2
+        if left < heap_size:
+            if (heap_areas[left] < heap_areas[smallest]
+                    or (heap_areas[left] == heap_areas[smallest]
+                        and heap_indices[left] < heap_indices[smallest])):
+                smallest = left
+        if right < heap_size:
+            if (heap_areas[right] < heap_areas[smallest]
+                    or (heap_areas[right] == heap_areas[smallest]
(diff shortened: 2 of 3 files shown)


brendancol:
## PR Review: polygonize: optimize visvalingam_whyatt with min-heap

Reviewed the heap-based rewrite of `_visvalingam_whyatt`. The algorithm is correct and the correctness test is solid. My only real concern is the two timing-based tests, which will likely flake on shared CI.

None.

- [ ] `xrspatial/tests/test_polygonize.py:1019` and `:1180`: both `test_visvalingam_whyatt_no_regression_small_inputs` and `test_visvalingam_whyatt_scales_subquadratic` assert wall-clock ratios (`ratio < 3.0`). Timing assertions are flaky on shared CI runners. Under load, a GC pause, numba recompilation, or a noisy neighbor can blow past 3x and fail a PR that has nothing wrong with it. The scaling test also runs n up to 32000 across 5 sizes x 10 iterations, which is slow for the unit suite. Consider moving the timing checks into the asv suite (`benchmarks/benchmarks/polygonize.py`) and keeping only `test_visvalingam_whyatt_heap_correctness` in the test file. The correctness test is what actually guards the behavior.

- [ ] The PR description says "Building the initial heap is O(n)", but the heap is built with n-2 sequential `_heap_push` calls at `polygonize.py:1879`, which is O(n log n), not a bottom-up O(n) heapify. The overall O(n log n) complexity still holds, so this is just a wording fix in the description, not the code.
- [ ] The O(n^2) baseline `_visvalingam_whyatt_o2` is duplicated verbatim in both test methods (`test_polygonize.py:962` and `:1037`). Pulling it into one module-level helper would keep the two copies from drifting apart.
- [ ] No asv benchmark covers the simplify path. `benchmarks/benchmarks/polygonize.py` exists but doesn't exercise simplification. Since issue #2539 is about simplify runtime, a benchmark there would guard against future regressions better than the wall-clock unit tests.

### What looks good
- Heap capacity `n*3` (`polygonize.py:1876`) is sized correctly. Total pushes over a run is at most 3(n-2): one initial push per interior vertex plus at most two re-pushes per removal. Worth noting because numba nopython mode does no bounds-checking, so an undersized buffer would be silent corruption rather than an IndexError.
- Lazy deletion is handled correctly. The `not removed[idx] and area == areas[idx]` check at `polygonize.py:1890` skips stale entries, and the float-equality test is reliable here because a fresh push stores exactly `areas[idx]`.
- Tie-break by smallest index (`polygonize.py:1770-1772`, `1797-1799`) matches the old linear scan's strict `<` plus index-order iteration, which is why the outputs stay identical.
- Endpoints never enter the heap: the `p > 0` and `nx_i < n - 1` guards preserve the sentinel handling from the original.
- The correctness test compares heap output against the O(n^2) baseline across 4 patterns, 5 sizes, and 3 tolerances. That is a real equivalence check, not a smoke test.

### Checklist
- [x] Algorithm matches reference (verified against the retained O(n^2) baseline)
- [x] Backends consistent (N/A: `_visvalingam_whyatt` is a CPU-only numba helper, not part of the DataArray backend dispatch)
- [x] NaN handling is correct (matches baseline: NaN areas are never selected)
- [x] Edge cases covered (n <= 2 early return; correctness test covers small n)
- [x] Dask chunk boundaries (N/A for this helper)
- [ ] Benchmark exists (no asv benchmark for simplify; see suggestion)
- [x] README feature matrix (N/A: private helper, no public API change)

brendancol:
@Melissari1997 [HUMAN] hey the review above is the `.claude/commands/review-pr.md` output.  What I've been doing is making sure to run this command, address any blockers/suggestions/nits and then do a follow up review as a final check.

brendancol (approved):
[HUMAN] there were a few small changes mentioned in the review-pr comment, besides that, LGTM

Melissari1997:
Thank you! @brendancol 
By the way, the commands are really well done. It would be useful in the future to set them up as pre-commit steps, to be run locally.

Melissari1997:
# PR Review: polygonize: optimize visvalingam_whyatt with min-heap (#2817)

## Blockers (must fix before merge)

None.

## Suggestions (should fix, not blocking)

None — all previous findings have been addressed.

## Nits (optional improvements)

None.

## What looks good

- **Algorithmic correctness verified against O(n²) baseline** across 4 patterns × 5 sizes × 3 tolerances = 60 combinations. The heap with lazy deletion produces bit-identical results.
- **Lazy deletion design is sound**: stale entries are filtered via both the `removed[idx]` flag and the `area == areas[idx]` value comparison. The monotonicity guarantee (`areas[p] = max(new_area, min_area)`) ensures an old stale entry's area never matches the updated value.
- **All edge cases handled**: NaN input returns all coords unchanged (no infinite loop), identical coords simplify to endpoints, two-point input returns early.
- **Performance verified**: manual benchmark shows O(n log n) scaling with doubling-time ratios of 1.57x–2.97x across 2000→32000 vertices. Timing tests use 200 iterations (small) / 50 iterations (large) for statistical stability.
- **Benchmark added**: `PolygonizeSimplify` class in `benchmarks/benchmarks/polygonize.py` covers both DP and VW methods.
- **Code is clean**: the vestigial `remaining` variable was removed, heap allocation rationale is documented (`3n - 6` max + safety margin), sentinel values are explained.

## Checklist

- [x] Algorithm matches reference/paper — heap-based Visvalingam–Whyatt is the standard fast implementation
- [x] All implemented backends produce consistent results — N/A (CPU-only internal helper, change is transparent to backends)
- [x] NaN handling is correct — tested, returns all coords unchanged
- [x] Edge cases are covered by tests — NaN, identical coords, straight-line, triangle, two-point, 60-way correctness comparison
- [x] Dask chunk boundaries handled correctly — N/A (operates on post-polygonization CPU arrays)
- [x] No premature materialization or unnecessary copies — `np.empty` pre-allocation, single pass
- [x] Benchmark exists — `PolygonizeSimplify` class added
- [x] README feature matrix updated — N/A (no new public function)
- [x] Docstrings present and accurate — all three heap helpers and `_visvalingam_whyatt` have docstrings

**Summary**: The PR is clean, correct, and complete. No blockers.

brendancol:
@Melissari1997 hey do you have a preference of WhatsApp vs. Slack vs. something else for chat / direct messaging?

Melissari1997:
I think slack is ok!

PR #2824, opened 2026-06-02
Synthesize x/y coords for no-georef VRT reads

### LINKED ISSUE #2818: VRT no-georef reads drop x/y coordinates entirely
**Describe the bug**

Reading a VRT that has no `<GeoTransform>` element drops the x and y coordinates. Both the eager path and the chunked/dask path return a DataArray with no x/y coords, while still setting `georef_status='none'`.

This diverges from the non-VRT read path. For a plain GeoTIFF with no georef tags, `_coords.py` synthesizes integer pixel coords `0..N-1` (via `coords_from_pixel_geometry(..., has_georef=False)`) and stamps `georef_status='none'`. The VRT paths start from `coords = {}` and only fill it in when a GeoTransform is present, so a no-GeoTransform VRT comes back with no coords at all.

Downstream code that assumes x/y always exist on a read result works on the plain GeoTIFF but breaks on the equivalent VRT.

**To Reproduce**

Read a VRT with no `<GeoTransform>` element, both eager and chunked. The result has `georef_status='none'` but no `x`/`y` coords.

**Expected behavior**

When a VRT has no GeoTransform, both VRT paths should synthesize integer x/y coords the same way the non-VRT path does, reusing the helper in `_coords.py`. The result should carry integer x/y coords with `georef_status='none'`, matching the non-VRT behavior.

**Affected code**

- `xrspatial/geotiff/_backends/vrt.py`, eager path (~line 561): `coords = {}` when no GeoTransform
- `xrspatial/geotiff/_backends/vrt.py`, chunked path (~line 1098): `coords = {}` when no GeoTransform
- Reference: `xrspatial/geotiff/_coords.py` non-VRT synthesis

**Severity:** High

brendancol:
Closes #2818

A VRT with no `<GeoTransform>` element came back with no x/y coords on both the eager and chunked read paths, while a plain GeoTIFF with no georef tags synthesizes integer pixel coords `0..N-1` (via `coords_from_pixel_geometry(..., has_georef=False)` in `_coords.py`). Downstream code that assumes x/y always exist broke on no-georef VRTs but worked on the equivalent plain GeoTIFF.

## What changed

- Both VRT paths now reuse the shared `coords_from_pixel_geometry(has_georef=False)` helper in the no-`<GeoTransform>` branch (eager: `_backends/vrt.py` ~line 561; chunked: ~line 1098) instead of leaving `coords = {}`.
- The result carries integer x/y coords with `georef_status='none'`, matching the non-VRT path. Windowed reads shift the coords to the window offset, same as the non-VRT windowed read.

## Backend coverage

The change touches the eager (numpy) and chunked (dask) VRT read paths. GPU reads go through the same chunked code, so the synthesized coords apply there too. No backend support changed.

## Test plan

- [x] New parity tests assert x/y coords exist and match the non-VRT synthesized coords for eager, chunked, and windowed no-georef VRT reads
- [x] Confirmed the new tests fail without the fix (empty coords) and pass with it
- [x] Full VRT suite: 494 passed
- [x] Full geotiff suite: 6052 passed, 81 skipped, 1 xfailed


brendancol (commented):
## PR Review: Synthesize x/y coords for no-georef VRT reads

### Blockers
None.

### Suggestions
None blocking.

### Nits
- `_backends/vrt.py`, eager else-branch (~610): the window/height/width setup repeats the georef branch right above it. The chunked path hoists `coord_window` out of the if/else; the eager path could do the same and drop the second `arr.shape[:2]` read. Cosmetic. The two branches are mutually exclusive and the duplication is tiny.

### What looks good
- Both fixed branches call the shared `coords_from_pixel_geometry(has_georef=False)` helper instead of re-deriving the arange logic, so the no-georef coord convention stays in one place.
- Windowed reads work: the eager branch reuses the georef branch's `max(0, window[0])` offset, and the chunked branch reuses `(win_r0, win_c0, win_r0+full_h, win_c0+full_w)`, which matches the non-VRT dask path's `(win_r0, win_c0, win_r1, win_c1)`.
- Tests check coord values and dtype against the non-VRT reference for eager, chunked, and windowed reads, and they fail without the fix.

### Checklist
- [x] Coords match non-VRT reference (values + dtype)
- [x] georef_status stays 'none'
- [x] Dask (chunked) path covered
- [x] Windowed reads covered
- [x] No premature materialization (coords built from extents, not data)
- [x] No new public API; README/notebook not applicable

brendancol (commented):
## PR Review (follow-up)

The nit from the first pass is resolved: the eager path now hoists `height/width` and `coord_window` above the `if gt is not None` block, so the georef and no-georef branches share the same setup. This matches the chunked path's structure.

### Blockers
None.

### Suggestions
None.

### Nits
None.

Both VRT read paths synthesize integer x/y pixel coords for no-GeoTransform VRTs via the shared `coords_from_pixel_geometry(has_georef=False)` helper, matching the non-VRT path. Tests cover eager, chunked, and windowed reads and fail without the fix. 494 VRT tests pass, flake8 clean.

PR #2826, opened 2026-06-02
Enforce stable_only=True for HTTP/fsspec sources on non-VRT read paths

### LINKED ISSUE #2821: stable_only=True does not reject HTTP/fsspec sources on non-VRT read paths
**Describe the bug**

`open_geotiff(..., stable_only=True)` is supposed to restrict reads to stable-tier sources, but it doesn't enforce that for HTTP/fsspec sources on the non-VRT paths.

`reader.fsspec` and `reader.http` are marked `advanced` in `SUPPORTED_FEATURES` (`xrspatial/geotiff/_attrs.py:292-293`). The `stable_only` check only runs on the VRT path, through `_validate_stable_only_vrt`. The eager non-VRT path calls `_read_to_array` directly (`xrspatial/geotiff/__init__.py:959`) and never re-checks the tier, and the dask path only forwards `stable_only` to `read_vrt` when the source ends in `.vrt` (`xrspatial/geotiff/_backends/dask.py:208`). For a plain HTTP or fsspec source, neither path rejects the advanced-tier reader.

**Reproduction**

Write a valid GeoTIFF into a `memory://` fsspec store and call `open_geotiff(path, stable_only=True)`. It succeeds on both the eager and dask paths, even though `reader.fsspec` is advanced. A local file under `stable_only=True` succeeds too, which is correct.

**Expected behavior**

With `stable_only=True`, the eager and dask non-VRT paths should reject advanced-tier HTTP/fsspec sources and raise the same error the VRT path raises (`VRTStableSourcesOnlyError`, a `GeoTIFFAmbiguousMetadataError` subclass), unless the caller passes `allow_experimental_codecs=True`. Local-file sources should keep working under `stable_only=True`.

**Additional context**

This makes the stability contract leaky: someone who sets `stable_only=True` to avoid advanced-tier readers can still get an HTTP/fsspec read.

brendancol:
Closes #2821

`reader.http` and `reader.fsspec` are advanced-tier in `SUPPORTED_FEATURES`, but `stable_only=True` was only gated on the VRT dispatch. The eager non-VRT path called `read_to_array` directly, and the dask path only forwarded `stable_only` to `read_vrt` for `.vrt` sources, so an HTTP/fsspec source read straight through `stable_only=True` unchecked.

## Changes

- Add `RemoteStableSourcesOnlyError` (a `GeoTIFFAmbiguousMetadataError` subclass), the non-VRT sibling of `VRTStableSourcesOnlyError`, and re-export it from `xrspatial.geotiff`.
- Add `_validate_stable_only_remote` and call it from `open_geotiff` (covers the eager/dask/GPU dispatch) and from `read_geotiff_dask` (direct callers). It rejects `http(s)://` and fsspec URIs under `stable_only=True` before any fetch or decode.
- `allow_experimental_codecs=True` stays the documented unlock; local-file reads still pass.
- Document the gate in `docs/source/reference/geotiff.rst`.

## Backends

The gate lives in `open_geotiff` ahead of the eager / dask / GPU dispatch, plus a direct check in `read_geotiff_dask`. CPU eager and CPU dask are covered by the new tests; the GPU dispatch path shares the same `open_geotiff` gate (no GPU hardware in CI to assert directly).

## Test plan

- [x] Eager fsspec source under `stable_only=True` raises `RemoteStableSourcesOnlyError`
- [x] Dask fsspec source (via `open_geotiff` and direct `read_geotiff_dask`) raises
- [x] `http://` source raises before any network fetch
- [x] `allow_experimental_codecs=True` unlocks the read
- [x] Local-file eager and dask reads still succeed under `stable_only=True`
- [x] Existing VRT stable_only and release-gate suites still green


brendancol (commented):
## PR Review: Enforce stable_only=True for HTTP/fsspec sources on non-VRT read paths

None.

- [ ] `read_geotiff_gpu` (`xrspatial/geotiff/_backends/gpu.py:71`) takes `stable_only` but does not gate remote HTTP/fsspec sources on its own. Through `open_geotiff` the new gate runs ahead of the GPU dispatch, so the public entry point is fine. A direct `read_geotiff_gpu(http_url, stable_only=True)` call still gets through, the same gap this PR closes for eager/dask. Out of scope for #2821 (eager + dask), but a follow-up would make all three direct entry points match. No GPU in CI to test it here anyway.

- [ ] The rejection message in `_validate_stable_only_remote` (`xrspatial/geotiff/_validation.py:798`) cites `release_gate_geotiff.rst` and epic #2342, matching `_validate_stable_only_vrt`. The doc that actually describes this gate is `geotiff.rst`, not the release-gate checklist. The sibling validator does the same, so leaving it is fine.

### What looks good
- The gate sits in `open_geotiff` ahead of the eager/dask/GPU dispatch (`__init__.py:917`), with a direct check in `read_geotiff_dask` (`_backends/dask.py`), so both the routed and direct dask paths are covered.
- `_validate_stable_only_remote` mirrors `_validate_stable_only_vrt`: same early returns, same `allow_experimental_codecs` unlock, same error family (`GeoTIFFAmbiguousMetadataError`).
- `source` is already normalized by `_coerce_path` before the gate, so path-like inputs are strings by the time `_is_fsspec_uri` sees them.
- Tests cover eager, dask via `open_geotiff`, direct `read_geotiff_dask`, the `http://` case (gated before any fetch), the experimental-codecs unlock, and both local-file backends still passing.

### Checklist
- [x] Behavior matches the VRT gate it mirrors
- [x] Targeted backends (eager, dask) gated and tested
- [x] Error type/message consistent with the existing VRT path
- [x] Edge cases covered (local file passes, http gated pre-fetch, unlock works)
- [x] No premature materialization (gate runs before any read)
- [ ] Benchmark not needed (validation-only change)
- [x] Doc updated (geotiff.rst)

brendancol (commented):
## Review follow-up

No new commits: both findings were resolved without code changes in this PR.

- Suggestion (direct `read_geotiff_gpu` does not gate remote sources): deferred to #2830. It is GPU-only and was out of scope for the eager + dask finding in #2821. The public `open_geotiff(..., gpu=True)` path is already covered by the shared gate in this PR; only the direct GPU entry point remains, which #2830 tracks.
- Nit (error message cites `release_gate_geotiff.rst` rather than `geotiff.rst`): dismissed. The citation matches the sibling `_validate_stable_only_vrt` validator on purpose. Changing one and not the other would split the convention, and the review itself flagged leaving it as fine.

PR #2832, opened 2026-06-02
test(slope): cover planar boundary modes on cupy and dask+cupy backends

brendancol:
Each planar slope backend handles the `boundary` parameter on its own. `test_boundary_modes` and `test_boundary_numpy_equals_dask` cover the numpy and dask+numpy paths across every mode, but neither GPU backend was ever tested with a non-default boundary.

That leaves two code paths unexercised. `_run_cupy` has its own pad-recurse-crop branch for `boundary != 'nan'`, and `_run_dask_cupy` passes the mode through `_boundary_to_dask(..., is_cupy=True)` and `map_overlap`. A regression in either would have shipped without a test catching it.

This adds two parametrized tests that compare the cupy and dask+cupy planar output against the numpy reference for `nan`, `nearest`, `reflect`, and `wrap`. The dask+cupy test runs both a multi-chunk and a single-chunk layout so the overlap path is covered.

Test-only, no source changes. Ran on a CUDA host: 12 new tests pass, full slope suite 103 passed.

---
Found by the test-coverage sweep against `slope`.

### DIFF
--- a/xrspatial/tests/test_slope.py
+++ b/xrspatial/tests/test_slope.py
@@ -111,6 +111,46 @@ def test_boundary_numpy_equals_dask(boundary, size, chunks):
         np_result.data, da_result.data.compute(), equal_nan=True, rtol=1e-5)
 
 
+@cuda_and_cupy_available
+@pytest.mark.parametrize("boundary", ['nan', 'nearest', 'reflect', 'wrap'])
+def test_boundary_numpy_equals_cupy(boundary):
+    # The planar cupy backend has its own non-'nan' boundary branch
+    # (_run_cupy pads, recurses, then crops). Pin it against the numpy
+    # reference for every boundary mode so a GPU-only boundary regression
+    # can't ship undetected.
+    rng = np.random.default_rng(42)
+    data = (rng.random((10, 15)).astype(np.float64) * 500)
+    numpy_agg = create_test_raster(data, backend='numpy', attrs={'res': (1, 1)})
+    cupy_agg = create_test_raster(data, backend='cupy', attrs={'res': (1, 1)})
+    np_result = slope(numpy_agg, boundary=boundary)
+    cupy_result = slope(cupy_agg, boundary=boundary)
+    np.testing.assert_allclose(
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## Review: test(slope) GPU boundary coverage

Read the full file from a worktree on the PR head and ran the 12 new tests on a CUDA host — all pass. No blockers.

### What this closes
The planar cupy and dask+cupy backends each have their own boundary handling — `_run_cupy` pads/recurses/crops for non-'nan' modes, and the dask+cupy path threads `boundary` through `_boundary_to_dask(is_cupy=True)` into `map_overlap`. Before this PR both were unpinned, so a GPU-only boundary regression could ship while the numpy/dask+numpy tests stayed green. The new tests pin both against the numpy reference across all four modes, which is the right reference and matches the existing `test_boundary_numpy_equals_dask` style.

### Suggestion (I'll push this myself)
`test_boundary_numpy_equals_dask_cupy` only parametrizes even chunks `(10,15)/(5,5)` and a single chunk `(10,15)/(10,15)`. Its numpy-vs-dask sibling deliberately also covers a ragged last chunk `(7,9)/(3,3)` and `(5,5)/(2,2)`, and a ragged trailing chunk under a depth-1 overlap is exactly where a dask+cupy boundary bug would hide. Adding a ragged case brings GPU dask coverage to parity with the CPU dask test. Pushing that to this branch now.

### Note on CI
The new tests are gated on `@cuda_and_cupy_available`, so the green CI runners (no CUDA) skip them — they were validated on a CUDA host, not by CI. That's the existing convention for this file, just calling it out so the green checks aren't read as having exercised the GPU paths.

### Good
- One file, test-only, scoped to the gap it names.
- `.get()` / `.compute().get()` used correctly; `equal_nan=True` is right for the 'nan' edges; seeded RNG.

Looks ready once the ragged case lands.

brendancol:
Pushed the ragged-chunk case (98a0537): `test_boundary_numpy_equals_dask_cupy` now parametrizes `(7,9)/(3,3)` alongside the even and single-chunk cases, matching the numpy-vs-dask sibling. All 12 dask+cupy variants pass on a CUDA host; flake8 clean on the changed lines.

PR #2834, opened 2026-06-02
Fix GPU focal std/var precision on large-offset rasters

### LINKED ISSUE #2831: GPU focal_stats std/var lose precision on large-offset rasters (one-pass variance)
## Description

The GPU focal statistics kernels `_focal_std_cuda` and `_focal_var_cuda` in `xrspatial/focal.py` compute variance with a one-pass formula:

    var = (sum_wx2 / w_sum) - mean * mean

in float32. This subtracts two nearly-equal large quantities (`E[x^2]` and `E[x]^2`), which is catastrophic cancellation. On rasters whose values carry a large offset (roughly 1e6 to 1e7), the float32 result loses its significant digits and collapses toward zero.

The numpy and dask+numpy backends route `std`/`var` through `np.nanstd`/`np.nanvar`, which are two-pass and stay in the input float dtype (float64 for a float64 raster). So the cupy and dask+cupy backends diverge from the CPU result on the same input, with no warning.

## Reproduce

```python
import numpy as np, cupy, xarray as xr
from xrspatial.focal import focal_stats
from xrspatial.convolution import custom_kernel

rng = np.random.default_rng(0)
data = (1e7 + rng.random((8, 8))).astype(np.float64)
kernel = custom_kernel(np.array([[0, 1, 0], [1, 1, 1], [0, 1, 0]]))

np_var = focal_stats(xr.DataArray(data), kernel, stats_funcs=['var']).sel(stats='var').data
cp_var = focal_stats(xr.DataArray(cupy.asarray(data)), kernel, stats_funcs=['var']).sel(stats='var').data.get()
# interior cell: numpy ~0.079, cupy ~0.109; some interior cells collapse to exactly 0
```

## Fix

Switch the two GPU kernels to a two-pass computation over the same window: accumulate the weighted mean first, then in a second sweep accumulate the weighted sum of squared deviations from that mean. The window is already re-readable inside the kernel, so this needs no extra memory. A two-pass float32 computation gives the right variance even at the 1e7 offset.

## Coverage gap

The existing cupy-vs-numpy parity tests for std/var use small values and pass within rtol=1e-4. No test exercises a large-offset raster, which is why this slipped through.

Found by the accuracy sweep against the focal module.

brendancol:
Closes #2831

The GPU `_focal_std_cuda` and `_focal_var_cuda` kernels computed variance with a one-pass `E[x^2] - E[x]^2` form in float32. On rasters with a large offset (~1e6-1e7) the two terms are nearly equal, so the subtraction loses all precision and std/var collapse toward zero. The numpy/dask paths use two-pass `np.nanstd`/`np.nanvar` in float64, so the GPU backends silently diverged.

- Rewrote both kernels to two passes over the same window: weighted mean first, then weighted sum of squared deviations from that mean. No extra memory, same NaN-skipping and `w_sum > 0` guard as before.
- Added a regression test comparing cupy vs numpy std/var on 1e6 and 1e7 offset rasters.

Backend coverage: the kernels are shared by the cupy and dask+cupy `focal_stats` paths; numpy and dask+numpy were already correct and unchanged.

Test plan:
- [x] `pytest xrspatial/tests/test_focal.py` (185 passed, includes cupy + dask+cupy)
- [x] New `test_focal_stats_std_var_large_offset_gpu_matches_numpy_2831` passes (cupy)
- [x] Verified pre-fix cupy var collapsed to ~0 at 1e7 offset; post-fix matches numpy to float64 epsilon


brendancol (commented):
## PR Review: Fix GPU focal std/var precision on large-offset rasters

None.

None.

- The new test exercises the cupy backend directly. The dask+cupy path shares the exact same `_focal_std_cuda`/`_focal_var_cuda` kernels (focal.py line 1145), so it is covered transitively. A one-line dask+cupy assertion would make that explicit, but it is not required.
- Each kernel now reads its window twice (one pass for the mean, one for the deviations), so std/var do roughly double the global-memory reads per pixel. That is the expected cost of a stable two-pass variance and only touches the std/var kernels. Worth knowing if someone benchmarks it later.

### What looks good
- The two-pass form is the standard fix for the cancellation. Subtracting the window mean before squaring holds precision at any offset; verified to float64 epsilon at 1e6 and 1e7.
- Both passes use identical iteration bounds, NaN-skip (`x != x`), kernel-zero skip, and the `w_sum <= 0` to NaN early return, so all-NaN windows and weighted kernels behave exactly as before apart from the precision gain.
- The numpy/dask paths route through np.nanstd/np.nanvar (population, ddof=0) and were already correct; the GPU kernels now match that definition.
- The regression test asserts the numpy reference is non-flat before comparing, so it cannot pass vacuously.

### Checklist
- Algorithm matches reference: yes, weighted population variance, ddof=0
- All implemented backends consistent: yes, cupy and dask+cupy share the fixed kernels; numpy/dask unchanged
- NaN handling correct: yes, x != x skip in both passes
- Edge cases tested: large-offset covered; all-NaN window preserved by the w_sum guard
- Dask chunk boundaries: unchanged, shared kernel
- Benchmark: not needed for a kernel-internal fix
- Docstrings present: comments explain the two-pass rationale

PR #2835, opened 2026-06-02
aspect: cover northness/eastness method='geodesic' path (#2829)

### LINKED ISSUE #2829: northness/eastness geodesic path is untested
## What's missing

`northness()` and `eastness()` in `xrspatial/aspect.py` take a `method` argument and pass it to `aspect()`. Every test in `xrspatial/tests/test_northness_eastness.py` uses the default `method='planar'`, so the `method='geodesic'` branch through these two wrappers has never run in CI on any backend.

The planar side is well covered: correctness, output range, NaN handling, custom names, and all four backends. The gap is only the geodesic branch of the two wrappers.

## Why it matters

`northness`/`eastness` forward `method`, `z_unit`, and `boundary` to `aspect()`. If that wiring broke for `method='geodesic'`, nothing would catch it. Geodesic aspect runs a different code path (ECEF projection, lat/lon coordinate grids) than planar, so the wrapper behavior should be pinned on its own rather than assumed from the planar tests.

## Proposed fix

Add tests that assert:
- `northness(agg, method='geodesic')` equals `cos(deg2rad(...))` of geodesic aspect, with flat cells (aspect -1) mapped to NaN
- `eastness(agg, method='geodesic')` equals `sin(deg2rad(...))` likewise
- numpy, dask+numpy, cupy, and dask+cupy agree, reusing the geodesic raster helpers from test_geodesic_aspect.py

Test-only. No source change expected.

brendancol:
Closes #2829

`northness()` and `eastness()` forward `method` to `aspect()`, but every existing test ran only the default `method='planar'`. The geodesic branch through these two wrappers had no coverage on any backend.

This adds:
- Correctness: `northness == cos(deg2rad(geodesic aspect))` and `eastness == sin(...)`, with flat cells (-1) mapped to NaN. The test surface is diagonal, so planar and geodesic aspect differ by about 80 degrees here; a wrapper that silently dropped `method='geodesic'` and fell back to planar would fail these assertions.
- Flat surface: geodesic aspect -1 maps to NaN for both functions.
- Backend parity across numpy, dask+numpy, cupy, and dask+cupy.

Backend coverage: all four. GPU-validated locally (the cupy and dask+cupy tests ran and passed on a CUDA host).

Test-only change. No source modification.

Test plan:
- [x] `pytest xrspatial/tests/test_northness_eastness.py` — 37 passed, including 9 new geodesic tests with GPU backends executed
- [x] flake8 clean on the edited file


brendancol (commented):
## PR Review: cover northness/eastness method='geodesic' path (#2829)

### Blockers
None.

### Suggestions
None.

### Nits
- The two value-checking tests (`test_northness_geodesic_correctness`, `test_eastness_geodesic_correctness`) only assert on numpy. cupy and dask correctness comes in transitively: the parity tests assert each backend equals numpy, and numpy is value-checked. Fine as written. Flagging it so it reads as a deliberate choice, not a gap.
- `_to_numpy` and `_make_geo_raster` repeat helpers that already live in `test_geodesic_aspect.py` and `test_aspect.py`. Three copies isn't worth a refactor, but if anyone touches `general_checks.py` later, that's the place to consolidate.

### What looks good
- The correctness tests use a diagonal surface where planar and geodesic aspect land about 80 degrees apart, and the expected value comes straight from `aspect(method='geodesic')`. If the wrapper dropped `method` and fell back to planar, these would fail. I ran a negative control to confirm that, so they're real guards rather than smoke tests.
- All four backends are exercised: numpy gets the value check, dask+numpy/cupy/dask+cupy get parity against numpy. The GPU backends ran and passed on a CUDA host.
- Flat surface asserts geodesic aspect -1 becomes NaN for both functions.
- The parity tests call `general_output_checks`, so attrs/coords/dims preservation is checked on the dask and cupy paths too.
- Test-only, matching the issue scope.

### Checklist
- Backends consistent: yes, parity covers all four
- NaN handling: yes, flat to NaN tested
- Edge cases: flat surface here; degenerate shapes already covered elsewhere
- Dask chunks: ragged (4,5) on 8x10
- No premature materialization: compute/get only in assertions
- Benchmark / README / docstrings: n/a or present

Issue #2733, opened 2026-05-30
focal_stats and hotspots: result .name differs across backends


brendancol:
## Summary

The `.name` of the DataArray returned by `focal_stats` and `hotspots` is not the same across the four backends. On the dask paths it leaks an internal dask graph token.

## Details

`focal_stats`:
- numpy / dask+numpy: `'focal_apply'` (inherited from the internal `apply` call)
- cupy: `None`
- dask+cupy: a dask graph token, e.g. `'_trim-d142fa3d...'`

`hotspots`:
- numpy / cupy: `None`
- dask+numpy / dask+cupy: a dask graph token, e.g. `'_trim-b75965c2...'`

Root cause: the dask paths build the output `xr.DataArray` without an explicit `name=`. When the backing array is a dask array, xarray adopts the dask array's internal graph-layer name as the public `.name`. That token changes from call to call and exposes an implementation detail. The numpy/cupy paths leave the name as `None` (or inherit a different value).

`.name` is user-visible: it shows in the repr, becomes the variable key in `to_dataset()`, and sets default plot labels. So a chained pipeline ends up with a different name depending on which backend ran.

## Fix

Set a stable `.name` on the returned DataArray in both functions so all four backends agree. `focal_stats` keeps the `'focal_apply'` name the numpy path already emits; `hotspots` uses `'hotspots'`.

The rest of the metadata (attrs `res`/`crs`, coords, dims) is already preserved consistently across all four backends. Only `.name` was off.

PR #2838, opened 2026-06-02
fix(slope): reset .name on dask backends so all four backends agree

### LINKED ISSUE #2837: slope() leaks the dask graph token as .name on dask backends
On the dask backends, `slope(agg, name=None)` returns a DataArray whose `.name` is the dask array's internal graph token (e.g. `_trim-<hash>` for planar, `getitem-<hash>` for geodesic) instead of `None`. The numpy and cupy backends correctly return `None`.

The cause: when `name=None` is passed to `xr.DataArray(dask_array, name=None, ...)`, xarray adopts the dask array's name rather than leaving it unset. So the public `.name` ends up backend-dependent and non-deterministic per call.

This is the same class of bug already fixed for zonal (#2611), focal (#2733), and proximity (#2723).

### Reproduce
```python
from xrspatial import slope

arr = xr.DataArray(da.from_array(np.random.rand(8, 8), chunks=4),
                   dims=['y', 'x'],
                   coords={'y': np.arange(8), 'x': np.arange(8)})
print(slope(arr).name)   # -> '_trim-...' ; expected: None
```

### Fix
Reset `result.name` after construction so all four backends agree (`None` when `name=None`, the given name otherwise). Regression tests assert name consistency across numpy/cupy/dask+numpy/dask+cupy for both planar and geodesic methods.

Found by a metadata sweep of the slope module.

brendancol:
Closes #2837.

`slope(agg, name=None)` leaked the dask array's internal graph token as `.name` on the dask backends (`_trim-<hash>` planar, `getitem-<hash>` geodesic), while numpy and cupy returned `None`. Passing `name=None` to `xr.DataArray(dask_array, ...)` lets xarray adopt the dask name, so the public `.name` came out backend-dependent and non-deterministic.

The fix resets `result.name = name` after construction, so every backend returns `None` when `name=None` and the given name otherwise. Same approach already used for zonal (#2611), focal (#2733), and proximity (#2723).

### Tests
Four parametrized regression tests (one per backend) assert name consistency for both planar and geodesic methods. Full slope + geodesic-slope suite: 99 passed on a CUDA host.

Found by a metadata sweep of the slope module.

### DIFF
--- a/xrspatial/slope.py
+++ b/xrspatial/slope.py
@@ -429,8 +429,13 @@ def slope(agg: xr.DataArray,
         )
         out = mapper(agg)(agg.data, lat_2d, lon_2d, WGS84_A2, WGS84_B2, z_factor, boundary)
 
-    return xr.DataArray(out,
-                        name=name,
-                        coords=agg.coords,
-                        dims=agg.dims,
-                        attrs=agg.attrs)
+    result = xr.DataArray(out,
+                          name=name,
+                          coords=agg.coords,
+                          dims=agg.dims,
+                          attrs=agg.attrs)
+    # On dask backends, xr.DataArray keeps the dask array's internal graph
+    # token as .name when name=None, so reset it post-construction to match
+    # the numpy/cupy backends. (Same fix as zonal #2611, focal #2733.)
+    result.name = name
+    return result
--- a/xrspatial/tests/test_slope.py
+++ b/xrspatial/tests/test_slope.py
@@ -210,6 +210,68 @@ def test_name_none_accepted():
     assert result.name is None
 
 
+# The output .name must agree across backends. xr.DataArray keeps a dask
+# array's internal graph token (e.g. '_trim-<hash>' planar, 'getitem-<hash>'
+# geodesic) as .name when name=None, so the dask backends used to disagree
+# with numpy/cupy. Regression for the slope .name leak (same class as zonal
+# #2611, focal #2733, proximity #2723, viewshed #2743).
+def _assert_name_planar(backend, name):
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## Review: fix(slope) reset .name on dask backends

Read both changed files in full from a worktree on the PR head. No blockers.

### Correctness
The fix is right. `xr.DataArray(dask_array, name=None)` lets xarray pick up the dask array's name, which is why the dask backends leaked `_trim-<hash>`/`getitem-<hash>` while numpy and cupy returned `None`. Assigning `result.name = name` after construction (slope.py:440) forces agreement: `None` stays `None`, and an explicit name is preserved (the assignment is a harmless no-op in that case). It sits at the single shared return, so all four backends get it without per-backend branching. No algorithm or NaN-handling change.

### Tests
`test_name_consistent_*` covers all four backends, parametrized over `name in {None, 'slope'}`, for both planar and geodesic. That's the right matrix for a name-leak regression — `None` is the case that actually used to break, and an explicit name guards against an over-eager reset. Helpers and decorators all resolve against general_checks. Reported 99 passed on a CUDA host.

### Nits (optional)
- `_assert_name_planar` passes `chunks=(3, 4)` to `create_test_raster` for every backend including numpy, where it's ignored. Harmless, just slightly misleading.
- Tests assert `.name` only, not values — correct here, since value correctness for both methods is already covered elsewhere in the file.

### Good
- Scoped to one finding: the 4-line return change plus its tests, nothing adjacent touched.
- Comment cites the sibling fixes (#2611, #2733), so the next person sees the pattern.

Backends consistent, NaN handling untouched, no materialization (`result.name = name` is O(1) metadata). Looks ready once CI is green.

PR #2844, opened 2026-06-02
Warn on degree/meter unit mismatch in planar aspect()

### LINKED ISSUE #2839: planar aspect() does not warn on degree/meter unit mismatch like slope() does
## Description

`aspect()` with the default `method='planar'` silently produces wrong results when given a lat/lon raster (coordinates in degrees) with elevation values in meters. The planar path in `xrspatial/aspect.py` calls `get_dataarray_resolution(agg)` directly, without first checking whether the horizontal units (degrees) match the vertical units (meters).

`slope()` already guards against this. Its planar path calls `warn_if_unit_mismatch(agg)` before computing resolution, so the user gets a `UserWarning` when degrees and meters are mixed. Aspect divides elevation differences by cell size the same way slope does, so it has the same problem, just without the warning.

## Expected behavior

Aspect's planar path should call the same `warn_if_unit_mismatch()` helper that slope uses, before computing resolution, so users get warned when horizontal units don't match vertical units.

## Steps to reproduce

Call `aspect()` on a DataArray with degree-based lat/lon coordinates and meter elevation values. `slope()` warns; `aspect()` does not.

## Scope

One-line behavior fix plus a test. It reuses slope's existing helper and import. The other aspect findings (dask memory guard, name leak, z_unit error message) are tracked separately.

brendancol:
Closes #2839

## What this does

- `aspect()`'s planar path now calls `warn_if_unit_mismatch(agg)` before computing resolution, the same guard `slope()` already has. Degree-based lat/lon rasters with meter elevations get a `UserWarning` instead of silently producing wrong values.
- Adds a "Notes" paragraph to the `aspect()` docstring mirroring slope's, explaining the warning and pointing to `method='geodesic'` or reprojection.

## Backend coverage

The warning lives in the shared `aspect()` entry point ahead of backend dispatch, so it fires for all four backends (numpy / cupy / dask+numpy / dask+cupy). No backend kernels changed.

## Test plan

- [x] `test_planar_warns_on_degree_coords_meter_elevation` asserts the warning fires on a degree-coord meter-elevation raster
- [x] `test_planar_no_warn_on_projected_meter_coords` asserts no warning on projected meter coords
- [x] Full `test_aspect.py` passes (200 tests)

## Scope

Just the warning. The other aspect findings (dask memory guard, name leak, z_unit error message) are separate PRs.

### DIFF
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -20,7 +20,7 @@ from xrspatial.geodesic import (INV_2R, WGS84_A2, WGS84_B2, _check_geodesic_memo
 from xrspatial.utils import (Z_UNITS, ArrayTypeFunctionMapping, _boundary_to_dask,
                              _extract_latlon_coords, _pad_array, _validate_boundary,
                              _validate_raster, cuda_args, get_dataarray_resolution, has_dask_array,
-                             ngjit)
+                             ngjit, warn_if_unit_mismatch)
 
 
 def _geodesic_cuda_dims(shape):
@@ -399,6 +399,14 @@ def aspect(agg: xr.DataArray,
         2D aggregate array of calculated aspect values.
         All other input attributes are preserved.
 
+    Notes
+    -----
+    The ``'planar'`` method uses the coordinate spacing directly as the cell
+    size. If the coordinates are in degrees (lat/lon) but the elevation values
+    are in meters, the result is wrong by orders of magnitude. When this
+    mismatch is detected, a ``UserWarning`` is emitted suggesting you reproject
+    to a projected CRS or use ``method='geodesic'``.
+
     References
     ----------
 
@@ -433,6 +441,7 @@ def aspect(agg: xr.DataArray,
     _validate_boundary(boundary)
 
     if method == 'planar':
+        warn_if_unit_mismatch(agg)
         cellsize_x, cellsize_y = get_dataarray_resolution(agg)
         mapper = ArrayTypeFunctionMapping(
             numpy_func=_run_numpy,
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Warn on degree/meter unit mismatch in planar aspect()

None.

One observation, not blocking: the new test helpers `_degree_coord_meter_elevation_raster` and `_projected_meter_raster` in test_aspect.py are copies of the same helpers in test_slope.py. The duplication keeps the aspect test file self-contained, which is how the suite is already organized, so it's fine as-is.

- The shared warning message in utils.py ends with "before calling `slope`" even when the call comes from aspect. That wording predates this PR and lives in the shared helper, so leaving it is consistent. A future cleanup could make it backend-agnostic.
- A few existing oracle tests (the cellsize=0.5 variants) now surface this UserWarning because their coordinate spacing looks degree-like to the heuristic. The tests still pass; that's the heuristic doing its job, not a regression.

### What looks good
- The fix mirrors slope.py: same helper, same import source, same placement just before `get_dataarray_resolution` in the planar branch.
- The warning sits in the shared `aspect()` entry point ahead of backend dispatch, so all four backends (numpy, cupy, dask+numpy, dask+cupy) get it with no kernel changes.
- Tests cover both directions: the warning fires on degree/meter coords and stays silent on projected meter coords. Both pass, and the full aspect suite (200 tests) is green.
- Added a Notes paragraph to the aspect docstring matching slope's.

### Checklist
- [x] No algorithm change; warning only
- [x] All backends consistent (warning is pre-dispatch)
- [x] NaN handling unchanged
- [x] Edge cases covered (warn / no-warn both tested)
- [x] Dask chunk boundaries unchanged
- [x] Benchmark not needed

PR #2845, opened 2026-06-02
Fix aspect() .name leak on dask backends when name=None

### LINKED ISSUE #2841: aspect(name=None) leaks Dask graph names on dask backends
**Describe the bug**

`aspect(..., name=None)` leaks Dask graph names. In `xrspatial/aspect.py` the output `xr.DataArray` is built but its `.name` is never reset. On Dask backends, `xr.DataArray` keeps the dask array's internal graph token as `.name` when `name=None`, so the result comes back named like `_trim-<hash>` (planar) or `getitem-<hash>` (geodesic) instead of `None`. The numpy and cupy backends return `.name is None`.

`slope()` already handles this by resetting `result.name` after construction (`xrspatial/slope.py`, fixed in #2838). `aspect()` should do the same so `name=None` gives `.name is None` on every backend.

**Expected behavior**

`aspect(raster, name=None).name is None` on all four backends (numpy, cupy, dask+numpy, dask+cupy), for both `method='planar'` and `method='geodesic'`.

**Additional context**

Same bug class as zonal #2611, focal #2733, and slope #2838. The fix resets `result.name` after constructing the output DataArray, mirroring slope.py.

brendancol:
Closes #2841

`xr.DataArray` keeps a dask array's internal graph token as `.name` when `name=None`, so `aspect()` returned something like `_trim-<hash>` (planar) or `getitem-<hash>` (geodesic) on the dask backends, while numpy and cupy returned `None`.

- Reset `result.name` after constructing the output DataArray so all four backends agree, matching the existing fix in `slope()` (#2838) and the same bug class in zonal (#2611) and focal (#2733).
- Surgical change: only the `.name` handling in `aspect()`. The other open aspect findings (dask memory guard, unit-mismatch warning, z_unit error message) are out of scope and handled separately.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy, for both `method='planar'` and `method='geodesic'`.

Test plan:
- [x] New parametrized name-consistency tests over all four backends (planar + geodesic), asserting `aspect(raster, name=name).name == name` for `name in {None, 'aspect'}`.
- [x] Confirmed the dask test fails before the fix (`_trim-<hash>` leak) and passes after.
- [x] Full `test_aspect.py` and `test_geodesic_aspect.py` suites pass (243 tests). cupy variants skip without a GPU.

### DIFF
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -469,11 +469,17 @@ def aspect(agg: xr.DataArray,
         )
         out = mapper(agg)(agg.data, lat_2d, lon_2d, WGS84_A2, WGS84_B2, z_factor, boundary)
 
-    return xr.DataArray(out,
-                        name=name,
-                        coords=agg.coords,
-                        dims=agg.dims,
-                        attrs=agg.attrs)
+    result = xr.DataArray(out,
+                          name=name,
+                          coords=agg.coords,
+                          dims=agg.dims,
+                          attrs=agg.attrs)
+    # On dask backends, xr.DataArray keeps the dask array's internal graph
+    # token as .name when name=None, so reset it post-construction to match
+    # the numpy/cupy backends. (Same fix as zonal #2611, focal #2733,
+    # slope #2838.)
+    result.name = name
+    return result
 
 
 @supports_dataset
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -555,11 +555,15 @@ def northness(agg: xr.DataArray,
     else:
         trig = np.cos(np.deg2rad(asp_data))
         out = np.where(asp_data == -1, np.nan, trig)
-    return xr.DataArray(out,
-                        name=name,
-                        coords=agg.coords,
-                        dims=agg.dims,
-                        attrs=agg.attrs)
+    result = xr.DataArray(out,
+                          name=name,
+                          coords=agg.coords,
+                          dims=agg.dims,
+                          attrs=agg.attrs)
+    # Reset .name post-construction so dask backends don't leak the graph
+    # token when name=None, matching aspect()/slope() (#2841, #2838).
+    result.name = name
+    return result
 
 
 @supports_dataset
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Fix aspect() .name leak on dask backends when name=None

None.

- `northness()` and `eastness()` in the same file have the same leak. Both call `aspect(..., name='_aspect')` internally, then build their own `xr.DataArray` (aspect.py:552 and aspect.py:632) without resetting `.name`. On dask backends they come back as `where-<hash>` when `name=None`. I confirmed it by running both with `name=None` on a dask raster. They share the file and take the exact same one-line fix, so worth folding the `result.name = name` reset into this PR.

None.

### What looks good
- The fix mirrors slope.py: assign to `result`, then reset `result.name`. The comment cites the prior fixes (zonal #2611, focal #2733, slope #2838).
- Tests cover all four backends for both planar and geodesic, parametrized over `name` in `{None, 'aspect'}`.
- Confirmed the dask test fails before the fix (`_trim-<hash>`) and passes after.

### Checklist
- Algorithm unchanged (metadata-only fix): N/A
- All implemented backends produce a consistent `.name`: yes
- NaN handling: unchanged
- Edge cases covered by tests: `name` in `{None, str}` across backends and methods
- No premature materialization: yes, no new compute
- Benchmark: not needed (no compute change)
- README feature matrix: not applicable (no new function)
- Docstrings: unchanged, accurate

brendancol (commented):
## PR Review (follow-up after 34fe2514)

### Blockers
None.

### Suggestions
None remaining. The earlier suggestion is resolved: `northness()` and `eastness()` now reset `result.name` post-construction (aspect.py), matching `aspect()`. The name-consistency tests were extended to cover both across all four backends.

### Nits
None.

### What looks good
- All three public functions in aspect.py (aspect, northness, eastness) reset `.name` so dask backends no longer leak the graph token when `name=None`.
- Tests parametrize over `name` in `{None, 'aspect'}` across numpy, cupy, dask+numpy, and dask+cupy; aspect also covers planar and geodesic. 243 tests pass locally (cupy variants skip without a GPU).

### Checklist
- All backends produce a consistent `.name`: yes
- NaN handling: unchanged
- No premature materialization: yes, no new compute
- Benchmark: not needed
- Docstrings: unchanged, accurate

PR #2846, opened 2026-06-02
Guard dask geodesic aspect against single-chunk OOM

### LINKED ISSUE #2840: Dask geodesic aspect skips the chunk memory guard
**Describe the bug**

The Dask path in geodesic aspect skips the chunk memory guard. In `xrspatial/aspect.py`, the `method='geodesic'` branch checks `is_dask` and skips `_check_geodesic_memory` for any Dask-backed array. So a Dask array whose single chunk spans the whole raster gets no guard at all, and it can still OOM at compute time.

The slope module already handles this. `slope()` calls `_check_geodesic_memory_backend_aware(agg, func_name='slope')` (in `xrspatial/slope.py`). That helper is backend-aware: for eager numpy/cupy it sizes against the full raster, and for Dask it sizes against the largest chunk plus the map_overlap halo. A single huge chunk has no memory advantage over eager, so the helper still rejects it.

Aspect should call the same helper so its Dask geodesic path gets a real guard instead of none.

**Expected behavior**

`aspect(raster, method='geodesic')` on a Dask array with one chunk covering the whole raster should raise `MemoryError` when the chunk exceeds available memory, the same way `slope()` does. A small-chunked raster should still stream through fine.

**Additional context**

The helper already lives in `xrspatial/geodesic.py`. The fix swaps the import and replaces the inline `is_dask` check in aspect with the shared helper.

brendancol:
Closes #2840

## What changed

- The geodesic `aspect()` path skipped the memory guard for any dask-backed array, so a dask array whose only chunk spans the whole raster got no guard and could still OOM at compute time.
- Switched aspect to `_check_geodesic_memory_backend_aware`, the same helper `slope()` already uses. It sizes against the full raster for eager numpy/cupy and against the largest chunk plus the `map_overlap` halo for dask, so a single huge chunk is still rejected while small-chunked rasters stream through.
- Dropped the now-unused `has_dask_array` import and the inline `is_dask` check.

## Backend coverage

The guard now fires correctly on numpy, cupy, dask+numpy, and dask+cupy. The fix only changes which guard runs before the existing per-backend compute paths; those paths are unchanged.

## Test plan

- [x] Added `test_dask_single_huge_chunk_still_rejected` (mirrors slope's equivalent). Verified it fails on the old code (`DID NOT RAISE`) and passes with the fix.
- [x] `pytest xrspatial/tests/test_geodesic_aspect.py xrspatial/tests/test_aspect.py` — 236 passed.
- [x] flake8 clean on the changed lines.

### DIFF
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -15,11 +15,12 @@ import xarray as xr
 from numba import cuda
 
 from xrspatial.dataset_support import supports_dataset
-from xrspatial.geodesic import (INV_2R, WGS84_A2, WGS84_B2, _check_geodesic_memory,
+from xrspatial.geodesic import (INV_2R, WGS84_A2, WGS84_B2,
+                                _check_geodesic_memory_backend_aware,
                                 _cpu_geodesic_aspect, _run_gpu_geodesic_aspect)
 from xrspatial.utils import (Z_UNITS, ArrayTypeFunctionMapping, _boundary_to_dask,
                              _extract_latlon_coords, _pad_array, _validate_boundary,
-                             _validate_raster, cuda_args, get_dataarray_resolution, has_dask_array,
+                             _validate_raster, cuda_args, get_dataarray_resolution,
                              ngjit)
 
 
@@ -450,14 +451,7 @@ def aspect(agg: xr.DataArray,
             )
         z_factor = Z_UNITS[z_unit]
 
-        # The full-raster memory guard only applies to in-memory (numpy/cupy)
-        # arrays, which materialize the whole (3, H, W) stack at once. Dask
-        # backends process the raster chunk-by-chunk via map_overlap, so peak
-        # memory is bounded by chunk size, not full-raster size.
-        is_dask = has_dask_array() and isinstance(agg.data, da.Array)
-        if not is_dask:
-            rows, cols = agg.shape[-2], agg.shape[-1]
-            _check_geodesic_memory(rows, cols, func_name='aspect')
+        _check_geodesic_memory_backend_aware(agg, func_name='aspect')
 
         lat_2d, lon_2d = _extract_latlon_coords(agg)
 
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Guard dask geodesic aspect against single-chunk OOM

None.

None.

- The new test only exercises `dask+numpy` for the single-huge-chunk rejection. That matches the existing convention in `test_geodesic_slope.py`, and the guard inspects `.chunks` rather than the array backend, so a separate dask+cupy case would not add real coverage. Fine to leave as is.

### What looks good
- The fix is the right one: aspect now calls `_check_geodesic_memory_backend_aware`, the same helper `slope()` already uses, instead of the inline `is_dask` check that skipped the guard for every dask array. A single chunk spanning the whole raster is now rejected, while small-chunked rasters still stream through.
- The removed `has_dask_array` import is confirmed unused after the change; `da` is still imported and used elsewhere in the module, so dropping only `has_dask_array` is correct.
- Compute paths are untouched. Only the pre-compute guard changed, so there is no numeric or backend-parity risk.
- The added test was verified to fail on the pre-fix code (DID NOT RAISE) and pass after, so it genuinely pins the bug.

### Checklist
- [x] All implemented backends produce consistent results (compute paths unchanged)
- [x] Edge cases are covered by tests (single huge chunk, chunked, eager)
- [x] Dask chunk boundaries handled correctly (guard sizes against largest chunk + halo)
- [x] Benchmark exists or is not needed (not needed; guard is a cheap pre-check)
- [x] README feature matrix updated (n/a; no new function or backend change)
- [x] Docstrings present and accurate (no API change)

brendancol (commented):
## Follow-up review

No follow-up commits were needed. The first review found no Blockers and no Suggestions. The single Nit (adding a dask+cupy single-huge-chunk case) was dismissed with reason: the guard inspects `.chunks` rather than the array backend, so a cupy variant would not exercise a new code path, and the existing `test_geodesic_slope.py` follows the same convention of only covering dask+numpy for that case.

The diff is unchanged since the first review. Re-checked:
- Aspect now calls `_check_geodesic_memory_backend_aware`; the inline `is_dask` skip is gone, so a single chunk spanning the whole raster is rejected.
- Removed `has_dask_array` import confirmed unused; `da` still used elsewhere.
- Compute paths untouched, so no backend-parity or numeric risk.

Disposition: 0 fixed, 0 deferred, 1 dismissed (with reason above). No code changes warranted by the review.

PR #2853, opened 2026-06-02
Raise a clear error for non-2D custom kernels

### LINKED ISSUE #2842: custom_kernel leaks a raw unpacking error for non-2D kernels
**Describe the bug**

Passing a 1D (or any non-2D) NumPy array as the kernel to `apply`, `focal_stats`, or `hotspots` raises a confusing error from `custom_kernel` in `xrspatial/convolution.py`:

```
ValueError: not enough values to unpack (expected 2, got 1)
```

The error comes from `rows, cols = kernel.shape`, which assumes the kernel is 2D. Nothing checks the kernel's dimensionality first, so the user sees an unpacking error instead of a message that says what went wrong.

**Expected behavior**

`custom_kernel` should check that the kernel is 2D and raise a `ValueError` that names the requirement and the bad shape, e.g. "kernel must be a 2D array; got shape (3,)".

**To reproduce**

```python
import numpy as np
import xarray as xr

agg = xr.DataArray(np.ones((5, 5)))
apply(agg, kernel=np.array([1, 1, 1]))  # raises the cryptic unpacking error
```

**Additional context**

`apply`, `focal_stats`, and `hotspots` all route the kernel through `custom_kernel`, so the check belongs there and covers all three entry points.

brendancol:
Closes #2842

`custom_kernel` in `xrspatial/convolution.py` unpacked `rows, cols = kernel.shape` without first checking the kernel was 2D. A 1D or 3D kernel passed to `apply`, `focal_stats`, or `hotspots` (all of which route through `custom_kernel`) leaked a raw `ValueError: not enough values to unpack (expected 2, got 1)`.

- Added an `ndim != 2` check in `custom_kernel` that raises a clear `ValueError` reporting the offending shape, before the shape unpacking.
- Covered the three public entry points with 1D and 3D kernels, plus a direct `custom_kernel` unit test.

The check is backend-independent (it runs on the kernel argument before any dispatch), so it applies to numpy, cupy, dask+numpy, and dask+cupy alike.

Test plan:
- [x] New tests in test_convolution.py and test_focal.py pass
- [x] Existing convolution tests still pass

### DIFF
--- a/xrspatial/convolution.py
+++ b/xrspatial/convolution.py
@@ -325,6 +325,12 @@ def custom_kernel(kernel):
             "The kernel received was of type {} and needs to be "
             "of type `ndarray`".format(type(kernel))
         )
+    elif kernel.ndim != 2:
+        raise ValueError(
+            "Received a custom kernel that is not a 2D array.",
+            "A custom kernel needs to be a 2D array, the supplied kernel "
+            "has shape {}.".format(kernel.shape)
+        )
     else:
         rows, cols = kernel.shape
 
--- a/xrspatial/tests/test_convolution.py
+++ b/xrspatial/tests/test_convolution.py
@@ -2,7 +2,7 @@ import numpy as np
 import pytest
 import xarray as xr
 
-from xrspatial.convolution import circle_kernel, convolve_2d
+from xrspatial.convolution import circle_kernel, convolve_2d, custom_kernel
 
 
 KERNEL = circle_kernel(1, 1, 1)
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Raise a clear error for non-2D custom kernels

None.

None.

- convolution.py:333-335: the new ndim branch uses the same two-string-argument ValueError form as the existing odd-shape branch below it ("Received ...", "A custom kernel ..."). Worth knowing: passing two positional args to ValueError stores them as a 2-tuple in `args`, so the printed message is the tuple rather than one sentence. That is a pre-existing pattern in this function, not something this PR introduced, and matching it is the right call. Noting it only for the record.

### What looks good
- The check sits before `rows, cols = kernel.shape`, so the cryptic unpacking error can no longer surface.
- The message reports the bad shape via `kernel.shape`, which reads correctly for both the 1D case (e.g. (3,)) and higher-dim cases (e.g. (3, 3, 3)).
- Tests cover all three public entry points (apply, focal_stats, hotspots) with both 1D and 3D kernels, plus a direct unit test on custom_kernel. Each entry-point/kernel combo passes independently.
- The validation is backend-independent: it runs on the kernel argument before any ArrayTypeFunctionMapping dispatch, so numpy, cupy, dask+numpy, and dask+cupy all behave the same. No per-backend tests needed.

### Checklist
- All implemented backends produce consistent results: yes, validation runs pre-dispatch
- Edge cases covered by tests: yes (1D and 3D kernels)
- No premature materialization or unnecessary copies: yes, no array ops added
- Docstrings present and accurate: yes; the custom_kernel docstring ("Validates a custom kernel") still holds

PR #2859, opened 2026-06-02
Call has_cuda_and_cupy() in has_rtx() so GPU dispatch gates correctly

### LINKED ISSUE #2849: has_rtx() never calls has_cuda_and_cupy, so GPU dispatch fires even when CUDA is unusable
**Describe the bug**

`has_rtx()` in `xrspatial/gpu_rtx/__init__.py` returns the function object `has_cuda_and_cupy` instead of calling it:

```python
def has_rtx():
    return has_cuda_and_cupy and RTX is not None
```

`has_cuda_and_cupy` is a function, so the bare reference is always truthy. `has_rtx()` therefore reports `True` whenever `rtxpy` imports successfully, even when CUDA or CuPy is not actually usable on the machine.

**Expected behavior**

`has_rtx()` should return `False` when `has_cuda_and_cupy()` is `False`, regardless of whether `rtxpy` imported. The line should call the function:

```python
return has_cuda_and_cupy() and RTX is not None
```

**Impact**

`has_rtx()` gates GPU dispatch in `xrspatial/viewshed.py` (line 1727) and `xrspatial/hillshade.py`. When `rtxpy` is installed but the CUDA driver is unusable, dispatch takes the GPU path instead of falling back or skipping. A reviewer running the viewshed test file saw 13 GPU cases fail with `cudaErrorInsufficientDriver` instead of skipping, because the `has_rtx()`-gated skip conditions in the tests all evaluated truthy.

**Additional context**

One missing pair of parentheses. Affects the GPU dispatch path only.

brendancol:
Closes #2849

## What changed

- `has_rtx()` referenced the `has_cuda_and_cupy` function object instead of calling it, so the condition was always truthy whenever `rtxpy` imported. Added the missing call parentheses.
- Added unit tests for `has_rtx()` that patch `has_cuda_and_cupy` and `RTX` to pin every combination, including the regression case (no usable CUDA but `rtxpy` present should be `False`).

## Why it matters

`has_rtx()` gates the GPU dispatch path in `viewshed()` and `hillshade()`. With the bug, a machine where `rtxpy` is installed but CUDA/CuPy is unusable took the GPU path instead of falling back to CPU or skipping, which surfaced as `cudaErrorInsufficientDriver` (13 viewshed GPU cases failing instead of skipping).

## Backend coverage

Logic-only fix in the dispatch gate. The tests are pure unit tests with mocked inputs, so they run on any backend without a GPU. No numerical kernel changed.

## Skipped steps

Pure bug fix, no new public API, so the user-guide notebook and README feature-matrix steps do not apply. No docs reference the internal `has_rtx()` helper.

## Test plan

- [x] `pytest xrspatial/tests/test_gpu_rtx_has_rtx.py` passes (4 tests)
- [x] Verified the tests fail when the fix is reverted (2 of 4 fail), then pass with the fix

### DIFF
--- a/xrspatial/gpu_rtx/__init__.py
+++ b/xrspatial/gpu_rtx/__init__.py
@@ -7,4 +7,4 @@ except ImportError:
 
 
 def has_rtx():
-    return has_cuda_and_cupy and RTX is not None
+    return has_cuda_and_cupy() and RTX is not None
--- a/xrspatial/tests/test_gpu_rtx_has_rtx.py
+++ b/xrspatial/tests/test_gpu_rtx_has_rtx.py
@@ -1,0 +1,50 @@
+"""Tests for ``has_rtx()`` dispatch gating (Issue #2849).
+
+``has_rtx()`` gates the GPU dispatch path in ``viewshed()`` and
+``hillshade()``.  It must return ``False`` whenever CUDA/CuPy is unusable,
+even if ``rtxpy`` imported.  A regression here once let ``has_rtx()`` report
+``True`` on machines without a working CUDA driver because the call
+parentheses on ``has_cuda_and_cupy`` were missing, so the bare function
+reference was always truthy.
+
+These are pure logic tests: they patch the two inputs to ``has_rtx`` and do
+not need a GPU.
+"""
+
+from unittest.mock import patch
+
+import xrspatial.gpu_rtx as gpu_rtx
+from xrspatial.gpu_rtx import has_rtx
+
+
+def test_has_rtx_false_when_no_cuda_even_if_rtx_present():
+    """If CUDA/CuPy is unusable, has_rtx() is False even when RTX imported."""
+    with patch.object(gpu_rtx, "has_cuda_and_cupy", return_value=False), \
+            patch.object(gpu_rtx, "RTX", object()):
+        assert has_rtx() is False
+
+
+def test_has_rtx_false_when_rtx_missing_even_if_cuda_present():
+    """If rtxpy did not import, has_rtx() is False even when CUDA is usable."""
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Call has_cuda_and_cupy() in has_rtx() so GPU dispatch gates correctly

None.

None.

- xrspatial/tests/test_gpu_rtx_has_rtx.py: the test patches `xrspatial.gpu_rtx.has_cuda_and_cupy` and `RTX`. That is the right target, since `has_rtx` looks both names up in the `gpu_rtx` module globals where they are imported. No change needed; noting it so the patch targets are not "fixed" later to point at `xrspatial.utils`, which would stop exercising the gate.

### What looks good
- The fix matches the reported bug. `has_cuda_and_cupy` was a bare function reference (always truthy) instead of a call. Adding `()` makes `has_rtx()` return False when CUDA/CuPy is unusable.
- Every other `has_rtx()` caller benefits for free (viewshed.py 2115/2156/2209, hillshade.py), since they all route through this one gate.
- `test_has_rtx_calls_has_cuda_and_cupy` asserts the mock was actually called, so a future edit that drops the parentheses fails loudly instead of slipping past a truthiness check. I reverted the fix locally and confirmed 2 of the 4 tests fail, then pass again with the fix.
- Pure logic tests with mocked inputs, so they run on any backend without a GPU.

### Checklist
- Algorithm matches reference: n/a (dispatch gate, no algorithm)
- All implemented backends consistent: n/a (gate is backend-independent; tests mock inputs)
- NaN handling: n/a
- Edge cases covered: yes (all four combinations of has_cuda_and_cupy x RTX)
- Dask chunk boundaries: n/a
- Benchmark: not needed
- README feature matrix: not needed (no public API change)

PR #2860, opened 2026-06-02
Fix hotspots() silent all-zeros on degenerate dask inputs

### LINKED ISSUE #2843: hotspots() silently returns all-zeros for degenerate dask inputs where NumPy/CuPy raise
## Describe the bug
`hotspots()` returns an all-zeros ("no significance") raster for degenerate dask inputs, while the NumPy and CuPy backends raise an error on the same input.

The NumPy and CuPy paths validate the global Gi* terms through `_gistar_global_stats`, which raises `ValueError` when there are fewer than 2 valid (non-NaN) cells, and `ZeroDivisionError` when the global standard deviation is 0 (a constant raster). The dask paths (`_hotspots_dask_numpy`, `_hotspots_dask_cupy`) keep the global mean, std, and valid-cell count as lazy 0-d dask arrays and never run that validation. `_gistar_zscore` maps every non-finite z-score to 0, so a constant raster, an all-NaN raster, or a raster with a single valid cell computes to an all-zeros result that looks valid instead of failing.

So the same input errors on NumPy/CuPy but quietly answers "no hotspots" on dask. A downstream user gets a plausible-looking statistic with no hint that the input was degenerate.

## To Reproduce
```python
import numpy as np
import xarray as xr
import dask.array as da
from xrspatial.focal import hotspots

kernel = np.ones((3, 3))
data = np.zeros((10, 10))  # constant raster -> global std == 0

# NumPy raises ZeroDivisionError
hotspots(xr.DataArray(data), kernel)

# dask quietly returns all zeros
hotspots(xr.DataArray(da.from_array(data, chunks=(5, 5))), kernel).compute()
```

## Expected behavior
The dask backends should signal the degenerate input the same way NumPy does instead of returning a silent all-zeros raster. The dask path is lazy by design (issue #2772), so the validation should fire at compute time, not when the graph is built.

## Additional context
- File: `xrspatial/focal.py`
- NumPy validation: `_gistar_global_stats`
- Dask paths missing it: `_hotspots_dask_numpy`, `_hotspots_dask_cupy`
- Affected backends: dask+numpy, dask+cupy

brendancol:
Closes #2843

## What

- The numpy and cupy `hotspots()` paths validate the global Gi* terms via `_gistar_global_stats`, raising for fewer than 2 valid cells (`ValueError`) or a zero global std (`ZeroDivisionError`). The dask paths kept those terms as lazy 0-d arrays and never ran the check, so constant, all-NaN, or single-valid-cell rasters classified to a silent all-zeros raster instead of raising.
- Added `_gistar_validate_lazy`, a `map_blocks` step that re-applies the numpy-path check at compute time and returns the z-score unchanged. The graph stays lazy on call (issue #2772); the error now fires at `compute()` to match numpy and cupy.

## Backends

numpy and cupy unchanged. dask+numpy and dask+cupy now raise the same errors as numpy on degenerate input.

## Test plan

- [x] `test_hotspots_degenerate_*_2843` across numpy / cupy / dask+numpy / dask+cupy for constant, all-NaN, and single-valid-cell rasters
- [x] Existing hotspots tests still pass, including the dask laziness checks (`test_hotspots_no_eager_compute`)

### DIFF
--- a/docs/source/reference/dask_laziness.rst
+++ b/docs/source/reference/dask_laziness.rst
@@ -78,8 +78,8 @@ Focal operations
      - Fully lazy
      - Multiple stats via ``map_overlap``, 3D output
    * - ``hotspots``
-     - Partially lazy
-     - Computes global mean and std, result is dask
+     - Fully lazy
+     - Global mean/std/count stay lazy; degenerate-input check fires at compute
 
 
 Classification
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -1371,6 +1371,20 @@ def _gistar_global_stats(global_mean, global_std, n):
     return global_mean, global_std, n
 
 
+def _gistar_validate_lazy(z_block, global_std, n):
+    """Validate the global Gi* terms inside the dask graph.
+
+    The dask backends keep ``global_std`` and ``n`` as lazy 0-d arrays, so
+    the eager ``_gistar_global_stats`` check never runs on them. This block
+    function re-applies that check at compute time and returns ``z_block``
+    unchanged, so degenerate inputs (constant raster, all-NaN raster, or a
+    single valid cell) raise the same errors as the numpy and cupy paths
+    instead of silently classifying to all zeros.
+    """
+    _gistar_global_stats(0.0, float(global_std), int(n))
+    return z_block
+
+
 def _gistar_zscore(weighted_sum, weight_sum, sq_weight_sum,
                    global_mean, global_std, n):
     """Getis-Ord Gi* z-score from the per-cell convolution terms.
@@ -1462,6 +1476,11 @@ def _hotspots_dask_numpy(raster, kernel, boundary='nan'):
     # per block.
     z_array = _gistar_zscore(weighted_sum, weight_sum, sq_weight_sum,
                              global_mean, global_std, n)
+    # Re-apply the numpy-path degenerate-input check lazily so constant /
+    # all-NaN / single-valid-cell rasters raise at compute time instead of
+    # classifying to a silent all-zeros raster (issue #2843).
+    z_array = da.map_blocks(_gistar_validate_lazy, z_array, global_std, n,
+                            dtype=z_array.dtype, meta=z_array._meta)
     out = z_array.map_blocks(_calc_hotspots_numpy,
                              meta=np.array((), dtype=np.int8))
     return out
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Fix hotspots() silent all-zeros on degenerate dask inputs

None.

None.

- `_gistar_validate_lazy` runs once per chunk via `map_blocks`, so the check repeats for every block of a multi-chunk raster. It is two scalar comparisons and only one block needs to raise, so this is harmless. Putting the guard in the block function (rather than a single up-front check) is what keeps the graph lazy, which is the point of #2772. (focal.py:1374)

### What looks good
- The fix keeps the laziness contract from #2772: validation is folded into the dask graph via `map_blocks` and fires at `compute()`, not at graph-build time. `test_hotspots_no_eager_compute` still passes.
- `_gistar_validate_lazy` reuses `_gistar_global_stats`, so the dask error messages and exception types match numpy/cupy exactly (ValueError for n<2, ZeroDivisionError for std==0).
- `meta=z_array._meta` keeps the chunk element type, so the dask+cupy path stays on-device. `._meta` is already used this way in kde.py and utils.py.
- Tests cover all three degenerate cases (constant, all-NaN, single valid cell) across numpy, cupy, dask+numpy, and dask+cupy, with the dask tests asserting the error fires at compute().

### Checklist
- [x] Algorithm matches reference (no algorithm change; validation parity only)
- [x] NaN handling is correct (all-NaN -> n=0 -> ValueError)
- [x] Dask chunk boundaries handled correctly (0-d scalars broadcast to every block)
- [x] No premature materialization (global terms stay lazy)
- [x] Benchmark not needed (bugfix, no perf change)

brendancol (commented):
## Follow-up review (after docs commit)

Re-reviewed after the second commit. The only change since the first review is a docs fix.

### Changes since last review
- `docs/source/reference/dask_laziness.rst`: the hotspots row still said "Partially lazy / Computes global mean and std," which went stale when #2772 made the dask path lazy. Updated it to "Fully lazy" with a note that the global terms stay lazy and the degenerate-input check fires at compute. This matches the code and `test_hotspots_no_eager_compute`.

### Disposition of earlier findings
- No Blockers or Suggestions were raised.
- The single Nit (validation runs per block) is intentional: putting the guard inside the block function is what keeps the graph lazy per #2772. Dismissed, not a defect.

No new issues. Code is unchanged from the first review.

PR #2862, opened 2026-06-02
Reject non-binary kernels in apply() and focal_stats()

### LINKED ISSUE #2848: apply() and focal_stats() give backend-dependent results on non-binary kernels
**Describe the bug**

`apply()` and `focal_stats()` in `xrspatial/focal.py` document the `kernel` argument as "2D array where values of 1 indicate the kernel", i.e. a binary membership mask. But the CPU and GPU paths disagree on what a non-binary value (say, 2) means, so the same call can return different numbers depending on the backend.

CPU: `_apply_numpy` only copies a neighbour into the window when `kernel[...] == 1`. A cell with weight 2 is neither masked out (0) nor included (1), so it gets dropped. `focal_stats` on numpy and dask+numpy delegates to `apply` and inherits this.

GPU: `_focal_sum_cuda` and `_focal_mean_cuda` treat every nonzero cell as a weight and accumulate `w * v`, so a 2 doubles that neighbour's contribution. The min/max/range/variety kernels treat nonzero as plain membership and ignore the weight, which is yet another interpretation of the same kernel.

So `focal_stats(..., stats_funcs=['mean', 'sum'])` and `apply(...)` can give different answers on the same non-binary kernel depending which backend runs.

**Expected behavior**

Same result across all four backends for a given kernel. The documented contract is a binary mask, and weighting is supposed to happen inside the user's `func` (the `apply` docstring example does exactly that). So `apply` and `focal_stats` should reject non-binary kernels with a clear error instead of silently producing backend-dependent numbers. `convolve_2d` and `hotspots` genuinely use weighted kernels and are unaffected.

**Additional context**

Affected functions: `apply`, `focal_stats`. Backends: numpy, cupy, dask+numpy, dask+cupy.

brendancol:
Closes #2848

`apply()` and `focal_stats()` document the kernel as a binary membership mask ("values of 1 indicate the kernel"), but the backends disagreed on what a non-binary value meant:

- CPU `_apply_numpy` only kept cells where `kernel == 1`, so a weight of 2 was dropped.
- GPU `_focal_sum_cuda` / `_focal_mean_cuda` weighted every nonzero cell by its value.

Same kernel, different numbers depending on the backend.

This rejects non-binary kernels in both APIs with a clear `ValueError`, so all four backends agree. Per-cell weighting still belongs inside the user's `func`, or in `convolve_2d` / `hotspots`, which handle weighted kernels directly and are untouched. The docstrings now state the binary contract.

Test plan:
- Non-binary kernels raise ValueError on every backend (cross-backend tests).
- Binary 0/1 kernels still work on every backend.
- Regression test pins that apply(mean) and focal_stats(mean) agree on a binary kernel.
- Full test_focal.py passes (214 tests).

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -33,6 +33,31 @@ from xrspatial.utils import (ArrayTypeFunctionMapping, _boundary_to_dask, _pad_a
                              ngjit)
 
 
+def _validate_binary_kernel(kernel, func_name):
+    """Reject non-binary kernels for the mask-based focal APIs.
+
+    ``apply`` and ``focal_stats`` document the kernel as "2D array where
+    values of 1 indicate the kernel" -- a binary membership mask, not a
+    weight array.  The CPU and GPU code paths disagree on what a value
+    other than 0 or 1 means: ``_apply_numpy`` only copies cells where
+    ``kernel == 1`` (so a weight of 2 is dropped), while the GPU sum/mean
+    kernels treat every nonzero cell as a weight (``w * v``).  The result
+    is backend-dependent output for the same call.
+
+    Weighting belongs inside the user-supplied ``func`` (see the ``apply``
+    docstring example) or in ``convolve_2d`` / ``hotspots``, which handle
+    weighted kernels directly.  Reject anything that is not strictly 0/1
+    here so all four backends agree.
+    """
+    if not np.all((kernel == 0) | (kernel == 1)):
+        raise ValueError(
+            f"{func_name}(): kernel must be binary (only 0 and 1 values); "
+            "it is used as a membership mask, not a weight array. Apply "
+            "per-cell weights inside the func argument, or use convolve_2d "
+            "for a weighted convolution."
+        )
+
+
 def _check_kernel_vs_raster_memory(kernel, rows, cols, func_name):
     """Reject kernel + raster combinations that would OOM the host.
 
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Reject non-binary kernels in apply() and focal_stats()

None.

None.

- focal.py: a kernel containing NaN also gets rejected by this guard, since `np.nan == 0` and `np.nan == 1` are both False. That is the right call (NaN is not a valid mask value), but the error message only mentions "0 and 1 values" and does not call out NaN. Not worth a code change; flagging it so the behaviour is on record.

### What looks good
- The fix follows the documented contract ("values of 1 indicate the kernel") instead of inventing weighted semantics, so results are unchanged for any kernel that was already valid. The existing 214 tests pass as-is.
- Validation sits in one helper called from both `apply` and `focal_stats` right after `custom_kernel`, so all four backends are gated at the same point. `convolve_2d` and `hotspots`, which genuinely use weighted kernels, are left alone.
- `np.all((kernel == 0) | (kernel == 1))` handles int, float, and bool kernels. `circle_kernel` / `annulus_kernel` emit float 0.0/1.0 and still pass.
- 3D rasters route through `_apply_per_band`, which recurses into the public function per band, so the guard fires there too.
- Tests exercise both APIs on numpy, cupy, dask+numpy, and dask+cupy: non-binary kernels raise, binary kernels still work, and the apply/focal_stats agreement from the issue is pinned.

### Checklist
- Algorithm matches contract: yes (binary mask)
- All backends consistent: yes (all reject non-binary)
- NaN handling: NaN kernel cells rejected, consistent with mask semantics
- Edge cases covered: non-binary, mixed, and fractional kernels tested
- Dask boundaries: unaffected (validation runs before dispatch)
- No premature materialization: validation is a single np.all on the kernel
- Benchmark: not needed (validation-only change)
- README matrix: not applicable (no new function, no backend change)
- Docstrings: updated to state the binary contract

brendancol (commented):
## Follow-up review

The one nit from the first pass is resolved: the binary-kernel error message now calls out NaN explicitly, and there is a new test (`test_apply_rejects_nan_kernel`) covering the NaN-kernel rejection path.

No new findings. No blockers or suggestions outstanding. The full focal suite still passes locally.

PR #2864, opened 2026-06-02
Fix proximity dask fallback mutating the caller's input chunks

### LINKED ISSUE #2847: proximity: dask unbounded fallback mutates the caller's input chunks
**Describe the bug**

`_process_dask()` in `xrspatial/proximity.py` mutates the caller's input DataArray when `max_distance >= max_possible_distance`. At line 1385 it runs:

```python
raster.data = raster.data.rechunk({0: height, 1: width})
```

This reassigns `.data` on the input DataArray in place. It happens on the unbounded dask fallback paths: GREAT_CIRCLE and the no-scipy (no cKDTree) fallback, where the code skips the KDTree branch and calls `_process_dask` directly.

**Expected behavior**

Calling `proximity()`, `allocation()`, or `direction()` should leave the caller's input alone. After the call, the input DataArray's `.data` should keep its original chunking and object identity. The rechunk should run on a local variable, not reassign `raster.data`.

**Impact**

Hidden side effect. Code that reuses the same DataArray after calling proximity sees changed chunking, which breaks the assumption that the input is read-only.

**Additional context**

Affects the dask backends (dask+numpy and dask+cupy) on the unbounded fallback paths only. The fix should add a regression test asserting that the input DataArray's `.data` chunking and identity are unchanged after the call.

Ref: xrspatial/proximity.py:1385

brendancol:
Closes #2847

`_process_dask()` did `raster.data = raster.data.rechunk(...)` on the unbounded GREAT_CIRCLE and no-scipy fallback paths, which rebound `.data` on the caller's input DataArray. Code that reused the same DataArray afterward saw changed chunking.

## Changes
- Rechunk into a local variable in `_process_dask` instead of reassigning `raster.data`, so the input is left untouched.
- Add regression tests asserting the input `.data` keeps its identity and chunking after `proximity`/`allocation`/`direction` on both unbounded fallback paths.

## Backend coverage
The bug and fix are dask-only. Tests run on dask+numpy and dask+cupy. numpy and cupy paths never reassigned `.data` and are unaffected.

## Test plan
- [x] New mutation regression tests fail on the old code and pass on the fix
- [x] Full `test_proximity.py` suite passes (291 passed)

### DIFF
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -1378,12 +1378,15 @@ def _process(
 
     def _process_dask(raster, xs, ys):
 
+        # Rechunk into a local variable instead of reassigning raster.data,
+        # which would mutate the caller's input DataArray (issue #2847).
+        data = raster.data
         if max_distance >= max_possible_distance:
             # consider all targets in the whole raster
             # the data array is computed at once,
             # make sure your data fit your memory
             height, width = raster.shape
-            raster.data = raster.data.rechunk({0: height, 1: width})
+            data = data.rechunk({0: height, 1: width})
             xs = xs.rechunk({0: height, 1: width})
             ys = ys.rechunk({0: height, 1: width})
             pad_y = pad_x = 0
@@ -1393,7 +1396,7 @@ def _process(
 
         out = da.map_overlap(
             _process_numpy,
-            raster.data, xs, ys,
+            data, xs, ys,
             depth=(pad_y, pad_x),
             boundary=np.nan,
             meta=np.array((), dtype=np.float32),
--- a/xrspatial/tests/test_proximity.py
+++ b/xrspatial/tests/test_proximity.py
@@ -1621,3 +1621,62 @@ def test_great_circle_numpy_off_by_more_than_a_metre_is_fixed(
         raster, x='lon', y='lat', distance_metric='GREAT_CIRCLE').data
 
     assert np.nanmax(np.abs(result - expected)) < 1.0
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Fix proximity dask fallback mutating the caller's input chunks

None.

None.

- test_proximity.py: on a host without CUDA, the `dask+cupy` parametrization runs the same code as `dask+numpy`, because `_backend_raster` only swaps the buffer to cupy when CUDA is present. So the dask+cupy cases give real coverage only on a GPU runner. That matches the rest of the file and needs no change; just flagging that the dask+cupy assertions are duplicate work off-GPU.

### What looks good
- The fix is minimal and correct. `data = raster.data` followed by `data = data.rechunk(...)` leaves the caller's DataArray untouched, since dask `.rechunk` returns a new array rather than mutating in place. The `map_overlap` call now reads from the local `data`.
- The dask+cupy unbounded path was already safe (it rebinds `raster` to a `.copy()` before reaching `_process_dask`); this fix covers the dask+numpy unbounded path and the converted dask+cupy case without touching that.
- Tests assert both object identity (`raster.data is data_before`) and chunking equality, and cover all three public entry points (proximity, allocation, direction) on both unbounded fallback paths. The tests fail on the pre-fix code and pass after.

### Checklist
- Algorithm unchanged; only the chunk-handling side effect was removed.
- NaN handling unchanged.
- Dask chunk boundaries handled correctly; rechunk semantics preserved.
- No premature materialization or extra copies introduced.
- No README or docstring change needed (pure bug fix, no API change).

PR #2868, opened 2026-06-02
Reject non-finite target_values in proximity tools

### LINKED ISSUE #2850: proximity/allocation/direction: non-finite target_values give backend-dependent output
**Describe the bug**

`proximity`, `allocation`, and `direction` accept a `target_values` list. When that list contains a non-finite value such as `np.inf`, the backends disagree on whether those values count as targets.

On NumPy, `_is_target_value` (xrspatial/proximity.py:346) checks only `v == target_values[k]` for explicit targets, with no finiteness guard. A pixel holding `inf` matches `target_values=[np.inf]`, so the search runs and returns full output.

On CuPy (proximity.py:516-518) and Dask (`_target_mask`, proximity.py:780-784), the mask is `isin(data, target_values) & isfinite(data)`. The finiteness term drops `inf` pixels even when the user asked for them explicitly. With no surviving targets the kernels fill the result with NaN.

So the same raster and the same `target_values=[np.inf]` returns a real distance/allocation/direction grid on NumPy and an all-NaN grid on Dask and CuPy.

**Expected behavior**

One behavior across all four backends: numpy, cupy, dask+numpy, dask+cupy.

A non-finite `target_values` entry (inf, -inf, nan) is not a meaningful raster category. `nan` can never match a pixel anyway because `nan == nan` is False, and `inf` only "matches" on NumPy by accident. The fix should reject non-finite `target_values` up front with a clear ValueError, consistent with how the function already validates `distance_metric` and `max_distance`. The empty-`target_values` default path, which already ignores non-finite pixels on every backend, stays unchanged.

**Additional context**

Refs: xrspatial/proximity.py:346, xrspatial/proximity.py:516, xrspatial/proximity.py:780. Central dispatch is `_process` (proximity.py:1152), which already normalizes `target_values` with `np.asarray` and is the natural place for the check.

brendancol:
Closes #2850

## What

`proximity`, `allocation`, and `direction` treated non-finite `target_values` differently per backend. NumPy matched `inf`-valued pixels for `target_values=[np.inf]` and returned a real grid, while dask and cupy masked non-finite pixels out and returned all-NaN for the same raster.

- Validate `target_values` in the shared `_process` dispatch, next to the existing `distance_metric` and `max_distance` checks. A non-finite entry (inf, -inf, nan) now raises `ValueError` on every backend.
- The empty-`target_values` default path is unchanged. It already ignores non-finite pixels everywhere.
- Note the finiteness requirement in the three `target_values` docstrings.

## Chosen resolution

Reject up front rather than silently ignore. A non-finite target is not a meaningful raster category: `nan` can never match a pixel (`nan == nan` is False), and `inf` only "matched" on NumPy by accident. An explicit error is clearer than dropping entries silently and matches the existing validation style in `_process`.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy. The validation lives in the shared dispatch ahead of backend branching, so all four paths raise identically.

## Test plan

- [x] New `test_non_finite_target_values_raises` over [inf, -inf, nan, [2, inf]] x {proximity, allocation, direction} x 4 backends
- [x] New `test_finite_target_values_run` confirms the default path and explicit finite targets still work on every backend
- [x] Full `test_proximity.py` passes (339 tests)

### DIFF
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -1180,10 +1180,15 @@ def _process(
     # matched target_values=[inf] and ran the search, while dask/cupy mask
     # non-finite pixels out and returned all-NaN for the same input. nan can
     # never match a pixel anyway (nan == nan is False). Fail fast instead of
-    # producing backend-dependent output (issue #2850).
-    if target_values.size and not np.isfinite(target_values).all():
+    # producing backend-dependent output (issue #2850). A non-numeric dtype
+    # (e.g. strings) can't index a raster either, so it gets the same error
+    # rather than a downstream TypeError from np.isfinite.
+    if target_values.size and (
+        target_values.dtype.kind not in "iuf"
+        or not np.isfinite(target_values).all()
+    ):
         raise ValueError(
-            "target_values must all be finite, got {0!r}.".format(
+            "target_values must all be finite numbers, got {0!r}.".format(
                 target_values.tolist()
             )
         )
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -1176,6 +1176,18 @@ def _process(
 
     target_values = np.asarray(target_values)
 
+    # Reject non-finite explicit target_values. On numpy a pixel holding inf
+    # matched target_values=[inf] and ran the search, while dask/cupy mask
+    # non-finite pixels out and returned all-NaN for the same input. nan can
+    # never match a pixel anyway (nan == nan is False). Fail fast instead of
+    # producing backend-dependent output (issue #2850).
+    if target_values.size and not np.isfinite(target_values).all():
+        raise ValueError(
+            "target_values must all be finite, got {0!r}.".format(
+                target_values.tolist()
+            )
+        )
+
     if max_distance is None:
         max_distance = np.inf
 
@@ -1547,7 +1559,8 @@ def proximity(
     target_values: list
         Target pixel values to measure the distance from. If this option
         is not provided, proximity will be computed from non-zero pixel
-        values.
+        values. All entries must be finite; a non-finite value (inf or
+        nan) raises ValueError.
 
     max_distance: float, default=np.inf
         The maximum distance to search. Proximity distances greater than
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Reject non-finite target_values in proximity tools

None.

None.

- proximity.py:1184: `np.isfinite` raises TypeError, not ValueError, if `target_values` ends up object/string dtype (e.g. `["a"]`). That path was already broken before this PR (the old code fails downstream too), so it is out of scope here, but a follow-up that rejects non-numeric target_values with a clear message would round this out.
- test_proximity.py:364: the `cupy` and `dask+cupy` params only exercise the GPU code when CUDA is present; otherwise the `test_raster` fixture falls back to numpy data, so on a CPU-only runner those two params re-test the numpy path. This matches the existing convention in `test_invalid_max_distance_raises` in the same file, and the validation runs ahead of backend dispatch anyway, so the gap is cosmetic.

### What looks good
- The check lives in `_process` ahead of backend branching, so all four backends raise identically. Right altitude for the fix.
- The `target_values.size` guard keeps the empty-list default path untouched, and the fixture already carries inf/nan pixels that the default path still ignores. `test_finite_target_values_run` pins that.
- Error message style and placement match the neighboring `max_distance` validation.
- Tests cover inf, -inf, nan, and a mixed finite+inf list across all three public functions.

### Checklist
- [x] Behavior is consistent across backends (validation precedes dispatch)
- [x] NaN/inf handling is correct (both rejected; default path still ignores them)
- [x] Edge cases covered (empty list preserved, mixed list rejected)
- [x] No dask/cupy materialization concerns (validation is on the small 1D target_values array)
- [x] Docstrings updated on all three public functions

brendancol (commented):
## Follow-up review (after c8761de0)

- Nit 1 (non-numeric target_values gave a downstream TypeError): fixed. The validation now guards `dtype.kind not in "iuf"`, so a string or other non-numeric target_values raises the same up-front `ValueError` as a non-finite one. Added `test_non_numeric_target_values_raises` across all four backends.
- Nit 2 (cupy params fall back to numpy on CPU-only runners): dismissed. This is the established convention in `test_invalid_max_distance_raises` in the same file, and the validation runs ahead of backend dispatch, so the GPU and CPU paths can't diverge here. Matching the sibling test is the right call.

### New change check
- The dtype guard short-circuits before `np.isfinite`, so the bool/complex/string cases never reach the float-only path. Empty-list default still skipped via `target_values.size`.
- No new findings. 78 target_values tests pass locally.

PR #2870, opened 2026-06-02
Fix invalid z_unit error to list accepted unit names

### LINKED ISSUE #2858: Invalid z_unit error lists numeric factors instead of accepted unit names
**Describe the bug**

When you pass an invalid `z_unit` to `aspect(..., method='geodesic')`, the `ValueError` lists numeric conversion factors instead of the unit names you're allowed to pass. The check at `xrspatial/aspect.py:446` builds the message from `sorted(set(Z_UNITS.values()))`, which are the factors, not the accepted strings.

Pass `z_unit='metres'` and you get:

```
z_unit must be one of [0.3048, 1.0, 1000.0, 1609.344], got 'metres'
```

A list of floats doesn't tell you which names are valid (`'meter'`, `'m'`, `'ft'`, `'km'`, and so on).

`xrspatial/slope.py:415` has the same bug in its own `z_unit` check.

**Expected behavior**

The error should list the accepted unit names (the keys of `Z_UNITS`), so the message says which names work:

```
z_unit must be one of ['feet', 'foot', 'ft', 'kilometer', ...], got 'metres'
```

**Additional context**

`Z_UNITS` lives in `xrspatial/utils.py`. Its keys are the unit strings; its values are the numeric factors. Only the message construction needs to change. The validation itself (`z_unit not in Z_UNITS`) is already correct.

brendancol:
Closes #2858

## What changed

- The geodesic `z_unit` check in `aspect()` and `slope()` built its error message from `sorted(set(Z_UNITS.values()))`, which are the numeric conversion factors (`0.3048`, `1.0`, `1000.0`, `1609.344`), not the unit names a user can pass. It now reports `sorted(Z_UNITS)` so the message lists the accepted unit names.
- Fixed in both `aspect.py` and `slope.py` since they share the same check and message format.

Before:

```
z_unit must be one of [0.3048, 1.0, 1000.0, 1609.344], got 'cubit'
```

After:

```
z_unit must be one of ['feet', 'foot', 'ft', 'kilometer', 'kilometers', 'km', 'm', 'meter', 'meters', 'mi', 'mile', 'miles'], got 'cubit'
```

## Backend coverage

This is a validation-path change that runs before any backend dispatch, so it applies to all four backends (numpy / cupy / dask+numpy / dask+cupy).

## Test plan

- [x] Extended `test_invalid_z_unit_raises` in `test_geodesic_aspect.py` and `test_geodesic_slope.py` to assert the message contains the accepted unit names and not bare numeric factors.
- [x] `pytest` passes for both updated tests.

### DIFF
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -445,7 +445,7 @@ def aspect(agg: xr.DataArray,
     else:  # geodesic
         if z_unit not in Z_UNITS:
             raise ValueError(
-                f"z_unit must be one of {sorted(set(Z_UNITS.values()), key=str)}, "
+                f"z_unit must be one of {sorted(Z_UNITS)}, "
                 f"got {z_unit!r}"
             )
         z_factor = Z_UNITS[z_unit]
--- a/xrspatial/slope.py
+++ b/xrspatial/slope.py
@@ -412,7 +412,7 @@ def slope(agg: xr.DataArray,
     else:  # geodesic
         if z_unit not in Z_UNITS:
             raise ValueError(
-                f"z_unit must be one of {sorted(set(Z_UNITS.values()), key=str)}, "
+                f"z_unit must be one of {sorted(Z_UNITS)}, "
                 f"got {z_unit!r}"
             )
         z_factor = Z_UNITS[z_unit]
--- a/xrspatial/tests/test_geodesic_aspect.py
+++ b/xrspatial/tests/test_geodesic_aspect.py
@@ -213,8 +213,17 @@ class TestGeodesicAspectValidation:
     def test_invalid_z_unit_raises(self):
         elev = _flat_surface()
         raster = _make_geo_raster(elev, 40.0, 41.0, 10.0, 11.0)
-        with pytest.raises(ValueError, match="z_unit"):
+        with pytest.raises(ValueError, match="z_unit") as excinfo:
             aspect(raster, method='geodesic', z_unit='cubit')
+
+        # The message must list the accepted unit-name strings (the keys a
+        # user is allowed to pass), not the numeric conversion factors.
+        msg = str(excinfo.value)
+        assert "'meter'" in msg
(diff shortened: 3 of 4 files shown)


brendancol (commented):
## PR Review: Fix invalid z_unit error to list accepted unit names

None.

None.

- The test guards against the old factors by name (`assert "0.3048" not in msg`). That works. A stricter version would reject any bare float, e.g. `assert not re.search(r"\b\d+\.\d+", msg)`, so a future regression that swaps in different factors still trips it. Optional; the current asserts already fail against the old code.

### What looks good
- Root cause is right. `Z_UNITS` keys are the unit names users pass; the values are numeric factors. `sorted(Z_UNITS)` reports the keys, which is what the message should say.
- The same bug was in `aspect.py:448` and `slope.py:415`. Fixing both keeps the two messages in sync.
- `sorted(Z_UNITS)` sorts the keys lexically, so the message order is stable across runs.
- This runs in the validation path before backend dispatch, so there's no numpy/cupy/dask/dask-cupy parity question.
- Tests check both the presence of accepted names and the absence of the old factors, and they pass.

### Checklist
- Algorithm matches reference: n/a (error-message fix)
- All implemented backends consistent: n/a (pre-dispatch validation)
- NaN handling correct: n/a
- Edge cases covered by tests: yes (invalid unit string)
- Dask chunk boundaries handled: n/a
- README feature matrix updated: not needed (no API change)
- Docstrings present and accurate: unchanged

brendancol (commented):

The earlier nit is addressed: both `test_invalid_z_unit_raises` tests now assert `not re.search(r"\d+\.\d+", msg)`, so any bare decimal in the message fails the test rather than just the two old factors. `import re` was added to both test files. Tests pass.

No new blockers, suggestions, or nits. The remaining `F841` flake8 warning on the unused `flat` variable in `test_geodesic_aspect.py` predates this PR and sits outside the changed lines, so it is left alone.

PR #2871, opened 2026-06-02
Fix viewshed() mutating the caller's input raster (#2852)

### LINKED ISSUE #2852: viewshed() mutates the caller's input raster (dtype/array-type)
**Describe the bug**

`viewshed()` mutates the caller's input raster in place.

- CPU path, `xrspatial/viewshed.py:1584`: `raster.data = raster.data.astype(np.float64, copy=False)` replaces the input DataArray's backing array with a float64 version.
- CuPy fallback (no rtxpy), `xrspatial/viewshed.py:1735`: `raster.data = cp.asnumpy(raster.data)` swaps the caller's CuPy array for a numpy one.

So an int16 (or any non-float64) input comes back as float64 after the call, and a CuPy-backed input gets silently converted to numpy. A library function shouldn't change its caller's input, and this bites hard when downstream code reuses the same raster after calling `viewshed`.

**Repro**

```python
import numpy as np
import xarray as xr

arr = xr.DataArray(
    np.zeros((5, 5), dtype="int16"),
    coords={"y": np.arange(5.0), "x": np.arange(5.0)},
    dims=["y", "x"],
)
print(arr.dtype)        # int16
viewshed(arr, x=2, y=2)
print(arr.dtype)        # int16 expected, but now float64
```

**Expected behavior**

The internal dtype/array-type conversion should run on a copy. After `viewshed(...)` returns, the input DataArray's dtype and data are unchanged. Applies to both the numpy/CPU path and the CuPy fallback path.

**Additional context**

Found during a review of the viewshed module. Pure bug fix, no public API change.

brendancol:
Closes #2852

`viewshed()` was changing its caller's input raster in place:

- The CPU path reassigned `raster.data` to a float64 array, so an int16 input came back as float64 after the call.
- The CuPy fallback (cupy present, rtxpy absent) reassigned `raster.data` to a numpy array, silently converting the caller's CuPy input.

## Changes

- CPU path: do the float64 conversion on a local copy and feed that to the sweep kernels; `raster.data` is left alone.
- CuPy fallback: build a new numpy-backed DataArray for the CPU run instead of reassigning the input's `.data`.
- Add regression tests that assert the input dtype and data are unchanged after `viewshed()`.

## Backend coverage

- numpy: covered (int16 stays int16, data unchanged)
- dask+numpy: covered (same assertion)
- cupy: regression test for the no-rtx CPU fallback, skipped when cupy/rtx unavailable
- dask+cupy: not separately exercised; relies on the same dask/cupy paths

Pure bug fix, no public API change. Skipped the user-guide-notebook and README feature-matrix steps for that reason. No docs `.rst` change needed (no new public function).

## Test plan

- [x] `pytest xrspatial/tests/test_viewshed.py` passes (65 passed, 1 skipped, 1 xfailed)
- [x] New `test_viewshed_does_not_mutate_input` (numpy, dask) passes
- [x] Manual repro confirms int16 input stays int16 after the call

### DIFF
--- a/xrspatial/tests/test_viewshed.py
+++ b/xrspatial/tests/test_viewshed.py
@@ -827,3 +827,57 @@ def test_viewshed_max_distance_anisotropic(backend, fine_axis):
     # A cell 9 fine-axis cells away (> max_distance) stays INVISIBLE.
     far = (10, 19) if fine_axis == "x" else (19, 10)
     assert result[far] == INVISIBLE
+
+
+# -------------------------------------------------------------------
+# Input-not-mutated regression tests (issue #2852)
+# -------------------------------------------------------------------
+
+def _build_int16_raster(backend):
+    """5x5 int16 raster with a single non-zero cell, optionally cupy-backed."""
+    ny, nx = 5, 5
+    arr = np.zeros((ny, nx), dtype="int16")
+    arr[2, 2] = 3
+    xs = np.arange(nx, dtype=float)
+    ys = np.arange(ny, dtype=float)
+    if backend == "cupy":
+        import cupy as cp
+        arr = cp.asarray(arr)
+    elif backend == "dask":
+        arr = da.from_array(arr, chunks=(3, 3))
+    return xa.DataArray(arr, coords=dict(x=xs, y=ys), dims=["y", "x"])
+
+
+@pytest.mark.parametrize("backend", ["numpy", "dask"])
+def test_viewshed_does_not_mutate_input(backend):
+    """viewshed() must not change the caller's input dtype or data (#2852)."""
+    raster = _build_int16_raster(backend)
+    before = np.asarray(raster.data.compute() if backend == "dask"
+                        else raster.data).copy()
+    before_dtype = raster.dtype
+
+    viewshed(raster, x=2, y=2, observer_elev=2)
+
+    assert raster.dtype == before_dtype == np.dtype("int16")
+    after = np.asarray(raster.data.compute() if backend == "dask"
+                       else raster.data)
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Fix viewshed() mutating the caller's input raster (#2852)

None.

None.

- viewshed.py:1583: the CPU path now always copies via `astype(np.float64)` (default copy=True), even when the input is already float64. That is the safe choice here, since the float64 case is exactly where the old `copy=False` aliased the caller's buffer. Cost is one HxW copy. No change needed, just noting the tradeoff is deliberate.

### What looks good
- Root cause fixed in both reported spots. The CPU path uses a local `raster_data` instead of reassigning `raster.data`, and the CuPy fallback builds a fresh DataArray rather than swapping the input's `.data`.
- The new DataArray in the CuPy fallback carries coords, dims, and attrs across.
- Checked the sweep kernels (`_init_event_list`, `_viewshed_cpu_sweep`): they only read from the raster array, so the float64 copy is not guarding against in-place writes, it purely protects the caller.
- The dask Tier B path was already safe (it works on `raster.copy()`), and the new dask regression test pins that contract end to end.
- Tests assert both dtype and array contents are unchanged on numpy and dask, plus a cupy-fallback test guarded on cupy-present / rtx-absent.

### Checklist
- [x] Algorithm unchanged (pure input-mutation fix)
- [x] NaN handling unaffected
- [x] Edge cases: int16 input covered; mutation contract is the focus
- [x] Dask chunk boundaries unaffected (output rebuilt from copy)
- [x] No premature materialization or unnecessary copies (one deliberate float64 copy)
- [x] README feature matrix not applicable (no new function/backend)

PR #2874, opened 2026-06-02
Validate viewshed max_distance (reject negative / non-finite)

### LINKED ISSUE #2855: viewshed: max_distance is not validated (negative/NaN/inf give confusing internal errors)
**Describe the bug**

`viewshed()` doesn't validate `max_distance`. Negative, NaN, and infinite values get passed straight into the windowing math (xrspatial/viewshed.py:2096) and blow up with internal errors that never mention the actual problem.

What you get:
- `max_distance=-1.0` raises `ValueError: zero-size array to reduction operation minimum which has no identity` (a negative radius produces an empty window).
- `max_distance=nan` raises `ValueError: cannot convert float NaN to integer`.
- `max_distance=inf` raises `OverflowError: cannot convert float infinity to integer`.

A caller seeing any of these has no way to tell that `max_distance` was the culprit.

**Expected behavior**

`viewshed()` should require a finite `max_distance >= 0` and raise a clear `ValueError` (something like "max_distance must be a finite number >= 0") for negative or non-finite values. The check belongs at the public entry point so it covers all four backends (numpy, cupy, dask+numpy, dask+cupy). Valid values keep working as before.

**To reproduce**
```python
data = np.zeros((20, 20))
t = xr.DataArray(data, dims=['y', 'x'])
t['y'] = np.linspace(0, 19, 20)
t['x'] = np.linspace(0, 19, 20)
viewshed(t, x=10, y=10, observer_elev=5, max_distance=-1.0)
```

**Additional context**

Pure input-validation fix. No new public API.

brendancol:
Closes #2855

`viewshed()` didn't validate `max_distance`. Negative, NaN, and infinite
values fell through to the windowing math and raised internal errors that
never mentioned the argument (for example "zero-size array to reduction
operation minimum" for -1.0 and "cannot convert float NaN to integer" for
NaN).

What changed:
- Validate `max_distance` at the public `viewshed()` entry point, before
  backend dispatch, and raise `ValueError("max_distance must be a finite
  number >= 0, ...")` for negative or non-finite values.
- Document the requirement on the `max_distance` parameter.

Backend coverage: the check runs before dispatch, so it applies to all
four backends (numpy, cupy, dask+numpy, dask+cupy).

Test plan:
- [x] New test asserts the clear `ValueError` for -1.0, -0.5, NaN, +inf, -inf.
- [x] New test confirms finite values (0.0, 3.0) still work on numpy and dask+numpy.
- [x] Full `test_viewshed.py` passes (72 passed, 1 xfailed).

Skipped the user-guide-notebook and README-matrix steps: this is a pure
input-validation bug fix with no new public API and no change to backend
support.

### DIFF
--- a/xrspatial/tests/test_viewshed.py
+++ b/xrspatial/tests/test_viewshed.py
@@ -580,6 +580,31 @@ def test_viewshed_custom_name(backend):
     assert result.name == "my_vs"
 
 
+@pytest.mark.parametrize("bad", [-1.0, -0.5, float("nan"),
+                                 float("inf"), float("-inf")])
+def test_viewshed_invalid_max_distance_raises(bad):
+    """Negative or non-finite max_distance raises a clear ValueError (#2855).
+
+    Validation lives at the public entry point, before backend dispatch,
+    so a single numpy raster covers every backend. Previously these
+    values fell through to confusing internal errors (e.g. "zero-size
+    array to reduction operation minimum" or "cannot convert float NaN
+    to integer").
+    """
+    raster = _make_raster("numpy")
+    with pytest.raises(ValueError, match="max_distance must be a finite"):
+        viewshed(raster, x=3, y=2, observer_elev=1, max_distance=bad)
+
+
+@pytest.mark.parametrize("backend", ["numpy", "dask+numpy"])
+@pytest.mark.parametrize("good", [0.0, 3.0])
+def test_viewshed_valid_max_distance_still_works(backend, good):
+    """Finite max_distance >= 0 passes validation and returns a result."""
+    raster = _make_raster(backend)
+    result = viewshed(raster, x=3, y=2, observer_elev=1, max_distance=good)
+    assert result.shape == raster.shape
+
+
 # -------------------------------------------------------------------
 # dask+cupy backend tests
 # -------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Validate viewshed max_distance (reject negative / non-finite)

None.

None.

- xrspatial/viewshed.py:1721 and 1729: there are now two back-to-back `if max_distance is not None:` blocks. You could fold the validation into the existing block, but keeping them separate reads fine too (validation first, windowing dispatch under its own comment). Optional.

### What looks good
- Validation sits at the public entry point before backend dispatch, so all four backends (numpy, cupy, dask+numpy, dask+cupy) are covered by one check. The test docstring spells out that reasoning.
- The error message names the argument and echoes the bad value, which is the point of the fix.
- The `try/except (TypeError, ValueError)` also turns a non-numeric `max_distance` into the same clear message instead of a raw `np.isfinite` TypeError.
- Tests cover -1.0, -0.5, NaN, +inf, -inf and confirm 0.0 and 3.0 still work on numpy and dask.
- Docstring updated to state the contract.

### Checklist
- [x] Validation runs before dispatch, so all backends behave the same
- [x] NaN / inf / negative handled with a clear ValueError
- [x] Edge cases (0.0, +inf, -inf, non-numeric) covered
- [x] No new public API, so README matrix and user guide are not needed

brendancol (commented):
## Follow-up review (after 9104c38)

The one nit from the first pass is resolved: the validation and the windowed dispatch now share a single `if max_distance is not None:` branch (xrspatial/viewshed.py:1721). Behavior is unchanged and the 24 max_distance tests still pass.

No remaining Blockers, Suggestions, or Nits.

PR #2875, opened 2026-06-02
Reject non-monotonic 1D coords in proximity/allocation/direction

### LINKED ISSUE #2851: proximity/allocation/direction silently accept non-monotonic 1D coords and return wrong results
## Describe the bug

The proximity / allocation / direction functions assume monotonic 1D x and y coordinates, but nothing enforces it. Several internal steps depend on that assumption:

- `max_possible_distance` comes from the coordinate endpoints only (`x_coords[0]`, `x_coords[-1]`), which is the maximum span only when coords are monotonic.
- The dask `map_overlap` halo depth assumes array adjacency maps to spatial adjacency.
- The NumPy line-sweep propagates a nearest-target candidate between neighbouring pixels, which works only when array order matches spatial order.
- The tiled KDTree fallback convergence check uses chunk-boundary coordinate gaps as a lower bound on the distance to points outside the search region (`_min_boundary_distance`, proximity.py:888, used at proximity.py:992). That bound is invalid for non-monotonic coords, so the expanding-ring search can stop early and return wrong proximity / allocation / direction versus the brute-force NumPy result.

Non-monotonic 1D coordinates are already outside the module's contract, but they get accepted silently and produce wrong answers instead of an error.

## Expected behavior

Reject non-monotonic 1D x or y coordinates up front with a clear ValueError, the same way across all four backends (numpy, cupy, dask+numpy, dask+cupy). The error should name which axis is non-monotonic and point at the monotonic-coordinate requirement.

## Proposed fix

Add an early validation in the shared `_process` entrypoint (the single chokepoint for proximity, allocation, and direction) that checks each 1D coordinate array is monotonic and raises a clear ValueError otherwise. Document the monotonic-coordinate requirement in the public docstrings. Add cross-backend tests that pin the raised error.

Refs: xrspatial/proximity.py:888, xrspatial/proximity.py:992

brendancol:
Closes #2851

## What

`proximity`, `allocation`, and `direction` assume monotonic 1D x and y coordinates but never enforced it. Several internal steps rely on that assumption:

- `max_possible_distance` is taken from the coordinate endpoints only.
- The dask `map_overlap` halo and the NumPy line-sweep treat array adjacency as spatial adjacency.
- The tiled KDTree convergence check lower-bounds the out-of-region distance with chunk-boundary coordinate gaps (`_min_boundary_distance`).

None of those hold for a non-monotonic axis, so such input was accepted silently and produced wrong proximity/allocation/direction versus the brute-force NumPy result.

Non-monotonic coords are outside the module's contract, so this rejects them up front rather than trying to make the tiled path provably correct on data the rest of the module already cannot handle.

## Change

- Validate both 1D coordinate axes once in the shared `_process` entrypoint and raise a `ValueError` naming the non-monotonic axis. A single-element axis is still allowed (no order to violate).
- Document the monotonic-coordinate requirement in the three public docstrings and on the proximity reference page.

## Backends

The check lives in `_process`, the single chokepoint for all three functions, so it applies to numpy, cupy, dask+numpy, and dask+cupy alike.

## Test plan

- [x] Non-monotonic x and y each raise `ValueError` for all three functions across all four backends (GPU/dask+cupy skip when unavailable).
- [x] Descending (monotonic) axis still produces the reference proximity.
- [x] Single-element axis is accepted.
- [x] Full `test_proximity.py` suite passes (309 tests).

### DIFF
--- a/docs/source/reference/proximity.rst
+++ b/docs/source/reference/proximity.rst
@@ -18,6 +18,13 @@ Proximity
    into a single chunk.  Set a finite ``max_distance`` to keep memory
    bounded.
 
+.. caution::
+
+   ``proximity()``, ``allocation()``, and ``direction()`` require monotonic
+   1D ``x`` and ``y`` coordinates (strictly increasing or strictly
+   decreasing). A non-monotonic axis raises a ``ValueError``; sort the raster
+   along that axis first.
+
 Allocation
 ==========
 .. autosummary::
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -254,6 +254,35 @@ def _distance(x1, x2, y1, y2, metric):
     return np.float32(d)
 
 
+def _check_monotonic_coords(x_coords, y_coords, x, y):
+    """Reject non-monotonic 1D coordinates.
+
+    Every backend in this module assumes the 1D axis coordinates are
+    monotonic: ``max_possible_distance`` is taken from the endpoints, the
+    dask halo and the NumPy line-sweep treat array adjacency as spatial
+    adjacency, and the tiled KDTree convergence check lower-bounds the
+    out-of-region distance with chunk-boundary coordinate gaps. None of
+    those hold when a coordinate axis is not monotonic, so a non-monotonic
+    axis silently yields wrong proximity/allocation/direction. Reject it up
+    front with a clear message instead.
+
+    A single-element axis has no order to violate and is allowed.
+    """
+    for coords, name in ((x_coords, x), (y_coords, y)):
+        if len(coords) < 2:
+            continue
+        diffs = np.diff(coords)
+        ascending = np.all(diffs > 0)
+        descending = np.all(diffs < 0)
+        if not (ascending or descending):
+            raise ValueError(
+                "proximity/allocation/direction require monotonic "
+                "(strictly increasing or strictly decreasing) 1D "
+                "coordinates, but the {0!r} axis is not monotonic. Sort the "
+                "raster along {0!r} before calling.".format(name)
+            )
+
+
 def _halo_depth(x_coords, y_coords, max_distance, distance_metric):
     """Overlap depth in pixels for the bounded dask map_overlap call.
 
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Reject non-monotonic 1D coords in proximity/allocation/direction

None.

- proximity.py:271-277. The check also rejects equal adjacent coordinates (a duplicate axis value makes both `ascending` and `descending` false) and coords that contain NaN. Both are reasonable to reject, but the error always says "non-monotonic", which is a bit off for the duplicate/NaN case. Not worth a code change unless you want a more specific message. Flagging it so it is a conscious call.

- test_proximity.py:1700. `test_single_element_axis_allowed` covers numpy and dask+numpy but not cupy. The single-element path runs before dispatch so it is backend-agnostic and this is fine, but a cupy parametrization would match the other two new tests.

### What looks good
- The validation lives in `_process`, the single chokepoint for all three public functions, so one check covers all four backends. Right place for it: it runs after dask coords are computed to numpy and before any backend-specific path.
- Length-1 axes are explicitly allowed, matching the existing `_halo_depth` single-coordinate handling.
- Tests cover both axes, all three functions, all four backends (GPU skipped when unavailable), plus the descending-accept and single-element-accept paths.
- Docstrings and the reference page document the new contract.

### Checklist
- [x] Reject is the defensible stance (the whole module assumes monotonic coords)
- [x] All backends consistent (check runs pre-dispatch)
- [x] NaN handling correct (NaN coords rejected, NaN raster data unaffected)
- [x] Edge cases covered (single-element axis, descending axis)
- [x] No premature materialization (coords already computed upstream)
- [x] Benchmark not needed; README matrix not applicable; docstrings accurate

brendancol (commented):
## Follow-up review (after 5c0cd237)

- Suggestion (error message): the message now spells out that duplicate and NaN coordinate values also fail the strict-monotonic check, so it reads accurately for those cases. proximity.py:278-283.
- Nit (test coverage): `test_single_element_axis_allowed` is now parametrized over cupy and dask+cupy, matching the other two new tests. test_proximity.py:1700.

No new findings. No blockers. Ready to merge once CI is green.

brendancol:
CI note: the `run (ubuntu-latest, 3.14)` fast-lane job fails on `test_polygonize.py::TestSimplifyHelpers::test_visvalingam_whyatt_scales_subquadratic`, a wall-clock timing assertion (`ratio < 3.0`) unrelated to this PR. It failed three runs in a row with ratios 3.47x, 3.43x, 3.46x, all just over the 3.0x threshold. The test lives on `main` and this branch touches only `xrspatial/proximity.py`, its tests, and the proximity reference page, so the failure is a pre-existing timing flake from slow runners, not a regression here. Every other check passes: `run (3.12)`, `label`, and the sibling `run (ubuntu-latest, 3.14)` workflow. The full `test_proximity.py` suite passes locally (311 tests).

PR #2876, opened 2026-06-02
Fix viewshed crash on NaN cells and mark them INVISIBLE (#2857)

### LINKED ISSUE #2857: viewshed crashes on NaN cell in observer row (ValueError: node not found)
**Describe the bug**

NaN/nodata handling in `viewshed` has a gap. `_init_event_list` (`xrspatial/viewshed.py` around line 1164) creates ENTERING/CENTER/EXITING events for NaN cells. A NaN cell on the observer's row, to the right of the observer, never gets inserted into the status structure because the pre-insert loop guards on `not np.isnan(data[1][i])`. But its EXITING event still fires during the sweep, so `_delete_from_tree` raises `ValueError: node not found`.

The test suite documents this today with an `xfail` at `xrspatial/tests/test_viewshed.py:742`.

Reproducer:

```python
import numpy as np, xarray as xa
from xrspatial.viewshed import viewshed

arr = np.full((5, 5), 2.0)
arr[2, 4] = np.nan        # observer row, to the right
raster = xa.DataArray(arr, coords=dict(x=np.arange(5.0), y=np.arange(5.0)),
                      dims=["y", "x"])
viewshed(raster, x=2.0, y=2.0, observer_elev=5)
# ValueError: node not found
```

There's a second problem even when the sweep doesn't crash. Downstream consumers treat `vs_data != INVISIBLE` as "visible" (`xrspatial/visibility.py:287` and `xrspatial/experimental/min_observable_height.py:171`). If a NaN cell produced a NaN output, `NaN != INVISIBLE` is True, so the cell would be counted as visible.

**Expected behavior**

NaN input cells shouldn't create events or crash the sweep. They should produce an output that downstream `!= INVISIBLE` checks don't count as visible. The natural sentinel is `INVISIBLE` (-1), which is already what the visibility grid is filled with.

**Additional context**

Pure bug fix, no public API change. The fix lives in the CPU sweep, which also backs the dask Tier B path.

brendancol:
Closes #2857

## What changed

- `_init_event_list` skips NaN cells so they no longer emit ENTERING/CENTER/EXITING events. This removes the `ValueError: node not found` that hit when a NaN sat on the observer row to the right of the observer (its EXITING event tried to delete a node that was never inserted).
- `_init_event_list` now returns the event count, and `_viewshed_cpu` trims the unused trailing rows of the pre-allocated event list before sorting. Without the trim, the leftover all-zero rows sort as CENTER events at cell (0, 0) and would mark it visible.
- NaN cells keep their `INVISIBLE` (-1) fill value. That is the sentinel downstream consumers already test for (`visibility.py` and `experimental/min_observable_height.py` both treat `!= INVISIBLE` as visible), so a NaN cell is never counted as visible.

## Backends

The fix lives in the CPU sweep, which backs the numpy backend and the dask Tier B path. The cupy/RTX viewshed is a separate implementation and is out of scope here.

## Test plan

- [x] Converted the `xfail` NaN test into passing tests.
- [x] NaN at every position around a 5x5 grid, including the observer row to the right.
- [x] Guard that skipped NaN cells don't leak visibility to cell (0, 0).
- [x] numpy and dask+numpy backends.
- [x] `test_viewshed.py`, `test_visibility.py`, `test_min_observable_height.py` all green.

No new public API, so the user-guide notebook and README feature-matrix steps are skipped. No docs change needed; the `viewshed` reference entry and signature are unchanged.

### DIFF
--- a/xrspatial/tests/test_viewshed.py
+++ b/xrspatial/tests/test_viewshed.py
@@ -717,13 +717,19 @@ def test_viewshed_metadata_preserved(backend, max_distance):
 # NaN / nodata input
 # -------------------------------------------------------------------
 
-@pytest.mark.parametrize("nan_pos", [(1, 1), (0, 0), (4, 4)])
-def test_viewshed_nan_input_supported_positions(nan_pos):
-    """A single NaN cell at these positions runs without error.
+# Every NaN position around the 5x5 grid, including the observer row to the
+# right of the observer (2, 4), which used to raise ValueError: node not found.
+_NAN_POSITIONS = [
+    (1, 1), (0, 0), (4, 4), (2, 4), (2, 0), (0, 2), (4, 2), (3, 4), (2, 3),
+]
 
-    The CPU sweep falls back to the centre elevation when bilinear
-    neighbours are NaN, so these cases complete.  See the companion
-    xfail test for the positions that still crash (issue #2693).
+
+@pytest.mark.parametrize("nan_pos", _NAN_POSITIONS)
+def test_viewshed_nan_input_supported_positions(nan_pos):
+    """A single NaN cell at any position runs without error (issue #2857).
+
+    NaN cells generate no sweep events, so the sweep never tries to delete an
+    uninserted node.  The NaN cell stays at its INVISIBLE fill value.
     """
     arr = np.full((5, 5), 2.0)
     arr[nan_pos] = np.nan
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Fix viewshed crash on NaN cells and mark them INVISIBLE (#2857)

None.

None.

- The new `if np.isnan(inrast[1][j]): continue` reads `inrast[1][j]`, the same value the line above stores into `e[E_ELEV_1]` (viewshed.py:1165). Reusing `e[E_ELEV_1]` for the guard would drop the second index, but reading `inrast[1][j]` directly next to the comment is clearer. Wash. Leave as-is.

### What looks good
- Root cause is right. NaN cells emitted EXITING events with no matching insert (the pre-insert loop at viewshed.py:1388 guards on `not np.isnan`), so `_delete_from_tree` raised. Skipping event creation for NaN cells removes the asymmetry.
- The trailing-row trim (`event_list = event_list[:count_event]`) is load-bearing, not cosmetic. Without it the leftover all-zero rows sort as CENTER events at (0, 0) and would mark that cell visible. `test_viewshed_nan_does_not_leak_visibility_to_origin` covers that path.
- Sentinel choice is correct. NaN cells stay at the INVISIBLE fill value, which is what visibility.py:287 and experimental/min_observable_height.py:171 test against. A NaN output would have read as visible; INVISIBLE does not.
- Verified two more edge cases locally: a NaN observer cell and an all-NaN raster both return cleanly (observer 180, rest INVISIBLE) with no crash.
- The dask Tier B path runs through `_viewshed_cpu`, so it picks up the fix. The dask+numpy test confirms it.
- The Tier C out-of-core distance sweep already skips NaN in `_sweep_ring` (`elev != elev`), so it was never affected and the fix correctly leaves it alone.

### Checklist
- [x] Algorithm matches reference (GRASS r.viewshed port; NaN cells are NODATA and excluded)
- [x] Implemented backends consistent (numpy + dask Tier B share the CPU sweep; cupy/RTX is a separate path, out of scope)
- [x] NaN handling correct
- [x] Dask chunk boundaries handled (Tier B computes to one in-memory array)
- [x] README feature matrix N/A (no new function)
- [x] Docstrings accurate (no API change)

PR #2877, opened 2026-06-02
Fix bounded-dask proximity crash on skinny rasters

### LINKED ISSUE #2854: Bounded Dask proximity crashes when halo depth exceeds an axis length
**Describe the bug**

Bounded Dask proximity crashes when the computed halo depth is larger than an axis. `_halo_depth()` turns `max_distance` into a pixel search radius using the densest coordinate spacing. On a skinny raster that radius can exceed the raster height or width. It then goes straight into `da.map_overlap(depth=(pad_y, pad_x))`, which rejects any depth larger than the array along that axis and raises `ValueError: The overlapping depth ... is larger than your array ...`.

Both Dask paths hit this: the dask+numpy path (`_process_dask`) and the dask+cupy path (`_process_dask_cupy`).

**Reproduce**

```python
from xrspatial.proximity import proximity

arr = np.zeros((3, 100), dtype=np.float64)
arr[1, 50] = 1.0
xs = np.linspace(0, 99, 100)
ys = np.linspace(0, 2, 3)
draster = xr.DataArray(da.from_array(arr, chunks=(3, 100)),
                       coords={'y': ys, 'x': xs}, dims=['y', 'x'])
proximity(draster, x='x', y='y', max_distance=10).data.compute()
# ValueError: The overlapping depth 10 is larger than your array 3.
```

Here the y axis is length 3 but the halo comes out at 10 pixels.

**Expected behavior**

A valid raster with a finite `max_distance` should not crash. Finite `max_distance` is what we point users at for scalable Dask, so it needs to work on skinny rasters and on great-circle rasters where the pixel pitch changes with latitude. Clamp the overlap depth per axis to something `map_overlap` accepts (no larger than the chunk size, which is `axis_length - 1` for a single chunk) and keep proximity, allocation, and direction results correct.

**Impact**

High. Finite `max_distance` is the recommended path for large Dask rasters, and right now it fails on legitimate input.

**Affected code**
- xrspatial/proximity.py:302 (`_halo_depth` returns the unclamped pixel radius)
- xrspatial/proximity.py:1391 (`_process_dask` passes the depth to `map_overlap`)
- xrspatial/proximity.py:589 (`_process_dask_cupy` passes the depth to `map_overlap`)

brendancol:
Closes #2854

### What

- `_halo_depth()` can return a pixel search radius larger than the raster height or width. That depth went straight into `da.map_overlap`, which rejects a depth larger than the array along an axis and raised `ValueError: The overlapping depth ... is larger than your array ...`. Reproduced on a 3x100 raster with one target and `max_distance=10`.
- Add `_fit_halo_to_chunks()`: when the halo is deeper than the smallest chunk on an axis, fold that axis into a single chunk and set its depth to 0. Every chunk then sees the full axis, so no target within `max_distance` is dropped and the result still matches the numpy backend. Applied to both the dask+numpy and dask+cupy `map_overlap` call sites.

### Backend coverage

Fix touches the dask+numpy and dask+cupy paths. numpy and cupy (non-dask) paths are unaffected. The normal (non-skinny) dask path keeps its existing chunking, so there is no extra materialisation in the common case.

### Test plan

- [x] New regression test `test_bounded_dask_skinny_raster_matches_numpy` covers the 3x100 skinny raster across proximity, allocation, and direction, pinned against the numpy backend.
- [x] Full `test_proximity.py` passes (282 tests).
- [x] Verified manually: multi-chunk skinny raster and a great-circle skinny raster near the pole also match numpy.

### DIFF
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -314,7 +314,10 @@ def _fit_halo_to_chunks(pad_y, pad_x, *arrays):
     a halo is too deep for the chunking, fold that whole axis into a single
     chunk and drop its depth to zero: every chunk then sees the full axis, so
     no target within ``max_distance`` is missed and the result still matches
-    the NumPy backend.
+    the NumPy backend. The fold deliberately trades chunking on that axis for
+    correctness; clamping the depth while keeping multiple chunks would
+    silently drop targets that fall in a non-adjacent chunk, so do not replace
+    it with a bare depth clamp.
 
     ``arrays`` are the dask arrays passed to the same ``map_overlap`` call
     (the raster and the coordinate grids); they all share the raster's
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -304,6 +304,40 @@ def _halo_depth(x_coords, y_coords, max_distance, distance_metric):
     return pad_y, pad_x
 
 
+def _fit_halo_to_chunks(pad_y, pad_x, *arrays):
+    """Make a bounded halo depth that ``da.map_overlap`` will accept.
+
+    ``map_overlap`` rejects a depth that is larger than the smallest chunk
+    along that axis. The pixel halo from ``_halo_depth`` can exceed that on
+    skinny rasters (e.g. a 3-row raster with a 10-pixel halo), or on
+    great-circle rasters where the pixel pitch shrinks toward the poles. When
+    a halo is too deep for the chunking, fold that whole axis into a single
+    chunk and drop its depth to zero: every chunk then sees the full axis, so
+    no target within ``max_distance`` is missed and the result still matches
+    the NumPy backend.
+
+    ``arrays`` are the dask arrays passed to the same ``map_overlap`` call
+    (the raster and the coordinate grids); they all share the raster's
+    chunking and are rechunked together so the call stays aligned.
+
+    Returns the adjusted ``(pad_y, pad_x)`` and the (possibly rechunked)
+    arrays in the same order they were given.
+    """
+    arrays = list(arrays)
+    height, width = arrays[0].shape
+    chunks_y, chunks_x = arrays[0].chunks
+    rechunk = {}
+    if pad_y > min(chunks_y):
+        rechunk[0] = height
+        pad_y = 0
+    if pad_x > min(chunks_x):
+        rechunk[1] = width
+        pad_x = 0
+    if rechunk:
+        arrays = [a.rechunk(rechunk) for a in arrays]
+    return pad_y, pad_x, arrays
+
+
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Fix bounded-dask proximity crash on skinny rasters

None.

None.

- xrspatial/proximity.py:307 `_fit_halo_to_chunks` rechunks an over-deep axis to a single chunk instead of just clamping the depth. That is the correct call: clamping the depth while keeping multiple chunks would silently drop targets that live in non-adjacent chunks. The cost is that the folded axis is fully materialised per task, which is cheap for the skinny axis that triggers it. Worth a one-line docstring note that the fold trades chunking on that axis for correctness, so a later reader does not "optimise" it back to a bare depth clamp.

### What looks good
- The fix lives at the two map_overlap call sites (dask+numpy at proximity.py:1431, dask+cupy at proximity.py:614) and leaves `_halo_depth` alone. The depth math is unchanged; only the chunking moves, and only when it has to.
- Folding the over-deep axis to a single chunk and zeroing its depth is correct. Every chunk then sees the whole axis, so no target within max_distance is lost. Checked against numpy for the skinny case, a multi-chunk skinny case, and a great-circle skinny case near the pole.
- The common non-skinny path is untouched. A halo smaller than the chunk leaves the chunking alone, so there is no extra materialisation in the normal case.
- The numpy path reassigns `raster.data`, which matches the unbounded branch right above it. The cupy path uses a local `raster_data` to avoid mutating the cupy-backed input. The asymmetry is deliberate.
- Regression test covers proximity, allocation, and direction on the 3x100 raster, pinned against numpy with equal_nan.

### Checklist
- Algorithm correctness: depth math unchanged; chunking fold is correct.
- Backend parity: dask+numpy and dask+cupy both fixed; numpy/cupy paths untouched. The dask+cupy fold is not exercised in CI (no GPU) but mirrors the numpy path.
- NaN handling: unchanged; boundary=np.nan still used.
- Edge cases: skinny, multi-chunk skinny, and great-circle skinny near the pole all checked.
- Dask chunk boundaries: handled by the rechunk-to-single-chunk fold.
- No premature materialisation in the common path.
- Docstring present on the new helper.
- README/benchmark: not applicable (pure bug fix, no new public API).

brendancol (commented):
## Follow-up review (after 9f2c39ef)

The single nit from the first pass is addressed: `_fit_halo_to_chunks` now carries a docstring note that the single-chunk fold is deliberate and must not be swapped for a bare depth clamp (which would drop targets in non-adjacent chunks).

No new findings. No blockers or suggestions outstanding. The skinny-raster regression test still passes and flake8 is clean on the changed file.

PR #2878, opened 2026-06-02
Gate remote sources in read_geotiff_gpu under stable_only (#2867)

### LINKED ISSUE #2867: read_geotiff_gpu(..., stable_only=True) does not gate remote sources
## Description

`read_geotiff_gpu(..., stable_only=True)` accepts the `stable_only` kwarg but never enforces it. The function runs `_validate_dispatch_kwargs` for the generic dispatcher checks, but it skips `_validate_stable_only_remote`, the gate that rejects advanced-tier HTTP / fsspec sources when a caller asks for stable-tier sources only.

Both sibling entry points already apply this gate:

- `open_geotiff` calls `_validate_stable_only_remote(source, stable_only=..., allow_experimental_codecs=...)` before dispatch (`xrspatial/geotiff/__init__.py`).
- `read_geotiff_dask` calls the same helper for its own direct entry point (`xrspatial/geotiff/_backends/dask.py`).

`read_geotiff_gpu` is also a public direct entry point, but it skips the gate. For HTTP / fsspec sources the GPU path routes through a CPU fallback, so a direct GPU call goes ahead and reads a remote source that `open_geotiff` and `read_geotiff_dask` reject. The GPU docstring makes it worse by calling `stable_only` a no-op on the GPU path, which is wrong for remote sources.

## Impact

Backend-level API behavior diverges. A downstream caller using the GPU read API directly does not get the stability guarantee they asked for with `stable_only=True`. They can read an advanced-tier remote source that the other two readers turn away.

## Expected behavior

`read_geotiff_gpu(remote_source, stable_only=True)` should raise `RemoteStableSourcesOnlyError`, matching `open_geotiff` and `read_geotiff_dask`. Passing `allow_experimental_codecs=True` should unlock it the same way.

## Coverage gap

There is no test for `read_geotiff_gpu(remote, stable_only=True)`. The existing regression suite for this gate (`test_stable_only_remote_2821.py`) covers the eager and dask paths but not the GPU entry point.

## Fix

Call `_validate_stable_only_remote(...)` in `read_geotiff_gpu` before the cupy import, the CUDA preflight, and the remote CPU fallback, the same way `open_geotiff` and `read_geotiff_dask` call it. Update the GPU docstring so `stable_only` is no longer described as a no-op.

brendancol:
Closes #2867

`read_geotiff_gpu(..., stable_only=True)` accepted the kwarg but never enforced it. For HTTP / fsspec sources the GPU path routes through the CPU fallback, so a direct GPU call could read an advanced-tier remote source that `open_geotiff` and `read_geotiff_dask` reject. The GPU docstring also called `stable_only` a no-op, which was wrong for remote sources.

- Call `_validate_stable_only_remote(...)` in `read_geotiff_gpu` before the cupy import, the CUDA preflight, and the remote CPU fallback, the same way `open_geotiff` and `read_geotiff_dask` call it. Running it ahead of the import means the rejection is independent of GPU availability.
- Fix the GPU docstring so `stable_only` is described as an enforced gate, not a no-op.

Backend coverage: the gate lives in the shared `read_geotiff_gpu` entry point, which serves the eager GPU (numpy/cupy) and chunked GPU (dask+cupy) reads. The other backends already had the gate.

Test plan:
- [x] Added GPU coverage to `test_stable_only_remote_2821.py`: fsspec and http sources rejected under `stable_only=True`, plus an experimental-optin unlock test gated behind cupy + CUDA.
- [x] Rejection tests pass on a CPU-only host (the gate fires before the cupy import).
- [x] `pytest xrspatial/geotiff/tests/test_stable_only_remote_2821.py` passes (12 passed locally with a GPU present).

### DIFF
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -210,10 +210,13 @@ def read_geotiff_gpu(source: str, *,
     stable_only : bool, default False
         [experimental] Read-side opt-in for stable-tier sources only.
         The GPU read path does not consume VRT sources directly (VRT
-        routing happens in ``open_geotiff``), so this kwarg is accepted
-        for cross-backend signature symmetry and is a no-op on the GPU
-        eager / chunked paths. See ``open_geotiff`` for the full
-        description.
+        routing happens in ``open_geotiff``), but it does route HTTP /
+        fsspec sources through the CPU fallback, and those advanced-tier
+        readers must be gated. With ``stable_only=True`` a remote source
+        raises ``RemoteStableSourcesOnlyError`` before any cupy import or
+        decode, matching ``open_geotiff`` and ``read_geotiff_dask``. Pass
+        ``allow_experimental_codecs=True`` to unlock the advanced tier.
+        See ``open_geotiff`` for the full description.
     allow_experimental_codecs : bool, default False
         [experimental] Read-side opt-in for Tier 3 experimental codecs
         (``lerc``, ``jpeg2000`` / ``j2k``, ``lz4``). The GPU read path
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Gate remote sources in read_geotiff_gpu under stable_only (#2867)

None.

None.

- The unlock test (`test_gpu_direct_fsspec_source_allowed_with_experimental_optin`) re-declares the `_gpu_available` / `_HAS_GPU` helper that a few other geotiff test modules already define locally (`read/test_basic.py`, `unit/test_predictor.py`). There is no shared fixture for it, so the local copy matches the existing pattern in this package. Fine to leave as is.

### What looks good
- The gate runs before the cupy import, the CUDA preflight, and the remote CPU fallback, so the rejection does not depend on GPU availability. That is what lets the two rejection tests run on CPU-only CI, which is the right call.
- The call signature matches how `open_geotiff` and `read_geotiff_dask` invoke `_validate_stable_only_remote` (same kwargs, same local-import style).
- The gate sits ahead of the `chunks is not None` branch, so it covers both the eager and chunked GPU paths, not just one.
- The docstring no longer claims `stable_only` is a no-op. The wording now matches actual behavior and the `geotiff.rst` note that already said all three paths run the check.
- Tests cover fsspec and http rejection plus the experimental-codecs unlock, mirroring the existing eager/dask cases.

### Checklist
- [x] Behavior matches the sibling readers (open_geotiff, read_geotiff_dask)
- [x] Gate fires before any GPU/network work
- [x] Covers eager and chunked GPU paths
- [x] Edge cases covered by tests (fsspec, http, unlock)
- [x] README feature matrix not affected (no new function, no backend change)

PR #2879, opened 2026-06-02
Validate MFD companion raster shape against grid (#2863)

### LINKED ISSUE #2863: MFD hydrology functions don't validate companion raster shape, causing OOB reads
**Describe the bug**

The MFD hydrology functions validate that secondary/companion rasters are 2-D DataArrays, but they never check that the companion raster's shape matches the primary MFD grid shape (H, W). Pass a mismatched companion raster and the CPU kernel reads out of bounds and returns uninitialized garbage instead of raising.

Reproduced with an (8, 3, 3) MFD flow-direction grid and a (1, 1) start_points raster:

```python
from xrspatial.hydro import flow_path_mfd
fr = xr.DataArray(np.zeros((8, 3, 3)), dims=('d', 'y', 'x'))
sp = xr.DataArray(np.array([[5.0]]), dims=('y', 'x'))
out = flow_path_mfd(fr, sp)  # no error; returns garbage
```

The same missing-shape-check shows up in:

- `flow_path_mfd` (start_points)
- `watershed_mfd` (pour_points)
- `hand_mfd` (flow_accum, elevation)
- `stream_order_mfd` (flow_accum)
- `stream_link_mfd` (flow_accum)

**Expected behavior**

Each function should check that the companion raster shape matches the primary grid (H, W) and raise a clear ValueError before any kernel indexing happens.

**Additional context**

Follow-up to #1425, which added `_validate_raster` on the secondary args but only checked type and ndim, not shape.

brendancol:
Closes #2863

The MFD hydrology functions checked that their companion rasters were 2-D DataArrays but never confirmed the companion's (H, W) matched the primary MFD grid. A mismatched companion let the CPU kernel read out of bounds and return uninitialized garbage instead of raising.

- Added a `_validate_matching_shape` helper in `xrspatial/utils.py` that raises a clear ValueError when a companion raster's shape does not match the primary grid.
- Wired it into the five affected MFD functions: `flow_path_mfd` (start_points), `watershed_mfd` (pour_points), `hand_mfd` (flow_accum, elevation), `stream_order_mfd` (flow_accum), `stream_link_mfd` (flow_accum). The check runs after the existing 3-D shape validation and before any kernel indexing.

The check lives in the shared public-API entry point, so it guards all four backends (numpy, cupy, dask+numpy, dask+cupy) before dispatch.

Test plan:
- [x] New `test_validate_mfd_companion_shape.py`: each function raises ValueError on a mismatched companion and still runs on a matching one (11 tests).
- [x] Existing MFD suites and the secondary-args validation suite pass (96 tests).
- [x] flake8 clean on changed lines.

### DIFF
--- a/xrspatial/hydro/flow_path_mfd.py
+++ b/xrspatial/hydro/flow_path_mfd.py
@@ -29,6 +29,7 @@ except ImportError:
     da = None
 
 from xrspatial.utils import (
+    _validate_matching_shape,
     _validate_raster,
     has_cuda_and_cupy,
     is_cupy_array,
@@ -429,6 +430,10 @@ def flow_path_mfd(flow_dir_mfd: xr.DataArray,
             f"flow_dir_mfd must have shape (8, H, W), got {data.shape}")
 
     _, H, W = data.shape
+
+    _validate_matching_shape(
+        start_points, (H, W), func_name='flow_path_mfd',
+        name='start_points', expected_name='flow_dir_mfd')
 
     if isinstance(data, np.ndarray):
         _check_memory(H, W)
--- a/xrspatial/hydro/hand_mfd.py
+++ b/xrspatial/hydro/hand_mfd.py
@@ -27,6 +27,7 @@ from xrspatial.hydro.watershed_mfd import (
     _to_numpy_f64,
 )
 from xrspatial.utils import (
+    _validate_matching_shape,
     _validate_raster,
     has_cuda_and_cupy,
     is_cupy_array,
@@ -709,6 +710,13 @@ def hand_mfd(flow_dir_mfd: xr.DataArray,
             f"flow_dir_mfd must have shape (8, H, W), got {data.shape}")
 
     _, H, W = data.shape
+
+    _validate_matching_shape(
+        flow_accum, (H, W), func_name='hand_mfd',
+        name='flow_accum', expected_name='flow_dir_mfd')
+    _validate_matching_shape(
+        elevation, (H, W), func_name='hand_mfd',
+        name='elevation', expected_name='flow_dir_mfd')
 
     if isinstance(data, np.ndarray):
         _check_memory(H, W)
--- a/xrspatial/hydro/stream_link_mfd.py
+++ b/xrspatial/hydro/stream_link_mfd.py
@@ -40,6 +40,7 @@ except ImportError:
     da = None
 
 from xrspatial.utils import (
+    _validate_matching_shape,
     _validate_raster,
     cuda_args,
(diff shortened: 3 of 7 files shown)


brendancol (commented):
## PR Review: Validate MFD companion raster shape against grid (#2863)

None.

- The check compares full (H, W) equality but does not verify chunk alignment for dask inputs. A correctly-shaped but differently-chunked companion (e.g. flow_accum chunked (1,4) while fractions is chunked (2,2)) still reaches the kernel. This PR's scope is shape validation, so chunk-alignment is reasonable to leave for a follow-up, but worth flagging.

- The matching-input tests assert only out.shape. A known-value assertion would be stronger, but those cases exist mainly to prove the guard does not false-positive, so shape is acceptable.
- The `_validate_matching_shape` default `expected_name='the primary raster'` is never used, since every call site passes an explicit name. Harmless.

### What looks good
- The check sits in the shared public entry point after the existing 3-D validation and before any kernel indexing, so all four backends are guarded before dispatch.
- For dask inputs it reads `.shape` only, which is known without computing, so no premature materialization.
- The error message names both the offending parameter and the reference grid.
- Tests hit all five functions with a mismatched (raises) and a matching (runs) case each.

### Checklist
- All implemented backends produce consistent results: yes, guard runs pre-dispatch for all four
- Edge cases covered by tests: mismatched and matching shapes per function
- Dask chunk boundaries handled correctly: shape guard is lazy; chunk-alignment not checked (see suggestion)
- No premature materialization: confirmed, `.shape` only
- Benchmark exists or not needed: not needed (validation guard)
- README feature matrix updated: n/a (no new public function)
- Docstrings present and accurate: yes, helper documented

brendancol (commented):
## Follow-up review (after review fixes)

Re-reviewed after the follow-up commit. Disposition of the first-pass findings:

- Nit (matching tests asserted only shape): fixed. The `flow_path_mfd` matching test now asserts the traced path values (all-east flow labels the top row 7.0, the rest stay NaN). The other matching cases keep the shape assertion since their purpose is to prove the guard does not false-positive on valid input, and the dedicated MFD suites already cover output correctness.
- Suggestion (dask chunk alignment not checked): deferred. Chunk-alignment validation is out of scope for this shape-only fix and overlaps sibling MFD work; folding it in here would widen the diff and risk merge conflicts.
- Nit (`_validate_matching_shape` default `expected_name` unused): dismissed. The default is a sensible self-documenting fallback for the reusable helper, matching the `name='raster'` default on the sibling `_validate_raster`.

No new findings. Shape guard and tests look correct.

PR #2880, opened 2026-06-02
viewshed: warn and document dask Tier C divergence from exact sweep

### LINKED ISSUE #2872: viewshed dask Tier C fallback diverges from exact sweep but is documented as minor
**Describe the bug**

The dask "Tier C" out-of-core fallback in `viewshed()` produces visibility masks that diverge substantially from the exact numpy sweep, but the docstring at `xrspatial/viewshed.py:1670` only describes "minor visibility differences near the boundary of occluded regions." On random terrain the divergence is far larger than that, and there is no runtime warning. Large dask rasters can feed downstream visibility logic with silently wrong results.

Tier C (starting at `xrspatial/viewshed.py:2227`) is a horizon-profile distance sweep that discretizes occlusion into angular bins. It is a different, approximate visibility model from the exact GRASS r.viewshed event sweep used by the CPU/Tier-B path. The error is geometric, not a resolution artifact: raising the angular bin count does not reduce the mismatch count.

**Reproduce**

Force Tier C on a small random terrain (call `_viewshed_distance_sweep` directly, or shrink the memory budget so `_viewshed_dask` skips Tier B) and diff the visibility mask against `_viewshed_cpu`.

Measured mask mismatches vs the exact sweep (observer at grid center, integer-coordinate grid, no observer/target elevation offset):

- 20x20, seed 0: 81 / 400 cells (20.2%)
- 20x20, seed 42: 25 / 400 cells (6.2%)
- 40x40, seed 42: 199 / 1600 cells (12.4%)

Mismatches include both false positives and false negatives. Raising `n_angles` from 360 to 500000 does not change the count.

**Expected behavior**

Either Tier C matches the exact sweep, or the documentation states the real accuracy contract and the code emits a runtime warning when the Tier C path runs, so users do not silently receive approximate visibility from large dask inputs.

**Additional context**

Reimplementing the exact event sweep out-of-core is a large undertaking, and it is the reason Tier C exists: the exact algorithm's working set does not fit when the grid exceeds memory. The horizon-profile model cannot be made exact without becoming the exact algorithm.

Affected: dask+numpy and dask+cupy backends on grids large enough to skip Tier B and without `max_distance` set.

brendancol:
Closes #2872

## What

The dask out-of-core "Tier C" fallback in `viewshed()` (`_viewshed_dask`, distance-sweep path) is a horizon-profile distance sweep. It is a different, approximate visibility model from the exact GRASS r.viewshed event sweep used by the numpy backend and the dask Tier-B (in-memory compute) path. The docstring described it as producing only "minor visibility differences near the boundary of occluded regions," which understates the problem: on rough terrain the visibility mask can differ for a substantial fraction of cells, with both false positives and false negatives.

I investigated whether Tier C could be made exact. It cannot, short of reimplementing the exact event sweep out-of-core. The divergence is geometric, not an angle-resolution artifact: raising the bin count from 360 to 500000 leaves the mismatch count unchanged. Making it exact would mean streaming the GRASS red-black status structure across chunks, which is a major new subsystem and defeats the reason Tier C exists (the exact algorithm's working set does not fit when the grid exceeds memory).

So this PR takes option (b) from the issue: make the contract honest and stop the silent divergence.

## Changes

- Emit a `UserWarning` when the Tier C path runs, so large dask rasters do not silently feed approximate visibility into downstream logic.
- Correct the `viewshed` docstring and the dask laziness reference table to state the real accuracy contract and point users to `max_distance` for exact results on large grids.
- Add a test that asserts the warning fires and quantifies the mask divergence vs the exact numpy sweep (pins the documented behavior; if Tier C is ever made exact, the test flags that it should switch to a parity assertion).
- Add a `filterwarnings` ignore for the new warning so the two existing Tier C tests stay quiet.

## Backend coverage

Affects the dask backends only (dask+numpy and dask+cupy take the same Tier C path; dask+cupy converts chunks to numpy before the sweep). numpy and cupy backends are unchanged. No new public API and no change to backend support, so the README matrix and user-guide notebook are unchanged.

## Test plan

- [x] `pytest xrspatial/tests/test_viewshed.py` (64 passed, 1 xfailed)
- [x] New test asserts the `UserWarning` and a material (>= 10 cell) mask divergence on a 20x20 random terrain
- [x] flake8 clean on the touched files

### DIFF
--- a/docs/source/reference/dask_laziness.rst
+++ b/docs/source/reference/dask_laziness.rst
@@ -154,7 +154,10 @@ Visibility
      - Notes
    * - ``viewshed``
      - Fully materialized
-     - Sweepline algorithm needs random access
+     - Sweepline needs random access. Grids that exceed memory (no
+       ``max_distance``) use an approximate out-of-core sweep that does
+       not match the exact result and emits a ``UserWarning``; set
+       ``max_distance`` for exact results.
    * - ``line_of_sight``
      - Fully materialized
      - Extracts 1D transect via ``.compute()``
--- a/setup.cfg
+++ b/setup.cfg
@@ -113,6 +113,7 @@ filterwarnings =
     ignore:cost_distance. max_cost is infinite:UserWarning:xrspatial
     ignore:surface_distance. max_distance is infinite:UserWarning:xrspatial
     ignore:proximity. target coordinates exceed:ResourceWarning:xrspatial
+    ignore:viewshed. grid exceeds memory:UserWarning:xrspatial
     ignore:Covariance of the parameters could not be estimated:scipy.optimize.OptimizeWarning
     ignore:'oneOf' deprecated:DeprecationWarning:matplotlib
     ignore:'parseString' deprecated:DeprecationWarning:matplotlib
--- a/xrspatial/tests/test_viewshed.py
+++ b/xrspatial/tests/test_viewshed.py
@@ -420,6 +420,63 @@ def test_viewshed_dask_distance_sweep_target_elev():
             err_msg="Tier C angles diverge too far from numpy reference")
 
 
+def test_viewshed_dask_tier_c_warns_and_quantifies_divergence():
+    """Tier C (out-of-core distance sweep) is an approximate visibility
+    model that does not match the exact numpy sweep cell-for-cell.
+
+    Pins the documented contract for issue #2872:
+    1. A UserWarning is emitted when the Tier C path runs, so users feeding
+       large dask rasters into downstream visibility logic are not silently
+       misled.
+    2. The visibility-mask divergence vs the exact sweep is real and
+       material on rough terrain (not "minor near occluded boundaries").
+    """
+    from unittest.mock import patch
+
+    ny, nx = 20, 20
+    rng = np.random.RandomState(42)
+    terrain = rng.rand(ny, nx) * 10.0
+    xs = np.arange(nx, dtype=float)
+    ys = np.arange(ny, dtype=float)
+    obs_x, obs_y = 10.0, 10.0
+
+    # Exact numpy reference.
+    raster_np = xa.DataArray(terrain.copy(), coords=dict(x=xs, y=ys),
+                             dims=["y", "x"])
+    v_np = viewshed(raster_np, x=obs_x, y=obs_y)
+
+    # Force Tier C: shrink the memory budget so Tier B (full compute + R2)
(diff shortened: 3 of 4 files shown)


brendancol (commented):
## PR Review: viewshed: warn and document dask Tier C divergence from exact sweep

None.

- `cumulative_viewshed` (visibility.py:281) calls `viewshed` once per observer, so when the Tier C path runs, each call hits the new UserWarning. Python's default filter dedupes by (message, category, module, lineno), so it fires once per process rather than spamming. Worth a one-line comment in the test noting the dedupe so a future reader doesn't worry about it. Not blocking.

- The test threshold `mismatches >= 10` is a magic number. It works as a floor (the measured value on seed 42 is 25), but a short comment stating the observed count would make a future regression easier to read. The docstring already covers the intent, so this is minor.

### What looks good
- Right call to document + warn instead of attempting an exact out-of-core rewrite. The PR body justifies it and the n_angles invariance backs it up.
- The UserWarning stacklevel=3 resolves to the user's `viewshed()` call.
- The new test pins both the warning and a material divergence, and leaves a clear note to switch to a parity assertion if Tier C is ever made exact.
- The filterwarnings ignore matches the existing proximity/cost_distance pattern, so the older Tier C tests stay quiet while `pytest.warns` still catches it in the new test.
- Docstring and dask laziness table now state the real contract and point to max_distance.

### Checklist
- Algorithm matches reference: n/a (documents an approximation, does not change it)
- All implemented backends consistent: dask+numpy and dask+cupy share the Tier C path; numpy/cupy unchanged
- NaN handling correct: unchanged by this PR
- Edge cases covered: warning + divergence pinned on random terrain
- No premature materialization: unchanged
- Benchmark: not needed (no perf change)
- README matrix: not needed (no API/backend change)
- Docstrings present and accurate: yes, corrected

brendancol (commented):
## Follow-up review (after comment-only fixes)

Both items from the first review are addressed:

- The Tier C test now states the observed mismatch count (25 of 400) next to the `>= 10` floor, so the threshold reads as a deliberate floor rather than a bare magic number.
- A comment in the test explains that `cumulative_viewshed` calls `viewshed` per observer and that the warning dedupes under Python's default filter, so it surfaces once per process.

No new findings. The change stays scoped to documentation, a runtime warning, and one test. `pytest xrspatial/tests/test_viewshed.py` passes (64 passed, 1 xfailed) and flake8 is clean on the touched Python files.

PR #2881, opened 2026-06-02
Make allocation/direction tie-breaking deterministic across backends

### LINKED ISSUE #2856: Allocation/direction tie-breaking differs across backends
**Describe the bug**

`allocation` and `direction` can return different categorical results across backends when two or more targets are equidistant from a pixel. The distance itself is fine (ties have equal distance by definition), but the target that gets picked, and therefore the allocated zone value and the direction angle, is not stable across backends.

Two paths disagree:

- The Dask unbounded path queries a SciPy `cKDTree` (proximity.py:754, proximity.py:1244). `cKDTree.query` does not promise which of several equidistant neighbours it returns, and the order in which target coordinates get collected depends on chunk iteration.
- The NumPy line-sweep (`_process_numpy_linesweep`) breaks ties as a side effect of its four-pass propagation order, which does not match the brute-force or CUDA backends.

The exact brute-force CPU path (`_process_numpy_bruteforce`) and the CUDA kernel (`_proximity_cuda_kernel`) already agree with each other: on a tie, the target with the lowest flat (row-major) index wins, because both scan targets in row-major order and keep the first one under a strict `<` comparison.

**Reproduce**

3x3 raster, targets at (1,0)=1 and (1,2)=2. The top-center pixel (0,1) is the same distance from both.

- NumPy `allocation` gives the top-center pixel 2.
- Dask `allocation` gives it 1.
- Brute-force and CUDA give it 1 (lowest flat index).

**Expected behavior**

All four backends (numpy, cupy, dask+numpy, dask+cupy) should break ties the same way, with one documented rule: on equal distance, the target with the lowest flat (row-major) index wins. That is what the brute-force and CUDA paths already do. Document the rule in the `allocation` and `direction` docstrings and pin it with cross-backend tests.

**Additional context**

Refs: xrspatial/proximity.py:754, xrspatial/proximity.py:1244

brendancol:
Closes #2856

When two or more targets are equidistant from a pixel, `allocation` and `direction` could return different results depending on the backend. The distance is the same for tied targets, but the chosen target was not stable. This pins one policy everywhere: on a tie, the target with the lowest flat (row-major) index wins. That is what the brute-force CPU path and the CUDA kernel already did.

## Changes

- ALLOCATION and DIRECTION now run through the exact brute-force search on the numpy and dask+numpy backends. The line-sweep broke ties as a side effect of its four-pass propagation order. PROXIMITY keeps the faster line-sweep, since the distance is identical for tied targets.
- The Dask KDTree query is pinned to the lowest target index on a tie. `cKDTree.query` does not promise which of several equidistant neighbours it returns, so a helper queries the two nearest and keeps the smaller index when they tie. Targets are stored in row-major order, so the smaller index is the lowest flat index. This covers both the global (lazy) and tiled (eager fallback) KDTree paths.
- The policy is documented in the `allocation` and `direction` docstrings.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy all resolve ties the same way. The unbounded dask+cupy path converts to dask+numpy and uses the KDTree; the bounded dask+cupy path uses the CUDA kernel, which already followed the policy.

## Test plan

- [x] New cross-backend tests pin the tie outcome for `allocation` (exact expected grid) and `direction` (pinned to the numpy reference) over all four backends.
- [x] New test forces the eager tiled KDTree fallback and checks the tie outcome.
- [x] Full `test_proximity.py` suite passes (288 tests).

### DIFF
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -745,7 +745,10 @@ def _kdtree_query_lowest_index(tree, query_pts, p, max_distance):
     the tie-break policy documented on ``allocation``/``direction``.
 
     Query the two nearest targets; wherever they are equidistant, keep the one
-    with the smaller index.
+    with the smaller index. This resolves 2-way ties, which is what grid
+    geometry produces in practice. A pixel equidistant to three or more targets
+    relies on cKDTree returning the lower index among the rest, which it does
+    for the row-major target order used here but does not strictly promise.
     """
     n_targets = tree.n
     if n_targets < 2:
@@ -1409,8 +1412,10 @@ def _process(
         # four-pass propagation order, which disagrees with the brute-force
         # and CUDA kernels. Route those modes through the brute-force search,
         # which keeps the lowest-flat-index target on a tie (see the
-        # `allocation`/`direction` docstrings). PROXIMITY only returns the
-        # distance, which is identical for tied targets, so the faster
+        # `allocation`/`direction` docstrings). Brute force is O(N*T) versus
+        # the line-sweep's O(N); the slower scan is a deliberate trade for a
+        # tie-break that matches every other backend. PROXIMITY only returns
+        # the distance, which is identical for tied targets, so the faster
         # line-sweep stays in use there.
         if process_mode in (ALLOCATION, DIRECTION):
             return _process_numpy_bruteforce(
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -735,6 +735,35 @@ def _process_proximity_line(
     return
 
 
+def _kdtree_query_lowest_index(tree, query_pts, p, max_distance):
+    """Nearest-target query that breaks ties by lowest target index.
+
+    ``cKDTree.query`` does not promise which of several equidistant targets
+    it returns, so allocation and direction can disagree with the brute-force
+    and CUDA backends on a tie. Target coordinates are stored in row-major
+    (flat-index) order, so the lowest target index is the lowest flat index --
+    the tie-break policy documented on ``allocation``/``direction``.
+
+    Query the two nearest targets; wherever they are equidistant, keep the one
+    with the smaller index.
+    """
+    n_targets = tree.n
+    if n_targets < 2:
+        return tree.query(query_pts, p=p, distance_upper_bound=max_distance)
+
+    dists2, idx2 = tree.query(query_pts, k=2, p=p,
+                              distance_upper_bound=max_distance)
+    dists = dists2[:, 0]
+    indices = idx2[:, 0]
+    # A tie exists where both neighbours are finite and equidistant. Prefer the
+    # smaller index in that case so the result is independent of cKDTree's
+    # internal traversal order.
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Make allocation/direction tie-breaking deterministic across backends

None.

- Performance tradeoff on the unbounded numpy/dask+numpy path: `_process_numpy` now sends ALLOCATION and DIRECTION through `_process_numpy_bruteforce`, which is O(N*T): every pixel scans every target. The line-sweep it replaces is roughly O(N). For a raster with many target pixels and no `max_distance` bound, that is a real slowdown. It buys a deterministic cross-backend tie-break and matches what the CUDA kernel already does, so the tradeoff is reasonable, but a one-line code comment would tell the next reader the cost was deliberate. (xrspatial/proximity.py, `_process_numpy`)

- `_kdtree_query_lowest_index` only resolves 2-way ties, since it queries k=2. A pixel equidistant to 3 or more targets could still land on a non-minimal index if cKDTree returns the 2nd and 3rd nearest as the top two. The global KDTree path stores targets in row-major order and cKDTree tends to return the lower index first, so this is unlikely to bite in practice, but a short comment noting the k=2 limitation would set expectations. (xrspatial/proximity.py, `_kdtree_query_lowest_index`)

### What looks good
- The tie-break policy (lowest flat row-major index) matches the brute-force and CUDA backends that already existed, so this aligns the two stragglers rather than inventing a new rule.
- Both KDTree entry points (global lazy `_kdtree_chunk_fn` and eager tiled `_tiled_chunk_query`) go through the same helper, so the fallback path is covered.
- The `np.isfinite(dists2[:, 1])` guard avoids treating an out-of-range second neighbour (inf distance under `distance_upper_bound`) as a tie.
- Tests pin the actual tie outcome across all four backends for both allocation and direction, plus a dedicated test for the tiled fallback. The direction test pins to the numpy reference rather than a hand-written grid, which is reasonable for an angle.
- Docstrings on both public functions now state the policy.

### Checklist
- [x] Algorithm matches reference/policy (lowest flat index, matches brute-force and CUDA)
- [x] NaN / out-of-range handling is correct (isfinite guard on the tie mask)
- [x] Edge cases covered by tests (tie on a 3x3, tiled fallback)
- [x] Dask chunk boundaries handled correctly (targets collected in row-major order)
- [ ] Performance: brute-force on unbounded numpy alloc/direction is slower than the line-sweep; deliberate tradeoff for determinism

brendancol (commented):
## Follow-up review

Both items from the first pass are addressed in 8ddf938 (comments only, no behavior change):

- The brute-force performance trade is now spelled out in `_process_numpy`: O(N*T) versus the line-sweep's O(N), chosen so allocation and direction break ties the same way on every backend.
- `_kdtree_query_lowest_index` now states that the k=2 query covers 2-way ties and leans on the row-major target order for the rarer 3-or-more-way case.

No new findings. Behavior is unchanged from the first pass; the proximity suite still passes (288 tests). No blockers.

PR #2882, opened 2026-06-02
Gate stable_only=True before bbox metadata read on remote/VRT sources

### LINKED ISSUE #2869: stable_only=True enforced after bbox metadata read on remote/VRT sources
**Describe the bug**

`open_geotiff(..., bbox=...)` resolves the bounding box to a pixel window before it runs the `stable_only=True` rejection for remote and VRT sources. The bbox resolver reads source geo metadata first. The TIFF path reads a header (a range GET for an HTTP/fsspec source); the VRT path parses the VRT XML. Both stable-only gates only fire after that, so a `stable_only=True` request can trigger remote metadata I/O or a VRT parse before it gets refused.

In `xrspatial/geotiff/__init__.py`, the bbox-to-window resolution sits ahead of the `_validate_stable_only_remote` call and ahead of the VRT dispatch into `read_vrt` (which holds the VRT stable-only gate). The docstring on `_validate_stable_only_remote` in `xrspatial/geotiff/_validation.py` states the remote rejection must happen before any range GET or decode work. The bbox path breaks that ordering.

**Expected behavior**

With `stable_only=True` and no `allow_experimental_codecs`:

- A remote (HTTP / fsspec) source raises `RemoteStableSourcesOnlyError` before any bbox metadata read.
- A `.vrt` source raises `VRTStableSourcesOnlyError` before the VRT XML is parsed for the bbox.

Passing `bbox=` should not change whether or when the rejection happens.

**Affected combinations (currently uncovered)**

- `open_geotiff(remote, bbox=..., stable_only=True)`
- `open_geotiff(remote_vrt, bbox=..., stable_only=True)`

**Fix**

Move the remote and VRT stable-only checks ahead of bbox resolution: after source classification and dispatcher kwarg validation, but before any geo-metadata read or VRT parse that bbox handling triggers.

**Additional context**

Part of the GeoTIFF release gate work tracked under epic #2342.

brendancol:
Closes #2869

`open_geotiff(..., bbox=...)` resolved the bounding box to a pixel window before the `stable_only=True` rejection ran for remote and VRT sources. The bbox resolver reads source geo metadata first: the TIFF path reads a header (a range GET for an HTTP/fsspec source) and the VRT path parses the VRT XML. Both stable-only gates only fired after that, so a `stable_only=True` request could trigger remote I/O or a VRT parse before it got refused. The docstring on `_validate_stable_only_remote` requires the rejection to land before any range GET or decode work.

## Changes

- Move both stable-only gates ahead of bbox resolution in `open_geotiff`: `_validate_stable_only_vrt` for `.vrt` sources, `_validate_stable_only_remote` otherwise. Each helper is a no-op for the wrong source type.
- Drop the old post-VRT-branch `_validate_stable_only_remote` call, now redundant since the gate runs earlier. `read_vrt` keeps its own gate for direct callers, so the VRT path still fails closed on both entry points.

## Backend coverage

Dispatcher-level validation, so it applies uniformly across numpy / dask. The fsspec + bbox case is exercised through both the eager and dask dispatch paths. No GPU or numerical kernel change.

## Test plan

- [x] `test_stable_only_bbox_ordering_2869.py`: remote (HTTP + fsspec) and VRT bbox + stable_only combinations, each pointed at a source whose bbox read would raise a different error (fetch failure / XML `ParseError`) if it ran first, so the typed stable-only error proves the gate runs ahead of the read.
- [x] Positive controls: `allow_experimental_codecs=True` still resolves the bbox and reads for both fsspec and VRT sources.
- [x] Existing `test_stable_only_remote_2821.py`, bbox suites, and the geotiff release-gate tests pass unchanged.

### DIFF
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -842,6 +842,33 @@ def open_geotiff(source: str | BinaryIO, *,
     _is_vrt_source = (
         isinstance(source, str) and source.lower().endswith('.vrt'))
 
+    # Gate ``stable_only=True`` BEFORE resolving ``bbox=``. The bbox
+    # resolver reads source geo metadata first (the TIFF path reads a
+    # header, which is a range GET for an HTTP / fsspec source; the VRT
+    # path parses the VRT XML), so the stable-only rejection has to run
+    # ahead of it or a stable-only request would trigger remote I/O or a
+    # VRT parse before it is refused. ``_validate_stable_only_remote``
+    # documents exactly this ordering contract ("before any range GET or
+    # decode work"). ``_validate_stable_only_vrt`` is the matching gate
+    # for ``.vrt`` sources; ``read_vrt`` runs it again on the direct-call
+    # path, so this is defence in depth, not the only gate. Each helper is
+    # a no-op for the wrong source type, but branching keeps the intent
+    # obvious.
+    from ._validation import _validate_stable_only_remote
+    from ._validation import _validate_stable_only_vrt
+    if _is_vrt_source:
+        _validate_stable_only_vrt(
+            source,
+            stable_only=stable_only,
+            allow_experimental_codecs=allow_experimental_codecs,
+        )
+    else:
+        _validate_stable_only_remote(
+            source,
+            stable_only=stable_only,
+            allow_experimental_codecs=allow_experimental_codecs,
+        )
+
     # Resolve ``bbox=`` to a pixel ``window=`` via a header-only
     # metadata read. Done at the dispatcher so every backend
     # (eager / dask / GPU / VRT) sees a uniform pixel window without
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Gate stable_only=True before bbox metadata read on remote/VRT sources

None.

None.

- `xrspatial/geotiff/__init__.py:859-870`: moving the stable-only gate above the bbox block also moves it above the `window=`/`bbox=` mutual-exclusion `ValueError` (now at :880-884). So `open_geotiff(remote, window=X, bbox=Y, stable_only=True)` now raises the stable-only error instead of the mutual-exclusion error. That precedence is defensible (reject the unsupported tier before validating kwargs that won't be honored anyway) and probably more correct, but it is a small change in error ordering that nothing tests. Optional: note it in the comment, or leave as-is.

### What looks good
- The fix matches the contract documented on `_validate_stable_only_remote` ("before any range GET or decode work") and extends the same ordering to the VRT XML parse.
- Tests are well-targeted. Each points at a source whose bbox read would raise a distinct error (fetch failure / xml `ParseError`) if it ran first, so the typed stable-only error proves the ordering rather than just the eventual rejection. The negative control (malformed VRT without stable_only still hits `ParseError`) confirms the fixture exercises the path being short-circuited.
- The redundant post-VRT-branch `_validate_stable_only_remote` call is removed cleanly. `read_vrt` keeps its own gate, so direct callers still fail closed.
- Positive controls with `allow_experimental_codecs=True` confirm the bbox still resolves and reads when the tier is unlocked.

### Checklist
- [x] Validation ordering matches the documented contract
- [x] Remote (HTTP + fsspec) and VRT bbox paths covered by tests
- [x] Eager and dask dispatch paths both exercised
- [x] No new public API; no README or notebook needed
- [x] No numerical/kernel change; backend parity not affected
- [x] Diff is surgical, no unrelated edits

brendancol (commented):
## Follow-up review

The one nit from the first pass is addressed in a1c75a2c:

- The comment at `xrspatial/geotiff/__init__.py:854-861` now spells out that the stable-only gate runs ahead of the `window=`/`bbox=` mutual-exclusion check, so the tier rejection wins.
- `test_stable_only_wins_over_window_bbox_conflict` pins that precedence: a remote source under `stable_only=True` with both `window=` and `bbox=` reports `RemoteStableSourcesOnlyError`, not the kwarg conflict.

No new findings. The full `test_stable_only_bbox_ordering_2869.py` suite (8 tests) passes, along with the existing stable-only, bbox, and geotiff release-gate suites. The branch is even with `origin/main` with no conflicts. Nothing blocking.

PR #2883, opened 2026-06-02
Validate MFD fraction values in public consumers (#2873)

### LINKED ISSUE #2873: MFD fraction-grid functions accept invalid fractions without validation
## Summary

The public MFD functions document a specific input contract: each cell's 8 fraction bands lie in [0, 1] and sum to either 1.0 (flow) or 0.0 (pit/flat/sink), with all-NaN bands at edges and nodata cells. The functions don't enforce any of that. They check the array is shape (8, H, W) and nothing else about the values.

So a caller can pass negative fractions, per-cell sums above 1.0, or a cell where some of the 8 bands are NaN and the rest are finite, and the function will happily run accumulation, length, order, or link math on it and hand back hydrology that looks fine but isn't. These are public APIs. Bad input should raise early with a clear message instead of producing quietly wrong output.

## Affected functions

The public fraction-grid consumers in `xrspatial/hydro`:

- `flow_accumulation_mfd`
- `flow_length_mfd`
- `stream_order_mfd`
- `stream_link_mfd`
- `flow_path_mfd`
- `hand_mfd`
- `watershed_mfd`

## Proposal

Add fraction-value validation that runs right after the existing (8, H, W) shape check in each function. Check three things:

- No negative fractions.
- Each cell's band sum is 1.0 (flow) or 0.0 (pit/flat/sink), within a small tolerance.
- NaN bands are all-or-nothing per cell: either all 8 directions are NaN (edge/nodata) or none are. Reject the partially-NaN case.

Raise `ValueError` on any violation. Keep the documented semantics (sum == 1.0 or 0.0). The tolerance needs to be loose enough that valid grids out of `flow_direction_mfd` still pass.

## Scope

Fraction value and topology only (negative, sum, NaN consistency). Secondary-raster shape validation, cyclic-graph detection, and dask-laziness work are separate issues.

brendancol:
Closes #2873

## What

The public MFD functions document a fraction-grid contract: bands in [0, 1], per-cell sum of 1.0 or 0.0, all-NaN bands at edges. They only enforced the (8, H, W) shape, though. Feed them bad fraction values and you got wrong-but-plausible hydrology back instead of an error.

This adds `_validate_mfd_fractions` to `xrspatial/utils.py` and calls it right after the shape check in all seven public consumers:

- flow_accumulation_mfd
- flow_length_mfd
- stream_order_mfd
- stream_link_mfd
- flow_path_mfd
- hand_mfd
- watershed_mfd

The check raises a clear ValueError for negative fractions, per-cell band sums that are neither ~1.0 nor ~0.0 (tolerance 1e-6), and partial-NaN cells (some of the 8 bands NaN, the rest finite).

## Backends

numpy and cupy inputs are validated eagerly. Dask inputs (numpy- or cupy-backed) skip the value check so laziness is preserved; lazy validation is a separate concern tracked elsewhere.

## Scope

Fraction value and topology only. Secondary-raster shape validation, cyclic-graph detection, and dask-laziness are sibling PRs and aren't touched here.

## Test plan

- [x] New tests in `test_validate_mfd_fractions.py` cover negative, bad-sum, and partial-NaN rejection across all seven functions
- [x] Valid grids from `flow_direction_mfd` still pass, including sink cells that sum to 0.0 and all-NaN edge cells
- [x] Dask input with invalid values does not raise eagerly
- [x] Existing MFD suite (144 tests) still green

### DIFF
--- a/xrspatial/hydro/flow_accumulation_mfd.py
+++ b/xrspatial/hydro/flow_accumulation_mfd.py
@@ -30,6 +30,7 @@ except ImportError:
     da = None
 
 from xrspatial.utils import (
+    _validate_mfd_fractions,
     _validate_raster,
     cuda_args,
     has_cuda_and_cupy,
@@ -841,6 +842,9 @@ def flow_accumulation_mfd(flow_dir_mfd: xr.DataArray,
             "flow_dir_mfd must be a 3-D array of shape (8, H, W), "
             f"got shape {data.shape}"
         )
+
+    _validate_mfd_fractions(data, func_name='flow_accumulation_mfd',
+                            name='flow_dir_mfd')
 
     if isinstance(data, np.ndarray):
         _check_memory(data.shape[1], data.shape[2])
--- a/xrspatial/hydro/flow_length_mfd.py
+++ b/xrspatial/hydro/flow_length_mfd.py
@@ -25,6 +25,7 @@ except ImportError:
 
 from xrspatial.hydro._boundary_store import BoundaryStore
 from xrspatial.utils import (
+    _validate_mfd_fractions,
     _validate_raster,
     get_dataarray_resolution,
     has_cuda_and_cupy,
@@ -1025,6 +1026,9 @@ def flow_length_mfd(flow_dir_mfd: xr.DataArray,
             "flow_dir_mfd must be a 3-D array of shape (8, H, W), "
             f"got shape {data.shape}"
         )
+
+    _validate_mfd_fractions(data, func_name='flow_length_mfd',
+                            name='flow_dir_mfd')
 
     cellsize_x, cellsize_y = get_dataarray_resolution(flow_dir_mfd)
     if not (np.isfinite(cellsize_x) and cellsize_x != 0
--- a/xrspatial/hydro/flow_path_mfd.py
+++ b/xrspatial/hydro/flow_path_mfd.py
@@ -29,6 +29,7 @@ except ImportError:
     da = None
 
 from xrspatial.utils import (
+    _validate_mfd_fractions,
     _validate_raster,
     has_cuda_and_cupy,
     is_cupy_array,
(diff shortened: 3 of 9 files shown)


brendancol (commented):
## PR Review: Validate MFD fraction values in public consumers (#2873)

None.

- `utils.py:244` builds `finite = data[~nan_mask]` just to test for negatives, which copies the whole grid when nothing is NaN. You can skip the copy: `bool((data < 0).any())` does the same thing, because `NaN < 0` is False so NaN cells never trip it. Only matters for big grids, but the copy is avoidable.

- The three error messages each rebuild the `func_name(): name ...` prefix. With only three it's fine, but a local prefix variable would trim the repetition.
- The helper docstring notes that only numpy and cupy inputs get value-checked. Worth saying the same in the public function docstrings so nobody assumes a dask grid gets the same guard. Optional.

### What looks good

- The check sits after the shape check and before any hydrology runs in all seven consumers, so bad input fails early like it should.
- Sinks (sum 0.0) and all-NaN edge cells both still pass, and the tests assert that.
- Partial-NaN detection via the per-cell band count (0 < count < 8) is the right topology test.
- Dask is skipped on purpose to keep laziness intact, which keeps this PR in scope and out of the dask-laziness sibling PR's way.
- Tests cover negative, bad-sum, and partial-NaN across all seven functions, plus the valid-input and dask-skip paths.

### Checklist

- [x] Matches the documented fraction contract
- [x] NaN handling correct (partial vs all-NaN distinguished)
- [x] Edge cases covered (sink, all-NaN edge, valid grid)
- [x] Dask path intentionally skipped and documented
- [x] No new public function, so no benchmark/README needed
- [x] Helper has a full docstring

brendancol (commented):
## Follow-up review (after 2b030e38)

Re-checked the helper in `xrspatial/utils.py` after the review fixes.

### Resolved
- Negative check no longer copies the grid. It now uses `bool((data < 0).any())`, and NaN cells don't trip it since `NaN < 0` is False. Confirmed no RuntimeWarning under `-W error::RuntimeWarning`.
- The repeated `func_name(): name` prefix is now a single local `prefix` variable shared by all three error messages.

### Still open (deferred, not blocking)
- Adding a "dask inputs aren't value-checked" note to each of the seven public docstrings. Deferred: it would edit docstrings in all seven MFD files, which are being touched by sibling PRs right now, so the extra conflict surface isn't worth it. The helper's own docstring already documents the dask-skip behavior. Can fold into a later docs pass.

Tests: full MFD suite plus the new validation tests pass (151 passed).

PR #2884, opened 2026-06-02
Reject cyclic MFD fraction grids instead of returning wrong results or hanging

### LINKED ISSUE #2866: MFD flow functions accept cyclic fraction grids: wrong results or infinite hang
**Describe the bug**

The MFD hydrology functions assume their input fraction grid is a directed acyclic graph (DAG) but never check it. A hand-built or user-supplied fraction grid with a cycle is not rejected, so the functions either return wrong numbers or never terminate.

Three functions are affected:

- `flow_accumulation_mfd` runs a Kahn topological traversal but never confirms every valid cell was processed. Cells inside a cycle keep their initial value. A 2-cell cycle returns `[[1., 1.]]` (each cell's own area only) instead of raising.
- `flow_length_mfd` has the same gap in the downstream and upstream CPU kernels. Cells in a cycle never enter the topological order, so they keep flow length 0.
- `flow_path_mfd` traces the dominant neighbor in an unconditional `while True` loop with no visited set and no step cap. A cyclic grid loops forever and hangs the process.

**Expected behavior**

A cyclic fraction grid should raise a clear `ValueError` instead of returning wrong values or hanging. After the Kahn traversal, `flow_accumulation_mfd` and `flow_length_mfd` should confirm the processed-cell count equals the valid-cell count and raise if a cycle remains. `flow_path_mfd` should detect the cycle, stop, and raise rather than spinning forever.

**Additional context**

Affected sites (approximate): the CPU BFS in `xrspatial/hydro/flow_accumulation_mfd.py`, the downstream and upstream CPU BFS in `xrspatial/hydro/flow_length_mfd.py`, and the `while True` traversal in `xrspatial/hydro/flow_path_mfd.py` (CPU and dask paths).

Scope is cycle-detection and non-termination hardening only. Secondary-raster shape validation, fraction-range validation, and dask-laziness work are tracked separately.

brendancol:
Closes #2866

The MFD flow functions assume their fraction grid is a DAG but never check it. A hand-built or user-supplied grid with a cycle slipped through: `flow_accumulation_mfd` and `flow_length_mfd` returned wrong numbers (cells in the cycle kept their initial value), and `flow_path_mfd` looped forever.

## Changes

- `flow_accumulation_mfd` and `flow_length_mfd`: after the Kahn traversal, compare the processed-cell count against the valid-cell count. If a cycle left some cells with nonzero in-degree, raise a `ValueError` naming the problem. Applied to the eager CPU kernels and the dask tile kernels (downstream and upstream).
- `flow_path_mfd`: the dominant-neighbor trace now caps at `H*W` steps. A path through a DAG visits each cell at most once, so exceeding that bound means a cycle. Applied to the CPU `while True` loop and the dask tracing loop. Both raise instead of spinning.

## Backend coverage

The cycle guard runs on numpy, dask+numpy, and dask+cupy (the cupy and dask+cupy paths route through the same CPU/tile kernels). cupy follows the CPU kernel as well.

## Test plan

- [x] New `ValueError` tests for a 2-cell cycle on numpy and dask, for all three functions and both flow_length directions
- [x] Existing 82 MFD tests still pass (90 total with the new ones)
- [x] Confirmed a valid `flow_direction_mfd` grid still computes on every backend

### DIFF
--- a/xrspatial/hydro/flow_accumulation_mfd.py
+++ b/xrspatial/hydro/flow_accumulation_mfd.py
@@ -163,6 +163,7 @@ def _flow_accum_mfd_cpu(fractions, height, width):
     valid = np.zeros((height, width), dtype=np.int8)
 
     # Pass 1: initialise
+    n_valid = 0
     for r in range(height):
         for c in range(width):
             v = fractions[0, r, c]
@@ -171,6 +172,7 @@ def _flow_accum_mfd_cpu(fractions, height, width):
             else:
                 valid[r, c] = 1
                 accum[r, c] = 1.0
+                n_valid += 1
 
     # Pass 2: compute in-degrees
     for r in range(height):
@@ -216,6 +218,15 @@ def _flow_accum_mfd_cpu(fractions, height, width):
                             queue_r[tail] = nr
                             queue_c[tail] = nc
                             tail += 1
+
+    # If a cycle remains, some valid cells never reached in_degree 0 and
+    # were never dequeued.  head counts the cells processed by the BFS.
+    if head < n_valid:
+        raise ValueError(
+            "flow_accumulation_mfd: the MFD fraction grid contains a cycle "
+            "(some cells never reach zero in-degree).  The input must be a "
+            "directed acyclic graph, as produced by flow_direction_mfd."
+        )
 
     return accum
 
@@ -402,12 +413,14 @@ def _flow_accum_mfd_tile_kernel(fractions, h, w,
     valid = np.zeros((h, w), dtype=np.int8)
 
     # Initialise
+    n_valid = 0
     for r in range(h):
         for c in range(w):
             v = fractions[0, r, c]
             if v == v:  # not NaN
                 valid[r, c] = 1
                 accum[r, c] = 1.0
+                n_valid += 1
             else:
                 accum[r, c] = np.nan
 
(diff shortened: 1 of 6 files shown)


brendancol (commented):
## PR Review: Reject cyclic MFD fraction grids instead of returning wrong results or hanging

### Blockers
None.

### Suggestions
None.

### Nits
- The same ValueError message is repeated across six sites (the CPU and dask-tile kernels in flow_accumulation_mfd.py and flow_length_mfd.py, plus two in flow_path_mfd.py). Numba njit functions can't pull a raise message from a module-level constant, so the duplication is hard to avoid for the kernels. Not worth refactoring.

### What looks good
- The cycle check is the standard Kahn-completion test: after the topological traversal, compare the processed-cell count against the valid-cell count. Correct, with O(1) extra work.
- `head` (processed) and `tail` (ordered) are equivalent here because the BFS drains its queue. Both stay below n_valid only when a cycle remains.
- flow_path_mfd caps the trace at H*W steps. A dominant-neighbor path through a DAG visits each cell at most once, so the bound is tight and won't false-positive. A full-length 2-cell valid path (H*W=2) does not trip the cap.
- No false positives on valid grids: pits and flats are in_degree-0 sources, and off-grid flow creates no in-grid edges. Confirmed on a random elevation-derived MFD grid across numpy and dask.
- The guard sits in both the eager and dask-tile kernels, so numpy, dask+numpy, and dask+cupy all raise. cupy routes through the CPU kernel.
- Tests cover numpy and dask for all three functions and both flow_length directions. 8 new tests; the existing 82 still pass.

### Checklist
- Algorithm matches reference: yes (Kahn completion check)
- All implemented backends consistent: yes
- NaN handling correct: unchanged; n_valid counts non-NaN cells
- Edge cases covered: cycle on numpy and dask; 1x1 and full-length paths checked manually
- Dask chunk boundaries: guard fires per tile; cross-tile cycles are out of scope for this PR
- No premature materialization: just an integer counter
- Benchmark: existing flow benchmarks apply; no new function added
- README matrix: not applicable, no new function or backend change
- Docstrings: unchanged, still accurate

Issue #2885, opened 2026-06-02
Extend lazy MFD dask assembly to stream/hand modules and study the convergence phase


brendancol:
Follow-up to #2865, which made the dask output assembly lazy in `flow_accumulation_mfd`, `flow_length_mfd`, and `watershed_mfd`.

## Remaining eager paths

The same eager-assembly pattern (`.compute()` per tile, then `da.from_array` / `da.block`) lives in other MFD dask paths that #2865 left untouched to keep its diff scoped and reduce conflicts with parallel hydro work:

- `xrspatial/hydro/stream_link_mfd.py`, `_stream_link_mfd_dask`
- `xrspatial/hydro/stream_order_mfd.py`
- `xrspatial/hydro/hand_mfd.py`

These can take the same `da.map_blocks` treatment. They have extra secondary inputs (accumulation, stream rasters) that need the same chunk-alignment handling used for pour points in `watershed_mfd`.

## Convergence phase

The iterative boundary-convergence sweep that runs before assembly still reads tile data eagerly. That is inherent to the algorithm: converging cross-tile boundaries needs real values across sweeps. Making it lazy would need a different approach (for example a fixed number of `map_overlap` rounds, or a delayed reduction), which is a separate design question worth its own investigation. Filing it here so it is tracked rather than lost.

PR #2886, opened 2026-06-02
Keep MFD dask output assembly lazy

### LINKED ISSUE #2865: MFD dask paths materialize the output raster eagerly during the API call
## Description

Several MFD dask code paths call `.compute()` eagerly while building the output DataArray. The final assembly step in `flow_accumulation_mfd`, `flow_length_mfd`, and `watershed_mfd` computes each tile and wraps the materialized numpy result with `da.from_array` / `da.block`. The result is that the full output raster is materialized in memory during the function call, before the user ever asks for `.compute()`. That defeats dask laziness and risks OOM on the large rasters hydrology typically handles.

Affected:
- `xrspatial/hydro/flow_accumulation_mfd.py`, `_assemble_result_mfd`
- `xrspatial/hydro/flow_length_mfd.py`, `_assemble_result`
- `xrspatial/hydro/watershed_mfd.py`, the assembly block in `_watershed_mfd_dask`

## Proposed fix

Rework the final assembly so the returned DataArray stays lazy. Build the result with `da.map_blocks` over the original lazy fractions array, passing the small converged boundary snapshot into the block function so the per-tile kernel runs at compute time instead of during the API call. Verify results still match the numpy path and watch for OOM.

## Scope note

The iterative boundary-convergence sweep that runs before assembly reads tile data eagerly by design, because converging cross-tile boundaries is inherently iterative. Lazifying that phase is a separate, larger design question and is out of scope here; track it separately.

brendancol:
Closes #2865.

## What changed

- The dask assembly step in `flow_accumulation_mfd`, `flow_length_mfd`, and `watershed_mfd` used to `.compute()` each tile and wrap the numpy result with `da.from_array` / `da.block`, which materialized the whole output raster during the function call.
- Rebuilt the assembly with `da.map_blocks` over the lazy fractions array. The converged boundary snapshot and fraction strips are small, so they ride along in a closure and the per-tile kernel runs at compute time. The returned DataArray is now genuinely lazy.
- Added a rechunk guard so the 8 direction bands stay in one chunk (every tile kernel needs all 8), and aligned the pour-points chunks to the fractions tile grid in `watershed_mfd`.

The boundary-convergence sweep that runs before assembly still reads tile data eagerly. That is inherent to the algorithm and is tracked separately in #2885, along with the same assembly rework for the stream and hand MFD modules.

## Backend coverage

numpy and cupy paths are unchanged. The lazy rework applies to dask+numpy and dask+cupy (the cupy variant converts to numpy, runs the same iterative path, and converts back).

## Test plan

- [x] `flow_accumulation_mfd`, `flow_length_mfd`, `watershed_mfd` dask results match numpy
- [x] New test asserts the assembly kernel runs at compute time, not during the call
- [x] New test covers an input chunked along the 8-band axis (rechunk guard)
- [x] New test covers pour points chunked differently from fractions
- [x] Full hydro suite passes (860 tests)

### DIFF
--- a/xrspatial/hydro/flow_accumulation_mfd.py
+++ b/xrspatial/hydro/flow_accumulation_mfd.py
@@ -698,6 +698,11 @@ def _flow_accum_mfd_dask_iterative(fractions_da, chunks_y, chunks_x):
     n_tile_y = len(chunks_y)
     n_tile_x = len(chunks_x)
 
+    # The 8 direction bands must stay in a single chunk: every tile kernel
+    # needs all 8 fractions, and the lazy assembly drops axis 0 per block.
+    if fractions_da.chunks[0] != (fractions_da.shape[0],):
+        fractions_da = fractions_da.rechunk({0: fractions_da.shape[0]})
+
     # Phase 0: extract boundary fraction strips
     frac_bdry = _preprocess_mfd_tiles(fractions_da, chunks_y, chunks_x)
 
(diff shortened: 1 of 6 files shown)


brendancol (commented):
## PR Review: Keep MFD dask output assembly lazy

None.

None blocking. The change is well scoped and the tests cover the laziness contract directly.

- The three `_tile` closures in flow_accumulation_mfd.py, flow_length_mfd.py, and watershed_mfd.py are nearly identical, including the searchsorted tile-index lookup. A shared helper could remove the duplication, but the duplication is small and the sibling rockout PRs touch these same files, so factoring it out now would raise the conflict risk. Reasonable to leave as is for this PR.

### What looks good
- The map_blocks rework keeps the returned DataArray lazy while leaving the numpy and cupy paths untouched. The boundary-convergence sweep stays eager by necessity, and that limitation is documented in the docstrings and tracked in the follow-up issue #2885.
- `block_info[0]['array-location']` reads from the input fractions coordinates, which is the right reference for the tile index even after `drop_axis=0` collapses the band axis. The watershed path confirms this works with a second 2-D input array.
- The rechunk guard for the 8-band axis and the pour-points chunk alignment are both covered by new tests, including a deliberate chunk mismatch.
- Correctness is checked against the numpy path across single-chunk, many-chunk, and band-axis-chunked inputs, with equal_nan comparisons.

### Checklist
- Algorithm matches reference: unchanged, kernels reused as is
- Backends consistent: numpy/cupy unchanged; dask+numpy verified against numpy in tests. dask+cupy shares the same iterative path but needs a GPU runner to verify directly.
- NaN handling correct: unchanged from prior tile kernels; equal_nan tests pass
- Edge cases covered: single tile, many tiles, band-axis chunked, pour-points mismatch
- Dask chunk boundaries handled: rechunk guard plus pour-points alignment
- No premature materialization: that is the point of this PR; assembly is now lazy
- Benchmark: not needed, no new public API
- README: not applicable, no new function
- Docstrings present and accurate: assembly helpers document the laziness contract

PR #2887, opened 2026-06-02
Build GPU rtx mesh with real x/y resolution (#2861)

### LINKED ISSUE #2861: GPU viewshed ignores x/y resolution: anisotropic rasters diverge from CPU
**Describe the bug**

The GPU ray-tracing viewshed builds its terrain mesh on integer grid coordinates and ignores the real cell resolution. In `xrspatial/gpu_rtx/mesh_utils.py`, the mesh vertices sit at `(w, h)` integer indices (`_triangulate_terrain_kernel`, around line 77/84). The real `ew_res`/`ns_res` only come into play later, in `xrspatial/gpu_rtx/viewshed.py` around line 243, when the output vertical angle is computed.

So on an anisotropic raster (`ew_res != ns_res`), the GPU traces a different terrain than the CPU sweep does. The CPU path (`xrspatial/viewshed.py`) scales horizontal distances by the real `ew_res`/`ns_res` wherever it computes gradients and distances, so occlusion gets decided in real-world units. The GPU decides occlusion in integer grid units. A ridge that blocks line of sight in real units may not block it in grid units, or the other way around, and the two backends end up reporting different sets of visible cells.

**Expected behavior**

GPU and CPU viewshed should agree on an anisotropic raster, within the small float32 ray-tracing tolerance. The GPU mesh geometry should reflect the real cell resolution so the ray tracer sees the same terrain the CPU sweep does.

**Reproduction**

```python
import numpy as np
import xarray as xa
import cupy as cp

ny, nx = 15, 15
terrain = np.full((ny, nx), 1.0)
terrain[:, 7] = 8.0  # ridge that occludes

xs = np.arange(nx, dtype=float) * 1.0    # ew_res = 1
ys = np.arange(ny, dtype=float) * 10.0   # ns_res = 10  (anisotropic)

obs_x, obs_y, obs_elev = xs[3], ys[7], 2

v_np = viewshed(xa.DataArray(terrain, coords=dict(x=xs, y=ys), dims=["y","x"]),
                x=obs_x, y=obs_y, observer_elev=obs_elev)
v_gpu = viewshed(xa.DataArray(cp.asarray(terrain), coords=dict(x=xs, y=ys), dims=["y","x"]),
                 x=obs_x, y=obs_y, observer_elev=obs_elev)

n = v_np.values
g = v_gpu.data.get()
print("visibility disagreements:", ((n > -1) != (g > -1)).sum(), "of", n.size)
# -> 105 of 225
```

**Additional context**

The fix is to build the GPU mesh (and the matching camera/viewpoint ray origins) in real-resolution coordinates instead of integer grid indices, and keep the downstream angle calculation consistent so resolution is not applied twice. `create_triangulation` is shared with the RTX hillshade path, so that path has to stay consistent too. GPU-only code path.

brendancol:
## Summary

Closes #2861.

The GPU ray-tracing mesh in `create_triangulation` (`xrspatial/gpu_rtx/mesh_utils.py`) placed vertices at integer grid indices and ignored the raster's `ew_res`/`ns_res`, while the CPU viewshed sweep works in real-world units. This builds the mesh in real-resolution coordinates so the GPU geometry matches the CPU coordinate convention.

- Mesh vertices now sit at `(col * ew_res, row * ns_res)`, and the camera rays in both the viewshed and hillshade RTX paths start at the matching real-world positions.
- The z-scale stays resolution-independent (`max(H, W) / maxH`). It must not depend on `ew_res`/`ns_res`, or it would cancel the x/y stretch.
- `create_triangulation` now returns `(scale, ew_res, ns_res)` so callers can cast rays at the matching positions.

## Investigation note on impact

While implementing this I checked the end-to-end effect, and it's smaller than the issue implies. The issue measured 105/225 cells disagreeing between GPU and CPU on an anisotropic raster. That disagreement isn't caused by the mesh ignoring x/y resolution. It comes from the separate z-scale heuristic that flattens the mesh.

Two things explain why the mesh x/y fix alone doesn't change the viewshed result:

1. Ray-triangle occlusion is invariant under a per-axis (linear) scaling of the mesh. Stretching the mesh in y by a constant maps lines to lines and preserves which triangles a ray hits, so the set of occluded cells doesn't change.
2. The viewshed output angle is already computed from the real `ew_res`/`ns_res` downstream.

I confirmed this empirically: across several anisotropic varied-terrain rasters the GPU output is byte-identical before and after the change (max abs diff 0.0, 0 occlusion flips), and isotropic hillshade output is byte-identical too. The CPU sweep is likewise occlusion-invariant to anisotropic resolution along any direction.

So this PR removes a latent geometry inconsistency and makes the GPU mesh trace the same shape the CPU works on, rather than fixing a numeric divergence. The larger GPU-vs-CPU difference on varied terrain is the z-scale heuristic, which is out of scope here. Happy to file a follow-up if maintainers want that looked at.

## Backend coverage

GPU only (cupy and dask+cupy route through the same RTX path). The numpy and dask+numpy viewshed backends don't use this mesh.

## Test plan

- [x] `test_cell_resolution_*` cover anisotropic coords and the no-coords unit fallback.
- [x] `test_triangulate_places_vertices_at_real_resolution` asserts vertices land at `(col*ew_res, row*ns_res)`.
- [x] `test_create_triangulation_returns_resolution` covers the new return tuple.
- [x] `test_viewshed_gpu_anisotropic_matches_cpu` (both anisotropy orientations) checks GPU vs CPU on an anisotropic raster.
- [x] Existing viewshed, hillshade, and mesh-guard tests pass.
- [x] GPU tests are gated on `has_rtx()` and skip without a CUDA / rtxpy device.

### DIFF
--- a/xrspatial/gpu_rtx/hillshade.py
+++ b/xrspatial/gpu_rtx/hillshade.py
@@ -200,7 +200,9 @@ def hillshade_rtx(raster: xr.DataArray,
     _check_gpu_memory("hillshade_rtx", H, W)
 
     optix = RTX()
-    scale, ew_res, ns_res = create_triangulation(raster, optix)
+    # hillshade does not use the z-scale; only viewshed scales the observer
+    # and target elevations by it.
+    _, ew_res, ns_res = create_triangulation(raster, optix)
 
     return _hillshade_rt(
         raster, optix, azimuth=azimuth, angle_altitude=angle_altitude,
--- a/xrspatial/gpu_rtx/hillshade.py
+++ b/xrspatial/gpu_rtx/hillshade.py
@@ -15,24 +15,28 @@ from .mesh_utils import create_triangulation
 
 
 @nb.cuda.jit
-def _generate_primary_rays_kernel(data, H, W):
+def _generate_primary_rays_kernel(data, H, W, ew_res, ns_res):
     """
     A GPU kernel that given a set of x and y discrete coordinates on a raster
     terrain generates in @data a list of parallel rays that represent camera
     rays generated from an orthographic camera that is looking straight down
     at the surface from an origin height 10000.
+
+    Ray origins are placed at the real-world cell centres (column * ew_res,
+    row * ns_res) so they land on the resolution-aware mesh built by
+    ``create_triangulation`` (issue #2861).
     """
     i, j = nb.cuda.grid(2)
     if i >= 0 and i < H and j >= 0 and j < W:
         if (j == W-1):
-            data[i, j, 0] = j - 1e-3
+            data[i, j, 0] = j * ew_res - 1e-3 * ew_res
         else:
-            data[i, j, 0] = j + 1e-3
+            data[i, j, 0] = j * ew_res + 1e-3 * ew_res
 
         if (i == H-1):
-            data[i, j, 1] = i - 1e-3
+            data[i, j, 1] = i * ns_res - 1e-3 * ns_res
         else:
-            data[i, j, 1] = i + 1e-3
+            data[i, j, 1] = i * ns_res + 1e-3 * ns_res
 
         data[i, j, 2] = 10000  # Location of the camera (height)
         data[i, j, 3] = 1e-3
@@ -42,9 +46,10 @@ def _generate_primary_rays_kernel(data, H, W):
         data[i, j, 7] = np.inf
 
 
-def _generate_primary_rays(rays, H, W):
+def _generate_primary_rays(rays, H, W, ew_res, ns_res):
     griddim, blockdim = calc_cuda_dims((H, W))
-    _generate_primary_rays_kernel[griddim, blockdim](rays, H, W)
+    _generate_primary_rays_kernel[griddim, blockdim](
+        rays, H, W, ew_res, ns_res)
     return 0
 
 
@@ -148,7 +153,9 @@ def _hillshade_rt(raster: xr.DataArray,
                   optix: RTX,
                   azimuth: int,
                   angle_altitude: int,
-                  shadows: bool) -> xr.DataArray:
+                  shadows: bool,
+                  ew_res: float,
+                  ns_res: float) -> xr.DataArray:
     H, W = raster.shape
     sun_dir = cupy.array(_get_sun_dir(angle_altitude, azimuth))
 
@@ -158,7 +165,7 @@ def _hillshade_rt(raster: xr.DataArray,
     d_aux = cupy.empty((H, W, 3), np.float32)
     d_output = cupy.empty((H, W), np.float32)
 
-    _generate_primary_rays(d_rays, H, W)
+    _generate_primary_rays(d_rays, H, W, ew_res, ns_res)
     device = cupy.cuda.Device(0)
     device.synchronize()
     res = optix.trace(d_rays, d_hits, W*H)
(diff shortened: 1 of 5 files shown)


brendancol (commented):
## PR Review: Build GPU rtx mesh with real x/y resolution (#2861)

None.

- [ ] `xrspatial/gpu_rtx/hillshade.py:203` -- `scale, ew_res, ns_res = create_triangulation(...)` unpacks `scale`, but hillshade never uses it (only viewshed scales observer/target elev by it). Harmless, but unpacking to `_` (`_, ew_res, ns_res = ...`) would signal that intent. Minor.

- [ ] `xrspatial/gpu_rtx/mesh_utils.py:14` -- `_axis_resolution` reads `index.values` and assumes evenly spaced coordinates (first and last divided by n-1). That matches how the CPU viewshed computes its own ew_res/ns_res, so the two are consistent, but neither path detects an irregular coordinate axis. Not introduced here, just flagging the shared assumption.

### What looks good
- The new `_axis_resolution` guards `n <= 1` and returns 1.0, which removes a latent divide-by-zero in `_viewshed_rt`. The old code did `ew_res = (x_range[1]-x_range[0])/(W-1)` and would have raised on a single-column GPU raster. Nice incidental fix.
- The z-scale stays resolution-independent, and the docstring explains why a resolution-dependent z-scale would cancel the x/y stretch. That's the subtle part and it's documented well.
- The sub-cell ray nudge is scaled by ew_res/ns_res so it stays sub-cell once the coordinates are in real-world units. Easy to miss.
- The PR is upfront that the end-to-end viewshed result doesn't change (occlusion is invariant under per-axis linear scaling), and the byte-identical before/after check backs that up.
- Hillshade and viewshed both go through the shared mesh, and both ray generators were updated together, so the two paths stay in sync.

### Test coverage notes
- Mesh-level tests assert the vertices land at (col*ew_res, row*ns_res) and cover the no-coords unit fallback. The end-to-end GPU-vs-CPU test covers both anisotropy orientations.
- The single-column (W=1) divide-by-zero fix in `_viewshed_rt` isn't tested directly on the GPU path. The CPU path already has `test_viewshed_single_row_or_column`; a GPU analogue would lock in the incidental fix, but it needs a CUDA device and the payoff is small.

### Checklist
- [x] Algorithm matches reference/convention (mesh now matches CPU coordinate units)
- [x] Backends consistent (GPU viewshed + hillshade share the mesh, both updated)
- [x] NaN handling unchanged (all-zero/all-NaN guard preserved)
- [x] Edge cases: single-column divide-by-zero now guarded
- [ ] Dask chunk boundaries -- n/a (GPU mesh path isn't chunked)
- [x] Benchmark -- n/a (no perf-sensitive change; output byte-identical)
- [x] README feature matrix -- n/a (no new public API)

brendancol (commented):
## Follow-up review (after 8c20f05)

### Suggestion -- fixed
- `hillshade.py` now unpacks the unused z-scale to `_` with a one-line comment about why hillshade ignores it. Hillshade GPU tests still pass.

### Nit -- dismissed (with reason)
- The even-spacing assumption in `_axis_resolution` is pre-existing: the CPU viewshed computes ew_res/ns_res the same way (first/last over n-1), so the GPU and CPU paths stay consistent. Detecting irregular coordinate axes would be a separate change across both backends, out of scope for this PR. Leaving it as-is.

No new findings. Geometry, ray generation, and the resolution-independent z-scale all look correct, and the end-to-end viewshed result stays byte-identical as documented.

PR #2898, opened 2026-06-03
Validate return_type before data work in contours() (#2891)

### LINKED ISSUE #2891: contours(): return_type validated too late; all-non-finite path silently accepts bad values
**Describe the bug**

`contours()` validates `return_type` too late. The only check is at the end of the
function (`xrspatial/contour.py:691`), inside the dispatch chain:

    if return_type == "numpy":
        return results
    elif return_type == "geopandas":
        ...
    else:
        raise ValueError(...)

That runs only after the raster has been validated, the levels computed, and the
contours extracted. Two problems fall out of the ordering:

1. A bad `return_type` isn't rejected until the whole extraction has already run, so
   the caller pays for work that was always going to fail.

2. The all-non-finite early return at `xrspatial/contour.py:653-656` short-circuits
   before that check ever happens:

        if not np.isfinite(vmin) or not np.isfinite(vmax):
            if return_type == "numpy":
                return []
            return _to_geopandas([], crs=agg.attrs.get('crs', None))

   Any non-`"numpy"` value falls through to the GeoDataFrame branch. So
   `contours(all_nan_raster, return_type="bad")` hands back an empty GeoDataFrame
   instead of raising. The bad argument is silently accepted.

**Expected behavior**

`return_type` should be checked near the top of `contours()`, right after
`_validate_raster` and before any level computation or data work. A bad value should
raise `ValueError` no matter what the input raster looks like, including the
all-non-finite case.

**Additional context**

Fix: add an explicit `return_type` check (allowed values `'numpy'` and `'geopandas'`)
at the top of the function. The existing late dispatch stays put but becomes
unreachable for bad values.

brendancol:
Closes #2891

`contours()` only checked `return_type` at the end of its dispatch chain.
The all-non-finite early return (`contour.py:653-656`) ran before that
check, so any non-`"numpy"` value fell through to the GeoDataFrame branch:
`contours(all_nan, return_type="bad")` handed back an empty GeoDataFrame
instead of raising.

- Add an explicit `return_type` check right after `_validate_raster`, before
  any level computation or extraction. Bad values raise `ValueError` with the
  same wording the late branch already used.
- The late `else: raise` stays in place but is now unreachable for bad values;
  the legitimate numpy/geopandas dispatch is untouched.

Backend coverage: pure validation at the DataArray boundary, so it applies
across numpy / cupy / dask+numpy / dask+cupy without backend-specific code.

Test plan:
- [x] `return_type="bad"` raises on a normal raster
- [x] `return_type="bad"` raises on the all-non-finite path (the silent case)
- [x] validation fires before data work on a degenerate all-NaN raster
- [x] `"numpy"` and `"geopandas"` still work
- [x] full `test_contour.py` passes (50 tests)

### DIFF
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -693,6 +693,8 @@ def contours(
         crs = agg.attrs.get('crs', None)
         return _to_geopandas(results, crs=crs)
     else:
+        # Unreachable: return_type is validated at the top of the function.
+        # Kept as a defensive guard.
         raise ValueError(
             f"Invalid return_type '{return_type}'. "
             "Allowed values are 'numpy' and 'geopandas'."
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -616,6 +616,11 @@ def contours(
     >>> level, coords = lines[0]
     """
     _validate_raster(agg, func_name='contours', name='agg', ndim=2)
+    if return_type not in ("numpy", "geopandas"):
+        raise ValueError(
+            f"Invalid return_type '{return_type}'. "
+            "Allowed values are 'numpy' and 'geopandas'."
+        )
     if agg.shape[0] < 2 or agg.shape[1] < 2:
         raise ValueError(
             "Input raster must have at least 2 rows and 2 columns"
--- a/xrspatial/tests/test_contour.py
+++ b/xrspatial/tests/test_contour.py
@@ -282,6 +282,46 @@ class TestEdgeCases:
         with pytest.raises(TypeError, match="xarray.DataArray"):
             contours(data, levels=[0.5])
 
+    def test_invalid_return_type_rejected(self):
+        """A bad return_type raises ValueError for a normal raster."""
+        data = _make_ramp(ny=5, nx=6)
+        agg = create_test_raster(data, backend='numpy')
+        with pytest.raises(ValueError, match="Invalid return_type"):
+            contours(agg, levels=[2.5], return_type="bad")
+
+    def test_invalid_return_type_rejected_all_nan(self):
+        """A bad return_type raises even on the all-non-finite path.
+
+        Previously the all-non-finite early return handed back an empty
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Validate return_type before data work in contours()

None.

None. The fix is well placed and the regression tests pin the exact bug.

- `xrspatial/contour.py:691`: the late `else: raise ValueError(...)` at the end of
  the dispatch chain is now unreachable, since a bad `return_type` gets rejected at
  the top. Leaving it is fine as a defensive guard and keeps the diff small, but a
  one-line comment saying so would save the next reader a double-take. Not blocking;
  the PR body already explains the intent.

### What looks good
- Validation moved to the top, right after `_validate_raster`, which is where the
  issue asked for it. The error message matches the existing late branch, so the
  behaviour stays consistent.
- `test_invalid_return_type_rejected_all_nan` is the test that matters here. I
  confirmed it fails on the pre-fix code and passes after.
- Tests cover the normal raster, the all-non-finite path, the degenerate input
  before any data work, and that both valid values still work.
- Backend-agnostic by construction: the check runs on the DataArray before any
  dispatch, so all four backends are covered without backend-specific code.

### Checklist
- Algorithm matches reference: n/a (validation-only change)
- All backends consistent: yes, validation runs before dispatch
- NaN handling correct: yes, the all-non-finite path now raises on a bad return_type
- Edge cases covered: yes
- Dask chunk boundaries: n/a
- No premature materialization: yes, it raises before any data work
- Benchmark: n/a
- Docstrings accurate: yes

brendancol (commented):
## Follow-up review (after dd01faa9)

The one Nit from the first pass is addressed: `contour.py:691` now carries a comment
noting the late `else: raise` is unreachable and kept as a defensive guard. No code
behaviour changed.

No new findings. Full `test_contour.py` still passes (50 tests). Nothing left
outstanding.

PR #2899, opened 2026-06-03
Validate n_levels contract in contours()

### LINKED ISSUE #2895: contours(): n_levels has no validation contract
**Describe the bug**

`contours()` doesn't validate its `n_levels` parameter. The value is only used on the auto-level branch (when `levels` isn't passed), at `xrspatial/contour.py:659`:

    levels = np.linspace(vmin, vmax, n_levels + 2)[1:-1]

Two things go wrong:

- `n_levels=0` or `n_levels=-1` silently returns no contours. `np.linspace(vmin, vmax, 2)[1:-1]` and `np.linspace(vmin, vmax, 1)[1:-1]` give back an empty or degenerate level set, so the caller gets an empty result with no hint that the argument was bad.
- `n_levels=2.5` (or any non-integer) raises a raw `TypeError` from inside numpy (`'float' object cannot be interpreted as an integer`), which leaks an implementation detail instead of pointing at the parameter.

**Expected behavior**

`n_levels` should have an explicit contract: an integer >= 1. Invalid values raise a clear error up front:

- non-integer (e.g. `2.5`, or a bool) -> `TypeError` that names `n_levels`
- integer < 1 (e.g. `0`, `-1`) -> `ValueError` that names `n_levels`

Since `n_levels` is ignored when explicit `levels` are supplied, the check should only fire on the auto-level branch, so `contours(agg, levels=[...], n_levels=0)` isn't rejected for no reason.

**Additional context**

Convention in the codebase: `TypeError` for wrong type, `ValueError` for out-of-range. See `landforms` (`inner_radius must be >= 1`) in terrain_metrics.py and `max_iterations must be >= 1` in balanced_allocation.py.

brendancol:
Closes #2895

## What changed

- `contours()` now validates `n_levels` on the auto-level branch (when `levels` is not passed). Non-integers (including `bool`) raise `TypeError`; values `< 1` raise `ValueError`. Both messages name the parameter.
- The check sits inside the `if levels is None:` block, so callers passing explicit `levels` are unaffected. `n_levels` is unused for them and is not validated.
- The `n_levels` docstring documents the contract.

Before this, `n_levels=0` or `-1` silently returned no contours, and `n_levels=2.5` leaked a raw numpy `TypeError` ("'float' object cannot be interpreted as an integer").

## Backends

Scalar argument validation in the shared `contours()` entry point. No backend-specific code paths touched, so numpy / cupy / dask+numpy / dask+cupy all behave the same.

## Test plan

- [x] `n_levels=0` and `n_levels=-1` raise `ValueError`
- [x] `n_levels=2.5` and `n_levels=True` raise `TypeError` naming the parameter
- [x] `n_levels=1` produces exactly one level
- [x] explicit `levels` with `n_levels=0` still works (validation skipped)
- [x] full `test_contour.py` suite passes (52 tests)

### DIFF
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -586,7 +586,9 @@ def contours(
 
     n_levels : int, default 10
         Number of contour levels to generate when ``levels`` is not
-        provided.
+        provided.  Must be an integer >= 1; otherwise a ``TypeError``
+        (non-integer) or ``ValueError`` (< 1) is raised.  Ignored when
+        ``levels`` is given, in which case it is not validated.
 
     return_type : str, default "numpy"
         Output format.  ``"numpy"`` returns a list of ``(level, coords)``
@@ -623,6 +625,17 @@ def contours(
 
     # Determine contour levels.
     if levels is None:
+        # n_levels is only consumed on this auto-level branch, so validate
+        # it here.  When explicit levels are supplied n_levels is ignored,
+        # and rejecting it then would be surprising.  Reject bools too:
+        # bool is an int subclass, but True/False as a level count is a bug.
+        if isinstance(n_levels, bool) or not isinstance(n_levels, (int, np.integer)):
+            raise TypeError(
+                f"n_levels must be an integer, got {type(n_levels).__name__}"
+            )
+        if n_levels < 1:
+            raise ValueError(f"n_levels must be >= 1, got {n_levels}")
+
         # Reduce over finite values only.  +/-inf cells would otherwise
         # poison the range and make np.linspace emit non-finite levels,
         # silently dropping every contour for the finite terrain.  An
--- a/xrspatial/tests/test_contour.py
+++ b/xrspatial/tests/test_contour.py
@@ -222,6 +222,54 @@ class TestAutoLevelsInf:
 
 
 # ---------------------------------------------------------------------------
+# n_levels validation contract (issue #2895)
+# ---------------------------------------------------------------------------
+
+class TestNLevelsValidation:
+
+    def test_n_levels_zero_raises(self):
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Validate n_levels contract in contours()

None.

None.

- Pre-existing, not from this PR: `xrspatial/tests/test_contour.py:5` imports `_contours_numpy`, which flake8 flags as F401 (unused). It is already unused on `main`, so it is out of scope for this change. Flagging it so it does not get attributed to this PR.

### What looks good
- Validation lives inside the `if levels is None:` block (contour.py:627-636), so it only fires when n_levels is actually consumed. Passing an explicit `levels` with a junk `n_levels` is not rejected, which matches the documented contract and is covered by `test_explicit_levels_skip_n_levels_validation`.
- bool is rejected explicitly even though it is an int subclass (contour.py:632). `np.bool_` also lands in the TypeError branch since it is neither `bool` nor `np.integer`.
- The type check runs before the `< 1` comparison, so the range check never sees a non-int.
- TypeError for wrong type and ValueError for out-of-range matches the convention in `terrain_metrics.landforms` and `balanced_allocation`.
- Tests cover 0, -1, 2.5, True, the valid n_levels=1 single-level case, and the explicit-levels skip path.
- Added line is 85 chars, under the repo's 100-char flake8 limit, so no wrap needed.

### Checklist
- [x] Algorithm matches reference/paper (n/a, validation only)
- [x] All implemented backends produce consistent results (shared entry point, no backend code touched)
- [x] No premature materialization or unnecessary copies (scalar checks only)
- [x] README feature matrix updated (contours already listed; no change needed)

PR #2900, opened 2026-06-03
Resolve contours output CRS the same way as polygonize (#2893)

### LINKED ISSUE #2893: contours and polygonize disagree on CRS resolution from the same raster
**Describe the bug**

`contours(..., return_type='geopandas')` resolves the output CRS only from `agg.attrs['crs']`. Both CRS sites in `xrspatial/contour.py` use `agg.attrs.get('crs', None)`: the all-non-finite early return (line 656) and the main geopandas return (line 688). If a raster carries its georeferencing under `attrs['crs_wkt']` or via `raster.rio.crs` instead, contours drops it and returns an unprojected GeoDataFrame.

`polygonize`, the other raster-to-vector function, already has a fuller resolver: `_detect_raster_crs` in `xrspatial/polygonize.py` (line 793). Its order is:

1. `attrs['crs']`
2. `attrs['crs_wkt']`
3. `raster.rio.crs`
4. `None`

So the two entry points disagree on how they pick up georeferencing from the same input. A raster that polygonizes with a CRS can come out of contours without one.

**Expected behavior**

`contours` should resolve the output CRS the same way `polygonize` does, so both georeference identically from the same raster. There's no import cycle between the two modules, so the fix can reuse `_detect_raster_crs` directly at both CRS sites.

**Additional context**

Found during a code review of the contours module.

brendancol:
Closes #2893

`contours(..., return_type='geopandas')` read the output CRS only from `agg.attrs['crs']`. The other raster-to-vector function, `polygonize`, uses a fuller resolver, `_detect_raster_crs` (attrs['crs'] -> attrs['crs_wkt'] -> raster.rio.crs -> None). So a raster georeferenced via `crs_wkt` or `rio.crs` would polygonize with a CRS but contour without one.

Now both functions resolve georeferencing the same way:

- contours reuses `polygonize._detect_raster_crs` at both CRS sites in `contour.py` (the all-non-finite early return and the main geopandas return). No import cycle: polygonize does not import contour.
- polygonize is unchanged.

Backend coverage: CRS resolution is backend-independent. The geopandas return path is numpy-only (it builds shapely geometries on the host), so the tests run on numpy.

Test plan:
- crs_wkt-only raster now yields a GeoDataFrame with its CRS set (was None before).
- attrs['crs'] still takes precedence over attrs['crs_wkt'].
- A raster with no CRS info yields crs None.
- Parity: contours(...).crs matches what _detect_raster_crs returns for the same raster.
- Full test_contour.py suite passes (52 tests).

### DIFF
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -14,6 +14,7 @@ from typing import TYPE_CHECKING, List, Optional, Sequence, Tuple, Union
 import numpy as np
 import xarray as xr
 
+from .polygonize import _detect_raster_crs
 from .utils import ArrayTypeFunctionMapping, _validate_raster, ngjit
 
 if TYPE_CHECKING:
@@ -653,7 +654,7 @@ def contours(
         if not np.isfinite(vmin) or not np.isfinite(vmax):
             if return_type == "numpy":
                 return []
-            return _to_geopandas([], crs=agg.attrs.get('crs', None))
+            return _to_geopandas([], crs=_detect_raster_crs(agg))
 
         # Exclude exact min/max to avoid tracing along the boundary.
         levels = np.linspace(vmin, vmax, n_levels + 2)[1:-1]
@@ -685,7 +686,7 @@ def contours(
     if return_type == "numpy":
         return results
     elif return_type == "geopandas":
-        crs = agg.attrs.get('crs', None)
+        crs = _detect_raster_crs(agg)
         return _to_geopandas(results, crs=crs)
     else:
         raise ValueError(
--- a/xrspatial/tests/test_contour.py
+++ b/xrspatial/tests/test_contour.py
@@ -751,23 +751,25 @@ class TestCRSPropagation:
         assert isinstance(gdf, gpd.GeoDataFrame)
         assert gdf.crs is None
 
-    @pytest.mark.parametrize("attrs", [
-        {'crs': 'EPSG:5070'},
-        {'crs_wkt': None},  # filled in below with a real WKT
-        {},
+    @staticmethod
+    def _wkt_4326():
+        from pyproj import CRS
+        return {'crs_wkt': CRS.from_epsg(4326).to_wkt()}
+
+    @pytest.mark.parametrize("attrs_factory", [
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Resolve contours output CRS the same way as polygonize (#2893)

None.

None.

- test_contour.py: the `{'crs_wkt': None}` entry in the `test_geopandas_crs_matches_detect_raster_crs` parametrize list is a placeholder that the test body overwrites with a real WKT. It works, and the comment explains it, but you have to read both the decorator and the body to see what actually runs, and the pytest id shows `attrs1` instead of something readable. Either build the WKT case directly or use `pytest.param(..., id="crs_wkt")`.

### What looks good
- Surgical fix: one import plus two one-line call-site swaps, both pointing at the resolver polygonize already uses. Since polygonize is untouched, the two functions stay in sync by sharing code, not by keeping a copied constant aligned.
- Both CRS sites are covered, including the easy-to-miss all-non-finite early return at line 657.
- `_detect_raster_crs` only reads attrs and rio.crs, so it never touches array data. That keeps it safe for numpy, cupy, and dask inputs even though the geopandas path is host-only.
- Tests assert the actual behavior change (crs_wkt-only used to drop to None), precedence, the no-CRS case, and parity against `_detect_raster_crs` directly.

### Checklist
- [x] No import cycle (polygonize does not import contour; verified by import).
- [x] Both CRS resolution sites updated.
- [x] crs_wkt precedence and no-CRS cases tested.
- [x] Parity with polygonize asserted directly.
- [x] No backend dispatch changes; CRS resolution is backend-independent.
- [x] No docstring or README changes needed (pure bug fix, no API change).

brendancol (commented):
## Follow-up review (after ac954ddd)

The one nit from the first pass is fixed. The parametrized parity test now uses per-case factory lambdas with explicit `pytest.param` ids (`crs` / `crs_wkt` / `no_crs`), so the cases read clearly and no longer rely on a `None` placeholder that the body overwrote. The crs_wkt WKT is built lazily in a staticmethod, so collection still works without pyproj installed.

No new findings. All 52 tests in test_contour.py pass locally, including the 8 in TestCRSPropagation.

PR #2901, opened 2026-06-03
Validate planar cell size in slope() (#2897)

### LINKED ISSUE #2897: slope() planar method does not validate cell size
## Describe the bug

The planar `slope()` path passes cell sizes straight into divisions in the slope kernel without validating them first. Bad resolutions produce confusing results:

- Zero resolution raises a raw `ZeroDivisionError`, which is cryptic and doesn't point at the cause.
- NaN resolution silently returns NaN across the interior, with no hint that the cell size was the problem.
- Inf resolution yields wrong values that look plausible, with no warning.

`curvature()` already guards against this. It reads the resolution, checks that both axes are non-zero and finite, and raises a clear `ValueError` naming the bad values. `slope()` should do the same.

## To Reproduce

```python
import numpy as np
import xarray as xr
from xrspatial import slope

data = np.zeros((5, 5), dtype=np.float32)
agg = xr.DataArray(data, attrs={'res': (0, 0)})
slope(agg)  # ZeroDivisionError instead of a clear message
```

## Expected behavior

`slope()` should raise a clear `ValueError` for zero, negative, NaN, or infinite cell sizes on either axis, matching the existing `curvature()` guard. The message should name the offending values and say how to fix the resolution.

## Additional context

The check belongs in the public `slope()` wrapper so it applies uniformly across all four backends (numpy, cupy, dask+numpy, dask+cupy) on the planar path. The geodesic path takes its scale from lat/lon coordinates and is unaffected.

brendancol:
## Summary

- Add cell-size validation to the planar `slope()` path, matching the guard already in `curvature()`.
- `slope(method='planar')` now raises a clear `ValueError` when the resolution on either axis is zero, negative, NaN, or infinite. Before this, zero gave a raw `ZeroDivisionError`, NaN gave a silently-NaN interior, and inf gave wrong values that looked plausible.
- The check sits in the public wrapper before backend dispatch, so it behaves the same on numpy, cupy, dask+numpy, and dask+cupy. The geodesic path is untouched since it takes its scale from lat/lon coordinates, not the `res` attribute.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy (planar). Validation runs in the wrapper, so all four behave identically.

## Test plan

- [x] Parametrized tests assert the `ValueError` for zero/negative/NaN/inf cell sizes across all four backends.
- [x] Regression test confirms a normal positive cell size still works.
- [x] Test confirms the geodesic path ignores a bogus `res` attribute and does not trip the planar guard.
- [x] `test_slope.py` and `test_geodesic_slope.py` pass (153 tests).

Closes #2897

### DIFF
--- a/xrspatial/slope.py
+++ b/xrspatial/slope.py
@@ -403,6 +403,8 @@ def slope(agg: xr.DataArray,
     if method == 'planar':
         warn_if_unit_mismatch(agg)
         cellsize_x, cellsize_y = get_dataarray_resolution(agg)
+        # Reject negatives too (curvature() only checks == 0): a negative cell
+        # size would silently flip the slope sign rather than error out.
         if (not np.isfinite(cellsize_x) or not np.isfinite(cellsize_y)
                 or cellsize_x <= 0 or cellsize_y <= 0):
             raise ValueError(
--- a/xrspatial/slope.py
+++ b/xrspatial/slope.py
@@ -360,7 +360,9 @@ def slope(agg: xr.DataArray,
     size. If the coordinates are in degrees (lat/lon) but the elevation values
     are in meters, the result is wrong by orders of magnitude. When this
     mismatch is detected, a ``UserWarning`` is emitted suggesting you reproject
-    to a projected CRS or use ``method='geodesic'``.
+    to a projected CRS or use ``method='geodesic'``. The ``'planar'`` method also
+    raises a ``ValueError`` if the cell size on either axis is zero, negative, or
+    non-finite, since those values cannot produce a valid slope.
 
     References
     ----------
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Validate planar cell size in slope() (#2897)

None.

None.

- slope.py:406-414 The guard uses `cellsize_x <= 0` while the sibling guard in curvature.py:256-263 uses `cellsize_x == 0`. The divergence is deliberate: issue #2897 asks slope to reject negative resolutions too, and a negative cell size would silently flip the slope sign. A one-line comment would stop a future reader from "harmonizing" the two guards back to `== 0`.
- test_slope.py:483-520 The four backend-specific parametrized tests are near-identical. Validation runs in the wrapper before dispatch, so the numpy test alone already covers the code path. The cupy/dask variants are cheap insurance that no backend swallows the error, so keeping them is fine; they could be collapsed if the file grows.

### What looks good
- The check sits in the public wrapper before `ArrayTypeFunctionMapping` dispatch, so all four backends behave the same without per-backend code. Matches the curvature.py precedent.
- NaN/inf ordering is correct: `not np.isfinite(...)` is evaluated first via `or`, so NaN never reaches the `<= 0` comparison (which is False for NaN and would let it through).
- The geodesic path is left untouched, with an explicit test confirming a bogus `res` attribute does not trip the planar guard.
- The error message names the offending values and says how to fix the resolution, in the curvature.py message style.
- Tests hit zero, negative, NaN, and inf on each axis independently, plus a positive-cellsize regression test.

### Checklist
- [x] Algorithm matches reference/paper (no algorithm change; validation only)
- [x] All implemented backends produce consistent results (validation is backend-agnostic, runs in wrapper)
- [x] NaN handling is correct
- [x] Dask chunk boundaries handled correctly (unchanged; guard runs before dispatch)
- [x] Benchmark exists or is not needed (validation guard, no perf impact)
- [ ] README feature matrix updated (not applicable; no new function or backend change)

brendancol (commented):
## Follow-up review (after f1a5eab7)

The single actionable nit from the previous pass is resolved: slope.py now carries a one-line comment explaining why the guard uses `<= 0` (rejecting negatives) where curvature() uses `== 0`. No new code paths, so the earlier checklist still holds.

The second nit (collapsing the four near-identical backend tests) is left as-is on purpose. The cupy and dask variants are cheap and confirm no backend wrapper swallows the error before it reaches the user. Collapsing them would trade that coverage for a few saved lines, which is not worth it here.

No remaining blockers or suggestions.

PR #2902, opened 2026-06-03
Drop zero-length contour geometry at exact-level corners

### LINKED ISSUE #2892: contours() emits zero-length, invalid LineStrings when corners equal the contour level
**Describe the bug**

`contours()` can emit zero-length, invalid LineString geometries when raster corner values land exactly on the contour level.

Corners are classified with `>= level` (`xrspatial/contour.py` lines 77-84), and `_emit_seg` (lines 163-209) interpolates edge crossings even when the "crossing" is really an edge or corner exactly equal to the level. When a corner equals the level, the interpolation parameter `t` lands on the corner, so both endpoints of a segment collapse to the same point. The result is a zero-length segment.

**Reproduce**

A 4x4 checkerboard of 0 and 1 at `level=1.0` produces 8 zero-length LineStrings such as `LINESTRING (0 1, 0 1, 0 1)`. Each has length 0 and `is_valid == False` in Shapely, and they propagate into the GeoDataFrame from `return_type="geopandas"`.

```python
import numpy as np

data = np.array([[0., 1., 0., 1.], [1., 0., 1., 0.], [0., 1., 0., 1.], [1., 0., 1., 0.]])
gdf = contours(agg, levels=[1.0], return_type='geopandas')
for g in gdf.geometry:
    print(g.is_valid, g.length, g.wkt)
```

**Expected behavior**

No zero-length or invalid geometry should reach the output. Degenerate segments (zero-length, or collapsed to a single distinct point) and degenerate polylines should be dropped before stitching and GeoDataFrame output, and equality with the level should be handled the same way every time. This should hold for all four backends (numpy, cupy, dask+numpy, dask+cupy).

**Additional context**

The `numpy`, `cupy`, and `dask` paths all feed through `_emit_seg`, and the results become LineStrings in `_to_geopandas`.

brendancol:
Closes #2892

## Problem

`contours()` classifies corners with `>= level`, so a corner exactly equal to the level is treated as above it. When that happens the marching-squares interpolation lands the crossing on the corner, and both endpoints of a segment can collapse onto the same point. The result is a zero-length segment that becomes an invalid LineString (`is_valid == False`, length 0) once it reaches the GeoDataFrame.

A 4x4 checkerboard of 0 and 1 at `level=1.0` produced 8 such zero-length geometries.

## Fix

Filter degenerate geometry in the backend-agnostic stitching/output layer, so both `return_type="numpy"` and `return_type="geopandas"` stay clean:

- `_emit_seg` skips writing a segment when its two endpoints are identical (zero length).
- `_stitch_segments` drops any polyline that collapses to a single distinct point.
- The policy is documented in the `contours()` docstring.

Corners are still classified with `>= level` (equality counts as above), and that rule is applied the same way every time. Genuine mid-edge crossings are untouched.

## Backends

The fix lives in shared Python helpers that all four backends route through (`_emit_seg`, `_stitch_segments`), so numpy, cupy, dask+numpy, and dask+cupy get the same behavior.

## Test plan

- [x] Checkerboard at `level=1.0` produces no zero-length segments and no invalid LineStrings (numpy and dask)
- [x] No repeated consecutive points in emitted geometries
- [x] Equality semantics consistent across checkerboard phase variants
- [x] Genuine ramp crossing still produces valid contours (regression guard)
- [x] Existing contour tests still pass (53 total, 7 new)

### DIFF
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -209,7 +209,10 @@ def _emit_seg(r, c, tl, tr, bl, br, level, edge_a, edge_b,
     # happens the interpolation parameter lands on the corner and both
     # endpoints can collapse onto the same point, producing a zero-length
     # segment.  Drop those here so no zero-length / single-point geometry
-    # reaches stitching or GeoDataFrame output.
+    # reaches stitching or GeoDataFrame output.  Exact float equality is
+    # intended: a real collapse computes both endpoints from the same corner
+    # value and yields bit-identical coordinates, while near-equal endpoints
+    # are genuine short segments and must be kept.
     if r0 == r1 and c0 == c1:
         return
 
@@ -571,7 +574,10 @@ def _to_geopandas(results, crs=None):
 
     records = []
     for level, coords in results:
-        if len(coords) >= 2:
+        # Require at least two distinct vertices.  _stitch_segments already
+        # drops single-point polylines, but guard here too so the geopandas
+        # path never emits a zero-length / invalid LineString on its own.
+        if _has_distinct_points(coords, 10):
             # coords are (row, col); convert to (x, y) = (col, row)
             geom = LineString(coords[:, ::-1])
             records.append({'level': level, 'geometry': geom})
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -204,6 +204,15 @@ def _emit_seg(r, c, tl, tr, bl, br, level, edge_a, edge_b,
         r1 = float(r) + t
         c1 = float(c)
 
+    # Degenerate-segment policy: corners are classified with ``>= level``,
+    # so a corner exactly equal to the level is treated as above.  When that
+    # happens the interpolation parameter lands on the corner and both
+    # endpoints can collapse onto the same point, producing a zero-length
+    # segment.  Drop those here so no zero-length / single-point geometry
+    # reaches stitching or GeoDataFrame output.
+    if r0 == r1 and c0 == c1:
+        return
+
     seg_rows[idx, 0] = r0
     seg_rows[idx, 1] = r1
     seg_cols[idx, 0] = c0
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Drop zero-length contour geometry at exact-level corners

None.

- `xrspatial/contour.py` `_to_geopandas` (around line 547): the geopandas builder still only guards `len(coords) >= 2`. After this PR the upstream `_stitch_segments` filter removes single-point polylines, so nothing degenerate reaches it in practice. But the output layer has no independent guard, so a future change to the stitching path could let a zero-length 2-point line through again. A one-line distinct-point check here would make the geopandas path self-defending. Optional, since the upstream filter covers the current code.

- `_emit_seg` uses exact float equality (`r0 == r1 and c0 == c1`). This is correct for the real collapse case, because both endpoints are computed from the same corner values and come out bit-identical. A one-line comment noting that exact equality is intended (near-equal points are genuine short segments) would save the next reader a double-take.

### What looks good
- The fix sits in shared helpers (`_emit_seg`, `_stitch_segments`) that all four backends route through, so numpy, cupy, dask+numpy, and dask+cupy get identical behavior without per-backend code.
- `_has_distinct_points` reuses the existing `DECIMALS=10` rounding tolerance from the stitching logic, so the distinct-point test matches how endpoints are already compared.
- The tests are hard, not forgiving. They assert no zero-length segments, no invalid LineStrings, and no repeated consecutive points, and they were confirmed to fail without the fix (6 of 7 fail on the pre-fix code; the 7th is the genuine-crossing regression guard).
- The `>= level` equality rule is documented in the docstring and kept consistent.

### Checklist
- [x] Algorithm matches reference (marching squares unchanged; only degenerate output filtered)
- [x] All four backends produce consistent results (shared helpers)
- [x] NaN handling unchanged (non-finite quads still skipped upstream)
- [x] Edge cases covered (checkerboard, phase variants, genuine crossing)
- [x] Dask chunk boundaries handled (filter runs in both per-chunk and re-stitch paths)
- [x] README feature matrix unchanged (no new function, backends unchanged)
- [x] Docstring updated

brendancol (commented):
## Follow-up review (after commit 561fdaa6)

### Suggestion (resolved)
`_to_geopandas` now calls `_has_distinct_points(coords, 10)` instead of the bare `len(coords) >= 2` check, so the geopandas path drops zero-length / single-point lines on its own rather than relying only on the upstream stitch filter.

### Nit (resolved)
The `_emit_seg` zero-length guard now carries a comment explaining that exact float equality is intentional: a real collapse yields bit-identical coordinates, while near-equal endpoints are genuine short segments and are kept.

### New findings
None. The follow-up is two small, contained edits. All 53 contour tests pass (7 new for this issue).

### Blockers
None.

PR #2903, opened 2026-06-03
Fix geodesic coord resolution to prefer real lat/lon over numeric y/x

### LINKED ISSUE #2896: Geodesic slope/aspect silently use numeric y/x index coords instead of real 2-D lat/lon
## Description

For a curvilinear raster with dims ('y', 'x'), numeric y/x index coordinates, and real 2-D latitude/longitude coordinates, the geodesic method silently uses the numeric y/x index coords as if they were lat/lon. The user gets wrong geodesic results with no warning.

The root cause is in `_find_coord()` in `xrspatial/utils.py`. It first accepts the dimension coordinate if it is numeric, before it ever scans for a coordinate explicitly named latitude or longitude. When the dimension coord is a numeric y/x pixel index, that index wins over the real 2-D lat/lon coords.

The range validation does not catch this: small pixel indices (0..N) fall inside the accepted geographic ranges of [-90, 90] for latitude and [-180, 360] for longitude, so no error is raised.

## Reproduction

A 6x8 east-tilted surface over a real 40-41N / 10-11E grid:
- Curvilinear raster (dims y/x with numeric y/x coords plus 2-D lat/lon coords): center-cell slope = 0.0074
- Reference raster (same grid, dims lat/lon with 1-D lat/lon coords): center-cell slope = 0.0675

Same physical surface, ~9x difference, no warning.

## Expected behavior

When real latitude/longitude coordinates exist on the DataArray, the geodesic path should prefer them over numeric y/x dimension index coords.

## Affected

slope, aspect, and surface_distance geodesic paths all resolve coords through `_extract_latlon_coords` / `_find_coord`, so all are affected. The fix is backend-agnostic (numpy, cupy, dask+numpy, dask+cupy).

Test gap: existing geodesic tests cover 1-D lat/lon coords and backend parity but not the dims=('y','x') + 2-D lat/lon + numeric y/x case where this bug lives.

brendancol:
Closes #2896

## Problem

`_find_coord()` in `xrspatial/utils.py` accepted the numeric dimension coordinate before it scanned for a coordinate named latitude/longitude. For a curvilinear raster with dims ('y', 'x'), numeric y/x pixel-index coords, and real 2-D lat/lon coords, the geodesic path used the pixel indices as lat/lon. The range check did not catch it because small indices (0..N) fall inside the accepted geographic ranges. The result was a wrong slope (~0.007 vs ~0.067 on the reproduction grid) with no warning.

## Change

- `_find_coord()` now prefers an explicitly named lat/lon coord (`lat`/`latitude`/`lon`/`longitude`) over the dimension coord. The dimension coord and the broader `y`/`x` name set remain as fallbacks, so rasters with y/x as the geographic coords still work.
- This fixes slope, aspect, and surface_distance, which share `_extract_latlon_coords` / `_find_coord`. The logic is coord resolution only, so it is backend-agnostic.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy. Coord resolution runs on the host before dispatch; the new curvilinear tests assert parity across all four (cupy / dask+cupy skip without a GPU).

## Test plan

- [x] Repro from the issue now matches the 1-D lat/lon reference exactly
- [x] New `TestGeodesicSlopeCurvilinear*` classes: dims y/x + 2-D lat/lon + numeric y/x, asserting correct slope and cross-backend parity
- [x] Existing geodesic slope/aspect and surface_distance suites still green (including `test_projected_coords_raises`, which relies on the y/x fallback)

### DIFF
--- a/xrspatial/tests/test_geodesic_slope.py
+++ b/xrspatial/tests/test_geodesic_slope.py
@@ -41,6 +41,40 @@ def _make_geo_raster(elev, lat_start, lat_end, lon_start, lon_end,
     return raster
 
 
+def _make_curvilinear_raster(elev, lat_start, lat_end, lon_start, lon_end,
+                             backend='numpy', chunks=(3, 3)):
+    """Build a curvilinear DataArray: dims ('y', 'x') with numeric y/x index
+    coords AND real 2-D lat/lon coords over the same geographic grid.
+
+    This is the layout that exposed the coordinate-resolution bug: the numeric
+    y/x pixel-index coords must not be used as lat/lon when real lat/lon coords
+    are present.
+    """
+    H, W = elev.shape
+    lat1d = np.linspace(lat_start, lat_end, H)
+    lon1d = np.linspace(lon_start, lon_end, W)
+    lon2d, lat2d = np.meshgrid(lon1d, lat1d)
+    raster = xr.DataArray(
+        elev.astype(np.float64),
+        dims=['y', 'x'],
+        coords={
+            'y': np.arange(H, dtype=np.float64),
+            'x': np.arange(W, dtype=np.float64),
+            'lat': (('y', 'x'), lat2d),
+            'lon': (('y', 'x'), lon2d),
+        },
+    )
+
+    if 'cupy' in backend:
+        import cupy
+        raster.data = cupy.asarray(raster.data)
+
+    if 'dask' in backend and da is not None:
+        raster.data = da.from_array(raster.data, chunks=chunks)
+
+    return raster
+
+
 def _flat_surface(H=6, W=8, elev=500.0):
     """Constant-elevation surface — slope should be 0 everywhere interior."""
     return np.full((H, W), elev, dtype=np.float64)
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Fix geodesic coord resolution to prefer real lat/lon over numeric y/x

None.

None.

- xrspatial/utils.py:967-972: The explicit-name scan picks the first matching coord in `agg.coords` iteration order. If a DataArray carries both a `lat` and a `latitude` coord (or `lon`/`longitude`) with different dimensionality, the winner depends on coord insertion order, and a 1-D winner on one axis paired with a 2-D winner on the other would trip the "both 1-D or both 2-D" error in `_extract_latlon_coords`. This is an unusual input and the resulting error is clear, so it does not need handling now, but a one-line docstring note that duplicate explicit names resolve by coord order would help a future reader.

### What looks good
- Root cause fixed at the right layer: reordering `_find_coord` precedence fixes slope, aspect, and surface_distance at once, since they all route through `_extract_latlon_coords`.
- The `y`/`x` fallback is preserved, so rasters that legitimately use y/x as geographic coords still resolve (confirmed by the still-passing `test_projected_coords_raises`).
- Tests assert against the 1-D lat/lon reference rather than just "runs without error", and the `> 0.05` interior check would catch a regression back to the pixel-index behavior (~0.007).
- Backend parity covered across numpy, cupy, dask+numpy, dask+cupy via the general_checks markers.

### Checklist
- Algorithm matches reference: n/a (coord resolution, not a numeric kernel)
- All implemented backends produce consistent results: yes, parity tests added
- NaN handling correct: unchanged by this PR
- Edge cases covered by tests: curvilinear path covered; duplicate-explicit-name case noted as a nit
- Dask chunk boundaries handled correctly: unchanged; coord resolution is host-side
- No premature materialization: coord `.values` read is host-side and pre-existing
- Benchmark exists or not needed: not needed (no perf-sensitive path changed)
- README feature matrix updated: not applicable (bug fix, no new API)
- Docstrings present and accurate: yes, `_find_coord` docstring updated

brendancol (commented):
## PR Review (follow-up pass)

Re-reviewed after commit 40a00653.

### Blockers
None.

### Suggestions
None.

### Nits
None outstanding. The earlier nit on `_find_coord` (utils.py) is resolved: the docstring now states that when several explicit lat/lon names are present, the first one in coord order wins.

### What looks good
- The fix and tests are unchanged from the first pass and still green: 103 passed across geodesic slope, geodesic aspect, and surface_distance suites locally (cupy / dask+cupy classes skip without a GPU).
- The diff stays scoped to coord resolution plus its tests. No unrelated edits.

Nothing further to address. Recommending merge once CI is green.

PR #2904, opened 2026-06-03
Set slope output units to 'degrees' instead of leaking input units

### LINKED ISSUE #2894: slope() output advertises stale elevation units instead of degrees
**Describe the bug**

`slope()` returns its result in degrees, but it copies the input DataArray's attrs verbatim via `attrs=agg.attrs` (xrspatial/slope.py near line 432). If the input DEM has `attrs={'units': 'm'}`, the slope output values are in degrees while the attrs still say `units='m'`. The metadata no longer matches the data.

This bites because `_infer_vertical_unit_type` in xrspatial/utils.py reads `attrs['units']` and treats `'degrees'` as an angle and `'m'` as elevation. A slope output carrying `units='m'` gets classified as elevation, so downstream diagnostics and users are misled about what the array holds.

**Expected behavior**

The slope output should report the angle unit it actually produces (`units='degrees'`), not the input elevation unit. Other input attrs should carry through unchanged.

**Steps to reproduce**

```python
from xrspatial import slope
agg = xr.DataArray(np.random.rand(5, 5), attrs={'units': 'm'})
out = slope(agg)
print(out.attrs['units'])  # prints 'm', should be 'degrees'
```

**Additional context**

Applies to all four backends (numpy, cupy, dask+numpy, dask+cupy) since the attrs are set in the shared public wrapper. aspect.py and curvature.py copy attrs the same way, but those are out of scope here.

brendancol:
Closes #2894

`slope()` returns degrees but copied the input DataArray's attrs verbatim, so a DEM with `units='m'` produced a slope array still labelled `units='m'`. `_infer_vertical_unit_type` reads that label and treated the slope output as elevation rather than an angle.

## Changes
- Override `attrs['units']` to `'degrees'` in the shared public wrapper, so the fix applies to all backends and to Dataset inputs. All other input attrs carry through unchanged, and the input array is not mutated.
- Update the docstring Returns note to state the units override.
- Add a `verify_attrs` passthrough to the cross-backend equality helpers in `general_checks.py`; the slope tests whose attrs now legitimately differ pass `verify_attrs=False`.
- Add tests asserting the degrees override on numpy, cupy, dask+numpy, dask+cupy, the no-input-units case, and the geodesic path.

## Backend coverage
numpy, cupy, dask+numpy, dask+cupy (attrs are set once in the wrapper).

## Test plan
- [x] `pytest xrspatial/tests/test_slope.py xrspatial/tests/test_geodesic_slope.py` (121 passed)
- [x] Regression check on aspect/curvature/hillshade/terrain_metrics/northness_eastness which share the helpers (565 passed)

### DIFF
--- a/xrspatial/slope.py
+++ b/xrspatial/slope.py
@@ -351,8 +351,8 @@ def slope(agg: xr.DataArray,
         If `agg` is a DataArray, returns a DataArray of the same type.
         If `agg` is a Dataset, returns a Dataset with slope computed
         for each data variable.
-        2D array of slope values.
-        All other input attributes are preserved.
+        2D array of slope values in degrees. The output ``attrs['units']``
+        is set to ``'degrees'``; all other input attributes are preserved.
 
     Notes
     -----
@@ -429,11 +429,17 @@ def slope(agg: xr.DataArray,
         )
         out = mapper(agg)(agg.data, lat_2d, lon_2d, WGS84_A2, WGS84_B2, z_factor, boundary)
 
+    # slope is reported in degrees, so override the input elevation 'units'
+    # (e.g. 'm') with the angle unit the result actually carries. Other input
+    # attrs are preserved.
+    attrs = dict(agg.attrs)
+    attrs['units'] = 'degrees'
+
     result = xr.DataArray(out,
                           name=name,
                           coords=agg.coords,
                           dims=agg.dims,
-                          attrs=agg.attrs)
+                          attrs=attrs)
     # On dask backends, xr.DataArray keeps the dask array's internal graph
     # token as .name when name=None, so reset it post-construction to match
     # the numpy/cupy backends. (Same fix as zonal #2611, focal #2733.)
--- a/xrspatial/tests/general_checks.py
+++ b/xrspatial/tests/general_checks.py
@@ -174,34 +174,36 @@ def assert_boundary_mode_correctness(numpy_agg, dask_agg, func, depth=1, rtol=1e
             )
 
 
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Set slope output units to 'degrees' instead of leaking input units

None.

- xrspatial/tests/test_slope.py: flipping the existing tests to `verify_attrs=False` means they no longer confirm that the other input attrs (`res`, `crs`) survive `slope()`. Only `test_units_overridden_to_degrees_numpy` checks `res`/`crs` propagation, and only on numpy. Worth having that numpy test assert the non-units attrs explicitly so the propagation contract stays covered now that the blanket attrs check is off.

- None.

### What looks good
- The override lives in the single shared wrapper, so all four backends and the Dataset path inherit it without per-backend edits. numpy/cupy/dask+numpy/dask+cupy and Dataset all report 'degrees'.
- `dict(agg.attrs)` copies before mutating, so the input DataArray's attrs aren't touched; a test asserts the input still has `units='m'`.
- The geodesic path is covered.
- The `verify_attrs` passthrough on the shared equality helpers defaults to True, so no other module's tests change behavior.

### Checklist
- [x] slope is degrees, units now say degrees
- [x] NaN handling unaffected by this change
- [x] Edge cases covered (no-input-units case, geodesic path)
- [x] No premature materialization or extra copies (one dict copy of attrs)
- [x] Benchmark not needed
- [x] Docstring updated to note the units override

brendancol (commented):
## Follow-up review

The one suggestion from the previous pass is addressed: `test_units_overridden_to_degrees_dask_numpy` now asserts `res` and `crs` survive `slope()` on the dask path, in addition to the numpy test. No blockers, suggestions, or nits remain.

PR #2909, opened 2026-06-03
Fix bounded dask proximity dropping in-range targets on irregular coords

### LINKED ISSUE #2908: Bounded dask proximity drops in-range targets on irregular coordinates
**Describe the bug**

On the bounded-distance dask path of proximity/allocation/direction (finite `max_distance`, dask+numpy backend), in-range cells turn into NaN when the raster has irregularly spaced coordinates and is chunked along the column axis.

The cause is in `_process_dask` (xrspatial/proximity.py). It captures `data = raster.data` before sizing the halo, then calls `_fit_halo_to_chunks`, which can fold an axis into a single chunk and drop that axis's halo depth to zero. The folded result gets assigned back to `raster.data`, `xs`, and `ys` instead of to the local `data`. Two things break:

1. `map_overlap` is still handed the stale, un-folded `data` while `xs`/`ys` are the folded versions, so data and coordinate grids disagree on chunking. Pixels whose nearest target sits in a neighbouring column chunk lose that target and come out NaN even though it is well inside `max_distance`.
2. Assigning to `raster.data` mutates the caller's input DataArray. That is the mutation issue #2847 was meant to prevent, and the comment right above the line says not to do it.

**Expected behavior**

The bounded dask result matches the numpy reference for irregularly spaced coordinates, and the caller's DataArray chunking is left unchanged.

**How to reproduce**

`coords = [0, 100, 101, 102, 103, 104, 105, 106]` on both axes, a single target at `data[3, 3]`, `max_distance=2.5`, dask `chunks=(4, 2)`. The numpy result keeps the in-range targets; the dask result drops the ones in adjacent column chunks.

A regression test (`test_bounded_dask_irregular_coords_matches_numpy`) already exists and fails for all three of proximity, allocation, and direction.

**Additional context**

The EUCLIDEAN and MANHATTAN bounded paths are affected; the great-circle and unbounded paths run different code. The dask+cupy path (`_process_dask_cupy`) shares the same fold logic, so its handling should be checked too.

brendancol:
Closes #2908

## What

- The bounded dask path of proximity/allocation/direction folded the halo with `_fit_halo_to_chunks` but assigned the folded array back to `raster.data` instead of the local `data` that `map_overlap` actually reads. The call then mixed a stale, un-folded data array with folded `xs`/`ys`, so chunking disagreed and in-range targets in adjacent column chunks dropped to NaN on irregularly spaced coordinates. The same assignment mutated the caller's input DataArray, the mutation issue #2847 already guards against.
- Assign the fold into the local `data`, matching the unbounded branch right above it. One line of behaviour change plus an explanatory comment.

## Backend coverage

- dask+numpy: fixed (this is where the bug lived).
- dask+cupy: already correct; `_process_dask_cupy` folds into a local `raster_data` and passes that to `map_overlap`, so it neither mutated the input nor mismatched chunking. No change needed.
- numpy / cupy: not on this code path.

## Test plan

- [x] `test_bounded_dask_irregular_coords_matches_numpy` passes for proximity, allocation, direction (was failing on all three).
- [x] New `test_bounded_dask_does_not_mutate_input` asserts the caller's `.data` is neither rebound nor rechunked; verified it fails on the pre-fix code.
- [x] Full `test_proximity.py` green (410 passed).

No new public API and no backend-support change, so the user-guide notebook and README feature-matrix steps do not apply. CHANGELOG left untouched.

### DIFF
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -1533,8 +1533,14 @@ def _process(
         else:
             pad_y, pad_x = _halo_depth(
                 x_coords, y_coords, max_distance, distance_metric)
-            pad_y, pad_x, (raster.data, xs, ys) = _fit_halo_to_chunks(
-                pad_y, pad_x, raster.data, xs, ys)
+            # Fold into the same local ``data`` the map_overlap call uses, not
+            # ``raster.data``: assigning back to ``raster.data`` would both
+            # mutate the caller's input (issue #2847) and leave map_overlap
+            # reading the stale, un-folded array while xs/ys are folded, so
+            # chunking disagrees and in-range targets in adjacent chunks drop
+            # to NaN (issue #2908).
+            pad_y, pad_x, (data, xs, ys) = _fit_halo_to_chunks(
+                pad_y, pad_x, data, xs, ys)
 
         out = da.map_overlap(
             _process_numpy,
--- a/xrspatial/tests/test_proximity.py
+++ b/xrspatial/tests/test_proximity.py
@@ -1530,6 +1530,35 @@ def test_bounded_dask_irregular_coords_matches_numpy(func):
 
 @pytest.mark.skipif(da is None, reason="dask is not installed")
 @pytest.mark.parametrize("func", [proximity, allocation, direction])
+def test_bounded_dask_does_not_mutate_input(func):
+    """Bounded dask must not rebind or rechunk the caller's .data (issue #2908).
+
+    When the halo is deeper than a chunk, the bounded path folds that axis into
+    a single chunk. The fold used to be assigned back to ``raster.data``, which
+    rebound the caller's DataArray (the same mutation issue #2847 guards
+    against) and left map_overlap reading a stale, un-folded array. Sizing the
+    halo at 3 px against width-2 column chunks forces the column-axis fold.
+    """
+    coords = np.array([0, 100, 101, 102, 103, 104, 105, 106], dtype=float)
+    data = np.zeros((8, 8), dtype=np.float64)
+    data[3, 3] = 1.0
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Fix bounded dask proximity dropping in-range targets on irregular coords

None.

None. The change is one line plus a comment, and it matches what the unbounded branch right above already does: keep the fold in the local `data`, never touch `raster.data`.

None.

### What looks good
- The root cause is right. `_process_dask` captured `data = raster.data`, then `_fit_halo_to_chunks` returned a folded array that got assigned to `raster.data` while `map_overlap` kept reading the stale local `data`. Routing the fold into the local `data` keeps data and xs/ys chunk-aligned.
- It also drops the input-mutation side effect from issue #2847, since `raster.data` is no longer rebound.
- I checked the dask+cupy twin (`_process_dask_cupy`) and it was already correct: it folds into a local `raster_data` and passes that to map_overlap, so it neither mutated the input nor mismatched chunking. Leaving it alone was the right call.
- Test coverage holds up. The existing `test_bounded_dask_irregular_coords_matches_numpy` checks correctness against the numpy reference for all three funcs, and the new `test_bounded_dask_does_not_mutate_input` guards the rebind/rechunk regression. I confirmed the new test fails on the pre-fix code, so it actually catches the bug.
- Full test_proximity.py is green (410 passed).

### Checklist
- [x] dask+numpy now matches the numpy reference; dask+cupy already matched
- [x] In-range cells no longer turn into false NaN
- [x] Edge cases covered (irregular coords; single-row/col tests still pass)
- [x] Dask chunk alignment fixed (the point of the change)
- [x] Benchmark not needed; no perf change
- [x] README matrix not applicable; no new function or backend change
- [x] Comment added explaining which array the fold targets

PR #2922, opened 2026-06-04
Zero kriging-matrix diagonal so the nugget is not placed on it

### LINKED ISSUE #2915: kriging(): nugget lands on the kriging-matrix diagonal (should be gamma(0)=0)
**Describe the bug**

`kriging()` builds the ordinary-kriging matrix in `_build_kriging_matrix` (`xrspatial/interpolate/_kriging.py`) with:

```python
K[:n, :n] = vario_func(D)
```

`D` is the pairwise distance matrix, so its diagonal is 0. Evaluating the variogram model at distance 0 returns the nugget `c0` (for example `_spherical(0, c0, c, a) == c0`), so the nugget ends up on the matrix diagonal.

By definition the semivariogram has gamma(0) = 0: gamma(h) = 0.5 * Var(Z(x+h) - Z(x)), and that difference is identically zero at h = 0. The nugget is the one-sided limit as h approaches 0 from above, not the value at 0. Standard ordinary-kriging implementations such as PyKrige zero the diagonal of the variogram block.

**Effect**

When the fitted nugget is zero the diagonal is already 0 and nothing changes, which is why the current tests pass: their trend-dominated data fits a near-zero nugget. When the nugget is non-zero (real data with measurement noise or microscale variation), putting it on the diagonal:

- forces the predictor to reproduce the data exactly instead of smoothing through the nugget, which is wrong for noisy data
- biases the kriging variance downward, since the diagonal term feeds into the variance via k0^T K_inv k0

**Expected behavior**

The variogram block diagonal should be 0. With a non-zero nugget the predictor should smooth rather than hit data points exactly, and the kriging variance should reflect the nugget.

**Fix**

Zero the diagonal of the variogram block after evaluating the model in `_build_kriging_matrix`.

Found during an accuracy sweep of `xrspatial/interpolate/_kriging.py`.

brendancol:
Closes #2915

`_build_kriging_matrix` set the variogram block to `vario_func(D)`, where `D` has a zero diagonal. `vario_func(0)` returns the nugget `c0`, so a non-zero nugget ended up on the matrix diagonal. The semivariogram is `gamma(0) = 0` by definition (the nugget is the one-sided limit as `h -> 0+`), and standard ordinary-kriging implementations zero that diagonal.

The bug only bit when the fitted nugget was non-zero: it forced exact interpolation of noisy data and shrank the kriging variance. The existing tests use trend-dominated data that fits a near-zero nugget, so they never exercised it.

Changes:
- Zero the variogram-block diagonal with `np.fill_diagonal` in `_build_kriging_matrix`.
- Add two regression tests: the matrix diagonal stays 0 under a non-zero nugget, and a non-zero nugget now smooths predictions and lifts the variance above the nugget floor.

Backends: `_build_kriging_matrix` is shared by all four backends (numpy / cupy / dask+numpy / dask+cupy), which build `K_inv` on the host before dispatch, so the fix applies uniformly.

Test plan:
- [x] `pytest xrspatial/tests/test_interpolation.py -k "Kriging or kriging"` (16 passed, includes cupy and dask+cupy parity)
- [x] New tests fail on the pre-fix code and pass after the fix


brendancol (commented):
## PR Review: Zero kriging-matrix diagonal so the nugget is not placed on it

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The fix matches the standard ordinary-kriging formulation: the semivariogram is gamma(0)=0, and the nugget is the one-sided limit as h -> 0+, so it should not sit on the matrix diagonal. PyKrige does the same.
- The change is minimal and well-commented, touching only `_build_kriging_matrix`.
- `_build_kriging_matrix` runs in numpy on the host before backend dispatch, so all four backends (numpy / cupy / dask+numpy / dask+cupy) consume the corrected `K_inv`. The fix is uniform without per-backend edits.
- The existing post-failure regularizer (`K[:n,:n] += 1e-10 * np.eye(n)`) is a separate Tikhonov term applied only when `inv` raises; it is unaffected and still correct.
- Both new tests were confirmed to fail on the pre-fix code and pass after, so they are genuine regression guards rather than tautologies.

### Notes
- The new tests exercise `_build_kriging_matrix` / `_kriging_predict` directly in numpy rather than across all four backends. That is appropriate here: the corrected line is shared host code, and the existing `test_cupy_*` / `test_dask_cupy_*` parity tests already cover dispatch. Adding nugget-specific backend variants would not increase coverage of the fix.

### Checklist
- [x] Algorithm matches reference (ordinary kriging, gamma(0)=0)
- [x] All implemented backends consume the corrected matrix
- [x] NaN handling unchanged
- [x] Regression tests fail before / pass after the fix
- [x] Benchmark not needed (bug fix, no perf-sensitive change)
- [x] README feature matrix unchanged (no new API)
- [x] Docstrings/comments present and accurate

PR #2924, opened 2026-06-04
Guard single-point input in kriging() variogram

### LINKED ISSUE #2917: kriging() raises an opaque numpy error for a single input point
## Description

Calling `kriging()` with a single valid input point crashes with a confusing low-level numpy error instead of either producing a result or raising a clear xrspatial error.

```python

tmpl = xr.DataArray(np.zeros((2, 2)), dims=['y', 'x'],
                    coords={'y': [0.0, 1.0], 'x': [0.0, 1.0]})
kriging([5.0], [5.0], [42.0], tmpl)
# ValueError: zero-size array to reduction operation maximum which has no identity
```

The same crash happens when all but one of the input points are NaN, since `validate_points()` drops the NaN rows and leaves a single point behind.

## Cause

In `_experimental_variogram` (`xrspatial/interpolate/_kriging.py`), the pairwise distances are reduced with `dists.max()` before the `max_dist <= 0` guard runs. With one point, `np.triu_indices(n, k=1)` is empty, so `dists` is a zero-size array and `.max()` raises.

## Expected

A single point carries no spatial structure, so kriging cannot fit a variogram from it. The function already handles the degenerate case where every point sits at the same location (zero maximum distance): `_experimental_variogram` returns empty arrays and `kriging()` falls back to a default variogram. The single-point case should reach that same fallback path instead of crashing on the empty reduction.

## Scope

Fix is limited to `xrspatial/interpolate/_kriging.py`. Guard the empty-distances case in `_experimental_variogram` so the existing fallback handles it.

brendancol:
Closes #2917

`kriging()` crashed with an opaque numpy `ValueError` ("zero-size array to reduction operation maximum which has no identity") whenever the input reduced to a single point: either one point passed directly, or all but one point dropped by NaN filtering.

- `_experimental_variogram` now returns empty arrays when there are no distance pairs, before calling `dists.max()`. This routes the single-point case to the existing default-variogram fallback that already handles coincident points, so the call produces a constant-valued raster instead of raising.

Backend coverage: numpy (the variogram and matrix build run on host arrays for all four backends, so the fix applies regardless of the template's backend).

Test plan:
- [x] `test_single_point` — one input point yields a finite constant prediction
- [x] `test_single_valid_point_after_nan_drop` — NaN filtering down to one point reaches the same fallback
- [x] existing kriging suite still passes (18 passed)


brendancol (commented):
## PR Review: Guard single-point input in kriging() variogram

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The guard is placed correctly: `dists.size == 0` is checked before `dists.max()`, which is exactly where the empty-reduction crash originated (`_kriging.py`).
- It reuses the existing default-variogram fallback rather than adding a parallel code path. The `dists.size == 0` (n<2) and `max_dist <= 0` (coincident points) cases stay distinct, and both are needed.
- Because the variogram and matrix build run on host numpy arrays ahead of backend dispatch, the fix covers all four backends without touching the dispatch layer.
- Two tests cover both routes into the single-point state: one point passed directly, and one point left after NaN filtering. Both assert a finite constant prediction and the expected warning.

### Verification
- Reproduced the original `ValueError` on `main`, confirmed it is gone after the change.
- Single point returns a constant raster equal to the point value (42.0 and 7.0 in the two tests); ordinary-kriging weight collapses to 1.0 with no spatial structure, which is the sensible degenerate result.
- Full kriging suite: 18 passed.

### Checklist
- [x] Algorithm correct for the degenerate case
- [x] All backends covered (fix is upstream of dispatch)
- [x] NaN handling correct (NaN-drop-to-one routes to fallback)
- [x] README not applicable (no API change)
- [x] Docstrings unchanged; behavior already documented via MemoryError/warning paths

PR #2926, opened 2026-06-04
Skip full-grid k0 memory term for dask kriging templates

### LINKED ISSUE #2923: kriging() memory guard raises spurious MemoryError on dask-backed templates
## Summary

`kriging()`'s memory guard `_check_kriging_memory` raises a spurious `MemoryError` on dask-backed templates. It estimates the prediction matrix (`k0`) allocation from the full grid size, but the dask backend builds `k0` per chunk via `map_blocks`. Peak per-task memory scales with chunk size, not the full grid, so the guard rejects the large-grid case that the dask backend is meant to handle.

## Reproduction

A 4000x4000 dask template with 200x200 chunks and 20 input points raises a MemoryError claiming it needs ~8.1 GB, when each chunk's `k0` only needs ~20 MB:

```python
import numpy as np, dask.array as da, xarray as xr
from unittest import mock

rng = np.random.RandomState(0)
N = 20
x = rng.uniform(0, 100, N); y = rng.uniform(0, 100, N); z = 2*x + 3*y
yc = np.linspace(0, 100, 4000); xc = np.linspace(0, 100, 4000)
arr = da.zeros((4000, 4000), chunks=(200, 200), dtype='float64')
tmpl = xr.DataArray(arr, dims=['y', 'x'], coords={'y': yc, 'x': xc})

with mock.patch('xrspatial.zonal._available_memory_bytes', lambda: 64 * 1024**2):
    kriging(x, y, z, tmpl)   # raises spurious MemoryError
```

## Expected

For a dask-backed template, the `k0` term should be excluded from the guard (or scaled to the largest chunk), since `k0` is allocated per chunk. The variogram-pair and matrix terms are point-based and materialized on the host regardless of backend, so they still apply.

## Severity

Performance / scalability, MEDIUM. Affected backends: dask+numpy and dask+cupy.

brendancol:
Closes #2923

## What changed

- `_check_kriging_memory` now takes an `is_dask` flag. When the template is dask-backed, it drops the prediction-matrix (`k0`) term from the estimate, since `map_blocks` builds `k0` one chunk at a time and peak per-task memory scales with the chunk, not the full grid.
- The point-based terms (variogram pair arrays and the kriging matrix) are materialized on the host regardless of backend, so they still apply on the dask path.
- `kriging()` detects a dask-backed template via `isinstance(template.data, da.Array)` and passes the flag through.

## Why

Before this, a large chunked dask template raised a spurious `MemoryError` for the exact case the dask backend exists to handle. A 4000x4000 template with 200x200 chunks claimed it needed ~8.1 GB when each chunk only needs ~20 MB.

## Backends

numpy guard behaviour unchanged. dask+numpy and dask+cupy no longer hit the full-grid k0 estimate. cupy (non-dask) unchanged.

## Test plan

- [x] `test_dask_template_skips_grid_memory_guard`: a 4000x4000 / 200x200 dask template runs under a 64 MB cap
- [x] `test_check_helper_dask_skips_k0_term`: the helper drops the k0 term when `is_dask=True`
- [x] `test_check_helper_dask_still_guards_matrix`: the point-based matrix guard still fires on the dask path
- [x] Existing numpy memory-guard tests still pass (k0 term unchanged for numpy)
- [x] Full `test_interpolation.py` green (44 passed, includes cupy and dask+cupy parity)


brendancol (commented):
## PR Review: Skip full-grid k0 memory term for dask kriging templates

### Blockers
None.

### Suggestions
None.

### Nits
- `_kriging.py` `_check_kriging_memory`: with `is_dask=True` the k0 term is dropped unconditionally. A dask array with a single chunk covering the whole grid still allocates the full k0 in that one chunk, so the guard under-estimates that case. It is unusual to hold a full-grid dask template in one chunk, and the old full-grid estimate was wrong for the normal chunked case, so dropping the term is the right tradeoff here. If you want to tighten it later, scale k0 to the largest chunk via `template.data.chunks`.

### What looks good
- The change is small: one keyword arg and a per-backend choice for a single term. The variogram-pair and matrix terms still apply on the dask path, which is correct because both are built on the host regardless of backend.
- `is_dask` detection checks `da is not None` before the isinstance, so it is safe with dask uninstalled, and it covers dask+cupy since that backend also wraps a `dask.array.Array`.
- Tests cover the helper (k0 skipped, matrix still guarded), an end-to-end 4000x4000 / 200x200 dask template under a 64 MB cap, and the unchanged numpy path.

### Checklist
- Algorithm matches reference: n/a (memory guard, not a numerical kernel)
- All implemented backends consistent: yes, dask+numpy and dask+cupy both covered
- NaN handling: unchanged
- Dask chunk boundaries: n/a
- No premature materialization: confirmed, no .values/.compute added
- Benchmark: not needed (guard logic)
- Docstrings: updated to document is_dask behavior

Issue #2931, opened 2026-06-04
[1.0.0] GeoTIFF / COG I/O


brendancol:
## 1.0.0 readiness: GeoTIFF / COG I/O

**Area:** I/O  **Functions:** open_geotiff, to_geotiff, read_geotiff_dask, read_geotiff_gpu, write_geotiff_gpu, read_vrt, write_vrt
**Source:** `xrspatial/geotiff/`

### Scope
- [ ] Public surface frozen against the README feature matrix
- [ ] Dark-corner decisions resolved (see #2930)

### Sweep clearance, no CRITICAL or HIGH (accuracy and security first)
- [ ] Accuracy  `/deep-sweep xrspatial/geotiff`
- [ ] Security
- [ ] Performance
- [ ] API consistency

Last swept: none yet

### Examples
- [ ] Notebook example
- [ ] (optional) standalone script

### Linked bug issues
- #2830 (High) stable_only=True does not reject HTTP/fsspec sources via direct read_geotiff_gpu()

brendancol:
```
  Final results — deep-sweep geotiff (7 sweeps)

  ┌─────────────────┬────────────────────────────────────────────────┬─────────────────────┬───────────────────┐
  │      Sweep      │                    Findings                    │     /rockout PR     │ State row written │
  ├─────────────────┼────────────────────────────────────────────────┼─────────────────────┼───────────────────┤
  │ accuracy        │ clean (0/0/0/0)                                │ --                  │ yes               │
  ├─────────────────┼────────────────────────────────────────────────┼─────────────────────┼───────────────────┤
  │ api-consistency │ clean (3 LOW re-confirmed intentional)         │ --                  │ yes               │
  ├─────────────────┼────────────────────────────────────────────────┼─────────────────────┼───────────────────┤
  │ metadata        │ clean (0/0/0/0)                                │ --                  │ yes               │
  ├─────────────────┼────────────────────────────────────────────────┼─────────────────────┼───────────────────┤
  │ performance     │ clean (0/0/0/0; 1 pre-existing LOW documented) │ --                  │ yes ✅ (re-run)   │
  ├─────────────────┼────────────────────────────────────────────────┼─────────────────────┼───────────────────┤
  │ security        │ clean (0/0/0/0)                                │ --                  │ yes               │
  ├─────────────────┼────────────────────────────────────────────────┼─────────────────────┼───────────────────┤
  │ style           │ clean (0/0/0/0)                                │ --                  │ yes               │
[...]

PR #2963, opened 2026-06-05
geotiff: align open_geotiff parameters with rioxarray's open_rasterio (#2961)

### LINKED ISSUE #2961: geotiff: align open_geotiff parameters with rioxarray's open_rasterio
## Reason or Problem

`xrspatial.geotiff.open_geotiff` and rioxarray's `open_rasterio` do the same job (read a GeoTIFF/COG/VRT into an `xarray.DataArray`), but their signatures diverge on parameter names and one default, so moving code between them needs edits:

- The nodata-masking flag is `mask_nodata` (default `True`, promotes the sentinel to `NaN`) here, versus `masked` (default `False`) in `open_rasterio`. The name and the default both differ.
- The array-name kwarg is `name` here, versus `default_name` in `open_rasterio`.
- `open_rasterio` has `mask_and_scale`, `parse_coordinates`, `lock`, and `cache`; `open_geotiff` has no counterpart for any of them.

## Proposal

Align the overlapping parameters with `open_rasterio` so a rioxarray-style call works on `open_geotiff` with minimal edits, while keeping the xrspatial-only parameters (`gpu`, `window`, `bbox`, `max_pixels`, `allow_*`, tier flags) and not renaming parameters that already match (`chunks`).

**Design:**

- Rename `mask_nodata` -> `masked` and flip the default from `True` to `False` to match `open_rasterio`. Keep `mask_nodata` as a deprecated alias: passing it emits a `DeprecationWarning`, passing both `masked` and `mask_nodata` raises `TypeError`. This reuses the sentinel-based deprecation pattern already used for `read_geotiff_gpu`'s `gpu` -> `on_gpu_failure` rename (`_runtime.py` sentinels + the resolution block in `_backends/gpu.py`).
- Rename `name` -> `default_name` with the same deprecated-alias mechanism.
- Add `mask_and_scale` (default `False`): read GDAL `SCALE`/`OFFSET` from `attrs['gdal_metadata']`, apply `data * scale + offset`, mask the nodata sentinel, and record `attrs['scale_factor']`/`attrs['add_offset']`, mirroring `open_rasterio`.
- Add `parse_coordinates` (default `True`): when `False`, skip building the `x`/`y` coordinate arrays while keeping `attrs['transform']` and `attrs['crs']`.
- Add `lock` and `cache` for signature compatibility. xrspatial's dask reader re-opens the source per window, so there is no shared GDAL handle to lock and no caching backend to toggle; passing a non-default value emits a `GeoTIFFFallbackWarning` explaining this rather than silently ignoring the kwarg.

The four backend readers keep their internal kwarg names; the public contract change lives only at `open_geotiff`.

**Usage:**

```python

da = open_geotiff("dem.tif", masked=True)              # was mask_nodata=True
da = open_geotiff("dem.tif", mask_and_scale=True)      # apply SCALE/OFFSET, mask nodata
da = open_geotiff("dem.tif", parse_coordinates=False)  # index coords, keep crs/transform attrs
```

**Value:** lowers the cost of switching from rioxarray and makes the reader's surface predictable for users who already know `open_rasterio`.

## Stakeholders and Impacts

Affects callers of `open_geotiff`. The masking-default flip is a breaking behavior change: a bare `open_geotiff(path)` stops promoting the nodata sentinel to `NaN`. The deprecated `mask_nodata` alias keeps the old name working through a deprecation window.

## Drawbacks

The masking-default flip changes the result of the default call path. `mask_and_scale` is new code and depends on GDAL writing scale/offset into the `GDAL_METADATA` tag. Sources without those items get a `(1.0, 0.0)` no-op.

## Alternatives

- Additive aliases that keep the current masking default — rejected because a bare call would still not match `open_rasterio`.
- Renaming every shared parameter including `chunks` — rejected because `chunks` already matches.

## Unresolved Questions

Whether `lock`/`cache` should map onto anything in the dask read path or stay accept-and-warn. The current plan is accept-and-warn.

brendancol:
Closes #2961

## What changed

Aligns `open_geotiff`'s read parameters with rioxarray's `open_rasterio` so callers can move between the two with minimal edits.

- Renamed `mask_nodata` -> `masked` and flipped the default from `True` to `False` to match `open_rasterio`. A bare `open_geotiff(path)` no longer promotes the nodata sentinel to NaN; pass `masked=True` for the old behavior. `mask_nodata` stays as a deprecated alias (`DeprecationWarning`; passing both raises `TypeError`).
- Renamed `name` -> `default_name`, with `name` kept as a deprecated alias on the same terms.
- Added `mask_and_scale`: reads GDAL `SCALE`/`OFFSET` from `attrs['gdal_metadata']`, returns `data * scale + offset`, masks the sentinel, and records `attrs['scale_factor']`/`attrs['add_offset']`. No-op when the source carries no scale/offset.
- Added `parse_coordinates`: `False` skips the `x`/`y` coordinate arrays while keeping `attrs['transform']`/`attrs['crs']`.
- Added `lock`/`cache` for signature compatibility. xrspatial's reader re-opens the source per window, so these have no analog; passing a non-default value emits a `GeoTIFFFallbackWarning` rather than silently ignoring the kwarg.
- `chunks` already matched rioxarray and is unchanged, as are the xrspatial-only parameters (`gpu`, `window`, `bbox`, `max_pixels`, `allow_*`, tier flags).

The deprecation aliases reuse the existing sentinel pattern from `read_geotiff_gpu`'s `gpu` -> `on_gpu_failure` rename.

## Scope note

`masked` and `default_name` thread through all four backends (eager, dask, GPU, VRT). `mask_and_scale` and `parse_coordinates` are implemented on the CPU eager and dask paths; combining either with `gpu=True` or a `.vrt` source raises `ValueError` with a message pointing at the supported path, matching how the dispatcher already gates `on_gpu_failure`/`missing_sources`/`max_cloud_bytes` per backend. This keeps the change inside the tiers where behavior is parity-gated rather than threading untested code through the GPU and VRT-mosaic readers.

## Backend coverage

- numpy: full (rename, default flip, mask_and_scale, parse_coordinates)
- dask+numpy: full
- cupy / dask+cupy: rename + default flip; `mask_and_scale`/`parse_coordinates=False` rejected with a clear error

## Test plan

- [x] New `xrspatial/geotiff/tests/read/test_rioxarray_compat_2961.py`: default flip, deprecation warnings, both-passed `TypeError`, alias equivalence, `mask_and_scale` (eager + dask, no-metadata no-op, `dtype=int` raises), `parse_coordinates`, `lock`/`cache` warnings, GPU/VRT gating.
- [x] Migrated the existing suite for the default flip (masking-behavior tests now pass `masked=True`; canonical names replace deprecated-alias usage).
- [x] Full `xrspatial/geotiff/tests` suite green (GPU available in the dev env, so cupy/dask+cupy/gpu params ran).
- [x] `xrspatial/tests/test_reproject.py`, `test_rasterize.py`, `test_accessor.py` green (the reproject datum-grid fallback now matches its rasterio primary path, which also does not mask).

The user-guide notebook and README feature-matrix steps were skipped: this is parameter alignment on an existing function, not a new spatial operation or backend-support change.

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -181,11 +181,12 @@ this section is the brief.
   (``xrspatial/geotiff/tests/write/test_nodata.py``).
 * Float nodata. The on-disk sentinel is recorded on
   ``attrs['nodata']`` and surfaces as NaN in pixel data only when the
-  read promotes via ``mask_nodata=True`` (the default for float
-  outputs). With ``mask_nodata=False`` the raw float sentinel passes
-  through, so downstream callers can branch on the exact value;
-  ``xrspatial/geotiff/tests/write/test_nodata.py`` pins this
-  split.
+  read promotes via ``masked=True``. The default is ``masked=False``
+  (matching rioxarray's ``open_rasterio``), so by default the raw float
+  sentinel passes through and downstream callers can branch on the exact
+  value; pass ``masked=True`` to get NaN-masked output.
+  ``xrspatial/geotiff/tests/write/test_nodata.py`` pins this split.
+  (``mask_nodata`` is a deprecated alias of ``masked``.)
 * NaN nodata. A file that declares ``nodata=NaN`` is read with NaN in
   both ``attrs['nodata']`` and pixel data (NaN propagates either way).
 * ``attrs['masked_nodata']``. Every read sets a boolean lifecycle
--- a/docs/source/reference/release_gate_geotiff.rst
+++ b/docs/source/reference/release_gate_geotiff.rst
@@ -199,7 +199,7 @@ Local GeoTIFF read and write
        keys (``transform``, ``crs``, ``crs_wkt``, ``nodata``,
        ``masked_nodata``, ``georef_status``, ``raster_type``) across
        four scenarios: integer-nodata, float-NaN-nodata, MinIsWhite,
-       and the ``mask_nodata=False`` raw-sentinel branch of the
+       and the ``masked=False`` raw-sentinel branch of the
        nodata lifecycle.
      - ``xrspatial/geotiff/tests/release_gates/test_stable_features.py``
        (eager / dask full parity section)
@@ -384,7 +384,7 @@ Nodata lifecycle
      - stable
      - The sentinel survives read and write across every backend; integer
        sentinels are preserved bit-exact, float sentinels surface as NaN
-       only when ``mask_nodata=True``.
+       only when ``masked=True`` (default ``masked=False``).
      - ``xrspatial/geotiff/tests/read/test_nodata.py``,
        ``xrspatial/geotiff/tests/write/test_nodata.py``
      - `#2341`_
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -89,8 +89,8 @@ write.
        replaced sentinel pixels with NaN (so the buffer is NaN-aware);
        ``False`` when the array still carries the literal sentinel
        values, including the case where the array is float dtype
-       because the caller passed ``mask_nodata=False`` together with
-       ``dtype=float...``. Only set when ``nodata`` is set; absence
+       because the caller passed ``masked=False`` (the default) together
+       with ``dtype=float...``. Only set when ``nodata`` is set; absence
        means no declared sentinel. See issue #2092.
    * - ``nodata_pixels_present``
      - bool
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -81,7 +81,8 @@ from ._geotags import RASTER_PIXEL_IS_AREA, RASTER_PIXEL_IS_POINT, GeoTransform 
 from ._reader import _MAX_CLOUD_BYTES_SENTINEL, CloudSizeLimitError, UnsafeURLError
 from ._reader import read_to_array as _read_to_array
 from ._runtime import (_CRS_WKT_DEPRECATED_SENTINEL, _GPU_DEPRECATED_SENTINEL,  # noqa: F401
-                       _MISSING_SOURCES_SENTINEL, _ON_GPU_FAILURE_SENTINEL, GeoTIFFFallbackWarning,
+                       _MASK_NODATA_DEPRECATED_SENTINEL, _MISSING_SOURCES_SENTINEL,
+                       _NAME_DEPRECATED_SENTINEL, _ON_GPU_FAILURE_SENTINEL, GeoTIFFFallbackWarning,
                        _geotiff_strict_mode, _gpu_fallback_warning_message)
 from ._validation import (_validate_3d_writer_dims, _validate_chunks_arg,  # noqa: F401
                           _validate_tile_size_arg)
@@ -491,7 +492,8 @@ def open_geotiff(source: str | BinaryIO, *,
                  bbox: tuple | None = None,
                  overview_level: int | None = None,
                  band: int | None = None,
-                 name: str | None = None,
+                 default_name: str | None = None,
+                 name: str | None = _NAME_DEPRECATED_SENTINEL,  # type: ignore[assignment]
                  chunks: int | tuple | None = None,
                  gpu: bool = False,
                  max_pixels: int | None = None,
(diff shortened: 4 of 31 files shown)


brendancol (commented):
## PR Review: align open_geotiff parameters with rioxarray's open_rasterio

Reviewed the four source files in full (`__init__.py`, `_attrs.py`, `_backends/dask.py`, `_runtime.py`) plus the test and doc changes. Full geotiff suite is green locally (6092 passed) with GPU present, and the new compat tests cover the renames, the default flip, both-passed errors, `mask_and_scale` eager/dask, `parse_coordinates`, the lock/cache warnings, and the GPU/VRT gates.

None.

- [ ] Round-trip double-apply with `mask_and_scale`. In `_attrs.py` (`_finalize_eager_read`) and `_backends/dask.py`, after applying `data * scale + offset` the source's `attrs['gdal_metadata']` still carries the raw `SCALE`/`OFFSET` items. If a caller reads with `mask_and_scale=True` and then writes the result with `to_geotiff`, the writer re-embeds those items, so a later `mask_and_scale=True` read applies the scale a second time. rioxarray sidesteps this by moving the values to `encoding`. Either drop `SCALE`/`OFFSET` from the propagated `gdal_metadata` once they have been applied, or document the round-trip caveat on the `mask_and_scale` parameter.
- [ ] Resolution edge case in `open_geotiff` (`__init__.py`). The "both passed" guard is `if mask_nodata is not <sentinel>: if masked is not False: raise`. Because `masked` has a real default of `False`, an explicit `masked=False` cannot be told apart from the default, so `open_geotiff(path, masked=False, mask_nodata=True)` does not raise and silently resolves to masking on. This matches the documented stance in `read_geotiff_gpu`, but a one-line comment in the resolution block noting the case would save the next reader the trace.

- [ ] `_extract_scale_offset` (`_attrs.py`) applies one scale/offset pair to all bands and falls back to band 0 for per-band metadata. The docstring already says this; consider a warning when per-band values actually disagree so a multi-band source with mismatched scaling does not silently use band 0.
- [ ] `__init__.py` computes `_is_vrt_source_early` for the new gate and then `_is_vrt_source` again a few lines later. Harmless, but the second could reuse the first.

### What looks good
- The deprecation aliases reuse the existing sentinel pattern from the `gpu` -> `on_gpu_failure` rename rather than inventing a new mechanism, including the both-passed `TypeError`.
- `mask_and_scale` is verified to match between the eager and dask paths, and the `dtype=<int>` + scale combination raises through the existing float-to-int guard.
- The default flip is the risk here, and the test migration is thorough: masking-behavior tests now opt in with `masked=True`, and the two tests that pinned the old default were rewritten to pin the new one.
- Scoping `mask_and_scale` / `parse_coordinates=False` to the CPU eager and dask paths with an explicit `ValueError` for GPU and VRT keeps the change inside the parity-gated tiers instead of threading untested code through the GPU and VRT readers.

### Checklist
- [x] Behavior matches rioxarray's `open_rasterio` for the aligned parameters
- [x] Renames thread through all four backends; new behavioral options gated where unsupported
- [x] NaN / sentinel masking correct under the new default
- [x] Edge cases covered (no-metadata no-op, dtype clash, both-passed, deprecation warnings)
- [x] Dask path matches eager for `mask_and_scale`
- [x] No premature materialization (dask scale/offset stays lazy)

brendancol (commented):
## Follow-up review (after commit a59264c9)

Disposition of the findings from the first pass:

- Fixed: round-trip double-apply with `mask_and_scale`. The reader sets both `attrs['gdal_metadata']` and `attrs['gdal_metadata_xml']`, and `to_geotiff` prefers the XML, so stripping only the dict would not have prevented re-embedding and stripping the serialized XML pulls the writer path into this read-focused change. Documented the caveat on the `mask_and_scale` parameter instead (drop the scale/offset tags and `scale_factor`/`add_offset` before writing if a clean round-trip matters). Filed as the proportionate fix; a writer-side strip can be a follow-up if it comes up in practice.
- Fixed: added a comment in the `open_geotiff` resolution block spelling out the `masked=False` + `mask_nodata=True` edge case.
- Fixed (nit): unified the duplicated `_is_vrt_source` computation; the gate and the later VRT branch now share one binding.
- Dismissed (nit): warning when per-band `SCALE`/`OFFSET` disagree. The band-0 fallback is already documented, the case is uncommon, and emitting warnings from the low-level attrs helper would expand the change without much payoff.

Full `xrspatial/geotiff/tests` suite is green after the follow-up (6114 passed, 81 skipped, 1 xfailed) with GPU present.

brendancol:
#2931

PR #2967, opened 2026-06-05
geotiff: write VRT index atomically via temp-then-rename (#2965)

### LINKED ISSUE #2965: VRT XML write in write_vrt is not atomic
**Describe the bug**

The VRT XML write in `write_vrt` (`xrspatial/geotiff/_vrt.py`) is not atomic. The final `.vrt` file is opened and written in place:

```python
xml = '\n'.join(lines) + '\n'
with open(vrt_path, 'w') as f:
    f.write(xml)
```

If the process crashes, the disk fills up, or the write is interrupted between `open` and the finished `write`, a truncated `.vrt` is left at the final path.

This bites hardest on the tiled path. `_write_vrt_tiled` (`xrspatial/geotiff/_writers/eager.py`) writes every tile into a staging directory, promotes that directory to its final name with an atomic `os.replace`, and only then writes the VRT index. The index is the last non-atomic step, so a partial index ends up pointing at a fully-promoted, complete tile set. A reader that picks up the half-written VRT sees a valid tile directory behind a malformed index.

The other local writers in the module already handle this. `_write_bytes` in `_writer.py` writes to a temp file in the same directory and then `os.replace`s it into place. The VRT index write skips that path and writes directly.

**Expected behavior**

The VRT XML write should be atomic: write to a temp file in the same directory as the destination, then `os.replace` it into the final path. An interrupted or failed write should never leave a partial `.vrt` behind. The temp file has to live on the same filesystem as the destination for the rename to stay atomic. The fix should reuse the existing `_write_bytes` helper instead of duplicating the temp-then-rename logic.

**Additional context**

- Cited line: the `open(vrt_path, 'w')` write at the end of `write_vrt` in `xrspatial/geotiff/_vrt.py`.
- Promotion-before-index ordering: the `os.replace(staging_dir, tiles_dir)` followed by `write_vrt(...)` sequence in `xrspatial/geotiff/_writers/eager.py`.
- Existing atomic helper: `_write_bytes` in `xrspatial/geotiff/_writer.py`.

brendancol:
Closes #2965

`write_vrt` in `xrspatial/geotiff/_vrt.py` wrote the final `.vrt` in place with `open(vrt_path, 'w')`. A crash, a full disk, or an interrupted write between `open` and the finished `write` could leave a truncated index at the final path.

The tiled writer makes this worse. `_write_vrt_tiled` promotes the staging tile directory to its final name with an atomic `os.replace` and only then writes the index, so the index was the last non-atomic step. A partial index would point at a complete, fully-promoted tile set, and a reader that picks it up sees a valid tile directory behind a malformed index.

## Change

Route the index write through the existing `_write_bytes` helper in `_writer.py`. It writes to a temp file in the destination directory and `os.replace`s it into place. The temp file lives on the same filesystem as the destination, so the rename stays atomic. No new temp-then-rename code.

## Backend coverage

Not applicable. `write_vrt` writes a filesystem index, it is not a backend-dispatched array op. The behavior is the same on every platform: `os.replace` is atomic on POSIX and is the standard atomic-rename primitive on Windows for same-directory targets.

## Notes

`_write_bytes` names its temp file with a `.tif.tmp` suffix. That suffix is cosmetic for a VRT, the file is renamed to the real `.vrt` immediately and is only visible if the process dies mid-write, so I left the shared helper alone rather than reshaping it for one caller.

## Test plan

- [x] `test_write_vrt_leaves_no_temp_file`: a successful write leaves the `.vrt` and no `.tmp` artifacts in the directory.
- [x] `test_write_vrt_no_partial_file_on_write_failure`: a forced failure during the write leaves no partial `.vrt` at the final path.
- [x] `test_write_vrt_failure_preserves_existing_file`: a failed rewrite leaves a pre-existing `.vrt` byte-for-byte intact.
- [x] Existing VRT write, metadata, nodata, and tiled-writer suites pass unchanged.

### DIFF
--- a/xrspatial/geotiff/_vrt.py
+++ b/xrspatial/geotiff/_vrt.py
@@ -2242,7 +2242,13 @@ def write_vrt(vrt_path: str, source_files: list[str], *,
     lines.append('</VRTDataset>')
 
     xml = '\n'.join(lines) + '\n'
-    with open(vrt_path, 'w') as f:
-        f.write(xml)
+    # Write the index atomically: a temp file in the destination
+    # directory followed by ``os.replace``. The tiled writer promotes the
+    # tile directory before this call, so a direct in-place write here is
+    # the last non-atomic step and an interrupted write would leave a
+    # partial ``.vrt`` pointing at a complete tile set. ``_write_bytes``
+    # already implements the temp-then-rename pattern for local paths.
+    from ._writer import _write_bytes
+    _write_bytes(xml.encode('utf-8'), vrt_path)
 
     return vrt_path
--- a/xrspatial/geotiff/tests/write/test_basic.py
+++ b/xrspatial/geotiff/tests/write/test_basic.py
@@ -1526,6 +1526,78 @@ def test_compatible_sources_succeed(tmp_path):
     assert os.path.exists(vrt)
 
 
+def test_write_vrt_leaves_no_temp_file(tmp_path):
+    """A successful write_vrt leaves the final .vrt and no stray temp
+    files in the destination directory (issue #2965)."""
+    d = _unique_dir(tmp_path, "atomic_clean")
+    a = os.path.join(d, "a.tif")
+    b = os.path.join(d, "b.tif")
+    _write_tif(a, h=4, w=4, dtype=np.float32)
+    _write_tif(b, h=4, w=4, dtype=np.float32, origin_x=4.0)
+    vrt = os.path.join(d, "out.vrt")
+    _priv_write_vrt(vrt, [a, b])
+    assert os.path.exists(vrt)
+    # No leftover temp artifacts from the temp-then-rename write.
+    leftovers = glob.glob(os.path.join(d, "*.tmp")) + glob.glob(
+        os.path.join(d, "*.tmp*"))
+    assert leftovers == [], f"temp files not cleaned up: {leftovers}"
+
+
+def test_write_vrt_no_partial_file_on_write_failure(tmp_path, monkeypatch):
+    """If the index write fails mid-flight, no partial .vrt is left at
+    the final path (issue #2965). The write goes to a temp file in the
+    destination directory and is os.replace'd into place, so a failure
+    before the rename never publishes a truncated index."""
+    d = _unique_dir(tmp_path, "atomic_fail")
+    a = os.path.join(d, "a.tif")
+    b = os.path.join(d, "b.tif")
+    _write_tif(a, h=4, w=4, dtype=np.float32)
+    _write_tif(b, h=4, w=4, dtype=np.float32, origin_x=4.0)
+    vrt = os.path.join(d, "out.vrt")
+
+    # Force the atomic helper to blow up while writing the temp file,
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: write VRT index atomically via temp-then-rename (#2965)

Reviewed the full diff and the surrounding code in `_vrt.py`, `_writer.py`, and the test helpers.

None.

None blocking. One thing worth calling out explicitly in the PR/commit, not a code change: the old write used `open(vrt_path, 'w')`, i.e. the platform default text encoding. The new write encodes UTF-8 explicitly. That is the correct direction, since `_read_vrt_xml` already decodes the file as UTF-8 (`.decode('utf-8')`), so writer and reader now agree. On Windows the old default could have been cp1252, which would have mangled non-ASCII CRS WKT on round-trip. So this is a small latent-bug fix riding along with the atomicity change, not a regression. Fine to keep, just good to have on record.

- `test_write_vrt_failure_preserves_existing_file` reads the file with bare `open(vrt, "rb").read()` twice (test_basic.py, in the new test). The handles get closed eventually by GC, but `pathlib.Path(vrt).read_bytes()` or a `with` block is cleaner and matches how the rest of the suite reads small files. Cosmetic.

### What looks good

- Reuses the existing `_write_bytes` helper instead of hand-rolling a second temp-then-rename. The temp file lands in the destination directory, so the `os.replace` stays on one filesystem and is a true atomic rename.
- The local `from ._writer import _write_bytes` import avoids any import-cycle risk and matches the local-import style already used between these modules (eager.py does `from .._vrt import write_vrt`).
- The tests cover the cases that matter: a clean write leaves no temp litter, a failed write leaves no partial `.vrt`, and a failed rewrite leaves a pre-existing `.vrt` byte-for-byte intact. The monkeypatch targets `writer_mod._write_bytes`, which is the right binding since `write_vrt` looks the name up at call time.
- The write-site comment says why atomicity matters here specifically: the tiled writer promotes the tile dir before the index write.

### Checklist
- [x] Behavior matches the intended fix (atomic temp-then-rename)
- [x] No backend dispatch involved; filesystem index writer
- [x] No NaN / float concerns in this path
- [x] Edge cases covered (write failure, rewrite over existing file)
- [x] No dask chunk-boundary concerns
- [x] No benchmark needed (I/O index write, not a hot array path)
- [x] README feature matrix not applicable (internal-only function, no API change)
- [x] Write-site comment present and accurate

brendancol (commented):
## Follow-up review (#2967)

### Blockers
None.

### Suggestions
None.

### Nits
The one nit from the first pass is resolved: `test_write_vrt_failure_preserves_existing_file` now reads the file inside `with open(...)` blocks instead of bare `open(...).read()`.

### What looks good
The production change is unchanged and correct: the VRT index now goes through `_write_bytes`, which writes a temp file in the destination directory and `os.replace`s it into place. Writer and reader agree on UTF-8 (`_read_vrt_xml` decodes UTF-8). Nothing else moved.

Nothing left to address. Good to merge once CI is green.

PR #2968, opened 2026-06-05
geotiff: clean up VRT staging dir on early validation failures

### LINKED ISSUE #2964: VRT tiled writes leak staging directories on early validation failures
**Describe the bug**

`_write_vrt_tiled` in `xrspatial/geotiff/_writers/eager.py` creates its staging directory (`<stem>_tiles.tmp-<uuid>`) at line ~1202, before the cleanup-protected `try`/`except` block at line ~1340. A few validation steps run in between: CRS validation, the georeferenced-transform requirement, and the "2D arrays only" check at line ~1309. If any of them raises, nothing removes the staging directory.

So a failed VRT write leaves an orphaned `*.tmp-*` directory on disk. For example, `to_geotiff(data_3d, "out.vrt")` raises the 2D-only `ValueError` and leaves `out_tiles.tmp-<uuid>` behind. That contradicts the function's own comment, which claims a failed write never leaves partial output ("tiles are staged in a temp dir ... so a retry after a failure starts from a clean slate").

Reproduced locally.

**Expected behavior**

When validation fails after the staging directory exists, the staging directory should be removed so a failed write leaves nothing behind, matching the documented contract.

**Fix**

Move the staging-directory creation inside the cleanup-protected block, or otherwise guarantee it gets removed on any early failure path.

**Additional context**

Affects the VRT tiled-write path: `to_geotiff(..., ".vrt")` and direct `_write_vrt_tiled` calls. Module: geotiff / VRT.

brendancol:
Closes #2964.

`_write_vrt_tiled` created its staging directory (`<stem>_tiles.tmp-<uuid>`) before the cleanup-protected `try` block. Validation that runs in between (CRS, the georeferenced-transform requirement, and the 2D-only check) could raise and leave the staging dir on disk, which contradicts the function's no-partial-output contract. A public `to_geotiff(data_3d, "out.vrt", tiled=True)` reproduced the leak: it raised the 2D-only `ValueError` and left `out_tiles.tmp-<uuid>` behind.

## Change

- Extend the existing `try`/`except` so it also covers the post-staging validation. Any early failure now runs `_cleanup_staging()` and removes the `*.tmp-*` directory.
- The write loop and final atomic-rename / VRT-index steps are unchanged.

## Backends

The VRT tiled-write path covers numpy, cupy, and dask (the dask streaming path shares the same staging dir and cleanup). No backend behavior changed; this only affects the failure path.

## Test plan

- [x] New regression test in `test_vrt_atomic.py`: a 3D VRT write fails the 2D-only check and leaves no `*.tmp-*` staging dir, no `_tiles` dir, and no VRT.
- [x] Existing `test_vrt_atomic.py` cases (per-tile failure cleanup, empty-dir reuse, dask cleanup) still pass.
- [x] Full `xrspatial/geotiff/tests/write/` suite: 1036 passed, 1 skipped.
- [x] Verified the leak on the pre-fix code and its absence after the fix.

### DIFF
--- a/xrspatial/geotiff/_writers/eager.py
+++ b/xrspatial/geotiff/_writers/eager.py
@@ -1201,143 +1201,148 @@ def _write_vrt_tiled(data, vrt_path, *, crs=None, nodata=None,
         vrt_dir, f'{tiles_dir_name}.tmp-{uuid.uuid4().hex}')
     os.makedirs(staging_dir, exist_ok=True)
 
-    # Resolve CRS. ``numbers.Integral`` covers numpy integer scalars
-    # (``np.int32``, ``np.int64``) so ``crs=np.int64(4326)`` does not
-    # silently fall through to ``epsg=None``. Validator already
-    # rejects bool.
-    _validate_crs_arg(crs)
-    epsg = None
-    wkt_fallback = None
-    if isinstance(crs, numbers.Integral):
-        epsg = int(crs)
-    elif isinstance(crs, str):
-        epsg = _wkt_to_epsg(crs)
-        if epsg is None:
-            wkt_fallback = crs
-
-    geo_transform = None
-    raster_type = RASTER_PIXEL_IS_AREA
-    x_res = None
-    y_res = None
-    res_unit = None
-    gdal_meta_xml = None
-    extra_tags_list = None
-
-    if isinstance(data, xr.DataArray):
-        raw = data.data
-        if epsg is None and crs is None:
-            crs_attr = data.attrs.get('crs')
-            if isinstance(crs_attr, str):
-                epsg = _wkt_to_epsg(crs_attr)
-                if epsg is None and wkt_fallback is None:
-                    wkt_fallback = crs_attr
-            elif crs_attr is not None:
-                # Same gate as the kwarg path: reject bool / non-int
-                # types and confirm the EPSG resolves before writing it
-                # to disk. Without this, ``attrs={'crs': True}`` round-
-                # trips as EPSG=1.
-                _validate_crs_arg(crs_attr)
-                epsg = int(crs_attr)
-            if epsg is None:
-                wkt = data.attrs.get('crs_wkt')
-                if isinstance(wkt, str):
-                    epsg = _wkt_to_epsg(wkt)
-                    if epsg is None and wkt_fallback is None:
-                        wkt_fallback = wkt
-        if nodata is None:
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: clean up VRT staging dir on early validation failures

None.

None.

- `xrspatial/geotiff/_writers/eager.py`: the validation and write body now sit inside one `try`, so the diff looks large (about 130 re-indented lines) even though the behavior change is small. The comment at line 1207 explains why the guard spans both. Most of the diff is whitespace; flagging so a reviewer reading it knows what to skim past.

### What looks good
- Root cause is right. `os.makedirs(staging_dir)` ran before the cleanup-protected `try`, so a validation failure in between (CRS, georef transform, or the 2D-only check) leaked the `*.tmp-*` dir. Extending the existing `try` to cover validation is the minimal fix and reuses the existing `except BaseException: _cleanup_staging(); raise`.
- The `_cleanup_staging` helper was hoisted above the `try` and the old duplicate definition removed, so there is one definition and no shadowing.
- The post-`try` sections (atomic rename, VRT index write) are unchanged and still reference function-scope locals (`tile_names`, `nodata`, `tiles_dir`) that get assigned on the success path before the `try` exits.
- New regression test fails a 3D VRT write at the 2D-only check and asserts no `*.tmp-*` dir, no `_tiles` dir, and no VRT remain. It sits next to the existing atomicity tests. I confirmed it fails on pre-fix code (leaks `out_tiles.tmp-*`) and passes after.

### Backend notes
The VRT tiled path covers numpy, cupy, and dask. The fix is on the shared failure path, so it applies to all three. No backend behavior changed.

### Pre-existing observation (not introduced by this PR)
`restore_sentinel` is only assigned inside the `if isinstance(data, xr.DataArray)` branch but is passed to `_write_single_tile` for every backing type. A direct `_write_vrt_tiled(numpy_array, ...)` call would hit a `NameError` before the tile write. This predates the PR and the indentation move does not change it. Noting it for awareness, not as a fix request.

### Checklist
- Contract restored: yes, the no-partial-output guarantee holds again
- Backends consistent: yes, shared failure path
- NaN handling: not relevant here
- Edge cases: 3D early-validation failure now tested; per-tile and dask failure paths already covered
- No premature materialization: unchanged
- Benchmark: not needed (failure-path fix)
- README/docs: not needed (private helper, no public API change)
- Docstrings: the test has a clear docstring; the new comment explains the guard

PR #2971, opened 2026-06-05
vrt: stop build_vrt fabricating a GeoTransform for non-georeferenced sources

### LINKED ISSUE #2966: build_vrt fabricates a GeoTransform for non-georeferenced TIFFs
**Describe the bug**

`build_vrt` writes a `<GeoTransform>` element into the output VRT even when the source TIFF has no georeferencing. A plain (non-georeferenced) TIFF reads back with `georef_status='none'`, but after wrapping it in a VRT with `build_vrt`, the VRT reads as `georef_status='transform_only'` with a fabricated identity transform `(1, 0, 0, 0, -1, 0)`.

This is silent metadata fabrication. Downstream spatial functions can then treat pixel coordinates as real CRS-space coordinates, and nothing distinguishes the fabricated transform from a real one.

Root cause in `xrspatial/geotiff/_vrt.py`: the per-source metadata dict always stores `geo.transform`, even when `geo.has_georef` is false. With no georeferencing, `geo.transform` is the default identity `GeoTransform(origin_x=0, origin_y=0, pixel_width=1, pixel_height=-1)`. The writer always emits a `<GeoTransform>` line, so that identity leaks into the VRT XML.

**Reproduction**

```python
from xrspatial.geotiff import to_geotiff, open_geotiff, build_vrt

d = tempfile.mkdtemp()
arr = xr.DataArray(np.arange(12, dtype='float32').reshape(3, 4))
tif = os.path.join(d, 'plain.tif')
to_geotiff(arr, tif)

print(open_geotiff(tif).attrs['georef_status'])   # none

vrt = os.path.join(d, 'plain.vrt')
build_vrt(vrt, [tif])
out = open_geotiff(vrt)
print(out.attrs['georef_status'])   # transform_only  (BUG)
print(out.attrs['transform'])       # (1.0, 0.0, 0.0, 0.0, -1.0, 0.0)
```

**Expected behavior**

When every source TIFF has no georeferencing (`geo.has_georef` is false), `build_vrt` should not emit a `<GeoTransform>` element. The resulting VRT then preserves `georef_status='none'`. A VRT without a `<GeoTransform>` already reads back as `none` with integer pixel coordinates, so the reader side needs no change.

Mixed sources (some georeferenced, some not) should be rejected with a clear error, matching the existing fail-loud gates in `build_vrt`. A non-georeferenced tile cannot be placed correctly on a georeferenced mosaic.

**Additional context**

Severity: high. The fabricated transform is indistinguishable from a real one to downstream consumers.

brendancol:
Closes #2966

## Problem

`build_vrt` wrote a `<GeoTransform>` into the VRT even when the source TIFF had no georeferencing. A non-georeferenced source carries a default identity transform (origin 0,0; pixel 1,-1) with `has_georef=False`, and that identity leaked into the VRT XML. A plain TIFF that reads as `georef_status='none'` came back from its VRT as `transform_only` with a fabricated `(1, 0, 0, 0, -1, 0)` transform.

## Change

- Track `has_georef` per source in `build_vrt` and emit `<GeoTransform>` only when the sources are georeferenced. A VRT without `<GeoTransform>` already reads back as `none` with integer pixel coords, so the reader side is unchanged.
- Add `_check_no_mixed_georef`, a fail-loud gate that rejects mixing georeferenced and non-georeferenced sources. A plain tile only has identity values, so it cannot be placed on a georeferenced mosaic.
- Two window-transform-shift tests in `test_signatures.py` relied on the fabricated identity GeoTransform. They now use a hand-built VRT with an explicit origin (0,0) GeoTransform so they still exercise the shift contract.

## Backend coverage

`build_vrt` is a writer that produces VRT XML on disk; it is backend-independent. Reads of the resulting VRT were verified through `open_geotiff` (eager numpy path).

## Test plan

- [x] Non-georeferenced source: VRT omits `<GeoTransform>` and reads back as `georef_status='none'` with `transform=None`.
- [x] Georeferenced source: VRT still emits `<GeoTransform>` and reads as a non-`none` status.
- [x] Mixed georef / non-georef sources: rejected with a clear `ValueError`.
- [x] Full `xrspatial/geotiff/tests/` suite passes (6130 passed).

### DIFF
--- a/xrspatial/geotiff/_vrt.py
+++ b/xrspatial/geotiff/_vrt.py
@@ -1881,6 +1881,40 @@ def _check_no_mixed_raster_type(sources_meta: list[dict]) -> None:
             )
 
 
+def _check_no_mixed_georef(sources_meta: list[dict]) -> None:
+    """Reject mixing georeferenced and non-georeferenced VRT sources.
+
+    A source TIFF with no georeferencing (``has_georef`` False) carries a
+    default identity transform, not a real mapping. The mosaic-placement
+    math reads ``origin_x`` / ``pixel_width`` off each source transform to
+    position its tile, so a non-georeferenced source dropped into an
+    otherwise georeferenced mosaic would be placed at pixel-space origin
+    ``(0, 0)`` as if those identity values were real CRS coordinates.
+    Refuse rather than emit a mosaic that mislocates the tile.
+
+    The all-georeferenced and all-non-georeferenced cases both pass:
+    build_vrt emits a ``<GeoTransform>`` only when every source is
+    georeferenced, so the all-non-georeferenced VRT preserves
+    ``georef_status='none'`` on read.
+    """
+    if not sources_meta:
+        return
+    georef_flags = [bool(m.get('has_georef')) for m in sources_meta]
+    if any(georef_flags) and not all(georef_flags):
+        georeffed = [m['path'] for m, g in zip(sources_meta, georef_flags)
+                     if g]
+        plain = [m['path'] for m, g in zip(sources_meta, georef_flags)
+                 if not g]
+        raise ValueError(
+            f"VRT sources mix georeferenced and non-georeferenced TIFFs. "
+            f"Georeferenced: {georeffed!r}; non-georeferenced: {plain!r}. "
+            f"A non-georeferenced source carries only a default identity "
+            f"transform, so it cannot be placed on a georeferenced "
+            f"mosaic. Either georeference every source or none of them "
+            f"before building the VRT."
+        )
+
+
 def _nodata_values_agree(a, b) -> bool:
     """Return True iff two nodata sentinels are the same value.
 
@@ -2039,6 +2073,7 @@ def write_vrt(vrt_path: str, source_files: list[str], *,
             'bps': bps,
             'sample_format': ifd.sample_format,
             'transform': geo.transform,
+            'has_georef': geo.has_georef,
             'crs_wkt': geo.crs_wkt,
             'nodata': geo.nodata,
             'raster_type': geo.raster_type,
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: stop build_vrt fabricating a GeoTransform for non-georeferenced sources

None.

None.

- `_vrt.py` (mosaic placement, ~lines 2178-2197): when more than one non-georeferenced source is passed, every tile gets a DstRect of (0,0). The identity transforms place each source at pixel origin, so later tiles overwrite earlier ones in the same window. This is pre-existing (the old fabricated-identity path did the same) and this PR does not make it worse, so it is out of scope here. Worth a separate follow-up if stacking multiple plain tiles is a real use case, since there is no spatial info to tile them by.
- `_vrt.py` (~line 2207): for a non-georeferenced source, a caller-supplied `crs_wkt` still emits an `<SRS>` with no `<GeoTransform>`, which reads back as `crs_only`. That is a reasonable, non-fabricated result (CRS without a transform), so no change needed. Just flagging the interaction.

### What looks good
- Minimal fix on the exact mechanism: `has_georef` is tracked per source, and the `<GeoTransform>` is emitted only when sources are georeferenced.
- `_check_no_mixed_georef` matches the existing fail-loud gate style (`_check_no_rotated_source_transforms`, `_check_no_mixed_raster_type`) and names the offending sources in the error.
- Tests cover the three meaningful cases: non-georef VRT omits the transform and reads as `none`, georef VRT still emits the transform, mixed sources raise.
- The two window-transform-shift tests that depended on the old fabricated identity were re-pointed at a VRT with an explicit GeoTransform, so they still cover the shift contract.
- Full `xrspatial/geotiff/tests/` suite passes (6130 passed).

### Checklist
- Contract matches the georef_status decision table: yes
- NaN handling: not applicable (metadata-only change)
- Edge cases covered by tests: single non-georef, georef, mixed
- No premature materialization: yes (writer emits XML, no array work)
- Benchmark: not needed (metadata path)
- README feature matrix: not applicable (no new function)
- Docstrings present and accurate: yes (new gate helper documented)

PR #2977, opened 2026-06-05
geotiff: reject non-string compression at the to_geotiff boundary (#2975)

### LINKED ISSUE #2975: to_geotiff leaks a raw AttributeError for non-string compression when compression_level is set
**Describe the bug**

`to_geotiff` raises a raw `AttributeError: 'int' object has no attribute 'lower'` when you pass a non-string `compression` (say an int) together with `compression_level`.

The public wrapper only validates compression names inside an `if isinstance(compression, str):` block (`xrspatial/geotiff/_writers/eager.py`). That block raises a clear `ValueError` listing the valid options for a bad string name, but a non-string value skips it entirely. Further down, the `compression_level` validation does `compression.lower()` to look up the codec's level range, and that's where the `AttributeError` comes from.

The low-level writer (`xrspatial/geotiff/_writer.py`) already guards against this: when `compression` is neither a `str` nor `None`, it raises a `TypeError` that names the bad type. The problem is the public wrapper can blow up on `.lower()` before the call ever reaches the low-level writer.

**Reproduction**

```python
arr = xr.DataArray(np.ones((4, 4), dtype="float32"),
                   coords={"y": np.arange(4)[::-1].astype("float64"),
                           "x": np.arange(4).astype("float64")},
                   dims=["y", "x"])
to_geotiff(arr, "out.tif", compression=5, compression_level=1)
# AttributeError: 'int' object has no attribute 'lower'
```

**Expected behavior**

A non-string `compression` should be rejected at the public boundary with the same `TypeError` the low-level writer already raises, rather than leaking an `AttributeError`. The `None` default (no compression) and valid string compressions must still work.

**Additional context**

The existing string block already handles bad string names. The fix is a type guard for the non-string case so the wrapper never reaches a `.lower()` call on something that isn't a string.

brendancol:
Closes #2975

`to_geotiff` raised a raw `AttributeError: 'int' object has no attribute 'lower'` when `compression` was a non-string (say an int) and `compression_level` was set. The public wrapper validated compression names only inside an `if isinstance(compression, str):` block, so a non-string skipped validation and later hit `compression.lower()` in the `compression_level` range check.

- Add a type guard at the public boundary in `eager.py` that rejects a non-string, non-`None` `compression` with a `TypeError`, matching the low-level writer's existing guard in `_writer.py`.
- The string-name `ValueError` path and the `None` exemption are unchanged.

Backends: validation happens in the shared public wrapper before backend dispatch, so this covers numpy, cupy, dask+numpy, and dask+cupy.

Test plan:
- New regression tests assert `to_geotiff(arr, path, compression=<non-str>, compression_level=1)` and the no-level variant raise `TypeError` instead of `AttributeError`, parametrized over int/float/list/object.
- A positive test confirms the default, the `'none'` string, and a string + `compression_level` combo still write.
- Full `write/test_basic.py` passes (289 tests); flake8 clean on changed files.

### DIFF
--- a/xrspatial/geotiff/_writers/eager.py
+++ b/xrspatial/geotiff/_writers/eager.py
@@ -496,6 +496,16 @@ def to_geotiff(data: xr.DataArray | np.ndarray,
     # Up-front validation: catch bad compression names before they reach
     # any of the deeper write paths (streaming, GPU, VRT, COG) where the
     # error surfaces from _compression_tag with a less obvious traceback.
+    if not isinstance(compression, str) and compression is not None:
+        # The string block below validates bad NAMES, but a non-string
+        # ``compression`` skips it entirely and later lands in
+        # ``compression.lower()`` during compression_level validation,
+        # surfacing as ``AttributeError`` instead of a typed error.
+        # Reject the bad TYPE here with the same shape as the low-level
+        # writer's guard in ``_writer.py``.
+        raise TypeError(
+            f"compression must be a str (in to_geotiff); "
+            f"got {type(compression).__name__}.")
     if isinstance(compression, str):
         if compression.lower() not in _VALID_COMPRESSIONS:
             raise ValueError(
--- a/xrspatial/geotiff/tests/write/test_basic.py
+++ b/xrspatial/geotiff/tests/write/test_basic.py
@@ -158,6 +158,36 @@ class TestWriteInvalidInput:
         with pytest.raises(ValueError, match="(Unsupported|Unknown) compression"):
             write(arr, str(tmp_path / 'bad.tif'), compression='bzip2')
 
+    @pytest.mark.parametrize('bad', [5, 5.0, ['deflate'], object()])
+    def test_to_geotiff_non_string_compression_typeerror(self, tmp_path, bad):
+        # Regression for #2975: a non-string ``compression`` skipped the
+        # public wrapper's name-validation block and later landed in
+        # ``compression.lower()`` during compression_level validation,
+        # surfacing as a raw ``AttributeError``. The wrapper now rejects
+        # the bad type with the same ``TypeError`` shape as the low-level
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: reject non-string compression at the to_geotiff boundary (#2975)

None.

None.

- `eager.py:499` -- The guard exempts `None` (`and compression is not None`) to mirror the low-level writer's guard, which is the right call for keeping the two shapes aligned. One thing to note: `to_geotiff(arr, path, compression=None)` still leaks a downstream `AttributeError` from `_compression_tag` (`_encode.py:311`), because the signature types `compression` as `str` (default `'zstd'`) and `None` was never a supported value here. That is pre-existing and out of scope for this PR. I'm flagging it only so `None` isn't read as a value the new guard now accepts. A follow-up could either treat `None` as an alias for `'none'` or reject it at the same boundary.

### What looks good
- The guard sits before every `compression.lower()` call (the name check, the JPEG and experimental gates, and all three `compression_level` range checks), so no non-string can reach `.lower()`.
- The `TypeError` type and message match the low-level writer's guard in `_writer.py:215`, so the public and private boundaries stay consistent.
- Tests cover the reported case plus int/float/list/object, with and without `compression_level`, and a positive test confirms the default, the `'none'` string, and a string + level combo still write. Temp files carry the issue number.
- Validation runs in the shared wrapper before backend dispatch, so the fix applies to numpy, cupy, dask+numpy, and dask+cupy without per-backend changes.

### Checklist
- Algorithm matches reference/paper: n/a (validation fix)
- All implemented backends produce consistent results: yes (pre-dispatch validation)
- Edge cases covered by tests: yes (int/float/list/object, level/no-level, valid codecs)
- README feature matrix updated: not needed (no new API)
- Docstrings present and accurate: unchanged; compression already documented as str

PR #2979, opened 2026-06-05
Privatize build_vrt; route to_geotiff's VRT path through it

### LINKED ISSUE #2974: Make build_vrt private; route to_geotiff's VRT path through it
## Summary

`build_vrt` is public right now: it's in `__all__`, it's documented, and it has a row in the README feature matrix. It takes a list of existing GeoTIFF files and writes a VRT mosaic index over them. I'd like to drop it from the public API and keep it as an internal helper that only runs when `to_geotiff` writes a `.vrt` path.

## Motivation

The public GeoTIFF surface is `open_geotiff` / `to_geotiff` / `build_vrt`. The first two are read and write. `build_vrt` is the odd one out because it doesn't take a DataArray at all, it just indexes files already sitting on disk. Anyone who wants a VRT mosaic from xrspatial-written tiles already gets one by writing to a `.vrt` path with `to_geotiff`, so the second entry point adds contract surface without adding much most people need.

The giveaway: `to_geotiff`'s `.vrt` branch (`_write_vrt_tiled`) doesn't even call `build_vrt`. It calls `_vrt.write_vrt` directly. So `build_vrt` is public API with no internal caller.

## Proposed change

- Rename the wrapper `build_vrt` to `_build_vrt` and drop it from `__all__` and the package docstring.
- Point `to_geotiff`'s VRT write path (`_write_vrt_tiled`) at `_build_vrt` so there's a single internal entry point for emitting the VRT index.
- Pull `build_vrt` out of the docs reference, the README feature matrix, and the release-gate contract table.
- Update the tests that import the public name to use `_build_vrt`, and flip the API-surface assertion to expect just `open_geotiff` and `to_geotiff`.

## Compatibility

`build_vrt` is advanced-tier, not stable, so it's allowed to change between releases. Anyone calling `xrspatial.geotiff.build_vrt` directly will need to either write to a `.vrt` path through `to_geotiff` or import `_build_vrt`.

## Backend coverage

No behavior changes. VRT writing already runs on the eager and dask paths; this is an API-surface refactor.

brendancol:
## Summary

Closes #2974.

- Drops `build_vrt` from the public GeoTIFF API and renames the wrapper to `_build_vrt` (out of `__all__` and the package docstring).
- Routes `to_geotiff`'s `.vrt` write path (`_write_vrt_tiled`) through `_build_vrt`, so it's the single internal entry point for VRT-index emission. That path used to call `_vrt.write_vrt` directly, which left the public `build_vrt` with no internal caller at all.
- Updates the docs (GeoTIFF reference, safe-IO guide, release-gate table, internal notes), the README feature matrix, and the user-guide notebook's VRT section to write a mosaic through `to_geotiff('.vrt')` instead of the removed helper.
- Updates tests to import `_build_vrt`; the public-surface assertions now expect only `open_geotiff` and `to_geotiff`.

`build_vrt` is advanced-tier, not stable, so it's allowed to change between releases. Anyone who called `xrspatial.geotiff.build_vrt` either writes to a `.vrt` path through `to_geotiff` or imports `_build_vrt`.

## Backend coverage

No behavior change. VRT writing already ran on the eager and dask paths; this is an API-surface refactor. Round-trip checked: `to_geotiff(da_chunked, 'mosaic.vrt')` tiles the array and `open_geotiff` reads it back matching the original.

## Test plan

- [x] Public-surface assertions flipped to `{open_geotiff, to_geotiff}`, plus a new test that `build_vrt` is not importable while `_build_vrt` is
- [x] Every test file that imported the public name now uses `_build_vrt`
- [x] VRT round-trip through `to_geotiff('.vrt')` passes on numpy and dask
- [x] flake8 + isort clean on changed files
- [x] User-guide notebook VRT cell re-run; output regenerated from a real execution

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,15 +162,14 @@ VRT is supported as a conservative advanced feature for simple GeoTIFF mosaics, 
 | Name | Description | NumPy | Dask | CuPy GPU | Dask+CuPy GPU | Cloud |
 |:-----|:------------|:-----:|:----:|:--------:|:-------------:|:-----:|
 | [open_geotiff](xrspatial/geotiff/__init__.py) | Read GeoTIFF / COG / VRT | ✅ | ✅ | 🧪 | 🧪 | 🔼 |
-| [to_geotiff](xrspatial/geotiff/__init__.py) | Write DataArray as GeoTIFF / COG | ✅ | ✅ | 🧪 | 🧪 | 🔼 |
-| [build_vrt](xrspatial/geotiff/__init__.py) | Generate VRT mosaic from existing GeoTIFFs | 🔼 |  |  |  |  |
+| [to_geotiff](xrspatial/geotiff/__init__.py) | Write DataArray as GeoTIFF / COG / VRT | ✅ | ✅ | 🧪 | 🧪 | 🔼 |
 
 `open_geotiff` and `to_geotiff` select the backend from their parameters
 (`gpu=`, `chunks=`, `.vrt` path); GPU read/write is reached with `gpu=True`,
 not a separate function:
 
 ```python
-from xrspatial.geotiff import build_vrt, open_geotiff, to_geotiff
+from xrspatial.geotiff import open_geotiff, to_geotiff
 
 open_geotiff('dem.tif')                              # NumPy
 open_geotiff('dem.tif', chunks=512)                  # Dask
@@ -187,7 +186,7 @@ to_geotiff(data, 'cog.tif', cog=True)                # COG with auto overviews
 to_geotiff(data, 'cog.tif', cog=True,                # COG with explicit levels
            overview_levels=[2, 4, 8],
            overview_resampling='nearest')
-build_vrt('mosaic.vrt', ['tile1.tif', 'tile2.tif'])  # mosaic existing tiles
+to_geotiff(data, 'mosaic.vrt')                       # write a tiled VRT mosaic
 
 open_geotiff('dem.tif', dtype='float32')             # half memory
 open_geotiff('dem.tif', dtype='float32', chunks=512) # Dask + half memory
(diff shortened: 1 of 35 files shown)


brendancol (commented):
## PR Review: Privatize build_vrt; route to_geotiff's VRT path through it

### Blockers
None.

### Suggestions
None blocking.

### Nits
- `_build_vrt`'s docstring opening now reads `[internal-only]`, but the five per-parameter tier tags below it still say `[advanced]` (`_writers/vrt.py:58-90`). Minor internal inconsistency now that the function is no longer a public entry point; worth normalizing to `[internal-only]`.

### What looks good
- The routing change is behavior-preserving. `_build_vrt(vrt_path, tile_paths, relative=True, nodata=nodata)` resolves to `write_vrt(vrt_path, tile_paths, relative=True, crs_wkt=None, nodata=nodata)`, the same call the path made before. No deprecation warning fires (`path` positional, `crs` unset), and the local import sidesteps a circular dependency.
- Public-surface assertions are flipped correctly, and the new `test_build_vrt_is_not_public` covers both directions: the public name is gone and `_build_vrt` is still importable.
- The pre-existing local helper `_build_vrt(tmp_path)` in `test_finalization.py` would have collided with the imported `_build_vrt` after the rename; it was renamed to `_make_one_source_vrt`, and the imported call inside it was left intact.
- The notebook's VRT output cell was regenerated from a real run rather than hand-written, so the printed tile sizes and round-trip result are accurate.

### Checklist
- [x] No algorithm change (API-surface refactor)
- [x] Backends unaffected; VRT write still runs eager + dask
- [x] No NaN / edge-case concerns
- [x] Routing change verified behavior-preserving
- [x] Docstrings present (one tier-tag nit above)

brendancol (commented):
## Follow-up review

The one nit from the first pass is fixed in d04a2526: the five per-parameter tier tags in `_build_vrt`'s docstring now read `[internal-only]`, matching the function summary. `test_polish.py::TestC5WriteVrtKwargs` still passes (the docstring keeps `relative` / `crs_wkt` / `nodata`).

No new findings. The change is docstring-only; no code paths moved.

PR #2980, opened 2026-06-05
geotiff: align direct backend mask_nodata default with open_geotiff

### LINKED ISSUE #2976: Direct reader backends default to masking nodata while open_geotiff defaults to unmasked
## Describe the bug

The direct reader backends default to masking nodata, but the public `open_geotiff` defaults to unmasked reads. Call a backend directly with no `mask_nodata` argument and you get a different dtype and different values than `open_geotiff` returns. The divergence is silent.

`open_geotiff` sets `masked: bool = False` and documents rioxarray-compatible unmasked reads. The wrapper forwards `mask_nodata=masked` to the backends explicitly on every dispatch path (GPU, dask, VRT, eager). But the three backend entry points each keep `mask_nodata: bool = True` as their own default:

- `_read_geotiff_dask` in `xrspatial/geotiff/_backends/dask.py`
- `_read_geotiff_gpu` in `xrspatial/geotiff/_backends/gpu.py`
- `_read_vrt` in `xrspatial/geotiff/_backends/vrt.py`

## To Reproduce

A public chunked read returns int16 with -9999 preserved (unmasked). A direct `_read_geotiff_dask()` call on the same file returns float64 with NaN (masked). Same split for `_read_geotiff_gpu` and `_read_vrt`.

## Expected behavior

A bare backend call with no `mask_nodata` argument should match `open_geotiff`: keep the source dtype, preserve the nodata sentinel, don't substitute NaN. The three backend defaults should be `False`.

## Additional context

The public wrapper always passes `mask_nodata=masked` explicitly, so the default only affects direct callers. The internal callers (GPU-to-dask fallback, dask-to-VRT dispatch) forward the value explicitly too, so flipping the default only touches code that calls a backend directly without the kwarg.

brendancol:
Closes #2976

## What

The three direct GeoTIFF backend entry points defaulted to `mask_nodata=True` while the public `open_geotiff` defaults to `masked=False`. A bare backend call returned a different dtype and NaN-substituted values than the public path, silently.

- Flip `_read_geotiff_dask`, `_read_geotiff_gpu`, and `_read_vrt` to `mask_nodata=False` so a direct call with no `mask_nodata` argument matches `open_geotiff` (unmasked, rioxarray-compatible).
- The public wrapper and the internal GPU-to-dask / dask-to-VRT callers all forward `mask_nodata` explicitly, so only direct callers are affected.

## Backend coverage

The dask and VRT entry points cover numpy and cupy graphs; the GPU entry point covers eager and dask+cupy. The default flip applies the same way on all of them.

## Test plan

- [x] New regression tests: each direct backend with no `mask_nodata` produces the same unmasked result (dtype + sentinel preserved, no NaN) as `open_geotiff`. The GPU test is guarded by `requires_gpu`.
- [x] Updated existing tests that relied on the old masked default. Where a test verifies the masking path it now passes `mask_nodata=True`; where a test pinned the bare-call behavior it asserts the new unmasked default.
- [x] Full geotiff suite green locally (6155 passed, GPU included).
- [x] flake8 clean on touched files.

### DIFF
--- a/xrspatial/geotiff/_backends/dask.py
+++ b/xrspatial/geotiff/_backends/dask.py
@@ -40,7 +40,7 @@ def _read_geotiff_dask(source: str, *,
                        allow_experimental_codecs: bool = False,
                        allow_internal_only_jpeg: bool = False,
                        band_nodata: str | None = None,
-                       mask_nodata: bool = True,
+                       mask_nodata: bool = False,
                        mask_and_scale: bool = False,
                        parse_coordinates: bool = True) -> xr.DataArray:
     """Read a GeoTIFF as a dask-backed DataArray for out-of-core processing.
@@ -99,14 +99,15 @@ def _read_geotiff_dask(source: str, *,
         mixed-band-metadata check. Forwarded
         verbatim to ``_read_vrt`` when the source is a ``.vrt`` file.
         Passing it with a non-VRT GeoTIFF source raises ``ValueError``.
-    mask_nodata : bool, default True
+    mask_nodata : bool, default False
         [stable] If True, replace the nodata sentinel with NaN per
         chunk (integer rasters get promoted to ``float64``). If False,
         skip the sentinel-to-NaN step so the source dtype survives.
         The raw sentinel is still carried on ``attrs['nodata']``
-        either way. Pass ``mask_nodata=False`` together with
-        ``dtype=<integer>`` to keep an integer source dtype; the
-        default promotes to ``float64`` and the cast then raises.
+        either way. The default matches ``open_geotiff`` (unmasked,
+        rioxarray-compatible): a bare ``_read_geotiff_dask(path)`` keeps
+        the source dtype and leaves the sentinel pixels untouched. Pass
+        ``mask_nodata=True`` to restore the promote-to-NaN behaviour.
     mask_and_scale : bool, default False
         [advanced] If True, apply the source's GDAL ``SCALE`` / ``OFFSET``
         (``data * scale + offset``) lazily on the assembled dask array and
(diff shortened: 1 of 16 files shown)


brendancol (commented):
## PR Review: geotiff: align direct backend mask_nodata default with open_geotiff

None.

One note, not blocking. The internal helpers that aren't public entry points kept `mask_nodata: bool = True`: `_read_geotiff_gpu_chunked_gds` in gpu.py, plus `_vrt_chunk_read` and the chunked dispatch in vrt.py. That's correct. They always get an explicit `mask_nodata` forwarded from the entry point and aren't a caller's direct surface. The tests in test_reader.py and test_kernels_and_kwargs.py now pass `mask_nodata=True` to both sides when they compare an entry point against an internal helper, so those comparisons stay apples-to-apples. Flagging it only so the next reader doesn't mistake the asymmetry for an oversight.

None.

### What looks good
- The source change is small: three default flips and the matching `default True` to `default False` docstring edits. Everything else in the diff is test updates.
- The public wrapper forwards `mask_nodata=masked` on every dispatch path (GPU, dask, VRT, eager), and the two internal cross-backend callers (gpu.py:1422 to `_read_geotiff_dask`, dask.py:227 to `_read_vrt`) forward it explicitly. So the default only affects direct callers. Confirmed by reading the call sites.
- The new regression tests check direct-backend-vs-`open_geotiff` parity for all three backends (GPU guarded by `requires_gpu`): dtype, sentinel survival, no NaN, bit-equal pixels.
- Existing tests that pinned the old masked default were split the right way. Masking-path tests now request `mask_nodata=True`; default-behavior tests assert the new unmasked default and were renamed for #2976.
- Full geotiff suite passes locally including GPU (6155 passed).

### Checklist
- [x] Behavior change matches the issue: direct backend default now matches open_geotiff
- [x] All three backends produce consistent unmasked results with no mask_nodata arg
- [x] NaN / sentinel handling correct: masking still available via mask_nodata=True
- [x] Edge cases covered: int sentinel hit, float sentinel, dtype preservation, dask + GPU graphs
- [x] Internal helper defaults left untouched (correct: they receive explicit values)
- [x] No premature materialization or new copies (no compute path changed)
- [x] Benchmark not needed (no perf-critical path changed)
- [x] README feature matrix not applicable (no new public function)
- [x] Docstrings updated to default False on all three backends

PR #2982, opened 2026-06-06
Surface standalone docstrings on .xrs accessor methods; fix broken hydro delegations

### LINKED ISSUE #2981: help() on .xrs accessor methods shows no docstring (and 17 hydro accessor methods are broken)
## What happens

`help(arr.xrs.slope)` shows a bare `slope(**kwargs)` with no parameter docs, while `help(slope)` (from `from xrspatial import slope`) shows the full docstring. The `.xrs` accessor methods are thin wrappers that delegate to the standalone functions but carry no docstrings of their own, so anyone exploring the API through the accessor gets nothing useful.

## What I'd expect

`help(arr.xrs.slope)` should show the same documentation as `help(slope)` — the parameters, notes, and examples.

## A bug I hit while looking into this

17 hydrology accessor methods raise `ModuleNotFoundError` when called. They still import from per-algorithm modules (`from .fill import fill`, `.watershed`, `.basin`, and so on) that were consolidated into `xrspatial/hydro.py`. Nothing in the accessor tests calls them, so it slipped through.

Affected: `flow_direction`, `flow_direction_dinf`, `flow_direction_mfd`, `flow_accumulation`, `flow_accumulation_mfd`, `watershed`, `basin`, `basins`, `sink`, `fill`, `stream_order`, `stream_link`, `snap_pour_point`, `flow_path`, `flow_length`, `twi`, `hand`.

```python
import numpy as np, xarray as xr, xrspatial
da = xr.DataArray(np.abs(np.random.rand(8, 8)) * 100, dims=['y', 'x'],
                  coords={'y': np.arange(8.), 'x': np.arange(8.)})
da.xrs.fill()   # ModuleNotFoundError: No module named 'xrspatial.fill'
```

## Fix

Repoint the broken hydrology delegations at `xrspatial.hydro`, then copy each standalone function's docstring onto its accessor method so `help()` works. The hydrology unified wrappers only carry a generic dispatcher docstring, so for those, take the help text from the documented default-algorithm `*_d8` variants instead. Add accessor tests for both the docstrings and the hydrology calls.

brendancol:
Closes #2981

`help(arr.xrs.slope)` showed nothing useful because the `.xrs` accessor methods are thin delegating wrappers with no docstrings of their own. This copies each standalone function's docstring onto its accessor method at import time, so `help(arr.xrs.slope)` now matches `help(slope)`.

While adding that, I found 17 hydrology accessor methods were broken: they imported from per-algorithm modules (`from .fill import fill`, `.watershed`, and so on) that had been consolidated into `xrspatial/hydro.py`, so calling them raised `ModuleNotFoundError`. Nothing in the accessor tests called them, so it had gone unnoticed. This PR repoints those delegations at `.hydro` too.

## Changes

- Repoint the 17 broken hydrology delegations (`fill`, `flow_direction`, `flow_accumulation`, `watershed`, `basin`, `basins`, `sink`, `stream_order`, `stream_link`, `snap_pour_point`, `flow_path`, `flow_length`, `twi`, `hand`, plus the dinf/mfd variants) at `xrspatial.hydro`, in both the DataArray and Dataset accessors.
- Attach standalone docstrings to accessor methods. Methods that already had hand-written docs (`plot`, `to_geotiff`, `open_geotiff`) are left alone. The hydrology unified wrappers only carry a generic dispatcher stub, so their help text is taken from the documented default-algorithm `*_d8` variants.

Doc-only — no `__signature__` is set, since several methods (`zonal_stats`, `crop`, `idw`/`kriging`/`spline`, `mahalanobis`) don't pass the accessor's array as the function's first argument, which would make a naive signature wrong. The copied Parameters section already documents everything.

## Backends

No backend-specific code. The docstring attachment is pure Python metadata, and the hydrology import fix routes through the same `xrspatial.hydro` functions that already support numpy / cupy / dask+numpy / dask+cupy.

## Test plan

- [x] `pytest xrspatial/tests/test_accessor.py` (64 passed)
- [x] Docstring surfacing matches the source function (slope, aspect, hillshade, curvature, focal_mean->mean, ndvi, fill->fill_d8, watershed->watershed_d8)
- [x] Drift guard: every public method on both accessor classes carries a non-stub docstring
- [x] Regression: all 17 hydrology delegations resolve; the 6 single-input ones match their standalone function

Not touched: no new public API, so no user-guide notebook, README matrix row, or Sphinx reference page. The accessor has no reference page today; that's a separate docs task.

### DIFF
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -1333,3 +1333,60 @@ class XrsSpatialDatasetAccessor:
     def rechunk_no_shuffle(self, **kwargs):
         from .utils import rechunk_no_shuffle
         return rechunk_no_shuffle(self._obj, **kwargs)
+
+
+# ---------------------------------------------------------------------------
+# Surface standalone-function docstrings on accessor methods so that, e.g.,
+# ``help(da.xrs.slope)`` shows the same documentation as ``help(slope)``.
+# ---------------------------------------------------------------------------
+
+# Accessor method name -> name of the standalone function (in the top-level
+# ``xrspatial`` namespace) whose docstring should be surfaced. Only needed
+# when the method name differs from the function name, or when the direct
+# delegate's docstring is a generic dispatcher: the hydrology unified wrappers
+# route by ``routing=`` and carry only a stub docstring, so their help text is
+# taken from the documented default-algorithm (``*_d8``) variants instead.
+_DOC_SOURCE_OVERRIDES = {
+    'focal_mean': 'mean',
+    'zonal_hypsometric_integral': 'hypsometric_integral',
+    'fill': 'fill_d8',
+    'flow_direction': 'flow_direction_d8',
+    'flow_accumulation': 'flow_accumulation_d8',
+    'basin': 'basin_d8',
+    'basins': 'basins_d8',
+    'watershed': 'watershed_d8',
+    'snap_pour_point': 'snap_pour_point_d8',
+    'flow_path': 'flow_path_d8',
+    'flow_length': 'flow_length_d8',
+    'sink': 'sink_d8',
+    'stream_link': 'stream_link_d8',
+    'stream_order': 'stream_order_d8',
+    'twi': 'twi_d8',
+    'hand': 'hand_d8',
+}
+
+
+def _delegated_doc(method_name):
+    """Return the docstring to surface for an accessor method, or None."""
+    if method_name == 'min_observable_height':
+        from .experimental.min_observable_height import min_observable_height
+        return min_observable_height.__doc__
+    import xrspatial
+    source_name = _DOC_SOURCE_OVERRIDES.get(method_name, method_name)
+    return getattr(getattr(xrspatial, source_name, None), '__doc__', None)
+
+
+def _attach_delegated_docs(cls):
+    for name, member in vars(cls).items():
+        if name.startswith('_') or not callable(member) or member.__doc__:
+            continue
+        try:
+            doc = _delegated_doc(name)
+        except Exception:
+            continue
+        if doc:
+            member.__doc__ = doc
+
+
+for _cls in (XrsSpatialDataArrayAccessor, XrsSpatialDatasetAccessor):
+    _attach_delegated_docs(_cls)
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -1373,7 +1373,8 @@ def _delegated_doc(method_name):
         return min_observable_height.__doc__
     import xrspatial
     source_name = _DOC_SOURCE_OVERRIDES.get(method_name, method_name)
-    return getattr(getattr(xrspatial, source_name, None), '__doc__', None)
+    func = getattr(xrspatial, source_name, None)
+    return func.__doc__ if func is not None else None
 
 
 def _attach_delegated_docs(cls):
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Surface standalone docstrings on .xrs accessor; fix broken hydro delegations

Reviewed the full diff and both changed files in the worktree. No blockers. The hydro import fix is verified end-to-end and the docstring mechanism is sound. A couple of robustness items in the new code are worth tightening.

- `xrspatial/accessor.py`, `_delegated_doc`: the lookup `getattr(getattr(xrspatial, source_name, None), '__doc__', None)` has a sharp edge. If a doc-source name is ever missing (a typo in `_DOC_SOURCE_OVERRIDES`, or a future rename), the inner `getattr` returns `None`, and `None.__doc__` is the truthy string `"The type of the None singleton."` — which then gets attached as the method's docstring. It doesn't trigger today (every name resolves, drift guard is green), but it fails silently and wrong rather than skipping. Resolve the function first and read `__doc__` only when it's not `None`: `func = getattr(xrspatial, source_name, None); return func.__doc__ if func is not None else None`.

- Dataset accessor multispectral methods (`ndvi`, `evi`, `savi`, etc.) now show the standalone function's docstring, which documents array inputs (`nir_agg`, `red_agg`). The Dataset accessor actually takes band-name strings (`ds.xrs.ndvi(nir='b8', red='b4')`), so the surfaced parameter docs don't match the call convention. The class docstring already explains the band-mapping, so this is informative-but-imperfect rather than wrong. Worth a note, or leave as-is given the goal was parity with the standalone help.

### Nits (optional)

- `xrspatial/accessor.py`: the module-level `for _cls in (...)` loop leaves `_cls` bound in the module namespace. Harmless; `del _cls` after the loop or a comprehension would keep it clean.
- `xrspatial/tests/test_accessor.py`: `import inspect` and the accessor-class imports sit mid-file with `# noqa: E402`. Fine and flake8-clean, but they could move to the top with the other imports.

### What looks good

- The hydro fix is the right root cause: the 17 methods imported from per-algorithm modules that were folded into `xrspatial.hydro`, and repointing the lazy imports there fixes all of them. Verified no remaining broken imports and that the six single-input methods match their standalone results.
- Docstrings are sourced from the documented `*_d8` variants for the hydro wrappers, avoiding the generic dispatcher stub.
- Tests compare against real reference functions (not just "it runs"), and the drift guard will catch any future accessor method added without a usable docstring.
- Hand-written docs on `plot`/`to_geotiff`/`open_geotiff` are correctly left untouched.

### Checklist
- [x] Behavior matches intent (help text parity; hydro calls succeed)
- [x] No backend-specific code; hydro fix routes through existing multi-backend functions
- [x] Edge cases covered by tests (delegation resolves, doc surfacing, drift guard)
- [x] Docstrings present and accurate (one Dataset-multispectral caveat above)
- [ ] Latent None.__doc__ fallback worth tightening (Suggestion 1)

brendancol (commented):
## Follow-up review (after fixes in 01a5348)

Re-read the diff after the follow-up commit. The earlier findings are resolved; nothing new surfaced.

- Suggestion 1 (None.__doc__ fallback) — fixed. `_delegated_doc` now resolves the source function first and returns `None` when it's missing, so a bad override name no longer attaches `"The type of the None singleton."`. Verified with a bogus override.
- Nit 1 (`_cls` loop leak) — fixed. Replaced with two explicit `_attach_delegated_docs(...)` calls.
- Nit 2 (mid-file test imports) — fixed. `inspect` and the accessor-class imports moved to the top of the test module.
- Suggestion 2 (Dataset multispectral docstrings describe array inputs, not the band-name-string API) — kept as-is. Sourcing the standalone docstring is the stated goal, the Dataset accessor class docstring already explains the band mapping, and bespoke per-method Dataset docs would be a larger change for marginal gain. Tracking it as a known imperfection rather than a fix here.

Tests: 64 passed, flake8 clean on both changed files.

PR #2985, opened 2026-06-06
geotiff: accumulate GPU overview mean in float64 for CPU byte parity (#2983)

### LINKED ISSUE #2983: GPU float32 COG overview mean diverges from CPU writer (float32 vs float64 accumulation)
## Describe the bug

The GPU COG writer's overview "mean" reduction doesn't match the CPU writer on float32 input. `_block_reduce_2d_gpu` in `xrspatial/geotiff/_gpu_decode.py` calls `cupy.nanmean(blocks, axis=(1, 3))` on a float32 `blocks` array, so each block mean accumulates in float32. The CPU path (`_reduce2x2_mean` in `_overview_kernels.py`, via `_block_reduce_2d`) accumulates the same block in float64 and downcasts to float32 only on store.

So the two writers emit different overview bytes for float32 rasters. That contradicts the GPU writer's stated contract: `_writers/gpu.py` says it "produces the same overview bytes as the CPU writer".

## To Reproduce

On a 64x64 float32 raster, 307 of 1024 overview pixels differ between the CPU and GPU mean reductions, max absolute difference about 1.2e-4. Only `mean` on float32 is affected. `min`, `max`, and `median` are exact and already match, and float64 input already matches because both paths accumulate in float64 there.

## Expected behavior

GPU overview mean byte-identical to the CPU overview mean for the same input, per the documented parity contract.

## Fix

Pass `dtype=cupy.float64` to the `nanmean` call so the GPU accumulates in float64 and downcasts on the existing final `astype(arr2d.dtype)`. Verified byte-identical to CPU across 64x64, odd 33x65, and 128x128 float32 inputs.

Found by the accuracy sweep.

brendancol:
Closes #2983

- `_block_reduce_2d_gpu` (`xrspatial/geotiff/_gpu_decode.py`) ran `cupy.nanmean` on a float32 `blocks` array, so the mean accumulated in float32. The CPU kernel accumulates in float64 and downcasts on store, so the two writers emitted different overview bytes for float32 rasters. That broke the GPU writer's documented byte-parity contract.
- Pass `dtype=cupy.float64` to the `nanmean` call so the GPU also downcasts a float64 accumulator on the existing final `astype`. `min` / `max` / `median` are exact and unchanged; float64 input was already correct on both paths.

Backend coverage: GPU (cupy) only. The change is isolated to the GPU overview reducer; the CPU and dask paths are untouched and serve as the reference.

Test plan:
- [x] New `test_gpu_overview_reduce_byte_matches_cpu_2983` asserts CPU/GPU byte parity for mean/min/max/median across 64x64, odd 33x65, and 128x128 float32 inputs
- [x] Confirmed the test fails on the `mean` cases without the fix and passes with it
- [x] Existing GPU overview / COG tests still pass

### DIFF
--- a/xrspatial/geotiff/_gpu_decode.py
+++ b/xrspatial/geotiff/_gpu_decode.py
@@ -3251,7 +3251,16 @@ def _block_reduce_2d_gpu(arr2d, method, nodata=None):
                         mask, cupy.float64('nan'), blocks)
 
     if method == 'mean':
-        result = cupy.nanmean(blocks, axis=(1, 3))
+        # Accumulate the block mean in float64 then downcast on the final
+        # ``astype(arr2d.dtype)`` below, matching the CPU kernel
+        # (``_reduce2x2_mean`` in ``_overview_kernels``) which sums into a
+        # float64 Python accumulator. Without ``dtype=cupy.float64`` a
+        # float32 ``blocks`` array makes ``cupy.nanmean`` accumulate in
+        # float32, so the GPU and CPU writers emit different overview bytes
+        # for float32 rasters (the GPU writer documents byte parity with
+        # the CPU writer). float64 input already accumulates in float64 on
+        # both paths, so this only changes the float32 result.
+        result = cupy.nanmean(blocks, axis=(1, 3), dtype=cupy.float64)
     elif method == 'min':
         result = cupy.nanmin(blocks, axis=(1, 3))
     elif method == 'max':
--- a/xrspatial/geotiff/tests/gpu/test_writer.py
+++ b/xrspatial/geotiff/tests/gpu/test_writer.py
@@ -1512,6 +1512,38 @@ def test_gpu_writer_overview_uses_make_overview_gpu_fresh_buffer_1948():
         )
 
 
+@_gpu_only
+@pytest.mark.parametrize("shape", [(64, 64), (33, 65), (128, 128)])
+@pytest.mark.parametrize("method", ["mean", "min", "max", "median"])
+def test_gpu_overview_reduce_byte_matches_cpu_2983(shape, method):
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: geotiff: accumulate GPU overview mean in float64 for CPU byte parity (#2983)

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- Root cause is right. `cupy.nanmean` on a float32 array accumulates in float32, while the CPU kernel `_reduce2x2_mean` accumulates in float64 and downcasts on store. Adding `dtype=cupy.float64` makes the GPU downcast a float64 accumulator on the existing final `astype`, so the two paths match.
- The fix is one line plus a comment, scoped to the `mean` branch. `min`/`max`/`median` are exact and left alone, which the test confirms.
- I checked the integer mean path too. The CPU integer path already accumulates in float64 (`arr2d.astype(np.float64)` then `np.nanmean`), so the GPU change keeps integer means consistent rather than opening a new divergence. Verified int32 mean is byte-identical after the change.
- The test fails on the three `mean` cases without the fix and passes with it, so it guards the behavior instead of just running it.
- The odd-shape case (33x65) hits the NaN-padded residual-block path, which is where accumulation order matters most.

### Checklist
- [x] Algorithm matches reference (CPU kernel is the reference; byte parity verified)
- [x] All implemented backends produce consistent results (CPU/GPU mean now byte-identical for float32 and int32)
- [x] NaN handling correct (unchanged; all-NaN block still yields NaN for the caller's rewrite)
- [x] Edge cases covered (regular + odd shapes; mean/min/max/median)
- [ ] Dask chunk boundaries -- N/A (the overview reducer is not a dask map_overlap)
- [x] Benchmark -- not needed (no perf change)
- [x] README feature matrix -- N/A (no new function)

PR #2986, opened 2026-06-06
geotiff: cover GPU writer degenerate shapes (1x1/1xN/Nx1)

### LINKED ISSUE #2984: geotiff: GPU writer lacks degenerate-shape (1x1 / 1xN / Nx1) round-trip coverage
The GPU writer for the geotiff subpackage has no degenerate-shape tests. The read side already covers 1x1 / 1xN / Nx1 sources on all four backends (`xrspatial/geotiff/tests/read/test_degenerate_shapes.py`), and the dask streaming writer covers them too (`xrspatial/geotiff/tests/integration/test_dask_pipeline.py`). The GPU write path is the gap left over: the smallest shape in `xrspatial/geotiff/tests/gpu/test_writer.py` is 2x2.

Two paths are missing coverage:

- `_write_geotiff_gpu`, the direct CuPy writer
- `to_geotiff(gpu=True)` over a dask+cupy DataArray, the dask+gpu write path

A single-pixel dimension is exactly the kind of input that stresses GPU tile padding and the encode kernels. The read side needed a fix for this same class recently (commit 04a16d47, "cover degenerate-shape reads on dask and GPU backends").

I ran both paths on a CUDA host against 1x1, 1xN, and Nx1 inputs with `none` and `deflate` compression, supplying an explicit `attrs['transform']` since pixel size cannot be inferred from a single-element coord axis. Everything round-tripped. So this is a coverage gap, not a bug, and the PR is test-only.

Scope:

- Add GPU-writer degenerate-shape round-trip tests under `xrspatial/geotiff/tests/gpu/`.
- Guard with the existing `requires_gpu` marker so non-CUDA hosts skip.
- No source changes.

brendancol:
Closes #2984.

Adds degenerate-shape (1x1 / 1xN / Nx1) round-trip coverage for the GPU write path. The read side and the dask streaming writer already cover these shapes; the GPU writer was the remaining gap (the smallest shape elsewhere in `gpu/test_writer.py` was 2x2).

- `test_write_geotiff_gpu_degenerate_round_trip`: direct `_write_geotiff_gpu`, parametrized over 1x1 / 1xN / Nx1 and `none` / `deflate` compression.
- `test_to_geotiff_dask_gpu_degenerate_round_trip`: the dask+cupy write path via `to_geotiff(gpu=True)`, parametrized over the three shapes.

Both supply an explicit `attrs['transform']` since pixel size cannot be inferred from a single-element coord axis, matching the read-side degenerate fixture. Tests use the existing `requires_gpu` marker so non-CUDA hosts skip.

No source changes. I checked that the GPU and dask+gpu write paths already handle these shapes before writing the tests, so this is coverage, not a fix.

Backend coverage: cupy (direct GPU writer) and dask+cupy (dask+gpu write path). The numpy eager writer and dask+numpy streaming writer already had this coverage.

Test plan:
- 9 new tests run and pass on a CUDA host.
- On a non-GPU host they skip via `requires_gpu`.

This PR also normalizes the sweep state CSV to one physical line per record, fixing pre-existing union-merge newline corruption.


brendancol (commented):
## PR Review: geotiff, cover GPU writer degenerate shapes (1x1/1xN/Nx1)

Test-only PR. No source changes, so the review is about whether the new tests are valid and have teeth, not about algorithm changes.

### Blockers
None.

### Suggestions
None blocking. One optional note below.

### Nits
- The dask+gpu test only exercises `compression="none"`, while the direct-writer test covers both `none` and `deflate`. Adding `deflate` to the dask+gpu parametrization would close the small asymmetry, but it is not required: the codec path is shared and already covered on the direct writer.

### What looks good
- The gap is real and the scope is right. Read-side degenerate shapes already cover all four backends (`read/test_degenerate_shapes.py`) and the dask streaming writer covers them (`integration/test_dask_pipeline.py`); the GPU write path was the only hole.
- The explicit `attrs['transform']` matches how the read-side degenerate fixture handles the single-element coord axis, so the two suites stay consistent.
- Assertions check both shape and pixel values via `assert_array_equal`, so a dropped window or a wrong-shape write would fail. I confirmed the shape mismatch raises.
- Temp files include the issue number and the shape/compression id, so no collisions across parametrized cases or parallel worktrees.
- Tests use the existing `requires_gpu` marker, so non-CUDA hosts skip rather than error.
- The state CSV normalization to one physical line per record fixes the pre-existing union-merge newline corruption.

### Checklist
- All implemented backends consistent: cupy + dask+cupy write paths now covered; numpy/dask+numpy already covered
- NaN handling correct: n/a (no nodata in fixtures)
- Edge cases covered: yes, that is the point of the PR
- Dask chunk boundaries: dask+cupy path exercised on degenerate chunks
- Benchmark: not needed
- README matrix: not needed (no new public function)
- Docstrings present: yes on every new test

brendancol (commented):
## Follow-up review

The one nit from the first pass is addressed in 357b454: the dask+gpu degenerate test is now parametrized over `none` and `deflate`, matching the direct-writer test. The degenerate suite is 12 tests now (3 shapes x 2 codecs for the direct writer, 3 shapes x 2 codecs for dask+gpu), all passing on a CUDA host.

No remaining findings. Test-only change, no source touched.

PR #2991, opened 2026-06-06
Surface chunked VRT decode-time hole gap under missing_sources='warn' (#2989)

### LINKED ISSUE #2989: Chunked VRT missing_sources='warn' under-reports decode-time holes in attrs['vrt_holes']
## Describe the bug

Reading a VRT with `chunks=` and `missing_sources='warn'` does not reliably record decode-time holes in `attrs['vrt_holes']`.

The chunked path (`xrspatial/geotiff/_backends/vrt.py`, around line 1160) pre-populates `attrs['vrt_holes']` from a parse-time `os.path.exists` sweep over every source. That sweep only catches sources whose backing file is missing on disk. A source file that exists but is corrupt, truncated, or fails to decode (zlib/zstd codec error, malformed IFD) passes the existence check, so it never lands in `attrs['vrt_holes']` at build time.

The eager path finds these holes at decode time. It catches `(OSError, ValueError, struct.error)` plus the codec decode exceptions per source and appends each to `vrt.holes`, which `read_vrt` surfaces as `attrs['vrt_holes']`. Under chunked dispatch each per-task decode catches its own failure and warns from the worker, but those records cannot be reduced back onto the parent DataArray's attrs.

So a caller can read a chunked VRT under `missing_sources='warn'`, get a DataArray whose `vrt_holes` attr lists only the missing-file holes (or is absent entirely), branch on `"vrt_holes" in da.attrs` the way the docstring suggests, then compute and still get a partial mosaic if the compute-time warnings are filtered or missed.

## Expected behavior

The chunked `attrs['vrt_holes']` contract should be documented as statically-detectable (missing-file) holes only, so callers stop treating the absence of the attr as proof of a complete mosaic. On top of that, a lenient chunked read should tell the caller up front that decode-time codec failures on existing sources surface as per-task warnings at compute and will not appear in `attrs['vrt_holes']`.

Forcing an eager decode of every source at graph-build time would surface every hole but defeats lazy reading, so that is off the table.

## Additional context

- Eager decode-time hole catch: `xrspatial/geotiff/_vrt.py` around line 1599.
- Chunked static sweep and raise guard: `xrspatial/geotiff/_backends/vrt.py` around line 1160.
- The existing `missing_sources='raise'` chunked guard already raises up front for missing files. The same silent-partial-mosaic risk applies to a `'raise'` caller whose corrupt-but-existing source only raises inside a worker task that may never run.

## Tradeoff

A documentation-only fix is cheap and honest but leaves the gap open. A build-time heads-up warning under `'warn'` keeps the lazy contract intact and points the caller at where to look, at the cost of one extra warning on every lenient chunked read that has decode-capable sources. Forcing eager decode is rejected.

brendancol:
Closes #2989

## What

The chunked VRT path pre-populates `attrs['vrt_holes']` from a parse-time `os.path.exists` sweep, which only catches sources whose backing file is missing. A source that exists but fails to decode at compute time (corrupt, truncated, codec error) reads as a hole inside the per-chunk worker and is warned there, but that record cannot be reduced back onto the parent DataArray's attrs without eagerly decoding every source. Eager decode would defeat lazy reading, so it is off the table.

This PR makes the gap explicit instead of silent:

- Documents `attrs['vrt_holes']` under chunked dispatch as a lower bound (statically-detectable missing-file holes only), in the `_read_vrt` docstring, the inline static-sweep comment, and `docs/source/reference/geotiff.rst`.
- Emits one `GeoTIFFFallbackWarning` up front under `missing_sources='warn'` when the requested window touches an existing (decode-capable) source, so a caller cannot treat the absence of the attr as proof of a complete mosaic. The warning is suppressed when no in-window source exists on disk, since nothing can decode-fail in that case.

No behavior change for `missing_sources='raise'` (still fails closed up front on missing files) or for the eager path.

## Backend coverage

VRT reads are CPU/dask only and do not go through the GPU decoder pipeline; the warning lands on the chunked dispatcher shared by dask+numpy and dask+cupy. No numpy/cupy eager-path change.

## Test plan

- [x] Present-but-corrupt source is absent from build-time `attrs['vrt_holes']`
- [x] Build-time heads-up `GeoTIFFFallbackWarning` fires under chunked `'warn'`
- [x] Corrupt source still warns at compute time
- [x] Heads-up suppressed when no in-window source exists on disk
- [x] Full `xrspatial/geotiff/tests/vrt/` suite green (502 passed)

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -510,6 +510,20 @@ integer bands without a sentinel). ``XRSPATIAL_GEOTIFF_STRICT=1``
 forces the raise in ``'warn'`` mode too, so CI environments can enforce
 fail-fast behavior globally.
 
+Under chunked dispatch (``chunks=`` set), ``attrs['vrt_holes']`` records
+statically-detectable missing-file holes only (#2989). A source file
+that exists but fails to decode at compute time -- corrupt, truncated,
+or a codec error -- reads as a hole in the per-chunk worker and emits a
+:class:`xrspatial.geotiff.GeoTIFFFallbackWarning` then, but cannot be
+reduced back onto the parent DataArray's ``attrs['vrt_holes']`` without
+eagerly decoding every source (which would defeat lazy reading). So the
+chunked ``attrs['vrt_holes']`` is a lower bound, not the eager reader's
+exhaustive list. The chunked ``'warn'`` path emits one
+``GeoTIFFFallbackWarning`` up front when the requested window touches any
+existing source, so do not treat the absence of ``attrs['vrt_holes']``
+as proof of a complete mosaic. Use ``missing_sources='raise'`` to fail
+closed if you need a hard guarantee.
+
 BigTIFF COG (issue #2303)
 =========================
 
--- a/xrspatial/geotiff/_backends/vrt.py
+++ b/xrspatial/geotiff/_backends/vrt.py
@@ -222,6 +222,17 @@ def _read_vrt(source: str, *,
         sentinel (or zero on integer bands without a sentinel).
         ``XRSPATIAL_GEOTIFF_STRICT=1`` forces a raise across the whole
         module regardless of this kwarg.
+        Under chunked dispatch (``chunks=`` set) ``attrs['vrt_holes']``
+        records statically-detectable missing-file holes only. A source
+        file that exists but fails to decode at compute time (corrupt,
+        truncated, or codec error) surfaces as a per-chunk
+        ``GeoTIFFFallbackWarning`` and does NOT appear in
+        ``attrs['vrt_holes']``; recording it would require eagerly
+        decoding every source, which defeats lazy reading. The chunked
+        ``'warn'`` path emits one ``GeoTIFFFallbackWarning`` up front
+        when the requested window touches any existing source, so do not
+        treat the absence of ``attrs['vrt_holes']`` as proof of a
+        complete mosaic.
     band_nodata : {'first', None}, optional
         [advanced] Opt-out for the fail-closed mixed-band-metadata
         check. ``None`` (the default) rejects a VRT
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Surface chunked VRT decode-time hole gap under missing_sources='warn' (#2991)

None.

- `xrspatial/geotiff/_backends/vrt.py:1290-1308` -- the `decode_capable` scan re-walks `vrt.bands` and every source with the same window-intersection test the static-hole loop at line ~1185 already runs. On a mosaic with many tiles that is a second full pass. Consider computing `decode_capable` inside the existing static-hole loop: a source that intersects the window and exists is decode-capable, one that intersects and is missing is a hole. That avoids the duplicate traversal at the cost of a slightly busier loop body. Not blocking: the scan breaks on the first existing in-window source, so the common case is cheap.

- `xrspatial/geotiff/_backends/vrt.py:1310-1312` -- `import warnings` and `from .._runtime import GeoTIFFFallbackWarning` sit inside the `if decode_capable` block. That matches the eager path's local-import style in `_vrt.py:1621`, so it is consistent. A module-level import would be marginally cleaner; leave it if matching the existing pattern is the intent.

### What looks good
- The window-intersection predicate is copied verbatim from the static-hole guard, so the heads-up and the attrs population agree on what "in window" means.
- The warning is gated on `missing_sources == 'warn'` and sits after the raise guard, so `'raise'` and strict mode are untouched.
- Suppressing the warning when no in-window source exists on disk keeps it from firing on all-missing VRTs where nothing can decode-fail, and a test pins that.
- Docstring, inline comment, and geotiff.rst describe the lower-bound contract consistently.

### Checklist
- [x] Algorithm matches reference (N/A, no numerical change)
- [x] All implemented backends consistent (VRT reads are CPU/dask only)
- [x] NaN handling correct (N/A, no pixel-value change)
- [x] Edge cases covered by tests (corrupt source, all-missing, present source)
- [x] Dask chunk boundaries handled correctly (window test reused from existing guard)
- [x] No premature materialization (build-time scan is path-existence only, no decode)
- [x] Benchmark not needed
- [x] Docstrings accurate

brendancol (commented):
## Follow-up review (after a69477a3)

- Suggestion (duplicate traversal, `_backends/vrt.py`): fixed. `decode_capable` is now computed inside the existing static-hole sweep. The loop applies the window-intersection test to every in-window source once and branches on `os.path.exists`: present sources set the flag, missing ones append to `chunked_holes`. No second pass.
- Nit (local import of `warnings` / `GeoTIFFFallbackWarning`): dismissed. It matches the eager path's local-import of the same symbols at `_vrt.py:1621`. A module-level import here would diverge from that established pattern, so leaving it keeps the two paths consistent.

### State now
- Single source-list walk; behaviour unchanged.
- Full `xrspatial/geotiff/tests/vrt/` suite green (502 passed); the four new #2989 tests pass.
- flake8 clean on the changed file.

No new findings.

PR #2992, opened 2026-06-06
geotiff: reject malformed SCALE/OFFSET under mask_and_scale

### LINKED ISSUE #2987: geotiff: malformed SCALE/OFFSET silently ignored under mask_and_scale
**Describe the bug**

`_extract_scale_offset` in `xrspatial/geotiff/_attrs.py` reads the SCALE and OFFSET items from a GeoTIFF's GDAL_METADATA when a caller passes `mask_and_scale=True`. Its inner `_num` helper does:

```python
try:
    return float(gdal_metadata[k])
except (TypeError, ValueError):
    return default
```

So when a SCALE or OFFSET key is present but its value won't parse as a float (say `SCALE="abc"`), the function quietly returns the default 1.0 / 0.0.

The caller passed `mask_and_scale=True` to ask the reader to honor the scale/offset metadata. A malformed value gets treated as if no scale/offset existed at all, so the raw unscaled pixels flow downstream and the file looks clean. That's the same silent-coercion trap the project already closed for integer nodata sentinels with `InvalidIntegerNodataError`.

**Expected behavior**

An absent key should stay silent: defaulting to 1.0 / 0.0 is correct, because the source genuinely carries no scale/offset. But a key that is present and unparseable under `mask_and_scale=True` should fail closed with a typed error from the existing `GeoTIFFAmbiguousMetadataError` family. The message should name the offending key and value.

**Additional context**

Both call sites (`_finalize_eager_read` in `_attrs.py` and the dask backend in `_backends/dask.py`) gate this on `mask_and_scale=True`. The fix needs to tell "key absent" apart from "key present but malformed" inside `_extract_scale_offset`.

brendancol:
Closes #2987

`_extract_scale_offset` returned the 1.0 / 0.0 identity default whenever a `SCALE` or `OFFSET` value failed to parse as a float, so a malformed item read the same as a missing one. Under `mask_and_scale=True` the caller asked the reader to honour the metadata, and the silent fallback let the raw unscaled pixels through as if the file were clean.

Changes:

- `_extract_scale_offset` now raises the new `MalformedScaleOffsetError` (a `GeoTIFFAmbiguousMetadataError`/`ValueError` subclass) when a SCALE or OFFSET key is present but unparseable. An absent key still returns the identity default, so a source with no scale/offset reads as before.
- The error names the offending key and value.
- Added the error to the geotiff package exports and a note to the reference docs.

Both `mask_and_scale` call sites (eager `_finalize_eager_read` and the dask backend) reach `_extract_scale_offset` during graph assembly, so the rejection fires before compute on both. Reads without `mask_and_scale` never touch the metadata and are unaffected.

Test plan:

- [x] `test_mask_and_scale_malformed_scale_raises` / `_offset_raises` (eager)
- [x] `test_mask_and_scale_malformed_scale_dask_raises` (dask)
- [x] `test_malformed_scale_ignored_without_mask_and_scale` (no opt-in, no rejection)
- [x] Existing `test_rioxarray_compat_2961.py` suite still green (29 passed)

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -201,6 +201,13 @@ this section is the brief.
   nodata sentinels raises ``MixedBandMetadataError`` by default. Pass
   ``band_nodata='first'`` to opt back into the legacy flatten-to-band-0
   behaviour; see ``xrspatial/geotiff/tests/vrt/test_metadata.py``.
+* Malformed scale/offset. Under ``mask_and_scale=True`` the read applies
+  the source's ``SCALE`` / ``OFFSET`` from ``GDAL_METADATA``. A key that
+  is present but does not parse as a float (for example ``SCALE="abc"``)
+  raises ``MalformedScaleOffsetError`` (a ``GeoTIFFAmbiguousMetadataError``
+  subclass) rather than falling back to the 1.0 / 0.0 default and reading
+  the raw pixels as clean. An absent key keeps the identity default; see
+  ``xrspatial/geotiff/tests/read/test_rioxarray_compat_2961.py``.
 
 The lifecycle is locked end-to-end by
 ``xrspatial/geotiff/tests/read/test_nodata.py``.
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -67,7 +67,8 @@ from ._coords import \
 from ._crs import _resolve_crs_to_wkt, _wkt_to_epsg  # noqa: F401
 from ._errors import (ConflictingCRSError, ConflictingNodataError, DuplicateIFDTagError,
                       GeoTIFFAmbiguousMetadataError, InconsistentGeoKeysError, InvalidCRSCodeError,
-                      InvalidIntegerNodataError, MixedBandMetadataError,
+                      InvalidIntegerNodataError, MalformedScaleOffsetError,
+                      MixedBandMetadataError,
                       NonRepresentableEPSGCRSError, NonUniformCoordsError,
                       RemoteStableSourcesOnlyError, RotatedTransformError, UnknownCRSModelTypeError,
                       UnparseableCRSError, UnsupportedGeoTIFFFeatureError,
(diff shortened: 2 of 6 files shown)


brendancol (commented):
## PR Review: geotiff: reject malformed SCALE/OFFSET under mask_and_scale

None.

None.

- `xrspatial/geotiff/_errors.py`: the hierarchy tree in the module docstring (`Exception -> GeoTIFFAmbiguousMetadataError -> ...`) already drops several existing subclasses, so leaving `MalformedScaleOffsetError` out of it stays consistent with what's there. The tree is already stale, so adding the new class is optional cleanup, not a fix.

### What looks good
- The change is confined to the one helper the finding named, which keeps the conflict surface small against the sibling PR editing the same function.
- The absent-key vs present-but-malformed split is handled right. The loop only inspects a key when it's actually in the dict and returns the identity default only when no key is present. A malformed dataset-level SCALE raises instead of falling through to the per-band key, which matches the old code: on a parse failure it returned the default, never the per-band value.
- The new error subclasses `GeoTIFFAmbiguousMetadataError`/`ValueError`, so existing `except ValueError` callers keep catching it. Same shape as the `InvalidIntegerNodataError` precedent for this class of silent-coercion bug.
- The message names the offending key and value.
- Both call sites hit `_extract_scale_offset` at graph-assembly time, so the dask path fails closed before compute rather than lazily. The dask test confirms it.
- Tests cover eager SCALE, eager OFFSET, dask, and the no-opt-in case, with uniquely named temp files carrying the issue number.

### Checklist
- Algorithm matches reference: n/a (validation guard, no numerical algorithm)
- Backends consistent: eager and dask both raise; GPU `mask_and_scale` is already rejected upstream
- NaN handling: untouched by this PR
- Edge cases tested: absent key, present-malformed SCALE, present-malformed OFFSET, no opt-in
- Dask chunk boundaries: n/a, the raise is at graph build
- No premature materialization: confirmed, no new array ops
- Benchmark: not needed (error path)
- README feature matrix: n/a (no new function, no backend change)
- Docstrings present and accurate: yes, on the new error class and the helper

brendancol (commented):

The one nit from the first pass is resolved: `MalformedScaleOffsetError` is now listed in the `_errors.py` hierarchy docstring tree. I left the other pre-existing subclasses missing from that tree alone, since backfilling them is unrelated to this PR.

No new findings. No blockers, no open suggestions. The change stays confined to the malformed-value handling in `_extract_scale_offset` plus the new error class and its exports.

PR #2993, opened 2026-06-06
geotiff: reject mixed per-band SCALE/OFFSET under mask_and_scale (#2988)

### LINKED ISSUE #2988: mask_and_scale=True silently corrupts multi-band rasters with differing per-band SCALE/OFFSET
**Describe the bug**

`open_geotiff(..., mask_and_scale=True)` can silently corrupt a multi-band raster when the source carries per-band SCALE/OFFSET that differ between bands.

`_extract_scale_offset` in `xrspatial/geotiff/_attrs.py` returns a single `(scale, offset)` pair: it prefers the dataset-level `SCALE`/`OFFSET` and falls back to band 0's per-band values, then applies that one pair to the whole array. When bands have different scale/offset, every band other than band 0 is scaled with the wrong factor. The attrs still look valid (`scale_factor` / `add_offset` are stamped), so downstream spatial functions consume numerically wrong values with no signal that anything is off.

This is documented as a limitation in the `mask_and_scale` docstring today, but for a read pipeline that feeds spatial ops, silent wrong numbers are a data-corruption bug, not just a caveat.

**Expected behavior**

When `mask_and_scale=True` and the source carries differing per-band SCALE/OFFSET:

- If a single band is selected with `band=`, apply that band's scale/offset.
- If no band is selected (the full multi-band array is returned), raise `MixedBandMetadataError` instead of silently applying band 0's values. This mirrors the existing per-band nodata handling, which already rejects conflicting per-band sentinels by default and tells the caller to disambiguate.

A source with a single dataset-level scale/offset, or uniform per-band values, keeps working as before.

**Additional context**

Both the eager (numpy/cupy) read path and the dask read path call `_extract_scale_offset`, so the fix has to be applied consistently across both.

brendancol:
Closes #2988

`mask_and_scale=True` read a multi-band raster with one `(scale, offset)` pair, falling back to band 0's per-band values when there was no dataset-level value. Bands with a different scale/offset came back numerically wrong while the attrs still looked valid, so downstream spatial functions had no signal that the data was corrupt.

This makes `_extract_scale_offset` band-aware:

- A dataset-level `SCALE`/`OFFSET` still applies to the whole array.
- Per-band values that agree across bands still apply to the whole array.
- Per-band values that disagree now raise `MixedBandMetadataError` when no band is selected, mirroring the existing per-band nodata handling. Selecting a single band with `band=` reads it with that band's own scale/offset.

Both the eager (numpy) and dask read paths thread the selected band into the helper, so they reject and scale identically.

Backend coverage: CPU eager (numpy) and dask. `mask_and_scale` already raises `ValueError` with `gpu=True` and on `.vrt` sources, so cupy / VRT paths are unaffected.

Test plan:
- [x] Mixed per-band SCALE raises on eager and dask
- [x] Mixed per-band OFFSET raises
- [x] `band=` selects that band's scale/offset (eager and dask agree)
- [x] Uniform per-band and dataset-level values still apply unchanged
- [x] Existing mask_and_scale tests still pass


brendancol (commented):
## PR Review: geotiff: reject mixed per-band SCALE/OFFSET under mask_and_scale (#2988)

### Blockers
None.

### Suggestions
- `xrspatial/geotiff/_attrs.py:1541` -- the docstring says the band-selected case falls back to "the dataset-level value, then the default", but the code returns early on any dataset-level key (line 1555), so the per-band branch is only reached when there is no dataset-level value. When a band is selected and that band has no per-band entry, the code returns `default`, never a dataset-level value. Trim the docstring to "that band's per-band value, else the default".

### Nits
- `xrspatial/geotiff/_attrs.py:1553` -- `_resolve(name, dataset_key, default)` is always called with `name == dataset_key` ('SCALE'/'SCALE', 'OFFSET'/'OFFSET'). The `dataset_key` parameter is redundant; drop it and use `name` for both lookups.

### What looks good
- The fix lands in the one shared helper both read paths call, and both eager and dask now pass `band`, so they reject and scale identically. The dask call resolves scale/offset at graph-build time, so the rejection raises eagerly rather than at compute, matching the eager path.
- Distinctness is compared on coerced floats, not raw strings, so "2.0" and "2.00" don't trip a false mixed-metadata error.
- Tests cover mixed SCALE, mixed OFFSET, band selection on both backends, uniform per-band, and the existing dataset-level cases still pass.
- Reuses the existing `MixedBandMetadataError`, consistent with the per-band nodata rejection already in the module.

### Checklist
- [x] NaN handling is correct for the masking path (unchanged)
- [x] Both implemented backends (eager, dask) produce consistent results
- [x] Docstring present; one overstated line noted above
- [ ] README feature matrix -- not applicable, no new public function or backend change

brendancol (commented):
## Follow-up review (after ee02548f)

- Suggestion (docstring overstated the band-selected fallback): fixed. The docstring now reads "that band's per-band value (or the default when that band carries none)", which matches the code.
- Nit (redundant `dataset_key` parameter on `_resolve`): fixed. `_resolve(name, default)` now uses `name` for both the dataset-level and per-band lookups.

No new findings. The 12 mask_and_scale tests still pass. Nothing else outstanding.

PR #2994, opened 2026-06-06
geotiff: align eager nodata masking with the lazy/dask path for int sources (#2990)

### LINKED ISSUE #2990: geotiff: eager vs lazy nodata masking diverge on dtype and masked_nodata for int sources with no sentinel pixels
**Describe the bug**

`open_geotiff(..., masked=True)` on an integer-dtype source that declares a nodata sentinel returns a different result from the eager path than from the lazy (dask) path when the read window has no pixels matching the sentinel.

Same file, same `masked=True`, only `chunks=` differs:

- Eager: `dtype=uint16`, `masked_nodata=False`, `nodata_pixels_present=False`
- Lazy: `dtype=float64`, `masked_nodata=True`, `nodata_pixels_present` absent

The lazy path declares `float64` and stamps `masked_nodata=True` from the graph dtype before any chunk is decoded (`xrspatial/geotiff/_backends/dask.py` near lines 424 and 521). The eager path only promotes the integer array to float64 when at least one sentinel pixel actually matches (`_apply_eager_nodata_mask` in `xrspatial/geotiff/_attrs.py` near line 1483); with no match it keeps the integer dtype and reports `masked_nodata=False`.

This matters because `masked_nodata` is read as semantic state, not just dtype decoration. The writer's `_should_restore_nan_sentinel` keys the NaN-to-sentinel rewrite off it, and the GPU writer reads it too. The two backends disagree on whether masking happened for the same input.

**Expected behavior**

The eager and lazy paths should report the same `masked_nodata` and the same dtype for the same input. rioxarray's `open_rasterio(..., masked=True)` always promotes an integer source to float regardless of whether a sentinel pixel is present, so the lazy path's behavior is the correct reference. The eager path should match: when `masked=True` on a maskable integer source with a declared sentinel, promote to float64 unconditionally and report `masked_nodata=True`, even when no sentinel pixel matches.

`nodata_pixels_present` is a separate signal and is allowed to stay absent on the lazy path, since a strict per-chunk reduction would force an eager compute. This issue does not propose forcing eager compute on the lazy path. It only proposes making the `masked_nodata` flag and the output dtype consistent.

**Reproduction**

Write a uint16 GeoTIFF declaring nodata=9999 whose pixels are all in the range 1..50, then open it with `masked=True` both eager and with `chunks=` set, and compare `dtype` and `masked_nodata`.

**Additional context**

Found during a code review of the geotiff backends. The eager promotion gate on a matching sentinel pixel is the divergent behavior.

brendancol:
Closes #2990

## Problem

`open_geotiff(..., masked=True)` on an integer source with a declared, in-range nodata sentinel returned different results from the eager path than from the lazy (dask) path when no pixel matched the sentinel:

- Eager: `uint16`, `masked_nodata=False`
- Lazy: `float64`, `masked_nodata=True`

The dask path declares `float64` up front from the in-range sentinel gate and stamps `masked_nodata=True` before any chunk decodes. The eager helper only promoted to `float64` after finding a matching pixel, so the same file came back two different ways depending on `chunks=`. This matters because `masked_nodata` is read as semantic state: the writer's NaN-to-sentinel rewrite and the GPU writer both key off it, not just off the dtype.

## Fix

`_apply_eager_nodata_mask` now promotes a maskable integer source to `float64` whenever masking is on, independent of whether any pixel matches. This matches the dask path and rioxarray's `open_rasterio(..., masked=True)`, both of which promote unconditionally. The promotion gate is unchanged for out-of-range, non-finite, and fractional sentinels, which still keep the source integer dtype and report `masked_nodata=False` on both paths.

`nodata_pixels_present` keeps its existing meaning (did any pixel match) and stays absent on the lazy path per the documented dask contract. No eager compute is forced on the dask path.

## Backend coverage

- numpy (eager): behavior changed (the fix)
- cupy (eager GPU): routes through the same helper via `_finalize_eager_read`, so it picks up the fix
- dask+numpy / dask+cupy (lazy): already correct, used as the reference

## Test plan

- [x] New `TestEagerLazyMaskParity2990` asserts eager and dask agree on dtype and `masked_nodata` for the no-hit, hit, and out-of-range cases
- [x] Updated the eager no-hit test to the corrected contract (float64 + `masked_nodata=True` + `nodata_pixels_present=False`)
- [x] Full geotiff suite passes (6177 passed, 81 skipped, 1 xfailed)
- [x] Updated docstrings (`_apply_eager_nodata_mask`, `open_geotiff`, module contract) and `attrs_contract.rst`

## Scope note

The VRT eager/chunked masking helpers (`_vrt.py`) carry their own per-band sentinel handling and a separate documented divergence; they are left unchanged here. Worth a follow-up look but out of scope for this fix.

### DIFF
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -85,13 +85,18 @@ write.
        ``_resolve_nodata_attr``.
    * - ``masked_nodata``
      - bool
-     - Paired with ``nodata``. ``True`` when the reader actually
-       replaced sentinel pixels with NaN (so the buffer is NaN-aware);
-       ``False`` when the array still carries the literal sentinel
-       values, including the case where the array is float dtype
-       because the caller passed ``masked=False`` (the default) together
-       with ``dtype=float...``. Only set when ``nodata`` is set; absence
-       means no declared sentinel. See issue #2092.
+     - Paired with ``nodata``. ``True`` when the reader ran the
+       sentinel-to-NaN step so the buffer is NaN-aware; ``False`` when
+       the array still carries the literal sentinel values, including the
+       case where the array is float dtype because the caller passed
+       ``masked=False`` (the default) together with ``dtype=float...``.
+       The flag tracks whether masking ran, not whether any sentinel
+       pixel matched: a masked read of a maskable integer source promotes
+       to float and sets ``True`` even when zero pixels match, so the
+       eager and dask paths agree for the same input (issue #2990). Use
+       ``nodata_pixels_present`` for the did-any-pixel-match question.
+       Only set when ``nodata`` is set; absence means no declared
+       sentinel. See issue #2092.
    * - ``nodata_pixels_present``
      - bool
      - Paired with ``nodata``. ``True`` iff the read window contained
--- a/xrspatial/geotiff/__init__.py
+++ b/xrspatial/geotiff/__init__.py
@@ -688,9 +688,10 @@ def open_geotiff(source: str | BinaryIO, *,
         (``masked=False``); note that earlier xrspatial releases masked
         by default (``mask_nodata=True``), so a bare ``open_geotiff(path)``
         no longer promotes the sentinel to NaN. Pass ``masked=True`` and
-        ``dtype=<integer>`` together and the read raises ``ValueError``
-        once the sentinel matches a pixel, because the float64 promotion
-        then makes the integer cast lossy.
+        ``dtype=<integer>`` together on a source with a maskable sentinel
+        and the read raises ``ValueError``, because the unconditional
+        float64 promotion (issue #2990) makes the integer cast lossy
+        whether or not a sentinel pixel is present.
     mask_nodata : bool
         [deprecated] Deprecated alias of ``masked``; emits a
         ``DeprecationWarning``. Passing both ``masked`` and ``mask_nodata``
@@ -841,9 +842,12 @@ def open_geotiff(source: str | BinaryIO, *,
     passing ``dtype=<integer>`` as well is not enough to keep an integer
     dtype: the sentinel-to-NaN promotion runs first and the subsequent
     integer cast then raises ``ValueError`` (float-to-int is lossy in a
-    way users rarely intend). When the file has no in-range sentinel
-    match, the promotion is skipped and ``dtype=<integer>`` works either
-    way.
+    way users rarely intend). The promotion runs whenever the sentinel is
+    maskable (finite, integer, in-range), whether or not any pixel matches
+    it, so the eager and dask paths return the same float64 dtype for the
+    same input (issue #2990). A sentinel that cannot match (out-of-range,
+    non-finite, or fractional) leaves the source dtype, so
+    ``dtype=<integer>`` works in that case.
 
     Examples
     --------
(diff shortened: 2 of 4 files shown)


brendancol (commented):
## PR Review: align eager nodata masking with the lazy/dask path for int sources (#2990)

None.

None blocking. One thing worth confirming rather than changing: in `_apply_eager_nodata_mask` (`xrspatial/geotiff/_attrs.py:1515-1517`) the mask is now computed after the float64 cast (`mask = arr == np.float64(nodata_int)`), whereas before it ran on the integer buffer. For the in-range integer dtypes that TIFF actually uses (uint8/16, int8/16/32) the cast round-trips exactly through float64, so the equality is unchanged. A sentinel above 2^53 on an int64/uint64 source could in theory lose precision, but the original code had the same exact-cast assumption via `arr.dtype.type(nodata_int)`, and 64-bit integer TIFF rasters with such sentinels are not a path this reader exercises. No action needed unless 64-bit integer sources become a supported tier.

None.

### What looks good

- The fix is the minimal change: drop the matching-pixel gate on the float promotion, keep the finite/integer/in-range gate that protects against unmatchable sentinels. The out-of-range and non-finite branches are untouched, so those still keep the integer dtype on both paths.
- The direction is correct. rioxarray's `open_rasterio(masked=True)` promotes an integer source to float unconditionally, and the dask path already did the same, so aligning the eager path to the lazy path (rather than the reverse) matches the ecosystem and avoids forcing an eager compute on dask.
- `nodata_pixels_present` keeps its separate meaning and is still computed eagerly, so the did-any-pixel-match signal is preserved while `masked_nodata` becomes the consistent did-masking-run flag.
- The GPU eager path routes through the same helper via `_finalize_eager_read`, so cupy picks up the fix without a separate kernel.
- Tests cover the three relevant cases (no-hit, hit, out-of-range) and assert eager-vs-dask agreement directly. The stale eager no-hit test was updated to the corrected contract rather than deleted.
- Docstrings and `attrs_contract.rst` were updated so the contract text no longer claims promotion depends on a matching pixel.

### Checklist

- [x] Algorithm matches reference (rioxarray masked=True semantics; dask path)
- [x] All implemented backends produce consistent results (eager numpy/GPU now match dask)
- [x] NaN handling is correct (sentinel pixels still rewritten to NaN when present)
- [x] Edge cases are covered by tests (no-hit, hit, out-of-range)
- [x] Dask chunk boundaries handled correctly (dask path unchanged, used as reference)
- [x] No premature materialization (lazy path untouched; no eager compute added)
- [x] Benchmark not needed (no new function, masking cost unchanged in the common hit case)
- [x] Docstrings present and accurate (updated for the new contract)

### Scope note

The PR explicitly leaves the VRT masking helpers (`_vrt.py`) alone. They carry per-band sentinel handling and their own documented divergence, and the VRT lazy graph dtype comes from `_effective_dtype_for_bands` rather than the in-range sentinel gate, so the VRT eager/lazy story is self-contained and different from the main reader. Folding it into this PR would expand scope into a separate subsystem. A follow-up issue for VRT parity would be reasonable.

brendancol (commented):
## Follow-up review (#2990)

No code changes were needed after the first review. The first pass found no blockers, suggestions, or nits requiring edits.

Disposition of the one confirm-don't-change item (float-cast equality precision in `_apply_eager_nodata_mask`): dismissed. The mask now runs after the float64 cast, but for the integer dtypes TIFF actually uses (uint8/16, int8/16/32) the cast round-trips exactly, so the equality is unchanged. The prior code carried the same exact-cast assumption via `arr.dtype.type(nodata_int)`, and 64-bit integer sentinels above 2^53 are not a path this reader exercises. No precision regression.

The full geotiff suite passes (6177 passed, 81 skipped, 1 xfailed), including the new cross-backend parity tests and the updated eager no-hit test.

PR #2999, opened 2026-06-06
geotiff: fail closed on malformed GDAL_METADATA XML under mask_and_scale

### LINKED ISSUE #2998: mask_and_scale=True silently ignores malformed GDAL_METADATA XML
**Describe the bug**

When `open_geotiff(..., mask_and_scale=True)` reads a file whose GDAL_METADATA XML tag is not well-formed XML, the reader drops the metadata silently and returns raw, unscaled pixels.

`_parse_gdal_metadata()` in `xrspatial/geotiff/_geotags.py` catches `ET.ParseError` and returns `{}`. Downstream, `_extract_scale_offset()` in `xrspatial/geotiff/_attrs.py` treats a missing scale/offset as identity scaling (1.0 / 0.0). A file that declared a scale/offset inside a corrupt XML payload therefore reads as if no scaling were declared, instead of failing.

PRs #2988 and #2992 already reject malformed numeric SCALE/OFFSET values under `mask_and_scale=True` via `MalformedScaleOffsetError`. Malformed XML itself slips through that guard.

**Expected behavior**

Under `mask_and_scale=True`, a malformed GDAL_METADATA XML payload should fail closed with a clear `MalformedScaleOffsetError`, like the existing malformed-numeric case. Reads that don't request `mask_and_scale` should keep working unchanged, since they never read the scale/offset metadata.

For this module, returning wrong pixels silently is worse than a hard error.

**Additional context**

The DOCTYPE / billion-laughs security drop should stay silent; that payload is refused on purpose. Existing tests cover malformed numeric scale/offset values but not malformed XML itself (`xrspatial/geotiff/tests/read/test_rioxarray_compat_2961.py`).

brendancol:
Closes #2998

## What

Under `mask_and_scale=True`, a GeoTIFF whose `GDAL_METADATA` tag holds non-well-formed XML now fails closed with `MalformedScaleOffsetError` instead of reading raw, unscaled pixels. `_parse_gdal_metadata()` caught `ET.ParseError` and returned `{}`, which `_extract_scale_offset()` then read as identity scaling. A scale/offset hidden inside a corrupt XML payload was lost silently. PRs #2988 / #2992 closed the same risk for malformed numeric SCALE/OFFSET; this covers the XML container itself.

- `_parse_gdal_metadata_strict()` returns `(dict, malformed)`. `_parse_gdal_metadata()` keeps its dict-only contract by discarding the flag, so the existing callers and the DOCTYPE/billion-laughs security drop are unchanged (that payload is still refused silently, on purpose).
- `GeoInfo.gdal_metadata_malformed` carries the flag to the consumer and is inherited by overview IFDs alongside the XML.
- `_extract_scale_offset()` raises `MalformedScaleOffsetError` when the flag is set.

## Backend coverage

The rejection lands on both `mask_and_scale` consumer sites: the eager path (`_finalize_eager_read`, used by numpy / cupy / GPU) and the dask path. numpy, cupy, dask+numpy, and dask+cupy all route through one of the two.

## Test plan

- [x] Eager read of malformed XML raises `MalformedScaleOffsetError`
- [x] Dask read (`chunks=2`) of malformed XML raises `MalformedScaleOffsetError`
- [x] Read without `mask_and_scale` is unchanged (no rejection)
- [x] Existing DOCTYPE security test still returns `{}` (silent drop)
- [x] Full geotiff suite green (6206 passed, 81 skipped, 1 xfailed)

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -1580,6 +1580,10 @@ def _extract_scale_offset(gdal_metadata, band=None, *, malformed=False):
     as if no scaling were declared.
     """
     scale, offset = 1.0, 0.0
+    # Check ``malformed`` before the empty-dict short-circuit below: an
+    # unparseable payload yields ``gdal_metadata == {}``, so a guard that
+    # returned the identity default on an empty dict first would silence
+    # the rejection. Keep this check at the top.
     if malformed:
         raise MalformedScaleOffsetError(
             "GDAL_METADATA XML is malformed and could not be parsed. "
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -1549,7 +1549,7 @@ def _apply_eager_nodata_mask(arr, *, mask_sentinel, mask_nodata):
     return arr, nodata_pixels_present
 
 
-def _extract_scale_offset(gdal_metadata, band=None):
+def _extract_scale_offset(gdal_metadata, band=None, *, malformed=False):
     """Pull SCALE / OFFSET from parsed GDAL_METADATA for ``mask_and_scale``.
 
     Returns ``(scale, offset)`` floats, defaulting to ``(1.0, 0.0)`` when the
@@ -1571,8 +1571,22 @@ def _extract_scale_offset(gdal_metadata, band=None):
     Raises :class:`MalformedScaleOffsetError` when a ``SCALE`` or ``OFFSET``
     item is present but does not parse as a float. An absent key keeps the
     1.0 / 0.0 identity default.
+
+    ``malformed=True`` signals that the source carried a GDAL_METADATA XML
+    payload that did not parse (see :func:`_parse_gdal_metadata_strict`).
+    Because the unparseable payload could have declared a scale / offset
+    that is now lost, ``mask_and_scale`` fails closed with a
+    :class:`MalformedScaleOffsetError` rather than reading the raw pixels
+    as if no scaling were declared.
     """
     scale, offset = 1.0, 0.0
+    if malformed:
+        raise MalformedScaleOffsetError(
+            "GDAL_METADATA XML is malformed and could not be parsed. "
+            "mask_and_scale=True cannot honour the scale / offset it may "
+            "declare, so the read is refused rather than returning raw, "
+            "unscaled pixels."
+        )
     if not gdal_metadata:
         return scale, offset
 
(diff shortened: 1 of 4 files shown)


brendancol (commented):
## PR Review: fail closed on malformed GDAL_METADATA XML under mask_and_scale

### Blockers
None.

### Suggestions
None blocking. Splitting `_parse_gdal_metadata_strict` (returns the malformed flag) from the thin `_parse_gdal_metadata` wrapper keeps the existing callers and the DOCTYPE security test unchanged. Right call.

### Nits
- `_attrs.py`: `_extract_scale_offset` raises on `malformed=True` before the `if not gdal_metadata` short-circuit. The order is correct, since a malformed payload yields an empty dict and the malformed check has to run first. A one-line comment noting the order matters would protect against a later refactor that moves the empty-dict guard up and quietly silences the rejection.

### What looks good
- The fix targets the actual gap: `ParseError` is now separated from the DOCTYPE `ValueError` drop, so the billion-laughs guard stays silent while genuinely corrupt XML fails closed under `mask_and_scale`.
- The malformed flag rides on `GeoInfo` and is inherited by overview IFDs alongside the XML it describes (`_geotags.py` overview-inheritance block), so a COG overview that inherits a corrupt base payload also fails closed.
- Both `mask_and_scale` consumer sites are covered (eager `_finalize_eager_read` and the dask path). VRT does not consume `_extract_scale_offset`, so there is no third site to wire.
- Reads without `mask_and_scale` are untouched: the flag is only read inside the `mask_and_scale` branch.
- Tests cover eager + dask rejection and the no-op-without-`mask_and_scale` case. The malformed XML is injected through the raw `gdal_metadata_xml` writer kwarg because the writer escapes normal values.

### Checklist
- Algorithm matches reference: n/a (bug fix)
- Implemented backends consistent: yes (eager + dask both raise)
- NaN handling: unaffected
- Edge cases covered by tests: yes (malformed XML, with and without `mask_and_scale`, eager and dask)
- Dask chunk boundaries: unaffected (rejection at graph-build, before chunking)
- Benchmark: not needed
- README feature matrix: not needed (no new API)

brendancol (commented):
## Follow-up review (after 097d32c7)

The one nit from the previous pass is addressed: `_extract_scale_offset` now carries a comment explaining why the `malformed` check must stay ahead of the empty-dict short-circuit.

No remaining findings. Comment-only change; the malformed/mask_and_scale tests still pass (19 passed).

PR #3006, opened 2026-06-07
Re-export open_geotiff and to_geotiff from the top-level xrspatial namespace

### LINKED ISSUE #3005: Re-export open_geotiff and to_geotiff from the top-level xrspatial namespace
## Reason or Problem
`open_geotiff` and `to_geotiff` are the two public entry points of the `geotiff` subpackage, but today you can only import them from the subpackage:

```python
```

Every other public function in xarray-spatial imports directly from the top level (`from xrspatial import slope`). The GeoTIFF entry points are the odd ones out, which trips people up.

## Proposal
Re-export both functions from the top-level `xrspatial` package so either spelling works:

```python
from xrspatial import open_geotiff, to_geotiff          # new, shorter
from xrspatial.geotiff import open_geotiff, to_geotiff   # still works
```

**Design:** Add two import lines to `xrspatial/__init__.py` next to the existing top-level re-exports. The functions themselves don't change.

**Value:** Same import style everywhere, and shorter imports for the GeoTIFF entry points.

## Drawbacks
None worth flagging. The subpackage import keeps working, so nothing breaks.

## Alternatives
Leave it as is and document the subpackage path, which keeps the inconsistency.

brendancol:
Closes #3005

`open_geotiff` and `to_geotiff` were the only public functions you had to import from the subpackage. Now they re-export from the top level like everything else:

- Add both to `xrspatial/__init__.py` next to the other top-level re-exports.
- `from xrspatial.geotiff import open_geotiff, to_geotiff` still works (same objects).
- Note the alias in the GeoTIFF reference docs.

No functional change to the readers/writers, so backend coverage (numpy / cupy / dask+numpy / dask+cupy) is untouched.

## Test plan
- [x] `test_top_level_reexports_are_the_subpackage_functions` pins both spellings to the same objects.
- [x] Existing `test_api_consolidation.py` suite still passes (14 passed).

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -216,7 +216,9 @@ Reading
 =======
 ``open_geotiff`` is the single read entry point. The backend follows the
 parameters: ``gpu=True`` returns a CuPy-backed array, ``chunks=N`` returns a
-lazy dask array, and a ``.vrt`` source reads a mosaic.
+lazy dask array, and a ``.vrt`` source reads a mosaic. It is re-exported at
+the top level, so ``from xrspatial import open_geotiff`` and
+``from xrspatial.geotiff import open_geotiff`` both work.
 
 .. autosummary::
     :toctree: _autosummary
@@ -228,7 +230,8 @@ Writing
 ``to_geotiff`` is the single write entry point (``gpu=True`` or CuPy data
 selects the GPU path; a ``.vrt`` output path writes tiles plus an index).
 Writing to a ``.vrt`` path is how you produce a VRT mosaic; the underlying
-index emitter is internal.
+index emitter is internal. Like ``open_geotiff``, it is re-exported at the
+top level, so ``from xrspatial import to_geotiff`` also works.
 
 .. autosummary::
     :toctree: _autosummary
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Re-export open_geotiff and to_geotiff from the top-level xrspatial namespace

None.

None.

- The two new imports sit between the `zonal` and `reproject` blocks in `xrspatial/__init__.py:144-145`. The file isn't strictly module-sorted anyway (hydro is scattered too), so this is fine as is. A `geotiff` block grouped with the other I/O imports would read a touch better, but not enough to warrant a re-push.

### What looks good
- Both spellings resolve to the same objects; `xrspatial.open_geotiff is xrspatial.geotiff.open_geotiff` holds.
- There's no top-level `__all__` to keep in sync, so the re-export matches how every other public function is exposed.
- The contract test pins identity, not just importability, so it catches a future accidental shadow.
- No circular import: the geotiff subpackage doesn't reach back into the top-level package.
- Backward compatible. The subpackage import keeps working.
- Docs mention the alias in both the Reading and Writing sections.

### Checklist
- Algorithm matches reference/paper: n/a (no algorithm change)
- All implemented backends produce consistent results: n/a (no backend change)
- Edge cases covered by tests: the identity test covers the contract
- No premature materialization or unnecessary copies: n/a
- README feature matrix updated: not applicable (no new function)
- Docstrings present and accurate: docs updated

PR #3008, opened 2026-06-07
Size streaming GeoTIFF write budget from source chunks, not output tiles

### LINKED ISSUE #3007: Streaming GeoTIFF writer budgets memory from output tiles, not source chunks — OOMs on slope/overlap pipelines
**Describe the bug**

`to_geotiff` streams a dask-backed DataArray one tile-row at a time and bounds peak memory with `streaming_buffer_bytes` (default 256 MB). The column-segmentation budget is sized from the *output* tile geometry:

```python
# _writer.py:1030
bytes_per_tile_col = th * tw * bytes_per_sample * samples
```

That matches reality for a plain windowed read. It does not match a `map_overlap` source. When the array is `open_geotiff(chunks=...).xrs.slope()` (same for aspect, curvature, hillshade), slicing one tile-row `dask_data[r0:r1, seg].compute()` materializes every *source* chunk-row the band intersects plus a one-chunk halo — not the tile-row's output footprint.

So when the source chunks are taller than the tile (e.g. `chunks=512` read, `tile_size=256`) and the raster is wide (LANDFIRE CONUS), each compute pulls several full-width source chunk-rows. The 256 MB budget never sees that, the column segmentation never tightens enough, and the compute OOMs.

Measured on a 2048x8192 float32 raster read at `chunks=512`, slope, one 256-row output strip:

- output strip the budget assumes: 8 MB
- source reads actually materialized: 32 MB (2 chunk-rows x 16 col-chunks; the depth-1 halo pulls the neighbor chunk-row)

The overshoot scales with `source_chunk_height / tile_height` and with raster width, so a tall chunk on a CONUS-width raster is multiple GB per compute.

**Expected behavior**

Peak materialized memory per compute stays within `streaming_buffer_bytes` regardless of how the source was chunked. The column budget should come from the source chunk geometry — the source chunk-rows a tile-row band touches, plus a one-chunk halo — instead of the output tile height.

**Repro**

```python
from xrspatial import slope

base = da.random.random((2048, 8192), chunks=(512, 512)).astype('float32')
agg = xr.DataArray(base, dims=['y', 'x'])
to_geotiff(slope(agg), 'slope.tif')  # each tile-row pulls ~4x the budgeted bytes
```

**Additional context**

- Default `cog=False` streaming path, `_writer.py:1024-1099`.
- Workaround today: read with short chunks matching the tile, e.g. `open_geotiff(path, chunks=256)`.
- Strip layout (`tiled=False`) does no horizontal segmentation at all; bounding its peak needs a separate change and is out of scope here.

brendancol:
Closes #3007

The streaming `to_geotiff` writer bounds peak memory with `streaming_buffer_bytes`, but sized its horizontal-segment budget from the *output* tile geometry (`tile_h * tile_w`). That undercounts a `map_overlap` source: slicing one tile-row out of `open_geotiff(chunks=...).xrs.slope()` materializes every source chunk-row the band touches plus a one-chunk halo. A source chunked taller than the tile (e.g. `chunks=512`, `tile_size=256`) on a wide raster pulled several full-width source chunk-rows per compute and OOMed, despite the buffer.

- Add `_max_streaming_row_span`: the worst-case source rows a tile-band materializes (touched source chunk-rows plus a one-chunk halo each side; `map_overlap` depth never exceeds one chunk, so that is an upper bound).
- Size `bytes_per_tile_col` from that span instead of `th`. Falls back to the tile height for numpy-from-dask or unknown-chunk arrays. The column halo of a 2D overlap adds a bounded couple of source chunk-columns, so `streaming_buffer_bytes` stays a soft cap.

On the repro (2048x8192 float32, `chunks=512`, slope, 8 MB cap) peak source bytes per compute drop from 32 MB to 12.6 MB.

Backend coverage: writer path is backend-agnostic (numpy / cupy / dask+numpy / dask+cupy all reach the same streaming writer); the fix affects the dask streaming branch.

Test plan:
- [x] `test_overlap_source_respects_buffer_3007` -- instruments source-chunk materialization and asserts peak per compute stays within 2x the buffer (fails pre-fix at 4x)
- [x] `test_overlap_source_roundtrip_small_buffer_3007` -- lazy slope under a tight cap matches eager slope
- [x] `test_max_streaming_row_span_3007` -- helper unit test
- [x] existing `TestStreamingBufferBudget` + dask integration suite still green (146 passed, 1 skipped)

### DIFF
--- a/xrspatial/geotiff/_writer.py
+++ b/xrspatial/geotiff/_writer.py
@@ -1068,8 +1068,11 @@ def _write_streaming(dask_data, path: str, *,
                 # the budget from that row span so a tall-chunk wide raster
                 # cannot blow past streaming_buffer_bytes (#3007). The column
                 # halo adds a bounded couple of source chunk-columns on top;
-                # streaming_buffer_bytes stays a soft cap. Falls back to the
-                # tile height for numpy-from-dask or unknown-chunk arrays.
+                # streaming_buffer_bytes stays a soft cap. A non-overlap read
+                # carries no halo, so this is intentionally conservative for
+                # it -- it may segment one step early, which only costs an
+                # extra compute call near the cap. Falls back to the tile
+                # height for numpy-from-dask or unknown-chunk arrays.
                 materialized_h = th
                 row_chunks = getattr(dask_data, 'chunks', None)
                 if row_chunks:
--- a/xrspatial/geotiff/_writer.py
+++ b/xrspatial/geotiff/_writer.py
@@ -673,6 +673,37 @@ write = _write
 # re-exported above for backwards compatibility.
 
 
+def _max_streaming_row_span(row_chunks, tile_h, height):
+    """Worst-case source rows a tiled streaming compute materialises.
+
+    The streaming writer computes one ``tile_h``-tall band per dask
+    ``.compute()``. For a plain windowed read that band materialises only
+    the source chunk-rows it overlaps. For a ``map_overlap`` source
+    (slope / aspect / curvature / hillshade) dask also pulls the
+    neighbouring chunk-row on each side to satisfy the halo, and a
+    windowed read materialises every touched chunk-row in full. ``depth``
+    can never exceed one chunk, so a one-chunk halo each side is an upper
+    bound. Return the largest span across all bands so the column budget
+    is sized from the source geometry, not the output tile height (#3007).
+    """
+    import bisect
+    offsets = [0]
+    for h in row_chunks:
+        offsets.append(offsets[-1] + int(h))
+    n = len(row_chunks)
+    worst = tile_h
+    for r0 in range(0, height, tile_h):
+        r1 = min(r0 + tile_h, height)
+        first = bisect.bisect_right(offsets, r0) - 1
+        last = bisect.bisect_right(offsets, r1 - 1) - 1
+        lo = max(0, first - 1)
+        hi = min(n - 1, last + 1)
+        span = offsets[hi + 1] - offsets[lo]
+        if span > worst:
+            worst = span
+    return worst
+
+
 def _write_streaming(dask_data, path: str, *,
                      geo_transform: 'GeoTransform | None' = None,
                      crs_epsg: int | None = None,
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: source-aware streaming write budget

Reviewed the full diff and the surrounding streaming-writer context. The change is narrow and the logic holds up.

### Blockers
None.

### Suggestions
- The one-chunk halo is added unconditionally, so the budget treats every source as if it had a `map_overlap` halo. A genuinely non-overlap wide read (plain windowed read, no slope/aspect) will therefore segment about one step earlier than it strictly needs to. The writer can't cheaply tell overlap from non-overlap, and over-segmenting only costs an extra compute call near the buffer ceiling -- the safe direction. Worth a one-line note in the code comment that the non-overlap case is intentionally conservative, but not worth detection machinery.

### Nits
- `_writer.py`: `import bisect` sits inside the helper rather than at module top. That matches the file's existing style (`os` / `tempfile` are imported inside `_write_streaming`), so it's consistent -- leaving as-is is fine.
- Strip layout (`tiled=False`) still has no horizontal segmentation, so an overlap source in strip mode keeps its old unbounded peak. The issue scopes that out and the docstring now says so; flagging only so it isn't forgotten.

### What looks good
- The budget is sized from the real driver. `_max_streaming_row_span` accounts for the source chunk-rows a tile-band touches plus the one-chunk halo, and the one-chunk bound is correct because `map_overlap` requires `depth <= chunk size`.
- The fix only resizes column segments; the slices stay tile-aligned and cover the full raster, so output bytes are unchanged. The 146-test streaming + dask-integration suite staying green backs that up.
- `test_overlap_source_respects_buffer_3007` instruments actual source-chunk materialization rather than asserting on an internal counter, and it fails on the pre-fix writer (32 MB vs 8 MB cap). That's a real regression guard, not a smoke test.
- Helper has direct unit coverage for the three interesting shapes (band inside a tall chunk, middle band with halo on both sides, single full-height chunk).
- Unknown-chunk / numpy-from-dask inputs fall back to the tile height via the `getattr` + `try/except`, so the new path can't raise on a chunkless array.

### Checklist
- [x] Output unchanged (slices still tile-aligned and complete)
- [x] All streaming backends reach the same path; fix is backend-agnostic
- [x] NaN handling untouched (no change to pixel transforms)
- [x] Edge cases covered (tall chunk, ragged halo, single chunk, fallback)
- [x] Dask chunk boundaries handled (halo bound is provably tight)
- [x] No premature materialization added (helper is pure arithmetic)
- [x] Benchmark not needed (memory-budget fix, no hot loop touched)
- [x] README matrix N/A (no new function)
- [x] Docstrings updated in both `_write_streaming` and public `to_geotiff`

brendancol (commented):
## Follow-up review (after aeec97fc)

- **Fixed** -- the suggestion to flag the non-overlap conservatism in-code. Added a comment at the budget block (`aeec97fc`) explaining that a non-overlap read carries no halo, so the one-chunk halo is intentionally conservative there and may segment one step early.
- **Dismissed (style match)** -- `import bisect` inside the helper. The file already imports `os` / `tempfile` inside `_write_streaming`, so the local import is consistent with the surrounding code.
- **Dismissed (out of scope, documented)** -- strip-layout (`tiled=False`) overlap peak. The issue scopes strip mode out and the docstring now states it does no horizontal segmentation.

No new findings. The one-line follow-up is comment-only; the streaming budget + helper tests stay green (7 passed).

PR #3010, opened 2026-06-07
Fix contours() return_type validation to fail fast on invalid values

### LINKED ISSUE #2788: contours() validates return_type too late, wasting compute on invalid values
**Describe the bug**

`contours()` in `xrspatial/contour.py` checks the `return_type` argument only at the end of the function, after all the contour work has already run. The branch that raises `ValueError` for an unknown `return_type` sits near the bottom of the function (around line 676), past level generation, backend dispatch, and the per-segment coordinate transform.

For a Dask-backed input this means an invalid `return_type` triggers a full `dask.compute()` before the function raises. For a large NumPy raster it can hit the memory guard in `_contours_numpy` first. Either way the call does a lot of work before failing on what is usually a typo.

**Expected behavior**

An invalid `return_type` should fail fast with a clear `ValueError` before any level generation, backend dispatch, or compute happens. Passing `return_type="geojson"` (or any other unsupported value) should raise right away, with no Dask execution and no memory allocation.

**Steps to reproduce**

- Build a Dask-backed DataArray.
- Call `contours(agg, return_type="bogus")`.
- The chunks get computed before the `ValueError` is raised.

**Additional context**

Move the `return_type` membership check to the top of `contours()`, next to the existing raster validation. The late branch can then drop its `else: raise` since the value is already known good. Add a test asserting an invalid `return_type` raises without triggering a compute on a Dask input.

Melissari1997:
Closes https://github.com/xarray-contrib/xarray-spatial/issues/2788

## Summary

- `contours()` now validates `return_type` at the top of the function, before any level computation, backend dispatch, or compute happens
- Removed the unreachable `else` block that could never execute (the validation at the top already guarantees valid values)
- Added a test that confirms invalid `return_type` raises without triggering `dask.compute()` on a Dask-backed input

## Backend coverage

No changes to backend logic - this is purely an input validation fix. All four backends (numpy, cupy, dask+numpy, dask+cupy) benefit from the early validation.

## Test plan

- `test_invalid_return_type_no_dask_compute` - verifies no dask.compute is called before the ValueError raise
- All existing contour tests pass (53 passed, 23 skipped for optional deps)


Melissari1997 (commented):
## PR Review: Fix contours() return_type validation (#2788)

### Blockers
None

### Suggestions
1. **Test spy function recursion risk** - The original `spy_compute` called `da.compute` after patching, which would infinite-recurse if compute were ever invoked. Fixed in follow-up commit by capturing `_original_compute = dask.compute` before patching.

### Nits
1. **contour.py:745-747** - The simplification from `if/elif/else` to `if/return` is clean and correct. The `return_type` validation at line 664-668 guarantees only "numpy" or "geopandas" reach this point, so the direct return for geopandas is safe.

2. **test_contour.py** - The test `test_invalid_return_type_no_dask_compute` correctly verifies the fast-fail behavior. It only covers `dask+numpy` backend, but since validation happens before backend dispatch, this is sufficient.

3. **Commit message** - The first commit message mentions "Move return_type validation check to top of contours()" but the validation was already at the top. The commit actually removes the dead else block and adds the test. Consider rewording to be more accurate.

### Summary
- 2 files changed, 2 files modified
- 48 tests passed, 22 skipped (optional deps)
- Change is focused and correct
- All findings are nits or already-fixed suggestions

brendancol (approved):

PR #3021, opened 2026-06-07
Fix contours() border collar on integer dask input (#3020)

### LINKED ISSUE #3020: contours(): integer dask input gets a spurious border collar vs numpy
**Describe the bug**

`contours()` returns different geometry for dask-backed arrays than for eager numpy arrays when the input raster is an integer dtype. The dask output traces a rectangular border (a "collar") around the whole data extent that the numpy output doesn't have.

The dask backend gives each chunk a one-cell halo before marching squares with `da.overlap.overlap(data, depth={0: 1, 1: 1}, boundary=np.nan)`. With floating-point input the halo cells are NaN, and the kernel skips any quad touching them, so the chunked result matches numpy. Integer arrays can't hold NaN, so dask casts the fill to the integer minimum instead (int32 gives -2147483648). `np.isfinite` returns True for that value, so the phantom halo cells get treated as real data. Any edge cell sitting above a contour level then crosses against the int-min halo corner, and those crossings stitch together into a frame around the raster.

The numpy backend casts to float64 before tracing and never adds a halo, so it's fine. The dask path runs the overlap while the array is still integer, so the int-min halo is already baked in by the time the per-chunk call casts to float.

**Expected behavior**

Same geometry from numpy and dask regardless of input dtype. No collar for integer input.

**Reproduction**

```python
from xrspatial import contours

z = np.tile(np.arange(20), (20, 1)).astype(np.int32)
co = {"y": np.arange(20), "x": np.arange(20)}
a_np = xr.DataArray(z, dims=["y", "x"], coords=co)
a_dk = xr.DataArray(da.from_array(z, chunks=(7, 7)), dims=["y", "x"], coords=co)

g_np = contours(a_np, levels=[5.0, 10.0, 15.0], return_type="geopandas")
g_dk = contours(a_dk, levels=[5.0, 10.0, 15.0], return_type="geopandas")

print(g_np.length.sum())  # 57.0
print(g_dk.length.sum())  # 168.0 -- the extra length is the collar
```

Casting to float first (`.astype(np.float64)`) makes both backends agree.

This shows up with real integer rasters like land cover classifications (e.g. LANDFIRE EVC).

**Fix**

Cast non-float dask input to float64 before the overlap call in `_contours_dask` and `_contours_dask_cupy`, so the halo is real NaN and matches what the numpy backend already does.

**Additional context**

Affects dask+numpy and dask+cupy. numpy and cupy backends are correct.

brendancol:
Closes #3020

## Problem

`contours()` returned different geometry for dask-backed integer rasters than for eager numpy. The dask output drew a rectangular border collar around the whole data extent.

The dask backends call `da.overlap.overlap(data, ..., boundary=np.nan)` to halo each chunk. NaN only fits in float dtypes; on an integer array dask fills the halo with the integer minimum, which `np.isfinite` accepts. Edge cells above a contour level then cross against those phantom corners, and the crossings stitch into a frame. The numpy backend casts to float64 before tracing, so it was never affected.

## Fix

Cast non-float dask input to float64 before the overlap, in a shared `_overlap_for_contours` helper used by both `_contours_dask` and `_contours_dask_cupy`. The halo is then real NaN and matches the numpy backend.

## Backends

- numpy: already correct, unchanged
- cupy: already correct, unchanged (no overlap)
- dask+numpy: fixed
- dask+cupy: fixed

## Test plan

- [x] New `TestIntegerDtypeCollar`: integer dask and dask+cupy results match numpy segment-for-segment; geopandas total length and bounds match numpy (no collar)
- [x] Confirmed the new tests fail before the fix and pass after
- [x] Full `test_contour.py` suite passes (79 tests)

### DIFF
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -400,6 +400,21 @@ def _contours_cupy(data, levels):
     return _contours_numpy(cpu_data, levels)
 
 
+def _overlap_for_contours(data):
+    """Add a NaN 1-cell halo to a dask array for marching squares.
+
+    ``boundary=np.nan`` only produces real NaN on float dtypes.  On an
+    integer array NaN can't be stored, so dask fills the halo with the
+    integer minimum, which ``np.isfinite`` accepts.  Those phantom cells
+    then emit spurious crossings that stitch into a border collar (see
+    issue #3020).  Cast non-float input to float64 first so the halo is
+    real NaN, matching the numpy backend which casts in ``_contours_numpy``.
+    """
+    if data.dtype.kind != 'f':
+        data = data.astype(np.float64)
+    return da.overlap.overlap(data, depth={0: 1, 1: 1}, boundary=np.nan)
+
+
 def _contours_dask(data, levels):
     """Dask backend: process each chunk with 1-cell overlap, then merge.
 
@@ -410,7 +425,7 @@ def _contours_dask(data, levels):
     if da is None:
         raise ImportError("Dask is required for chunked contour extraction")
 
-    padded = da.overlap.overlap(data, depth={0: 1, 1: 1}, boundary=np.nan)
+    padded = _overlap_for_contours(data)
     orig_row_chunks = data.chunks[0]
     orig_col_chunks = data.chunks[1]
     padded_blocks = padded.to_delayed()
@@ -445,7 +460,7 @@ def _contours_dask_cupy(data, levels):
     if da is None:
         raise ImportError("Dask is required for chunked contour extraction")
 
-    padded = da.overlap.overlap(data, depth={0: 1, 1: 1}, boundary=np.nan)
+    padded = _overlap_for_contours(data)
     orig_row_chunks = data.chunks[0]
     orig_col_chunks = data.chunks[1]
     padded_blocks = padded.to_delayed()
--- a/xrspatial/tests/test_contour.py
+++ b/xrspatial/tests/test_contour.py
@@ -33,6 +33,23 @@ def _make_peak():
     return data
 
 
+def _segments_by_level(results, decimals=8):
+    """Decompose contour polylines into canonicalized segments per level.
+
+    Each segment is stored with its smaller endpoint first so the result is
+    direction-independent, and segments are sorted for stable comparison
+    across backends.
+    """
+    by_level = defaultdict(list)
+    for level, coords in results:
+        for i in range(len(coords) - 1):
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Fix contours() border collar on integer dask input (#3020)

### Blockers
None.

### Suggestions
- [ ] `xrspatial/tests/test_contour.py`: the tests only exercise int32 via `_int_ramp`. The fix covers every non-float kind (int16, uint8, int64, bool). A `@pytest.mark.parametrize` over a couple of those dtypes would lock in the broader contract cheaply. Not blocking, since the mechanism is the same across integer widths.

### Nits
- [ ] `xrspatial/tests/test_contour.py` `_collect_segments` reuses `TestBackendEquivalence._collect_segments(self, results)` by passing a `TestIntegerDtypeCollar` instance as `self`. It works because the helper doesn't read instance state, but it reads oddly. A module-level helper or a `@staticmethod` would be cleaner.

### What looks good
- Root cause is right: `boundary=np.nan` can't store NaN in an integer halo, so dask fills with int-min and `np.isfinite` lets the phantom cells through. Casting to float before the overlap matches what `_contours_numpy` already does.
- `astype` preserves the dask chunk layout, so reading `data.chunks` after the helper call still lines up with the padded blocks.
- The shared `_overlap_for_contours` helper keeps the cast in one place across both dask backends.
- Regression tests were confirmed to fail before the fix and pass after, on dask+numpy and dask+cupy. Ragged chunks (7,7 on a 20-wide raster) get exercised.

### Checklist
- [x] Matches reference behavior (numpy backend)
- [x] NaN handling is correct
- [x] No premature materialization (lazy astype)
- [x] README feature matrix not needed (no API change)

brendancol (commented):
## PR Review (follow-up): Fix contours() border collar on integer dask input (#3020)

### Blockers
None.

### Suggestions
None remaining. The earlier suggestion (parametrize over integer widths) was addressed: `test_int_dask_equals_numpy` now runs across int16, int32, int64, and uint8.

### Nits
None remaining. The awkward cross-class `_collect_segments` call was replaced with a module-level `_segments_by_level` helper.

### What looks good
- Fix is unchanged and correct: cast non-float dask input to float64 before the overlap so the NaN halo is real, matching the numpy backend.
- Test coverage now spans four integer dtypes plus the no-collar geometry check, on dask+numpy and dask+cupy.
- Full `test_contour.py` suite passes (82 tests).

PR #3025, opened 2026-06-07
Route remote sidecar parse failures through the shared warn/raise policy

### LINKED ISSUE #3022: Remote sidecar parse failures are swallowed instead of warning or surfacing the parse error
**Describe the bug**

`discover_remote_sidecar` in `xrspatial/geotiff/_sidecar.py` swallows download and parse failures of a remote `.ovr` sidecar that was already successfully probed. When a probe succeeds but the subsequent download or parse fails, it silently falls back to a base-only read.

That diverges from the local eager path in `_reader.py` and the fsspec metadata path in `__init__.py`. Both route the same failure through `handle_sidecar_parse_failure`, which warns on a malformed sidecar and, when an explicit `overview_level` needs the sidecar, raises a `ValueError` chained from the real parse error instead of degrading into a generic out-of-range failure from `select_overview_ifd`.

For remote sources this means a malformed sidecar gets dropped with no warning, and an explicit `overview_level` that depends on the sidecar fails with a misleading generic error rather than the actual parse cause.

**Expected behavior**

The remote path should follow the same policy as the local one: warn when a probed sidecar fails to parse and fall back to base-only, and surface the actual parse error when the caller asked for an `overview_level` the base file alone cannot serve. The fix should reuse the existing `handle_sidecar_parse_failure` helper so all three reader paths share one policy.

**Additional context**

`handle_sidecar_parse_failure` already exists and is used by the local eager CPU/GPU paths and the fsspec metadata path. Only the remote/HTTP dask metadata path via `discover_remote_sidecar` skips it. `CloudSizeLimitError` must still propagate unchanged.

brendancol:
Closes #3022

## What changed

- `discover_remote_sidecar` now routes a probed-but-unreadable `.ovr` sidecar through the existing `handle_sidecar_parse_failure` helper instead of swallowing the exception. A malformed sidecar warns and falls back to base-only; an explicit `overview_level` only the sidecar could serve re-raises the parse error (chained), matching the eager local and fsspec readers.
- Threaded `overview_level` into `discover_remote_sidecar` and wired it from the `_cog_http` metadata caller.
- `CloudSizeLimitError` budget breaches still propagate unchanged.

## Backend coverage

This touches the GeoTIFF reader's remote/HTTP and dask metadata path only. No spatial-op kernels, so the numpy/cupy/dask matrix does not apply. The fix aligns the dask+HTTP metadata path with the eager CPU/GPU and fsspec paths that already used the shared policy.

## Test plan

- [x] `test_discover_remote_sidecar_falls_back_when_load_fails` updated to assert the `RuntimeWarning` now fires on fallback.
- [x] New `test_discover_remote_sidecar_surfaces_parse_error_for_sidecar_level` covers an explicit `overview_level=1` against a single-IFD base raising the chained parse error.
- [x] `test_discover_remote_sidecar_propagates_cloud_size_limit` still passes (budget breach re-raises).
- [x] Full `test_sidecar.py` + `test_http_sources.py` suites pass (416 passed).

### DIFF
--- a/xrspatial/geotiff/_cog_http.py
+++ b/xrspatial/geotiff/_cog_http.py
@@ -236,7 +236,8 @@ def _parse_cog_http_meta(
     if source_path is not None:
         from ._sidecar import discover_remote_sidecar
         ifds, sidecar, sidecar_ifd_ids = discover_remote_sidecar(
-            source_path, ifds, max_cloud_bytes=max_cloud_bytes,
+            source_path, ifds, overview_level=overview_level,
+            max_cloud_bytes=max_cloud_bytes,
         )
 
     ifd = select_overview_ifd(ifds, overview_level)
--- a/xrspatial/geotiff/_sidecar.py
+++ b/xrspatial/geotiff/_sidecar.py
@@ -292,6 +292,7 @@ def discover_remote_sidecar(
     source,
     base_ifds: list[IFD],
     *,
+    overview_level: int | None = None,
     max_cloud_bytes: int | None = None,
 ) -> tuple[list[IFD], 'SidecarOverviews | None', set[int]]:
     """Probe for a sibling ``.ovr`` and return a merged IFD list.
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Route remote sidecar parse failures through the shared warn/raise policy

None.

None.

- `xrspatial/geotiff/_cog_http.py:238`: three mock HTTP tests (`test_read_cog_http_uses_coalesced_fetches`, `test_read_cog_http_perf_with_mock_rtt`, `test_dask_http_parses_ifds_once`) now emit a RuntimeWarning because their mock `read_all()` rejects the `max_bytes` kwarg, so `load_sidecar` raises a TypeError that the new policy surfaces as a warning. The tests still pass and the warning is the intended behavior change. If the noise is unwanted those mocks could accept `**kwargs`, but that is out of scope here.

### What looks good
- The fix reuses the existing `handle_sidecar_parse_failure` helper instead of duplicating the warn/raise logic, so the eager CPU/GPU, fsspec metadata, and remote/HTTP dask metadata paths now share one policy.
- `base_ifd_count=len(base_ifds)` matches the gating the local paths use, so a base TIFF with its own internal overviews can still serve `level=1` without forcing a raise when the sidecar is corrupt.
- `CloudSizeLimitError` still re-raises before the new handler, so the byte-budget contract is preserved.
- The release contract (`reader.sidecar_ovr` row) already promised this behavior for the dask metadata path; the code was the outlier, so no doc change is needed.
- Tests cover both new branches: warn-and-fall-back for an implicit level, and the chained parse error for an explicit `overview_level=1` against a single-IFD base.

### Checklist
- Algorithm matches reference/paper: N/A (reader robustness fix, no kernel)
- All implemented backends produce consistent results: N/A (no spatial-op backends touched)
- NaN handling is correct: N/A
- Edge cases are covered by tests: yes
- Dask chunk boundaries handled correctly: N/A (metadata path only)
- README feature matrix updated: N/A (no new public API)
- Docstrings present and accurate: yes, docstring updated to document the new behavior and the ValueError path

PR #3026, opened 2026-06-07
Remove stale external-overview skips on dask/fsspec/GPU golden-corpus backends

### LINKED ISSUE #3024: Stale golden-corpus skips hide external-overview validation on dask/fsspec/GPU backends
## Describe the bug

The golden-corpus backend parity modules list `overview_external_ovr_uint16` in `_OVERVIEW_READER_GAPS`, which skips the overview-level oracle comparison for the dask, fsspec, GPU, and dask+GPU backends:

- `xrspatial/geotiff/tests/golden_corpus/test_dask_numpy.py:81`
- `xrspatial/geotiff/tests/golden_corpus/test_fsspec.py:87`
- `xrspatial/geotiff/tests/golden_corpus/test_gpu.py:87`
- `xrspatial/geotiff/tests/golden_corpus/test_dask_gpu.py:85`

The gate switches off the `candidate_factory` that reads each overview level via `open_geotiff(..., overview_level=lvl)`. The base-IFD parity check still runs, but the overview levels stored in the sibling `.tif.ovr` are never checked on these backends.

The eager numpy module cleared its own `_OVERVIEW_READER_GAPS` entry once the reader gained external `.ovr` sidecar discovery (`_sidecar.py`). That sidecar code is shared across backends, so the four remaining skips look stale and leave overview reads unverified on every backend except eager numpy.

## Expected behavior

Backends whose read path resolves the sidecar pyramid should run the overview-level comparison for `overview_external_ovr_uint16`, the same as eager numpy. A skip should only remain where a backend genuinely cannot serve the level.

## To reproduce

Drop the `overview_external_ovr_uint16` entry from `_OVERVIEW_READER_GAPS` in a backend module and run that backend's golden-corpus parity test for the fixture id.

brendancol:
Closes #3024

## What

The golden-corpus dask, fsspec, GPU, and dask+GPU parity modules listed `overview_external_ovr_uint16` in `_OVERVIEW_READER_GAPS`, which switched off the overview-level oracle comparison and ran only the base-IFD check. The reader has external `.ovr` sidecar discovery (`_sidecar.py`) shared across backends, so these gates were stale.

- Removed the `overview_external_ovr_uint16` entry from `_OVERVIEW_READER_GAPS` in all four modules. The overview-level comparison now runs and passes.
- For fsspec, also pushed the sibling `.tif.ovr` into the `memory://` filesystem so the reader's fsspec sidecar discovery (`_probe_fsspec` -> `load_sidecar`) can resolve the external overview levels. Without it the memory filesystem held only the base `.tif` and `overview_level>=1` failed as out of range. That was a test-harness gap, not a reader limitation.

## Backends

Test-only change. All four backends (dask+numpy, fsspec, GPU/cupy, dask+GPU) run and pass the overview comparison in the dev environment used here, which has a working CUDA device. Eager numpy already had its gap cleared.

## Test plan

- [x] `test_dask_numpy_parity[overview_external_ovr_uint16]` passes
- [x] `test_fsspec_parity[overview_external_ovr_uint16]` passes
- [x] `test_gpu_parity[overview_external_ovr_uint16]` passes
- [x] `test_dask_gpu_parity[overview_external_ovr_uint16]` passes
- [x] Full golden-corpus suite for the four modules: 127 passed, 9 skipped
- [x] `test_taxonomy_ids_are_in_manifest` passes in each module

### DIFF
--- a/xrspatial/geotiff/tests/golden_corpus/test_dask_gpu.py
+++ b/xrspatial/geotiff/tests/golden_corpus/test_dask_gpu.py
@@ -82,11 +82,12 @@ _DASK_GPU_SKIPS: dict[str, str] = {}
 # xrspatial reader. The base-IFD parity check still runs; the
 # overview-level loop driven by ``candidate_factory`` is skipped for
 # these ids. See the eager module for the rationale.
-_OVERVIEW_READER_GAPS: dict[str, str] = {
-    "overview_external_ovr_uint16": (
-        "External .ovr sidecar reader is not implemented in xrspatial."
-    ),
-}
+#
+# Empty now that the dask+GPU reader resolves the sibling ``.tif.ovr``
+# pyramid through the shared ``_sidecar`` discovery the eager numpy
+# backend uses, so the overview-level comparison runs for
+# ``overview_external_ovr_uint16`` on this backend too.
+_OVERVIEW_READER_GAPS: dict[str, str] = {}
 
 _INTENTIONAL_SKIPS: dict[str, str] = {
     "nodata_miniswhite_uint8": (
--- a/xrspatial/geotiff/tests/golden_corpus/test_dask_numpy.py
+++ b/xrspatial/geotiff/tests/golden_corpus/test_dask_numpy.py
@@ -78,11 +78,12 @@ _DASK_SKIPS: dict[str, str] = {}
 # xrspatial reader. The base-IFD parity check still runs; the
 # overview-level loop driven by ``candidate_factory`` is skipped for
 # these ids. See the eager module for the rationale.
-_OVERVIEW_READER_GAPS: dict[str, str] = {
-    "overview_external_ovr_uint16": (
-        "External .ovr sidecar reader is not implemented in xrspatial."
-    ),
-}
+#
+# Empty now that the dask reader resolves the sibling ``.tif.ovr``
+# pyramid through the shared ``_sidecar`` discovery the eager numpy
+# backend uses, so the overview-level comparison runs for
+# ``overview_external_ovr_uint16`` on this backend too.
+_OVERVIEW_READER_GAPS: dict[str, str] = {}
 
 _INTENTIONAL_SKIPS: dict[str, str] = {
     "nodata_miniswhite_uint8": (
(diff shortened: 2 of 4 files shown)


brendancol (commented):
## PR Review: Remove stale external-overview skips on dask/fsspec/GPU golden-corpus backends

None.

None.

- test_fsspec.py: the sidecar is served inside `test_fsspec_parity` via the `memory_fs_clean` fixture handle, while the base `.tif` goes through the module-level `_serve_via_memory` helper. Both write to the same process-global memory filesystem so it works, but a reader scanning the test sees two write paths for related files. Folding the sidecar write into `_serve_via_memory` (or a small sibling helper) keeps the memory-FS setup in one place. Cosmetic; the current form is correct.

### What looks good
- The four skip removals are verified rather than assumed. Each backend's overview-level comparison was run and passes, and the taxonomy guard tests still pass because the entries were removed cleanly, not left dangling.
- The fsspec change is the substantive one and is right. That skip was masking a test-harness gap (only the base `.tif` was in the memory filesystem), not a reader limitation. Serving the sibling `.tif.ovr` lets the real fsspec sidecar discovery path (`_probe_fsspec` -> `load_sidecar`) resolve the external overview levels, so the test exercises production code instead of routing around it.
- The sidecar path construction (`path.name` + `.ovr`) matches `find_sidecar`'s `source + ".ovr"` convention, so the served URL lines up with what the reader probes.
- Comments left in place of the removed entries explain why the gap is gone, matching the eager numpy module's existing note.

### Checklist
- [x] Algorithm matches reference/paper (n/a, test-only)
- [x] All implemented backends produce consistent results (all four verified passing)
- [x] Edge cases covered by tests (this PR adds the previously-skipped overview coverage)
- [x] Dask chunk boundaries handled correctly (n/a, no production change)
- [x] Benchmark exists or is not needed (not needed, test-only)
- [x] Docstrings present and accurate (skip-list comments updated)

brendancol (commented):
## Follow-up review (after edcf7bb4)

The single nit from the first pass is resolved. The fsspec sidecar write now lives in `_serve_via_memory` alongside the base `.tif` write, so all memory-filesystem setup is in one helper and the test body no longer carries a parallel write path. Full `test_fsspec.py` module passes (32 passed, 2 skipped).

No new findings. Nothing outstanding.

PR #3027, opened 2026-06-07
Thread sidecar_origin through the HTTP/fsspec sidecar georef path

### LINKED ISSUE #3023: Remote .tif.ovr sidecars with their own GeoKeys can return wrong georeferencing
## Describe the bug

When a GeoTIFF is read over HTTP/fsspec and it has an external `.tif.ovr` sidecar carrying its own GeoKeys (ModelPixelScale / ModelTiepoint / ModelTransformation / GeoKeyDirectory), the remote reader parses those georef tags against the base file's byte order instead of the sidecar's own bytes. If the selected overview IFD lives in the sidecar and declares its own georef, the transform and CRS can be misread, silently misplacing the data.

The local eager and metadata-only readers already handle this. They build a `sidecar_origin` mapping and pass it to `extract_geo_info_with_overview_inheritance`, so the sidecar's GeoKeys win and are parsed against the sidecar's bytes. The HTTP/fsspec path in `_cog_http.py` does not thread `sidecar_origin` through, so it falls back to base-file bytes.

## To Reproduce

Read a remote GeoTIFF whose `.tif.ovr` sidecar has its own GeoKeys that differ from the base file, requesting the sidecar overview level. The returned transform/CRS comes from parsing against the wrong byte source.

## Expected behavior

The HTTP/fsspec sidecar path should honor sidecar-owned GeoKeys the same way the local path does: the sidecar's georef wins and is parsed against the sidecar's own bytes and byte order.

## Additional context

The existing HTTP/fsspec sidecar tests only cover mixed-endian pixel bytes, not sidecar-owned GeoKeys, so this gap went uncaught. The local "sidecar GeoKeys win" tests have no remote analog.

brendancol:
Closes #3023

## What this does

- Build a `sidecar_origin` mapping (`id(ifd)` -> `(sidecar bytes, sidecar byte order)`) in `_cog_http.py` and pass it into `extract_geo_info_with_overview_inheritance`, so the HTTP/fsspec sidecar path matches the eager local and metadata-only readers in `_reader.py`.
- Add remote test coverage that pins the observable contract: when the selected overview IFD lives in a sidecar carrying its own GeoKeys, the sidecar's transform and CRS win over the base file's. Tests cover the HTTP eager, HTTP chunked, fsspec chunked, and direct `_parse_cog_http_meta` paths.

## Note on the mechanism

While verifying the original finding I traced where `data`/`byte_order` (and therefore `sidecar_origin`) actually get consumed. `extract_geo_info` forwards them only to `_parse_geokeys`, which reads already-materialized values off `ifd.entries`. IFD tag values, including out-of-line georef arrays, are read eagerly in `parse_ifd` using each file's own byte order. The sidecar IFDs are parsed with the sidecar's byte order in `discover_remote_sidecar`/`load_sidecar`, so their GeoKeys are already correct before `extract_geo_info` runs.

In other words, with the current eager-parse IFD model the HTTP path did not actually return wrong georeferencing, because the values were parsed against the sidecar bytes upstream. This change is a defensive consistency alignment with the local path rather than a correctness fix on today's code, and it removes the stale "intentionally not threaded" comment so a future change that makes `data`/`byte_order` load-bearing (e.g. lazy tag reads) stays correct on both surfaces. The added tests assert the user-visible contract regardless of the internal mechanism.

## Backend coverage

GeoTIFF reader change. Exercised on the HTTP eager, HTTP chunked (dask), and fsspec chunked (dask) paths.

## Test plan

- [x] `pytest xrspatial/geotiff/tests/integration/test_sidecar.py` (93 passed)
- [x] `pytest xrspatial/geotiff/tests -k "http or sidecar or cog or geotags or geo_info"` (824 passed, 7 skipped)
- [x] `flake8` clean on the changed files

### DIFF
--- a/xrspatial/geotiff/_cog_http.py
+++ b/xrspatial/geotiff/_cog_http.py
@@ -250,15 +250,19 @@ def _parse_cog_http_meta(
     # typically carry no out-of-line geokeys and inherit from level-0
     # (which sits in the base buffer).
     #
-    # Map the sidecar's IFDs to their own (bytes, byte_order) so a
-    # sidecar that declares its own geokeys is parsed against the
-    # sidecar bytes, not the base file's. Sidecar IFDs without geokeys
-    # still inherit from the base level-0 via the overview path. This
-    # mirrors the ``georef_origin`` mapping the eager local / fsspec
-    # reader builds in ``_reader.py``; without it a big-endian ``.ovr``
-    # with its own georef (or any sidecar whose byte order differs from
-    # the base file) is read with the wrong byte source and returns
-    # corrupt georeferencing.
+    # Map the sidecar's IFDs to their own (bytes, byte_order) and pass
+    # it as ``sidecar_origin`` so this path stays symmetric with the
+    # eager local / fsspec reader, which builds the same mapping in
+    # ``_reader.py``. On today's eager-parse IFD model the mapping does
+    # not change the result: ``extract_geo_info`` only forwards
+    # ``data`` / ``byte_order`` to ``_parse_geokeys``, which reads
+    # values already materialized on ``ifd.entries`` (every tag,
+    # including out-of-line georef arrays, is decoded in ``parse_ifd``
+    # against its own file's byte order, and the sidecar IFDs are
+    # parsed with the sidecar header in ``discover_remote_sidecar``).
+    # We thread it anyway to match the local path and to stay correct
+    # if tag-value reads ever become lazy and start depending on the
+    # buffer / byte order handed to ``extract_geo_info``.
     georef_origin = (
         {id(sc_ifd): (sidecar.data, sidecar.header.byte_order)
          for sc_ifd in sidecar.ifds}
--- a/xrspatial/geotiff/_cog_http.py
+++ b/xrspatial/geotiff/_cog_http.py
@@ -250,15 +250,25 @@ def _parse_cog_http_meta(
     # typically carry no out-of-line geokeys and inherit from level-0
     # (which sits in the base buffer).
     #
-    # The ``sidecar_origin`` kwarg used by the eager local / fsspec
-    # paths is intentionally not threaded here; the HTTP / dask
-    # sidecar-byte-order case is handled separately. When that lands,
-    # this call should pick up the same mapping so an HTTP sidecar with
-    # its own geokeys is parsed against the sidecar bytes too.
+    # Map the sidecar's IFDs to their own (bytes, byte_order) so a
+    # sidecar that declares its own geokeys is parsed against the
+    # sidecar bytes, not the base file's. Sidecar IFDs without geokeys
+    # still inherit from the base level-0 via the overview path. This
+    # mirrors the ``georef_origin`` mapping the eager local / fsspec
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Thread sidecar_origin through the HTTP/fsspec sidecar georef path

- None.

- [ ] `xrspatial/geotiff/_cog_http.py:255-258` -- the new comment claims that "without it a big-endian `.ovr` with its own georef ... is read with the wrong byte source and returns corrupt georeferencing." That overstates the effect on today's code. `extract_geo_info` forwards `data`/`byte_order` only to `_parse_geokeys`, which reads already-materialized values off `ifd.entries`. IFD tag values (including out-of-line georef arrays) are read eagerly in `parse_ifd` against each file's own byte order, and the sidecar IFDs are parsed with the sidecar header in `discover_remote_sidecar`. So `sidecar_origin` does not change the result on the current eager-parse path. The change is a defensive consistency alignment with the local reader, not a live correctness fix. Reword the comment to match (the PR body already states this honestly).

- [ ] The four new tests share an identical HTTP-server setup/teardown block. A small fixture could remove the duplication, but the existing mixed-endian tests in this file follow the same inline pattern, so matching them is defensible.

### What looks good
- The fix mirrors the `georef_origin` construction in `_reader.py:295-299` exactly (id -> (bytes, byte_order)), guarded on `sidecar is not None`, so it is a faithful parallel of the local path.
- Tests cover the HTTP eager, HTTP chunked, fsspec chunked, and direct `_parse_cog_http_meta` paths, asserting both transform and CRS. Temp files use issue-scoped, pid+uuid memory keys so parallel runs do not collide.
- No public API change; README/docs/benchmark updates are not needed.

### Checklist
- [x] Algorithm matches reference -- N/A, no algorithm change
- [x] All implemented backends produce consistent results -- HTTP/fsspec eager and chunked paths tested
- [x] NaN handling is correct -- N/A
- [x] Edge cases are covered by tests -- sidecar-with-geokeys vs base covered
- [x] Dask chunk boundaries handled correctly -- chunked paths tested
- [x] No premature materialization or unnecessary copies -- none introduced
- [x] README feature matrix updated -- not applicable
- [x] Docstrings present and accurate -- comment accuracy flagged above

brendancol (commented):
## PR Review (follow-up): Thread sidecar_origin through the HTTP/fsspec sidecar georef path

Re-reviewed after commit 4fbd09ac.

### Blockers
- None.

### Suggestions
- The earlier suggestion on `_cog_http.py` (the comment overstating that the change fixes corrupt georeferencing) is resolved. The reworded comment now states plainly that the mapping does not change the result on the current eager-parse IFD model and is threaded for symmetry with the local reader and to stay correct if tag reads ever become lazy. Accurate.

### Nits
- The four new tests still share an inline HTTP-server setup/teardown. Left as-is to match the surrounding mixed-endian tests in the same file, which use the same pattern. Not worth a divergent fixture.

### What looks good
- Production change is symmetric with `_reader.py`, comment now matches the actual mechanism, and the tests pin the observable "sidecar GeoKeys win" contract on the HTTP eager, HTTP chunked, fsspec chunked, and direct `_parse_cog_http_meta` paths.

No further changes requested.

PR #3028, opened 2026-06-08
Fix UnboundLocalError on restore_sentinel in plain-ndarray VRT tiled write (#2969)

### LINKED ISSUE #2969: VRT tiled write of a plain ndarray raises UnboundLocalError on restore_sentinel
**Describe the bug**

`to_geotiff` accepts `np.ndarray` (per its signature), and `.vrt` paths route to `_write_vrt_tiled`. In that function, `restore_sentinel` is only assigned inside the `if isinstance(data, xr.DataArray)` branch (`xrspatial/geotiff/_writers/eager.py`). When `data` is a plain ndarray, the `else: raw = data` branch runs and `restore_sentinel` is never set, so the per-tile write at `_write_single_tile(..., restore_sentinel=restore_sentinel)` raises `UnboundLocalError`.

Repro:

```python
import numpy as np
to_geotiff(np.arange(256, dtype="float32").reshape(16, 16),
           "out.vrt", tiled=True, tile_size=16, compression="none")
# UnboundLocalError: cannot access local variable 'restore_sentinel' ...
```

**Expected behavior**

A plain ndarray VRT write should either complete (with `restore_sentinel` defaulting to its kwarg default of `True`, matching the non-VRT path) or raise a clear, intentional error. It should not raise `UnboundLocalError`.

**Fix**

Initialize `restore_sentinel = True` (the kwarg default) before the `if isinstance(data, xr.DataArray)` branch so the non-DataArray path has a defined value. Add a regression test for a plain-ndarray VRT write.

**Context**

Found during review of #2968 (staging-dir cleanup). Pre-existing and unrelated to that PR's staging-dir contract, so split out per the one-fix-per-PR convention.

brendancol:
Closes #2969

## Problem

`to_geotiff` accepts a plain `np.ndarray`, and `.vrt` paths route to `_write_vrt_tiled`. That function only assigned `restore_sentinel` inside the `if isinstance(data, xr.DataArray)` branch. When `data` was a plain ndarray, the `else: raw = data` branch ran and `restore_sentinel` was never set, so the per-tile write raised `UnboundLocalError`.

## Fix

- Initialize `restore_sentinel = True` (the kwarg default) before the `if isinstance(data, xr.DataArray)` branch, so the non-DataArray path has a defined value. This matches the non-VRT write path.
- Add a regression test for a plain-ndarray tiled VRT write that round-trips the data.

## Backend coverage

The change is in the eager writer path. It covers numpy and cupy ndarray inputs to the VRT tiled writer; dask and DataArray inputs were already correct.

## Test plan

- [x] Reproduced the `UnboundLocalError` before the fix
- [x] `test_plain_ndarray_tiled_write_round_trips` passes after the fix
- [x] Existing `test_vrt_atomic.py` suite still passes (6 passed)

### DIFF
--- a/xrspatial/geotiff/_writers/eager.py
+++ b/xrspatial/geotiff/_writers/eager.py
@@ -1240,6 +1240,10 @@ def _write_vrt_tiled(data, vrt_path, *, crs=None, nodata=None,
         res_unit = None
         gdal_meta_xml = None
         extra_tags_list = None
+        # Default to the kwarg default so a plain-ndarray write (which
+        # skips the DataArray branch below) still has a defined value
+        # when it reaches the per-tile ``restore_sentinel=`` pass-through.
+        restore_sentinel = True
 
         if isinstance(data, xr.DataArray):
             raw = data.data
--- a/xrspatial/geotiff/tests/write/test_vrt_atomic.py
+++ b/xrspatial/geotiff/tests/write/test_vrt_atomic.py
@@ -183,3 +183,21 @@ def test_dask_failed_tiled_write_cleans_up(tmp_path, monkeypatch):
     to_geotiff(da, vrt_path, tiled=True, compression='none')
     assert os.path.isfile(vrt_path)
     assert os.path.isdir(tiles_dir)
+
+
+def test_plain_ndarray_tiled_write_round_trips(tmp_path):
+    """A plain ndarray VRT write must not raise UnboundLocalError on
+    ``restore_sentinel`` (issue #2969). The non-DataArray path skips the
+    branch that assigns ``restore_sentinel``, so it has to default to the
+    kwarg default before the per-tile write reads it."""
+    arr = np.arange(256, dtype='float32').reshape(16, 16)
+    vrt_path = str(tmp_path / 'ndarray_2969.vrt')
+
+    to_geotiff(arr, vrt_path, tiled=True, tile_size=16, compression='none')
+
+    assert os.path.isfile(vrt_path)
+    assert os.path.isdir(_tiles_dir_for(vrt_path))
+
+    result = open_geotiff(vrt_path)
+    np.testing.assert_array_almost_equal(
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Fix UnboundLocalError on restore_sentinel in plain-ndarray VRT tiled write (#2969)

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The fix mirrors the non-VRT writer path exactly. `_write_vrt_tiled` now sets `restore_sentinel = True` before the `if isinstance(data, xr.DataArray)` branch (eager.py:1240-1243), matching the same pattern already in the non-VRT writer at eager.py:807-811, down to the rationale comment about bare numpy/cupy inputs keeping the NaN-to-sentinel rewrite. So `True` is the right default here, not an arbitrary pick.
- The regression test in test_vrt_atomic.py walks the exact failure path from the issue: a plain ndarray, a tiled VRT write, then a round-trip readback. Without the fix it raises UnboundLocalError; with it, the values come back intact.
- Scope is tight. Four lines of source plus one test, nothing adjacent touched.

### Notes
- I confirmed the bug reproduces on the unpatched code and that the fix clears it. The full write suite passes (1059 passed, 1 skipped).
- This is the eager writer path, so it covers numpy and cupy ndarray inputs. Dask and DataArray inputs already assigned `restore_sentinel` and were never affected. No backend dispatch or README change is needed since no public API changed.

### Checklist
- Algorithm matches reference: n/a (bug fix, no algorithm)
- Backends consistent: fix is in the shared eager path; ndarray inputs now match DataArray inputs
- NaN handling correct: default True preserves the existing NaN-to-sentinel rewrite for bare arrays
- Edge cases covered: plain ndarray tiled write now tested
- Dask boundaries: unaffected; dask path already set the variable
- No premature materialization: no behavior change
- Benchmark: not needed for a bug fix
- README matrix: not applicable, no new function
- Docstrings: unchanged

PR #3029, opened 2026-06-08
Reject to_geotiff(compression=None) with a clear TypeError (#2978)

### LINKED ISSUE #2978: to_geotiff(compression=None) leaks a raw AttributeError instead of a clear error
**Describe the bug**

`to_geotiff(arr, path, compression=None)` raises a raw `AttributeError: 'NoneType' object has no attribute 'lower'` from `_compression_tag` (`xrspatial/geotiff/_encode.py`).

The `compression` parameter is typed and documented as `str` (default `'zstd'`), so `None` was never a supported value. PR #2977 added a type guard at the public boundary that rejects non-string compression, but it exempts `None` to keep the same guard shape as the low-level writer, so `None` still falls through to the downstream `.lower()` call.

**Reproduction**

```python
import numpy as np
to_geotiff(np.ones((4, 4), dtype="float32"), "out.tif", compression=None)
# AttributeError: 'NoneType' object has no attribute 'lower'
```

**Decision needed**

This needs a small design call before fixing:
- Treat `None` as an alias for the `'none'` codec (no compression), or
- Reject `None` at the public boundary with the same `TypeError` as other invalid types.

Either one is a few lines, but the choice affects the documented contract, which is why I'm splitting it out from the type-guard fix in #2977.

**Context**

Surfaced during review of #2977 (issue #2975).

brendancol:
## Summary

`to_geotiff(arr, path, compression=None)` raised a raw `AttributeError: 'NoneType' object has no attribute 'lower'` from `_compression_tag`. The type guards added in #2977 reject non-string compression but exempted `None`, so it fell through to the downstream `.lower()` call.

The documented contract is `compression: str` (default `'zstd'`), with the `'none'` string used to disable compression. `None` was never a supported value, so this rejects it with the same `TypeError` as any other non-string type rather than aliasing it to `'none'`.

- Dropped the `None` exemption from the public `to_geotiff` boundary guard (`_writers/eager.py`) and the low-level writer guard (`_writer.py`).
- Updated the stale `_validate_lowlevel_write_kwargs` docstring that claimed non-string values were not rejected.

Closes #2978.

## Backend coverage

Validation-layer change at the write entry points, so it is backend-agnostic: numpy, cupy, dask+numpy, and dask+cupy all route through the same guards.

## Test plan

- [x] `compression=None` raises `TypeError` at the public `to_geotiff` boundary
- [x] `compression=None` raises `TypeError` at the low-level `write` boundary
- [x] Existing non-string type-guard tests and valid-compression writes still pass
- [x] `pytest xrspatial/geotiff/tests/write/` green

### DIFF
--- a/xrspatial/geotiff/_writer.py
+++ b/xrspatial/geotiff/_writer.py
@@ -178,8 +178,9 @@ def _validate_lowlevel_write_kwargs(*,
     ----------
     compression : str or other
         Codec name. Validated against :data:`_VALID_COMPRESSIONS` and
-        the JPEG-in-TIFF opt-in gate. Non-string values are not
-        rejected here; ``_compression_tag`` raises downstream.
+        the JPEG-in-TIFF opt-in gate. Non-string values (including
+        ``None``) are rejected here with ``TypeError`` so they never
+        reach ``_compression_tag``'s ``.lower()`` (#2978).
     allow_internal_only_jpeg : bool
         If False (the default), ``compression='jpeg'`` is rejected
         because the encoder writes JFIF tiles without the
@@ -212,11 +213,13 @@ def _validate_lowlevel_write_kwargs(*,
             raise ValueError(
                 f"Unknown compression {compression!r} (in {entry_point}). "
                 f"Valid options: {list(_VALID_COMPRESSIONS)}.")
-    elif compression is not None:
-        # Unreachable from ``to_geotiff`` (which only forwards ``str``
-        # or the default), but direct callers can hit this. Without the
+    else:
+        # ``None`` and every other non-string value land here. Without the
         # explicit guard the downstream ``compression.lower()`` would
-        # surface as ``AttributeError`` instead of a typed error.
+        # surface as ``AttributeError`` instead of a typed error. ``None``
+        # is rejected rather than aliased to ``'none'``: the contract is
+        # ``compression: str`` (use the ``'none'`` string to disable
+        # compression) (#2978).
         raise TypeError(
             f"compression must be a str (in {entry_point}); "
             f"got {type(compression).__name__}.")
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Reject to_geotiff(compression=None) with a clear TypeError (#2978)

None.

None.

- `test_basic.py:191` / `:204`: the new `None` case could fold into the existing `test_to_geotiff_non_string_compression_typeerror` parametrize list (`[5, 5.0, ['deflate'], object()]`) instead of two standalone tests. I'd keep them separate, though: `None` was the specific exemption that caused the bug, and the dedicated tests carry the #2978 rationale and cover the low-level `write` path that the parametrized test does not. Not worth changing.

### What looks good
- The fix is minimal and surgical. It drops the `None` exemption from both the public `to_geotiff` guard (eager.py:499) and the low-level `_validate_lowlevel_write_kwargs` guard (_writer.py:215), so `None` now raises the same `TypeError` as any other non-string at both entry points.
- The decision (reject rather than alias to `'none'`) matches the documented `compression: str` contract and is recorded in both code comments and the test docstrings.
- No internal caller passes `compression=None`, so nothing downstream relied on the old fall-through. The remaining `not isinstance(compression, str)` guards in eager.py (max_z_error, jpeg/experimental warnings) are unaffected since `None` is rejected before reaching them.
- The stale docstring that claimed non-string values are not rejected was updated to match the new behavior.
- Both the public and low-level paths have direct regression tests, and the existing non-string and valid-compression tests still pass.

### Checklist
- [x] Behavior matches the documented contract (compression: str)
- [x] Both write entry points (public + low-level) covered
- [x] Error type is consistent with sibling guards (TypeError)
- [x] Edge cases covered by tests (None at both boundaries)
- [x] No premature materialization or unnecessary copies (validation-only change)
- [x] Benchmark not needed (validation path, no perf impact)

PR #3034, opened 2026-06-08
Validate regular grid spacing in viewshed (#2789)

### LINKED ISSUE #2789: viewshed: validate regular grid spacing instead of silently assuming it
## Description

The viewshed CPU, windowed, and dask paths all derive a single constant cell resolution from the coordinate endpoints:

```python
ew_res = (x_range[1] - x_range[0]) / (width - 1) if width > 1 else 1.0
ns_res = (y_range[1] - y_range[0]) / (height - 1) if height > 1 else 1.0
```

This appears at `xrspatial/viewshed.py:1568` (CPU), `:2090` (windowed), and `:2196` (dask). It assumes the x and y coordinates are uniformly spaced. If the coordinates are irregular (non-uniform spacing), the endpoint-derived resolution does not match the actual cell sizes, so distances and vertical angles come out physically wrong with no warning.

## Proposed fix

The algorithm already requires a regular grid. Make that requirement explicit and enforced:

- Validate that the x and y coordinate arrays are monotonic and uniformly spaced (within a small tolerance) before the sweep runs.
- Raise a clear `ValueError` that names which axis is irregular and the expected uniform spacing.
- Run the validation on all three dispatch paths via a shared helper.
- Document the regular-grid requirement in the `viewshed()` docstring.

A full coordinate-aware rewrite of the sweep algorithm is out of scope here; if that is warranted it should be a separate design effort.

brendancol:
Closes #2789

## What

The viewshed sweep derives one constant cell resolution per axis from the coordinate endpoints (`(end - start) / (n - 1)`). That is only correct when the grid is uniformly spaced. On an irregular grid the derived resolution did not match the real cell sizes, so distances and vertical angles came out physically wrong with no warning.

- Add a `_validate_regular_grid` helper that checks the x and y coordinates are monotonic and uniformly spaced within a small (unit-scaled) tolerance.
- Call it once at the top of the public `viewshed()`, so the CPU, windowed (`max_distance`), and dask dispatch paths are all covered.
- Raise a `ValueError` that names the offending axis and the expected uniform spacing.
- Document the regular-grid requirement in the `viewshed()` docstring.

A coordinate-aware rewrite of the sweep is out of scope; this just makes the existing assumption explicit and enforced.

## Backend coverage

Validation runs in the shared `viewshed()` entry point using `raster.indexes`, so it applies to numpy, cupy, dask+numpy, and dask+cupy alike before any backend-specific dispatch.

## Test plan

- [x] Irregular x and irregular y each raise `ValueError` naming the axis (numpy and dask)
- [x] The windowed `max_distance` path is validated too
- [x] A regularly spaced grid still runs and passes `general_output_checks`
- [x] `linspace` float roundoff is accepted (not a false positive)
- [x] Full `test_viewshed.py` suite: 96 passed, 1 skipped (cupy-only)

### DIFF
--- a/xrspatial/tests/test_viewshed.py
+++ b/xrspatial/tests/test_viewshed.py
@@ -1077,3 +1077,68 @@ def test_viewshed_does_not_mutate_cupy_input():
     assert isinstance(raster.data, cp.ndarray)
     assert raster.dtype == np.dtype("int16")
     np.testing.assert_array_equal(raster.data.get(), before)
+
+
+# -------------------------------------------------------------------
+# Regular-grid validation (#2789)
+# -------------------------------------------------------------------
+
+def _build_grid_2789(backend, xs, ys):
+    """5x5 raster with a single peak, given explicit x/y coordinates."""
+    arr = np.zeros((len(ys), len(xs)), dtype="float64")
+    arr[2, 2] = 5.0
+    if backend == "cupy":
+        import cupy as cp
+        arr = cp.asarray(arr)
+    elif backend == "dask":
+        arr = da.from_array(arr, chunks=(3, 3))
+    return xa.DataArray(arr, coords=dict(x=xs, y=ys), dims=["y", "x"])
+
+
+@pytest.mark.parametrize("backend", ["numpy", "dask"])
+def test_viewshed_irregular_x_raises(backend):
+    """Non-uniform x spacing must raise a clear ValueError naming the axis."""
+    ys = np.arange(5, dtype=float)
+    xs = np.array([0.0, 1.0, 2.0, 5.0, 6.0])  # gap between index 2 and 3
+    raster = _build_grid_2789(backend, xs, ys)
+    with pytest.raises(ValueError, match="x coordinates are not uniformly spaced"):
+        viewshed(raster, x=2, y=2, observer_elev=2)
+
+
+@pytest.mark.parametrize("backend", ["numpy", "dask"])
+def test_viewshed_irregular_y_raises(backend):
+    """Non-uniform y spacing must raise a clear ValueError naming the axis."""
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Validate regular grid spacing in viewshed (#2789)

None.

None.

- viewshed.py: `_validate_regular_grid` skips axes with fewer than 3 points. That is the right call (two points are trivially uniform, a single point has no spacing), and the inline comment says so. No change needed; flagging it so a future reader does not mistake the gap for an oversight.

### What looks good
- The check lives in the shared `viewshed()` entry point and reads `raster.indexes`, so it runs once and covers numpy, cupy, dask+numpy, and dask+cupy before any backend dispatch. That is the shared-helper approach the issue asked for.
- The tolerance scales with the spacing magnitude (`1e-6 * abs(expected) + 1e-9`), so it is unit-agnostic. I tried a UTM-like grid with 30 m spacing on ~500000 m coordinates and a descending y-axis from `linspace`; neither tripped a false positive.
- The error message names the offending axis and the expected spacing.
- Tests cover irregular x, irregular y, the windowed `max_distance` path, a passing regular grid, and `linspace` roundoff, parametrized across numpy and dask.

### Checklist
- Algorithm matches reference: n/a (validation only)
- All implemented backends consistent: yes, validated in the shared entry point
- NaN handling: unchanged
- Edge cases covered by tests: yes (both axes irregular, roundoff, windowed path)
- Dask chunk boundaries: unchanged; coords are 1D and small
- No premature materialization: `raster.indexes` reads only the small 1D coordinate arrays
- Benchmark: not needed (validation only)
- README feature matrix: not needed (no new function or backend change)
- Docstrings: regular-grid requirement documented

PR #3035, opened 2026-06-08
Fix broken bounds-check guard in viewshed _calculate_event_row_col() (#2793)

### LINKED ISSUE #2793: Broken bounds-check guard in viewshed _calculate_event_row_col()
**Describe the bug**

`_calculate_event_row_col()` in `xrspatial/viewshed.py` ends with an internal corruption guard:

```python
if abs(x - event_col > 1) or abs(y - event_row > 1):
    raise ValueError("_calculate_event_row_col()")
```

Operator precedence binds `>` before `abs()`, so this evaluates as `abs((x - event_col) > 1)`: `abs()` applied to a boolean comparison result, which is always 0 or 1. The guard is meant to assert that the computed `(x, y)` stays within one cell of `(event_col, event_row)`, but it never actually checks that.

The function is `@ngjit`-compiled, so the dead guard is silent at runtime. Normal output is unaffected because the algorithm already keeps results within one cell, but the guard does not do its job.

**Expected behavior**

The comparison should sit outside `abs()` so the guard raises when the computed neighbour drifts more than one cell from the event cell:

```python
if abs(x - event_col) > 1 or abs(y - event_row) > 1:
    raise ValueError("_calculate_event_row_col()")
```

**Additional context**

Every branch of the function assigns `x` to `event_col`, `event_col + 1`, or `event_col - 1`, and likewise for `y` relative to `event_row`. The corrected guard holds for all valid inputs and will not fire spuriously. Pure internal bug fix, no public API change.

brendancol:
## Summary

Fixes a broken bounds-check guard in `_calculate_event_row_col()` in `xrspatial/viewshed.py`.

- Operator precedence bound `>` before `abs()`, so the guard `abs(x - event_col > 1)` applied `abs()` to a boolean comparison result (always 0 or 1) instead of checking the distance. The guard never did its job.
- Moved the comparison outside `abs()`: `abs(x - event_col) > 1 or abs(y - event_row) > 1`. The corrected guard holds for every valid branch and does not fire spuriously.
- Pure internal bug fix, no public API change.

## Backend coverage

`_calculate_event_row_col()` is a single `@ngjit` CPU helper shared by all backends. No backend-specific code paths are involved.

## Test plan

- [x] New tests check the one-cell invariant across the quadrant and edge branches for both ENTERING and EXITING events
- [x] New test confirms CENTER events raise
- [x] Full `test_viewshed.py` suite passes (89 passed, 1 skipped)

Closes #2793


brendancol (commented):
## PR Review: Fix broken bounds-check guard in viewshed _calculate_event_row_col() (#2793)

None.

None.

- The new tests assert the one-cell invariant directly, which is the right check. They cannot drive the guard itself to raise, since every branch of `_calculate_event_row_col` already keeps the result within one cell. That is fine for an internal defensive guard, but a one-line comment noting the tests document the invariant (rather than triggering the raise) would help the next reader. Not blocking.

### What looks good
- The fix matches the issue exactly: the comparison now sits outside `abs()`, so the guard checks distance instead of `abs()` of a boolean.
- Tests cover the eight neighbours plus the four axis-aligned two-step cells for both ENTERING and EXITING events, exercising the quadrant and edge branches.
- Scope is tight: one operator-precedence fix plus tests, no unrelated changes, CHANGELOG left alone.

### Checklist
- [x] Fix matches the issue's described correction
- [x] Single @ngjit CPU helper, no backend dispatch involved
- [x] No NaN/float concerns (integer row/col indices)
- [x] Edge and quadrant branches covered by tests
- [x] No dask/cupy paths touched
- [x] Benchmark not needed (internal helper, no perf change)
- [x] No public API change, docs not required

brendancol (commented):
## Follow-up review (after commit 549a3687)

The one nit from the previous pass is addressed: the invariant test now carries a comment noting it documents the guard's invariant rather than driving the raise.

No new findings. Blockers: none. Suggestions: none. Nits: none. The `2793` tests still pass (2 passed).

PR #3036, opened 2026-06-08
Name the calling function in the unit-mismatch warning (#2782)

### LINKED ISSUE #2782: Consider unit-mismatch warning for aspect() planar path
Follow-up from #2764 / PR #2777.

`slope()` now warns when the planar path sees degree (lat/lon) coordinates with meter elevation. The review noted `aspect()` has a related footgun.

The case is weaker for aspect than for slope: aspect returns the direction of steepest descent (an atan2 of gradients), so a uniform degree-to-meter scale on both axes mostly cancels out. The planar aspect dispatch does not even read the coordinate resolution. Where it does matter is anisotropy: at non-equatorial latitudes the degree-to-meter ratio differs between the x and y axes, which skews the computed direction.

Worth deciding whether `aspect()` should call `warn_if_unit_mismatch` (xrspatial/utils.py:781) on its planar branch, or whether the anisotropy effect is small enough to leave alone. Filing separately so PR #2777 stays scoped to slope.

brendancol:
Closes #2782.

The decision the issue asked for (whether `aspect()` should call `warn_if_unit_mismatch` on its planar branch) was already made in #2839: aspect's planar path calls the helper. The leftover footgun is that the shared warning text was hard-coded to say "before calling `slope`", so an `aspect()` caller was told to fix a slope call they never made.

- Add an optional `func_name` argument to `warn_if_unit_mismatch` so the message names the operation the user actually called. Default stays `slope`, so slope's wording is unchanged.
- Have `aspect()` pass `func_name='aspect'`.
- Extend the message to mention the anisotropy effect called out in the issue: away from the equator the degree-to-meter ratio differs between the x and y axes, which skews the computed aspect direction.

This is a warning-message change in a private helper. No public API, no new function, and no backend support change, so it applies to all backends equally (the warning fires in the shared planar dispatch before backend selection).

Test plan:
- [x] `warn_if_unit_mismatch` names `slope` by default and `aspect` when passed `func_name='aspect'` (test_utils.py)
- [x] `aspect()` planar path emits a warning that names `aspect` (test_aspect.py)
- [x] Existing slope/aspect/utils warning tests still pass (message stem unchanged)

### DIFF
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -442,7 +442,7 @@ def aspect(agg: xr.DataArray,
     _validate_boundary(boundary)
 
     if method == 'planar':
-        warn_if_unit_mismatch(agg)
+        warn_if_unit_mismatch(agg, func_name='aspect')
         cellsize_x, cellsize_y = get_dataarray_resolution(agg)
         mapper = ArrayTypeFunctionMapping(
             numpy_func=_run_numpy,
--- a/xrspatial/tests/test_aspect.py
+++ b/xrspatial/tests/test_aspect.py
@@ -511,6 +511,14 @@ def test_planar_warns_on_degree_coords_meter_elevation():
         aspect(raster)  # method='planar' is the default
 
 
+def test_planar_warning_names_aspect_not_slope():
+    # The shared warning helper defaults to naming `slope`; aspect must pass
+    # its own name so the advice is not misleading. Regression for #2782.
+    raster = _degree_coord_meter_elevation_raster()
+    with pytest.warns(UserWarning, match="before calling `aspect`"):
+        aspect(raster)
+
+
 def test_planar_no_warn_on_projected_meter_coords():
     raster = _projected_meter_raster()
     with warnings.catch_warnings():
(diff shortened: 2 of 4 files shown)


brendancol (commented):
## PR Review: Name the calling function in the unit-mismatch warning (#2782)

None.

None.

- utils.py:1005-1006: the message still opens with "Slope/aspect operations expect...", and with `func_name='aspect'` it then closes with "before calling `aspect`", so the reader sees both "Slope/aspect" and "aspect" in one warning. It reads fine and is arguably more informative, so leaving it is reasonable. Dropping the "/aspect" and leaning on `func_name` would tighten it, but that is cosmetic.

### What looks good
- Default `func_name='slope'` keeps slope's existing message byte-for-byte, so no current behavior changes.
- The added anisotropy sentence matches the issue rationale: the x/y degree-to-meter ratio diverges away from the equator.
- Tests cover the helper default, the helper override, and the aspect() integration path. The "appears to have coordinates in degrees" stem is unchanged, so pre-existing match-based tests keep passing.
- The warning fires in the shared planar dispatch before backend selection, so all four backends behave identically. No dispatch change needed.

### Checklist
- Algorithm matches reference: n/a (warning-text change)
- Backends consistent: yes, warning is pre-dispatch
- NaN handling: unchanged
- Edge cases covered by tests: yes for the warning path
- Benchmark: not needed
- README feature matrix: not needed (private helper, no API change)
- Docstrings present and accurate: yes, Parameters added for func_name

PR #3039, opened 2026-06-08
Keep stream_link_mfd and stream_order_mfd dask assembly lazy (#2885)

### LINKED ISSUE #2885: Extend lazy MFD dask assembly to stream/hand modules and study the convergence phase

brendancol:
Follow-up to #2865, which made the dask output assembly lazy in `flow_accumulation_mfd`, `flow_length_mfd`, and `watershed_mfd`. This does the same for the two eager paths #2885 lists:

- `_stream_link_mfd_dask`
- the strahler and shreve paths in `stream_order_mfd`

(`hand_mfd` was already lazified separately, so it is left alone.)

Both modules used to call `.compute()` per tile, run the per-tile kernel eagerly, and stitch the numpy results with `da.from_array` / `da.block`, which materialized the whole output raster during the API call. They now build the result with `da.map_blocks` over the `flow_accum` array. The `flow_accum` array is realigned onto the fractions' spatial tile grid (the same rechunk handling `watershed_mfd` uses for pour points), and the matching 3D fractions strip is sliced inside the block function, so the per-tile kernel runs at compute time.

The boundary-convergence sweep that runs before assembly still reads tile data eagerly. That is inherent to the cross-tile convergence algorithm and stays tracked as the second half of #2885.

## Backend coverage

numpy and cupy paths are unchanged. dask+numpy now assembles lazily; dask+cupy goes through the same CPU dask path via the existing `map_blocks` get()/asarray() wrappers, so it inherits the change.

## Test plan

- [x] existing dask-vs-numpy parity tests still pass
- [x] new: flow_accum chunked differently from fractions still matches numpy (exercises the rechunk realignment)
- [x] new: assembly is deferred to compute time (spy on the tile kernel; assert it is invoked during `.compute()`, not at API-call time)

Closes #2885

### DIFF
--- a/xrspatial/hydro/stream_link_mfd.py
+++ b/xrspatial/hydro/stream_link_mfd.py
@@ -924,38 +924,48 @@ def _stream_link_mfd_dask(fractions_da, accum_da, threshold):
     _mask_bdry = mask_bdry
     _threshold = threshold
 
-    # Assemble result via da.block
-    rows = []
-    for iy in range(n_tile_y):
-        row = []
-        for ix in range(n_tile_x):
-            y_start = sum(chunks_y[:iy])
-            y_end = y_start + chunks_y[iy]
-            x_start = sum(chunks_x[:ix])
-            x_end = x_start + chunks_x[ix]
+    # Lazy assembly: each tile is recomputed on demand from the converged
+    # boundary state.  Driver memory holds only the small boundary/mask
+    # snapshots, so peak memory scales with chunk size rather than the full
+    # grid.  ``fractions_da`` is 3D (8, H, W) and cannot align with the 2D
+    # output via map_blocks, so we map over ``accum_da`` (aligned to the
+    # fractions' spatial tile grid) and slice the matching fractions strip
+    # inside the closure.  ``chunk-location`` gives the tile index directly.
+    accum_da = accum_da.rechunk((chunks_y, chunks_x))
+    cum_y = np.zeros(n_tile_y + 1, dtype=np.int64)
+    np.cumsum(chunks_y, out=cum_y[1:])
+    cum_x = np.zeros(n_tile_x + 1, dtype=np.int64)
+    np.cumsum(chunks_x, out=cum_x[1:])
 
-            frac_chunk = np.asarray(
-                fractions_da[:, y_start:y_end, x_start:x_end].compute(),
-                dtype=np.float64)
-            ac_chunk = _to_numpy_f64(
-                accum_da[y_start:y_end, x_start:x_end].compute())
+    def _tile(ac_block, block_info=None):
+        iy, ix = block_info[0]['chunk-location']
+        y_start = int(cum_y[iy])
+        y_end = int(cum_y[iy + 1])
+        x_start = int(cum_x[ix])
+        x_end = int(cum_x[ix + 1])
 
-            sm = np.where(ac_chunk >= _threshold, 1, 0).astype(np.int8)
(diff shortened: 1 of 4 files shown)


brendancol (commented):
## PR Review: Keep stream_link_mfd and stream_order_mfd dask assembly lazy (#2885)

None.

None. The change follows the merged #2865 pattern (and the already-lazy hand_mfd), so the established approach carries over directly.

- The `_tile` closures in both files drop the defensive `if block_info is None` guard that `_hand_mfd_dask` keeps. That is safe here: all three map_blocks calls pass an explicit `meta=`, so dask never does a `block_info=None` meta dry-run (verified the closure is not called during graph build). Worth a one-line comment so a later edit that removes `meta` does not silently reintroduce a None-deref.

### What looks good
- `accum_da.rechunk((chunks_y, chunks_x))` matches the validated spatial shape and reuses the same realignment watershed_mfd uses for pour points, so a flow_accum array chunked differently from the fractions gets correct tile indices instead of wrong ones.
- Output dtype (float64) and tiling match the previous da.block assembly, so the downstream coord/attr wrapping is unchanged.
- Tests cover the rechunk-realignment path (flow_accum 3x3 vs fractions 2x2) and the laziness contract (spy on the tile kernel, assert assembly runs during .compute(), not at API-call time), parametrized across strahler and shreve.

### Checklist
- [x] Algorithm unchanged (assembly-only refactor); per-tile kernels untouched
- [x] dask+numpy and dask+cupy stay consistent with numpy (existing + new parity tests)
- [x] NaN handling unchanged (same stream-mask construction inside the closure)
- [x] Edge cases: chunk mismatch covered; single-cell covered by the existing suite
- [x] Dask chunk boundaries handled; convergence sweep still eager by design, tracked in #2885
- [x] No premature materialization: full-raster da.block / per-tile .compute() removed from the assembly path
- [x] Benchmark: none for hydro MFD; #2865 added none either
- [x] README / docs: no API or backend-support change
- [x] Docstrings: internal functions; comments updated to describe the lazy assembly

brendancol (commented):
## Follow-up review (after 9e66632)

The single nit from the first pass is addressed: both `_tile` closures now carry a one-line comment noting `block_info` is always populated because `meta` is passed to `map_blocks`, so the dropped `block_info is None` guard is intentional.

No new findings. The 31 stream_link_mfd / stream_order_mfd tests pass locally, including the new rechunk-realignment and laziness tests. Nothing else changed between the two pushes.

PR #3040, opened 2026-06-08
Compute the Getis-Ord Gi* statistic in emerging_hotspots()

### LINKED ISSUE #2804: emerging_hotspots() repeats the local-mean z-score bug instead of computing Gi*
Follow-up to #2767 / #2803.

While fixing `hotspots()` to compute the real Getis-Ord Gi* statistic, I found that `emerging_hotspots()` carries its own copy of the same wrong formula. In `xrspatial/emerging_hotspots.py` it convolves each time step with a normalized kernel and z-scores the local mean against the global mean/std (see `_emerging_hotspots_numpy` around line 414, `_emerging_hotspots_cupy` around line 447, and `_convolve_and_zscore_chunk` around line 474). It reuses `_calc_hotspots_numpy` for the banding but never computes the Gi* weight sum, squared weight sum, sample count n, or the variance adjustment term.

The fix mirrors #2803: compute the per-cell Gi* z-score from the kernel weight sum, squared weight sum, valid-cell count, global mean, and population std, on all four backends. The Gi* helpers added in #2803 (`_gistar_zscore`, `_gistar_global_stats`) can be reused. The space-time hot/cold trend classification (Mann-Kendall step) stays as is; only the per-step z-score is wrong.

Kept separate from #2803 to keep one fix per PR.

brendancol:
Closes #2804

`emerging_hotspots()` carried its own copy of the local-mean z-score bug that `hotspots()` had before #2803. It convolved each time step with a normalized kernel and z-scored the local mean against the global mean and std, which drops the neighborhood weight sum, the squared weight sum, the sample count n, and the variance adjustment term.

What changed:

- Replace the per-step local-mean z-score with the real Gi* z-score, reusing the `_gistar_zscore` and `_gistar_global_stats` helpers from #2803. The global mean, population std, and valid-cell count n are taken over the whole space-time cube; the per-step weight sums come from convolving the validity mask with the kernel and squared kernel, so raster edges and interior NaN cells drop out of every neighborhood sum.
- Rework the dask backends to run the per-step convolutions through the dispatching `convolve_2d` (matching the `hotspots()` dask path) instead of a manual `map_overlap`, so the dask edge handling matches numpy and cupy.
- Keep empty-neighborhood cells as NaN so the Mann-Kendall step still detects pixels with no data across the whole series. The trend classification is unchanged.

Backends: numpy, cupy, dask+numpy, dask+cupy all use the same Gi* math through the existing `ArrayTypeFunctionMapping` dispatch.

Test plan:

- [x] Per-step z-score matches a Gi* reference for a binary kernel.
- [x] Same for a weighted (non-binary) kernel, where W_i != W2_i.
- [x] Output no longer matches the old local-mean formula for a weighted kernel.
- [x] Interior NaN excluded from the neighborhood; neighbors stay finite.
- [x] numpy/dask parity (interior matches, edge ring NaN on both).
- [x] `pytest xrspatial/tests/test_emerging_hotspots.py xrspatial/tests/test_focal.py` pass (327 tests).

### DIFF
--- a/xrspatial/emerging_hotspots.py
+++ b/xrspatial/emerging_hotspots.py
@@ -142,6 +142,10 @@ def _gistar_step_zscore(step_data, kernel, sq_kernel, global_mean,
     does. ``global_mean``, ``global_std``, and ``n`` are the global terms
     over the whole space-time cube.
     """
+    # Single-array backends pass a concrete numpy or cupy slice, so pick the
+    # matching module to build the validity mask. The dask helper below works
+    # off da.* instead, since its slices are lazy dask arrays of either chunk
+    # type and convolve_2d dispatches on the chunk type at compute time.
     xp = cupy.get_array_module(step_data) if cupy is not None else np
     valid = (~xp.isnan(step_data)).astype(step_data.dtype)
     filled = xp.where(valid > 0, step_data, step_data.dtype.type(0.0))
--- a/xrspatial/emerging_hotspots.py
+++ b/xrspatial/emerging_hotspots.py
@@ -22,11 +22,14 @@ import numpy as np
 import xarray as xr
 from numba import prange
 
-from xrspatial.convolution import convolve_2d, _convolve_2d_numpy, _convolve_2d_cupy
-from xrspatial.focal import _calc_hotspots_numpy
+from xrspatial.convolution import convolve_2d
+from xrspatial.focal import (
+    _calc_hotspots_numpy,
+    _gistar_global_stats,
+    _gistar_zscore,
+)
 from xrspatial.utils import (
     ArrayTypeFunctionMapping,
-    _boundary_to_dask,
     _validate_boundary,
     ngjit,
     not_implemented_func,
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Compute the Getis-Ord Gi* statistic in emerging_hotspots()

This swaps the per-step local-mean z-score for the real Gi* statistic, following the #2803 fix to hotspots(). I read the full emerging_hotspots.py and the focal.py Gi* helpers it reuses, ran the existing suite plus the new tests, and checked numpy/dask parity across all four boundary modes.

None.

None.

- benchmarks/benchmarks/focal.py has a time_hotspots benchmark but none for emerging_hotspots. Pre-existing gap, not introduced here, and not needed for a bugfix. Worth a follow-up only if the function's runtime starts to matter.
- _gistar_step_zscore (emerging_hotspots.py:130) uses cupy.get_array_module guarded by `cupy is not None`, while _gistar_step_dask uses da.* directly. Same convolution terms, two idioms. Fine as is, but a one-line comment on why the single-array path needs the xp switch and the dask path does not would save the next reader a double-take.

### What looks good
- The core fix is correct. Global mean, population std, and valid-cell count n come from the whole space-time cube, and the per-step weight sums come from convolving the validity mask with the kernel and squared kernel. The per-step output matches a hand-rolled Gi* reference to 1e-5 for both a binary and a weighted kernel.
- Reworking the dask backends to call convolve_2d instead of a manual map_overlap was the right move. It drops the edge-handling discrepancy the old map_overlap path had: numpy and dask now agree to 2.4e-7 on the interior with identical NaN positions for nan, nearest, reflect, and wrap.
- Keeping empty-neighborhood cells as NaN in _gistar_step_finalize is well reasoned and tested. A plain 0 would have broken the all-NaN-pixel detection in the Mann-Kendall step.
- Orphaned imports from the removed map_overlap path (_convolve_2d_numpy, _convolve_2d_cupy, _boundary_to_dask) were cleaned up, and the pre-existing unused not_implemented_func import was left alone, which is the right call for a surgical bugfix.
- New tests cover the binary kernel, the weighted kernel where W != W2, a regression guard that the output no longer matches the old local-mean formula, and interior-NaN exclusion.

### Checklist
- Algorithm matches reference: yes, the Gi* formula from #2803
- All implemented backends consistent: yes, numpy/dask verified, cupy paths reuse the same helpers
- NaN handling correct: yes
- Dask chunk boundaries handled correctly: yes, via convolve_2d dispatch
- No premature materialization: yes, global stats computed once with a single da.compute
- Benchmark exists or not needed: not needed for a bugfix
- Docstrings present and accurate: yes, unchanged and still correct

brendancol (commented):
## Follow-up review (after 86fc634)

Re-reviewed after the follow-up commit. The only change is a 4-line comment in _gistar_step_zscore explaining why the single-array path picks numpy vs cupy via cupy.get_array_module while the dask path uses da.* directly. That clears the one nit worth fixing in-PR.

The other nit (no emerging_hotspots benchmark in benchmarks/benchmarks/focal.py) is deferred: it is a pre-existing gap, the benchmark file is not touched by this PR, and adding one would push past the scope of a bugfix.

No blockers, no suggestions, nothing else outstanding. All 49 tests in test_emerging_hotspots.py pass locally.

PR #3041, opened 2026-06-08
polygonize: fix dask 8-connectivity float divergence at large rtol (#2677)

### LINKED ISSUE #2677: Dask 8-connectivity float polygonize diverges from numpy at large rtol
**Describe the bug**

With `connectivity=8` and a large `rtol`, `polygonize` on a dask-backed float raster can produce a different number of polygons (and different DN values) than the same raster as a single numpy chunk. This is separate from the rtol scan-order orientation fixed in #2666 / #2675.

**Reproduce**

```python

arr = np.array([
    [2.098, 2.43,  2.206, 2.09 ],
    [1.847, 2.292, 1.875, 2.784],
    [2.927, 1.767, 2.583, 2.058],
    [2.136, 2.851, 1.142, 1.174],
])
v_np, _ = polygonize(xr.DataArray(arr), atol=0.0, rtol=0.1, connectivity=8)
v_dk, _ = polygonize(
    xr.DataArray(da.from_array(arr, chunks=(2, 2))),
    atol=0.0, rtol=0.1, connectivity=8)
print(len(v_np), len(v_dk))  # 6 vs 4
```

**What I found**

This is a ring-merge topology problem, not a boundary close-value problem. It fails identically across chunk shapes that do not even split the relevant pixels (for example chunks of `(1, 4)`, `(4, 1)`, and `(2, 2)` all reproduce it), so the cross-chunk boundary decision is not the cause. The likely area is the degree-4 vertex pairing in `_pick_next_edge` / the figure-8 handling in `_merge_polygon_rings` where 8-connected diagonal regions meet.

#2675 fixes the 4-connectivity rtol case completely and the large majority of 8-connectivity cases; this issue tracks the remaining 8-connectivity divergence at large `rtol`.

**Expected behavior**

A chunked 8-connectivity float raster produces the same polygon partition as the unchunked input for the same `atol` / `rtol`.

brendancol:
Closes #2677.

## Problem

With `connectivity=8` and a large `rtol`, `polygonize` on a dask-backed float raster gave a different polygon partition (count and DN values) than the same raster as a single numpy chunk. It reproduced even on chunk shapes that do not split the diverging pixels, which points at the diagonal-pairing decision in the cross-chunk merge rather than the chunk-boundary close test.

numpy's connected-component labelling treats the 8-connected diagonal as conditional: a pixel checks its SW diagonal only when its W neighbour did not match, and its SE diagonal only when its S neighbour did not match. The dask cross-chunk merge honoured every diagonal pair unconditionally, so at a large `rtol` it unioned two regions numpy keeps separate whenever the intervening orthogonal neighbour already matched.

## Change

- Replace the per-pair 8-conn diagonal close test with a pass that replays numpy's CCL predicate over the union of boundary cells, with the same conditional SW/SE diagonal.
- The diagonal reads the real orthogonal neighbour value when that neighbour is itself a boundary cell (the cross-chunk case), and falls back to a recorded in-chunk match flag when the neighbour is interior to the cell's own chunk.
- 4-connectivity and integer rasters keep the existing ring-edge-keyed pairwise test unchanged.

## Backends

This touches the dask+numpy and dask+cupy merge paths; cupy float routes through the same numpy CCL and cross-chunk merge. The numpy and cupy single-chunk paths are unchanged.

## Test plan

- [x] New `test_polygonize_issue_2677.py`: the issue repro matches the unchunked numpy partition for every chunking; a deterministic fuzz over random float rasters across `rtol` and chunkings checks 8-conn parity; 4-conn and integer 8-conn paths stay put.
- [x] Full polygonize suite passes (562 passed, 16 skipped).
- [x] Fuzz of 20000 chunked-vs-unchunked comparisons: zero 8-conn divergences. (The only signature differences are a pre-existing 4-conn ring-simplification artifact at rtol=0.3, unrelated to this issue.)

### DIFF
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -397,6 +397,9 @@ def _boundary_ccl_unions(boundary_polys, atol, rtol, union):
     transitive closure matches numpy's single-chunk labelling.
     """
     # Global (col, row) -> value and -> set(owner polygon index) maps.
+    # A coordinate is one physical pixel, so every owner sharing it
+    # carries the same value and the same W / S match flags; overwriting
+    # ``field`` and reading the flags from any one owner is therefore safe.
     field = {}
     owners = {}
     for idx, item in enumerate(boundary_polys):
--- a/xrspatial/polygonize.py
+++ b/xrspatial/polygonize.py
@@ -332,14 +332,18 @@ def _ranges_close(range_a, range_b, atol, rtol):
     return False
 
 
-def _cells_close_directed(cells_a, cells_b, atol, rtol, connectivity_8):
+def _cells_close_directed(cells_a, cells_b, atol, rtol, connectivity_8,
+                          cell_field=None):
     """Direction-aware close-value test between two chunk-boundary polygons.
 
-    ``cells_a`` / ``cells_b`` map ``(global_col, global_row) -> value`` for
-    the pixels each polygon places on an internal chunk edge.  Return True
-    if any pixel in ``cells_a`` is grid-adjacent to any pixel in ``cells_b``
-    (orthogonally for 4-conn, plus diagonally for 8-conn) AND the pair
-    compares close under the *same orientation* numpy CCL uses.
+    ``cells_a`` / ``cells_b`` map ``(global_col, global_row) ->
+    (value, w_match, s_match)`` for the pixels each polygon places on an
+    internal chunk edge.  ``w_match`` / ``s_match`` record whether numpy
+    CCL connected that pixel to its in-chunk W / S orthogonal neighbour
+    (same region).  Return True if any pixel in ``cells_a`` is grid-adjacent
+    to any pixel in ``cells_b`` (orthogonally for 4-conn, plus diagonally
+    for 8-conn) AND the pair compares close under the *same orientation*
+    numpy CCL uses.
 
     numpy's ``_calculate_regions`` walks pixels in raster-scan order and
     tests ``_is_close(reference=current, value=earlier_neighbour)`` where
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: polygonize 8-conn dask divergence fix (#3041)

None.

- `_boundary_ccl_unions` (polygonize.py:417-425): `field[coord]` keeps the last value written for a coordinate, and `w_flag`/`s_flag` come from whichever owner the iteration hits first. Both are safe because a coordinate maps to a single physical pixel, so every owner sharing it carries the same value and the same numpy W/S match flags. That invariant is load-bearing but only implied by the code. A one-line comment stating it would help the next reader.

- `TestEightConnFuzzParity` runs 40 random rasters per rtol across all chunkings, deterministic via the seed, and prints the array and chunks on failure (good). Consider also asserting the numpy reference polygon count is stable so a numpy-side change is caught separately.

### What looks good
- The diagnosis is correct: numpy's 8-conn diagonal is conditional on the orthogonal neighbour, and the old pairwise test honoured it unconditionally. Replaying the predicate over the boundary-cell union is the right fix rather than patching the pairwise heuristic.
- The dual-source orthogonal-match decision (real neighbour value when it is a boundary cell, recorded in-chunk flag otherwise) covers both the cross-chunk and interior-guard cases. I checked the nx==1 padding path and the W/S indexing holds.
- 4-conn and integer paths are untouched; the pairwise test bails early for 8-conn float, so there is no double-decision.
- Tests pin the issue repro across every chunking plus a deterministic fuzz, and confirm the 4-conn and integer 8-conn paths are unaffected.

### Checklist
- [x] Algorithm matches numpy CCL semantics (verified against _calculate_regions)
- [x] dask+numpy and dask+cupy merge paths consistent with numpy
- [x] NaN handling unchanged (float mask path untouched)
- [x] Edge cases covered (every chunking incl. 1x1, single row/col)
- [x] No premature materialization or copies in the new code
- [x] Benchmark already exists (benchmarks/benchmarks/polygonize.py)
- [x] No README change needed (bugfix, no API change)

brendancol (commented):
## Follow-up review (after 780bdc67)

- Suggestion (single-pixel invariant): added a comment in `_boundary_ccl_unions` stating that a coordinate is one physical pixel, so overwriting `field` and reading the match flags from any one owner is safe.
- Nit (numpy reference stability): `TestEightConnFuzzParity` now accumulates the numpy polygon count over the fixed-seed sweep and asserts it against pinned per-rtol totals, so a numpy-side CCL change is caught directly instead of silently moving the dask parity reference.

No logic changed; the follow-up is a comment plus a test assertion. Full file `test_polygonize_issue_2677.py` passes (41 passed) and flake8 is clean. No new findings.

PR #3045, opened 2026-06-08
Add cross-backend NaN-input parity test for contours()

### LINKED ISSUE #3044: contours(): no cross-backend parity test for NaN input
The backend-equivalence tests in `test_contour.py` (`TestBackendEquivalence`) all use `elevation_raster_no_nans`, so none of them feed a raster with NaN cells through the four backends. The one NaN-input test, `test_partial_nan`, runs on numpy only.

That leaves a gap. The numpy backend skips quads with a non-finite corner in the interior. The dask backend adds a NaN halo via `da.overlap` and stitches segments across chunk boundaries. A NaN cell next to a chunk edge could be handled differently from a NaN cell in the numpy interior, and nothing asserts the two agree.

I checked on a CUDA host that all four backends (numpy, cupy, dask+numpy, dask+cupy) currently produce identical contour segments for a partial-NaN raster, so this guards against a future regression rather than flagging an open bug.

Proposed: add a test class that runs a partial-NaN raster (an interior NaN cell plus a NaN edge row) through all four backends and asserts segment equality against the numpy result. Reuse the existing `_segments_by_level` helper and the `dask_array_available` / `cuda_and_cupy_available` decorators. Test-only, no source change.

Found by the deep-sweep test-coverage sweep on module `contour`.

brendancol:
Closes #3044

Found by the deep-sweep test-coverage sweep on module `contour`.

The backend-equivalence tests use `elevation_raster_no_nans`, and the one NaN-input test runs on numpy only. Nothing pinned numpy/cupy/dask parity when a NaN cell sits next to a chunk edge, where the dask NaN halo (`da.overlap`) meets the numpy interior NaN-skip.

- Adds `TestNaNBackendParity` with three tests (dask+numpy, cupy, dask+cupy), each asserting `_segments_by_level` equality against the numpy result on a partial-NaN ramp (a NaN edge row plus an interior NaN cell inside a non-edge chunk).
- Reuses the existing `_segments_by_level` helper and the `dask_array_available` / `cuda_and_cupy_available` decorators.
- Test-only. `contour.py` is untouched. The four backends already agree, so this guards against a future regression.

Backend coverage: numpy (reference), cupy, dask+numpy, dask+cupy.

Test plan:
- [x] `TestNaNBackendParity` runs and passes on a CUDA host (3 passed, cupy and dask+cupy executed, not skipped).
- [x] Full `test_contour.py` green: 89 passed, 0 skipped.


brendancol (commented):
## PR Review: Add cross-backend NaN-input parity test for contours()

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/tests/test_contour.py` (TestNaNBackendParity): the three tests repeat the same build-and-compare body. A `@pytest.mark.parametrize` over backend would cut the duplication, but `TestBackendEquivalence` and `TestIntegerDtypeCollar` already use the same one-method-per-backend layout, so matching them is the right call. Not worth changing.

### What looks good
- The NaN placement is deliberate. A NaN edge row plus an interior NaN cell at `data[5,6]`, which with `chunks=(4,4)` lands inside the middle (non-edge) chunk. That is exactly where the dask NaN halo could diverge from the numpy interior, so the test has real teeth rather than just exercising the easy edge case.
- Reuses the module-level `_segments_by_level` helper and the existing `dask_array_available` / `cuda_and_cupy_available` decorators. No new helpers, no source change.
- GPU cells are not skip-guarded away on a CUDA host. The PR reports all three ran and passed, full file 89 passed / 0 skipped.
- The state CSV change also drops a stray duplicate header line left by an earlier union-merge. Clean side effect, no other module rows touched.

### Checklist
- [x] NaN handling is correct
- [x] Edge cases covered (NaN edge row + interior NaN at a chunk boundary)
- [x] Docstrings present

Test-only PR, source untouched.

PR #3047, opened 2026-06-08
Make reproject output-size guard backend-aware (#3046)

### LINKED ISSUE #3046: reproject() output-size guard blocks legitimate lazy dask outputs
**Describe the bug**

`reproject()` rejects any computed output grid larger than 1 billion pixels via a hard guard in `_compute_output_grid` (`xrspatial/reproject/_grid.py`). The guard runs before backend detection and treats every backend the same. It doesn't distinguish between the paths that materialize the full output array (in-memory numpy, in-memory cupy, the streaming fallback) and the dask path, which builds a lazy `map_blocks` graph and never holds the whole grid in memory.

So a dask-backed input can't be reprojected when the output grid is large, even though peak memory there is bounded by the chunk size, not the total pixel count.

Example: a CONUS-wide 30 m LANDFIRE EVC raster in EPSG:5070, opened with `chunks=2048`, reprojected to EPSG:3857. The output grid computes to 186095 x 112377 (about 20.9 billion pixels), which is the correct size for Web Mercator over a continental extent, and the call raises:

\`\`\`
ValueError: Computed output grid is too large (186095 x 112377 = 20,912,797,815 pixels, limit is 1,000,000,000). Increase the resolution parameter or reduce the output extent.
\`\`\`

**Expected behavior**

For dask-backed (lazy) outputs the call should proceed and return a lazy dask DataArray that computes chunk by chunk on demand. The 1 billion pixel cap should apply only to backends that allocate the full output in memory: in-memory numpy, in-memory cupy, and the streaming fallback.

**Additional context**

The guard's own comment justifies the cap on memory grounds ("1 billion pixels ~= 8 GB for a single float64 array"), which only holds for the materializing backends. The fix is to make the guard backend-aware: bypass it when the output is a lazy dask array, keep it otherwise. No new public parameter is needed.

brendancol:
Closes #3046

`reproject()` computed the output grid and applied the 1-billion-pixel memory guard before it detected the backend, so it treated every backend the same. That blocked dask-backed inputs whose output is lazy and never held in memory all at once.

This moves backend detection ahead of the grid computation and threads a `lazy_output` flag into `_compute_output_grid`. When the output is a lazy dask array the size guard is skipped; for the materializing backends (in-memory numpy, in-memory cupy, the streaming fallback) it stays in force.

- Detect the backend (including large-numpy auto-promotion to dask) before computing the output grid
- Skip the `_MAX_OUTPUT_PIXELS` guard for lazy dask outputs, keep it for materializing backends
- No new public parameter; behavior keys purely on the backend

Backend coverage: numpy and cupy keep the guard; dask+numpy and dask+cupy now produce lazy outputs above the old cap.

Test plan:
- [x] `_compute_output_grid(..., lazy_output=True)` over 1e9 pixels does not raise; default still raises
- [x] dask input forced over 1e9 output pixels reprojects lazily and a single block computes
- [x] full `xrspatial/tests/test_reproject.py` suite passes (425 tests)

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -936,20 +936,9 @@ def reproject(
     y_desc = _is_y_descending(raster)
     x_desc = _is_x_descending(raster)
 
-    # Compute output grid
-    grid = _compute_output_grid(
-        src_bounds, src_shape, src_crs, tgt_crs,
-        resolution=resolution, bounds=bounds,
-        width=width, height=height,
-        bounds_policy=bounds_policy,
-    )
-    out_bounds = grid['bounds']
-    out_shape = grid['shape']
-
-    # Output coordinates
-    y_coords, x_coords = _make_output_coords(out_bounds, out_shape)
-
-    # Detect backend
+    # Detect backend before computing the output grid so the grid's
+    # output-size guard can tell a lazy dask output (never materialized
+    # in full) from a materializing backend.
     from ..utils import has_dask_array, is_cupy_array
 
     data = raster.data
@@ -993,6 +982,21 @@ def reproject(
             except ImportError:
                 # dask not available -- fall back to streaming
                 _use_streaming = True
+
+    # Compute output grid. A dask output is lazy, so the output-size
+    # guard is skipped for it (peak memory is bounded by chunk size).
+    grid = _compute_output_grid(
+        src_bounds, src_shape, src_crs, tgt_crs,
+        resolution=resolution, bounds=bounds,
+        width=width, height=height,
+        bounds_policy=bounds_policy,
+        lazy_output=is_dask,
+    )
+    out_bounds = grid['bounds']
+    out_shape = grid['shape']
+
+    # Output coordinates
+    y_coords, x_coords = _make_output_coords(out_bounds, out_shape)
 
     # Serialize CRS for pickle safety
     src_wkt = src_crs.to_wkt()
--- a/xrspatial/reproject/_grid.py
+++ b/xrspatial/reproject/_grid.py
@@ -179,7 +179,7 @@ def _transform_boundary(source_crs, target_crs, xs, ys):
 
 def _compute_output_grid(source_bounds, source_shape, source_crs, target_crs,
                          resolution=None, bounds=None, width=None, height=None,
-                         bounds_policy="auto"):
+                         bounds_policy="auto", lazy_output=False):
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Make reproject output-size guard backend-aware (#3046)

None.

- `xrspatial/tests/test_reproject.py` `test_reproject_dask_output_over_limit_stays_lazy`: the assertion computes `out.data.blocks[0, 0]`, which with `chunk_size=20000` is a 20000x20000 float64 block, about 3.2 GB in RAM. That is heavy for CI runners and could OOM or run slowly. A smaller `chunk_size` (e.g. 1024) still forces a >1e9 total output while keeping the computed block cheap.
- `xrspatial/reproject/__init__.py:2310` `merge()` has the same shape as the bug fixed here: it computes the output grid (and so runs the 1e9 guard) before detecting whether the result will be dask, and it even auto-promotes large in-memory merges to `_merge_dask`. A large lazy merge would still hit the guard. Out of scope for #3046, but worth a follow-up issue.

None.

### What looks good
- Backend detection moved ahead of the grid computation cleanly; nothing in that block depended on the grid, so the reordering is safe.
- `lazy_output` defaults to False, so the other two `_compute_output_grid` callers and direct callers keep the strict guard.
- Tests cover both directions: lazy bypass succeeds and the non-lazy path still raises.

### Checklist
- [x] Logic matches the intended fix (skip guard only for lazy dask)
- [x] All implemented backends behave correctly (numpy/cupy keep guard, dask bypasses)
- [x] No NaN-handling changes in this PR
- [x] Edge cases covered (lazy over limit, non-lazy still raises)
- [x] Dask path stays lazy; a single block computes without materializing the grid
- [ ] Benchmark not needed (no perf-sensitive change)
- [ ] README feature matrix not applicable (no new function, no backend change)
- [x] Docstring added for the new parameter

brendancol (commented):
## Follow-up review (after de6d70be)

Disposition of the prior findings:

- Fixed: the integration test now uses `chunk_size=1024`, so the computed block is 1024x1024 (about 8 MB) instead of 20000x20000 (about 3.2 GB). It runs in ~1.4s and no longer risks OOM on CI.
- Deferred: the same guard-before-backend-detection pattern in `merge()` is tracked in #3048. Fixing it cleanly needs a small reorder around merge's auto-promote step, which would expand this PR beyond #3046.

No new findings. The reproject change is self-contained and the full `test_reproject.py` suite passes.

PR #3051, opened 2026-06-08
Test: parametrize dask+cupy reproject parity over resampling modes (#3050)

### LINKED ISSUE #3050: reproject() dask+cupy backend: nearest/bilinear resampling lack parity tests
## Coverage gap

`reproject()`'s dask+cupy backend is only parity-checked with `resampling='cubic'`. The test `TestCupyPyprojFallbackParity::test_projected_to_projected_dask_cupy_match` in `xrspatial/tests/test_reproject.py` reprojects an EPSG:32633 -> EPSG:3857 raster on a dask-of-cupy array and compares against the numpy reference, but only for cubic.

`nearest` and `bilinear` are covered on numpy (end-to-end) and on eager cupy (the parametrized `test_projected_to_projected_numpy_cupy_match` runs all three), but never on the dask+cupy chunk-assembly path. So if the dask+cupy path threaded the resampling mode through to `_resample_cupy_native` incorrectly per chunk, the two non-cubic modes would slip through.

## Proposed fix

Parametrize `test_projected_to_projected_dask_cupy_match` over `['nearest', 'bilinear', 'cubic']`. Each mode asserts the dask+cupy result matches the numpy reference: same finite/nodata pattern, finite values within rtol/atol 1e-6.

Test-only. I ran all three modes on a CUDA host and they already pass, so this closes a coverage gap rather than fixing a bug.

## Backends affected

dask+cupy (test coverage). numpy, eager cupy, and dask+numpy already have per-mode coverage.

_Filed by the test-coverage deep-sweep on the reproject module._

brendancol:
Closes #3050

`reproject()`'s dask+cupy backend was parity-tested only with `resampling='cubic'`. `nearest` and `bilinear` were covered on numpy (end-to-end) and eager cupy (the parametrized `test_projected_to_projected_numpy_cupy_match`), but never on the dask+cupy chunk-assembly path.

- Parametrize `TestCupyPyprojFallbackParity::test_projected_to_projected_dask_cupy_match` over `['nearest', 'bilinear', 'cubic']`.
- Each mode asserts the dask+cupy result matches the numpy reference: identical finite/nodata pattern, finite values within rtol/atol 1e-6.

Test-only, no source change. The dask+cupy `cubic` path already passed; this extends the same assertion to the two non-cubic modes.

Backend coverage: dask+cupy (test). numpy, eager cupy, and dask+numpy already had per-mode coverage.

Test plan:
- [x] `pytest test_projected_to_projected_dask_cupy_match` runs all three modes green on a CUDA host
- [x] Source untouched

_Generated by the test-coverage deep-sweep on the reproject module._


brendancol (commented):
Domain-aware review (test-coverage sweep, reproject):

Blockers: none.
Suggestions: none.
Nits: none.

Notes:
- Parity assertion (exact finite/nodata pattern + finite values within rtol/atol 1e-6) matches the sibling eager-cupy test `test_projected_to_projected_numpy_cupy_match`, so the dask+cupy variant now mirrors it across all three modes.
- The EPSG:32633 -> EPSG:3857 projected-to-projected fixture is the right choice: neither side is WGS84/NAD83, so both backends take the pyproj coordinate fallback, which is exactly the path the dask+cupy chunk assembly threads `_resample_cupy_native` through per chunk.
- Skip guards are correct: class-level `HAS_CUPY` plus method-level `HAS_DASK`.
- Source untouched, as required for the test-coverage sweep. All three parametrized cases ran and passed on a CUDA host.

PR #3059, opened 2026-06-09
Reject fills the output dtype cannot represent exactly

### LINKED ISSUE #3054: rasterize: integer/boolean fill can silently corrupt output and nodata attrs
**Describe the bug**

`rasterize()` only rejects `fill=NaN` against integer output dtypes. Two related cases slip through and corrupt both the output and its nodata metadata:

- An integer fill that doesn't fit the output dtype is cast silently. `dtype=np.uint8, fill=-9999` writes an actual fill of 241, but the returned DataArray's `nodata`/`_FillValue`/`nodatavals` attrs still advertise -9999.
- Boolean dtypes aren't covered by the integer guard. `dtype=np.bool_, fill=np.nan` casts every unwritten pixel to `True`, and since the fill is NaN no nodata attrs get emitted at all.

In both cases the stored fill value disagrees with what was actually burned into the array. Downstream masking and GeoTIFF output key off these attrs, so unwritten pixels get treated as valid data.

**Where**

- The guard in `xrspatial/rasterize.py` rejects only NaN-for-integer.
- The final cast happens per backend (`out.astype(dtype)` in `_run_numpy`, plus the cupy/dask equivalents).
- The attrs block stores the original `fill`, not the cast value.

**Expected behavior**

Reject a fill that can't be represented exactly in the output dtype (integer overflow and boolean both included), or only store a fill in attrs after confirming it round-trips through the dtype. Boolean needs explicit handling.

**Reproduction**

```python
import numpy as np, geopandas as gpd

gdf = gpd.GeoDataFrame({'v': [1]}, geometry=[box(0, 0, 1, 1)])
r = rasterize(gdf, column='v', width=4, height=4, dtype=np.uint8, fill=-9999)
print(r.values.max(), r.attrs.get('nodata'))   # 241 burned, attrs say -9999
```

**Additional context**

Found during a metadata-propagation review of rasterize.py.

brendancol:
Closes #3054

`rasterize()` casts its float64 work buffer to the output dtype at the end of every backend, while the attrs block stores the original `fill` verbatim. The old guard only caught `fill=NaN` against integer dtypes, so two cases slipped through and left the burned array disagreeing with its `nodata`/`_FillValue`/`nodatavals` attrs:

- `dtype=np.uint8, fill=-9999` burned an actual fill of 241 but advertised -9999.
- `dtype=np.bool_, fill=np.nan` turned every unwritten pixel into `True` with no nodata attrs (since `np.issubdtype(np.bool_, np.integer)` is False, bool dodged the integer-only guard).

## Change

- Replace the NaN-only-integer guard with a round-trip representability test: for integer and boolean output dtypes, reject any fill that changes value when cast to the dtype and back. Float dtypes are left alone (NaN is representable, ordinary rounding is expected).
- Update the `fill` docstring to describe the broadened guard.
- Relax the #2504 regression match strings to the shared `"cannot be represented"` substring.

## Backend coverage

The guard runs before any host/device allocation, so it fires identically on numpy, cupy, dask+numpy, and dask+cupy. New tests parametrize the out-of-range-int and bool cases across all four (cupy/dask paths skip when unavailable).

## Test plan

- [x] New regression tests in `test_rasterize_fill_dtype_3054.py` pass
- [x] Existing #2504 guard tests still pass
- [x] `test_rasterize.py` full suite passes (249 passed, 2 skipped)

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -3039,13 +3039,16 @@ def rasterize(
         length ``len(columns)`` as its ``props`` argument.
     fill : float, default np.nan
         Value for pixels not covered by any geometry.  When ``dtype``
-        resolves to an integer type (either via the ``dtype`` argument or
-        via ``like`` carrying an integer dtype), the default NaN fill is
-        rejected with ``ValueError`` because NaN has no integer
-        representation and the cast would silently land on a
-        platform-specific sentinel with no ``_FillValue`` attr to mark
-        it; pass an explicit integer sentinel (e.g. ``fill=0`` or
-        ``fill=-9999``) or use a floating dtype.
+        resolves to an integer or boolean type (either via the ``dtype``
+        argument or via ``like``), a fill the dtype cannot represent
+        exactly is rejected with ``ValueError``.  This covers NaN into an
+        integer dtype (which would land on a platform sentinel), an
+        out-of-range integer (``fill=-9999`` into ``uint8`` wraps to
+        241), and any non-False value into ``bool`` (which collapses to
+        ``True``).  Without the guard the burned array and its emitted
+        ``nodata`` / ``_FillValue`` attrs would disagree.  Pass a fill the
+        dtype can hold (e.g. ``fill=0`` or ``fill=-9999`` for integers,
+        ``fill=False`` for bool) or use a floating dtype.
     dtype : numpy dtype, optional
         Data type of the output array.  Defaults to np.float64, or
         to the dtype of ``like`` if provided.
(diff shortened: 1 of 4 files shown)


brendancol (commented):
## PR Review: Reject fills the output dtype cannot represent exactly

None.

- `xrspatial/rasterize.py:3378` — the round-trip cast originally caught only `(TypeError, ValueError)`. A fill wider than the platform C long (e.g. `fill=2**100`) lands in an object array whose `astype(<int>)` raises `OverflowError`, which would escape as a stray traceback instead of the guard's actionable message. Fixed during review by adding `OverflowError` to the caught set and a regression test (`test_huge_int_fill_raises_cleanly`).

None.

### What looks good
- The round-trip representability test is the right generalization. It subsumes the old NaN-into-integer case and extends cleanly to out-of-range integers and bool, without special-casing each.
- Bool is handled explicitly via the `final_dtype_np == np.bool_` branch, since `np.issubdtype(np.bool_, np.integer)` is False.
- Float dtypes are deliberately untouched, so NaN fills and ordinary float rounding still pass.
- The guard sits before backend dispatch in the single public entry point, so all four backends (numpy, cupy, dask+numpy, dask+cupy) hit it before any allocation. Tests parametrize each.
- The `RuntimeWarning` from the NaN-to-int probe cast is suppressed locally, matching the existing pattern in `_run_numpy`.
- The valid-fill round-trip tests confirm the nodata/_FillValue/nodatavals triplet stays consistent with the burned value.

### Checklist
- [x] Algorithm matches the stated contract (exact round-trip)
- [x] All four backends hit the guard before allocation
- [x] NaN handling correct (float NaN passes, int/bool NaN rejected)
- [x] Edge cases covered (out-of-range int, bool, huge int, valid round-trips)
- [x] No premature materialization (guard is a scalar check)
- [x] Docstrings present and updated for the broadened guard

PR #3060, opened 2026-06-09
Reject large integer burn values that float64 cannot hold exactly (#3056)

### LINKED ISSUE #3056: rasterize() loses precision on large integer burn values (float64 cast)
**Describe the bug**

`rasterize()` casts burn values to float64 before rasterization, so large integer burn values lose precision even when `dtype` is an integer type. The internal raster and the merge contract are float64 throughout: internal props are float64 (`xrspatial/rasterize.py` near line 102), GeoDataFrame columns are cast with `.astype(np.float64)` (near lines 2773 and 2783), and `(geometry, value)` pairs go through `float()` (near line 2797). Integers above 2**53 are not exactly representable in float64, so the cast rounds them without warning.

This hits identifiers that often get burned into rasters: zone IDs, parcel IDs, category IDs, and uint64 identifiers.

**Reproduction**

```python
import numpy as np

v = 2**53 + 1  # 9007199254740993
r = rasterize([(box(0, 0, 5, 5), v)], width=4, height=4, fill=0, dtype=np.int64)
print(int(r.values[0, 0]))  # 9007199254740992 -- off by one
```

**Expected behavior**

When the output dtype is an integer type, a burn value that cannot be represented exactly as float64 should not be silently corrupted. Reject it with a clear `ValueError` that names the offending value, the same way the existing guard rejects a NaN fill against an integer dtype.

**Additional context**

The internal pipeline is float64 end to end, so preserving integer props throughout would be a large change. Rejecting unrepresentable integer values when the output dtype is integer is the smaller fix, and it follows the precedent of the NaN-fill-vs-integer-dtype guard already in the function.

brendancol:
Closes #3056

## Problem

`rasterize()` computes in float64 and casts burn values to float64 at parse time. Integers above 2**53 - 1 are not exactly representable in float64, so a large integer ID burned into an integer-dtype raster was silently rounded. The reproduction burned 2**53 + 1 with dtype=np.int64 and got 9007199254740992 back instead of 9007199254740993. Zone IDs, parcel IDs, and uint64 identifiers fall in this range.

## Fix

Reject burn values whose magnitude exceeds the float64 safe-integer limit (2**53 - 1) when the output dtype is an integer type. The guard raises a ValueError naming the offending value, placed next to the existing NaN-fill-vs-integer-dtype guard. It runs before any host or device allocation, so it fires identically across all four backends. Float output dtypes are unaffected: the float64 value is what the user asked to store.

This is the reject approach rather than threading integer props through the whole pipeline. The internal computation is float64 end to end, so preserving exact integers throughout would be a much larger change.

## Backend coverage

numpy, cupy, dask+numpy, dask+cupy. The check sits in the shared front-end before backend dispatch, so all four trip it the same way.

## Test plan

- [x] Reproduction value now raises instead of returning an off-by-one
- [x] int64 and uint64 dtypes both reject the unsafe value
- [x] Negative large values rejected (magnitude check)
- [x] 2**53 - 1 (max safe integer) burns exactly on numpy and dask
- [x] Small integer IDs unaffected
- [x] Float output dtype accepts the value unchanged
- [x] `like` with integer dtype trips the guard
- [x] GeoDataFrame column and columns burn paths trip the guard
- [x] dask+numpy backend; cupy / dask+cupy tests gated on CUDA
- [x] Existing rasterize suite still green (233 passed, 2 skipped)

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -3048,7 +3048,14 @@ def rasterize(
         ``fill=-9999``) or use a floating dtype.
     dtype : numpy dtype, optional
         Data type of the output array.  Defaults to np.float64, or
-        to the dtype of ``like`` if provided.
+        to the dtype of ``like`` if provided.  When this resolves to an
+        integer type, burn values are validated against the float64 safe
+        integer range: the rasterizer computes in float64, so a value
+        with magnitude above ``2**53 - 1`` cannot be cast back to an
+        exact integer (e.g. ``2**53 + 1`` would land on ``2**53``).  Such
+        a value is rejected with ``ValueError`` rather than silently
+        rounded; use a floating dtype to burn identifiers larger than
+        that.
     all_touched : bool, default False
         If True, every pixel a polygon boundary passes through is
         burned in addition to the normal center-fill, using a
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -3370,6 +3370,39 @@ def rasterize(
             f"explicit integer fill (e.g. fill=0 or fill=-9999) or use a "
             f"floating dtype.")
 
+    # Reject burn values that float64 cannot hold exactly when the output
+    # dtype is integer.  ``_parse_input`` casts every burn value to float64
+    # (props_array is float64, GeoDataFrame columns go through
+    # ``.astype(np.float64)``, and ``(geom, value)`` pairs go through
+    # ``float()``), and the whole rasterize pipeline computes in float64
+    # before the final ``astype(int_dtype)``.  Integers with magnitude above
+    # 2**53 - 1 (the IEEE-754 "max safe integer") are not exactly
+    # representable, so an ID like ``2**53 + 1`` silently lands on
+    # ``2**53`` -- off by one -- by the time it reaches the output.  Zone
+    # IDs, parcel IDs, and uint64 identifiers are exactly the values that
+    # exceed this range, so a silent round is a correctness failure rather
+    # than a rounding nicety.  Reject up front with the offending value
+    # named, mirroring the NaN-fill guard above.  Float dtypes are
+    # unaffected: the float64 value is what the user asked to store.
+    #
+    # Checked before any host / device allocation so the error surfaces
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Reject large integer burn values that float64 cannot hold exactly (#3056)

None.

None.

- rasterize.py:3399-3402: the guard rejects exactly 2**53, which is itself representable in float64. This is conservative: float64 cannot distinguish a user passing 2**53 from one passing 2**53 + 1 (both round to 2**53), so rejecting at |value| > 2**53 - 1 is the defensible choice and the message/docstring state the boundary plainly. No change needed unless you want to special-case the exact even power. Recording it so the boundary choice is on the record.

### What looks good
- The guard sits in the shared front-end before backend dispatch, so numpy, cupy, dask+numpy, and dask+cupy all reject the same inputs with no per-backend code. Tests confirm this on numpy and dask+numpy, with cupy/dask+cupy gated on CUDA availability.
- np.isfinite masks NaN and inf props out before the magnitude compare, so a NaN burn (used by some merge modes) does not trip the guard spuriously.
- Placement mirrors the existing NaN-fill-vs-integer-dtype guard and runs before any allocation, matching that precedent.
- Tests cover the reproduction, int64/uint64, negative magnitudes, the max-safe-integer round-trip, small ints, float dtype passthrough, the like template path, and both GeoDataFrame column and columns paths.
- Float output dtype is correctly left alone.

### Checklist
- Algorithm matches reference: the 2**53 - 1 safe-integer boundary is the IEEE-754 definition, correct.
- All implemented backends produce consistent results: yes, single front-end check.
- NaN handling is correct: NaN/inf props excluded via np.isfinite.
- Edge cases covered by tests: max-safe boundary, negative, small, float, like, GeoDataFrame paths.
- Dask chunk boundaries: not applicable, the check runs before graph construction.
- No premature materialization: the check operates on the already-parsed props array, no new compute.
- Benchmark: not needed, this is a validation guard on an existing function.
- README feature matrix: not applicable, no new function and no backend-support change.
- Docstrings present and accurate: the dtype parameter documents the new behavior.

PR #3061, opened 2026-06-09
Warn on non-atomic custom GPU merge over overlap

### LINKED ISSUE #3057: rasterize: custom CUDA merge functions are documented as supported but race on overlap
## Description

`rasterize()` tells users in its docstring that they can pass a `@numba.cuda.jit(device=True)` merge function for the GPU backend, no caveat attached. The GPU path only uses CUDA atomics for the six built-in merges (sum, count, min, max, first, last). A custom merge callable instead falls into the legacy non-atomic read-modify-write closure path in `_apply_merge_gpu`. When geometries overlap, several GPU threads can read-modify-write the same pixel at once, so the value for an overlapping pixel is nondeterministic and need not match the CPU backend.

The internal code already knows this (`_GPU_ATOMIC_MERGES`, and the comments in `_ensure_gpu_kernels`), but nothing in the public `merge` docstring tells the caller.

## Where

- Built-in atomic merges: `xrspatial/rasterize.py`, `_apply_merge_gpu` (the sum/count/min/max/first/last branches)
- Non-atomic path for custom callables: same function, the trailing `else` branch
- Public docstring: the `merge` parameter of `rasterize()`

## Proposed fix

Warn about it in the public docstring: a custom GPU merge function over overlapping geometries is nondeterministic because the per-pixel update is a non-atomic read-modify-write, and a built-in string merge is the safe choice when overlap determinism matters. Also emit a `UserWarning` when a callable `merge` is paired with `use_cuda=True`, so a caller sees it when they run the code, not only if they read the docs.

## Backends affected

cupy and dask+cupy. The CPU backends (numpy, dask+numpy) write each pixel from a single thread and are not affected.

brendancol:
Closes #3057

The GPU backends only use CUDA atomics for the six built-in merges. A custom callable `merge` falls into the non-atomic read-modify-write path in `_apply_merge_gpu`, so overlapping geometries race and the overlap pixels are nondeterministic. The public docstring said nothing about this.

- Added a `.. warning::` to the `merge` docstring explaining that a custom GPU merge is nondeterministic over overlap, and that the built-in string merges stay deterministic because they use atomics.
- Emit a `UserWarning` when a callable `merge` is combined with `use_cuda=True` (covers both `cupy` and `dask+cupy`). The warning fires up front, before the geometry burn and the CuPy check.

No behavior change for existing results; the GPU callable path is unchanged.

## Backends

- numpy / dask+numpy: unaffected (CPU writes each pixel single-threaded; no warning).
- cupy / dask+cupy: warning now fires for callable merges.

## Test plan

- [x] New tests in `test_rasterize_gpu_callable_warn_3057.py`: warning fires for callable + `use_cuda=True` and for the `chunks` (dask+cupy) variant; no warning for a CPU callable merge or for a built-in GPU merge. They record warnings manually so they pass without a GPU.
- [x] Existing `test_rasterize.py` suite still passes (215 passed, 2 skipped).
- [x] flake8 clean on the changed files.

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -3107,6 +3107,21 @@ def rasterize(
         - *props*: 1D float64 array of property values for the geometry
         - *is_first*: 1 on first write to this pixel, 0 otherwise
 
+        .. warning::
+
+           On the GPU backends (``use_cuda=True``, with or without
+           ``chunks``) a custom callable does not use CUDA atomics.  Its
+           per-pixel update is a non-atomic read-modify-write, so when
+           geometries overlap, several threads may update the same pixel
+           at once and the result for those pixels is nondeterministic --
+           it can vary between runs and need not match the CPU backend.
+           The built-in string merges (``'sum'``, ``'count'``, ``'min'``,
+           ``'max'``, ``'first'``, ``'last'``) do use atomics and stay
+           deterministic over overlap; pass one of those if you need a
+           stable result where geometries overlap on the GPU.  Calling
+           ``rasterize`` with a callable ``merge`` and ``use_cuda=True``
+           emits a ``UserWarning`` to this effect.
+
     chunks : int or (int, int), optional
         If given, use the dask backend and split the output raster into
         tiles of this size ``(row_chunk, col_chunk)``.  Both axes must be
@@ -3163,6 +3178,23 @@ def rasterize(
         merge_fn = merge
         should_write_cpu = _should_write_any
         _merge_fn_gpu = merge  # same object for GPU path
+        if use_cuda:
+            # The GPU path has no atomic variant for an arbitrary callable,
+            # so it falls back to a non-atomic read-modify-write per pixel
+            # (see _apply_merge_gpu).  Overlapping geometries then race and
+            # the overlap pixels are nondeterministic.  Warn at call time;
+            # a built-in string merge stays deterministic over overlap.
+            warnings.warn(
+                "A custom callable merge on the GPU backend "
+                "(use_cuda=True) uses a non-atomic read-modify-write, so "
+                "values for pixels where geometries overlap are "
+                "nondeterministic and may not match the CPU backend. Use a "
+                "built-in string merge ('sum', 'count', 'min', 'max', "
+                "'first', 'last') if you need a deterministic result over "
+                "overlap.",
+                UserWarning,
+                stacklevel=2,
+            )
     elif isinstance(merge, str):
         if merge not in _MERGE_FUNCTIONS:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Warn on non-atomic custom GPU merge over overlap

### Blockers
None.

### Suggestions
None blocking. One thing worth weighing (see nits).

### Nits (optional)
- `xrspatial/rasterize.py` (~line 3181): the warning fires before the `cupy is None` ImportError guard at ~line 3413. A caller with no CuPy who passes a callable merge plus `use_cuda=True` gets the UserWarning first, then the ImportError. The ordering is harmless since the warning is true either way, but if you'd rather a missing-dependency caller only see the hard error, move the warn after the cupy check. Not a blocker.
- `test_rasterize_gpu_callable_warn_3057.py` (`_overlap_warnings`): the bare `except Exception: pass` is intentional and documented (no GPU in CI), but it would also swallow a real regression where rasterize raised before the warn line. The `len(matched) == 1` assert still catches the "warning never fired" case, so coverage holds. Flagging only so it stays a conscious choice.

### What looks good
- Warning sits in the single callable-merge branch, so it covers both cupy and dask+cupy without duplicated logic.
- The docstring warning names the deterministic built-in merges, so the user gets an actionable alternative.
- Tests check both directions: callable+GPU warns; CPU callable and built-in GPU merge stay silent. They run without a GPU by recording warnings manually.
- No behavior change to the existing GPU callable path; the existing rasterize suite stays green.

### Checklist
- Algorithm matches reference: n/a (doc + warning only)
- All implemented backends consistent: unchanged
- NaN handling correct: unchanged
- Edge cases covered by tests: yes (CPU vs GPU, callable vs string)
- Dask chunk boundaries handled: unchanged; chunks path has a warning test
- Benchmark exists or not needed: not needed (no perf change)
- README feature matrix updated: not needed (no new function, no backend change)

brendancol (commented):

Both nits from the first pass are addressed.

### Resolved
- Warning now fires inside the `else` (callable) branch after the `cupy is None` ImportError check (`xrspatial/rasterize.py` ~line 3420). A caller with no CuPy now gets only the ImportError, not a UserWarning they can't act on. Still covers both the cupy and dask+cupy paths since `gpu_merge_name` is resolved there for both.
- The two GPU warning tests are gated with `@skip_no_cupy`, matching the new ordering. The CPU "does not warn" test completes without a GPU, so it no longer relies on the broad `except` to pass. The remaining `except Exception` only swallows the device-less kernel-launch error, and the exact-count asserts still fail loudly if the warning never fires.

### Remaining
None. No blockers, no open suggestions.

### Verification
- New file `test_rasterize_gpu_callable_warn_3057.py`: 4 passed.
- `test_rasterize.py`: 215 passed, 2 skipped.
- flake8 clean on both changed files.

PR #3062, opened 2026-06-09
Guard rasterize() against geometry/like CRS mismatch (#3058)

### LINKED ISSUE #3058: rasterize() silently ignores CRS mismatch between geometries and like template
**Describe the bug**

`rasterize(geometries, like=template)` ignores any mismatch between the CRS of the input geometries and the CRS of the `like` template. `_parse_input` reads a GeoDataFrame's geometry and property columns but drops `gdf.crs`. When `like` is passed, `_extract_grid_from_like` carries the template's CRS through its non-dim coords (rioxarray's `spatial_ref`) and its `attrs` (`attrs['crs']`), and those land on the output.

So a lat/lon (EPSG:4326) GeoDataFrame can be rasterized against an EPSG:3857 template, and the output comes back tagged EPSG:3857 with no warning. The geometry coordinates were never reprojected, so the raster is wrong, but it carries an authoritative-looking CRS.

**Expected behavior**

When both the geometries and the `like` template carry a CRS, rasterize should compare them and raise when they disagree instead of emitting a mislabeled raster. A parameter should let callers turn the check off when they already know the coordinates line up.

**Additional context**

- `_parse_input` drops `gdf.crs` (xrspatial/rasterize.py around line 2764)
- `_extract_grid_from_like` keeps the template CRS via extra_coords and attrs (around lines 2954, 2970)
- output attrs propagate from `like_attrs` (around line 3462)

The accessor already does this elsewhere: `_open_geotiff_windowed` normalizes both CRS values through pyproj and raises on mismatch unless `auto_reproject=True`.

brendancol:
Closes #3058

## What

`rasterize(geometries, like=template)` no longer silently burns geometries onto a template in a different CRS.

- `_parse_input` now returns the GeoDataFrame's `.crs` instead of dropping it.
- A new `_like_crs` helper reads the template CRS (`attrs['crs']`, `attrs['crs_wkt']`, the `spatial_ref` coord, or `rio.crs`), matching `polygonize._detect_raster_crs`.
- When both sides carry a CRS, `rasterize` normalizes them through pyproj and raises `ValueError` on mismatch. A new `check_crs=True` parameter (default) gates this; `check_crs=False` opts out.

## Why

`_parse_input` discarded `gdf.crs` while `_extract_grid_from_like` propagated the template CRS onto the output. So an EPSG:4326 GeoDataFrame rasterized onto an EPSG:3857 template came back tagged EPSG:3857, with no reprojection and no warning. It looked authoritative but was wrong.

The geometries are never reprojected; reproject them to match the template, or pass `check_crs=False` to burn onto the template grid anyway.

## Backend coverage

The guard runs before backend dispatch, so it behaves the same for numpy, cupy, dask+numpy, and dask+cupy. Tests cover the numpy path plus the dask and cupy paths.

## Test plan

- [x] Matching CRS passes; mismatched CRS raises; `check_crs=False` bypasses
- [x] CRS read from `attrs['crs']`, `attrs['crs_wkt']`, and the `spatial_ref` coord
- [x] No-op when either side lacks a CRS, and for the `(geometry, value)` iterable path
- [x] int EPSG vs `"EPSG:xxxx"` string compare equal
- [x] Invalid CRS value raises instead of disabling the guard
- [x] Existing rasterize suite still green (215 passed)

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -2742,13 +2742,16 @@ def _run_dask_cupy(geometries, props_array, bounds, height, width, fill,
 # ---------------------------------------------------------------------------
 
 def _parse_input(geometries, column=None, columns=None):
-    """Normalise input to (geometry_list, props_array, bounds).
+    """Normalise input to (geometry_list, props_array, bounds, crs).
 
     Returns
     -------
     geom_list : list of shapely geometries
     props_array : (N, P) float64 array of property values
     bounds : tuple or None
+    crs : the GeoDataFrame's ``.crs`` (any pyproj-parseable value) or
+        ``None``.  Only a GeoDataFrame exposes a CRS; the
+        ``(geometry, value)`` iterable path always returns ``None``.
     """
     # Handle dask-geopandas by materializing eagerly.  Geometry data is
     # typically much smaller than the output raster, so this is fine.
@@ -2782,7 +2785,7 @@ def _parse_input(geometries, column=None, columns=None):
                     column = numeric_cols[0]
                 props_array = geometries[column].values.astype(
                     np.float64).reshape(-1, 1)
-            return geom_list, props_array, total_bounds
+            return geom_list, props_array, total_bounds, geometries.crs
     except ImportError:
         pass
 
@@ -2798,13 +2801,13 @@ def _parse_input(geometries, column=None, columns=None):
 
     if not geom_list:
         props_array = np.empty((0, 1), dtype=np.float64)
-        return geom_list, props_array, None
+        return geom_list, props_array, None, None
 
     props_array = np.array(value_list, dtype=np.float64).reshape(-1, 1)
 
     # Bounds computation is deferred: return None here and let the
     # caller compute bboxes only when bounds are actually needed.
-    return geom_list, props_array, None
+    return geom_list, props_array, None, None
 
 
 def _check_uniform_axis(axis_name, coords, expected_step):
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Guard rasterize() against geometry/like CRS mismatch (#3058)

None.

- `_like_crs` wraps the `rio.crs` lookup in a bare `except Exception` (rasterize.py:3010-3015). That matches the broad catch in `polygonize._detect_raster_crs`, so it is consistent with the codebase, but it also swallows a real rioxarray error (say a corrupt spatial_ref) and falls through to `None`, which disables the guard for that template. Since the resolution order checks `attrs['crs']`, `attrs['crs_wkt']`, and the `spatial_ref` coord first, this only bites templates that rely solely on `rio.crs`, so it is minor. A short comment on the intent would help; matching the existing helper is defensible either way.

- The mismatch error embeds `geom_crs!r`, which for a geopandas frame is a multi-line pyproj CRS repr, so the message runs long. `accessor._open_geotiff_windowed` does the same thing, so it is consistent, but a `to_epsg()`-style summary would read better. Not blocking.

### What looks good
- The guard runs before backend dispatch and before any host or device allocation, so it behaves the same across numpy, cupy, dask+numpy, and dask+cupy with no per-backend code.
- CRS resolution in `_like_crs` follows the same order as `polygonize._detect_raster_crs`, so a template round-trips the same CRS through rasterize and polygonize.
- Both sides go through `pyproj.CRS` before `.equals()`, so int EPSG, "EPSG:xxxx" strings, and WKT all compare correctly. Same precedent as `accessor.py`.
- The `check_crs=False` opt-out is documented and tested, and the no-op-when-either-side-is-None behavior keeps CRS-less inputs working unchanged.
- Tests cover the attrs, crs_wkt, and spatial_ref resolution paths, the bypass, int-vs-string equivalence, invalid-CRS surfacing, and the dask and cupy backends.

### Checklist
- [x] Algorithm matches reference/paper (N/A, validation guard)
- [x] All implemented backends produce consistent results (guard pre-dispatch)
- [x] Dask chunk boundaries handled correctly (guard runs before chunking)
- [x] Benchmark exists or is not needed (not needed, validation-only)
- [x] README feature matrix updated (N/A, no new function or backend change)

brendancol (commented):

Both findings from the first pass are resolved.

- Suggestion (bare except in `_like_crs`): added a comment at rasterize.py:3014 explaining the broad catch is deliberate. `.rio` raises AttributeError without rioxarray, and `rio.crs` can raise on a malformed georeference, so either way the template is treated as having no detectable CRS and the attrs/spatial_ref paths stay the source of truth. Behavior is unchanged on purpose; narrowing the type would break the no-rioxarray path.
- Nit (long pyproj repr in the error): the mismatch message now labels each side with a compact `EPSG:xxxx`, falling back to the CRS name when no EPSG code exists. A new test asserts the message is single-line and contains both codes.

Tests: 17 in the new file, all green; the full rasterize suite is still 231 passed, 2 skipped. flake8 clean.

No new issues found.

PR #3063, opened 2026-06-09
Rasterize LinearRing as a line; warn on dropped geometries

### LINKED ISSUE #3055: rasterize: non-empty LinearRing and other unsupported geometries are silently dropped
**Describe the bug**

`_classify_geometries` in `xrspatial/rasterize.py` sorts input geometries into polygon, line, and point buckets so `rasterize` can burn them into the output raster. Two gaps in the type classification cause non-empty geometries to be dropped with no warning.

1. LinearRing (Shapely type id 2) is left out. The fast-path `line_mask` matches only LineString (1) and MultiLineString (5), and the slow-path `_classify_one` matches only the `'LineString'`/`'MultiLineString'` names. A LinearRing is a closed line, so it should rasterize like one, but instead it falls through every bucket and vanishes.

2. Any other non-empty unsupported type vanishes the same way. The masks are inclusive: a geometry has to match a bucket to be kept, so anything non-empty that matches no bucket is dropped with no signal to the caller.

A user who passes a LinearRing gets back an all-fill raster with no error or warning, which looks like a successful no-op.

**Expected behavior**

- LinearRing should be treated as a line and rasterized.
- Any remaining non-empty geometry that matches no supported bucket should raise a warning naming the dropped type, so data never disappears silently.

**Additional context**

The comment at the type-id mapping already documents `2=LinearRing`, so the exclusion looks like an oversight, not a deliberate choice. Confirmed by reproduction: a non-empty LinearRing comes back from `_classify_geometries` in all three empty buckets with no warning.

brendancol:
Closes #3055

`_classify_geometries` sorts inputs into polygon / line / point buckets. The line bucket matched only LineString (1) and MultiLineString (5), so a non-empty LinearRing (shapely type id 2) fell through every bucket and disappeared. `rasterize` then returned an all-fill raster with no error or warning.

- Add type id 2 to the fast-path line bucket and the `'LinearRing'` name to the slow-path (GeometryCollection) line branch, so a LinearRing rasterizes as its closed boundary.
- Warn, naming the type, whenever a non-empty geometry matches no bucket, so data is never dropped silently. None / empty geometries stay silent (None reports as non-empty with type id -1, so it is now treated as empty to match the slow path's existing guard).
- Update the `rasterize` docstring to list LinearRing and the new drop-with-warning behavior.

`_classify_geometries` is the single classifier shared by every backend path, so the fix applies to numpy, cupy, dask+numpy, and dask+cupy.

### Test plan
- [x] LinearRing routes to the line bucket on both the fast path and the GeometryCollection slow path
- [x] End-to-end `rasterize` burns a LinearRing boundary and matches the equivalent closed LineString (numpy, dask, cupy, dask+cupy)
- [x] Non-empty unsupported type ids warn; None and empty geometries stay silent
- [x] Full existing rasterize suite passes (586 passed, 2 skipped); flake8 clean

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -4,7 +4,7 @@ Converts vector geometries (GeoDataFrame or list of (geometry, value) pairs)
 to a 2D xr.DataArray.  No GDAL dependency.
 
 - Polygons/MultiPolygons: scanline fill
-- Lines/MultiLineStrings: Bresenham line rasterization
+- Lines/LinearRings/MultiLineStrings: Bresenham line rasterization
 - Points/MultiPoints: direct pixel burn
 
 Supports numpy, cupy, dask+numpy, and dask+cupy backends.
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Rasterize LinearRing as a line; warn on dropped geometries

- None.

- None blocking. The `else` branch in `_classify_one` (slow path, `rasterize.py:366`) is effectively unreachable through the public API, since shapely won't build a GeometryCollection that holds an unsupported member. It is reasonable as defense-in-depth, and the helper itself is unit-tested, so I would keep it.

- `_warn_dropped_geometries` warns with the default `UserWarning`. That matches the rest of the module (no explicit warning category is used anywhere in `rasterize.py`), so no change is needed. Noting it only so the choice is visible.

### What looks good
- LinearRing now routes to the line bucket on both the vectorized fast path (`rasterize.py:308`) and the GeometryCollection slow path (`rasterize.py:363`), and the tests confirm a ring burns the same pixels as the equivalent closed LineString.
- The `None` case is handled correctly: `valid = ~empty & (type_ids >= 0)` (`rasterize.py:300`) drops type id -1, so a `None` geometry is skipped silently, matching the slow path's existing `geom is None` guard instead of warning on it.
- The drop warning fires once per batch with the unique offending types, not once per geometry, so it will not spam.
- Backend coverage is real: `_classify_geometries` is the single classifier shared by numpy, cupy, dask+numpy, and dask+cupy, and the tests exercise all four end to end.

### Checklist
- [x] Algorithm matches intended behavior (LinearRing = closed line)
- [x] NaN / None / empty handling is correct
- [x] Edge cases covered (None, empty, unsupported type id, GC slow path)
- [x] Dask chunk boundaries handled (shared classifier; ragged chunks tested)
- [x] Benchmark not needed (classification-only bug fix)

PR #3065, opened 2026-06-09
Add pack=True to to_geotiff: the inverse of mask_and_scale (#3064)

### LINKED ISSUE #3064: Add pack=True to to_geotiff as the inverse of open_geotiff(mask_and_scale=True)
## Reason or Problem
`open_geotiff(mask_and_scale=True)` reads an integer raster (say int8), applies the source's SCALE/OFFSET, masks the nodata sentinel to NaN, and promotes the array to float64. Getting back to the original integer encoding on write is awkward. `to_geotiff` has no `dtype` parameter, it does not reverse the scale/offset, and it re-embeds the source's SCALE/OFFSET tags. Reopening that file with `mask_and_scale=True` then applies the scale a second time (the round-trip caveat documented on `open_geotiff`). Today the user has to hand-roll the inverse: `(arr - offset) / scale`, `fillna`, `round`, `astype`, and strip the stale metadata tags.

## Proposal
Add a `pack=True` parameter to `to_geotiff` that reverses what `mask_and_scale=True` did and restores the original on-disk dtype, with no extra arguments. The name follows CF Conventions section 8.1 "Packed Data": `mask_and_scale` unpacks, `pack` re-packs.

**Design:**
- Read side records the pre-promotion source dtype in a new canonical attr `mask_and_scale_dtype` (attrs contract bumps from 4 to 5). This is set in the eager path (`_finalize_eager_read` in `_attrs.py`) and the dask path (`_backends/dask.py`) whenever masking and/or scaling promoted an integer array to float.
- A `_pack` transform reverses scale/offset (`(data - add_offset) / scale_factor`), fills NaN back to the nodata sentinel, rounds and casts to the recorded dtype, and strips the SCALE/OFFSET entries from `gdal_metadata` plus drops `gdal_metadata_xml` so the writer rebuilds clean tags. It runs as a single DataArray-level step at the top of `to_geotiff`, before the write-path branch, so it covers all backends.
- The writer already reverses the NaN-to-sentinel masking via `attrs['masked_nodata']`; `pack` adds the unscale, dtype restore, and tag cleanup on top.

**Usage:**
```python
da = open_geotiff("in_int8.tif", mask_and_scale=True)   # float64 in memory
da.xrs.to_geotiff("out.tif", pack=True)                 # written back as int8
```

**Value:** Makes the `mask_and_scale` read round-trip cleanly instead of leaving the user to reconstruct the integer encoding by hand or silently double-scaling on the next read.

## Drawbacks
Adds a parameter to an already-wide `to_geotiff` signature, and a new attrs-contract key that downstream consumers reading the contract version need to account for.

## Alternatives
- A standalone `da.xrs.pack()` accessor method instead of a `to_geotiff` flag.
- Following xarray's `.encoding` model rather than storing the state in `.attrs`. Rejected because this module deliberately keeps its metadata in `.attrs` under a versioned contract.

mask_and_scale is supported on the CPU eager and dask read paths only (GPU/VRT reject it), so the round-trip tests target numpy and dask.

### LINKED ISSUE #3066: Eager/dask dtype gap: bare mask_and_scale read of an int source with no scale/offset and no nodata
Surfaced while reviewing #3065 (the `pack=True` work).

`open_geotiff(int_source, mask_and_scale=True)` returns different dtypes on the eager and dask paths when the source has no SCALE/OFFSET metadata and no declared nodata sentinel:

- eager keeps the integer dtype (the `mask_and_scale` block only promotes when a real scale/offset applies, and there is no sentinel to mask)
- dask returns float64 (`_backends/dask.py:435` promotes any int source to float64 for `mask_and_scale` regardless of whether a scale/offset or sentinel exists)

Repro:

```python
data = np.array([[1, 2, 3], [4, 5, 6]], dtype=np.uint8)
to_geotiff(xr.DataArray(data, dims=("y", "x"),
                        coords={"y": [1.5, 0.5], "x": [0.5, 1.5, 2.5]},
                        attrs={"crs": 4326}), "plain.tif")
print(open_geotiff("plain.tif", mask_and_scale=True).dtype)            # uint8
print(open_geotiff("plain.tif", mask_and_scale=True, chunks=2).dtype)  # float64
```

This is a corner case (a `mask_and_scale` read of a file with nothing to scale or mask is close to a no-op), so #3065 left it alone and only kept its new `mask_and_scale_dtype` attr consistent across backends by keying off the real transform rather than the dask graph dtype. The underlying dtype divergence is still there.

Fix would be to make the dask path's `effective_dtype` promotion match eager: only promote when a scale/offset is actually present or a fittable sentinel exists. Needs care because the scale/offset values are not known until after the gdal_metadata is extracted later in the function, and some callers may rely on the float64 graph dtype.

brendancol:
Closes #3064

Adds `pack=True` to `to_geotiff` as the inverse of `open_geotiff(mask_and_scale=True)`. A `mask_and_scale` read promotes an integer raster to float64, applies the GDAL SCALE/OFFSET, and masks the nodata sentinel to NaN; `pack=True` reverses that so the file round-trips back to its integer encoding without the user reconstructing it by hand.

## What it does
- Read side records the integer source dtype on a new canonical attr `mask_and_scale_dtype` when `mask_and_scale` promoted the array to float (attrs contract v4 -> v5), on both the eager and dask paths.
- `to_geotiff(pack=True)` runs a `_pack` transform before any write-path dispatch: reverse `(data - add_offset) / scale_factor`, fill NaN back to the nodata sentinel, round, and cast to the recorded dtype.
- The SCALE/OFFSET GDAL_METADATA is kept, not dropped. Because the float values are un-scaled first, the packed integer file unpacks to the original values on the next `mask_and_scale` read instead of double-scaling. (My first pass dropped the tags, which silently changed the stored values' meaning; the test for the round-trip identity caught it.)

## Scope notes
- The dtype is recorded only for `mask_and_scale` reads, not plain `masked=True` reads, matching the attr name and the feature. Recording it on every masked read broke existing masked round-trip attr-stability tests, since writing a masked float array changes the on-disk dtype.
- Surfaced (but did not fix) a pre-existing eager/dask gap: a bare `mask_and_scale` read of an int source with no scale/offset and no sentinel returns int on eager but float64 on dask. The attr recording keys off the real transform, not the dask graph's promoted dtype, so the new attr stays consistent across backends.
- `pack` is a no-op on the GPU writer (the transform runs before dispatch); it exists there only for signature parity.

## Backends
numpy and dask. `mask_and_scale` is a CPU eager + dask read feature (GPU/VRT reject it), so the round-trip is exercised on those two.

## Test plan
- [x] int8 and uint16 round trips (numpy + dask)
- [x] SCALE/OFFSET round trip: raw reopen matches the source, and a `mask_and_scale` reopen reproduces the original decode (no double-scale)
- [x] eager/dask `mask_and_scale_dtype` parity
- [x] guards: bare array and no-mask_and_scale-state arrays raise
- [x] contract-version pin and docstring gates bumped to v5
- [x] full geotiff suite green (6247 passed); flake8 + isort clean

Also closes #3066 (eager/dask mask_and_scale dtype parity), surfaced during this PR's review: the dask reader no longer over-promotes a bare mask_and_scale read of an int source with no scale/offset and no sentinel to float64.

### DIFF
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -114,6 +114,12 @@ write.
        float-because-masked from float-because-promoted, which a
        ``masked_nodata`` lookup alone cannot disambiguate. See issue
        #2135.
+   * - ``mask_and_scale_dtype``
+     - str
+     - Integer source dtype name (e.g. ``"int8"``), set only when a
+       ``mask_and_scale=True`` read promoted an integer array to float.
+       ``to_geotiff(pack=True)`` reads it to reverse the promotion and
+       restore the on-disk dtype. Added in contract v5 (issue #3064).
    * - ``raster_type``
      - str
      - ``'point'`` when the file declares ``RasterPixelIsPoint``;
@@ -142,7 +148,7 @@ write.
        ``ResolutionUnit`` ids 1, 2, 3).
    * - ``_xrspatial_geotiff_contract``
      - int
-     - Contract version. Currently ``4``. See `Versioning`_.
+     - Contract version. Currently ``5``. See `Versioning`_.
    * - ``_xrspatial_no_georef``
      - bool
      - Stamped ``True`` on reads of files with no GeoTIFF transform
@@ -376,3 +382,12 @@ canonical tier. The attr surfaces the rotated 6-tuple from
 so callers can recover the rotated mapping. The writer drops it on
 round-trip until ``to_geotiff`` learns to emit
 ``ModelTransformationTag`` (issue #2115 follow-up).
+
+Contract v5 (issue #3064) added the ``mask_and_scale_dtype`` attr to
+the canonical tier. The attr records the integer source dtype when a
+``mask_and_scale=True`` read promoted the array to float, so
+``to_geotiff(pack=True)`` can reverse the scale / offset, fill NaN back
+to the nodata sentinel, and restore the on-disk dtype. The packed file
+keeps its ``SCALE`` / ``OFFSET`` tags, so reopening it with
+``mask_and_scale=True`` unpacks to the same values instead of scaling a
+second time.
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -25,7 +25,7 @@ is present, and pass-through keys are kept when the writer can
 reconstruct them from canonical state.
 
 The contract version is recorded in ``attrs['_xrspatial_geotiff_contract']``
-(currently ``4``). Consumers can branch on this integer if the tier
+(currently ``5``). Consumers can branch on this integer if the tier
 split changes in a future release.
 
 Canonical (xrspatial owns these; round-trip stable):
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -76,6 +76,11 @@ Canonical (xrspatial owns these; round-trip stable):
   passed an explicit ``dtype=`` kwarg. Records that a post-mask cast
   happened so consumers can tell float-because-masked from
   float-because-promoted.
+- ``mask_and_scale_dtype`` (#3064, contract v5): string dtype name of
+  the integer source (e.g. ``"int8"``), only emitted when masking
+  and / or ``mask_and_scale`` promoted an integer array to float on
+  read. ``to_geotiff(pack=True)`` reads it to reverse the promotion and
+  restore the on-disk dtype.
 - ``raster_type``: ``'area'`` (implicit / RasterPixelIsArea) or ``'point'``
   (explicit / RasterPixelIsPoint).
 - ``georef_status``: one of ``'full'``, ``'transform_only'``, ``'crs_only'``,
@@ -547,7 +552,12 @@ _TIFF_SHORT = 3
 # writer drops it on round-trip until ``to_geotiff`` grows a
 # ``ModelTransformationTag`` emit path. Existing keys keep their pre-v4
 # shape.
-_ATTRS_CONTRACT_VERSION = 4
+#
+# Version 5 adds ``attrs['mask_and_scale_dtype']``: the integer source
+# dtype recorded when ``mask_nodata`` / ``mask_and_scale`` promoted the
+# array to float on read. ``to_geotiff(pack=True)`` reads it to restore
+# the on-disk dtype. Existing keys keep their pre-v5 shape.
+_ATTRS_CONTRACT_VERSION = 5
 
 
 # Canonical ``attrs['georef_status']`` values. One attr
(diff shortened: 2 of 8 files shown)


brendancol (commented):
## PR Review: Add pack=True to to_geotiff

### Blockers
None.

### Suggestions
- Pre-existing eager/dask dtype gap, surfaced but not fixed (`_backends/dask.py:435`): a bare `mask_and_scale=True` read of an int source with no scale/offset and no sentinel returns int on eager but float64 on dask. The PR documents it in a comment and works around it for the new attr, which is the right call for scope, but it is worth a follow-up issue so it does not get lost.

### Nits
- Stale comment in `_writers/eager.py` (the `pack` block): the inline comment still says "drop the SCALE / OFFSET tags so the written file round-trips cleanly." The code keeps the tags (both the docstring and `_pack`'s docstring say so). Leftover from an earlier approach; the comment now contradicts the implementation and should be corrected so it does not mislead the next reader.
- `_pack` (`_attrs.py`): the pre-v5 fallback `np.asarray(nodata).dtype.name` can resolve to a wider integer than the true source (a Python-int sentinel gives int64), so packing an array read before this PR could land a wider dtype than the file had. Current reads always carry `mask_and_scale_dtype`, so this only affects old reads, and the docstring already calls it best-effort. Flagging for awareness.
- `out.round()` uses numpy's round-half-to-even. For clean packed round trips the unscaled values are exact integers, so this never bites; a non-identity scale on hand-built data could round a true .5 down. Not worth changing.

### What looks good
- The keep-the-tags decision is correct and the round-trip identity test proves it: a `mask_and_scale` reopen of the packed file reproduces the original decode rather than double-scaling. The PR body is honest that the first pass got this wrong.
- Read-side recording mirrors between eager and dask and is scoped to real transforms, so the new attr stays consistent across backends even on top of the pre-existing dtype gap.
- The transform runs once at the top of `to_geotiff` before any dispatch, so it covers numpy, dask, COG, and GPU without per-writer duplication, and stays lazy on dask (the only eager compute is the error-path NaN check, skipped when a sentinel exists).
- Guards and contract-gate updates are present; full geotiff suite passes.

### Checklist
- [x] Algorithm matches the inverse of mask_and_scale
- [x] Eager and dask record the attr consistently
- [x] NaN handling correct (filled before integer cast)
- [x] Edge cases covered (no-scale, scale/offset, guards)
- [x] Dask stays lazy on the happy path
- [x] Benchmark not needed (thin transform)
- [x] README not applicable (parameter, not new function)
- [x] Docstrings present and accurate (modulo the one stale inline comment)

brendancol (commented):
## Follow-up review (after review fixes)

- **Fixed** -- stale comment in `_writers/eager.py` that claimed `pack` drops the SCALE/OFFSET tags. Corrected in b205fbc to match the code and `_pack`'s docstring (the tags are kept; un-scaling first makes that correct).
- **Deferred** -- pre-existing eager/dask dtype gap for a bare `mask_and_scale` read of an int source with no scale/offset and no sentinel. Filed as #3066. Out of scope here: the fix changes dask read-dtype behavior and needs its own design pass.
- **Dismissed (with reason)** -- the pre-v5 `nodata`-width fallback in `_pack` and the round-half-to-even note. Both only touch pre-v5 or hand-built edge cases, are documented as best-effort, and never bite a clean round trip. No change warranted.

No blockers. `test_pack_3064.py` (11 tests) and flake8 still pass after the comment fix.

brendancol (commented):
## Update: also closes #3066

Pulled the deferred follow-up into this PR per request. The dask reader now extracts the scale/offset up front and promotes an integer source to float64 only for a real `mask_and_scale` transform (non-identity scale/offset or a fittable sentinel), matching the eager path. A bare `mask_and_scale` read of a plain int source with no scale/offset and no sentinel now stays integer on both backends.

- `_backends/dask.py`: gate the float64 promotion on the real transform; reuse the up-front scale/offset when applying the scaling lazily.
- New `tests/read/test_mask_and_scale_dtype_parity_3066.py`: eager/dask dtype + value parity across bare / masked / scaled / masked+scaled.
- Full geotiff suite: 6251 passed.

PR #3068, opened 2026-06-09
Pick reproject resampling by raster type in open_geotiff(auto_reproject=True) (#3067)

### LINKED ISSUE #3067: open_geotiff(auto_reproject=True) should pick reproject resampling from raster type
## Reason or Problem

When `myarr.xrs.open_geotiff(source, auto_reproject=True)` reads a GeoTIFF whose CRS differs from the caller's, it reprojects the windowed read back to the caller's CRS in `_open_geotiff_windowed` (`xrspatial/accessor.py:242-246`). That call passes no `resampling=` argument, so `reproject()` falls back to its `'bilinear'` default.

Bilinear averages neighbouring pixels. That's right for continuous rasters like elevation, but wrong for categorical rasters such as land cover: averaging class IDs 5 and 9 into 7 invents a class that doesn't exist. The caller has no way to influence the mode through `open_geotiff` today.

## Proposal

Add a `resampling` parameter to both `open_geotiff` accessor methods (DataArray and Dataset), defaulting to `'auto'`.

**Design**
- `'auto'` resolves to `'nearest'` when the read result looks categorical and `'bilinear'` otherwise. A raster counts as categorical when it carries a paletted colormap (`attrs['colormap']`, from TIFF tag 320) or has an integer dtype.
- Explicit `'nearest'`, `'bilinear'`, or `'cubic'` pass straight through to `reproject()`.
- Resolution happens in the accessor layer, so only concrete modes reach `reproject()`. `reproject()` itself is unchanged.
- The parameter is validated early, so a bad value fails before any read.

**Usage**
```python
template.xrs.open_geotiff(path, auto_reproject=True)                        # nearest for land cover, bilinear for elevation
template.xrs.open_geotiff(path, auto_reproject=True, resampling='bilinear') # force bilinear
```

**Value**
Categorical rasters survive auto-reprojection intact instead of being silently averaged into nonexistent classes, continuous rasters keep bilinear quality, and callers can override either way.

## Drawbacks

The integer-dtype signal treats integer-typed continuous DEMs (such as int16 SRTM) as categorical and resamples them with nearest. Those callers need to pass `resampling='bilinear'` explicitly. This caveat will be documented on the parameter.

## Alternatives

- Keep the bilinear default and require callers to always pass `resampling` explicitly.
- Use colormap presence alone as the categorical signal, which would miss palette-less integer class rasters.

## Additional Notes

`resampling` only affects the `auto_reproject` reproject step. It is ignored when no CRS reprojection happens.

brendancol:
Closes #3067

## What

`open_geotiff(auto_reproject=True)` reprojected the windowed read back to the caller's CRS with no `resampling=` argument, so `reproject()` always used its `'bilinear'` default. Bilinear is fine for continuous data but averages class IDs in categorical rasters, producing classes that do not exist.

This adds a `resampling` parameter to both `open_geotiff` accessor methods (DataArray and Dataset), defaulting to `'auto'`:

- `'auto'` picks `'nearest'` for categorical rasters (a paletted colormap is present, or the data has an integer dtype) and `'bilinear'` otherwise.
- Explicit `'nearest'`, `'bilinear'`, or `'cubic'` pass straight through to `reproject()`.
- Resolution happens in the accessor layer (`_resolve_resampling`), so only concrete modes reach `reproject()`. `reproject()` itself is untouched.
- The value is validated early, so a bad mode fails before any read.

Documented caveat: an integer-typed continuous DEM (such as int16 elevation) is treated as categorical under `'auto'`. Pass `resampling='bilinear'` for those.

## Backend coverage

The change sits in the accessor dispatch layer and only chooses the string handed to `reproject()`, which already supports numpy, cupy, dask+numpy, and dask+cupy. `result.dtype` and `result.attrs` read the same way across all four.

## Test plan

- [x] `_resolve_resampling` unit tests: float to bilinear, integer to nearest, colormap to nearest, explicit modes unchanged
- [x] End-to-end spy: auto float forwards bilinear, auto integer forwards nearest, explicit override forwarded
- [x] Invalid mode raises ValueError before any read
- [x] No reproject when CRS matches (resampling ignored)
- [x] Existing open_geotiff / auto_reproject tests still pass

### DIFF
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -146,7 +146,29 @@ def _bbox_edge_samples(x_min, y_min, x_max, y_max, n_per_side=20):
     return xs, ys
 
 
-def _open_geotiff_windowed(obj, source, *, auto_reproject=False, **kwargs):
+_RESAMPLING_CHOICES = ('auto', 'nearest', 'bilinear', 'cubic')
+
+
+def _resolve_resampling(resampling, result):
+    """Resolve an ``'auto'`` resampling choice for the reproject step.
+
+    ``'auto'`` picks ``'nearest'`` for categorical rasters (a paletted
+    colormap is present, or the data has an integer dtype) and
+    ``'bilinear'`` otherwise. Any explicit mode is returned unchanged
+    for :func:`xrspatial.reproject.reproject` to validate.
+    """
+    if resampling != 'auto':
+        return resampling
+    import numpy as np
+    categorical = (
+        result.attrs.get('colormap') is not None
+        or np.issubdtype(result.dtype, np.integer)
+    )
+    return 'nearest' if categorical else 'bilinear'
+
+
+def _open_geotiff_windowed(obj, source, *, auto_reproject=False,
+                           resampling='auto', **kwargs):
     """Shared implementation for ``.xrs.open_geotiff`` on DataArray and
     Dataset.
 
@@ -169,6 +191,12 @@ def _open_geotiff_windowed(obj, source, *, auto_reproject=False, **kwargs):
     """
     from .geotiff import open_geotiff, _read_geo_info, _extent_to_window
     from .utils import _classify_backend
+
+    if resampling not in _RESAMPLING_CHOICES:
+        raise ValueError(
+            f"resampling must be one of {list(_RESAMPLING_CHOICES)}, "
+            f"got {resampling!r}"
+        )
 
     if 'y' not in obj.coords or 'x' not in obj.coords:
         raise ValueError(
@@ -243,6 +271,7 @@ def _open_geotiff_windowed(obj, source, *, auto_reproject=False, **kwargs):
             result,
             target_crs=caller_crs,
             source_crs=file_crs,
+            resampling=_resolve_resampling(resampling, result),
         )
 
     return result
@@ -783,7 +812,8 @@ class XrsSpatialDataArrayAccessor:
         from .geotiff import to_geotiff
         return to_geotiff(self._obj, path, **kwargs)
 
-    def open_geotiff(self, source, *, auto_reproject=False, **kwargs):
+    def open_geotiff(self, source, *, auto_reproject=False,
+                     resampling='auto', **kwargs):
         """Read a GeoTIFF windowed to this DataArray's spatial extent.
 
         Uses ``self``'s ``y``/``x`` coordinates to compute a pixel window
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: pick reproject resampling by raster type

Small, well-scoped change. Resolution lives in the accessor layer, `reproject()` is untouched, and the tests cover the unit helper, end-to-end forwarding, validation, and the no-op path. No blockers.

### Suggestions

- **Nodata masking can defeat the integer-dtype signal.** When an integer raster carries a nodata sentinel, the reader promotes it to `float64` during masking (`xrspatial/geotiff/_backends/dask.py:424`, `xrspatial/geotiff/_backends/_gpu_helpers.py:45`). So a categorical integer raster that has a nodata value and no colormap reads back as `float64`, and `_resolve_resampling` (`xrspatial/accessor.py`) picks `'bilinear'` rather than `'nearest'`. This is not a regression (that case already got bilinear before), and the colormap branch still catches paletted rasters, but it's a real gap in the `'auto'` heuristic. Worth one sentence in the docstring caveat so users with nodata-carrying class rasters know to pass `resampling='nearest'` explicitly.

### Nits

- The forwarding tests assert the resampling string handed to `reproject()` via the spy, which is the right unit to check. If you want one belt-and-suspenders case, a single test that reprojects a tiny categorical raster for real and asserts the output holds only original class IDs (no averaged values) would lock in the end-to-end behaviour. Optional; the spy tests already prove the wiring.

### What looks good

- `_resolve_resampling` reads `result.dtype` / `result.attrs` lazily, so no dask materialization.
- Early validation against the allowed set fails before any file read.
- The int16-DEM caveat is already documented on both accessor methods.
- Backend-agnostic: only the mode string changes, and `reproject()` already covers all four backends.

brendancol (commented):
## Follow-up review (after cf663c07)

- **Suggestion (nodata-promotion gap):** fixed. Both accessor docstrings now spell out that an integer raster read with a nodata sentinel is promoted to float, so a categorical raster carrying nodata and no colormap resolves to `'bilinear'` under `'auto'`, and the user should pass `resampling='nearest'`.
- **Nit (real end-to-end check):** added. `test_categorical_real_reproject_preserves_class_ids` reprojects an int16 class raster across a CRS mismatch under `'auto'` and asserts every finite output value is one of the original class IDs, so nearest is genuinely in effect rather than just the right string being forwarded.

Full suite for the file is green (12 passed) and flake8 is clean. No remaining findings.

PR #3070, opened 2026-06-09
Add coregister=True to open_geotiff: mask_and_scale + reproject + match grid (#3069)

### LINKED ISSUE #3069: Add coregister=True to .xrs.open_geotiff: mask_and_scale + reproject + match grid
## Reason or Problem

Reading a GeoTIFF against an existing template array (`myarr`) usually means you want the result lined up with `myarr`: nodata masked and scale/offset applied, reprojected into `myarr`'s CRS, and resampled onto `myarr`'s exact grid (same y/x coordinates, same shape, same cell centers). Today that is three manual steps. `open_geotiff(auto_reproject=True)` only does the middle one: it reprojects to the caller's CRS but keeps the file's native resolution, so the result does not line up cell-for-cell with `myarr`.

## Proposal

Add a `coregister=False` flag to both `open_geotiff` accessor methods (DataArray and Dataset). When `coregister=True`, the read is masked and scaled, reprojected into the caller's CRS, and resampled onto the caller's exact grid in one call (rioxarray's `reproject_match` plus `mask_and_scale`).

```python
out = myarr.xrs.open_geotiff("test.tif", coregister=True)
# out.coords['x'] == myarr.coords['x'], out.coords['y'] == myarr.coords['y']
# masked + scaled, in myarr's CRS, ready to stack or compute with myarr
```

**Design**
- Resolution lives in the accessor layer. It composes the existing `open_geotiff(mask_and_scale=True)` read with `reproject(bounds=..., width=..., height=...)`, which already emits pixel-center coordinates, so passing the caller's edge bounds and shape reconstructs the caller's exact grid.
- coregister always snaps to the caller's grid, even when the CRS already matches (a pure resample for differing resolution or alignment).
- The resample mode reuses the `resampling='auto'` parameter (categorical to nearest, continuous to bilinear), with explicit override.
- `mask_and_scale` is unsupported on `gpu=True` and `.vrt` sources, so coregister raises a clear `ValueError` there rather than silently dropping the scaling.

**Relationship to existing params**
`coregister` is a new flag. `auto_reproject` stays as the lighter CRS-only option that keeps native resolution.

**Value**
One call produces a raster that is grid-aligned with the template, so downstream stacking and cell-wise math work without a manual regrid.

## Drawbacks

No GPU path, because the `mask_and_scale` step does not support GPU reads. Single-cell (1x1) templates cannot derive a resolution from cell centers and need a fallback to the template's transform or a clear error.

## Alternatives

- Expose a standalone `reproject_match`-style function instead of a flag on `open_geotiff`.
- Keep the three steps manual.

## Additional Notes

Depends on the `resampling='auto'` work in #3067 (PR #3068) for the `_resolve_resampling` helper.

brendancol:
Closes #3069

Stacked on #3068 (issue #3067). Review and merge that first. This branch is based on `issue-3067`, so its diff includes #3068's commits until that merges.

## What

`auto_reproject=True` reprojects a GeoTIFF into the caller's CRS but keeps the file's native resolution, so the result does not line up cell-for-cell with the template array. This adds `coregister=True` to both `open_geotiff` accessor methods, producing a result on the template's exact grid in one call:

- Reads masked and scaled (`mask_and_scale=True`): nodata to NaN, integer promotion, GDAL scale/offset.
- Reprojects into the template's CRS.
- Resamples onto the template's exact grid (same y/x coordinates, same shape), even when the CRS already matches.

```python
out = myarr.xrs.open_geotiff("test.tif", coregister=True)
# out.coords == myarr.coords; masked, scaled, in myarr's CRS
```

## Design

- All in the accessor layer. `_caller_grid` turns the template's center coordinates into the edge bounds + shape that make `reproject(bounds=..., width=..., height=...)` reproduce those exact centers (reproject emits pixel-center coords).
- Reuses `resampling='auto'` from #3068 for the resample mode.
- A masked read leaves `attrs['nodata']` as the raw sentinel while the holes are NaN, so `reproject` is now passed `nodata=nan` when `masked_nodata` is set; otherwise it would resample the holes as ordinary values. This also applies to the existing auto_reproject path and is a no-op when nothing was masked.
- `mask_and_scale` is unsupported on `gpu=True` and `.vrt` sources, so `coregister=True` raises a clear ValueError there.
- Single-cell templates derive resolution from `attrs['res']` / `attrs['transform']`, else raise.

## Backend coverage

numpy and dask templates covered. cupy / dask+cupy are rejected by the gpu guard (mask_and_scale has no GPU path).

## Test plan

- [x] Exact grid match on CRS mismatch and on matching CRS
- [x] dask template coregisters onto the exact grid
- [x] mask_and_scale forwarded to the read (spy); nodata masked to NaN end-to-end
- [x] categorical raster keeps original class IDs (nearest); explicit resampling overrides
- [x] gpu=True and .vrt sources raise ValueError
- [x] _caller_grid reconstructs centers; single-cell needs a resolution
- [x] existing open_geotiff / resampling tests still pass

### DIFF
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -197,6 +197,11 @@ def _caller_grid(obj):
     emits pixel-centre coords ``linspace(left+res/2, right-res/2, W)``, so
     bounds offset half a pixel outside the caller's centre extent map the
     output cells back onto the caller's coordinates.
+
+    Assumes a regular grid with ``x`` ascending and ``y`` descending (the
+    convention ``_open_geotiff_windowed`` already works in); resolution is
+    the centre spacing. An irregular or descending-``x`` template would
+    not line up cell-for-cell.
     """
     y = obj.coords['y'].values
     x = obj.coords['x'].values
@@ -337,11 +342,15 @@ def _open_geotiff_windowed(obj, source, *, auto_reproject=False,
                 "coregister=True needs a CRS: set attrs['crs'] on the "
                 "caller or read a file that declares one"
             )
+        # A file that declares no CRS is treated as already in the
+        # caller's CRS, so coregister stays a pure resample instead of
+        # failing reproject's source-CRS detection.
+        source_crs = file_crs if file_crs is not None else target_crs
         bounds, width_out, height_out = _caller_grid(obj)
         result = reproject(
             result,
             target_crs=target_crs,
-            source_crs=file_crs,
+            source_crs=source_crs,
             bounds=bounds,
             width=width_out,
             height=height_out,
@@ -924,7 +933,8 @@ class XrsSpatialDataArrayAccessor:
             ``self``'s ``y``/``x`` coordinates and shape. The grid snap
             happens even when the CRS already matches. ``mask_and_scale``
             is unsupported with ``gpu=True`` or ``.vrt`` sources, so
-            ``coregister=True`` raises ``ValueError`` there. The resample
+            ``coregister=True`` raises ``ValueError`` there, and it
+            overrides an explicit ``mask_and_scale=False``. The resample
             mode follows ``resampling``. This is the heavier counterpart
             of ``auto_reproject``, which keeps the file's native
             resolution.
@@ -1453,7 +1463,8 @@ class XrsSpatialDatasetAccessor:
             grid snap happens even when the CRS already matches.
             ``mask_and_scale`` is unsupported with ``gpu=True`` or
             ``.vrt`` sources, so ``coregister=True`` raises ``ValueError``
-            there. The resample mode follows ``resampling``.
+            there, and it overrides an explicit ``mask_and_scale=False``.
+            The resample mode follows ``resampling``.
         resampling : {'auto', 'nearest', 'bilinear', 'cubic'}
             Resampling mode for the ``auto_reproject`` / ``coregister``
             reproject step; ignored when no reprojection happens.
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -167,8 +167,51 @@ def _resolve_resampling(resampling, result):
     return 'nearest' if categorical else 'bilinear'
 
 
+def _axis_res(coord, n, obj, axis):
+    """Resolution (cell size) for one axis of the caller's grid.
+
+    With 2+ cells it is the centre spacing. A single-cell axis cannot
+    give a spacing, so fall back to the caller's ``attrs['res']`` or
+    ``attrs['transform']`` pixel size, and raise if neither is present.
+    """
+    if n >= 2:
+        return abs(float(coord[-1]) - float(coord[0])) / (n - 1)
+    res = obj.attrs.get('res')
+    if res is not None:
+        return abs(float(res[0] if axis == 'x' else res[1]))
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: coregister=True for open_geotiff

The grid math is exact (verified: reproject's pixel-center linspace over the half-pixel-expanded bounds reproduces the template centers), the masked-nodata → NaN fix is correct and also helps the existing auto_reproject path, and the gpu/.vrt guard and tests are solid. No blockers.

### Suggestions

- **CRS-less file + georeferenced template fails in reproject.** In the coregister branch (`xrspatial/accessor.py`), `source_crs` is always `file_crs`. When the file declares no CRS (`georef_status` `'transform_only'`) but the template has one, `target_crs` falls back to `caller_crs` while `source_crs` stays `None`, and `reproject` then raises "Could not detect source CRS." For coregister semantics (snap to the template grid), treating a CRS-less file as already in the template's CRS is the sensible default: pass `source_crs = file_crs if file_crs is not None else target_crs`. A short test with a transform-only file would lock it in.

### Nits

- `_caller_grid` (`xrspatial/accessor.py:192`) assumes a regular, x-ascending / y-descending template: it derives resolution from `(coord[-1]-coord[0])/(n-1)` and bounds from `min`/`max`, while `reproject` always emits x-ascending / y-descending coords. A descending-x or irregularly spaced template would not match cell-for-cell. One docstring line noting the assumption would set expectations (it matches the convention the rest of `_open_geotiff_windowed` already assumes).
- The `.vrt` guard uses `str(source).lower().endswith('.vrt')`, so a VRT handed in as an open file object slips past. Path inputs are the norm, so this is minor.
- coregister forces `mask_and_scale=True`, silently overriding an explicit `mask_and_scale=False`. It's noted in a code comment; a word in the docstring (or raising on the conflict) would make it visible to callers.

### What looks good

- Grid reconstruction is exact and reuses `_resolve_resampling` from #3068.
- The `nodata=nan`-when-`masked_nodata` fix is the right call and quietly corrects the auto_reproject path too.
- Clear, early gpu/.vrt `ValueError`; single-cell templates fall back to `attrs['res']`/`transform` or raise.
- Tests cover CRS match + mismatch, dask, categorical class-id preservation, mask_and_scale forwarding, nodata masking, and the guards.

brendancol (commented):
## Follow-up review (after 50bf5e2a)

Dispositions:

- **Suggestion (CRS-less file fails source-CRS detection):** fixed. The coregister branch now uses `source_crs = file_crs if file_crs is not None else target_crs`, so a transform-only file is treated as already in the template's CRS and resamples onto the grid. Added `test_coregister_crs_less_file`.
- **Nit (`_caller_grid` grid assumptions):** fixed. Docstring now states the regular, x-ascending / y-descending assumption.
- **Nit (`mask_and_scale=False` override):** fixed. Both accessor docstrings now note that `coregister=True` overrides an explicit `mask_and_scale=False`.
- **Nit (`.vrt` guard misses file-object inputs):** dismissed. coregister/VRT inputs are path-based in practice, and `str(file_obj)` has no reliable extension to key on; not worth special-casing.

Tests green (coregister + resampling + open_geotiff integration, 42 passed), flake8 clean. No remaining actionable findings.

PR #3074, opened 2026-06-09
Add dask+numpy nodata-masking value-parity tests for resample

### LINKED ISSUE #3073: resample: add dask+numpy nodata-masking value-parity tests
## Coverage gap

`resample` masks nodata sentinels before resampling in `_apply_nodata_mask`, which goes through xarray's `.where()` / `.astype()`. Those dispatch per backend, but the dask+numpy masking path has no value-level test. The existing dask+numpy nodata tests only check the error paths (`TestNodataOutOfRange`) and the identity-attr refresh (`TestIdentityNodataMetadata`). Nothing checks that sentinel cells actually become NaN, with values matching the numpy backend, after a real (non-identity) resample.

## Proposed tests

Add `xrspatial/tests/test_resample_nodata_dask_parity_3073.py`. For the dask+numpy backend it asserts that resample output matches the numpy backend (equal_nan=True) and that the expected masked cell is NaN. Parametrized over method (`nearest`, `average`) and sentinel source: explicit `nodata=` arg, `_FillValue` attr, an integer `-9999` sentinel, and a float `-1.0` sentinel. Guarded with the existing `dask_array_available` marker.

## Scope

Test-only. No source changes.

## Note on cupy

No cupy / dask+cupy nodata tests here. On this environment (xarray 2025.12 + cupy 13.6), `DataArray.where()` on a cupy backend raises `AttributeError: module 'cupy' has no attribute 'astype'`. That comes from an xarray/cupy Array-API mismatch hitting every `.where` call on a cupy array (~221 sites across xrspatial), not from resample, so it's out of scope for this coverage closure.

brendancol:
Closes #3073

## What this does

- Adds `xrspatial/tests/test_resample_nodata_dask_parity_3073.py` covering the dask+numpy nodata-masking value path, which had no value-level test (existing dask+numpy nodata tests only hit the error paths and identity-attr refresh).
- Each test asserts the dask+numpy output matches the numpy backend (equal_nan=True) and that the all-sentinel block becomes NaN, across method (`nearest`, `average`) and sentinel source (explicit `nodata=` arg, `_FillValue` attr, `nodata` attr; integer `-9999` and float `-1.0`).

## Backend coverage

numpy and dask+numpy. cupy / dask+cupy nodata tests are left out on purpose: on xarray 2025.12 + cupy 13.6, `DataArray.where()` on a cupy array raises `AttributeError: module 'cupy' has no attribute 'astype'`, an xarray/cupy Array-API mismatch (~221 `.where` sites across xrspatial), not a resample bug.

## Scope

Test-only. `resample.py` is unchanged.

## Test plan

- [x] 10 new tests pass locally (CUDA host)
- [x] Full `test_resample.py` + coverage suite still green (309 passed)

### DIFF
--- a/xrspatial/tests/test_resample_nodata_dask_parity_3073.py
+++ b/xrspatial/tests/test_resample_nodata_dask_parity_3073.py
@@ -1,0 +1,112 @@
+"""dask+numpy nodata-masking value parity for xrspatial.resample (#3073).
+
+`resample` replaces nodata sentinels with NaN before resampling via
+`_apply_nodata_mask`, which dispatches through xarray's `.where()` /
+`.astype()` per backend. The existing dask+numpy nodata tests only cover
+the error-raising paths (`TestNodataOutOfRange`) and the identity-attr
+refresh (`TestIdentityNodataMetadata`); none assert that sentinel cells
+become NaN with values matching the numpy backend after a real
+(non-identity) resample. These tests close that gap.
+
+cupy / dask+cupy are intentionally not covered: on xarray 2025.12 +
+cupy 13.6, `DataArray.where()` on a cupy backend raises
+`AttributeError: module 'cupy' has no attribute 'astype'` (an
+xarray/cupy Array-API mismatch affecting every `.where` on a cupy array,
+not a resample bug). Source untouched -- this is a test-only closure.
+"""
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+from xrspatial.resample import resample
+from xrspatial.tests.general_checks import (
+    create_test_raster,
+    dask_array_available,
+)
+
+
+def _to_numpy(arr):
+    data = arr.data
+    if hasattr(data, 'compute'):
+        data = data.compute()
+    return np.asarray(data)
+
+
+# A 4x4 grid whose top-left 2x2 block is the sentinel; downsampling by 0.5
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: Add dask+numpy nodata-masking value-parity tests for resample

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The `isnan(out[0,0])` assertions carry their weight. Without masking the top-left 2x2 sentinel block would average or select to a finite -9999 / -1.0, so the test fails loudly if masking ever silently becomes a no-op.
- Branch coverage of `_apply_nodata_mask` is good: the float -1.0 sentinel takes the float-input path (no cast), the int -9999 sentinel takes the integer->float32 cast path.
- Parity is checked against the numpy backend with equal_nan=True, which is exactly the gap this closes.
- The cupy exclusion is written into the module docstring with the concrete xarray/cupy error, so a future reader on a fixed environment knows what to re-enable.
- Filename carries the issue token 3073, so no temp/file collision with parallel runs.

### Checklist
- Backends: numpy + dask+numpy covered; cupy/dask+cupy excluded with a documented reason.
- NaN handling: masked cells assert NaN; parity uses equal_nan.
- Edge cases: all-sentinel block (all-NaN window) is exercised on the average path.
- Source untouched; test-only.
- No benchmark or README change needed.

PR #3075, opened 2026-06-09
Rename open_geotiff mask_and_scale to unpack and add GPU / dask+GPU support

### LINKED ISSUE #3071: Rename open_geotiff mask_and_scale to unpack and add GPU / dask+GPU support
## Summary

`open_geotiff` has a `mask_and_scale` parameter (named after rioxarray's
`open_rasterio`) that reads a source's GDAL SCALE/OFFSET metadata and returns
`data * scale + offset`, masking the nodata sentinel to NaN. Two things are off:

1. The name describes the implementation, not what the caller wants. The
   operation unpacks CF-packed data: integers stored with a scale/offset that
   recover floats. `unpack` says that, and it pairs with the existing `pack`
   write option.
2. It only runs on the CPU eager and dask paths. `gpu=True` and dask+GPU
   (`gpu=True, chunks=N`) raise `ValueError`, so the option is missing from the
   GPU pipelines the rest of the module already supports.

## Proposed change

1. Add an `unpack: bool = False` parameter. Keep `mask_and_scale` as a
   deprecated alias: passing it emits a `DeprecationWarning` and forwards to
   `unpack`, and passing both raises `TypeError`. Same sentinel pattern already
   used for `mask_nodata` -> `masked`. Behaviour does not change: `unpack=True`
   does what `mask_and_scale=True` did (scale/offset plus nodata masking).

2. Implement `unpack` on the GPU eager and dask+GPU backends. The scale/offset
   arithmetic already runs on CuPy arrays via numpy duck-typing in the shared
   eager finalizer, so the work is threading the flag through the GPU eager
   sites and the dask+GPU chunked path, then dropping the `gpu=True` rejection.

## Scope

- VRT sources stay gated and still raise. Out of scope here.
- `parse_coordinates=False` GPU gating is left as-is. Out of scope.
- The internal `attrs['mask_and_scale_dtype']` round-trip key keeps its name; it
  pairs with the writer's pack path. Only the public parameter is renamed.

## Backends

numpy and dask+numpy already work. This adds cupy and dask+cupy.

brendancol:
Closes #3071.

Renames `open_geotiff`'s `mask_and_scale` read option to `unpack` and
extends it to the GPU and dask+GPU backends.

## What changed

- New `unpack: bool = False` parameter on `open_geotiff`. The operation
  unpacks CF-packed data: integers stored with a SCALE/OFFSET that recover
  floats. It pairs with the writer's `pack` option.
- `mask_and_scale` is now a deprecated alias. Passing it warns and forwards
  to `unpack`; passing both raises `TypeError`. Same sentinel pattern used
  for `mask_nodata` -> `masked`. Behaviour is unchanged: `unpack=True`
  applies SCALE/OFFSET and masks the nodata sentinel to NaN.
- `unpack` now runs on the GPU eager, GPU-via-CPU, and dask+GPU paths. The
  scale/offset arithmetic already ran on CuPy via numpy duck-typing in the
  shared eager finalizer; the change threads the flag through the GPU eager
  sites and the dask+GPU chunked path and drops the `gpu=True` rejection.
  The dask+GPU disk->GPU GDS fast path has no scale step, so an `unpack`
  request takes the CPU-decode-then-upload route, which applies the scale
  lazily before the device upload.
- The merged `coregister` feature (#3072) read with `mask_and_scale=True`
  internally; it now passes `unpack=True` so it no longer trips its own
  deprecation warning. Its CPU-only gpu/.vrt rejection is unchanged.

## Scope

- VRT sources still raise for `unpack`. Out of scope.
- `parse_coordinates=False` GPU gating is unchanged. Out of scope.
- The `attrs['mask_and_scale_dtype']` round-trip key keeps its name; it
  pairs with the writer's pack path. Only the public parameter is renamed.

## Backends

numpy, dask+numpy, cupy, dask+cupy. Verified all four produce identical
values on a packed uint8 raster (`data*2+10`, sentinel -> NaN).

## Test plan

- [x] GPU eager and dask+GPU parity vs CPU (`@requires_gpu`)
- [x] Deprecation alias warns and matches `unpack`; both -> `TypeError`
- [x] `unpack` + `.vrt` still raises
- [x] Signature canonical-order test updated for the new param
- [x] coregister forwards `unpack`; coregister suite passes warning-strict
- [x] Existing pack / dtype-parity / rioxarray-compat suites pass

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -201,7 +201,8 @@ this section is the brief.
   nodata sentinels raises ``MixedBandMetadataError`` by default. Pass
   ``band_nodata='first'`` to opt back into the legacy flatten-to-band-0
   behaviour; see ``xrspatial/geotiff/tests/vrt/test_metadata.py``.
-* Malformed scale/offset. Under ``mask_and_scale=True`` the read applies
+* Malformed scale/offset. Under ``unpack=True`` (formerly
+  ``mask_and_scale=True``) the read applies
   the source's ``SCALE`` / ``OFFSET`` from ``GDAL_METADATA``. A key that
   is present but does not parse as a float (for example ``SCALE="abc"``)
   raises ``MalformedScaleOffsetError`` (a ``GeoTIFFAmbiguousMetadataError``
--- a/docs/source/user_guide/attrs_contract.rst
+++ b/docs/source/user_guide/attrs_contract.rst
@@ -116,10 +116,11 @@ write.
        #2135.
    * - ``mask_and_scale_dtype``
      - str
-     - Integer source dtype name (e.g. ``"int8"``), set only when a
-       ``mask_and_scale=True`` read promoted an integer array to float.
-       ``to_geotiff(pack=True)`` reads it to reverse the promotion and
-       restore the on-disk dtype. Added in contract v5 (issue #3064).
+     - Integer source dtype name (e.g. ``"int8"``), set only when an
+       ``unpack=True`` (formerly ``mask_and_scale=True``) read promoted an
+       integer array to float. ``to_geotiff(pack=True)`` reads it to reverse
+       the promotion and restore the on-disk dtype. The attr keeps its
+       ``mask_and_scale_dtype`` name. Added in contract v5 (issue #3064).
    * - ``raster_type``
      - str
      - ``'point'`` when the file declares ``RasterPixelIsPoint``;
@@ -384,10 +385,11 @@ round-trip until ``to_geotiff`` learns to emit
 ``ModelTransformationTag`` (issue #2115 follow-up).
 
 Contract v5 (issue #3064) added the ``mask_and_scale_dtype`` attr to
-the canonical tier. The attr records the integer source dtype when a
-``mask_and_scale=True`` read promoted the array to float, so
+the canonical tier. The attr records the integer source dtype when an
+``unpack=True`` read promoted the array to float, so
 ``to_geotiff(pack=True)`` can reverse the scale / offset, fill NaN back
 to the nodata sentinel, and restore the on-disk dtype. The packed file
 keeps its ``SCALE`` / ``OFFSET`` tags, so reopening it with
-``mask_and_scale=True`` unpacks to the same values instead of scaling a
-second time.
+``unpack=True`` unpacks to the same values instead of scaling a
+second time. (The read option was renamed from ``mask_and_scale`` to
+``unpack`` in issue #3071; the attr keeps its name.)
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -254,13 +254,13 @@ def _open_geotiff_windowed(obj, source, *, auto_reproject=False,
         is_vrt = str(source).lower().endswith('.vrt')
         if gpu_requested or is_vrt:
             raise ValueError(
-                "coregister=True applies mask_and_scale, which is not "
-                "supported with gpu=True or .vrt sources. Read on CPU, "
-                "or pass coregister=False."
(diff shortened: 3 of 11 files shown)


brendancol (commented):
## PR Review: Rename open_geotiff mask_and_scale to unpack and add GPU / dask+GPU support

Reviewed from the head-branch worktree. No blockers. The rename follows the
existing `mask_nodata` -> `masked` sentinel precedent and the GPU support
reuses the shared eager finalizer, so backend parity is structural rather
than reimplemented. I ran the four-backend smoke check and the GPU edge
cases (no-metadata no-op, per-band `MixedBandMetadataError`, band-selected
scale); all match the CPU path.

### Blockers
None.

### Suggestions
- [ ] `unpack` docstring (`xrspatial/geotiff/__init__.py:703`): the dask+GPU
  path silently disqualifies the disk->GPU GDS fast path when `unpack=True`
  and falls back to CPU-decode-then-upload (`_backends/gpu.py:1425`). That is
  correct, but it is a performance cliff documented only in a code comment.
  A user reading on a local tiled COG with `gpu=True, chunks=` loses the GDS
  route the moment they add `unpack=True`. One sentence in the docstring
  would set the expectation.

### Nits
- [ ] GPU edge-case coverage (`tests/read/test_rioxarray_compat_2961.py`):
  the new `@requires_gpu` tests cover the happy path (dataset-level scale).
  The no-metadata no-op and per-band paths are correct on GPU (verified
  locally) but only asserted on CPU. A couple of GPU asserts would lock the
  parity in CI rather than leaving it to the shared-finalizer argument.
- [ ] `_read_geotiff_gpu_chunked_gds` (`_backends/gpu.py:1470`) does not take
  `mask_and_scale`. Safe today because the caller disqualifies it, but a
  future direct caller would silently drop the unpack. A short comment on the
  helper noting it is unpack-free would prevent that.

### What looks good
- Sentinel deprecation matches the `mask_nodata`/`name` precedent exactly,
  including the both-passed `TypeError` and the documented `unpack=False`
  resolve-to-alias corner.
- `mask_and_scale_dtype` attr name preserved, so the writer pack path and the
  v5 attrs contract are untouched.
- The merged `coregister` feature was migrated to `unpack` internally; its
  suite passes warning-strict, so coregister no longer trips its own
  deprecation warning.
- Signature canonical-order test updated alongside the new parameter.

### Checklist
- [x] Behaviour unchanged vs `mask_and_scale` (scale/offset + nodata mask)
- [x] All four backends produce identical values (verified)
- [x] NaN handling correct (sentinel -> NaN survives the scale)
- [x] GPU edge cases match CPU (no-op, per-band, band select)
- [x] Deprecation alias warns; both -> TypeError; VRT still rejects
- [x] Docstrings and attrs contract updated
- [x] No new function, so README matrix unchanged (justified)

brendancol (commented):
## Follow-up review (after 9b9921b3)

Re-reviewed the follow-up commit. All three findings from the first pass are
addressed; the changes are docs and tests only, no logic moved.

### Disposition
- Suggestion (dask+GPU perf cliff): fixed. The `unpack` docstring now states
  the dask+GPU read takes the CPU-decode route instead of the GDS fast path
  (`xrspatial/geotiff/__init__.py:715`).
- Nit (thin GPU edge-case coverage): fixed. Added `@requires_gpu` tests for
  the no-metadata no-op and the integer-dtype-cast raise on GPU
  (`tests/read/test_rioxarray_compat_2961.py`). Per-band parity was already
  verified manually; the new asserts lock the no-op and dtype paths in CI.
- Nit (`_read_geotiff_gpu_chunked_gds` unpack-free): fixed. The helper
  docstring now records that it is unpack-free and names the fallback a
  future direct caller must use (`_backends/gpu.py:1490`).

No remaining open items. Nothing here blocks merge.

PR #3077, opened 2026-06-09
Silence pyproj PROJ-string warning in _get_datum_params (#3076)

### LINKED ISSUE #3076: reproject leaks a pyproj 'lose important projection information' warning on every call
**Describe the bug**

`reproject` prints a pyproj `UserWarning` on every call, and so does anything built on it (`open_geotiff(..., auto_reproject=True)` and `coregister=True`):

```
UserWarning: You will likely lose important projection information when converting to a PROJ string from another format.
```

The warning is not raised by user code. It comes from `_get_datum_params` in `xrspatial/reproject/_projections.py`, which calls `crs.to_dict()`. pyproj's `to_dict()` runs `to_proj4()` under the hood, and pyproj warns whenever it downgrades a CRS to a PROJ string.

`_get_datum_params` only reads the `datum`/`ellps` keys so it can look up a custom 7-parameter datum shift in `_DATUM_PARAMS`. For a CRS with no entry (e.g. EPSG:3857) it returns `None` and the PROJ string is never used for anything. So the warning is a side effect of probing the datum, not a real loss of information in the output.

**Reproduce**

```python
from xrspatial import open_geotiff
study_area = open_geotiff("N35W112.tif", chunks=2048)
proj = study_area.xrs.reproject("3857", resampling="nearest")
# UserWarning: You will likely lose important projection information ...
```

Call chain: `reproject` -> `_compute_output_grid` -> `_transform_boundary` -> `transform_points` -> `_get_datum_params` -> `crs.to_dict()`.

**Expected behavior**

`_get_datum_params` should read the datum/ellipsoid without surfacing pyproj's PROJ-string warning to the caller. The datum lookup still needs the proj4 short names (`tokyo`, `BD72`, `CH1903`, ...) that only `to_dict()` exposes, so the fix is to suppress that one pyproj warning around the probe rather than drop `to_dict()`.

**Additional context**

- File: `xrspatial/reproject/_projections.py`, `_get_datum_params`
- The probe already swallows arbitrary exceptions and returns `None`; suppressing the warning fits the same "best-effort lookup" intent.

brendancol:
Closes #3076

## What

`reproject` (and `open_geotiff` with `auto_reproject=True` / `coregister=True`) printed a pyproj `UserWarning` on every call:

> You will likely lose important projection information when converting to a PROJ string from another format.

The warning came from `_get_datum_params`, which calls `crs.to_dict()` to read the datum/ellipsoid short names for a custom 7-parameter shift lookup. pyproj's `to_dict()` runs `to_proj4()` under the hood and warns about the lossy conversion. The probe never uses that string, so the warning was noise leaking to callers.

- Wrap the `to_dict()` probe in `warnings.catch_warnings()` and filter out that one pyproj message. The datum lookup still uses the proj4 short names, so behaviour is unchanged.

## Backend coverage

CRS-metadata helper shared by all backends (numpy / cupy / dask+numpy / dask+cupy); no backend-specific code touched.

## Test plan

- [x] `TestDatumProbeNoProjWarning::test_get_datum_params_silences_proj_warning` — no PROJ-string warning leaks for EPSG:3857 (fails without the fix, passes with it).
- [x] `TestDatumProbeNoProjWarning::test_get_datum_params_still_resolves_known_datum` — NAD27 (EPSG:4267) still resolves to shift params.
- [x] Full `test_reproject.py` suite passes (429 passed).

### DIFF
--- a/xrspatial/reproject/_projections.py
+++ b/xrspatial/reproject/_projections.py
@@ -22,6 +22,7 @@ All other CRS pairs fall back to pyproj.
 from __future__ import annotations
 
 import math
+import warnings
 
 import numpy as np
 from numba import njit, prange
@@ -127,7 +128,16 @@ def _get_datum_params(crs):
     Returns None for WGS84/NAD83/GRS80 (no shift needed).
     """
     try:
-        d = crs.to_dict()
+        # crs.to_dict() goes through pyproj's to_proj4(), which warns that a
+        # PROJ string drops detail. We only read the datum/ellps short names
+        # here, never the lossy string, so silence that one warning.
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                'ignore',
+                message='.*lose important projection information.*',
+                category=UserWarning,
+            )
+            d = crs.to_dict()
     except Exception:
         return None
     datum = d.get('datum', '')
--- a/xrspatial/tests/test_reproject.py
+++ b/xrspatial/tests/test_reproject.py
@@ -1,6 +1,8 @@
 """Tests for xrspatial.reproject module."""
 from __future__ import annotations
 
+import warnings
+
 import numpy as np
 import pytest
 import xarray as xr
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Silence pyproj PROJ-string warning in _get_datum_params

### Blockers
None.

### Suggestions
None.

### Nits (optional)
- `xrspatial/reproject/_projections.py:131` — `warnings.catch_warnings()` mutates global warning-filter state and is not thread-safe. Not a practical problem here: `_get_datum_params` runs during eager grid setup on the main thread (`transform_points` -> `_transform_boundary` -> `_compute_output_grid`), not inside dask per-chunk tasks, so concurrent threads will not trample each other's filters. Mental note only; no change needed.

### What looks good
- Root cause is correct: `to_dict()` -> `to_proj4()` is what pyproj warns on, and the probe never consumes the lossy string, so suppressing the one message is the right call rather than dropping `to_dict()`. The `_DATUM_PARAMS` keys are proj4 short names that only `to_dict()` exposes.
- Surgical scope: one filter, matched by message and `category=UserWarning`, so unrelated warnings still propagate. The existing `except Exception: return None` best-effort behaviour is preserved.
- The unit test genuinely guards the fix: it fails without the change and passes with it. The companion test confirms the NAD27 datum lookup still resolves, so the suppression did not break the lookup it wraps.

### Checklist
- [x] Mechanism matches the reported cause (to_proj4 lossy-conversion warning)
- [x] NaN/units/edge cases -- n/a (CRS-metadata helper, no array math)
- [x] Backend parity -- n/a (eager helper, not in the dask graph)
- [x] Test fails without fix, passes with it; datum lookup regression covered
- [x] No new public API, so no README/docstring/benchmark changes needed

Trail note: an end-to-end `reproject(...)` test was written and then removed because it passed with and without the fix. The synthetic 4326->3857 path does not construct the CRS object in a way that triggers pyproj's warning, so it was a hollow guard. The unit test on `_get_datum_params` with `from_epsg(3857)` reproduces the exact warning and is the reliable regression guard.

PR #3079, opened 2026-06-09
Restore nodata sentinel for float rasters in to_geotiff(pack=True)

### LINKED ISSUE #3078: to_geotiff(pack=True) leaves float nodata pixels as NaN instead of the declared sentinel
**Describe the bug**

`to_geotiff(..., pack=True)` is supposed to be the inverse of an `open_geotiff(mask_and_scale=True)` read: un-scale, restore the source dtype, and fill the masked NaN pixels back to the declared nodata sentinel. It does that for integer source rasters. For float source rasters it doesn't: the masked NaN pixels stay NaN in the output, even though the file still writes a GDAL_NODATA tag carrying the original sentinel.

So you get an inconsistent file. The header says nodata = -9999, but the nodata pixels on disk are NaN. Anything that reads the file without applying masking (plain GDAL, QGIS, rasterio without `masked=True`) sees NaN where it expects -9999 and won't treat those pixels as nodata.

**Root cause**

In `_pack` (`xrspatial/geotiff/_attrs.py`), the `fillna(nodata)` call lives inside the `if tgt.kind in ('i', 'u')` branch, so it only runs for integer target dtypes. Float targets skip the fill and the NaN pixels survive into the written file. The docstring already says the function "fills NaN back to the declared nodata sentinel" with no dtype qualifier, so the code just doesn't match its own contract.

**Reproduce**

Open a float32 raster with a nodata sentinel of -9999 using `mask_and_scale=True`, then write it back with `pack=True`. The sentinel pixel comes out as NaN, but the file still carries a GDAL_NODATA tag of -9999.

**Expected behavior**

When a nodata sentinel is declared, `pack=True` should restore the masked NaN pixels to that sentinel for float source rasters too, so the pixels on disk match the declared GDAL_NODATA tag.

**Environment**
- xarray-spatial main
- numpy and dask CPU backends (`mask_and_scale` is a CPU read feature)

brendancol:
Closes #3078

`to_geotiff(pack=True)` is the inverse of an `open_geotiff(unpack=True)` read. For integer rasters it un-scales, fills the masked NaN pixels back to the nodata sentinel, and restores the integer source dtype. For float rasters it did everything except the fill: the NaN pixels stayed NaN while the file still wrote a GDAL_NODATA tag for the sentinel. That leaves an inconsistent file (header says nodata=-9999, pixels are NaN), and a non-masking reader like plain GDAL or QGIS treats those NaN pixels as valid data.

The `fillna(nodata)` call lived inside an `if tgt.kind in ('i', 'u')` branch, so float targets skipped it. The fill now runs for every dtype when a sentinel is declared, which is what the docstring already promised. Integer behavior is unchanged: the no-sentinel-but-NaN error path and the int-only `round()` both still apply.

`_pack` is plain xarray (`fillna` / `round` / `astype`) and runs before the writer dispatches, so the fix is backend-agnostic. Exercised here on numpy and dask.

Test plan:
- Added `test_pack_restores_float_nodata_sentinel` (numpy + dask): a float32 raster with a -9999 sentinel, read with `unpack=True` (masks it to NaN), packed back. Asserts no NaN survives, the pixel equals the sentinel, the GDAL_NODATA tag matches, and an `unpack=True` read of the packed file round-trips back to NaN.
- Confirmed the new test fails without the fix and passes with it.
- Full geotiff write + attrs suites pass (1116 passed, 1 skipped).

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -1699,17 +1699,22 @@ def _pack(data):
         target = np.asarray(nodata).dtype.name
     tgt = np.dtype(target) if target is not None else np.dtype(str(out.dtype))
 
+    if nodata is not None:
+        # Restore the masked sentinel for every dtype. Integers can't hold
+        # NaN at all; floats would otherwise write NaN pixels under a
+        # GDAL_NODATA tag that still declares the sentinel, leaving an
+        # inconsistent file that non-masking readers misread (#3078).
+        out = out.fillna(nodata)
+    elif tgt.kind in ('i', 'u'):
+        # ``isnull().any()`` forces a compute on dask; only reached on the
+        # error path where no sentinel exists to fill an integer's holes.
+        if bool(out.isnull().any()):
+            raise ValueError(
+                f"pack=True: cannot restore integer dtype {tgt.name}: "
+                "NaN pixels are present but no nodata sentinel is "
+                "declared to fill them.")
+
     if tgt.kind in ('i', 'u'):
-        if nodata is None:
-            # ``isnull().any()`` forces a compute on dask; only reached on
-            # the error path where no sentinel exists to fill the holes.
-            if bool(out.isnull().any()):
-                raise ValueError(
-                    f"pack=True: cannot restore integer dtype {tgt.name}: "
-                    "NaN pixels are present but no nodata sentinel is "
-                    "declared to fill them.")
-        else:
-            out = out.fillna(nodata)
         out = out.round()
 
     out = out.astype(tgt)
--- a/xrspatial/geotiff/tests/write/test_pack_3064.py
+++ b/xrspatial/geotiff/tests/write/test_pack_3064.py
@@ -83,6 +83,39 @@ def test_pack_restores_uint16_no_scale(tmp_path, chunks):
 
 
 # ---------------------------------------------------------------------------
+# Float source: the masked sentinel must be restored, not left as NaN (#3078)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("chunks", [None, 2], ids=["numpy", "dask"])
+def test_pack_restores_float_nodata_sentinel(tmp_path, chunks):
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Restore nodata sentinel for float rasters in to_geotiff(pack=True)

None.

None.

- `xrspatial/geotiff/tests/write/test_pack_3064.py:98` -- the new test hands a float32 array to `_write_int_tiff`, whose docstring says "an integer ndarray". The helper is dtype-agnostic so it works, but the name reads a little off now. Not worth renaming a shared helper for one caller; just noting it.

### Observations (out of scope, pre-existing)
- A float32 source comes back as float64 from pack. `_pack` infers the target dtype from the nodata scalar (`np.asarray(nodata).dtype.name`, `_attrs.py:1699`) when `mask_and_scale_dtype` is absent, and a float nodata is float64. The sentinel value is right, so this is a width-only fidelity gap, separate from the nodata bug this PR fixes. Flagging in case it deserves its own issue.

### What looks good
- The branch restructure is behavior-preserving for integers: the no-sentinel-but-NaN error path and the int-only `round()` both still fire exactly as before, confirmed by the 11 existing pack tests.
- `fillna` stays lazy on dask. The only forced compute (`isnull().any()`) is still on the nodata-None integer error path, same as before.
- The new test asserts real values (sentinel pixel, GDAL_NODATA tag, no surviving NaN, unpack round-trip), not just that it runs. Covers numpy and dask, and the temp files carry the issue number.
- Confirmed the test fails without the fix.

### Checklist
- [x] Algorithm matches the documented contract (the docstring already promised the fill for all dtypes)
- [x] Integer and float paths produce consistent results
- [x] NaN handling is correct (sentinel restored, no NaN left under a nodata tag)
- [x] Edge cases covered (float source, numpy + dask)
- [x] Dask handled correctly (lazy fillna, no new materialization)
- [ ] Benchmark -- not needed (bug fix, no perf-critical path)
- [ ] README feature matrix -- not applicable (no new function)
- [x] Docstring present and accurate (matches the fixed behavior)

PR #3081, opened 2026-06-09
Make merge() output-size guard backend-aware (#3048)

### LINKED ISSUE #3048: merge() output-size guard blocks legitimate lazy dask merges
**Describe the bug**

`merge()` (`xrspatial/reproject/__init__.py`) computes the output grid with `_compute_output_grid` before it decides whether the result will be dask-backed, so the 1-billion-pixel memory guard runs against every merge. The function then detects dask inputs and even auto-promotes large in-memory merges to the lazy `_merge_dask` path. A merge whose output exceeds 1e9 pixels but runs lazily through `_merge_dask` is still rejected by the guard.

This is the same pattern fixed for `reproject()` in #3046 / #3047.

**Why it wasn't fixed alongside #3046**

The reproject fix detects the backend before computing the grid and passes `lazy_output` into `_compute_output_grid`. For `merge()` the ordering is trickier: the auto-promote-to-dask decision (around line 2328) depends on `out_shape`, which only exists after the grid is computed. Real dask inputs can be detected up front, but the "large in-memory output auto-promotes to dask" case needs a small restructuring to know the output will be lazy before the guard fires. Kept separate to avoid expanding the scope of #3046.

**Expected behavior**

Lazy dask merges (dask inputs, or large in-memory merges that auto-promote to `_merge_dask`) should not be blocked by the 1e9 guard. The guard should still protect the in-memory merge path.

brendancol:
Closes #3048

`merge()` computed the output grid with the default `lazy_output=False`, so the 1-billion-pixel guard in `_compute_output_grid` fired for every merge, before `merge()` had decided whether the result would be dask-backed. That rejected merges whose output exceeds 1e9 pixels but actually run lazily through `_merge_dask`, where the full array is never materialized. It hit two cases:

- an input is already a dask array, and
- a large in-memory output that auto-promotes to the dask path.

## Fix

- Detect dask inputs up front and compute the output grid with the guard disabled (`lazy_output=True`).
- Re-apply the pixel check only when the final path is the in-memory merge, so a genuinely in-memory output over the limit still raises.
- Hoist the auto-promote byte budget to a module constant (`_MERGE_OOM_THRESHOLD`) so the in-memory guard branch is reachable in tests.

Same approach as the `reproject()` fix in #3046 / #3047.

## Backend coverage

`merge()` runs the same grid logic for all inputs; the change is in the path selection, not in any backend kernel. The added tests cover numpy inputs (in-memory and auto-promote) and dask inputs.

## Test plan

- [x] dask-input merge whose output exceeds 1e9 pixels stays lazy and does not raise
- [x] large in-memory merge that auto-promotes to dask does not raise
- [x] genuinely in-memory merge over the limit still raises the "too large" ValueError
- [x] full `test_reproject.py` suite passes (432 tests)

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -23,6 +23,7 @@ from ._crs_utils import (
     _resolve_crs,
 )
 from ._grid import (
+    _MAX_OUTPUT_PIXELS,
     _chunk_bounds,
     _compute_chunk_layout,
     _compute_output_grid,
@@ -2346,8 +2347,8 @@ def merge(
         # output array. Re-apply the output-size guard that was skipped
         # during grid computation so a genuinely in-memory merge over the
         # pixel limit still raises (the guard skip only applies to the
-        # lazy dask path).
-        _MAX_OUTPUT_PIXELS = 1_000_000_000
+        # lazy dask path). _MAX_OUTPUT_PIXELS is imported from _grid so the
+        # limit stays in sync with _compute_output_grid's own check.
         if out_shape[0] * out_shape[1] > _MAX_OUTPUT_PIXELS:
             raise ValueError(
                 f"Computed output grid is too large ({out_shape[1]} x "
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -63,6 +63,10 @@ __all__ = [
 _Y_NAMES = {'y', 'lat', 'latitude', 'Y', 'Lat', 'Latitude'}
 _X_NAMES = {'x', 'lon', 'longitude', 'X', 'Lon', 'Longitude'}
 
+# Output byte budget above which merge() auto-promotes an in-memory mosaic
+# to the lazy dask path instead of allocating the whole array.
+_MERGE_OOM_THRESHOLD = 512 * 1024 * 1024  # 512 MB
+
 # Map friendly vertical datum tokens to EPSG codes so attrs['vertical_crs']
 # from reproject output matches the convention used by xrspatial.geotiff,
 # which also writes EPSG ints to attrs['vertical_crs'].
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Make merge() output-size guard backend-aware (#3048)

None.

- xrspatial/reproject/__init__.py:2350-2354: the "too large" guard and its `_MAX_OUTPUT_PIXELS = 1_000_000_000` constant are duplicated from _grid.py:447-451. If the limit or the message changes in _grid.py, this copy drifts silently. The auto-promotion chicken-and-egg (it needs out_shape) makes a single grid call awkward, so the duplication is understandable. A short comment pointing at _grid.py as the source of truth, or importing the constant, would keep the two in sync.

None.

### What looks good
- The fix follows the reproject() approach from #3046 and matches what the issue asked for: detect dask up front, compute the grid with the guard off, re-apply the pixel check only on the in-memory branch.
- The in-memory guard is preserved. A genuinely in-memory output over the limit still raises, and the third test proves it by monkeypatching the auto-promote threshold so the in-memory path is actually reached.
- Error message wording and the width-by-height ordering match _grid.py.
- Tests cover all three cases from the issue: dask input over the limit stays lazy, in-memory auto-promote over the limit stays lazy, genuinely in-memory over the limit still raises.

### Checklist
- Algorithm matches reference: N/A, path-selection fix with no kernel change
- Backend results consistent: path selection per backend is unchanged
- NaN handling: untouched
- Edge cases covered by tests: yes, the three guard cases
- Dask chunk boundaries: untouched
- No premature materialization: the lazy path is no longer blocked, which is the point
- Benchmark: not needed for a bug fix
- README feature matrix: N/A, no new function
- Docstrings: merge() docstring unchanged and still accurate

brendancol (commented):
## Follow-up review (after c632c661)

The one Suggestion from the first pass is resolved.

- The duplicated 1e9 pixel constant is gone. `_MAX_OUTPUT_PIXELS` now lives as a module constant in _grid.py and merge() imports it, so the in-memory guard and `_compute_output_grid`'s own check read the same value. If the limit changes, both move together.

No new findings. Full test_reproject.py suite still passes (432 tests).

PR #3109, opened 2026-06-09
Reject non-finite burn values against integer and bool output dtypes (#3085)

### LINKED ISSUE #3085: rasterize: NaN/inf burn values silently cast to INT_MIN under integer output dtype
**Describe the bug**

`rasterize()` computes in float64 and casts to the output dtype at the end. When the output dtype is an integer type and a burn value is non-finite (NaN or +/-inf, e.g. a GeoDataFrame column with missing data), the final `astype` lands on a platform sentinel:

```python
import numpy as np

res = rasterize([(box(0, 0, 5, 5), np.nan), (box(5, 5, 10, 10), 7.0)],
                width=10, height=10, dtype=np.int32, fill=0)
np.unique(res.data)  # array([-2147483648, 0, 7], dtype=int32)
```

All four backends produce INT_MIN for the NaN pixels. The numpy path is silent because `_run_numpy` suppresses the "invalid value encountered in cast" RuntimeWarning; the dask tile path leaks the warning from workers but still writes the sentinel. A bool output dtype is worse: NaN casts to `True`.

The same class of problem is already guarded for two neighbours of this case:

- NaN **fill** into an integer dtype raises (#2504)
- burn values above 2**53 - 1 into an integer dtype raise (#3056)

The non-finite **burn value** case is the remaining gap.

**Expected behavior**

A non-finite burn value combined with an integer or bool output dtype should raise `ValueError` up front, naming the offending value, before any host or device allocation, on all four backends. Float output dtypes are unaffected since NaN/inf are representable there.

`merge='count'` never burns property values, so NaN attributes under `count` should keep working.

Found by the accuracy sweep against `xrspatial/rasterize.py`.

brendancol:
Closes #3085

`rasterize()` computes in float64 and casts to the output dtype at the end. A NaN or inf burn value (a GeoDataFrame column with missing data is the usual source) against an integer dtype landed on a platform sentinel: `np.unique(result)` showed `-2147483648` where the NaN geometry burned. The numpy backend was fully silent because `_run_numpy` suppresses the cast RuntimeWarning; bool dtypes collapsed NaN to `True`.

- Add an up-front `ValueError` for non-finite burn values when the output dtype is integer or bool, placed next to the unsafe-integer guard from #3056 and worded like the NaN-fill guard from #2504. It names the offending value and fires before any host or device allocation, so all four backends (numpy, cupy, dask+numpy, dask+cupy) reject identically.
- `merge='count'` is exempt: it burns overlap counts and never reads property values, so NaN attribute columns keep working there.
- Float output dtypes are untouched; NaN/inf are representable.
- Document the new rejection in the `dtype` docstring paragraph.

Test plan:
- [x] New `test_rasterize_nonfinite_burn_3085.py`: 24 tests covering NaN/+inf/-inf across int16/int32/uint8/int64, bool, the `like=` integer-template path, GeoDataFrame `column`/`columns` paths, the count-merge exemption, float-dtype passthrough, and the guard firing on all four backends (cupy and dask+cupy ran on GPU).
- [x] Full rasterize suite: 665 passed, 2 skipped.
- [x] flake8 clean on both touched files.


brendancol (commented):
## PR Review: Reject non-finite burn values against integer and bool output dtypes (#3085)

None.

None.

- [ ] xrspatial/rasterize.py:3623 - the new guard exempts `merge='count'`, but the unsafe-integer guard from #3056 directly above (line 3595) does not, so `merge='count'` with an attribute above 2**53-1 still raises even though count never burns the value. Pre-existing behavior, not made worse by this PR; noting it in case anyone wants to align the two guards later.
- [ ] xrspatial/rasterize.py:3626 - for an integer dtype this recomputes `np.isfinite(props_array)` right after the #3056 block computed the same thing. Props tables are small, so this costs nothing in practice; merging the two blocks would save a pass but make the conditions (bool inclusion, count exemption) harder to read. Fine as is.

### What looks good
- Guard sits before backend dispatch and before any host or device allocation, so all four backends reject identically with no per-backend code.
- `merge` is validated earlier in the function, so the `merge != 'count'` comparison only ever sees a known string or a callable, and callables correctly fall on the guarded side.
- Error message names the offending value and mirrors the wording of the #2504 and #3056 guards; docstring updated in the same paragraph that documents the 2**53 rule.
- Tests cover NaN/+inf/-inf across four integer dtypes plus bool, the `like=` template path, both GeoDataFrame burn paths, the count exemption, float passthrough, and all four backends (GPU tests ran on hardware). 24 new tests pass; full rasterize suite at 665 passed.

### Checklist
- [x] Algorithm matches reference/paper (n/a, validation guard)
- [x] All implemented backends produce consistent results (guard fires pre-dispatch)
- [x] NaN handling is correct
- [x] Benchmark exists or is not needed (validation-only change)
- [x] README feature matrix updated (no new function)

PR #3110, opened 2026-06-09
Add complementary non-finite burn-value tests for rasterize (#3088)

### LINKED ISSUE #3088: rasterize: NaN/inf burn values silently become INT_MIN under integer output dtypes
#### Describe the issue

`rasterize()` rejects a NaN *fill* against an integer dtype (#2504, #3059) and rejects finite burn values above 2**53 - 1 (#3056), but non-finite *burn values* slip through both guards. The cast at the end of each backend then turns them into garbage with no error and no warning:

```python
import numpy as np
from xrspatial import rasterize

r = rasterize([(box(0, 0, 5, 5), float('nan'))],
              width=8, height=8, fill=0, dtype=np.int32)
np.unique(r.values)   # [-2147483648]
```

- `nan` or `inf` into any integer dtype burns INT_MIN (-2147483648 for int32).
- `nan` into `dtype=bool` burns `True`.
- Same result on all four backends (numpy, cupy, dask+numpy, dask+cupy).
- The numpy backend suppresses the "invalid value encountered in cast" RuntimeWarning in `_run_numpy`, so there is no signal at all.

The practical case is a GeoDataFrame burn column with missing values plus `like=` an integer template raster: every feature with a NaN attribute lands on INT_MIN, which then masquerades as valid data (it does not match the `nodata` attr either).

#### Expected behavior

A `ValueError` before any backend dispatch, same as the NaN-fill guard: name the offending value and suggest using a float dtype or cleaning the column. NaN burns into *float* dtypes must keep working; that propagation is intended and tested (#2255).

#### Where

`xrspatial/rasterize.py` - the burn-value guard at the bottom of `rasterize()` only checks `np.isfinite(props_array)` values against the safe-int range, so `~finite` entries are never examined.

brendancol:
Closes #3088

The guard itself landed in #3109 (#3085 was the same bug, found independently by the accuracy sweep). After rebasing onto main, this PR carries what #3109 doesn't have:

- `test_rasterize_nonfinite_burn_3088.py`: 22 tests covering all 8 integer dtypes plus bool, NaN/inf/-inf burn values, mixed finite+NaN input, `merge='sum'`, a GeoDataFrame burn column with missing data, `like=` int template, and all four backends (CUDA exercised locally). Float-dtype NaN propagation (#2255) pinned as unaffected.
- The security-sweep state CSV row for the rasterize module.

All 22 tests pass against the guard on main, including its `merge='count'` exemption.

Test plan:
- [x] `pytest xrspatial/tests/test_rasterize_nonfinite_burn_3088.py`: 22 passed
- [x] flake8 clean


brendancol (commented):
## PR Review: Reject non-finite burn values for integer/bool output dtypes (#3088)

None.

- [ ] xrspatial/rasterize.py (new guard block, ~line 3607): the check is unconditional across merge modes. Under `merge='count'` the property value never reaches the output, and with `columns=` only `props[0]` is written by the built-in merges, so a NaN in a secondary column now raises even though the output would be fine. This matches how the adjacent #3056 safe-int guard already behaves, which is a fair call, but the comment block doesn't say the strictness is deliberate. One sentence noting "unconditional across merge modes, consistent with the safe-int guard below" would stop a future reader from 'fixing' it.

- [ ] test_rasterize_nonfinite_burn_3088.py: the guard is exercised only through the `(geom, value)` pairs path. A GeoDataFrame with a NaN attribute column is the failure mode the issue describes, and `_parse_input` builds props_array differently there (`.values.astype(float64)` vs `float()` per item). Same code path after parsing, so low risk, but one gdf-based test would pin the realistic entry point. Skip it if geopandas isn't a test dependency you want here.

### What looks good
- Guard sits before backend dispatch, so numpy, cupy, dask+numpy, and dask+cupy all reject identically; the cupy/dask+cupy tests actually ran on this host.
- Float dtypes are explicitly untouched and the #2255 NaN-propagation behaviour is pinned by a test rather than assumed.
- Error message names the offending value and gives both remediations (float dtype, or clean the column), matching the tone of the neighbouring guards.
- Docstring for `dtype` was updated to document the new rejection.

### Checklist
- [x] Algorithm matches reference/paper (n/a, validation guard)
- [x] NaN handling is correct (float propagation preserved)
- [x] Edge cases are covered by tests (mixed finite/NaN, bool, like=, empty short-circuit)
- [x] Dask chunk boundaries handled correctly (guard fires before graph construction)
- [x] Benchmark exists or is not needed (not needed: O(N) bool mask)

brendancol (commented):
Follow-up review after 310eb4b3:

- Suggestion (document deliberate unconditional strictness): fixed. The guard's comment now states it applies across merge modes and property columns on purpose, matching the safe-int guard below it.
- Nit (GeoDataFrame entry point untested): fixed. `test_nan_attribute_column_gdf_int_dtype_raises` burns a gdf with a NaN attribute column into int32 and asserts the ValueError, behind a geopandas skipif.

No new findings. 55 tests across the three burn/fill guard files pass, flake8 clean.

brendancol:
Rebased now that #3109 has merged. It landed the same guard (#3085 and #3088 turned out to be the same bug, found independently by two parallel sweeps), so I dropped the duplicate source change from this PR.

What remains is the test coverage #3109 doesn't have: all eight integer dtypes, mixed finite/NaN burn columns, merge='sum', a GeoDataFrame NaN column, and the dask/cupy backends, plus the security-sweep state row for rasterize. All 22 tests pass against the guard on main, including its merge='count' exemption. Retitled to match.

PR #3113, opened 2026-06-09
Propagate GeoDataFrame CRS to rasterize output when like has none (#3087)

### LINKED ISSUE #3087: rasterize: GeoDataFrame CRS dropped when no like= template is given
**Describe the bug**

`rasterize(gdf, width=..., height=...)` returns a DataArray with no CRS metadata even when the input GeoDataFrame has one. `_parse_input` extracts `gdf.crs`, but since #3058 it is only used for the mismatch check against `like`. On the no-`like` path the value is thrown away, so the output has no `crs` attr, no `crs_wkt`, and no `spatial_ref` coord.

```python
import geopandas as gpd

gdf = gpd.GeoDataFrame({'value': [1.0], 'geometry': [box(2, 2, 8, 8)]}, crs='EPSG:32610')
out = rasterize(gdf, width=10, height=10, bounds=(0, 0, 10, 10), column='value', fill=0)
print(out.attrs)  # {'nodata': 0, '_FillValue': 0, 'nodatavals': (0,)} -- no CRS
```

Consequences:

- `to_geotiff(rasterize(gdf, ...))` writes a GeoTIFF with no CRS.
- `polygonize(rasterize(gdf, ...))` returns a GeoDataFrame with `crs=None`, so a polygonize -> rasterize -> polygonize round trip loses the CRS.

All four backends (numpy, cupy, dask+numpy, dask+cupy) behave the same because they share the final DataArray constructor.

**Expected behavior**

The output grid is in the geometry CRS (bounds come from the geometry coordinates), so the output should carry that CRS. Following the geotiff attrs convention in `xrspatial/geotiff/_attrs.py`: emit `attrs['crs']` as an EPSG int when `pyproj.CRS.to_epsg()` resolves, otherwise `attrs['crs_wkt']` as a WKT string.

When `like` is given, the output keeps inheriting the template georeference (the #3058 check already guards mismatches). Only fill in the geometry CRS when the output would otherwise have none, so a template CRS is never overridden.

**Additional context**

Found by the metadata propagation sweep. One fix site in `rasterize()` after `out_attrs` is built (xrspatial/rasterize.py).

brendancol:
Closes #3087

`rasterize(gdf, width=..., height=...)` dropped the GeoDataFrame's CRS: `_parse_input` extracted `gdf.crs` but it was only consumed by the #3058 mismatch check, so the no-`like` output carried no `crs` attr, no `crs_wkt`, and no `spatial_ref` coord. `to_geotiff` then wrote an un-georeferenced file and `polygonize(rasterize(gdf))` came back with `crs=None`.

- After `out_attrs` is built, emit the geometry CRS when the output has no CRS from any other source: `attrs['crs']` as an EPSG int when `pyproj.CRS.to_epsg()` resolves, else `attrs['crs_wkt']` as WKT. This matches the geotiff attrs convention in `xrspatial/geotiff/_attrs.py`, and both forms are readable by `polygonize._detect_raster_crs` and `_like_crs`.
- A template CRS always wins: when `like` supplies `attrs['crs']`, `attrs['crs_wkt']`, or a `spatial_ref` coord, nothing changes (the #3058 `check_crs` guard has already compared it to the geometry CRS).
- `(geometry, value)` pair input is unchanged (no CRS to propagate).

Backends: numpy / cupy / dask+numpy / dask+cupy all route through the same final `xr.DataArray` constructor, so this is one fix site. The new tests run all four (GPU tests ran locally on real CUDA).

Test plan:
- [x] `xrspatial/tests/test_rasterize_geom_crs_3087.py` (11 tests): EPSG int emission, non-EPSG WKT fallback, CRS-less GeoDataFrame, pair-list input, `like` attrs/`spatial_ref` precedence, `like` without CRS, polygonize round trip, dask+numpy / cupy / dask+cupy backends
- [x] All existing rasterize tests: 652 passed, 2 skipped
- [x] flake8 clean on touched files


brendancol (commented):
## PR Review: Propagate GeoDataFrame CRS to rasterize output when like has none (#3087)

- none found

- [ ] `xrspatial/tests/test_rasterize_geom_crs_3087.py:131` builds the spatial_ref WKT via `gpd.GeoDataFrame(geometry=[], crs='EPSG:32610').crs.to_wkt()`. `pyproj.CRS('EPSG:32610').to_wkt()` says the same thing without constructing a throwaway frame, and the file already imports pyproj in two other tests.

- [ ] The state CSV diff swaps the focal/contour row order. That comes from the sweep tooling's sorted rewrite, so nothing to do here, just noting it for anyone reading the diff.

### What looks good
- The guard condition checks all three CRS carriers (`crs` attr, `crs_wkt` attr, `spatial_ref` coord) before emitting, so a template CRS is never overridden. The one case it does not check, a `like` whose CRS is only reachable through `.rio.crs` with no spatial_ref coord, is effectively unreachable in rioxarray's data model, and even there the #3058 check has already confirmed the two CRS values match before this code runs.
- EPSG-int-else-WKT matches what `attrs_to_metadata` in `xrspatial/geotiff/_attrs.py` parses, and both forms feed `polygonize._detect_raster_crs` and `_like_crs` without changes. The polygonize round-trip test pins the consumer side, not just the attr value.
- One fix site after backend dispatch, so numpy / cupy / dask+numpy / dask+cupy stay identical by construction; the four backend tests confirm it (GPU tests ran on real CUDA).
- `_PyprojCRS(geom_crs)` runs only on the GeoDataFrame path where `geom_crs` is already a pyproj CRS, so no new import burden for pair-list callers and no failure mode worth guarding.

### Checklist
- [x] Algorithm matches reference (geotiff attrs convention)
- [x] NaN handling is correct (n/a, metadata-only change)
- [x] Edge cases covered (no-EPSG CRS, CRS-less gdf, pair-list input, template precedence)
- [x] Benchmark not needed (attrs emission only)
- [x] Docstrings updated (Returns section documents the new behavior)

brendancol (commented):
## Follow-up review (after e882869b)

The one suggestion from the first pass is fixed: `test_like_spatial_ref_coord_wins` now builds the WKT with `pyproj.CRS('EPSG:32610').to_wkt()` instead of a throwaway GeoDataFrame. All 11 tests in the file still pass and flake8 is clean.

The CSV row-order nit needed no action (sweep tooling writes rows sorted).

No new findings. Disposition: 1 suggestion fixed, 1 nit dismissed with reason, 0 blockers.

PR #3115, opened 2026-06-09
Name unpack instead of the deprecated mask_and_scale alias in geotiff docs and errors

### LINKED ISSUE #3086: geotiff docs and error messages still tell users to pass the deprecated mask_and_scale kwarg
#### Description

PR #3075 (issue #3071) renamed `open_geotiff`'s scale/offset kwarg from `mask_and_scale` to `unpack`, keeping `mask_and_scale` as a deprecated alias that emits a `DeprecationWarning`. A few user-facing strings never got the memo and still treat the old name as current:

- The `pack` entry in `to_geotiff`'s docstring (`xrspatial/geotiff/_writers/eager.py`) describes the option as "Inverse of ``open_geotiff(mask_and_scale=True)``" and says "reopening it with ``mask_and_scale=True`` unpacks to the original values".
- Errors raised on the public `unpack=True` read path name the alias: `MalformedScaleOffsetError` ("mask_and_scale=True cannot honour the scale / offset it may declare", "mask_and_scale=True cannot honour a malformed SCALE") and `MixedBandMetadataError` ("mask_and_scale=True but the source declares distinct per-band SCALE values ... or drop mask_and_scale") in `xrspatial/geotiff/_attrs.py`.
- Errors on the `pack=True` write path point at the alias too: "It was not produced by open_geotiff(mask_and_scale=True)." (`_attrs.py`) and "pack=True requires a DataArray carrying mask_and_scale attrs" (`_writers/eager.py`).

So a caller who passed `unpack=True` gets told to "drop mask_and_scale", a kwarg they never typed. And a caller who follows the `pack` docstring's advice gets a `DeprecationWarning` for their trouble.

#### Expected behaviour

Docstrings and error messages name `unpack`. Out of scope on purpose: the internal backend kwargs named `mask_and_scale` (private modules) and the `attrs['mask_and_scale_dtype']` key (contract v5) keep their names.

No API change here. The deprecated alias keeps working; only doc text and error strings change.

(Found by the api-consistency sweep against the geotiff module.)

brendancol:
Closes #3086.

- Error messages on the public `unpack=True` / `pack=True` paths now name `unpack` instead of the deprecated `mask_and_scale` alias: the two `MalformedScaleOffsetError` messages and the `MixedBandMetadataError` remedy ("or drop unpack") in `_attrs.py`, plus the `_pack` ValueError and the `to_geotiff(pack=True)` bare-array ValueError in `_writers/eager.py`.
- `to_geotiff`'s `pack` docstring now points at `open_geotiff(unpack=True)`. `attrs['mask_and_scale_dtype']` keeps its name (contract v5) and the docstring now says the attr name is historical.
- Left alone on purpose: the private backend kwargs still named `mask_and_scale` (`_backends/*`, `_finalize_eager_read`) and the attrs key itself.

Backend coverage: text-only change, no backend logic touched. The reworded strings are raised from the shared `_attrs.py` helpers, so eager, dask, and GPU reads all see the same wording.

Test plan:
- [x] New regression tests pin that the messages name `unpack` and never the alias (`tests/read/test_rioxarray_compat_2961.py` #3086 section, `tests/write/test_pack_3064.py`)
- [x] `pytest xrspatial/geotiff/tests/read/test_rioxarray_compat_2961.py xrspatial/geotiff/tests/write/test_pack_3064.py` -- 59 passed
- [x] `pytest xrspatial/geotiff/tests/release_gates/test_features.py` -- 180 passed, 4 skipped

Also carries the api-consistency sweep state CSV update for the geotiff module.


brendancol (commented):
## PR Review: Name unpack instead of the deprecated mask_and_scale alias in geotiff docs and errors

None.

- [ ] `xrspatial/geotiff/_attrs.py:1659-1676` -- the `_pack` helper's docstring still reads "Re-pack a ``mask_and_scale=True`` read for writing", "reopening with ``mask_and_scale=True`` unpacks to the same values", and "carries no mask_and_scale state to reverse". It is a private helper, so the PR's stated scope (user-facing strings) technically excludes it, but it documents exactly the public round-trip this PR rewords everywhere else, and the ValueError it raises was just changed to say "no unpack state" -- the docstring and the error it describes now disagree. Three small line edits.

- [ ] `xrspatial/geotiff/tests/read/test_rioxarray_compat_2961.py:578-593` -- `test_malformed_scale_error_names_unpack_not_alias` runs the same four assertions twice against two fixtures. A `pytest.mark.parametrize` over the fixture builder would halve it. Fine as-is for a two-case test.

### What looks good
- The reworded messages keep every loose regex in the pre-existing tests alive (`per-band SCALE`, `SCALE`, `OFFSET`, `XML`), so no unrelated test churn.
- The new assertions are careful about the one old-name token that is allowed to stay: `test_pack_3064.py` checks `"mask_and_scale=True" not in msg` rather than the bare token, since the message legitimately names the `mask_and_scale_dtype` attrs key.
- Scope control is explicit: private backend kwargs and the contract-v5 attrs key keep their names, and the `pack` docstring now says the attr name is historical instead of silently mixing old and new.
- All five string sites named in #3086 are covered, and the strings live in shared `_attrs.py` helpers, so eager, dask, and GPU reads emit identical wording.

### Checklist
- [x] Docstrings present and accurate (after the `_pack` docstring suggestion)
- [x] Tests cover the changed behaviour, including the dask graph-build rejection path (existing tests) and both malformed-metadata variants (new tests)
- [x] Temporary test files are uniquely named (`t3086_*`)
- [x] No algorithm, backend, NaN, or performance surface touched -- text-only change
- [x] README / feature matrix: not applicable, no API change

brendancol (commented):
Follow-up review after 75d6e186.

Both findings from the first pass are resolved:

- `_pack`'s docstring (`xrspatial/geotiff/_attrs.py:1659-1676`) now matches the error it raises: "Re-pack an ``unpack=True`` read", "reopening with ``unpack=True``", "carries no unpack state". The remaining `mask_and_scale` mentions in `_attrs.py` (module docstring, `_extract_scale_offset` docstring, line 1844) all describe the internal `mask_and_scale` kwarg that the private read pipeline still uses by design, so they are correct as written.
- The malformed-metadata test is parametrized over the two fixture builders (`bad_scale_value` / `bad_xml`) instead of running the assertions twice inline.

60 tests pass across the two touched test files. Nothing further from me.

Issue #3100, opened 2026-06-09
_reproject_streaming crashes on 3-D (multi-band) sources


brendancol:
The streaming fallback in `reproject()` allocates its output as a 2-D buffer (`result = np.full(out_shape, nodata, dtype=np.float64)`), but the per-tile worker (`_reproject_chunk_numpy`) returns `(rows, cols, bands)` tiles for 3-D sources. Assembly then fails:

```
ValueError: could not broadcast input array from shape (16,16,3) into shape (16,16)
```

Both assembly loops in `_reproject_streaming` (the dask.bag branch and the local ThreadPoolExecutor branch) have the same problem.

**Repro** (direct call, since the end-to-end trigger needs a >512 MB source and no dask):

```python
import importlib
import numpy as np
import xarray as xr
rp = importlib.import_module('xrspatial.reproject')
from xrspatial.reproject._crs_utils import _resolve_crs
from xrspatial.reproject._grid import _compute_output_grid

raster = xr.DataArray(
    np.random.rand(32, 32, 3), dims=['y', 'x', 'band'],
    coords={'y': np.linspace(55, 45, 32), 'x': np.linspace(-5, 5, 32), 'band': [1, 2, 3]},
    attrs={'crs': 'EPSG:4326', 'nodata': np.nan},
)
src_crs = _resolve_crs('EPSG:4326')
tgt_crs = _resolve_crs('EPSG:32633')
src_bounds = rp._source_bounds(raster)
grid = _compute_output_grid(src_bounds, (32, 32), src_crs, tgt_crs)
rp._reproject_streaming(
    raster, src_bounds, (32, 32), True,
    src_crs.to_wkt(), tgt_crs.to_wkt(),
    grid['bounds'], grid['shape'],
    'bilinear', float('nan'), 16,
    16, rp._parse_max_memory('1GB'),
    x_desc=False, band_nodata=None,
)
```

To hit it end to end: reproject a multi-band raster over 512 MB in an environment without dask. `reproject()` sets `_use_streaming = True` and hands the 3-D source straight to `_reproject_streaming`.

Expected: a `(out_h, out_w, bands)` result, matching the in-memory and dask paths. The band-axis handling those paths got in #2027 never reached the streaming fallback.

Found while adding test coverage for the streaming path. A strict xfail in that test PR pins this crash until it's fixed.

PR #3118, opened 2026-06-09
Add tests for the streaming reproject fallback path (#3101)

### LINKED ISSUE #3101: Streaming reproject fallback has zero test coverage
`reproject()` has five execution paths: in-memory numpy, in-memory cupy, dask, dask+cupy, and a streaming fallback that kicks in when the source is over 512 MB and dask can't be imported. The first four are covered in test_reproject.py. Nothing runs the streaming one: `_reproject_streaming`, `_process_tile_batch`, and `_parse_max_memory` (`xrspatial/reproject/__init__.py`) never execute under pytest.

The path has its own tile-job construction, a ThreadPoolExecutor batching loop with a serial fallback, mosaic assembly from `(row_offset, col_offset, tile)` triples, and the max_memory parser. An off-by-one in the assembly offsets, or a parser regression, would not turn any test red.

The existing max_memory tests (`test_reproject_max_memory_string_arg` / `test_reproject_max_memory_int_arg`) push a 32x32 raster through `reproject()`, which stays on the in-memory path. `_parse_max_memory` only runs when the streaming branch is selected, so those tests never reach it.

Proposed tests (test-only):

- Call `_reproject_streaming` directly on a small raster and compare against the in-memory numpy output for the same grid: multi-tile threaded run, serial run (max_memory small enough to disable the thread pool), single-tile run.
- `_parse_max_memory` unit cases: None default, KB/MB/GB/TB suffixes, int passthrough, bare numeric string.
- A strict xfail pinning the 3-D crash tracked in #3100.

I checked locally: the 2-D streaming output matches the in-memory result, so this is a coverage gap rather than a behaviour bug.

brendancol:
Closes #3101.

- `reproject()`'s streaming fallback (`_reproject_streaming`, `_process_tile_batch`, `_parse_max_memory`) only runs when the source is over 512 MB and dask is not importable, so nothing in CI ever executed it. The new test file calls the helpers directly on small rasters and checks the output against the in-memory numpy path on the same grid: multi-tile threaded, serial (`max_memory=1` byte), single tile, and nearest resampling with NaN cells. `_parse_max_memory` gets 10 unit cases (int, float, KB/MB/GB/TB suffixes, lowercase with whitespace, bare numeric string, None default).
- The 3-D source crash in the streaming assembly is pinned with a strict xfail referencing #3100. The source fix stays with that issue; this PR is test-only.
- Backend note: the streaming path is CPU/numpy-only, so no cupy or dask variants apply.

Test plan:
- [x] `pytest xrspatial/tests/test_reproject_streaming_3101.py` -> 14 passed, 1 xfailed
- [x] No source changes (the only non-test diff is the sweep state CSV)


brendancol (commented):
## PR Review: Add tests for the streaming reproject fallback path (#3101)

None.

- [ ] `xrspatial/tests/test_reproject_streaming_3101.py:101-123` -- the threaded vs serial split depends on the budget arithmetic inside `_process_tile_batch` (`max_concurrent >= 2 and len(batch) > 1`). If the `tile_mem` formula in `_reproject_streaming` changes, both tests could silently run the same branch while parity still passes. A cheap guard: assert the computed concurrency in each test (budget // tile_mem >= 2 for the threaded case, == 1 for the serial case) so a formula change that re-routes the branch is visible.
- [ ] `xrspatial/tests/test_reproject_streaming_3101.py:140-152` -- when #3100 is fixed, the strict xfail flips and forces a revisit, which is the point. The post-fix assertion only checks `ndim`/`shape[2]`, though. Compare against the in-memory 3-D `reproject()` output so the un-xfailed test has real teeth from day one.

- [ ] The dask.bag distributed branch of `_reproject_streaming` (`xrspatial/reproject/__init__.py:1592-1619`) stays uncovered; it needs a live dask.distributed client. The sweep CSV records it as a carried LOW, fine to leave -- just noting that "streaming path covered" excludes that branch.

### What looks good
- Parity tests compare against the public `reproject()` output rather than stored fixtures, so the tile assembly is checked against the contract that matters.
- `_parse_max_memory` cases cover every suffix plus int, float, bare numeric string, and the None default.
- The 3-D crash is pinned strict with `raises=ValueError` and links the source issue; source stays untouched.

### Checklist
- [x] Algorithm matches reference -- N/A, test-only PR
- [x] Backend consistency -- streaming path is CPU-only by design; no GPU variants apply
- [x] NaN handling -- covered by the nearest-resampling NaN-block test
- [x] Edge cases -- single-tile, multi-tile, serial, 3-D strict xfail
- [x] Dask chunk boundaries -- N/A
- [x] No premature materialization -- N/A (tests)
- [x] Benchmark -- not needed for tests
- [x] README feature matrix -- not applicable
- [x] Docstrings -- module docstring explains why the helpers are called directly

brendancol (commented):
## Follow-up review (after d07dd09)

Both suggestions from the first pass are addressed:

- `_max_concurrent` mirrors the `tile_mem` arithmetic from `_reproject_streaming` / `_process_tile_batch`, and the threaded/serial tests now assert which batching branch they engage (`>= 2` vs `== 1`), plus a multi-tile shape check on the threaded run. A formula change upstream now fails loudly instead of collapsing both tests onto one branch. Fixed.
- `test_3d_source_streams` compares against the in-memory 3-D `reproject()` output, so the assertion is full parity once #3100 lands and the strict xfail comes off. Fixed.

Remaining open item is the nit about the dask.bag distributed branch (`xrspatial/reproject/__init__.py:1592-1619`), deferred with reason: it needs a live dask.distributed client fixture, and it is recorded as a carried LOW in the sweep state CSV.

File re-ran clean: 14 passed, 1 xfailed. No blockers.

brendancol:
CI note: the first run's macos-3.14 job died with SIGABRT during test_multi_tile_threaded -- two pool threads entered the @njit(parallel=True) projection kernels at once, and numba's workqueue threading layer aborts on concurrent entry. That is a real bug in the streaming path itself (the production ThreadPoolExecutor does the same thing); filed as #3141.

Commit 6432929 keeps the coverage but avoids the unsupported pattern: the threaded parity test runs with transform_precision=0 (exact pyproj, one Transformer per thread), and the multi-tile NaN test plus the 3-D xfail are pinned to the serial loop so the numba fast path is still exercised, just never concurrently. The windows-3.14 "failure" was fail-fast cancellation; its suite had fully passed.

PR #3119, opened 2026-06-09
Gate CUDA reproject fast path on WGS84-compatible datums (#3094)

### LINKED ISSUE #3094: reproject: cupy backend skips datum handling for non-WGS84 datums (try_cuda_transform missing the #2651 guard)
**Describe the bug**

#2651 disabled the numba fast path when either CRS uses a non-WGS84 datum: the projection kernels run in WGS84, so skipping the datum shift returned corrupt coordinates. That fix gated `try_numba_transform` and `transform_points` in `_projections.py`. `try_cuda_transform` in `_projections_cuda.py` never got the same guard.

The CUDA dispatcher matches projected CRSs through `_tmerc_params` / `_lcc_params` / etc., and those accept any datum present in the Helmert table via `_is_wgs84_compatible_ellipsoid` (OSGB36/airy, DHDN/potsdam, MGI, ED50, NAD27, ...). So a cupy or dask+cupy reproject between EPSG:4326 and EPSG:27700 runs the WGS84 Krueger series with no datum shift and no Airy ellipsoid handling.

Measured on a GPU host:

- `try_cuda_transform(EPSG:4326 -> EPSG:27700)` coordinates differ from pyproj by up to ~0.0014 deg (~96 m).
- End to end, `reproject(raster, 'EPSG:27700')` on the cupy backend differs from the numpy backend by up to 0.094 on data in [0, 1] (the sampler reads roughly 10 pixels away from the right spot), and 5 cells flip between NaN and finite on a 64x64 input.

numpy and dask+numpy are unaffected because `try_numba_transform` already bails to pyproj for these pairs.

**Expected behavior**

cupy and dask+cupy should match numpy. When either side uses a non-WGS84 datum, `try_cuda_transform` should return None so the chunk worker falls back to the CPU path (which then defers to pyproj), the same way #2651 handled the CPU kernels.

**Additional context**

Found by the accuracy sweep. Affected pairs: geographic WGS84/NAD83 to or from any projected CRS whose proj4 dict carries a datum/ellps key in `_DATUM_PARAMS`. EPSG:27700 confirmed live; the DHDN, MGI, ED50, and NAD27 entries hit the same dispatch.

brendancol:
Closes #3094

- Adds the non-WGS84 datum guard from #2651 to `try_cuda_transform`. The CUDA dispatcher matched projected CRSs whose datum sits in the Helmert table (OSGB36, DHDN, MGI, ED50, NAD27), then projected them with WGS84 constants and no datum shift, putting cupy and dask+cupy results ~100 m off while numpy went through pyproj. With the guard, those pairs return None and the chunk worker falls back to the CPU path.
- Adds `TestNonWgsDatumCudaFastPath`: bail-out checks for EPSG:27700 (both directions) and NAD27, a check that the WGS84 UTM fast path stays active, and an end-to-end numpy-vs-cupy parity test for a 4326 -> 27700 reproject (NaN masks equal, values within 1e-4).
- Also carries the accuracy-sweep state CSV update for the reproject module.

Backend coverage: numpy and dask+numpy were already correct; this fixes cupy and dask+cupy.

Test plan:
- [x] `pytest xrspatial/tests/test_reproject.py -k NonWgsDatum` (10 passed, CPU + CUDA classes, run on a GPU host)
- [x] `pytest xrspatial/tests/test_reproject.py -k "cupy or cuda or CuPy"` (48 passed on a GPU host)
- [x] Verified pre-fix divergence: `try_cuda_transform` coordinates off by ~96 m vs pyproj; end-to-end cupy vs numpy max diff 0.094 on [0,1] data


brendancol (commented):
## PR Review: Gate CUDA reproject fast path on WGS84-compatible datums (#3094)

None found.

None.

- `_get_datum_params` runs `crs.to_dict()` (which goes through pyproj's `to_proj4`) twice per chunk now that the guard sits in `try_cuda_transform` (`xrspatial/reproject/_projections_cuda.py:751`). The CPU fast path pays the same per-chunk cost in `try_numba_transform`, so this is parity rather than a regression. If chunk-loop profiling ever shows it, a small memo cache keyed on the CRS would fix both paths at once. Not needed for this PR.

### What looks good
- The guard is a faithful mirror of the #2651 CPU fix and sits before the cupy array allocation, so bailed pairs don't pay for device buffers they never use.
- Root cause is correctly identified: the projected-CRS matchers (`_tmerc_params`, `_lcc_params`, ...) accept any datum in the Helmert table via `_is_wgs84_compatible_ellipsoid`, so EPSG:27700 reached the WGS84 Krueger kernels. The guard cuts that off for every affected datum (OSGB36, DHDN, MGI, ED50, NAD27, ...) rather than special-casing one code.
- Test coverage is the right shape: both directions for 27700, the NAD27 geographic case, a WGS84-UTM check that proves the fast path is not over-blocked, and an end-to-end numpy-vs-cupy parity test with exact NaN-mask agreement. The parity test would have caught the original bug (~0.094 value error on [0,1] data).
- Post-guard, the cupy worker falls back to the CPU coordinate path, which already defers these pairs to pyproj, so all four backends now agree by construction.

### Checklist
- [x] Algorithm matches reference (#2651 CPU guard)
- [x] All implemented backends produce consistent results (numpy-vs-cupy parity test, run on a GPU host)
- [x] NaN handling is correct (mask equality asserted)
- [x] Edge cases are covered by tests (both directions, geographic + projected non-WGS84 sides, fast-path-still-active)
- [x] Dask chunk boundaries handled correctly (guard is per-chunk, no shape changes)
- [x] No premature materialization or unnecessary copies (guard placed before allocation)
- [x] Benchmark not needed (correctness fix; affected pairs previously returned wrong data)
- [x] README feature matrix unchanged (no API or backend-support change)
- [x] Docstrings/comments explain the why, with issue references

Issue #3102, opened 2026-06-09
rasterize(all_touched=True) has no effect on LineString geometries


brendancol:
**Describe the bug**

`all_touched=True` only changes how polygons are burned. In `_run_numpy` (rasterize.py, step 1b) the supercover traversal is applied to polygon boundary segments only; lines always go through the Bresenham burner (`_burn_lines_cpu`, step 2). The cupy and dask tile paths are laid out the same way. So for LineString / MultiLineString input, `rasterize(..., all_touched=True)` returns exactly the same pixels as `all_touched=False`.

rasterio's `features.rasterize(..., all_touched=True)` does burn every pixel a line crosses. The `all_touched` docstring here claims pixel-for-pixel parity with rasterio, so a diagonal line comes back with gaps that rasterio would fill.

```python
import numpy as np
from shapely.geometry import LineString
from xrspatial.rasterize import rasterize
import rasterio.features
from rasterio.transform import from_bounds

line = LineString([(0.2, 0.2), (4.8, 4.8)])
xs = rasterize([(line, 1.0)], width=5, height=5, bounds=(0, 0, 5, 5),
               fill=0, all_touched=True)
rio = rasterio.features.rasterize(
    [(line, 1)], out_shape=(5, 5),
    transform=from_bounds(0, 0, 5, 5, 5, 5),
    fill=0, all_touched=True)
print(xs.data.astype(int))  # 5 pixels, identical to all_touched=False
print(rio)                  # 9 pixels (supercover)
```

Verified on numpy, cupy, dask+numpy, and dask+cupy: all four return the Bresenham result.

**Expected behavior**

Lines should participate in `all_touched` the way polygons do. The supercover machinery already exists (`_burn_lines_supercover_cpu` and its GPU twins are used for polygon boundaries), so the fix is plumbing line segments through it when `all_touched=True`. If instead the intended contract is "all_touched affects polygons only", the docstring should say so.

**Additional context**

Found by the test-coverage sweep (2026-06-09) while auditing the `all_touched` x geometry-type interaction, which had no tests. A test-only PR pins the current no-op behavior on all four backends so a fix here will show up as a test update.

Issue #3105, opened 2026-06-09
rasterize: no tests for all_touched with line input or the non-iterable input error path


brendancol:
Test-coverage sweep pass 5 over `xrspatial/rasterize.py` (2026-06-09). Two parameter-coverage gaps, both MEDIUM.

1. `all_touched=True` is tested only with polygon input (rasterio parity in `test_rasterize_all_touched_supercover_2169.py`, monotonicity and thin-polygon cases in `test_rasterize_accuracy.py`, GPU race tests in `test_rasterize_gpu_race_2167.py`). No test on any backend passes a LineString with `all_touched=True`. The flag is currently a no-op for lines, which diverges from the rasterio behavior the docstring promises; that is filed separately as #3102. Whichever way #3102 is resolved, the interaction needs pinning so a behavior change is visible in CI.

2. `_parse_input` raises `TypeError("geometries must be a GeoDataFrame or iterable of (geometry, value) pairs")` for non-iterable input (rasterize.py:2834-2837). Nothing exercises it. The other validation guards in rasterize() were pinned in passes 1 through 4; this one was missed.

Plan: test-only PR. Pin lines + all_touched to the current contract (output equals `all_touched=False` and equals rasterio's default mode) across numpy / cupy / dask+numpy / dask+cupy, add a strict xfail for rasterio all_touched parity linked to #3102, and pin the TypeError path.

PR #3120, opened 2026-06-09
Add rasterize all_touched-line and non-iterable-input coverage tests

brendancol:
Test-only PR from the test-coverage sweep (pass 5 over `xrspatial/rasterize.py`).

## Proposed Changes

- Pins `all_touched=True` as a no-op for LineString input on numpy, cupy, dask+numpy, and dask+cupy, parametrized over a corner-crossing diagonal and a shallow interior-crossing line, plus equality with rasterio's default-mode burn. A strict xfail asserts rasterio all_touched parity and flips when #3102 (lines skip the supercover burn) is fixed, so the behavior change cannot land silently.
- Pins the non-iterable `geometries` TypeError in `_parse_input` over int / float / None / object inputs.
- Merged latest `main` (rasterize source fixes, reproject improvements, geotiff updates); resolved conflict in `.claude/sweep-test-coverage-state.csv` by keeping the PR's rasterize Pass 5 row and taking main's updated reproject row.

No source changes beyond the merge. The line/all_touched divergence itself is tracked in #3102.

Backend coverage: the no-op pin runs on all four backends; cupy and dask+cupy executed on a CUDA host (not skipped). Tests: 10 passed, 6 skipped.


brendancol (commented):
## PR Review: Add rasterize all_touched-line and non-iterable-input coverage tests

Test-only PR; reviewed the new file in full plus the surrounding rasterize source and the four prior coverage files.

None.

- [ ] test_rasterize_coverage_2026_06_09.py:96-99: `_DIAG` crosses pixel corners, which is exactly where rasterization tie-breaking lives. The pins hold today, but a second line geometry that crosses cell interiors (not corners) would make the no-op pin less sensitive to tie-break changes that are unrelated to #3102. Cheap to add if a follow-up touches this file; not worth a new commit on its own.

- [ ] test_rasterize_coverage_2026_06_09.py:185: a string input is iterable, so `rasterize("abc")` takes a different failure path (`item[0]` on a 1-char string). The new tests cover the documented non-iterable guard only; if anyone wants the string case pinned it belongs with #3105's LOW leftovers, not this PR.

### What looks good
- The no-op pin runs on all four backends with ragged dask chunks ((3,2),(3,2) over 5x5), so the tile path crosses the diagonal.
- The strict xfail referencing #3102 plus the equality pins mean the eventual fix cannot land without touching this file, which is the point of a pin.
- Bresenham expectation is a hand-checkable known value (anti-diagonal), not a snapshot.
- rasterio gating matches the convention in test_rasterize_all_touched_supercover_2169.py.
- State CSV row is a single physical line, last-write-wins, prior pass notes preserved.

### Checklist
- [x] Algorithm matches reference (rasterio default mode verified directly)
- [x] All implemented backends produce consistent results (cupy and dask+cupy executed on a CUDA host)
- [x] NaN handling: n/a for these paths
- [x] Edge cases covered (corner-crossing diagonal, MultiLineString explode, ragged chunks)
- [x] No premature materialization (single compute per assertion)
- [x] README feature matrix: n/a
- [x] Docstrings present on every test class and the module header

brendancol (commented):
## Follow-up review (after 96d0b653)

- Suggestion (interior-crossing line): fixed. The no-op pin is now parametrized over a corner-crossing diagonal and a shallow interior-crossing line on all four backends (16 cells total; 15 passed + 1 strict xfail on a CUDA host).
- Nit (string input): dismissed. A string is iterable, so it takes the pair-unpacking path and dies with an incidental IndexError. Pinning that would lock in accidental behavior for garbage input rather than a contract; left with the LOW leftovers on #3105.

No new findings. The diff since the first review is the parametrize change plus the `_INTERIOR` constant and comment.

brendancol:

PR #3123, opened 2026-06-09
Skip duplicate try_numba_transform probe in the numpy reproject chunk worker

### LINKED ISSUE #3106: reproject: numpy chunk worker runs try_numba_transform twice per chunk on pyproj-fallback CRS pairs
## Problem

`_reproject_chunk_numpy` (xrspatial/reproject/__init__.py) calls `try_numba_transform` to check for a numba fast path. When the CRS pair has no fast path the call returns None and the worker falls back to `_transform_coords`, which calls `try_numba_transform` again before building the pyproj control grid. Both calls do the same work and both return None.

Each wasted call re-parses CRS parameters (around ten pyproj `to_dict()` / `to_authority()` round-trips across the per-projection param extractors) and allocates four chunk-sized float64 arrays (`np.tile`, `np.repeat`, two `np.empty`) before concluding there is no fast path.

Measured on a 512x512 chunk, EPSG:4326 -> ESRI:54009 (Mollweide, no fast path):

- wasted `try_numba_transform` call: ~0.3-0.5 ms
- `_transform_coords` with the retry: 1.57 ms; without: 1.06 ms
- full chunk worker: 5.3 ms, of which ~11% is the duplicated call

The waste repeats for every output chunk on the dask+numpy path, and merge's per-block adapter routes through the same worker.

## Fix

In `_reproject_chunk_numpy`, pass `src_crs=None, tgt_crs=None` to `_transform_coords`. The inner retry is gated on both being non-None, so this skips it. The cupy workers (`_reproject_chunk_cupy`, `_reproject_dask_cupy` CPU fallback) must keep passing the CRS objects: they only tried the CUDA path, so the retry inside `_transform_coords` is their first and only numba attempt.

Found by the performance sweep audit of xrspatial/reproject. OOM verdict for the module: SAFE; bottleneck: compute-bound.

brendancol:
Closes #3106

- `_reproject_chunk_numpy` probed the numba fast path, then `_transform_coords` probed it again before building the pyproj control grid. The worker now passes no CRS objects to `_transform_coords`, which skips the inner retry (it is gated on both CRSes being non-None). Saves one CRS-param parse (~10 pyproj `to_dict()`/`to_authority()` round-trips) and four chunk-sized float64 allocations per output chunk on pyproj-fallback CRS pairs, measured at ~11% of the chunk worker for a 512x512 chunk (4326 -> Mollweide).
- The cupy CPU fallbacks (`_reproject_chunk_cupy`, `_reproject_dask_cupy`) are unchanged: they only try the CUDA path first, so the probe inside `_transform_coords` is their first and only numba attempt.
- Also commits the performance-sweep state row for the reproject pass that found this (OOM verdict SAFE, compute-bound).

Backends: affects numpy and dask+numpy (plus merge's per-block adapter, which routes through the same worker). cupy and dask+cupy behavior unchanged; validated on a CUDA host (numpy/cupy max abs diff 2e-12, on-device transform confirmed).

Test plan:
- [x] New `TestNoDuplicateNumbaFastPathProbe`: spy asserts exactly one `try_numba_transform` call per chunk on a no-fast-path pair; `_transform_coords` keeps its own probe when handed CRS objects; pyproj-reference value parity for the fallback pair
- [x] Full reproject suite: 447 passed (test_reproject.py, coverage, cupy gate)


brendancol (commented):
## PR Review: Skip duplicate try_numba_transform probe in the numpy reproject chunk worker

- None.

- None. The change is two lines plus a comment, and the gating logic in `_transform_coords` (xrspatial/reproject/__init__.py:284) is unchanged for callers that pass CRS objects.

- [ ] The state CSV rewrite reorders the `interpolate_spline` / `interpolate-kriging` rows. This is just the canonical writer's `sorted()` ordering (`-` sorts before `_`), not data loss, but it makes the diff look bigger than it is.
- [ ] `test_fallback_pair_values_match_pyproj_reference` exercises `_transform_coords` directly rather than the chunk worker, so it overlaps a little with `TestExactPrecisionEscapeHatch`. Fine to keep since it pins the no-CRS-kwargs call form the worker now uses.

### Verification performed
- Reverted `xrspatial/reproject/__init__.py` to the pre-fix state and re-ran the new test class: `test_chunk_numpy_probes_fast_path_exactly_once` fails (probe count 2), the other two pass. With the fix it passes. The regression test is genuinely red on the old code.
- Full reproject suite (test_reproject.py + coverage + cupy gate): 447 passed.

### What looks good
- The fix targets only `_reproject_chunk_numpy`; the cupy CPU fallbacks (`_reproject_chunk_cupy` line ~571, `_reproject_dask_cupy` line ~1790) still pass CRS objects, and the new `test_transform_coords_still_probes_when_given_crs` pins that contract so a future cleanup can't silently remove their only numba probe.
- `transform_precision=0` semantics are untouched: both the worker's probe and the inner retry were already gated off for the exact path, and the existing `TestExactPrecisionEscapeHatch` spies still pass.
- WKT construction moved into a helper so collection does not touch pyproj when it is not installed.

### Checklist
- [x] Algorithm matches reference (identical inputs to both probes; skipping the second cannot change results)
- [x] All implemented backends produce consistent results (numpy/dask+numpy affected; cupy paths unchanged)
- [x] Edge cases covered (probe-count spy, CRS-passing contract, pyproj value parity)
- [x] Dask chunk boundaries handled correctly (per-chunk worker logic unchanged)
- [x] No premature materialization or unnecessary copies (removes allocations)
- [x] Benchmark exists (benchmarks/benchmarks/reproject.py covers reproject)
- [x] Docstrings: no public API change

PR #3124, opened 2026-06-09
Sort dask KDTree targets row-major so the tie-break matches numpy (#3090)

### LINKED ISSUE #3090: Dask allocation/direction ignore the documented tie-break when targets span multiple chunk columns
**Describe the bug**

`allocation()` and `direction()` promise a deterministic tie-break: when two targets are equidistant from a pixel, the one with the lowest flat (row-major) index wins, on every backend. The dask+numpy KDTree path breaks that promise whenever the raster has more than one chunk column. dask+cupy with unbounded `max_distance` converts to the same path, so it is affected too.

The cause is in `_collect_region_targets` (xrspatial/proximity.py). It concatenates per-chunk target lists in chunk order (chunk row outer, chunk column inner), so the KDTree's target order is not global row-major. `_kdtree_query_lowest_index` then resolves ties by lowest tree index, which is the wrong target. Its own docstring assumes "Target coordinates are stored in row-major (flat-index) order" - that only holds when there is a single chunk column.

The existing tie-break tests (`test_allocation_tie_break_lowest_flat_index`, `test_allocation_tie_break_tiled_path`) put both targets in the same raster row, where chunk order happens to agree with row-major order, so they pass.

**Repro**

```python
import dask.array as da
from xrspatial import allocation

data = np.array([
    [0., 0., 0., 0., 0.],
    [0., 1., 0., 2., 0.],
    [0., 0., 3., 0., 0.],
    [0., 0., 0., 0., 0.],
    [0., 0., 0., 0., 0.]])

def make(backing):
    r = xr.DataArray(backing, dims=['y', 'x'])
    r['y'] = np.arange(5)[::-1].astype(float)
    r['x'] = np.arange(5).astype(float)
    return r

ref = allocation(make(data), target_values=[1, 2, 3], max_distance=10.0)
out = allocation(make(da.from_array(data, chunks=(5, 3))),
                 target_values=[1, 2, 3], max_distance=10.0)
print(ref.data[2, 3], out.data.compute()[2, 3])  # 2.0 vs 3.0
```

Pixel (2, 3) is at distance 1 from target 2 (flat index 8) and from target 3 (flat index 12). The documented policy picks 2; numpy, cupy, and the bounded map_overlap paths all return 2. The dask KDTree path returns 3 because target 3 lives in chunk (0, 0) and gets collected before target 2 in chunk (0, 1). Same divergence in `direction()` at the tied pixels. Chunking only along rows, or a single chunk, matches numpy.

**Expected behavior**

The dask KDTree paths (global and tiled) resolve ties to the target with the lowest global flat index, matching the other backends and the docstrings of `allocation`/`direction`.

**Additional context**

Found by the api-consistency sweep against the proximity module. Fix idea: have `_collect_region_targets` track each target's global flat index and sort the concatenated coords/values by it before building the tree. That fixes both `_build_global_kdtree` and `_tiled_chunk_query`, since both collect through the same helper.

brendancol:
Closes #3090

- `_collect_region_targets` now tracks each target's global flat (row-major) pixel index and sorts the concatenated coords/values by it before the KDTree is built. The chunk-major collection order was leaking into the tree, so `_kdtree_query_lowest_index`'s lowest-tree-index tie-break picked the wrong target whenever the raster had more than one chunk column.
- `_stream_target_counts` caches the flat indices alongside coords/values (same 25% budget, accounted in `entry_bytes`), and the cache is threaded through both the global and tiled KDTree paths.
- No change to proximity results (tied targets have equal distance); only `allocation` and `direction` values at tied pixels change, and only to the documented winner.

Backend coverage: the bug lives in the dask+numpy KDTree path; dask+cupy with unbounded `max_distance` converts to that path and is fixed by the same change. numpy, cupy, and the bounded map_overlap paths already followed the documented policy and are untouched.

Test plan:
- [x] New regression tests with diagonal tied targets across two chunk columns: `test_allocation_tie_break_chunk_column_order`, `test_direction_tie_break_chunk_column_order` (dask+numpy and dask+cupy), and `test_allocation_tie_break_tiled_path_chunk_column_order` for the eager tiled fallback (which also exercises the uncached recompute branch via the tiny-budget patch).
- [x] All 5 new tests fail on the unfixed code, pass with the fix.
- [x] Full `xrspatial/tests/test_proximity.py` passes (415 tests) on a CUDA host.


brendancol (commented):
## PR Review: Sort dask KDTree targets row-major so the tie-break matches numpy (#3090)

### Blockers
None.

### Suggestions
- [ ] `_kdtree_query_lowest_index` (xrspatial/proximity.py:809) still asserts "Target coordinates are stored in row-major (flat-index) order" as if it were a property of the caller's iteration order. That assumption is exactly what this PR found to be false; the order is now an invariant enforced by the sort in `_collect_region_targets`. Add a cross-reference there so the next reader knows where the invariant comes from and does not "optimize away" the sort.

### Nits
- [ ] The flat-index expression `(y_offsets[iy] + rows) * width + (x_offsets[ix] + cols)` appears twice (`_stream_target_counts` and the uncached branch of `_collect_region_targets`). A small helper would keep the two from drifting apart.

### What looks good
- The regression geometry is the right one: targets in different rows AND different chunk columns, which the existing same-row tie tests could not catch. All five new tests fail on the unfixed code.
- Coverage spans both collection branches: the chunk-column tests hit the cached path, the tiled test's 1-byte budget forces the uncached recompute path.
- `flat.nbytes` is included in the cache budget accounting, so the 25% memory bound still holds.
- int64 flat indices cannot overflow for any realistic raster, and `argsort` over unique keys is deterministic, so no stable-sort flag is needed.
- `proximity` results are unaffected (tied targets are equidistant by definition); only the tied `allocation`/`direction` pixels change, and only to the documented winner.

### Checklist
- [x] Algorithm matches the documented tie-break policy (lowest global flat index)
- [x] All implemented backends produce consistent results (numpy reference pinned in tests; dask+numpy, dask+cupy, tiled fallback verified)
- [x] NaN handling unchanged (no-target and out-of-range paths untouched)
- [x] Edge cases covered (single chunk, row-only chunks, 1x1 chunks, tiled fallback)
- [x] Dask chunk boundaries handled correctly (fix is in target collection, not chunk math)
- [x] No premature materialization (same `.compute()` pattern as before; sort is O(T log T) on the driver)
- [x] Benchmark exists (benchmarks/benchmarks/proximity.py; sort cost is negligible next to tree build)
- [x] README feature matrix not applicable (no API or backend change)
- [x] Docstrings updated where behavior context changed (`_collect_region_targets`, `_stream_target_counts`)

brendancol (commented):
Follow-up review after fc96c806:

- Suggestion (stale invariant wording in `_kdtree_query_lowest_index`): fixed. The docstring now says the row-major order is produced by the sort in `_collect_region_targets` and points at #3090.
- Nit (duplicated flat-index expression): fixed. Both call sites now use the new `_global_flat_indices` helper.

Test suite re-run after the changes: 415 passed. The two flake8 E127 warnings in test_proximity.py (lines 1134, 1160) predate this PR and sit in helpers the diff does not touch, so they were left alone.

No remaining findings.

Issue #3083, opened 2026-06-09
Lint cleanup in reproject/__init__.py: unused import, dead local, isort drift


brendancol:
flake8 and isort (run with the repo's setup.cfg config: max-line-length=100, line_length=100) flag three things in `xrspatial/reproject/__init__.py`:

- F401: `_merge_arrays_cupy` is imported from `._merge` at line 38 but never used. The function itself lives in `_merge.py` and stays; only the import is dead.
- F841: `_use_native_cuda` is assigned at lines 563 and 602 and never read. Leftover from #2620, which made the cupy chunk path always go through the native CUDA kernels, so nothing reads the flag anymore.
- isort: the top-of-file import blocks and the function-local `from ._vertical import _load_geoid, _interp_geoid_2d` (around line 1278) don't match the configured ordering.

Lint cleanup only, no behaviour change.

Found by the style sweep.

PR #3125, opened 2026-06-09
Accept cupy and dask+cupy inputs in reproject.merge() (#3095)

### LINKED ISSUE #3095: reproject.merge() raises TypeError on cupy-backed inputs
`reproject()` documents and supports all four backends (numpy, cupy, dask+numpy, dask+cupy). Its sibling `merge()` crashes on cupy-backed inputs:

```python
from xrspatial.reproject import merge

data = np.random.rand(32, 32).astype(np.float32)
coords = {'y': np.linspace(50, 45, 32), 'x': np.linspace(-5, 0, 32)}
r = xr.DataArray(cp.asarray(data), dims=['y', 'x'], coords=coords,
                 attrs={'crs': 'EPSG:4326'})
merge([r, r])
```

```
TypeError: Implicit conversion to a NumPy array is not allowed.
Please use `.get()` to construct a NumPy array explicitly.
```

The crash comes from `_merge_inmemory` calling `info['raster'].values` (xrspatial/reproject/__init__.py:2572), which xarray refuses for cupy data. Dask-of-cupy inputs fail the same way at compute time, since `_merge_block_adapter` routes every tile through `_reproject_chunk_numpy` and `np.asarray`.

Two API-consistency problems here:

1. A user who reprojects on GPU and then mosaics hits a low-level xarray TypeError instead of either a working merge or a clear "not supported" message. `merge()`'s docstring says nothing about backends, while `reproject()` advertises all four.
2. `_merge.py` contains a complete `_merge_arrays_cupy` implementation that is imported in `__init__.py` but never called from anywhere (the unused import is also flagged by the lint sweep in #3083). The GPU merge backend was plumbed in and never wired up.

Suggested fix: accept cupy and dask+cupy inputs in `merge()` by moving them to the host on entry (the same round-trip `_apply_vertical_shift` already uses for its CPU-only geoid lookup) and returning the mosaic on GPU when the inputs were on GPU. Document the backend behaviour in the merge docstring. Native GPU merge via `_merge_arrays_cupy` can come later if profiling justifies it.

Found by the API consistency sweep (cupy smoke test of public entry points).

brendancol:
Closes #3095

`merge()` crashed with xarray's "Implicit conversion to a NumPy array is not allowed" TypeError on cupy-backed inputs, while `reproject()` handles all four backends. The README matrix already listed Merge as CPU-fallback on GPU backends; this makes that true.

- `_merge_inputs_to_host()` converts cupy inputs to host copies on entry (dask+cupy stays lazy via `map_blocks(cp.asnumpy)`), the merge runs through the existing numpy pipeline, and the mosaic moves back to GPU at the end (`cp.asarray` eager, `map_blocks(cp.asarray)` lazy). Same host round-trip `_apply_vertical_shift` already uses for the geoid lookup.
- merge docstring now states the backend handling.
- One commit also carries the api-consistency sweep state CSV update for the reproject module.

Backend coverage: numpy and dask+numpy unchanged; cupy and dask+cupy now work (CPU execution, GPU result). Mixed numpy+cupy input lists return a cupy mosaic.

Not done here: native GPU merge via the existing-but-unwired `_merge_arrays_cupy`. The unused import it left behind is tracked in #3083.

Test plan:
- [x] New `TestMergeCupyBackends`: cupy round-trip, mixed inputs, source non-mutation, dask+cupy laziness (`_meta` is cupy, no compute at graph build), and exact value parity vs the numpy path for all five strategies.
- [x] Existing merge suites pass (34 tests).
- [x] Verified on a CUDA host.


brendancol (commented):
## PR Review: Accept cupy and dask+cupy inputs in reproject.merge() (#3095)

### Blockers
None.

### Suggestions
- [ ] `_merge_inputs_to_host` (xrspatial/reproject/__init__.py, the `elif is_dask_cupy(r)` branch) calls `utils.is_dask_cupy()`, which does `isinstance(agg.data, da.Array)` with no guard. `utils.py` sets `da = None` when dask is not installed, so on a host with cupy but no dask, merging a list that contains any non-cupy raster raises `AttributeError: 'NoneType' object has no attribute 'Array'` instead of merging. The rest of `merge()` is careful about this (`if has_dask_array(): import dask.array as _da`). Guard the branch with `has_dask_array() and is_dask_cupy(r)`.

### Nits
- [ ] `test_cupy_inputs_do_not_mutate_sources` only checks the inputs are still cupy arrays; it would not catch an in-place value mutation. Asserting the input values still equal the originals would pin what the test name promises.

### What looks good
- The host round-trip matches the `_apply_vertical_shift` precedent, and the dask+cupy path stays lazy (`map_blocks(cp.asnumpy)` in, `map_blocks(cp.asarray)` out) with the `_meta` assertion to prove no compute happens at graph build.
- Value parity vs the numpy path is pinned with `assert_array_equal` (exact, not allclose) across all five strategies, plus the mixed numpy+cupy case.
- The GPU re-wrap keys off `any_dask` after the auto-promote decision, so a large in-memory cupy merge that gets promoted to the lazy path still comes back as dask+cupy rather than crashing on `cp.asarray(dask_array)`.

### Checklist
- [x] All implemented backends produce consistent results (exact parity tests)
- [x] NaN handling unchanged (canonicalization happens in the existing numpy pipeline)
- [x] Dask chunk boundaries handled correctly (per-chunk conversion only)
- [ ] No premature materialization: `cp.asnumpy` on eager cupy inputs is a deliberate, documented host copy; dask inputs stay lazy
- [x] README feature matrix already listed Merge as CPU-fallback on GPU backends; this PR makes that accurate
- [x] Docstring updated with backend behaviour
- [x] Benchmark not needed (no new algorithm; conversion overhead is the documented cost)

brendancol (commented):
Follow-up review after 614c96e2:

- The `is_dask_cupy` branch is now guarded by `has_dask_array()`, so a cupy-without-dask install takes the eager path instead of hitting AttributeError. Fixed as suggested.
- `test_cupy_inputs_do_not_mutate_sources` now compares input values against the originals, so an in-place mutation would fail the test, not just a device change.

Both findings from the first pass are resolved. No new issues spotted in the follow-up diff.

Issue #3114, opened 2026-06-09
No GPU / dask+GPU test coverage for the to_geotiff(pack=True) round-trip


brendancol:
`tests/write/test_pack_3064.py` exercises the `unpack=True` read -> `pack=True` write round-trip on numpy and dask+numpy only. Its module docstring still says "mask_and_scale is a CPU eager + dask read feature (GPU/VRT reject it)", which #3075 made stale: `unpack=True` now works with `gpu=True` and with `gpu=True, chunks=N`. Nothing in the suite writes a GPU-read array back with `pack=True`.

The gap hides a real crash. Both GPU-backed variants raise today (#3112): the pack transform runs in `to_geotiff` before backend dispatch, so the GPU writer never even sees the data.

Wanted:

- GPU and dask+GPU `pack=True` round-trip tests next to the existing numpy/dask ones, behind the GPU-skip marker, `xfail(strict=True)` on #3112 until the crash is fixed so they flip to failures when it is.
- Fix the stale module docstring while in there.

Test-only change; the source fix belongs to #3112.

Found by /sweep-test-coverage (category 1, backend coverage; severity HIGH).

brendancol:
Already done by #3127 (merged 2026-06-10); the PR body didn't use a closing keyword, so this stayed open. `test_pack_round_trip_gpu` in `tests/write/test_pack_3064.py` covers the eager GPU and dask+GPU `pack=True` round-trips behind `requires_gpu`, with `xfail(strict=True, raises=(AttributeError, TypeError))` pinned to the #3112 crash. The stale module docstring was fixed in the same PR. Re-verified on a CUDA host today: 13 passed, 2 xfailed.

Note for anyone working on the pack path: the xfail is strict, so whatever fixes the cupy `fillna` crash (the #3112 fix, or possibly a #3080 fix touching `_pack`) will turn these into hard failures. Remove the marker in the same PR as the fix.

Issue #3112, opened 2026-06-09
to_geotiff(pack=True) crashes on cupy and dask+cupy input


brendancol:
**Describe the bug**

`open_geotiff(path, unpack=True, gpu=True)` works since #3075, but writing the result back with `to_geotiff(out, pack=True)` crashes before any bytes hit disk. `_pack()` in `xrspatial/geotiff/_attrs.py` calls `out.fillna(nodata)`, which goes through xarray's `duck_array_ops.where`, and that breaks on cupy-backed arrays:

- eager GPU read: `AttributeError: module 'cupy' has no attribute 'astype'` (xarray calls `xp.astype(condition, bool)` with `xp=cupy`)
- dask+GPU read: `TypeError: Unsupported type <class 'numpy.ndarray'>` from `cupy.where` when the streaming writer computes the packed graph and the host-side fill value meets a cupy chunk

Repro:

```python
import numpy as np, xarray as xr
from xrspatial.geotiff import open_geotiff, to_geotiff

data = np.array([[1, 2, 255], [4, 5, 6]], dtype=np.uint8)
da = xr.DataArray(
    data, dims=("y", "x"),
    coords={"y": np.arange(2, 0, -1) - 0.5, "x": np.arange(3) + 0.5},
    attrs={"crs": 4326, "nodata": 255,
           "gdal_metadata": {"SCALE": "2.0", "OFFSET": "10.0"}})
to_geotiff(da, "src.tif")

g = open_geotiff("src.tif", unpack=True, gpu=True)
to_geotiff(g, "out.tif", pack=True)    # AttributeError

dg = open_geotiff("src.tif", unpack=True, gpu=True, chunks=2)
to_geotiff(dg, "out2.tif", pack=True)  # TypeError at compute time
```

**Expected behavior**

The same round-trip works on numpy and dask+numpy (covered by `tests/write/test_pack_3064.py`). The GPU writer comment in `_writers/gpu.py` says `to_geotiff` applies the re-pack before dispatch "so every write path sees the re-packed array", so the GPU path is meant to work, not raise.

The eager failure looks like the known cupy 13.6 / xarray 2025.12 `where`/`astype` incompatibility. The dask+gpu one is `_pack` mixing a host-side fill value into a cupy graph. Either way `_pack` needs a cupy-safe NaN fill, or the dependency pin plus a device-side fill value.

**Environment**

[...]

PR #3127, opened 2026-06-09
Add xfail'd GPU and dask+GPU pack=True round-trip tests (#3114)

brendancol:
- Adds `test_pack_round_trip_gpu` to `tests/write/test_pack_3064.py`, parametrized over eager GPU and dask+GPU reads: `open_geotiff(unpack=True, gpu=True[, chunks=2])` followed by `to_geotiff(pack=True)`, asserting the uint8 source ints round-trip and the repacked file unpacks to the same values as the CPU path.
- Both legs crash today (#3112: `_pack`'s `fillna` breaks on cupy-backed arrays), so the tests run behind `requires_gpu` with `xfail(strict=True, raises=(AttributeError, TypeError))` against that issue — pinning to the known exception types so unrelated assertion failures surface as real failures instead of hiding under the known crash. When the crash is fixed they flip to hard failures and the marker comes off.
- Fixes the module docstring, which still claimed GPU rejects `mask_and_scale`; that has been stale since #3075.
- Updates the test-coverage sweep state CSV for the geotiff module.

Test-only change; the source fix stays with #3112.

Backend coverage: numpy and dask+numpy were already tested; this adds the cupy and dask+cupy legs (currently xfail).

## Proposed Changes

  - Add `test_pack_round_trip_gpu` parametrized over eager GPU and dask+GPU backends, gated by `requires_gpu` with `xfail(strict=True, raises=(AttributeError, TypeError))` pinned to the #3112 crash.
  - Pin `xfail` to `raises=(AttributeError, TypeError)` so unrelated failures (e.g. a GPU read-side regression dropping the `mask_and_scale` attr) surface as real failures rather than silently matching the known bug.
  - Fix stale module docstring claiming GPU rejects `mask_and_scale`.
  - Update `.claude/sweep-test-coverage-state.csv` for the geotiff module.
  - Merge `origin/main` into branch, resolving conflict in `.claude/sweep-test-coverage-state.csv` (kept our updated geotiff row; took main's updated rasterize and reproject rows).

Test plan:
- [x] `pytest xrspatial/geotiff/tests/write/test_pack_3064.py` on a CUDA host: 13 passed, 2 xfailed
- [x] Confirmed the new tests execute (not skip) on GPU and fail for the exact #3112 crash before xfail was applied


brendancol (commented):
## PR Review: Add xfail'd GPU and dask+GPU pack=True round-trip tests (#3114)

- None.

- [ ] `xrspatial/geotiff/tests/write/test_pack_3064.py:160` -- the bare `xfail(strict=True)` absorbs *any* failure, not just the #3112 crash. The assert at line 173 (`mask_and_scale_dtype == "uint8"`) runs before the pack call, so a GPU read-side regression that drops the attr would xfail silently and look like the known bug. Pin the expected failure with `raises=(AttributeError, TypeError)` so an unrelated assertion failure surfaces as a real failure instead.

- [ ] The two failure modes in #3112 may be version-dependent (the eager leg matches the known cupy 13.6 / xarray 2025.12 `where`/`astype` incompatibility). If a GPU CI host carries different pins where `fillna` works, `strict=True` turns an accidental pass into a CI failure. That failure would be informative (it means the env, not the code, decided the outcome), so keeping strict seems right -- but watch the first GPU CI run.

### What looks good
- The tests were executed on a CUDA host and observed to fail for the exact #3112 crash before the xfail was applied; 13 passed, 2 xfailed.
- Parametrization, tmp-file naming (`*_3114`), `requires_gpu` gating, and the `_write_int_tiff` fixture reuse all match the surrounding file.
- The stale module docstring (claiming GPU rejects `mask_and_scale`) is corrected in the same place the new coverage lands.
- Test-only: no source changes, the fix stays with #3112.

### Checklist
- [x] Algorithm matches reference (round-trip parity asserted against the CPU eager path)
- [x] All implemented backends produce consistent results (gpu/dask+gpu pinned as xfail on #3112)
- [x] NaN handling is correct (sentinel -> NaN -> sentinel round-trip asserted with equal_nan)
- [x] Edge cases covered (scale/offset + sentinel in one fixture)
- [x] Dask chunk boundaries handled (chunks=2 leg)
- [x] No premature materialization (np.asarray only on results under test)
- [x] Benchmark not needed (test-only PR)
- [x] README feature matrix unchanged (no API change)
- [x] Docstrings accurate after the stale-claim fix

brendancol (commented):
Follow-up review after 5751a115:

- The one suggestion is addressed: the xfail now carries `raises=(AttributeError, TypeError)` with a comment mapping each exception to its backend leg, so an unrelated assertion failure (e.g. the attr check before the pack call) surfaces instead of hiding under the known #3112 crash.
- Re-ran the file on the CUDA host: 13 passed, 2 xfailed.
- The nit (strict xfail vs possible GPU CI dependency pins) stands as a watch item for the first GPU CI run, not a change request.

No remaining findings.

brendancol:

PR #3128, opened 2026-06-09
Mask 64-bit integer sentinels at native width in eager reads (#3098)

### LINKED ISSUE #3098: open_geotiff(masked=True) falsely masks valid pixels near 64-bit integer sentinels on eager backends
**Describe the bug**

`_apply_eager_nodata_mask` in `xrspatial/geotiff/_attrs.py` promotes integer buffers to float64 and then compares against the sentinel:

```python
arr = arr.astype(np.float64)
mask = arr == np.float64(nodata_int)
```

For int64/uint64 rasters with a sentinel above 2**53, float64 rounding makes nearby valid values compare equal to the sentinel. With `nodata=INT64_MAX`, every value in `[INT64_MAX - 512, INT64_MAX]` rounds to 2**63 and gets masked to NaN. With UINT64_MAX the window is 1024 values wide.

The dask chunk path (`_delayed_read_window` in `_backends/dask.py`), the GPU GDS chunk path (`_apply_nodata_mask_gpu` in `_backends/_gpu_helpers.py`), and the VRT path all compute the mask at native integer width before promoting, so they only mask exact sentinel hits. The same file read with `masked=True` therefore gives different results on the eager backends than on dask:

```python
i64max = np.iinfo(np.int64).max
data = np.array([[i64max, i64max - 1, i64max - 100],
                 [i64max - 511, i64max - 512, i64max - 513],
                 [1000, 2000, 3000]], dtype=np.int64)
da = xr.DataArray(data, dims=('y', 'x'),
                  coords={'y': [2.5, 1.5, 0.5], 'x': [0.5, 1.5, 2.5]})
to_geotiff(da, path, nodata=i64max, compression='deflate')

open_geotiff(path, masked=True)            # 4 NaNs (3 are valid pixels)
open_geotiff(path, masked=True, chunks=2)  # 1 NaN (correct)
```

The function is also inconsistent with itself: the `mask_nodata=False` scan a few lines down compares at native width (`arr == arr.dtype.type(nodata_int)`), so `nodata_pixels_present` and the actual masking can disagree about which pixels are nodata.

The write side already defends against exactly this. `_overview_kernels.py` keeps integer reductions on the numpy path because the sentinel mask has to be computed at native integer width before any float64 promotion (64-bit sentinels like INT64_MAX round when cast), and `tests/write/test_overview.py` pins that behavior for UINT64_MAX. The eager read path contradicts the module's own convention.

**Expected behavior**

Eager masked reads mask only pixels that exactly equal the sentinel at the source dtype's width, matching the dask, GPU-chunked, and VRT read paths. Affected backends: numpy eager and cupy eager (both route through `_apply_eager_nodata_mask`; the GPU eager sites call it via duck-typing from `_finalize_eager_read`).

**Additional context**

Found by the accuracy sweep against the geotiff module. Promoting int64 data above 2**53 to float64 is lossy regardless, but the mask should not convert valid pixels to NaN, and the four backends should agree.

brendancol:
Closes #3098

- `_apply_eager_nodata_mask` compared the integer nodata sentinel after promoting the buffer to float64. For sentinels above 2**53 (INT64_MAX, UINT64_MAX), float64 rounding made up to 512 (int64) or 1024 (uint64) nearby valid values compare equal to the sentinel, so eager masked reads turned them into NaN. The fix computes the mask at the source dtype's width before the promotion, the same order the dask per-chunk mask, the GPU GDS chunk path, and the VRT path already use.
- The unconditional float64 promotion from #2990 is unchanged, and the mask now agrees with the native-width `nodata_pixels_present` scan in the same function.
- One site fixes both eager backends: the GPU eager read calls the same helper on CuPy buffers via duck typing.

Backend coverage: numpy and cupy eager were affected and are fixed; dask+numpy and dask+cupy were already correct and now match the eager output (verified on hardware with CUDA).

Test plan:
- [x] 5 regression tests in `tests/read/test_nodata.py`: int64 exact-hit masking, eager-vs-dask NaN-pattern parity, uint64 variant, near-sentinel file with no exact hit (`nodata_pixels_present` False, still promoted to float64), and the cupy eager path
- [x] `tests/read/` + `tests/unit/`: 2199 passed, 5 skipped
- [x] `tests/parity/test_finalization.py` and `tests/test_round_trip.py` pass
- [x] flake8 clean on the touched files


brendancol (commented):
## PR Review: Mask 64-bit integer sentinels at native width in eager reads (#3098)

None.

None.

- [ ] `xrspatial/geotiff/tests/read/test_nodata.py:3517` (test_eager_masked_int64_near_sentinel_without_hit_3098 docstring): "the native\n-width scan" wraps mid-word and reads as "native -width". Rejoin as "native-width".

### What looks good
- The fix is the minimal reorder: compute the equality mask before `astype(np.float64)`, leave the unconditional promotion from #2990 alone. For dtypes of 32 bits and below every value is exact in float64, so behavior is unchanged there; only the broken 64-bit case moves.
- It also resolves the internal inconsistency where the `mask_nodata=False` scan in the same function compared at native width but the masking branch did not, so `nodata_pixels_present` and the actual mask could disagree.
- One site covers both affected backends. The GPU eager read reaches this helper via duck typing from `_finalize_eager_read`, and all the ops involved (`arr == scalar`, boolean-mask assignment, `astype`) were already exercised on CuPy by the previous code, just in a different order. The GPU regression test confirms it on hardware.
- Tests pin the four interesting cases: exact-hit-only masking for INT64_MAX, eager-vs-dask NaN-pattern parity on the same file, the wider UINT64_MAX window, and the no-exact-hit file where `nodata_pixels_present` must be False while the buffer still promotes to float64.

### Checklist
- [x] Algorithm matches reference (dask `_delayed_read_window`, `_apply_nodata_mask_gpu`, and the VRT mask all use the same native-width order)
- [x] All implemented backends produce consistent results (verified numpy, cupy, dask+numpy, dask+cupy with CUDA)
- [x] NaN handling is correct (only exact sentinel hits become NaN)
- [x] Edge cases are covered by tests (near-sentinel values, no-hit file, uint64 max)
- [x] Dask chunk boundaries handled correctly (no dask code changed; parity test reads with chunks=2 across the affected pixels)
- [x] No premature materialization or unnecessary copies (mask now computed once on the int buffer instead of the promoted one)
- [x] Benchmark not needed (mask is a single vectorized compare either way)
- [x] README feature matrix not applicable (bug fix, no API change)
- [x] Docstrings present and accurate (helper comment explains the 2**53 rounding and cites #3098)

brendancol (commented):
Follow-up review after f9392191: the docstring nit is fixed and the 5 regression tests still pass. No remaining findings. Disposition of the original review: 0 blockers, 0 suggestions, 1 nit (fixed in-PR).

PR #3129, opened 2026-06-09
Reject zero and non-finite SCALE/OFFSET under unpack=True; guard pack=True against divide-by-zero (#3104)

### LINKED ISSUE #3104: unpack=True accepts SCALE=0 and non-finite SCALE/OFFSET; pack=True then divides by zero and writes a corrupt file
**Describe the bug**

`_extract_scale_offset` (`xrspatial/geotiff/_attrs.py`) rejects SCALE/OFFSET values that don't parse as floats (#2992), but accepts values that parse and can't be honoured: `SCALE="0"`, `SCALE="nan"`, `SCALE="inf"`, and non-finite `OFFSET`.

Repro on main, with a uint8 file carrying `SCALE=0.0`, `OFFSET=10.0`, `GDAL_NODATA=0` and pixel values 1..25:

```python
da = open_geotiff(path, unpack=True)   # every pixel is now 10.0, no warning
to_geotiff(da, out, pack=True)         # output file: every pixel == 0 (the sentinel)
```

On read, `arr * 0.0 + 10.0` flattens all 25 distinct values to the offset. `SCALE="nan"` turns the whole array to NaN. Neither case warns.

On write, `_pack` computes `(data - offset) / scale` with no zero or finiteness check. With scale 0 the division yields NaN (0/0), `fillna(nodata)` rewrites the whole raster to the sentinel, and the packed file reads back as all nodata. When values differ from the offset the division yields inf and the integer cast writes garbage instead.

The numpy and dask read paths share `_extract_scale_offset`, so both are affected.

**Expected behavior**

Same fail-closed handling that #2992 gave unparseable values: a SCALE of 0 or a non-finite SCALE/OFFSET can't be honoured on read and can't be inverted on write, so `unpack=True` should raise `MalformedScaleOffsetError`. `pack=True` should refuse `attrs['scale_factor']` values of 0 or non-finite instead of writing a corrupt file.

**Additional context**

Found by the security sweep (Cat 3, division without zero check; Cat 6, silent wrong results). Medium severity: it takes malformed metadata in the source file, but the corruption is silent on every path.

brendancol:
Closes #3104

- `_extract_scale_offset` now raises `MalformedScaleOffsetError` when SCALE is zero or non-finite, or when OFFSET is non-finite. This is the same fail-closed handling #2992 gave unparseable values: a SCALE of 0 collapses every pixel to the offset on read and has no inverse for `pack=True`.
- `_pack` refuses a zero or non-finite `scale_factor` / non-finite `add_offset` (reachable only via hand-edited attrs once the read side rejects them) instead of dividing by zero and writing a file where every pixel is the nodata sentinel.

Backend coverage: the fix lives in the one metadata helper every unpack path resolves through. Eager numpy and GPU reads call it via `_finalize_eager_read`; dask and dask+GPU call it in `_backends/dask.py`. The VRT path rejects `unpack=True` at the boundary and is unaffected. GPU and dask+GPU rejection is covered by `requires_gpu`-gated tests and was verified on a live CUDA host.

Test plan:
- [x] New tests in `xrspatial/geotiff/tests/read/test_scale_zero_3104.py`: SCALE 0/-0.0/nan/inf/-inf and OFFSET nan/inf rejection on numpy and dask, GPU and dask+GPU rejection plus a valid-scale GPU unpack (gated on `requires_gpu`), zero SCALE ignored without unpack, negative finite SCALE still honoured, pack guards for hand-edited attrs (28 cases on a CUDA host, 25 without)
- [x] `xrspatial/geotiff/tests/read` + `tests/write`: 2352 passed, 5 skipped

Also carries the security-sweep state CSV update for the geotiff module (audit pass 2026-06-09).


brendancol (commented):
## PR Review: Reject zero and non-finite SCALE/OFFSET under unpack=True; guard pack=True against divide-by-zero (#3104)

None.

- [ ] The PR body says unpack is "rejected at the boundary on GPU and VRT paths". True for VRT, stale for GPU: #3075 added unpack on the GPU and dask+GPU read paths. Both route through `_extract_scale_offset` (GPU eager via `_finalize_eager_read` in `_attrs.py`, dask+GPU via `_backends/dask.py:434`), so the fix already covers them. Confirmed with a live CUDA run: `open_geotiff(path, unpack=True, gpu=True)` and the `chunks=2` variant both raise `MalformedScaleOffsetError` on a SCALE=0 file, and a valid SCALE=2 file still unpacks. Update the PR body and add a `requires_gpu`-gated case to `test_scale_zero_3104.py` so the GPU path stays pinned by CI on GPU hosts.

- [ ] The module docstring of `test_scale_zero_3104.py` names only the eager and dask paths. Mention the GPU paths once the gated test lands.

### What looks good
- The rejection lives in the one helper all read paths share, so there is no per-backend drift to maintain.
- `MalformedScaleOffsetError` keeps the #2992 contract (a `ValueError` subclass), so existing `except ValueError` callers keep catching it.
- Tests cover `-0.0` (which compares equal to 0.0) and pin negative finite SCALE as still honoured, so legitimate inverted-scale files keep working.
- The `_pack` guard names both attrs and the reason in its message, and the hand-edited-attrs route has its own tests.

### Checklist
- [x] Algorithm matches reference (n/a: validation change)
- [x] Backends consistent: numpy/dask tested; GPU verified by hand, gated test requested above
- [x] NaN handling correct (non-finite values rejected up front)
- [x] Edge cases covered: 0, -0.0, nan, +/-inf, offset variants, no-unpack passthrough
- [x] Dask chunk boundaries: n/a, rejection happens at metadata time before graph build
- [x] No premature materialization (raises before any pixel work)
- [x] Benchmark not needed
- [x] Docstrings updated

brendancol (commented):
Follow-up review of 85544b10 (GPU-gated tests added after the first pass):

- Both suggestions are addressed: `test_unpack_rejects_zero_scale_gpu` covers GPU and dask+GPU through the `requires_gpu` marker (the suite's usable-CUDA probe, not a bare cupy import), `test_valid_scale_still_unpacks_gpu` pins the non-rejection side, and the module docstring now names all four read paths. The PR body's backend-coverage paragraph was corrected too.
- The `hasattr(arr, "get")` guard in the valid-scale GPU test is right: small GPU reads can come back as host arrays, so an unconditional `.get()` would break.
- Ran the file on this CUDA host: 28 passed. No new findings.

Disposition of the first-pass findings: both fixed in-PR, nothing deferred or dismissed.

PR #3130, opened 2026-06-09
Fix bounded dask GREAT_CIRCLE proximity missing antimeridian targets (#3108)

### LINKED ISSUE #3108: Bounded dask GREAT_CIRCLE proximity misses targets across the antimeridian
**Describe the bug**

`proximity`/`allocation`/`direction` with `distance_metric='GREAT_CIRCLE'` and a finite `max_distance` disagree between dask and numpy when the raster spans the antimeridian.

The bounded dask path sizes its `map_overlap` halo in array space: `_halo_depth` divides `max_distance` by the densest per-column step distance. That works for planar metrics, but great-circle distance is periodic in longitude. Haversine takes the short way around, so a pixel at lon 179.5 is about 111 km from a target at lon -179.5. In array space that target is at the opposite end of the axis, far outside any per-chunk halo, and the chunk never sees it.

Repro:

```python

lon = np.arange(-179.5, 180.0, 1.0)
lat = np.arange(-5.0, 6.0, 1.0)
data = np.zeros((lat.size, lon.size))
data[5, 0] = 1.0  # target at (lat 0, lon -179.5)

r = xr.DataArray(data, dims=['lat', 'lon'], coords={'lat': lat, 'lon': lon})
kw = dict(x='lon', y='lat', distance_metric='GREAT_CIRCLE', max_distance=250_000.0)

proximity(r, **kw).values[5, -1]                                                   # 111319.49
proximity(r.copy(data=da.from_array(data, chunks=(11, 60))), **kw).values[5, -1]   # nan
```

numpy and cupy run a whole-raster brute-force search and find the target. dask+numpy and dask+cupy both return NaN.

The linear column halo has a second problem at high latitude: a great-circle chord between two points on a parallel is shorter than the sum of the per-column steps, so the halo can be undersized even without a wrap. The worst-latitude sizing usually rescues this by blowing up the pad until the axis folds, but that is not guaranteed.

**Expected behavior**

All four backends return the same result. The chord bound `dist >= 2R asin(cos(lat_max) |sin(dlon/2)|)` holds for any pair of grid points, so the column halo can be derived from it. When the seam gap (or the 180-degree chord at the worst-case latitude) is still within `max_distance`, no array-space halo works and the x axis has to be folded into a single chunk, which `_fit_halo_to_chunks` already does for oversized pads.

**Additional context**

`_halo_depth` is shared by the dask+numpy and dask+cupy bounded paths, so one fix covers both.

brendancol:
Closes #3108.

Bounded `proximity`/`allocation`/`direction` with `distance_metric='GREAT_CIRCLE'` returned NaN on dask for targets across the +/-180 seam, while numpy and cupy found them. The `map_overlap` halo was sized as a linear sum of per-column parallel-arc steps, but great-circle distance is periodic in longitude and its chords shorten toward the poles, so array-space adjacency is not a lower bound on spherical distance.

- `_halo_depth` now derives the GREAT_CIRCLE column halo from the chord bound `dist >= 2R asin(cos(lat_max) |sin(dlon/2)|)`, which holds for every pair of grid points.
- When the seam gap (or the 180-degree chord at the worst-case latitude) is within `max_distance`, no array-space halo can cover the wrap; the returned depth exceeds the axis length so the existing `_fit_halo_to_chunks` folds the x axis into a single chunk. This also covers over-pole shortcuts.
- The row halo stays linear: great-circle distance is never smaller than the meridian separation, so the per-row step sum remains a valid lower bound.
- Regional rasters away from the seam and poles keep a small finite halo (no needless fold).

Backend coverage: the fix is in `_halo_depth`, shared by the bounded dask+numpy and dask+cupy paths. numpy and cupy were already correct (whole-raster brute force).

Test plan:
- [x] New regression test: global 1-degree raster spanning the antimeridian, all three ops, dask+numpy and dask+cupy match numpy (wrap pixel finite at ~111319.49 m, previously NaN).
- [x] New unit test: `_halo_depth` folds on wrap-reachable and pole-adjacent rasters, returns a finite halo for a regional raster.
- [x] Full `test_proximity.py` suite: 417 passed on a CUDA host (cupy and dask+cupy parametrizations executed).


brendancol (commented):
## PR Review: Fix bounded dask GREAT_CIRCLE proximity missing antimeridian targets (#3108)

- None found. The chord bound is valid for every grid-point pair (both latitudes are bounded by lat_max, and haversine's `|sin(dlon/2)|` takes the short way around), and the non-fold branch is sound: when `seam_gap > dlon_max`, every wrap-side separation is at least `360 - span = seam_gap`, so only direct separations matter and `ceil(dlon_max / min_step)` covers them.

- [ ] `xrspatial/tests/test_proximity.py` (new `test_great_circle_halo_folds_on_wrap_and_pole`): all three cases use ascending lon coords. `_great_circle_col_halo` relies on `abs()` for the span and `np.abs(np.diff(...))` for the step, which should make descending coords work, but nothing pins that. Add a descending-lon case (the module explicitly supports descending monotonic coords; see `test_descending_coords_allowed`).

- [ ] `xrspatial/tests/test_proximity.py`: `_antimeridian_raster` returns `(raster, data)` but both callers only use `raster` (the dask test re-reads `raster.data`). Return just the raster.
- [ ] `xrspatial/proximity.py:313`: `radius = 6378137.0` is the fourth copy of this constant in the module (`great_circle_distance` default, a comment, and the GPU device function). Consistent with existing style, so optional, but a module-level constant would keep them from drifting.

### What looks good
- The fix lands in `_halo_depth`, which both bounded dask paths share, so dask+numpy and dask+cupy are covered by one change.
- Fold is signalled as `width + 1`, reusing `_fit_halo_to_chunks` instead of adding a second fold mechanism; a single-chunk axis folds cleanly to depth 0 rather than padding a full chunk width of NaN.
- The row halo is left linear with a comment explaining why that remains a valid lower bound (meridian separation), so the change is scoped to where the old bound was actually wrong.
- The regression test asserts the wrap pixel is finite on numpy before comparing, so it cannot pass with both backends returning NaN.
- 417 tests pass on a CUDA host with the cupy and dask+cupy parametrizations executed.

### Checklist
- [x] Algorithm matches reference (chord lower bound derived and checked)
- [x] All implemented backends produce consistent results (verified on CUDA host)
- [x] NaN handling is correct (unchanged; boundary NaN padding still excluded via `isfinite`)
- [x] Edge cases covered (wrap, pole-adjacent, regional no-fold; max_distance=0 yields pad 0 via dlon_max=0)
- [x] Dask chunk boundaries handled correctly (fold guarantees full-axis visibility)
- [x] No premature materialization (halo math uses only 1D coord arrays; graph stays lazy)
- [x] Benchmark exists (benchmarks/benchmarks/proximity.py); no new function added
- [x] README feature matrix: not applicable (no new function, no backend change)
- [x] Docstrings present and accurate (`_great_circle_col_halo`, updated `_halo_depth`)

brendancol (commented):
## Follow-up review (after 90c9cca5)

Delta-only re-review of the follow-up commit:

- Descending-coordinate coverage added to `test_great_circle_halo_folds_on_wrap_and_pole`: regional case asserts pad equality with the ascending raster, and a descending wrap raster still folds. Verified `_great_circle_col_halo` is direction-independent (span via `abs()`, step via `np.abs(np.diff(...))`), so the assertions pin real behavior.
- `_antimeridian_raster` now returns just the raster; both callers updated.

Disposition of the original findings:
- Suggestion (descending-lon coverage): fixed.
- Nit (unused helper return): fixed.
- Nit (shared Earth-radius constant): dismissed. Unifying the constant would mean editing the public `great_circle_distance` signature and the CUDA device function, code this PR does not otherwise touch; the new code matches the module's existing style of a local literal.

No new findings. 48 great-circle/halo tests pass locally; flake8 clean on the changed files.

PR #3133, opened 2026-06-09
Keep integer dtype for empty reproject chunks (#3096)

### LINKED ISSUE #3096: reproject: dask backend computes float64 for integer rasters when output chunks miss the source footprint
**Describe the bug**

`_reproject_dask` builds its map_blocks template with the source's integer dtype (#2505), but the empty-chunk returns are hardcoded to float64:

- `_reproject_chunk_numpy`: the non-finite, no-overlap, and oversized-window early returns all do `np.full(_empty_shape, nodata, dtype=np.float64)`
- `_reproject_chunk_cupy`: same, with `cp.full`
- `_reproject_block_adapter`: the footprint-skip path returns `np.full(chunk_shape, nodata, dtype=np.float64)`

When any output chunk has no overlap with the source footprint, that chunk comes back float64 while the data chunks come back int16/uint8/etc. dask stitches the blocks with type promotion, so the computed array is float64 even though the lazy array advertises the integer dtype.

Repro:

```python

data = (np.random.rand(200, 200) * 1000).astype(np.int16)
r = xr.DataArray(da.from_array(data, chunks=(64, 64)), dims=['y', 'x'],
                 coords={'y': np.linspace(52, 51, 200), 'x': np.linspace(-2, -1, 200)},
                 attrs={'crs': 'EPSG:4326', 'nodata': -32768})
out = reproject(r, 'EPSG:3857', bounds=(-2e6, 5e6, 2e6, 9e6),
                resolution=20000, chunk_size=64)
print(out.dtype)            # int16
print(out.compute().dtype)  # float64
```

The eager numpy backend returns int16 for the same input, so the backends disagree on output dtype. Values are still correct (the integer sentinel is representable in float64); the dtype contract is what breaks.

**Expected behavior**

Empty chunks should use the dtype the template advertises: integer source means `np.full(shape, nodata, dtype=src_dtype)` (the resolved nodata is guaranteed representable for integer rasters per #2185/#2572), float source stays float64.

brendancol:
Closes #3096

- Empty-chunk fills in `_reproject_chunk_numpy`, `_reproject_chunk_cupy`, and the footprint-skip path in `_reproject_block_adapter` now use the source's integer dtype instead of a hardcoded float64. One no-overlap chunk used to promote the whole computed dask array to float64 while the lazy array (and the eager backend) advertised int16/uint8/etc. Float sources keep float64 fills.
- The resolved nodata is always representable for integer rasters (#2185/#2572), so the integer fill is safe.
- Adds `TestEmptyChunkDtype`: dask+numpy int16 with forced empty chunks (advertised == computed == int16, sentinel in the empty corners), eager numpy with a fully no-overlap output grid, a float64 guard, and a dask+cupy variant.
- Also carries the accuracy-sweep state CSV update for the reproject module.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy (the cupy worker and block adapter paths got the same fix).

Test plan:
- [x] `pytest xrspatial/tests/test_reproject.py -k EmptyChunkDtype` (4 passed on a GPU host)
- [x] `pytest xrspatial/tests/test_reproject.py` (436 passed)
- [x] `pytest xrspatial/tests/test_reproject_coverage_2026_05_27.py xrspatial/tests/test_reproject_cupy_gate_2564.py` (12 passed)


brendancol (commented):
## PR Review: Keep integer dtype for empty reproject chunks (#3096)

None found.

None.

- In `_reproject_block_adapter`, the 2-D footprint-skip path on dask+cupy still returns a numpy fill while the data chunks are cupy arrays (`xrspatial/reproject/__init__.py:2003`, pre-existing; the 3-D branch does return `cp.full`). The dask meta is numpy either way, so the stitched result behaves, but the array-type mix is worth a follow-up note if dask+cupy empty chunks ever grow a device-side consumer. Out of scope here since this PR only changes the fill dtype.
- A non-integral float nodata on an integer raster (e.g. explicit `nodata=0.5` with int16) would truncate silently in the new `np.full(..., dtype=int16)` fills. The data path's round-and-clip cast already collapses such a sentinel the same way, so this PR doesn't make it worse, and `_detect_nodata` range-checks but doesn't integrality-check. Pre-existing input-validation gap, not introduced here.

### What looks good
- All eleven empty-fill sites got the same treatment: four in `_reproject_chunk_numpy`, five in `_reproject_chunk_cupy`, and the footprint-skip path in `_reproject_block_adapter`, so the fix can't be dodged by which early-return a chunk happens to hit.
- The dtype rule matches the template logic in `_reproject_dask` and the eager cast-back exactly (`np.issubdtype(..., np.integer)`), so bool and float sources keep float64 fills and nothing else shifts.
- Leaning on the #2185/#2572 guarantee that integer rasters always resolve an in-range nodata is the right call; the integer fill is safe by construction and the comment says why.
- Tests pin the contract from four angles: dask advertised-vs-computed dtype with forced empty chunks (plus a sentinel check in the empty corner and a has-real-data guard so the test can't pass vacuously), the eager fully-no-overlap case, a float64 regression guard, and a dask+cupy run.
- Merge paths are unaffected: the canonicalize step already promotes via `np.where(arr == r_nd, np.nan, arr)`, which handles integer fills the same as the old float64 ones.

### Checklist
- [x] Algorithm unchanged (fill dtype only; values identical)
- [x] All implemented backends produce consistent results (numpy, dask+numpy, cupy worker, dask+cupy covered)
- [x] NaN handling is correct (float sources keep NaN fills, guarded by test)
- [x] Edge cases are covered by tests (empty corners, fully-no-overlap output, float guard)
- [x] Dask chunk boundaries handled correctly (per-chunk fills now match the template dtype)
- [x] Benchmark not needed (no hot-path change; fills were already allocated)
- [x] README feature matrix unchanged (no API or backend-support change)
- [x] Docstrings/comments explain the why, with issue references

brendancol (commented):
## Follow-up review (after 6a24085b)

The first nit from the previous review is addressed: the 2-D footprint-skip path in `_reproject_block_adapter` now returns `cp.full` on dask+cupy, matching the 3-D branch and the data chunks. Full reproject suite re-run on a GPU host: 436 + 12 passed.

Remaining item, dismissed with reason: the non-integral-float-nodata-on-integer-raster truncation is a pre-existing validation gap in `_detect_nodata` (unmodified by this PR), and the data path's round-and-clip cast already collapses such a sentinel identically. No change needed here.

No new findings.

PR #3134, opened 2026-06-09
Document actual return types of vertical datum functions (#3097)

### LINKED ISSUE #3097: reproject vertical functions: Returns docstrings claim 'same type as input' but ndarray comes back
The five vertical-datum functions in `xrspatial/reproject/_vertical.py` describe their return value as "same type as input", which is wrong for most input types.

`geoid_height` (lines 203-258) accepts "float, array-like, or xr.DataArray" and its Returns section says "N : same type as input". What actually happens:

```python
>>> geoid_height(xr.DataArray([-74.0]), xr.DataArray([40.7]))
numpy.ndarray          # not a DataArray
>>> geoid_height(-74.0, 40.7)
float                  # scalar path does float(out[0])
```

The four conversion wrappers (`ellipsoidal_to_orthometric`, `orthometric_to_ellipsoidal`, `depth_to_ellipsoidal`, `ellipsoidal_to_depth`) say "same type as height" but return `np.asarray(height) +/- N`, so scalar input produces `np.float64` and list/DataArray input produces `np.ndarray`. There's also a small drift among siblings: `geoid_height` converts scalars back to Python float, the wrappers do not.

These are documentation bugs. A user passing a DataArray and expecting coords/attrs back gets a bare array. The fix is to state the actual return types in each Returns section; changing the runtime behaviour to round-trip DataArrays would be a separate feature decision.

Found by the API consistency sweep (category: docstring/signature divergence).

brendancol:
Closes #3097

The five vertical functions in `xrspatial/reproject/_vertical.py` described their return value as "same type as input", which is wrong: DataArray input returns a plain ndarray, and the four conversion wrappers return numpy scalars for scalar input (only `geoid_height` converts back to Python float).

- Returns sections now state what actually comes back: `geoid_height` gives a Python float for scalar input, otherwise an ndarray shaped like the inputs; the wrappers pass `height`/`depth` through `np.asarray`, so scalar in gives a numpy scalar, array-like or DataArray in gives an ndarray.
- New `TestVerticalReturnTypes` pins those types so a future behaviour change has to update the docs too.
- One commit also carries the api-consistency sweep state CSV update for the reproject module.

Docs-only change at runtime; no behaviour modified. Making these functions round-trip DataArrays would be a separate feature decision, noted in the issue.

Test plan:
- [x] New return-type pin tests pass.
- [x] flake8 clean on the touched file.


brendancol (commented):
## PR Review: Document actual return types of vertical datum functions (#3097)

### Blockers
None.

### Suggestions
None.

### Nits
- [ ] The same three-line np.asarray explanation is repeated in four Returns sections (_vertical.py). Numpy-style docs favour explicit over DRY, so this is fine as-is; flagging only in case a shared "Notes" section is preferred later.

### What looks good
- Each documented claim matches the code: `geoid_height` does `float(out[0])` on the scalar path and `out.reshape(np.shape(lon))` otherwise, and the wrappers return `np.asarray(height) +/- N`, so the new wording ("numpy scalar for scalar input, plain ndarray otherwise, DataArray coords not carried through") is verifiably accurate.
- `TestVerticalReturnTypes` pins the types with `type(...) is` checks rather than `isinstance`, so a silent change from float to np.float64 (or ndarray to DataArray) fails loudly and forces a doc update.
- No runtime behaviour touched; the diff is Returns sections plus tests.

### Checklist
- [x] Docstrings present and accurate (verified against the implementation)
- [x] Edge cases: scalar, 1-D array, and DataArray inputs all covered by the pin tests
- [x] No behaviour change, so backend parity and NaN handling are unaffected

PR #3136, opened 2026-06-09
Group streaming-write rows into bands so source chunks compute once (#3117)

### LINKED ISSUE #3117: Streaming dask writer recomputes every source chunk once per 256-row band
**Describe the bug**

`_write_streaming` (xrspatial/geotiff/_writer.py) writes a dask-backed array by calling `.compute()` once per output tile-row (tiled layout, default `tile_size=256`) or once per 256-row strip (`tiled=False`). Each compute call is an independent dask graph execution, so a source chunk taller than 256 rows gets re-read and re-decoded once for every band it overlaps.

The amplification factor is `chunk_height / tile_height`. Counting `map_blocks` executions on the source array during `to_geotiff`:

- `open_geotiff(path, chunks=512)` then `to_geotiff(...)`: every source chunk executes 2x
- `open_geotiff(path, chunks=1024)` then `to_geotiff(...)`: every source chunk executes 4x

When the source is a computed pipeline (say `slope()` of a chunked read), the whole upstream graph re-runs that many times, not just the file read.

The memory budgeting from #3007 already accounts for each tile-row compute materialising every touched source chunk-row in full (`_max_streaming_row_span`), and sizes the horizontal-segment budget from that span. But the writer keeps only one 256-row slice of what it materialised and throws the rest away, then recomputes the same chunk-rows for the next tile-row.

**Expected behavior**

Source chunks should be computed roughly once. The per-compute peak is already set by the source chunk-row span (capped by `streaming_buffer_bytes`), so the writer can group consecutive tile-rows or strips into a row band whose materialised span fits the existing budget, compute the band once, and carve all tiles or strips in the band out of the materialised array. Peak memory stays inside the documented `streaming_buffer_bytes` cap. The only remaining recompute is the chunk straddling a band boundary.

Wide rasters that need horizontal segmentation (one tile-row exceeds the budget) can keep the current per-tile-row path. Banding only applies when a full-width tile-row fits the budget.

**Repro**

```python
import threading
from collections import Counter
import numpy as np
import xarray as xr

data = xr.DataArray(
    np.random.default_rng(0).random((1024, 1024)).astype('float32'),
    dims=['y', 'x'],
    coords={'y': np.arange(1024.0), 'x': np.arange(1024.0)})
to_geotiff(data, 'src.tif', compression='zstd')

lazy = open_geotiff('src.tif', chunks=512)
counts, lock = Counter(), threading.Lock()
def record(block, block_info=None):
    loc = tuple(block_info[0]['chunk-location']) if block_info else None
    with lock:
        counts[loc] += 1
    return block
instrumented = lazy.copy(data=lazy.data.map_blocks(record, dtype=lazy.dtype))
to_geotiff(instrumented, 'dst.tif', compression='zstd')
print(sorted(v for k, v in counts.items() if k is not None))
# [2, 2, 2, 2]  -- every 512x512 source chunk decoded twice
```

**Additional context**

- Affected path: dask to `to_geotiff` (non-COG, string path), both tiled and strip layouts. The eager and GPU writers are unaffected.
- Found by the performance sweep. OOM verdict for the module: SAFE (this is compute amplification, not a memory problem). Bottleneck: IO-bound.

brendancol:
Closes #3117

- `_write_streaming` ran one dask `.compute()` per 256-row tile-row (or strip), so a source chunk taller than the band re-executed once per band it overlapped: 2x at `chunks=512`, 4x at `chunks=1024`, with the whole upstream graph re-running for computed pipelines like `slope()`.
- New `_stream_row_bands` groups consecutive tile-rows/strips into row bands sized by the source chunk-row span (including the one-chunk overlap halo, same accounting as #3007) under `streaming_buffer_bytes`. Each band is computed once and the tiles/strips are carved out of the materialised array, so peak memory stays where it was.
- Wide rasters that need horizontal segmentation keep the per-tile-row path. `streaming_buffer_bytes` stays a soft cap.

Backend coverage: only the dask+numpy write path changes (`to_geotiff` on a dask-backed array, non-COG, string path). The eager, cupy, and dask+cupy writers are untouched.

Test plan:
- [x] New `TestRowBandRecompute3117`: instrumented chunk-execution counts equal 1 for tiled and strip layouts; segmented wide-raster round trip; NaN-sentinel parity between banded streaming and eager writes
- [x] New `test_stream_row_bands_3117` unit coverage of the band grouping, budget cutoff, and unknown-chunks fallback
- [x] `xrspatial/geotiff/tests/write/` + `integration/` + `parity/` suites pass (2195 passed, 44 skipped)
- [x] #3007 budget tests still pass (peak source bytes per compute stay bounded)
- [x] Measured: per-chunk executions drop from chunk_height/tile_height to 1 on the default `open_geotiff(chunks=512)` -> `to_geotiff` round trip


brendancol (commented):
## PR Review: Group streaming-write rows into bands so source chunks compute once (#3117)

- None found.

- [ ] `.claude/sweep-performance-state.csv`: the geotiff row's notes column dropped the 13-pass audit trail that previous sweeps kept (`Pass 13 (2026-05-20): ... | Pass 12 ...`). Other long-lived rows in this file (rasterize, polygonize, zonal) prepend the new pass and keep the old text. Restore the prior notes behind the new entry so the history survives.
- [ ] `xrspatial/geotiff/tests/write/test_streaming.py`: the budget contract for the new banded path has no direct peak-bytes test. `test_overlap_source_respects_buffer_3007` pins peak source bytes per compute, but only on the segmented path (its tight 8 MB budget turns banding off). A variant with a budget large enough to enable banding but small enough to force several bands would pin that `_stream_row_bands` actually bounds the per-compute span on the path this PR adds.

- [ ] PR has no labels; `performance` and `geotiff` fit, and the repo uses them for sweep-driven fixes.
- [ ] `_writer.py` banded branch: when NaNs are present the sentinel rewrite copies the whole band (`band_np.copy()`), which can transiently double a budget-sized buffer. Same pattern as the old per-tile-row copy and bounded by 2x the soft cap, so fine as-is; a comment noting the bound would help the next reader.

### What looks good
- `_stream_row_bands` reuses the #3007 span accounting (touched chunk-rows plus a one-chunk halo) so the banding and the segment budget agree on what a compute materialises.
- The segmented wide-raster path is untouched: `band_np` stays `None` there and the fallback band list is one tile-row per band, byte-identical behaviour to before.
- Regression tests count actual chunk executions through `map_blocks` and assert exactly 1, for both tiled and strip layouts; the unit test covers grouping, the budget cutoff, unknown chunks, and ragged tails.
- The strip path previously had no budget participation at all; it now shares the same banding code instead of growing a second mechanism.
- 2195 write/integration/parity tests pass, including the #3007 budget tests and the eager-vs-streaming byte parity checks.

### Checklist
- [x] Algorithm matches reference (#3007 span accounting reused)
- [x] All implemented backends produce consistent results (dask+numpy only path changed; eager/cupy writers untouched)
- [x] NaN handling is correct (per-band sentinel restore, parity test vs eager writer)
- [x] Edge cases are covered by tests (ragged tail, unknown chunks, budget below one band)
- [x] Dask chunk boundaries handled correctly (bands extend only while the span fits)
- [x] No premature materialization beyond the documented soft cap
- [ ] Benchmark exists or is not needed (no benchmark; covered by deterministic execution-count test instead)
- [x] README feature matrix update not applicable
- [x] Docstrings updated (`_write_streaming` notes the banding for both layouts)

brendancol (commented):
## Follow-up review (after ebd155a5)

- CSV audit trail: fixed. The geotiff row now prepends Pass 14 and keeps the 13-pass history behind a `|` separator, matching the rasterize/polygonize/zonal precedent.
- Banded budget contract: fixed. `test_banded_compute_respects_buffer` asserts banding groups tile-rows (computes < tile-rows) and that no single compute materialises more source bytes than `streaming_buffer_bytes`. The bound is strict because the spy source has no overlap halo.
- Labels: `performance`, `geotiff`, `dask` added.
- Copy-bound comment: added at the band NaN-sentinel rewrite, noting the 2x-soft-cap transient and that `astype` shares the same factor.

Re-checked the follow-up diff: the `da.Array.compute` spy accumulates inside the compute call boundary, so the before/after delta per compute is race-free under the threaded scheduler. 77 streaming tests pass, flake8 clean. Nothing further from me.

PR #3137, opened 2026-06-09
Build dask proximity coordinate grids with chunk-aligned broadcast (#3132)

### LINKED ISSUE #3132: proximity: dask coordinate grids built with tile/repeat+rechunk bloat the task graph
## Problem

The bounded dask paths in `xrspatial/proximity.py` build the 2D coordinate grids like this (dask+numpy around line 1622, dask+cupy around line 638):

```python
xs = da.tile(x_coords_da, (raster.shape[0], 1)).rechunk(raster.chunks)
ys = da.repeat(y_coords_da, raster.shape[1]).reshape(raster.shape).rechunk(raster.chunks)
```

The values are right, but the graph cost is steep. On a 2560x2560 raster with 256x256 chunks (100 chunks):

- xs (tile + rechunk): 2,772 tasks
- ys (repeat + reshape + rechunk): 10,449 tasks
- full bounded `proximity(max_distance=...)` graph: 18,535 tasks, ~185 per chunk

Worse, the ys term scales with raster height, not chunk count: 10,449 tasks at 2560 rows, 21,299 at 5120, 44,199 at 10240 (about 4.3 tasks per row). On rasters with millions of rows the scheduler chokes on the coordinate grids alone, before any proximity work is scheduled.

Building the same grids from per-axis chunked 1D arrays with `da.broadcast_to` gives identical values at about 1 task per chunk:

```python
xs = da.broadcast_to(x_1d[None, :], shape, chunks=chunks)   # 120 tasks
ys = da.broadcast_to(y_1d[:, None], shape, chunks=chunks)   # 120 tasks
```

and the full bounded proximity graph drops from 18,535 to 5,554 tasks (3.3x) with no change in results.

## Fix

Replace the tile/repeat constructions in both dask paths (`_process_dask` setup in `_process`, and `_process_dask_cupy`) with chunk-aligned `da.broadcast_to`. The chunk functions only read the grids, so the read-only broadcast views are safe; `map_overlap` materializes padded copies per chunk as before.

Affects dask+numpy and dask+cupy backends. Found during a performance sweep of the proximity module.

brendancol:
Closes #3132

- Replaces the `da.tile` / `da.repeat(...).reshape(...)` + `.rechunk()` construction of the 2D coordinate grids with chunk-aligned `da.broadcast_to`, in both the dask+numpy path (`_process` setup) and the dask+cupy path (`_process_dask_cupy`).
- The old construction cost ~100 graph tasks per chunk, with the repeat term growing linearly with raster height (44K tasks for the ys grid alone on a 10240-row raster). The broadcast grids are ~1 task per chunk and hold identical values.
- Bounded proximity on a 2560x2560 raster with 256-chunks drops from 18,535 to 5,554 graph tasks (3.3x).
- Adds a regression test that bounds tasks-per-chunk (old code: 100.4/chunk, new: 58.7/chunk, threshold 80) and checks value parity against the numpy backend on uniform and ragged chunks.

Backend coverage: dask+numpy and dask+cupy are the changed paths; numpy and cupy untouched. The chunk functions only read the grids, so the read-only broadcast views are safe; `map_overlap` still materializes padded per-chunk copies as before.

Test plan:
- [x] `pytest xrspatial/tests/test_proximity.py` — 411 passed on a CUDA host (dask+cupy parity tests included)
- [x] New test `test_proximity_dask_coord_grid_graph_size` fails against the old construction (measured 100.4 tasks/chunk vs the 80/chunk bound)


brendancol (commented):
## Review

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/tests/test_proximity.py::test_proximity_dask_coord_grid_graph_size`: the 80 tasks/chunk threshold sits between the measured old (100.4) and new (58.7) values. If dask's map_overlap internals ever add tasks, the margin is ~35%. Acceptable, but worth knowing the bound is empirical rather than derived.

### What was checked
- Both construction sites changed the same way (dask+numpy in `_process`, dask+cupy in `_process_dask_cupy`); values are identical because the broadcast grids come from the same 1D coordinate arrays.
- The chunk functions only read xs/ys, so read-only broadcast views are safe. When `depth > 0`, map_overlap pads into freshly allocated arrays; when chunks get folded or rechunked (`_fit_halo_to_chunks`, the unbounded single-chunk branch), rechunk materializes real arrays. The zero-depth skinny-raster and unbounded GREAT_CIRCLE paths are covered by the existing suite.
- `da.from_array(coords, chunks=(raster.data.chunks[1],))` uses the explicit tuple-of-tuples form, so ragged chunks align exactly; the new test checks parity on (200, 150) ragged chunks against the numpy backend.
- Graph effect confirmed: 2560x2560 with 256-chunks drops 18,535 -> 5,554 tasks; the old ys construction grew ~4.3 tasks per raster row, which the broadcast form eliminates.
- All 411 proximity tests pass on a CUDA host, including dask+cupy parity.

PR #3140, opened 2026-06-10
Add proximity coverage tests: int rasters, bounded GPU metrics, empty input (#3139)

### LINKED ISSUE #3139: proximity: test gaps for integer rasters, bounded dask+cupy metrics, empty rasters
Third test-coverage pass over `xrspatial/proximity.py` (prior passes: #2692). Three gaps remain in `xrspatial/tests/test_proximity.py`. All three behaviors check out as correct on a CUDA host, so this is test-only work to pin them down.

1. Integer-dtype rasters are never tested. Every raster in the test file is float64, but integer class grids are the main use case for `allocation`. The bounded dask path calls `da.map_overlap(..., boundary=np.nan)` on the raw input array, and for an int dtype the NaN pad casts to INT_MIN, which is a non-zero finite "target" value. The only reason halo pixels don't become phantom targets is that the coordinate grids' pads are real NaNs, so the phantom distances come out NaN and never win the nearest-target scan. No test pins that down.

2. Bounded dask+cupy only runs with EUCLIDEAN. `_process_dask_cupy` sizes its overlap halo using the active distance metric (`_halo_depth`). Metric-unit mismatches in that sizing are the #2809/#2854 bug family, but those regression tests only cover dask+numpy. No test runs a finite `max_distance` with MANHATTAN or GREAT_CIRCLE through the GPU chunked path.

3. Empty rasters are unpinned. A (0, N) or (N, 0) raster currently fails fast with an IndexError from the coordinate-endpoint lookup. Nothing asserts that it raises instead of returning garbage.

Planned tests for `xrspatial/tests/test_proximity.py`:
- integer-dtype raster (int32) x {proximity, allocation, direction} x 4 backends x {bounded, unbounded}, compared against the float64 numpy baseline, with float32 output dtype asserted
- bounded dask+cupy x {MANHATTAN, GREAT_CIRCLE} x 3 functions, compared against numpy
- empty raster (0 rows / 0 cols) raises

No source changes.

brendancol:
Closes #3139

Test-only PR; `xrspatial/proximity.py` is untouched.

- Integer-dtype rasters (int32) now run on all four backends, bounded and unbounded, pinned against the float64 numpy baseline (24 tests + 4 explicit-target_values tests). The bounded dask case is the interesting one: `boundary=np.nan` on an int array casts the pad to INT_MIN, a finite non-zero "target" value, and the only thing neutralizing those phantom targets is that the coordinate-grid pads are real NaNs.
- The bounded dask+cupy path (`_process_dask_cupy`) was only ever exercised with EUCLIDEAN. Added MANHATTAN and GREAT_CIRCLE x {proximity, allocation, direction} against the numpy baseline, with a spy asserting the bounded GPU path actually ran (6 tests). Mutation check: zeroing the halo depth flips all 6 red.
- Empty (0, N) / (N, 0) rasters now have a fail-fast pin: they raise (currently IndexError from the coordinate-endpoint lookup) instead of returning garbage (6 tests).

Backend coverage: numpy / cupy / dask+numpy / dask+cupy all executed locally on a CUDA host.

Test plan:
- [x] 40 new tests pass on a CUDA host
- [x] full `test_proximity.py` passes (450 passed, 0 skipped)
- [x] mutation check on the bounded-GPU halo flips the new tests red; source restored byte-identical (md5)


brendancol (commented):
## PR Review: Add proximity coverage tests: int rasters, bounded GPU metrics, empty input (#3139)

### Blockers
None.

### Suggestions
- Int coverage was int32-only. uint8 is the usual land-cover dtype and exercises a different NaN-pad cast (lands on a non-target 0 instead of int32's phantom-target INT_MIN). Fixed in be721dd1 by parametrizing dtype.
- `test_integer_raster_explicit_target_values` ran proximity only; allocation and direction never saw integer pixels with explicit `target_values`. Fixed in be721dd1 by parametrizing func.

### Nits
- The empty-raster test is numpy-only. That's fine: the raise happens in `_process` before backend dispatch (the `max_possible_distance` endpoint lookup), so one backend covers all four. No change needed.

### What looks good
- The spy on `_process_dask_cupy` means the bounded-GPU metric tests can't silently pass through a fallback path if dispatch thresholds drift.
- The mutation check (zeroing the halo depth) flipped all 6 bounded tests red, and the source was restored byte-identical.
- The int tests hit both the unbounded dask KDTree path (max_distance=inf) and the bounded map_overlap halo path (max_distance=10), which is where the NaN-pad cast lives.

### Checklist
- Algorithm reference: n/a, test-only
- Backend consistency: 72 new tests across numpy / cupy / dask+numpy / dask+cupy, executed on a CUDA host
- NaN handling: pinned via the int-pad tests
- Edge cases: empty-raster raise pinned
- Dask chunk boundaries: bounded halo path now exercised with non-default metrics
- Benchmark / README / docstrings: not applicable, test-only

Issue #1370, opened 2026-04-29
owa() unbounded xr.concat allocation across criteria


brendancol:
## Summary

`xrspatial.mcda.combine.owa()` builds a temporary stack of every weighted criterion via `xr.concat` along a new `__mcda_criterion` axis, then sorts that stack descending along axis 0. There is no size check, so a caller passing many criteria over a moderate raster will exhaust host memory before any error is raised.

## Where

`xrspatial/mcda/combine.py`, function `owa()`:

```python
weighted_layers = []
for var_name in criteria.data_vars:
    w = criterion_weights[var_name]
    weighted_layers.append(criteria[var_name] * w * n)

# Stack, sort descending along criterion axis, apply order weights
stacked = xr.concat(weighted_layers, dim="__mcda_criterion")
sorted_data = _sort_descending(stacked.data, axis=0)
```

## Memory cost

The stack holds every criterion as float64:

- bytes per pixel = `n_criteria * 8`
- total = `n_criteria * H * W * 8`

`_sort_descending` then runs `-np.sort(-data, axis=0)`, which roughly doubles the working set during the sort.

## Worked example

100 criteria over a 10000 x 10000 raster:

`100 * 10000 * 10000 * 8 = 80 GB`

The sort step roughly doubles that to ~160 GB peak. A 32 GB laptop will swap or OOM.

1000 criteria over a 4000 x 4000 raster:

`1000 * 4000 * 4000 * 8 = 128 GB`

## Proposed fix

Add a working-memory guard at the top of `owa()` that estimates `n_criteria * H * W * 8` and raises `MemoryError` if it exceeds 50% of available host RAM. The dask path is bounded per chunk, so skip the guard when the input data is dask-backed.

Same shape of guard as #1319 / #1361 / #1367 / #1369.

## Audit reference

Recorded in the mcda security audit alongside #1311 (the NaN/Inf weight fix series), as "MEDIUM Cat 1: combine.owa stacks all criteria via xr.concat without size guard".

PR #3153, opened 2026-06-10
Guard monte_carlo dask materialization against OOM (#3145)

### LINKED ISSUE #3145: sensitivity monte_carlo materializes dask criteria with no memory check
**Describe the bug**
`sensitivity(..., method="monte_carlo")` calls `criteria.compute()` when the input Dataset is dask-backed (`_monte_carlo` in `xrspatial/mcda/sensitivity.py`). That pulls every criterion layer into host memory at once, plus two full-size float64 accumulators for the Welford pass. There is no size check before the compute, so a Dataset that only fits in memory chunked (the reason to use dask in the first place) can OOM the host instead of failing cleanly.

`owa()` had the same hazard and got a guard in #1370. `_monte_carlo` predates that guard and never got one.

**Expected behavior**
Before materializing a dask-backed Dataset, estimate the working set (criterion layers plus accumulators, float64) and raise MemoryError when it exceeds the available-memory budget, with a message that suggests coarsening or subsetting the rasters. Same approach as the `owa()` guard in `combine.py`.

**Additional context**
The eager compute itself is intentional: it avoids chaining ~35 graph nodes per sample into an unparallelizable 35k-task graph (see the code comment). The guard is the missing piece, not a different evaluation strategy.

brendancol:
Closes #3145.

- `sensitivity(method="monte_carlo")` now estimates the working set before materializing a dask-backed criteria Dataset: (n_criteria + 3 working buffers) x pixels x 8 bytes, checked against 50% of available memory. Oversize inputs raise MemoryError naming the criterion count and grid shape instead of OOMing mid-compute. Reuses `_available_memory_bytes` and the budget from the `owa()` guard (#1370).
- The check runs before the Welford accumulators are allocated, so nothing full-size gets allocated when the input cannot fit.
- numpy-backed inputs skip the check (the layers are already resident) and the lazy `one_at_a_time` path is untouched.

Backend coverage: numpy behavior unchanged; dask+numpy gets the guard. No GPU code touched (monte_carlo accumulates through numpy buffers on every backend).

Test plan:
- [x] New `TestMonteCarloMemoryGuard`: oversize dask input raises before compute, numpy input skips the guard, small dask input still works, error message names count and shape, `one_at_a_time` unaffected
- [x] `pytest xrspatial/tests/test_mcda.py` -- 180 passed
- [x] flake8 on changed files -- only pre-existing warnings

Also includes the security-sweep state CSV update for the mcda module (sweep bookkeeping, one row).


brendancol (commented):
## PR Review: Guard monte_carlo dask materialization against OOM (#3145)

None.

- [ ] sensitivity.py:188 -- the guard sizes the working set from `n_criteria = len(weights)`, but `criteria.compute()` materializes every data var in the Dataset. Weight/criteria key equality is only enforced later, inside `combine_fn` (after the compute), so a Dataset with more vars than weights gets an underestimated budget and still materializes everything before the ValueError fires. Passing `len(criteria.data_vars)` makes the estimate match what compute() actually allocates. For valid inputs the two counts are equal, so this only changes the mismatched-input path.

- [ ] sensitivity.py:184-189 -- a dask+cupy-backed Dataset passes `_is_dask_dataset` and gets checked against host RAM, but its compute() lands in GPU memory. That path already dies later at `score.values` (cupy refuses implicit numpy conversion), so the guard is never the deciding factor; a short comment noting the guard is host-memory-only would save the next reader the analysis.

### What looks good
- The guard runs before the Welford accumulators are allocated, so an oversize input raises before any full-size buffer exists.
- Reuses `_available_memory_bytes` and `_BYTES_PER_VALUE` from combine.py and keeps the same 50% budget as the owa() guard, so the two guards stay in sync.
- Error message names the criterion count, grid shape, required and available GB, and what to do about it.
- Tests cover the raise-before-compute case, the numpy skip, the happy path, the message contents, and that one_at_a_time stays lazy and unguarded.

### Checklist
- [x] Algorithm matches reference (mirrors the #1370 owa guard)
- [x] All implemented backends produce consistent results (numpy unchanged, dask+numpy guarded; no GPU code touched)
- [x] NaN handling is correct (guard is shape-only, no value math)
- [x] Dask chunk boundaries handled correctly (guard is chunk-agnostic, runs pre-compute)
- [x] No premature materialization or unnecessary copies (the point of the change)
- [x] Benchmark exists or is not needed (O(1) check, not needed)

brendancol (commented):
## Follow-up review after a640ade

- The guard now sizes the estimate from `len(criteria.data_vars)`, matching what `compute()` materializes. The new `test_estimate_counts_data_vars_not_weights` pins it: a 3-var Dataset with 2 weights reports "3 criteria" in the MemoryError.
- The comment at the call site now states the guard is host-memory-only and why the dask+cupy path is out of scope (it fails later at `.values` regardless).

No new issues. 181 tests pass, flake8 clean on the changed module. Nothing further from me.

PR #3154, opened 2026-06-10
Preserve input attrs in mcda.constrain() (#3147)

### LINKED ISSUE #3147: mcda.constrain() drops raster attrs when exclusion masks are applied
**Describe the bug**

`constrain()` in `xrspatial/mcda/constrain.py` returns a DataArray with empty `attrs` whenever `exclude` contains at least one mask. The loop rewrites the result with `xr.where(mask, fill, result)`, and `xr.where` takes its attrs from the first value argument, which is the scalar `fill`. So `res`, `crs`, and `nodatavals` are gone from the constrained surface. With `exclude=[]` the attrs survive, since the function just returns a copy.

```python
import numpy as np
import xarray as xr
from xrspatial.mcda import constrain

suit = xr.DataArray(
    np.full((2, 3), 0.5), dims=["y", "x"],
    attrs={"res": (1.0, 1.0), "crs": "EPSG:32611", "nodatavals": (-9999.0,)},
)
mask = xr.DataArray(np.zeros((2, 3), dtype=bool), dims=["y", "x"])
print(constrain(suit, exclude=[]).attrs)      # attrs kept
print(constrain(suit, exclude=[mask]).attrs)  # {} -- attrs gone
```

**Expected behavior**

The output keeps the input's attrs regardless of how many masks are passed. constrain is usually the last step before exporting the suitability surface, so losing `crs`/`res` here means the exported GeoTIFF has no georeferencing.

**Additional context**

Affects numpy, dask+numpy, and dask+cupy the same way (verified on all three). Bare cupy currently fails earlier inside `xr.where` because of the known cupy 13.6 / xarray 2025.12 incompatibility, which is a separate issue.

Found by the metadata propagation sweep against the mcda module.

brendancol:
Closes #3147

- `constrain()` lost all input attrs (`res`, `crs`, `nodatavals`) whenever `exclude` was non-empty: `xr.where(mask, fill, result)` takes attrs from the scalar `fill`. With `exclude=[]` attrs survived, so behavior also depended on whether any mask was passed.
- The fix restores `suitability.attrs` (as a copy) on the output after the masking loop.
- New `TestConstrainAttrs` covers attrs survival with a mask, with a custom fill, with an empty exclude list, attrs-dict isolation from the input, and the dask+numpy and dask+cupy paths.

Backend coverage: numpy, dask+numpy, and dask+cupy verified end-to-end. On bare cupy, `xr.where` itself raises before the fix is reached (known cupy 13.6 / xarray 2025.12 incompatibility, tracked separately); the dask+cupy test asserts attrs on the lazy result for the same reason.

Test plan:
- [x] `pytest xrspatial/tests/test_mcda.py` (181 passed, includes 6 new attrs tests)
- [x] Manual 4-backend probe: attrs now kept on numpy / dask+numpy / dask+cupy

Found by the metadata propagation sweep. The sweep state CSV update for the mcda module rides along in this PR.


brendancol (commented):
## PR Review: Preserve input attrs in mcda.constrain() (#3147)

None.

None.

- [ ] `xrspatial/mcda/constrain.py:30-33`: the Returns section could state that the output keeps the input's attrs. The fix establishes that contract; documenting it makes the next regression visible in review rather than in a sweep.

### What looks good
- The root cause is identified precisely: `xr.where` takes attrs from its first value argument, the scalar `fill`, so any non-empty `exclude` stripped attrs. Restoring `dict(suitability.attrs)` after the loop fixes every mask count and fill value in one place, and the dict copy keeps the output from aliasing the input's attrs (covered by `test_attrs_not_shared_with_input`).
- Test coverage matches the bug shape: mask present, custom fill, empty exclude list, dask+numpy with value parity against numpy, and dask+cupy asserting attrs on the lazy result. The comment in `test_attrs_kept_dask_cupy` explaining why values are not computed there (pre-existing cupy/xarray `xr.where` incompatibility) is accurate; I reproduced that failure on main without this change.
- Hand-rolled fixtures match the existing test_mcda.py style, which does not use general_checks helpers anywhere.

### Checklist
- [x] Algorithm matches reference/paper (N/A, metadata-only fix)
- [x] All implemented backends produce consistent results (numpy, dask+numpy verified by value; dask+cupy attrs verified lazily; bare cupy blocked upstream before this code runs)
- [x] NaN handling is correct (unchanged; NaN fill still propagates)
- [x] Edge cases are covered by tests (empty exclude, custom fill, attrs aliasing)
- [x] Dask chunk boundaries handled correctly (no chunk-level code touched)
- [x] Benchmark exists or is not needed (not needed, no compute change)
- [ ] Docstrings present and accurate (see nit: attrs contract not documented)

The sweep state CSV row riding along is expected for metadata-sweep PRs.

brendancol (commented):
Follow-up review after 9f-commit addressing the docstring nit:

- The Returns section of `constrain` now states the output keeps the input's dims, coords, and attrs. That closes the only finding from the first pass.
- No new code paths were touched; constrain tests still pass (13 passed locally).

Nothing outstanding.

PR #3155, opened 2026-06-10
Fix mcda API consistency: owa weights param, boolean_overlay Dataset input, ahp_weights docs, ConsistencyResult export (#3148)

### LINKED ISSUE #3148: mcda: owa weights naming, boolean_overlay input type, ahp_weights docstring, ConsistencyResult export
An API consistency pass over `xrspatial.mcda` (10 public functions) found four user-facing inconsistencies.

## 1. `owa()` calls its criterion-weights dict `criterion_weights`; the other combiners call it `weights`

`wlc(criteria, weights)`, `wpm(criteria, weights)`, and `sensitivity(criteria, weights)` all take the per-criterion weight dict as `weights`. `owa(criteria, criterion_weights, order_weights)` is the outlier, even though the dict has identical semantics and goes through the same `_validate_weights` helper. Switching from `wlc` to `owa` by keyword fails:

```python
owa(ds, weights=w, order_weights=ow)
# TypeError: owa() got an unexpected keyword argument 'weights'
```

A rename is breaking, so the fix needs a deprecation shim: rename the parameter to `weights`, keep `criterion_weights` accepted as a keyword alias, emit `DeprecationWarning` on the old name, raise `TypeError` if both are given. Positional callers are unaffected.

## 2. `boolean_overlay()` documents `criteria: dict[str, xr.DataArray]`; every sibling combiner takes `xr.Dataset`

`wlc`/`wpm`/`owa`/`fuzzy_overlay` take an `xr.Dataset` of criterion layers. `boolean_overlay` annotates and documents a dict only. A Dataset actually works today because Dataset implements the Mapping interface the function uses, but that's undocumented, and a dict built for `boolean_overlay` can't be passed to `fuzzy_overlay` (raises `TypeError`). Fix: document and annotate `xr.Dataset | dict[str, xr.DataArray]`, with Dataset as the primary form.

## 3. `ahp_weights()` docstring says incomplete comparisons raise; they warn

The Raises section claims `ValueError` "If criteria list has fewer than 2 items or comparisons are incomplete." The code (and `TestAHPIncomplete`) emit a `UserWarning` and default missing pairs to 1. The docstring should match the tested behavior.

## 4. `ConsistencyResult` is returned by a public function but not exported

`ahp_weights` returns `(weights, ConsistencyResult)`, but the class is missing from `xrspatial/mcda/__init__.py` and from `docs/source/reference/mcda.rst`. Anyone wanting to type-annotate or isinstance-check the second return value has to import from `xrspatial.mcda.weights` directly. Export it and add it to the reference docs.

## Out of scope here, recorded for other sweeps

- `owa()` fails on cupy input: the order-weights array stays numpy and cupy refuses the mixed multiply (`combine.py` around line 336).
- `owa()` fails on dask input at graph construction: `dask.array` has no `sort` (`combine.py:356`). The OWA `MemoryError` message recommends switching to dask, which then crashes.
- `sensitivity(method="monte_carlo")` fails on cupy: `template.values` hits cupy's implicit-conversion guard.
- `constrain()` drops `attrs` (via `xr.where`) while the other nine public functions preserve them.

Found by the api-consistency sweep on 2026-06-10.

brendancol:
Closes #3148

- `owa()` now takes its criterion-weight dict as `weights`, matching `wlc`/`wpm`/`sensitivity`. `criterion_weights` still works as a keyword-only alias: it emits a `DeprecationWarning`, and passing both names raises `TypeError`. Positional callers are unchanged.
- `boolean_overlay()` is annotated and documented as accepting `xr.Dataset | dict[str, xr.DataArray]`. Datasets already worked through the Mapping interface; now the contract says so, and it matches the other combiners.
- `ahp_weights()` docstring no longer claims incomplete comparisons raise `ValueError`. They warn (`UserWarning`) and missing pairs default to 1, which is what the code and `TestAHPIncomplete` have always done. Added a Warns section.
- `ConsistencyResult` is exported from `xrspatial.mcda` and listed in `docs/source/reference/mcda.rst`, since it's half of the public `ahp_weights` return value.

Backend coverage: these are signature/doc/export changes shared by all four backends (mcda dispatches internally via `_get_xp`). The `owa` shim and `boolean_overlay` Dataset input were verified on numpy and cupy entry points. Note `owa` has pre-existing execution bugs on cupy and dask (numpy order-weights mixed into a cupy multiply; `da.sort` doesn't exist) that are recorded in #3148 for a separate fix; kwargs acceptance was confirmed up to that point on cupy.

Test plan:
- [x] `pytest xrspatial/tests/test_mcda.py` (188 passed)
- [x] New tests: `weights=` keyword, deprecated alias warning, both-names `TypeError`, missing-arg `TypeError`s, positional parity, OWA==WLC under equal order weights, Dataset/dict parity for `boolean_overlay`, empty-Dataset error, `ConsistencyResult` import + `__all__`, docstring regression
- [x] cupy smoke: new kwargs accepted, deprecation warning fires, `boolean_overlay(Dataset)` returns cupy-backed result


brendancol (commented):
## PR Review: mcda API consistency (#3148)

None found.

- [ ] `xrspatial/mcda/combine.py:240-241` -- the shim makes `weights` and `order_weights` default to `None`, so the signature no longer tells a reader they're required; the docstring entry for `weights` also still reads `dict` with no hint that it can be passed as `criterion_weights` positionally never. Consider a one-line note in the docstring ("required; defaults to None only to support the deprecated alias") so help() output isn't misleading.
- [ ] `xrspatial/tests/test_mcda.py` (TestAHPIncompleteDocstring3148) -- `doc.split("Raises")[1].split("Warns")[0]` will IndexError rather than fail with a useful message if someone reorders the docstring sections. An `assert "Warns" in doc` first would make the failure readable.

- [ ] `xrspatial/mcda/combine.py:300` -- the missing-argument TypeErrors wrap a short string across three lines; they'd fit on one.
- [ ] `xrspatial/mcda/combine.py:498` -- error message changed from "criteria dict is empty" to "criteria is empty"; the existing test matches on "empty" so this is safe, but worth knowing it's a user-visible string change.

### What looks good
- The shim follows the same pattern as the focal `raster->agg` and reproject vertical-CRS renames: keyword-only old name, DeprecationWarning with stacklevel=2, TypeError on conflict, positional callers untouched.
- Tests pin behavior parity (alias result equals canonical result, OWA==WLC under equal order weights) instead of just "doesn't crash".
- The ahp_weights docstring now matches what TestAHPIncomplete has always asserted.
- Internal-only rename confirmed: `criterion_weights` had no callers outside combine.py.

### Checklist
- [x] All implemented backends produce consistent results (signature/doc change; numpy and cupy entry points verified, pre-existing owa cupy/dask execution bugs recorded in #3148)
- [x] Edge cases are covered by tests (both-names conflict, missing args, empty Dataset)

brendancol (commented):
Follow-up review after 5865643e.

- Suggestion (weights docstring vs None default): fixed -- docstring now states the parameter is required and that the None default exists only for the deprecated alias.
- Suggestion (brittle docstring test): fixed -- the test asserts both section headers exist before slicing, so a reordered docstring fails with a readable message.
- Nit (multi-line TypeErrors): fixed -- collapsed to single lines.
- Nit (error string "criteria dict is empty" -> "criteria is empty"): no change needed; recorded as a deliberate user-visible string update now that the function accepts Datasets too. The existing test matches on "empty".

Re-checked the new diff: no new issues. 188 tests pass locally; flake8 over the touched files reports only pre-existing findings (unused import in standardize.py and old test-file warts), none introduced here.

Issue #3146, opened 2026-06-10
mcda: owa, piecewise/categorical standardize, and monte_carlo sensitivity raise on cupy/dask backends


brendancol:
Found while auditing `xrspatial/mcda` for backend parity. Several code paths written to support cupy and dask raise instead of returning results. The numpy paths are fine.

**owa fails on every dask backend.** `_sort_descending` in `combine.py` calls `da.sort`, which does not exist in dask.array (checked against dask 2025.7.0):

```
AttributeError: module 'dask.array' has no attribute 'sort'
```

This raises at graph construction for both dask+numpy and dask+cupy. The rest of `owa` is written for dask (the memory guard skips dask inputs, the order weights get wrapped with `da.from_array`), and there is no dask test for `owa`, so it slipped through.

**owa also fails on cupy.** The reshaped order-weight array stays numpy and is multiplied against the sorted cupy stack:

```
TypeError: Unsupported type <class 'numpy.ndarray'>
```

**standardize(method="piecewise") fails on cupy and dask+cupy.** On cupy, `xp.interp` gets numpy breakpoint/value arrays, which `cupy.interp` rejects:

```
NotImplementedError: Only int or ndarray are supported for a
```

On dask+cupy, `_interp_block` calls `np.asarray(block)` on a cupy chunk:

```
TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.
```

The same `np.asarray(block)` pattern breaks `standardize(method="categorical")` on dask+cupy.

**sensitivity(method="monte_carlo") fails on cupy and dask+cupy.** `_monte_carlo` in `sensitivity.py` reads `template.values` and `score.values`. For cupy-backed inputs (including dask+cupy after the eager `criteria.compute()`), `.values` triggers the implicit-conversion TypeError above.

Environment: cupy 13.6.0, dask 2025.7.0, CUDA available.

Not included here: `constrain` on cupy raises `AttributeError: module 'cupy' has no attribute 'astype'` inside `xr.where`. That is the known cupy 13.6 + xarray incompatibility affecting call sites repo-wide and is handled by a dependency pin, not an mcda change.

PR #3156, opened 2026-06-10
Add cupy, dask, metadata, and edge-case test coverage for mcda (#3149)

### LINKED ISSUE #3149: mcda: no cupy/dask+cupy test coverage; owa/wpm/constrain dask paths and metadata preservation untested
test_mcda.py has 175 tests, and every one of them exercises the numpy or dask+numpy path. Gaps found while auditing `xrspatial/mcda` for the test-coverage sweep:

**No cupy or dask+cupy tests.** standardize.py has explicit cupy branches (`_get_xp`, cupy chunk handling in `_interp_block` / `_apply_mapping`) and combine.py has a dedicated cupy branch in `_sort_descending`, but no test constructs a cupy-backed DataArray. Running the GPU paths by hand: most work (standardize linear/sigmoidal/gaussian/triangular/categorical, wlc, wpm, fuzzy_overlay, OAT sensitivity), several raise (#3146 covers those bugs).

**owa has no dask test.** The dask branch of `_sort_descending` calls `da.sort`, which doesn't exist in dask.array, so owa raises AttributeError on any dask-backed input. The bug shipped because nothing exercises the path (filed in #3146). Tests should land as xfail until that's fixed.

**Other dask+numpy paths untested:** wpm (its positivity check has a dask branch using `da.nanmin`), constrain, fuzzy_overlay's gamma operator, boolean_overlay.

**No metadata preservation tests.** Nothing asserts that attrs (res/crs), coords, or dim names survive standardize/combine/constrain. Current behavior is correct (checked by hand) but unguarded.

**Edge cases:** all-NaN criterion through wpm's positivity check and Inf propagation through wlc are untested.

Plan: a test-only PR adding cross-backend parity tests via the general_checks.py helpers, xfail tests for the paths broken by #3146, and metadata preservation tests. No source changes.

Found by the test-coverage sweep. Severity per the sweep rubric: HIGH (backend coverage), MEDIUM (metadata, edge cases).

brendancol:
Closes #3149.

- Adds cross-backend parity tests for mcda: standardize (all 7 value functions), wlc, wpm, owa, fuzzy_overlay, boolean_overlay, constrain, and sensitivity, each compared against the numpy result on cupy, dask+numpy, and dask+cupy. test_mcda.py previously had no cupy or dask+cupy tests at all.
- Adds metadata preservation tests (attrs, coords, dim names, output name) for every pipeline stage, plus edge cases: all-NaN criterion through wpm's positivity check and Inf propagation through wlc and fuzzy "and".
- Test-only, no source changes. Paths that currently raise are pinned with strict xfail markers that flip when the bugs get fixed: owa on dask and cupy, piecewise/categorical standardize on GPU backends, and monte_carlo sensitivity on cupy (#3146); constrain dropping attrs under masking (#3147). constrain on cupy/dask+cupy is xfail(strict=False) because that failure is the known cupy 13.6 + xarray xr.where incompatibility, not an mcda bug.

Backend coverage: numpy (existing), cupy, dask+numpy, dask+cupy. All new tests executed locally on a CUDA host.

Test plan:
- [x] `pytest xrspatial/tests/test_mcda.py`: 233 passed, 11 xfailed (CUDA host, cupy 13.6)
- [x] GPU classes ran rather than skipped: 32 passed, 9 xfailed under `-k Cupy`
- [x] flake8: no new warnings (3 pre-existing ones in the file left alone)


brendancol (commented):
## PR Review: Add cupy, dask, metadata, and edge-case test coverage for mcda (#3149)

- None.

- [ ] `xrspatial/tests/test_mcda.py:2186` (TestConstrainBackends) and `:2225` (TestBooleanOverlayBackends): the dask variants don't assert the result stays lazy, unlike `_assert_standardize_matches_numpy` and `_assert_combine_matches_numpy` which check `hasattr(result.data, "compute")`. A regression that eagerly materializes constrain/boolean_overlay on dask input would pass these tests. Add the same laziness assertion.
- [ ] `xrspatial/tests/test_mcda.py:2037,2070` : the xfail split for the broken standardize methods uses positional slices (`STANDARDIZE_CASES[:-1]`, `[:-2]`). If someone appends a case to the list the slices silently shift and the wrong method gets the xfail. Selecting by method name (e.g. filtering on `case[0]`) keeps the markers pinned to the right cases.

- [ ] `xrspatial/tests/test_mcda.py:2056` (`test_result_stays_on_gpu`): the inline `import cupy` is fine under the `@cuda_and_cupy_available` gate, but the parity helper already proves dispatch returns cupy via `_to_numpy`'s `.get()` path; the extra test is borderline redundant. Harmless, keep or drop.

### What looks good
- The xfail markers are strict for the deterministic #3146 breakages and non-strict only for the environment-dependent xr.where/cupy incompatibility, so the markers flip loudly when #3146 is fixed.
- Parity tests reuse `create_test_raster` and compare every backend against the numpy result on data that includes a NaN cell, so NaN handling parity is covered, not just the happy path.
- The new metadata tests pin attrs/coords/dims/name per stage, with the constrain attrs drop correctly split out and xfailed against #3147 instead of papering over it.
- Test-only diff; the source bugs surfaced by the new coverage are tracked in #3146/#3147 rather than bundled here.

### Checklist
- [x] Algorithm matches reference (n/a, test-only)
- [x] All implemented backends produce consistent results (asserted; broken paths xfailed with issue links)
- [x] NaN handling is correct (NaN-bearing parity data, all-NaN wpm case)
- [x] Dask chunk boundaries handled correctly (ragged 3x3 chunks on 5x4 data)
- [x] No premature materialization in tests beyond final comparisons
- [x] Benchmark exists or is not needed (test-only)
- [x] README feature matrix update not applicable
- [x] Docstrings present where useful

brendancol (commented):
Follow-up review after 80cb5c9.

- Suggestion (missing laziness assertions in constrain/boolean_overlay dask tests): fixed. Both `_assert_matches_numpy` helpers now assert `hasattr(result.data, "compute")` for dask backends.
- Suggestion (positional `STANDARDIZE_CASES[:-1]` / `[:-2]` slices): fixed. Cases are now selected by method name via `_standardize_cases(exclude=...)` and `_standardize_case(name)`, so appending a case can't shift the xfail markers.
- Nit (`test_result_stays_on_gpu` redundancy): dismissed, and the opposite turned out to be true. `_to_numpy` passes numpy arrays through silently, so without a type check the cupy parity tests would pass even if dispatch fell back to numpy. Kept the explicit test and additionally added `isinstance(result.data, type(input.data))` assertions to both parity helpers, which extends the backend-type check to every parametrized case.

Re-ran the suite on the CUDA host: 233 passed, 11 xfailed. No new findings.

PR #3157, opened 2026-06-10
Fix mcda cupy and dask backend failures in owa, standardize, and sensitivity (#3146)

### LINKED ISSUE #3146: mcda: owa, piecewise/categorical standardize, and monte_carlo sensitivity raise on cupy/dask backends

brendancol:
Closes #3146.

- `owa` no longer raises on dask input: `_sort_descending` rechunks the criterion axis and sorts each block with `np.sort` (dask.array has no `sort()`), passing `meta` explicitly so cupy-backed chunks keep their type. The order-weight array is converted to match the sorted stack's backend, which fixes the cupy `TypeError`.
- `standardize(method="piecewise")` converts breakpoints to the data's array module and forces C-contiguity, both of which `cupy.interp` requires. The piecewise and categorical dask chunk functions use the block's own array module instead of `np.asarray`, which raises on cupy chunks.
- Monte-carlo `sensitivity` accumulates Welford state with the array module matching the computed criteria and reads `score.data` instead of `.values`, so cupy and dask+cupy inputs work.

Backend coverage: numpy behavior unchanged (all 175 pre-existing tests pass). dask+numpy, cupy, and dask+cupy now run and match numpy (atol 1e-14 / rtol 1e-10 across standardize methods, owa, and monte-carlo sensitivity; NaN positions identical).

Test plan:
- [x] `xrspatial/tests/test_mcda.py`: 184 passed (9 new tests: 2 dask owa, 7 GPU-gated)
- [x] GPU tests executed on a CUDA host (cupy 13.6.0, dask 2025.7.0)
- [x] flake8 on changed files: no new warnings (two pre-existing ones left alone: unused `warnings` import in `_sigmoidal`, unused `finite` variable in a test)

README: sensitivity GPU columns move from unsupported to experimental.


brendancol (commented):
## PR Review: Fix mcda cupy and dask backend failures in owa, standardize, and sensitivity (#3146)

None found.

- [ ] `standardize(method="categorical")` gets a dask+cupy parity test (`test_standardize_categorical_dask_cupy`) but no plain-cupy one. The plain-cupy path worked before this PR, so it is not a regression, but the new `TestCupyBackends` class is the natural place to pin it down. A backend parametrize like the piecewise test would close the gap cheaply.

- [ ] `sensitivity.py` now imports the private `_get_xp` from `standardize.py` (xrspatial/mcda/sensitivity.py:9). Fine within the package, but if a third mcda module ever needs it, it should move to a shared helper rather than grow more cross-module private imports.
- [ ] `_interp_block` converts `bp`/`vl` to the block's array module on every block (xrspatial/mcda/standardize.py:270-277). The arrays are a handful of floats so the per-block transfer is noise; noting it only in case piecewise ever takes large breakpoint tables.

### What looks good
- The `meta=data._meta` fix in `_sort_descending` (xrspatial/mcda/combine.py:360-371) addresses the actual root cause: dask's meta inference ran the lambda on an empty meta block and fell back to numpy, which then steered the order-weight array to the wrong backend. Easy to get wrong, and the comment explains it.
- Descending-sort NaN behavior is unchanged: both the numpy path and the new per-block path use the same negate-sort-negate construction, so NaN placement matches and the existing `test_owa_nan` semantics carry over to dask (covered by the new `test_owa_dask_nan_propagates`).
- Parity was verified empirically on a CUDA host across all four backends for standardize (six methods), owa, and monte-carlo sensitivity, with NaN positions identical and atol 1e-14 / rtol 1e-10.
- The 3x3 fixtures chunked at 2x2 give ragged chunks, so the dask tests exercise uneven chunk boundaries, not just clean splits.
- Welford accumulation in `_monte_carlo` is unchanged numerically for numpy input (zeros init is equivalent since NaN flows in from the first score), and the reproducibility tests confirm it.

### Checklist
- [x] Algorithm matches reference (no algorithm change; backend plumbing only)
- [x] All implemented backends produce consistent results (verified on CUDA host)
- [x] NaN handling is correct
- [x] Edge cases are covered by tests (NaN, ragged chunks, single criterion unchanged)
- [x] Dask chunk boundaries handled correctly (criterion axis rechunked to one chunk before sort)
- [x] No premature materialization beyond the pre-existing, documented eager paths (`_linear` bounds reduction, monte-carlo `criteria.compute()`)
- [x] Benchmark exists or is not needed (bug fix, no benchmark needed)
- [x] README feature matrix updated (sensitivity GPU columns now experimental)
- [x] Docstrings present and accurate (no public signature changes)

brendancol (commented):
Follow-up review after 7aecac0e.

The one suggestion from the first pass is resolved: `test_standardize_categorical` is now parametrized over cupy and dask+cupy, so the plain-cupy categorical path has a parity test (185 tests pass locally, GPU tests executed on a CUDA host).

The two nits were left as-is, on purpose:
- The private `_get_xp` import stays in `sensitivity.py`. Only two modules use the helper; moving it to a shared home before a third consumer exists would churn files for no behavior change.
- The per-block `bp`/`vl` conversion in `_interp_block` stays. Breakpoint lists are a handful of floats, so hoisting would complicate the chunk function to save microseconds.

PR #3158, opened 2026-06-10
Fix owa() dask path and batch wpm() validation into one scheduler pass

### LINKED ISSUE #3150: mcda: owa() crashes on dask input; wpm() validation runs one compute per criterion
Found by the performance sweep against `xrspatial/mcda` (dask 2025.7.0). Two problems in `xrspatial/mcda/combine.py`, both on the dask path.

## 1. `owa()` raises AttributeError on dask-backed Datasets

`_sort_descending` calls `da.sort`, which does not exist in dask:

```python
from xrspatial.mcda import owa

ds = xr.Dataset({
    "a": xr.DataArray(np.random.rand(20, 20), dims=["y", "x"]).chunk({"y": 10}),
    "b": xr.DataArray(np.random.rand(20, 20), dims=["y", "x"]).chunk({"y": 10}),
})
owa(ds, {"a": 0.5, "b": 0.5}, [0.6, 0.4])
# AttributeError: module 'dask.array' has no attribute 'sort'
```

This is worse than a missing feature. When the criteria stack won't fit in RAM, the eager path's memory guard raises MemoryError telling the user to "Use a dask-backed Dataset for out-of-core processing", and that path then crashes. There is no dask test for `owa` (`TestDaskChunkAlignment` covers standardize, wlc, and fuzzy_overlay only), which is how this slipped through.

Fix: rechunk the criterion axis to a single chunk and sort per block with `map_blocks` + `np.sort(axis=0)`. Sorting an axis fully contained in one chunk is exact, and peak memory stays bounded by the block size.

## 2. `wpm()` validation calls `.compute()` once per criterion

`_check_wpm_positive` loops over `criteria.data_vars` and runs `float(da.nanmin(arr).compute())` for each variable. With N dask-backed criteria that is N separate scheduler runs, serialized, each a full pass over one layer, all at call time before the actual product is computed. Measured: a 4-criterion dataset triggers 4 scheduler invocations during graph construction.

Fix: collect the `nanmin` reductions and run them through one `dask.compute(*mins)` call so validation is a single parallel pass.

Backends affected: dask+numpy and dask+cupy. The numpy paths behave the same as before.

brendancol:
Closes #3150

- `_sort_descending` called `da.sort`, which does not exist in dask, so `owa()` raised AttributeError on any dask-backed Dataset. It now rechunks the criterion axis to a single chunk and sorts each block with `np.sort` via `map_blocks` (exact, since the axis lives in one chunk; memory stays bounded by block size). `meta` is forwarded so dask+cupy blocks keep their cupy meta.
- While testing the dask+cupy path I found the order-weights array was always numpy, which cupy kernels reject. `owa()` now copies the order weights to the device when the data is cupy-backed (plain or dask), so both cupy paths work instead of raising TypeError.
- `_check_wpm_positive` ran `float(da.nanmin(arr).compute())` once per criterion, i.e. one serialized scheduler pass per layer at call time. The nanmin reductions are now collected and run through a single `dask.compute()` call.

Backend coverage: numpy unchanged; dask+numpy fixed (owa) and faster validation (wpm); cupy and dask+cupy owa now work and were verified on a CUDA host.

Test plan:
- [x] `pytest xrspatial/tests/test_mcda.py` (182 passed, includes 7 new tests)
- [x] New dask tests: owa parity vs numpy, owa uniform-order-weights == wlc, wpm parity, wpm single-scheduler-pass assertion, wpm non-positive rejection on dask
- [x] New GPU tests (gated on `cuda_and_cupy_available`): owa cupy and dask+cupy parity, run on a CUDA host
- [x] Graph probe: owa on 2560x2560 / 256-chunks builds ~15 tasks per chunk, no fan-in blowup


brendancol (commented):
## PR Review: Fix owa() dask path and batch wpm() validation into one scheduler pass

None.

- [ ] `xrspatial/tests/test_mcda.py` (TestOWADask): the dask tests use even 10x10 chunks only. A ragged-chunk case (e.g. `chunk({"y": 7, "x": 9})` on a 20x20 grid) would catch block-alignment mistakes in the rechunk+map_blocks sort, which is exactly the code this PR adds.

- [ ] `xrspatial/mcda/combine.py:387` (`_sort_block_descending`): a one-line note that NaN ordering matches the eager `-np.sort(-data)` path (NaNs sort to the same end) would save the next reader from re-deriving it.

### What looks good
- The per-block sort is exact because the criterion axis is rechunked into a single chunk first, and `meta=data._meta` keeps dask+cupy blocks typed correctly.
- The scheduler-pass regression test pins the actual behavior (one `dask.compute` call with all three reductions) instead of timing anything.
- The order-weights device copy fixes a second crash (plain-cupy owa) that the issue did not even cover, with parity tests on both GPU backends.

### Checklist
- [x] Algorithm matches reference (descending sort + order weights, parity vs eager path)
- [x] All implemented backends produce consistent results (numpy/dask verified in tests; cupy/dask+cupy gated tests ran on a CUDA host)
- [x] NaN handling is correct (NaN fixture in the dask parity test)
- [x] Edge cases are covered by tests (non-positive rejection on dask, uniform order weights == wlc)
- [x] Dask chunk boundaries handled correctly (axis fully contained per block)
- [x] No premature materialization (validation now a single batched compute; sort path stays lazy)
- [x] README feature matrix: no tier changes needed (owa GPU cells already marked experimental)
- [x] Docstrings present and accurate (helper docstrings added)

brendancol (commented):
Follow-up review after b-commit:

- Ragged-chunk suggestion: fixed. `test_owa_dask_matches_numpy` is now parametrized over even (10x10) and ragged (7x9) chunkings; both pass.
- NaN-ordering nit: fixed with a note in `_sort_block_descending` explaining parity with the eager `-np.sort(-data)` expression.

No remaining findings. 183 tests pass locally, including the GPU-gated cases on a CUDA host.

Issue #3152, opened 2026-06-10
mcda memory: monte_carlo sensitivity materializes the full dask dataset; constrain() deep-copies its input


brendancol:
Two memory findings from the performance sweep against `xrspatial/mcda`.

## 1. `sensitivity(method="monte_carlo")` calls `criteria.compute()` on dask input

`_monte_carlo` materializes the whole dataset up front. The comment next to it is right that the alternative it considered (chaining ~1000 lazy combine graphs) is worse, but the cure means the function only works when the full criteria stack fits in host memory. On a larger-than-memory raster the `compute()` call OOMs.

A chunk-wise version avoids both problems: draw all `n_samples` weight vectors up front (they are tiny), stack the criterion layers, and run the sample loop inside one `map_blocks` chunk function with Welford accumulation per block. Graph size stays at one task per chunk and peak memory is bounded by chunk size. The numpy path can share the same per-block kernel, so the two backends produce identical values for the same seed.

Note this changes the dask return type from eagerly-computed numpy to a lazy dask array, which is the point. `test_monte_carlo_dask_returns_numpy` codifies the eager behavior and needs updating with it.

## 2. `constrain()` deep-copies the suitability raster

`constrain` starts with `result = suitability.copy()` (deep by default). The copy is never mutated: `xr.where` returns new objects, and the only in-place change is the `.name` assignment at the end. On the eager path that is a full extra allocation and copy of the input for nothing. A shallow copy keeps the name assignment from leaking into the caller's object (the empty `exclude` case) without duplicating the buffer.

Issue #3151, opened 2026-06-10
mcda GPU paths fail: piecewise standardize on cupy, piecewise/categorical on dask+cupy, monte_carlo sensitivity on cupy


brendancol:
Found by the performance sweep against `xrspatial/mcda`, verified on a CUDA host (cupy 13.6). `standardize()` documents support for numpy, cupy, dask+numpy, and dask+cupy, but three GPU paths fail outright.

## 1. `standardize(method="piecewise")` on cupy input

`_piecewise` calls `xp.interp(data, bp, vl)` with `xp = cupy` while `bp` and `vl` are numpy arrays:

```
NotImplementedError: Only int or ndarray are supported for a
```

## 2. `standardize(method="piecewise" | "categorical")` on dask+cupy input

The `map_blocks` chunk functions call `np.asarray(block)` with a comment claiming this "handles cupy chunks". It does not; cupy refuses implicit conversion:

```
TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.
```

Even if the conversion worked, it would be a device-to-host copy per chunk with the interpolation running on the CPU and the result left there.

## 3. `sensitivity(method="monte_carlo")` on cupy input

`_monte_carlo` reads `score.values` every iteration, which hits the same implicit-conversion TypeError on cupy-backed data. Were the conversion allowed, it would be one device-to-host transfer per sample (1000 by default) with the Welford accumulation running on the host.

## Proposed fixes

- piecewise on cupy: hand `cupy.interp` cupy copies of the breakpoint and value tables so everything stays on device.
- piecewise/categorical chunk functions: detect cupy blocks and use the matching array module instead of converting to numpy.
- monte_carlo: accumulate with the array's own module (`score.data` rather than `.values`) and convert to numpy once at the end.

PR #3159, opened 2026-06-10
Keep standardize piecewise/categorical on the device for cupy inputs

brendancol:
Fixes the `standardize()` part of #3151 (items 1 and 2). The `monte_carlo` sensitivity crash from that issue is handled separately together with #3152, since both need the same `_monte_carlo` rework.

- `_piecewise` handed numpy lookup tables to `cupy.interp`, which rejects them (`NotImplementedError`). The tables are now converted with the data's own array module (a no-op for numpy), so plain-cupy piecewise works and stays on the device.
- The `_piecewise` and `_categorical` dask chunk functions called `np.asarray(block)` with a comment claiming it handles cupy chunks. cupy refuses implicit conversion, so dask+cupy raised TypeError. The chunk functions now pick the block's own array module; cupy blocks are interpolated/mapped on the GPU and the result keeps its cupy meta.

Backend coverage: numpy and dask+numpy unchanged (same code path, `xp` is numpy); cupy piecewise and dask+cupy piecewise/categorical go from crashing to working, verified on a CUDA host with results staying in device memory end to end.

Test plan:
- [x] `pytest xrspatial/tests/test_mcda.py` (178 passed)
- [x] New `TestStandardizeCupy` tests gated on `cuda_and_cupy_available`: cupy piecewise, dask+cupy piecewise, dask+cupy categorical, each asserting parity with the numpy result and that the output is still a cupy array

### DIFF
--- a/xrspatial/mcda/standardize.py
+++ b/xrspatial/mcda/standardize.py
@@ -268,15 +268,18 @@ def _piecewise(data, *, breakpoints, values):
         import dask.array as da
 
         def _interp_block(block):
-            # Ensure block is numpy for np.interp (handles cupy chunks)
-            arr = np.asarray(block)
-            return np.interp(arr, bp, vl)
+            # Use the block's own array module so cupy chunks stay on
+            # the device (np.asarray on a cupy block raises TypeError).
+            xp_b = _get_xp(block)
+            return xp_b.interp(block, xp_b.asarray(bp), xp_b.asarray(vl))
 
         result = da.map_blocks(_interp_block, data, dtype=np.float64)
         result = da.where(da.isfinite(data), result, np.nan)
         return result
 
-    result = xp.interp(data, bp, vl)
+    # cupy.interp rejects numpy operands, so move the lookup tables to
+    # the data's array module first (a no-op copy for numpy).
+    result = xp.interp(data, xp.asarray(bp), xp.asarray(vl))
     result = xp.where(xp.isfinite(data), result, xp.nan)
     return result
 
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Keep standardize piecewise/categorical on the device for cupy inputs

None.

- [ ] `xrspatial/mcda/standardize.py:272` (`_interp_block`): the breakpoint/value tables are converted with `xp_b.asarray` inside the chunk function, so cupy chunks re-upload the (tiny) tables on every block. Converting once outside the closure, based on the dask array's `_meta`, removes the per-chunk transfer. Same applies to `keys`/`vals` in `_apply_mapping`. Minor, since the tables are a few dozen bytes, but it is one line each.

- [ ] `xrspatial/tests/test_mcda.py` (TestStandardizeCupy): the `raw` fixture is 2x4 with (1, 2) chunks, which exercises multiple blocks but not ragged chunking. Not critical for elementwise ops.

### What looks good
- The fix picks the array module per block instead of guessing, so dask+numpy keeps its existing behavior byte for byte while dask+cupy stops crashing.
- All three failure modes from #3151 items 1-2 have device-residency assertions (`isinstance(..., cupy.ndarray)`), not just value parity, so a silent fallback to host arrays would fail the test.

### Checklist
- [x] Algorithm matches reference (np.interp/cupy.interp parity asserted)
- [x] All implemented backends produce consistent results (numpy path untouched; GPU paths verified on a CUDA host)
- [x] NaN handling is correct (NaN cell in the fixture, equal_nan parity)
- [x] Edge cases are covered by tests (unmapped categorical values stay NaN via parity with the numpy reference)
- [x] Dask chunk boundaries handled correctly (elementwise per block)
- [x] No premature materialization or unnecessary copies (apart from the per-chunk table upload noted above)
- [x] README feature matrix: standardize GPU cells already marked advanced; no tier change

brendancol (commented):
Follow-up review after the second commit:

- Per-chunk table upload: fixed. `_piecewise` and `_categorical` now resolve the array module once from `data._meta` and convert the lookup tables a single time outside the chunk closure.
- Ragged-chunk nit: the GPU fixture now chunks (1, 3) on the 2x4 grid, giving an uneven 3+1 split.

No remaining findings. 178 tests pass locally, including the three GPU-gated cases on a CUDA host.

PR #3160, opened 2026-06-10
Make monte_carlo sensitivity chunk-bounded on dask and drop constrain() deep copy

### LINKED ISSUE #3151: mcda GPU paths fail: piecewise standardize on cupy, piecewise/categorical on dask+cupy, monte_carlo sensitivity on cupy

### LINKED ISSUE #3152: mcda memory: monte_carlo sensitivity materializes the full dask dataset; constrain() deep-copies its input

brendancol:
Closes #3152
Closes #3151 (the `monte_carlo` item; the `standardize` items are fixed in #3159)

- `_monte_carlo` called `criteria.compute()` on dask input, materializing the whole dataset before sampling. The sample loop now runs inside a single `map_blocks` chunk function: weight vectors are drawn up front (sequentially, so a given seed reproduces the old sample stream), criterion layers are stacked per chunk, and Welford accumulation happens per block. The graph is ~8 tasks per chunk regardless of `n_samples` (measured with n_samples=1000 on a 2560x2560 / 256-chunk dataset) and peak memory is bounded by chunk size.
- The dask result is now lazy instead of eagerly-computed numpy. `test_monte_carlo_dask_returns_numpy` codified the old behavior and is replaced by lazy/parity/NaN/validation tests. Numpy results for a given seed are unchanged (same draws, same per-pixel arithmetic).
- The same kernel also fixes the cupy crash from #3151: the old loop read `score.values` every iteration, which cupy rejects. Accumulation now runs on the device and the result stays a cupy array.
- `constrain()` deep-copied the suitability raster even though nothing mutates it (`xr.where` returns new objects). It now takes a shallow copy, which still keeps the `.name` assignment from leaking into the caller's object.

Backend coverage: numpy values unchanged for a given seed; dask+numpy goes from materialize-everything to chunk-bounded and lazy; cupy and dask+cupy go from crashing to working, verified on a CUDA host with parity against numpy.

Test plan:
- [x] `pytest xrspatial/tests/test_mcda.py` (181 passed)
- [x] New dask tests: lazy result with small graph, wlc/wpm parity vs numpy with same seed, NaN propagation, missing-weight validation
- [x] New cupy test (gated on `cuda_and_cupy_available`): MC parity vs numpy, result stays on device
- [x] dask+cupy MC verified manually for wlc and wpm (parity vs numpy)
- [x] Graph probe: 800 tasks for 100 chunks at n_samples=1000 (no scaling with sample count)
- [x] New constrain test: output rename does not leak into the input

### DIFF
--- a/README.md
+++ b/README.md
@@ -514,7 +514,7 @@ For a broader catalog of spectral indices and sensor-specific band combinations,
 | [Fuzzy Overlay](xrspatial/mcda/combine.py) | Combines criteria using fuzzy set operators (AND, OR, sum, product, gamma) | Eastman 1999 | ✅ | 🔼 | 🧪 | 🧪 |
 | [Boolean Overlay](xrspatial/mcda/combine.py) | Combines binary criterion masks using AND/OR logic | Standard | ✅ | 🔼 | 🧪 | 🧪 |
 | [Constrain](xrspatial/mcda/constrain.py) | Masks exclusion zones from a suitability surface | Standard | ✅ | 🔼 | 🧪 | 🧪 |
-| [Sensitivity](xrspatial/mcda/sensitivity.py) | Assesses weight stability via one-at-a-time or Monte Carlo perturbation | Standard | ✅ | 🔼 | 🚫 | 🚫 |
+| [Sensitivity](xrspatial/mcda/sensitivity.py) | Assesses weight stability via one-at-a-time or Monte Carlo perturbation | Standard | ✅ | 🔼 | 🧪 | 🧪 |
 
 -------
 
--- a/xrspatial/mcda/constrain.py
+++ b/xrspatial/mcda/constrain.py
@@ -35,7 +35,11 @@ def constrain(
     if name is None:
         name = suitability.name
 
-    result = suitability.copy()
+    # Shallow copy: xr.where returns new objects and nothing below
+    # mutates the data, so a deep copy would only duplicate the full
+    # raster. The copy still keeps the .name assignment from leaking
+    # into the caller's object when exclude is empty.
+    result = suitability.copy(deep=False)
     for mask in exclude:
         result = xr.where(mask, fill, result)
 
--- a/xrspatial/mcda/sensitivity.py
+++ b/xrspatial/mcda/sensitivity.py
@@ -5,7 +5,14 @@ from __future__ import annotations
 import numpy as np
 import xarray as xr
 
-from xrspatial.mcda.combine import wlc, wpm
+from xrspatial.mcda.combine import (
+    _check_wpm_positive,
+    _validate_criteria,
+    _validate_weights,
+    wlc,
+    wpm,
+)
+from xrspatial.mcda.standardize import _get_xp
 
 
 def sensitivity(
@@ -46,7 +53,9 @@ def sensitivity(
         For ``"one_at_a_time"``: Dataset with per-criterion DataArrays
         showing the absolute score difference under perturbation.
         For ``"monte_carlo"``: DataArray of per-pixel coefficient of
-        variation across random weight samples.
+        variation across random weight samples. Dask-backed criteria
+        produce a lazy dask-backed result; the sample loop runs inside
+        each chunk, so memory stays bounded by chunk size.
     """
     combine_fn = {"wlc": wlc, "wpm": wpm}.get(combine_method)
     if combine_fn is None:
@@ -63,7 +72,7 @@ def sensitivity(
                 f"n_samples must be a positive integer >= 1, got {n_samples!r}"
             )
         return _monte_carlo(
-            criteria, weights, combine_fn, int(n_samples), seed, name
+            criteria, weights, combine_method, int(n_samples), seed, name
         )
     else:
         raise ValueError(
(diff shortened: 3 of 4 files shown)


brendancol (commented):
## PR Review: Make monte_carlo sensitivity chunk-bounded on dask and drop constrain() deep copy

None.

- [ ] `xrspatial/mcda/sensitivity.py:222` (`_monte_carlo` dask branch): the mixed-input path (`da.from_array` for numpy-backed variables inside an otherwise dask Dataset) has no test. A Dataset with one chunked and one numpy variable would cover it.
- [ ] `xrspatial/tests/test_mcda.py` (TestSensitivityDask): chunk sizes are even 10x10 everywhere; a ragged-chunk parity case would pin the stack/rechunk behavior.

- [ ] `xrspatial/mcda/sensitivity.py:158` (`_mc_block`): the wlc/wpm accumulation re-implements the combine arithmetic on raw arrays. A short comment pointing at `combine.wlc`/`combine.wpm` as the canonical definitions would help keep them in sync if the combine semantics ever change.

### What looks good
- The Welford accumulation now runs per block with weights drawn up front, so the graph is ~8 tasks per chunk at n_samples=1000 and nothing materializes the criteria stack. That is the actual fix for the OOM, not a bigger buffer.
- Sequential dirichlet draws keep seed-for-seed parity with the old implementation on numpy, and the dask/numpy parity test asserts it.
- The same kernel removes the per-sample `.values` device round trip, which is why the cupy path now works at all; the test asserts the result stays a cupy array.
- `constrain` keeps the rename-isolation property (new test) without the full-raster deep copy.
- README sensitivity row updated to reflect the new GPU support tier.

### Checklist
- [x] Algorithm matches reference (CV via Welford; per-pixel arithmetic identical to the previous loop)
- [x] All implemented backends produce consistent results (numpy/dask in tests; cupy gated test; dask+cupy verified manually for wlc and wpm)
- [x] NaN handling is correct (dedicated NaN propagation test on dask)
- [x] Edge cases are covered by tests (missing-weight validation preserved; single-sample and single-criterion cases still pass)
- [x] Dask chunk boundaries handled correctly (criterion axis rechunked to one chunk before map_blocks, drop_axis=0)
- [x] No premature materialization or unnecessary copies (criteria.compute() removed; deep copy removed)
- [x] README feature matrix updated (sensitivity GPU cells)
- [x] Docstrings present and accurate (Returns section documents the lazy dask contract)

brendancol (commented):
Follow-up review after the third commit:

- Mixed-input path: fixed. New `test_monte_carlo_mixed_numpy_and_dask_vars` covers a Dataset with one chunked and one numpy-backed variable, asserting a lazy result with numpy parity.
- Ragged chunks: the dask/numpy parity test is now parametrized over 10x10 and 7x9 chunkings, for both wlc and wpm.
- Kernel sync nit: `_mc_block` now carries a comment pointing at `combine.wlc`/`combine.wpm` as the canonical definitions.

No remaining findings. 184 tests pass locally; the cupy-gated test ran on a CUDA host.

PR #3169, opened 2026-06-10
Fix compression corpus fixture path so oracle tests actually run (#3164)

### LINKED ISSUE #3164: Golden-corpus compression oracle tests never run: FIXTURES_DIR resolves to tests/golden_corpus/golden_corpus/fixtures
`tests/golden_corpus/test_compression.py:46-50` builds the fixture path as:

```python
FIXTURES_DIR = (
    pathlib.Path(__file__).resolve().parent
    / "golden_corpus"
    / "fixtures"
)
```

`__file__` already lives in `tests/golden_corpus/`, so this resolves to the nonexistent `tests/golden_corpus/golden_corpus/fixtures`. All 6 compression fixtures (none, deflate-pred2, deflate-pred3, lzw-pred2, lerc, jpeg-ycbcr) skip on every machine, including CI, with the reason "fixture ... not present (codec unavailable?)", even though all 6 .tif files are committed under `fixtures/`.

The other corpus test files get it right, e.g. `test_eager_numpy.py:82` uses `Path(generate.__file__).parent / "fixtures"`.

I checked that the tests pass once pointed at the correct directory, so this is a dead release gate rather than a hidden failure. One-line fix, but worth landing soon: skips that read as codec availability mask the fact that the codec parity gate has never actually gated anything.

brendancol:
Closes #3164

`test_compression.py` built `FIXTURES_DIR` as `Path(__file__).resolve().parent / "golden_corpus" / "fixtures"`. The file already lives in `golden_corpus/`, so the path pointed at the nonexistent `golden_corpus/golden_corpus/fixtures` and all 6 compression oracle tests skipped everywhere, with "fixture ... not present (codec unavailable?)" as the reason.

This drops the doubled segment so the path resolves to the committed `fixtures/` directory, where the sibling corpus test files already look.

Backend coverage: not applicable, test-infrastructure only.

Test plan:
- [x] `pytest xrspatial/geotiff/tests/golden_corpus/test_compression.py` -- 15 passed, 0 skipped (the 6 oracle tests skipped before this change)


brendancol (commented):
## PR Review: Fix compression corpus fixture path so oracle tests actually run (#3164)

### Blockers
None.

### Suggestions
None.

### Nits
None. One observation, not a change request: most corpus test files anchor the fixtures path with `Path(generate.__file__)`, while this fix uses the bare `__file__` form. Both resolve to the same directory since the test files sit next to `generate.py`, and `test_corpus_determinism.py:58-60` already uses the `__file__` form, so there is precedent for both. Switching would mean adding a module-level `generate` import that this file otherwise gets via `importlib` inside two tests, which isn't worth it for a path that can't differ.

### What looks good
- The fix is the minimal one: drop the doubled `"golden_corpus"` segment. The diff touches nothing else.
- Checked all 12 golden-corpus test files for the same pattern; `test_compression.py` was the only one pointing at `golden_corpus/golden_corpus/fixtures`.
- Ran the file locally: 15 passed, 0 skipped. Before the fix the 6 oracle tests skipped with the "codec unavailable?" reason.
- The defensive skip in `_fixture_path` is now reachable only when a fixture file is genuinely absent, which is what its docstring describes.

### Checklist
- Algorithm / backend / NaN / dask / benchmark items: N/A, this is a test-infrastructure path fix with no library code changes.
- Edge cases covered: the oracle tests themselves are the coverage being restored; no new tests needed.
- Docstrings / README: N/A, no public API change.

PR #3170, opened 2026-06-10
Fix README to_geotiff example that used the gated JPEG codec (#3162)

### LINKED ISSUE #3162: README GeoTIFF example to_geotiff(compression='jpeg') raises ValueError on default settings
The README's GeoTIFF section (README.md:184) shows this as a headline example:

```python
to_geotiff(data, 'ortho.tif', compression='jpeg')    # JPEG for orthophotos
```

Running it raises:

```
ValueError: compression='jpeg' is not supported: the encoder writes self-contained
JFIF streams without the required JPEGTables tag (347)...
```

JPEG is internal_only tier and requires `allow_internal_only_jpeg=True` (`_attrs.py:291`, `_writers/eager.py:302-315`). The gate is intentional, so the bug is in the README: the front-page example crashes and implies JPEG output is interoperable when the contract says it is not.

Either switch the example to a supported codec (deflate/zstd) or show the opt-in with a note about the interoperability limitation.

brendancol:
Closes #3162

- The README's GeoTIFF write example used `compression='jpeg'`, which raises `ValueError` on default settings because the JPEG encoder is internal-only and gated behind `allow_internal_only_jpeg=True`.
- Replaced it with `compression='zstd'`, which works out of the box.
- Marked the JPEG entry in the codec list as internal-only (opt-in flag named, not GDAL-readable) so the list no longer implies interoperable JPEG output.

Docs-only change to README.md. No code, tests, or backend behavior touched, so there is no backend-coverage impact.

Test plan:
- [x] Reproduced the `ValueError` from the issue with the old example
- [x] Ran the new example: `to_geotiff(data, 'out.tif', compression='zstd')` writes and the file reads back with matching values
- [x] Verified the codec-list note is accurate: `compression='jpeg'` with `allow_internal_only_jpeg=True` writes (with `GeoTIFFFallbackWarning`) and reading the result also requires the flag


brendancol (commented):
## PR Review: Fix README to_geotiff example that used the gated JPEG codec (#3162)

None.

- [ ] README.md:203 now flags JPEG as internal-only but still lists JPEG 2000 (glymur) with no qualifier. `compression='jpeg2000'` is gated too: it raises `ValueError` unless `allow_experimental_codecs=True` (experimental tier, see `xrspatial/geotiff/_writers/eager.py` docstring for `allow_experimental_codecs`). I verified this by running the writer with each listed codec; deflate, lzw, zstd, and packbits work on defaults, jpeg2000 does not. The edited line fixes the misleading entry for JPEG but leaves the same problem in place for JPEG 2000. Suggest tagging it, e.g. "JPEG 2000 (glymur, experimental: requires `allow_experimental_codecs=True`)".

None. Comment alignment in the example block (column 54) matches the neighboring lines.

### What looks good
- The replacement example runs on default settings: zstd write and read-back round-trip with matching values.
- The internal-only note for JPEG matches the actual gate behavior. The write needs `allow_internal_only_jpeg=True` and emits `GeoTIFFFallbackWarning`; reading the result back requires the same flag.
- The diff is two lines, scoped to exactly what the issue describes.

### Checklist
- [x] README example verified runnable on defaults
- [x] Codec-list claims checked against writer behavior
- [ ] Algorithm / backend parity / NaN handling / dask chunking / benchmarks: N/A, docs-only change with no code touched

brendancol (commented):
## Follow-up review (after 7e7eac3e)

The one suggestion from the first review is addressed: README.md:203 now tags JPEG 2000 as experimental with the `allow_experimental_codecs=True` flag named. I checked the claim by running the writer: `compression='jpeg2000'` raises `ValueError` without the flag and writes (with `GeoTIFFFallbackWarning`) with it, on uint8 input. The remaining codecs in the list (deflate, lzw, zstd, packbits, uncompressed) all work on defaults, so the list now matches actual writer behavior.

No new findings. No blockers, suggestions, or nits remain.

PR #3171, opened 2026-06-10
Fix to_geotiff dask+cupy crash in CPU streaming writer (gpu=False)

### LINKED ISSUE #3165: to_geotiff on dask+cupy with gpu=False crashes: streaming writer calls np.asarray before the cupy .get() handoff
Both segment loops in the streaming writer do the conversion in the wrong order (`_writer.py:1141-1148` for tiled segments, `1241-1247` for strips):

```python
seg_np = np.asarray(dask_data[r0:r1, seg_c0:seg_c1].compute())
if hasattr(seg_np, 'get'):
    seg_np = seg_np.get()
```

`cupy.ndarray.__array__` raises `TypeError: Implicit conversion to a NumPy array is not allowed`, so `np.asarray` throws before the `.get()` branch can run. The handoff is dead code, and any dask-of-cupy array routed to the CPU streaming writer crashes.

Repro: `to_geotiff(dask_cupy_da, path, gpu=False)` raises the TypeError above. `gpu=False` is the documented way to force the CPU writer, and the docstring promises dask-backed inputs are written in streaming mode.

The VRT per-tile writer gets the order right (`_writers/eager.py:1107-1112`): compute first, `.get()` if the chunk is cupy, then `np.asarray`. The streaming loops should do the same.

brendancol:
Closes #3165

Both streaming segment loops in `_writer.py` (tiled segments and strips) did `np.asarray(...compute())` before the `hasattr(seg, 'get')` check. `cupy.ndarray.__array__` raises `TypeError`, so the `.get()` branch was dead code and `to_geotiff(dask_cupy_da, path, gpu=False)` crashed. Reordered both loops to do what the VRT per-tile writer in `_writers/eager.py` already does: compute, `.get()` if the chunk is cupy, then `np.asarray`.

The GPU writer path (`gpu=True`) is untouched; numpy and dask+numpy behavior is unchanged.

Test plan:
- [x] New GPU-marked regression test writes a dask+cupy DataArray via `to_geotiff(gpu=False)` for both layouts (`tiled=True`/`False`) and round-trips it, including the NaN-to-nodata-sentinel path
- [x] Test fails on main (TypeError in both layouts), passes with the fix
- [x] `tests/write`, `tests/gpu/test_writer.py`, `tests/test_round_trip.py`: 1212 passed, 1 skipped


brendancol (commented):
## PR Review: Fix to_geotiff dask+cupy crash in CPU streaming writer (gpu=False)

- None.

- [ ] The `ndim == 3` branches of both loops (`xrspatial/geotiff/_writer.py:1136-1139` tiled, `1241-1242` strips) carry the same reorder but the regression test only exercises 2D input. A multiband dask+cupy case (e.g. a `(y, x, band)` array through the same `to_geotiff(gpu=False)` call) would pin those branches too.

- None.

### What looks good
- The conversion order now matches the VRT per-tile writer in `_writers/eager.py:1107-1112`, so all dask-to-disk paths agree.
- `np.asarray` after `.get()` is a no-op for numpy chunks; dask+numpy output is byte-identical to before.
- The test fails on main with the exact TypeError from the issue and covers both layouts, a chunk grid that forces multiple cupy->numpy handoffs per segment, and the NaN-to-sentinel rewrite on cupy-origin data.
- Temp filenames carry the issue number, and the test uses the shared `requires_gpu` marker.

### Checklist
- [x] Fix matches the reference implementation (eager.py per-tile writer)
- [x] dask+numpy behavior unchanged; GPU writer path untouched
- [x] NaN handling correct (sentinel restore runs on host arrays)
- [ ] 3D/multiband dask+cupy path untested (see suggestion)
- [x] No premature materialization beyond the per-segment compute the streaming writer already does
- [x] Benchmark not needed (bug fix, no perf-sensitive change)
- [x] No docstring/README changes needed (no API or backend-matrix change)

brendancol (commented):
Follow-up review after 41ebdf1d:

The one suggestion from the first pass is addressed. The new multiband test pushes a (y, x, band) dask+cupy array through both layouts of the streaming writer with gpu=False, so the ndim == 3 branches at _writer.py:1136-1139 and 1241-1242 are now pinned alongside the 2D ones. Confirmed locally that the new test fails on the pre-fix writer (TypeError in both layouts) and passes with the fix.

No remaining findings. All four 3165 regression tests pass on real hardware.

Issue #3166, opened 2026-06-10
dask+cupy writes auto-dispatch to the GPU writer and materialize the full array, contradicting the streaming contract


brendancol:
The `to_geotiff` docstring says dask-backed DataArrays are "written in streaming mode: one tile-row at a time, without materialising the full array into RAM" (`_writers/eager.py:102-109`). That is not what happens for dask+cupy input.

`_is_gpu_data` detects dask-of-cupy via `_meta` (`_backends/_gpu_helpers.py:23-37`), and the `use_gpu` dispatch at `eager.py:633`/`757` runs before the dask streaming branch at `918`. So dask+cupy auto-routes to `_write_geotiff_gpu`, which calls `data.compute()` (`_writers/gpu.py:565`) and materializes the whole array on device. Verified by execution: the write succeeds eagerly with no warning, and `streaming_buffer_bytes` does nothing on this path.

The no-op is documented only in the private `_write_geotiff_gpu` docstring (`gpu.py:207-210`). The public docs say `streaming_buffer_bytes` is "Only relevant for dask-backed inputs", which is exactly the input type where it gets ignored.

For out-of-core GPU pipelines this defeats the point of chunking. Short term, the public docstring should state that dask+cupy writes materialize on device and that `streaming_buffer_bytes` does not apply; longer term the GPU writer could stream per block. Related to the gpu=False escape hatch being broken, filed separately.

PR #3173, opened 2026-06-10
Warn when the GPU writer materializes dask input; scope the streaming docs to the CPU path

brendancol:
Refs #3166 (short-term fix; per-block GPU streaming stays open).

The `to_geotiff` docstring says dask-backed input is written in streaming mode. That only holds on the CPU path: dask+cupy auto-dispatches to `_write_geotiff_gpu`, which calls `.compute()` and materializes the full array on device, and `streaming_buffer_bytes` is silently ignored there.

- Docstring now scopes the streaming contract to the CPU path and documents that dask+cupy materializes on device with `streaming_buffer_bytes` ignored.
- `_write_geotiff_gpu` emits a `GeoTIFFFallbackWarning` from both of its `.compute()` sites (DataArray and positional input) when dask-backed data gets materialized.
- No behavior change beyond the warning.

Per-block GPU streaming is the long-term item and out of scope here, so #3166 stays open.

Backends: the warning fires on the gpu and dask+gpu write paths; numpy and dask+numpy are untouched.

Test plan:
- [x] New tests in `xrspatial/geotiff/tests/gpu/test_writer.py`: warning fires for dask+cupy via `to_geotiff` auto-dispatch and for positional dask input via `_write_geotiff_gpu`; no warning for plain cupy (ran on a CUDA device)
- [x] `gpu/test_writer.py` + `write/test_streaming.py` pass locally (185 passed, 1 skipped)
- [x] flake8 clean on edited files


brendancol (commented):
## PR Review: Warn when the GPU writer materializes dask input; scope the streaming docs to the CPU path

None.

- [ ] `xrspatial/geotiff/_writers/gpu.py:35`: the `GeoTIFFFallbackWarning` class docstring (`_runtime.py:114-121`) says `XRSPATIAL_GEOTIFF_STRICT=1` promotes these warnings to exceptions, but promotion happens per call site and this new site never checks `_geotiff_strict_mode()`. Skipping promotion is correct here: there's no opt-out flag, so strict mode would turn every dask+cupy GPU write into a hard failure, and the jpeg / experimental-codec opt-in warnings already skip it the same way. But the helper docstring should say this out loud so the next strict-mode audit doesn't "fix" it.
- [ ] `xrspatial/geotiff/tests/gpu/test_writer.py:2287`: the public docstring claims "The same applies when `gpu=True` is passed with any dask-backed input", and the tests don't cover that combination. dask+cupy gets there via `_is_gpu_data`; dask+numpy with `gpu=True` gets there via forced dispatch. A small fourth test would pin the claim.

- [ ] `xrspatial/geotiff/_writers/gpu.py:42-46`: worth a line in the helper docstring noting the warning stays truthful even when the GPU write later falls back to CPU, since the compute has already run by the time the fallback fires. Easy spot for a future reader to misread.

### What looks good
- The warning sits after the `import cupy` gate, so cupy-less hosts hit the CPU streaming fallback without a spurious materialise warning.
- `stacklevel=3` from the helper lands on the same frame the file's existing `stacklevel=2` calls do from the function body.
- Both compute sites (DataArray and positional) warn and have tests; the plain-cupy negative test pins that eager device data stays silent.
- The docstring updates agree with each other across the streaming paragraph, the `gpu` param, and `streaming_buffer_bytes`, and the GPU writer's own param doc points at the warning.
- Tmp files carry the issue number per project convention.

### Checklist
- [x] Algorithm matches reference/paper (n/a, doc + warning change)
- [x] All implemented backends produce consistent results (no numeric change)
- [x] Edge cases are covered by tests (both compute sites + negative case; dask+numpy `gpu=True` untested, see suggestion)
- [x] No premature materialization or unnecessary copies introduced (the materialization is pre-existing; this PR only makes it visible)
- [x] README feature matrix updated (no backend support change)

brendancol (commented):
## Follow-up review (after d470c737)

- The `_warn_dask_materialised` docstring now explains why this site skips strict-mode promotion (no opt-out flag, same as the jpeg / experimental-codec warnings).
- `test_to_geotiff_dask_numpy_gpu_true_warns_3166` covers the forced-dispatch path (`gpu=True` with dask+numpy), so the docstring claim is now tested.
- The docstring also notes the warning stays accurate when the write later falls back to CPU.

All four #3166 tests pass on a CUDA device, flake8 is clean. Nothing new to flag.

brendancol:
The failing pytest check is main breakage, not this PR: the strict xfail added in #3118 now XPASSes after #3111 landed, so every merge-ref CI run is red. Filed #3178 with the details. Once that's fixed on main, a rerun of the pytest workflow here should go green; the geotiff suites this PR touches pass locally and in the corpus/cog-validator checks.

PR #3174, opened 2026-06-10
Thread the nodata kwarg into _pack's NaN fill so pack=True holes match GDAL_NODATA (#3168)

### LINKED ISSUE #3168: to_geotiff(pack=True, nodata=X) fills NaN holes with the attrs sentinel but tags GDAL_NODATA with X
`_pack` runs before the `nodata` kwarg is consulted (`_writers/eager.py:398-403`), and its NaN fill uses the sentinel from the array attrs (`_attrs.py:1702-1707`). The GDAL_NODATA tag is then written from the kwarg, per the documented precedence "nodata kwarg overrides attrs" (`_validation.py:1133-1139`).

Repro: read with `unpack=True` from a file whose sentinel is -32768, then write with `to_geotiff(da, path, pack=True, nodata=-1)`. The NaN holes are filled with -32768 (from attrs, inside `_pack`) while the tag says -1. A `masked=True` re-read treats the -32768 holes as valid pixels.

The trigger is narrow (the caller has to pass a nodata that conflicts with the attrs sentinel), but the two halves of one write disagree and the result is an internally inconsistent file. Either thread the kwarg into `_pack`'s fill step or raise on the conflict.

brendancol:
Closes #3168

`to_geotiff(pack=True, nodata=X)` filled NaN holes with the attrs sentinel inside `_pack`, but wrote the GDAL_NODATA tag from the kwarg. The two halves of one write disagreed, so a masked re-read returned the holes as valid pixels.

- `_pack` now takes the kwarg and uses it as the fill value, matching the "kwarg overrides attrs" precedence the tag already follows.
- The pack step runs after `_validate_nodata_arg` now, so a bad kwarg gets rejected before it's used as a fill value. An override that doesn't fit the packed integer dtype (out of range, or fractional) raises instead of wrapping in the cast; wrapping would just recreate the same fill-vs-tag mismatch.
- The pre-v5 dtype-width fallback still keys off the attrs sentinel, since that carries the source dtype. The kwarg is a plain Python number with no width information.

The fix sits in the shared pre-dispatch pack step, so every write path (eager, GPU, VRT, streaming) fills with the same value the tag declares. Tests run on numpy and dask, same as the existing pack tests (`unpack=True` is a CPU read feature).

Test plan:
- [x] unpack read with sentinel -32768, pack write with nodata=-1: holes carry -1, tag says -1, masked re-read masks them (numpy + dask)
- [x] SCALE/OFFSET round trip with a swapped sentinel still unpacks to the original decoded values
- [x] no kwarg: attrs sentinel still drives both fill and tag
- [x] out-of-range / fractional override raises
- [x] full xrspatial/geotiff suite: 6265 passed, 81 skipped


brendancol (commented):
## PR Review: Thread the nodata kwarg into _pack's NaN fill (#3168)

### Blockers
None.

### Suggestions
- [ ] `xrspatial/geotiff/tests/write/test_pack_nodata_kwarg_3168.py` only exercises integer pack targets. The float path takes a different branch in `_pack` (the range guard at `_attrs.py:1716` is int-only, and the fill happens without a cast), so a float-source case is worth pinning: read a float32 raster with sentinel -9999 via `unpack=True`, write with `pack=True, nodata=-5`, check the hole carries -5, the tag says -5, and a masked re-read masks it. I ran this by hand and it behaves, but nothing in the suite would catch a regression there.

### Nits
- [ ] The `Raises` paragraph of the `_pack` docstring (`_attrs.py:1685-1687`) still lists only the two pre-existing ValueErrors. The new unrepresentable-kwarg rejection at `_attrs.py:1725` is documented only in a code comment; add it to the docstring.
- [ ] When the kwarg overrides the attrs sentinel, the DataArray returned by `_pack` keeps the old `attrs['nodata']` (say -32768) while its pixels now hold the override (-1). This is harmless today: every dispatch path (`eager.py:441` -> `737`/`779`/`961`/`1062`) receives the kwarg, which wins before the attr is ever consulted. But the intermediate is internally inconsistent, and a future write path that forgets to thread the kwarg would regress silently. Setting `attrs['nodata']` to the fill value in `_pack` when the override is given would close that off. Low confidence this matters in practice; flagging it rather than insisting.

### What looks good
- The fix lives in the shared pre-dispatch pack step (`eager.py:436-441`), so eager, GPU, VRT, and streaming writes all fill with the value the tag declares.
- Moving the pack block below `_validate_nodata_arg` means a bad kwarg is rejected with the standard message instead of failing inside `fillna`.
- The range guard (`_attrs.py:1716-1730`) stops an out-of-range or fractional override from wrapping in the `astype` and quietly recreating the fill-vs-tag mismatch the PR fixes.
- The pre-v5 dtype-width fallback (`_attrs.py:1708-1713`) still keys off the attrs sentinel, which is the only place width information exists. Threading the kwarg into it would have been wrong.
- Tests cover numpy and dask, the masked re-read from the issue repro, the SCALE/OFFSET round trip, the no-kwarg default, and the guard. Tmp filenames carry the issue number.

### Checklist
- [x] Algorithm matches the documented kwarg-overrides-attrs precedence
- [x] All write paths see the same packed array (pack runs before dispatch)
- [x] NaN handling is correct (holes fill with the tag value; NaN kwarg on int targets is rejected by the range guard)
- [ ] Edge cases: float pack target with an override is untested (see suggestion)
- [x] Dask path stays lazy (fillna with a scalar; the guard uses scalars only)
- [x] Benchmark not needed (metadata bug fix)
- [x] README untouched, correctly (no API or backend change)
- [x] Docstrings updated, minus the Raises nit above

brendancol (commented):
## Follow-up review (after 0a73ebd)

- Float-target coverage: `test_pack_nodata_kwarg_float_target` now pins the kwarg fill on the float branch (no range guard, no cast) on numpy and dask, including the masked re-read.
- The `Raises` paragraph of the `_pack` docstring now lists the unrepresentable-kwarg ValueError.
- `_pack` syncs `attrs['nodata']` to the kwarg on the packed intermediate, so the attr matches the pixels even if a future write path forgets to thread the kwarg.

Checked the new attrs sync against the full geotiff suite (6267 passed); no caller depends on the old stale attr. Nothing further from me.

PR #3176, opened 2026-06-10
Validate compression_level before GPU dispatch and pass it to fallback codecs (#3167)

### LINKED ISSUE #3167: compression_level is ignored and unvalidated on the GPU write path; CPU path validates and applies it
Verified by execution: `to_geotiff(cupy_da, path, compression='zstd', compression_level=999)` succeeds, while the identical numpy call raises `compression_level=999 out of range for zstd (valid: 1-22)`.

Two problems compound here:

1. Validation runs only after the GPU dispatch returns. The range checks at `_writers/eager.py:934-941` and `1038-1045` sit below the GPU branch at `757-790`, so GPU writes never reach them.
2. The level is dropped even where it could apply: `gpu_compress_tiles` takes no level argument (`_gpu_decode.py:2995`), and the GPU writer's CPU-codec fallback paths (lzw/packbits/lerc tiles) call the CPU compressor without the level, so those tiles compress at the default.

The ignore is documented only in the private `_write_geotiff_gpu` docstring (`gpu.py:165-168`, "[experimental] ... currently ignored"). The public `to_geotiff` docstring tags `compression_level` `[stable]` with documented valid ranges and no backend caveat (`eager.py:173-177`).

Output files are valid TIFFs, just compressed at a different level than requested, with bogus levels accepted on one backend and rejected on the other. Suggested fix: hoist the validation above the dispatch so both backends reject bad levels, pass the level through the CPU-codec fallback, and add a GPU caveat to the public docstring for codecs where the GPU encoder cannot honor it.

brendancol:
Closes #3167

`to_geotiff(cupy_da, path, compression='zstd', compression_level=999)` succeeded while the identical numpy call raised, because the range validation sat below the GPU dispatch in `to_geotiff`. The level was also dropped on the GPU writer's CPU-codec fallback tiles.

Changes:

- Hoist the `compression_level` range check above all backend dispatch in `to_geotiff` and mirror it in `_write_geotiff_gpu` for direct callers. The two now-dead copies on the CPU branches are removed. Both backends raise the same `ValueError` now.
- Pass `compression_level` through `gpu_compress_tiles` to the CPU codec fallback, so fallback tiles (deflate/zstd/lz4 when nvCOMP is missing) compress at the requested level instead of the default 6.
- nvCOMP has no level knob. When the device encoder handles the tiles and the caller asked for a level, a `UserWarning` says so instead of the level vanishing silently. The `to_geotiff` and `_write_geotiff_gpu` docstrings now carry the caveat.

Backend coverage: the hoisted check applies to numpy, cupy, dask+numpy, and dask+cupy, since all of them dispatch through it. The level threading only affects the GPU writer's CPU-codec fallback.

Test plan:
- [x] numpy and cupy reject `compression_level=999` with identical error messages (zstd/deflate/lz4)
- [x] `_write_geotiff_gpu` rejects out-of-range levels directly; no partial file left behind
- [x] CPU fallback tiles match `compress(tile, tag, level)` byte-for-byte (nvCOMP monkeypatched out)
- [x] end-to-end GPU fallback write at level 1 vs 9 produces different bytes, both round-trip exactly
- [x] nvCOMP path warns when a level is requested, stays silent when it is None
- [x] geotiff write/gpu/unit/parity/vrt/round-trip suites pass locally on a CUDA machine


brendancol (commented):
## PR Review: Validate compression_level before GPU dispatch and pass it to fallback codecs (#3167)

- None found.

- [ ] `xrspatial/geotiff/_gpu_decode.py:3146-3151` -- the `if compression_level is None` branch runs once per tile inside the fallback loop. `compress()` already defaults to `level=6`, so you can resolve the level once before the loop (`level = 6 if compression_level is None else compression_level`) and always pass it. Drops the per-tile branch and four lines.

- [ ] `xrspatial/geotiff/_gpu_decode.py:3126-3136` -- the nvCOMP ignored-level warning is emitted from `gpu_compress_tiles`, which `_write_geotiff_gpu` calls once per IFD part. Under `-W always`, a `cog=True` write with N overview levels fires it N+1 times. The default filter dedups by location so normal runs see it once. A short comment would save the next reader from wondering.
- [ ] `xrspatial/geotiff/_writers/gpu.py:355-366` -- in `_write_geotiff_gpu` the level gate sits below the experimental-codec opt-in warning, so `compression='lz4', allow_experimental_codecs=True, compression_level=99` warns and then raises. `to_geotiff` raises before that warning. Cosmetic ordering difference between the two entry points.
- [ ] `xrspatial/geotiff/tests/gpu/test_writer.py:2005-2011` -- the no-warning assertion uses `warnings.simplefilter("error")`, which also escalates unrelated warnings (a cupy deprecation, say) into a failure. Recording warnings and asserting none match the nvCOMP message would be sturdier.

### What looks good
- The hoisted check in `eager.py:615-626` runs after the compression-name/type validation (so `compression.lower()` is safe) and before VRT/GPU/streaming/eager dispatch, so all four backends share one gate. I ran the repro: numpy and cupy now raise the identical message for `zstd, level=999`.
- The two removed CPU-branch checks were dead after the hoist (same function, strictly below the new check).
- `_write_geotiff_gpu` mirrors the gate for direct callers, matching the file's existing pattern of duplicated entry-point gates.
- Bool/float `compression_level` values behave exactly as on main (range compare unchanged; an in-range float still fails deeper with TypeError), so no semantics drift outside the fix.
- Tests update the two stale pins that enshrined the old accepted-but-ignored contract, and the new fallback test compares tile bytes against `compress(tile, tag, level)` byte-for-byte instead of just round-tripping.
- Out-of-range tests assert no partial file is left on disk.

### Checklist
- [x] Algorithm matches reference (codec level ranges match `_LEVEL_RANGES`)
- [x] All implemented backends produce consistent results (numpy/cupy parity test compares error strings verbatim; dask paths share the hoisted check)
- [x] NaN handling is correct (untouched by this PR)
- [x] Edge cases are covered by tests (out-of-range both directions, in-range bounds, level=None)
- [x] Dask chunk boundaries handled correctly (n/a -- validation happens before dispatch)
- [x] No premature materialization or unnecessary copies (fallback loop already materialized tiles on main)
- [x] Benchmark exists or is not needed (validation + level threading; not perf-critical)
- [x] README feature matrix updated (n/a -- no new function, no backend support change)
- [x] Docstrings present and accurate (`to_geotiff`, `_write_geotiff_gpu`, `gpu_compress_tiles` all updated)

brendancol (commented):
## Follow-up review (after 226def51)

- Suggestion (fallback loop branch): fixed. `level` resolves once before the loop in `_gpu_decode.py` and `cpu_compress` is always called with it.
- Nit (warning repeats per IFD part): fixed with a comment on the nvCOMP warn site explaining the dedup-by-location behavior and the `-W always` case.
- Nit (gate ordering vs to_geotiff): fixed. The level range gate in `_write_geotiff_gpu` now runs above the experimental-codec opt-in warning, so a rejected call raises without warning first.
- Nit (test warning strictness): fixed. The no-warning assertion records warnings and checks for nvCOMP messages instead of escalating everything via `simplefilter("error")`.

GPU writer and codec round-trip suites pass locally (237 tests), flake8 clean. No new findings.

PR #3196, opened 2026-06-10
zonal.crosstab: tighten zone_ids/cat_ids type hints and fix stale nodata docstring

### LINKED ISSUE #3188: zonal.crosstab: tighten zone_ids/cat_ids type hints and fix stale nodata docstring
Two small API-consistency issues in `crosstab()`, found by the api-consistency sweep on the zonal module. Both are non-breaking (type hints and docstring text only).

**1. Type-hint drift on `zone_ids` / `cat_ids`.** Both default to `None` but are annotated `List[Union[int, float]] = None`. The sibling functions `stats()` and `crop()` wrap the same optional-list parameter in `Optional[...]`. A type checker reads the current crosstab hints as "must pass a list" and flags the `None` default. Wrapping both in `Optional[...]` matches the siblings.

**2. Stale docstring on `nodata_values`.** The parameter is named `nodata_values`, but its docstring body says "Cells with `nodata` do not belong to any zone". The `nodata` token is left over from a copy-paste of `apply()`, which has a parameter actually named `nodata`. The prose should reference `nodata_values`.

No behavior change, no signature change beyond the annotation, no deprecation needed.

Found by /sweep-api-consistency on the zonal module.

brendancol:
Closes #3188

Two non-breaking API-consistency fixes in `crosstab()`:

- Wrap the `zone_ids` and `cat_ids` annotations in `Optional[...]`. Both default to `None`, and the sibling functions `stats()` and `crop()` already annotate the same parameter as `Optional`. A type checker no longer flags the `None` default.
- Fix the `nodata_values` docstring, which referenced `nodata` (a copy-paste leftover from `apply()`).

Backend coverage: no backend code touched. The change is a type annotation plus docstring text, so numpy / cupy / dask+numpy / dask+cupy behavior is unchanged.

Test plan:
- [x] `pytest xrspatial/tests/test_zonal.py -k crosstab` (17 passed)
- [x] Verified the runtime signature now reports `Optional[List[Union[int, float]]]` for both params
- [x] Verified the docstring references `nodata_values`

Found by /sweep-api-consistency on the zonal module.


brendancol (commented):
## PR Review: zonal.crosstab type hints + docstring fix

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The diff is exactly two lines plus context: `zone_ids` and `cat_ids` gain `Optional[...]`, and the `nodata_values` docstring stops saying `nodata`. Nothing else is touched.
- `Optional`, `List`, and `Union` are already imported at the top of the module, so the annotation change needs no new import.
- The new hints match the sibling functions `stats()` and `crop()`, which is the consistency the change is after.
- No runtime behavior changes, so the existing crosstab tests (17 passed) are the right gate. A dedicated test would only assert a type annotation equals itself.

### Checklist
- Algorithm matches reference/paper: n/a (no logic change)
- Backends produce consistent results: n/a (no backend code touched)
- NaN handling correct: n/a
- Edge cases covered: existing crosstab suite passes
- Dask chunk boundaries: n/a
- Benchmark: not needed
- README feature matrix: not needed (no new function)
- Docstrings present and accurate: yes, this PR fixes a docstring inaccuracy

PR #3199, opened 2026-06-10
Compute dask source once in cumulative_viewshed (#3185)

### LINKED ISSUE #3185: cumulative_viewshed recomputes the dask source raster once per observer
## Description

`cumulative_viewshed` in `xrspatial/visibility.py` loops over `observers` and calls `viewshed(raster, ...)` once per observer, accumulating a count. A code comment ("Detect dask backend to keep accumulation lazy") implies the dask path stays lazy. It doesn't.

For a dask-backed raster with no `max_distance`, each `viewshed()` call goes through `_viewshed_dask` Tier B, which calls `raster.data.compute()` to run the exact CPU sweep. Since `cumulative_viewshed` hands the same dask raster to every call, the full source raster gets materialized N times, once per observer.

## Reproduction

A probe that wraps the source in `da.map_blocks` and counts block evaluations shows one source compute per observer:

```
4 observers => 4 eager source computes during cumulative_viewshed
```

The lazy accumulation buys nothing here: the expensive sweep has already run eagerly N times before the accumulation graph is even built.

## Impact

The per-observer sweep is unavoidable; each observer needs its own visibility grid. What's redundant is recomputing the same source dask graph N times instead of once. When the source is backed by an expensive graph (chained ops, file reads), that multiplies the source cost by the observer count.

Peak memory stays around one grid, so this is not an OOM problem. The output graph grows by roughly 64 tasks per observer, so a very large observer count produces a long graph.

## Proposed fix

When the raster is dask-backed and neither the function-level `max_distance` nor any per-observer `max_distance` is set (every observer takes the full-grid Tier B path), compute `raster.data` once into a numpy-backed DataArray before the loop and pass that to `viewshed()`. When any `max_distance` is present, keep the current per-observer behavior so dask windowing still loads only each observer's window.

Results stay identical: Tier B already computes to numpy and runs `_viewshed_cpu`, and the existing `test_dask_matches_numpy` already asserts dask/numpy parity.

Affected backends: dask+numpy and dask+cupy.

brendancol:
Closes #3185

## What changed

- `cumulative_viewshed` called `viewshed()` once per observer. On the dask path with no `max_distance`, every call ran `raster.data.compute()` on the same source, so the source was materialised once per observer.
- When every observer takes the full-grid path (no function-level or per-observer `max_distance`), the source is now computed once up front and the observers run against the in-memory raster. The accumulated count is re-wrapped as dask so the output backend still matches the dask input.
- Observers that set a `max_distance` keep the dask windowing path, which loads only each observer's window.

## Backend coverage

- numpy: unchanged.
- dask+numpy and dask+cupy: source computed once instead of N times on the no-`max_distance` path; per-observer `max_distance` behaviour unchanged.
- cupy: unchanged.

## Test plan

- [x] `test_dask_source_computed_once` asserts a single source compute for four observers and a dask-backed output.
- [x] `test_dask_per_observer_max_distance_stays_lazy` asserts the windowed path stays dask and matches numpy.
- [x] Existing `test_dask_matches_numpy` parity check still passes.
- [x] Full `test_visibility.py` suite: 27 passed.


brendancol:
## PR Review: Compute dask source once in cumulative_viewshed

(Posted as a comment via REST; `gh pr review`/`gh pr comment` GraphQL path returned HTTP 401.)

### Blockers
None.

### Suggestions
- The `da.from_array(count, chunks=_input_dask_chunks)` wrap at the end builds a dask array whose data is already a concrete numpy buffer. Intentional and correct (it restores the dask output type without re-running the sweep), but the comment only says "Restore the dask-backed output type the dask input would have given." A half-line noting the count is already concrete, so nothing lazy is recovered, would save the next reader a double-take. Not blocking.

### Nits
- The new guard `not any('max_distance' in obs for obs in observers)` walks the observer list once before the main loop. Fine for normal observer counts.

### What looks good
- Surgical change matching the issue: compute the dask source once when every observer takes the full-grid path, otherwise keep per-observer dask windowing.
- Output backend preserved: a dask input still returns a dask-backed result, output chunks match the input (verified).
- Caller's raster left alone; `raster.copy()` does not compute or mutate the input (verified).
- Correctness holds: Tier B already computed to numpy and ran `_viewshed_cpu`; existing `test_dask_matches_numpy` still passes.
- Tests: `test_dask_source_computed_once` asserts one source compute for four observers and a dask output; `test_dask_per_observer_max_distance_stays_lazy` asserts the windowed path stays dask with numpy parity.
- Verified locally: source computes dropped from N to 1; computing the result adds zero further source computes.

### Checklist
- Algorithm matches reference: N/A (scheduling change only)
- Backends consistent: yes
- NaN handling: unchanged
- Edge cases tested: yes (no max_distance, per-observer max_distance, parity)
- Dask chunk boundaries: yes (output chunks match input)
- No premature materialization or unnecessary copies: improved (removes N-1 source materializations)
- Benchmark: none; not needed for a scheduling fix
- README feature matrix: N/A
- Docstrings: accurate (no signature change)

PR #3201, opened 2026-06-10
Use largest chunk for clip_polygon rasterize mask on dask (#3191)

### LINKED ISSUE #3191: clip_polygon crop=True builds an over-fragmented mask chunking on dask backends
## Description

When `clip_polygon` runs with `crop=True` on a dask-backed raster, the dask task graph ends up much bigger than the output size needs. The culprit is the chunk size picked for the internal rasterize mask.

`_crop_to_bbox` slices the dask raster down to the geometry bounding box. Slicing a dask array leaves irregular chunk sizes at the cut edges. For a 2560x2560 raster chunked at (256, 256), clipping to a box that starts mid-chunk gives x-chunks like `(12, 256, 256, 256, 256, 256, 208)`.

`clip_polygon` then takes the rasterize mask chunk size from the first chunk of each axis:

```python
rc, cc = raster.data.chunks[-2], raster.data.chunks[-1]
kw.setdefault('chunks', (rc[0], cc[0]))
```

`rc[0]` / `cc[0]` is the leading edge chunk, which after slicing is often a tiny partial chunk (12 px). `rasterize` builds a uniform mask at that size, so a 1500-px-wide output gets 125 chunks of 12 px each. `xarray.where` then has to align the irregular raster chunks against the tiny uniform mask chunks, and the task count blows up.

## Evidence

Graph construction only, no `.compute()`. 2560x2560 raster, chunks=(256, 256), clip to `box(500, 500, 2000, 2000)`, crop=True:

- output shape: 1500x1500
- mask chunks: (8, 125)
- task count: 13169

Using the largest chunk per axis `(max(rc), max(cc))` instead of the first:

- mask chunks: (6, 6)
- task count: 1045

About a 12.6x smaller graph, same output values.

## Impact

- Backends affected: dask+numpy and dask+cupy (both go through the same chunk selection).
- Bottleneck: graph-bound. This is scheduler and graph-build overhead, not peak memory. Peak memory still scales with chunk size, so it is not an OOM risk.
- crop=False is unaffected (no slicing, chunks stay uniform). numpy and cupy non-dask paths are unaffected.

## Fix

Pick a representative interior chunk size instead of the leading partial chunk:

```python
kw.setdefault('chunks', (max(rc), max(cc)))
```

That keeps the mask grid coarse and roughly aligned with the raster's interior chunk size.

brendancol:
Closes #3191

`clip_polygon` with `crop=True` on a dask raster fragments the rasterize mask into tiny chunks and blows up the task graph. `_crop_to_bbox` slices the dask array, leaving a small partial chunk at the leading edge, and the code picked that first chunk (`chunks[-1][0]`) as the mask chunk size. A 1500-px-wide output ended up with ~125 narrow mask chunks, and `xarray.where` had to align them against the irregular raster chunks.

## Changes

- Pick the largest chunk per axis (`max(rc), max(cc)`) for the rasterize mask instead of the leading partial chunk, so the mask grid stays coarse and tracks the raster's interior chunk size.
- For `box(500, 500, 2000, 2000)` on a 2560x2560 raster (256 chunks), this drops the task graph from 13169 to 1045 tasks (graph construction only, no `.compute()`), with identical output values.
- Add regression tests: one asserts the crop-path graph and `numblocks` stay bounded without computing, the other checks `crop=True` values match `crop=False`.

## Backend coverage

- dask+numpy: fixed, tested.
- dask+cupy: fixed (same code path), tested on a CUDA host.
- numpy / cupy non-dask: unaffected (no slicing, no chunk selection).
- `crop=False`: unaffected (chunks stay uniform).

## Test plan

- [x] `pytest xrspatial/tests/test_polygon_clip.py` (25 passed, including CuPy and dask+cupy)
- [x] New graph-size regression test passes and uses graph construction only
- [x] crop vs no-crop value parity test passes


brendancol (commented):
## PR Review: Use largest chunk for clip_polygon rasterize mask on dask (#3191)

None.

None.

- `test_crop_matches_nocrop_values` runs on an all-zeros raster, so inside the polygon every value is 0.0 and outside is NaN. That checks mask placement and NaN positions but not that real data values survive the coarser chunking. `TestClipPolygonDask.test_crop_matches_numpy` already covers value preservation with arange data, so this is fine as written; swapping in arange data here too would just make the one test self-sufficient.

### What looks good
- One-line change with a comment that names the root cause: slicing leaves a tiny leading edge chunk, and the old code picked that as the mask chunk size.
- `max(rc)` / `max(cc)` is safe. Chunk tuples off a dask array are always non-empty with positive ints, and for a single-chunk-per-axis raster `max(rc) == rc[0]`, so that case is unchanged.
- The change only touches the dask branch. numpy/cupy non-dask paths and crop=False are untouched, which matches the PR description.
- The graph-size regression test reads `__dask_graph__()` and never calls `.compute()`, which is the right way to assert graph shape.
- Both dask+numpy and dask+cupy hit the changed line and both are tested; the suite passes on a CUDA host.

### Checklist
- Algorithm matches reference/paper: N/A, chunking change only
- All implemented backends produce consistent results: yes, crop vs no-crop parity tested
- NaN handling correct: yes, mask placement unchanged
- Edge cases covered: mid-chunk crop covered; single-chunk raster is a no-op by construction
- Dask chunk boundaries handled correctly: yes, that is the fix
- No premature materialization or unnecessary copies: confirmed, no .compute()
- Benchmark exists or not needed: no benchmark, not needed for a chunking fix
- README feature matrix updated: N/A, no API change
- Docstrings present and accurate: unchanged, still accurate

PR #3204, opened 2026-06-10
Align clip_polygon dask+cupy mask chunks before map_blocks (#3186)

### LINKED ISSUE #3186: clip_polygon dask+cupy path fails on rasters with non-uniform chunks
**Describe the bug**

`clip_polygon` raises on the dask+cupy backend when the input raster has non-uniform chunks. In `xrspatial/polygon_clip.py`, the dask+cupy branch builds the rasterized mask using chunk sizes taken only from the first chunk along each axis (`rc[0]`, `cc[0]`), then passes both the raster and the mask condition into `da.map_blocks(_apply_mask, raster.data, cond)`. `map_blocks` requires its array arguments to share a chunk layout. When the raster's chunks are irregular, the uniform mask layout differs, so `map_blocks` pairs up mismatched blocks and raises `IndexError` ("boolean index did not match indexed array") or `ValueError` ("Shapes do not align"). If block shapes happen to coincide, it could instead write the mask to the wrong cells without raising.

The dask+numpy path is unaffected because it goes through `xarray.DataArray.where`, which rechunks the mask to align with the raster.

**Expected behavior**

`clip_polygon` on a dask+cupy raster with non-uniform chunks should match the numpy backend, the same as the dask+numpy path does.

**Reproduction**

A raster of shape (8, 6) rechunked to `((3, 5), (6,))` on the dask+cupy backend, clipped with a box polygon, raises `ValueError: Shapes do not align`. The same raster on dask+numpy returns the correct masked array.

**Fix**

Rechunk the mask condition to `raster.data.chunks` before calling `da.map_blocks`. Also migrate the deprecated `rasterize(use_cuda=True)` call in that branch to `gpu=True`.

**Additional context**

Found during the deep-sweep accuracy audit of the polygon_clip module. The existing dask+cupy test only uses regular chunks, so the irregular-chunk case is uncovered.

brendancol:
Closes #3186

`clip_polygon` on the dask+cupy backend raised (or risked stamping the mask onto the wrong cells) when the input raster had non-uniform chunks.

- The dask+cupy branch rasterizes the polygon mask with a uniform chunk size taken from the raster's first chunk, then passes raster and mask into `da.map_blocks`, which pairs blocks positionally. A raster with irregular chunks gave the mask a different block layout, so `map_blocks` raised `IndexError`/`ValueError` ("Shapes do not align"). The fix rechunks the condition to `raster.data.chunks[-2:]` before `map_blocks`.
- The dask+numpy path was already correct because it goes through `xarray.DataArray.where`, which rechunks the mask to align.

Backend coverage: numpy and cupy unchanged; dask+numpy already correct (now covered by a test); dask+cupy fixed.

Test plan:
- [x] `pytest xrspatial/tests/test_polygon_clip.py` (25 passed, CUDA host)
- [x] New `test_nonuniform_chunks_matches_numpy` for dask+numpy and dask+cupy, exercising `((3,5),(6,))` and `((3,2,3),(2,1,3))` layouts against the numpy reference
- [x] Confirmed the repro raised before the fix and matches numpy after

Found during the deep-sweep accuracy audit of `polygon_clip`.


brendancol (commented):
## PR Review: Align clip_polygon dask+cupy mask chunks before map_blocks (#3186)

None.

None.

- xrspatial/polygon_clip.py:247 -- `cond` is always 2D (the rasterize mask) and `raster.data.chunks[-2:]` is the last-two-axis chunk tuple, so the rechunk aligns correctly only for 2D rasters. A raster with a leading band dimension would already fail one line down at `da.map_blocks(_apply_mask, raster.data, cond)` on an ndim mismatch, so this is a pre-existing limitation the PR neither fixes nor worsens. A one-line comment or an explicit 2D assertion would make the 2D-only assumption visible, if you want it.

### What looks good
- Root cause is right: `da.map_blocks` pairs blocks positionally, so a uniform mask layout silently mismatches an irregular raster layout. Rechunking the condition to the raster's chunks is the minimal fix.
- The comment names both failure modes (raise and silent mis-stamp), which matches the reproduction.
- Tests cover two irregular layouts (a 2-chunk and a 3-chunk split) on both dask+numpy and dask+cupy against the numpy reference with equal_nan. They fail before the fix and pass after.
- numpy, cupy, and dask+numpy code paths are untouched. The dask+numpy test is a guard for a path that was already correct via xarray.where.

### Checklist
- [x] Algorithm matches reference/paper (n/a; masking, no numerical algorithm)
- [x] NaN handling is correct (nodata default np.nan; equal_nan in tests)
- [x] Edge cases covered by tests (irregular 2- and 3-chunk layouts)
- [x] Dask chunk boundaries handled correctly (the fix itself)
- [x] No premature materialization or unnecessary copies (mask stays lazy)
- [x] Benchmark exists or is not needed (not needed; correctness fix)
- [x] README feature matrix updated (n/a; no new function or backend)
- [x] Docstrings present and accurate (unchanged public API)

PR #3205, opened 2026-06-10
Fix cupy backend crash in cumulative_viewshed/visibility_frequency

### LINKED ISSUE #3193: cumulative_viewshed and visibility_frequency crash on the cupy backend
**Describe the bug**

`cumulative_viewshed` and `visibility_frequency` in `xrspatial/visibility.py` crash on the cupy (GPU) backend.

`cumulative_viewshed` initializes its accumulator `count` as a numpy array for any non-dask raster (visibility.py:271-272), cupy-backed rasters included. The loop then does `count + (vs_data != INVISIBLE).astype(np.int32)` at visibility.py:287. On the GPU backend `vs_data` comes back from `viewshed()` as a cupy array, so adding a numpy array to it raises `TypeError: Unsupported type <class 'numpy.ndarray'>`. `visibility_frequency` calls `cumulative_viewshed`, so it breaks the same way.

A metadata-propagation sweep turned this up. Because the cupy backend raises instead of returning, it produces no attrs, coords, or dims at all, while numpy and dask+numpy return a DataArray carrying the input's coords, dims, and attrs (crs included). That's a backend-consistency gap in the returned metadata on top of the crash.

**Reproduction**

```python
from xrspatial.visibility import cumulative_viewshed

data = np.random.RandomState(1).rand(20, 20).astype(float) * 100
r = xr.DataArray(data, dims=['y', 'x'],
                 coords={'y': np.arange(20.0), 'x': np.arange(20.0)})
r.data = cp.asarray(r.data)
cumulative_viewshed(r, [{'x': 5.0, 'y': 5.0, 'observer_elev': 50}])
# TypeError: Unsupported type <class 'numpy.ndarray'>
```

**Expected behavior**

A cupy-backed raster returns a cupy-backed int32 DataArray with the same coords, dims, and attrs as the numpy backend, matching the cross-backend contract `viewshed` already follows.

**Proposed fix**

Initialize `count` as a cupy array when the raster is cupy-backed (mirroring the existing dask branch) so the accumulation stays on-device and the final DataArray wraps a cupy array. These functions have no cupy test coverage today; add cross-backend tests.

brendancol:
Closes #3193

`cumulative_viewshed` initialized its accumulator `count` as a numpy array for every non-dask raster, so accumulating cupy `viewshed` results raised `TypeError: Unsupported type <class 'numpy.ndarray'>` on the GPU backend. `visibility_frequency` wraps `cumulative_viewshed`, so it failed the same way.

- Initialize `count` as a cupy array when the raster is cupy-backed, mirroring the existing dask branch. The accumulation stays on-device and the result is a cupy-backed DataArray.
- On cupy the returned DataArray now carries the input coords, dims, and attrs (crs included), matching numpy and dask+numpy.
- Add cupy cross-backend tests for both functions.

Backend coverage: numpy and dask+numpy unchanged; cupy now works instead of crashing. dask+cupy isn't exercised by these functions (the dask branch runs before the cupy branch).

Test plan:
- [x] `pytest xrspatial/tests/test_visibility.py` (28 passed, including 3 new cupy tests on an RTX host)
- [x] Confirmed the cupy result is a real `cupy.ndarray` with int32/float64 dtype and preserved coords/dims/attrs
- [x] Confirmed cupy `cumulative_viewshed` matches the cupy `viewshed` binary mask for a single observer


brendancol (commented):
## PR Review: Fix cupy backend crash in cumulative_viewshed/visibility_frequency

### Blockers
None.

### Suggestions
None blocking. One thing worth a sentence in the docstring or a comment: `cumulative_viewshed` checks the dask branch before the cupy branch, so a dask-of-cupy raster goes through the dask path and never hits the new `cp.zeros` branch. That matches how the function already worked and the PR body calls it out, but a one-line comment at the branch would save the next reader the trace.

### Nits
The new `_is_cupy` guard recomputes `has_cuda_and_cupy()`; that is cheap and matches the style of the surrounding code, so leaving it is fine.

### What looks good
- The cupy branch mirrors the existing dask branch exactly: detect backend, allocate the accumulator on the right device, let the final DataArray wrap whatever `count` is. Minimal and consistent.
- Tests assert the result is a real `cupy.ndarray`, check dtype, and confirm coords/dims/attrs (crs included) survive, which is the actual contract that was broken.
- The single-observer test compares against the cupy `viewshed` binary mask rather than the numpy result, which is correct: the RTX viewshed and the CPU sweep disagree on visible-cell counts, so a numpy-vs-cupy value assertion would be wrong here. Good call.

### Checklist
- [x] Fix matches the backend-dispatch pattern used for dask
- [x] cupy result is consistent with cupy viewshed
- [x] coords/dims/attrs preserved on cupy
- [x] Edge cases: empty observers still raises (unchanged path)
- [x] No new public API, so README/benchmark unaffected
- [x] Tests gated on cuda+rtx availability

PR #3206, opened 2026-06-10
Add clip_polygon backend/edge-case/metadata test coverage (#3197)

### LINKED ISSUE #3197: clip_polygon: backend, edge-case, and metadata test gaps
## Description

A test-coverage sweep of `xrspatial.polygon_clip.clip_polygon` turned up several untested code paths. The existing test file covers numpy well and has one parity test for each of the dask/cupy backends, but the paths below aren't hit by any test.

**Backend coverage (cupy / dask+cupy)**
- The cupy and dask+cupy backends only test `crop=False` with the inner polygon. Custom `nodata`, `all_touched=True`, and NaN-input preservation never run on a GPU backend, even though those branches go through separate code (`_apply_mask` map_blocks for dask+cupy, raw-array masking for cupy).

**NaN / Inf / nodata edge cases**
- Inf and -Inf inputs, all-NaN input, and integer-dtype input with a sentinel `nodata` value are all untested.
- NaN-in-input preservation is tested only on numpy, not across backends.

**Geometric edge cases**
- Nx1 and 1xN strip rasters (single-column / single-row grids) aren't tested.

**Metadata preservation**
- Nothing asserts that output coords equal input coords for `crop=False`, or that cropped output coords are a contiguous subset of the input coords for `crop=True`.

I probed all of these locally on a CUDA host and they behave correctly today, so this is a coverage gap, not a source bug. The fix is test-only.

Found by the deep-sweep test-coverage sweep on 2026-06-10.

brendancol:
Closes #3197.

Test-only changes from the deep-sweep test-coverage sweep. Adds 13 tests to `xrspatial/tests/test_polygon_clip.py`. No source files changed.

What's covered:
- cupy and dask+cupy each now assert custom `nodata`, `all_touched=True`, and NaN-input preservation against the numpy result. Before this, the GPU backends only ran `crop=False` with the inner polygon.
- Edge inputs: +Inf / -Inf survive the clip, all-NaN input clips to all-NaN, and an int32 raster with a sentinel `nodata=-1`.
- Strip rasters: Nx1 single-column and 1xN single-row grids.
- Metadata: output coords equal input coords for `crop=False`, and the cropped coords are a contiguous subset of the input coords for `crop=True`.

Test plan:
- Ran the full file on a CUDA host: 36 passed, 0 skipped.
- The 6 new GPU tests executed (not skipped) and passed.

Notes for reviewers (out of scope here, not bundled into this PR):
- The docstring says `clip_polygon` accepts named y and x dims, but the underlying `rasterize()` requires literal y/x dims, so lat/lon-dim rasters raise. Dim-name preservation is unsupported by contract today, so it is not a test gap.
- `polygon_clip.py` still calls `rasterize(use_cuda=True)` on the dask+cupy path. `use_cuda` was renamed to `gpu` in #3089, so this hits the deprecation alias. Harmless for now, worth an api-consistency follow-up.


brendancol (commented):
## PR Review: clip_polygon backend/edge-case/metadata test coverage

Test-only PR, so the review is about whether the new tests exercise the right paths and assert the right things.

### Blockers
None.

### Suggestions
- `test_integer_dtype_sentinel_nodata` (test_polygon_clip.py) asserts the sentinel value and that kept cells came from the input, but not the output dtype. The docstring promises "same dtype as raster," and the int32 path does preserve int32 today (numpy `where` with an int fill). Adding `assert result.dtype == np.int32` would pin that contract so a future change that silently upcasts to float gets caught.

### Nits
None.

### What looks good
- The 6 GPU tests actually ran on the CUDA host (not skipped) and assert parity against numpy via `.data.get()` / `.data.compute().get()`, which matches the project convention.
- Edge inputs are real edge inputs: NaN sits at an interior cell, Inf and -Inf are checked with `isposinf` / `isneginf`, all-NaN is end-to-end.
- The crop-coords test checks a contiguous run, not just set membership, so a reordering or stride bug would fail.
- No source changes, and the two out-of-contract observations (lat/lon dims, lingering `use_cuda=`) are called out in the PR body rather than silently fixed.

### Checklist
- [x] All implemented backends exercised (numpy / cupy / dask+numpy / dask+cupy)
- [x] NaN / Inf handling tested
- [x] Edge cases covered (1xN, Nx1, all-NaN, int sentinel)
- [x] Dask chunk boundaries exercised (ragged (4,3) chunks)
- [x] No premature materialization in test helpers
- [ ] Output dtype assertion missing on the int path (see Suggestions)

brendancol:
## Follow-up review

The suggestion from the prior review pass is addressed in 192f78ba: `test_integer_dtype_sentinel_nodata` now asserts `result.dtype == np.int32`, pinning the docstring's same-dtype contract for the integer + sentinel path. Verified passing locally.

PR #3209, opened 2026-06-10
Fix cumulative_viewshed TypeError on cupy rasters (#3194)

### LINKED ISSUE #3194: cumulative_viewshed and visibility_frequency raise TypeError on cupy-backed rasters
## Summary

`cumulative_viewshed` (and `visibility_frequency`, which calls it) raise `TypeError: Unsupported type <class 'numpy.ndarray'>` on a cupy-backed raster. The numpy and dask+numpy backends work. The cupy path has no test coverage, so this slipped through.

## Root cause

In `xrspatial/visibility.py`, the non-dask branch initializes the accumulator as numpy:

```python
count = np.zeros(raster.shape, dtype=np.int32)
```

For a cupy-backed raster, `viewshed()` returns a cupy-backed DataArray, so `vs_data` is a `cupy.ndarray`. The accumulation line then mixes numpy and cupy:

```python
count = count + (vs_data != INVISIBLE).astype(np.int32)
```

cupy rejects the mixed-array operation and raises.

## Reproduce

```python
from xrspatial.visibility import cumulative_viewshed

data = (np.random.RandomState(3).rand(15, 15) * 100).astype(float)
r = xr.DataArray(cupy.asarray(data), dims=['y', 'x'],
                 coords={'y': np.arange(15.0), 'x': np.arange(15.0)})
cumulative_viewshed(r, [{'x': 3.0, 'y': 3.0, 'observer_elev': 50}])
# TypeError: Unsupported type <class 'numpy.ndarray'>
```

## Suggested fix

Pull each per-observer viewshed result to numpy before accumulating into the numpy `count`. The module already does this elsewhere (`_extract_transect` calls `.get()` on cupy data). When `vs_data` is a cupy array, call `.get()` before the comparison.

## Expected

cupy-backed input returns the same int32 counts as the numpy backend, aside from the RTX-vs-GRASS visibility differences already documented on `viewshed`.

Found by the accuracy sweep (Cat 5: backend inconsistency).

brendancol:
Fixes #3194

## Problem

`cumulative_viewshed` and `visibility_frequency` raised
`TypeError: Unsupported type <class 'numpy.ndarray'>` on a cupy-backed raster.
`viewshed()` returns a cupy-backed DataArray for cupy input, but the accumulator
was a numpy array (`count = np.zeros(...)`), and cupy refuses a numpy+cupy add.
The cupy path had no test coverage, so this went unnoticed.

## Fix

Pull each per-observer viewshed result to numpy with `.get()` before
accumulating into the numpy `count`. This matches how `_extract_transect`
already handles cupy data in the same module.

## Test

Added a cupy regression test (gated on `rtxpy`) confirming both functions run
and return valid int32 counts. It does not assert cell-for-cell parity with
numpy: the GPU viewshed uses RTX ray tracing and the numpy backend uses the
GRASS sweep, a divergence already documented on `viewshed`.

Verified on an RTX A6000: both functions run, return int32 / float64 with
valid ranges, 26 visibility tests pass.

Found by the accuracy sweep (Cat 5: backend inconsistency).


brendancol:
Closing in favor of #3205, which fixes the same cumulative_viewshed cupy crash but keeps the accumulator on-device, so cupy input yields a cupy-backed result (consistent with the cross-backend parity convention). This PR's .get() approach returns numpy output for cupy input, and the two test suites assert contradictory result types, so they can't both land. #3205 also covers visibility_frequency and metadata preservation.

PR #3210, opened 2026-06-10
Fix clip_polygon GPU dtype promotion for NaN nodata on integer rasters

### LINKED ISSUE #3187: clip_polygon: integer raster + default nodata=np.nan crashes on cupy/dask+cupy but works on CPU
**Describe the bug**

`clip_polygon` behaves differently across backends when the input raster has an integer dtype and `nodata` is left at its default of `np.nan`.

- numpy and dask+numpy: `raster.where(cond, other=np.nan)` promotes the output to float64 and masks pixels with NaN. Works.
- cupy and dask+cupy: the code does `out = raster.data.copy()` (keeping the integer dtype) then `out[~cond] = nodata`, which raises `ValueError: cannot convert float NaN to integer`.

So `clip_polygon(int_raster, poly)` with default arguments succeeds on CPU but crashes on GPU.

**Reproduce**

```python
import numpy as np
from xrspatial.polygon_clip import clip_polygon

poly = Polygon([(0.6, 0.6), (0.6, 2.9), (1.9, 2.9), (1.9, 0.6)])
data = np.arange(48, dtype=np.int32).reshape(8, 6)

r = create_test_raster(data, backend='cupy')
clip_polygon(r, poly)  # ValueError: cannot convert float NaN to integer
```

dask+cupy raises the same error on compute.

**Expected behavior**

The cupy and dask+cupy paths should match the numpy path: when the requested nodata value cannot be represented in the input integer dtype (e.g. NaN), promote the output to a float dtype that can hold it instead of crashing.

**Location**

`xrspatial/polygon_clip.py`, the cupy branch around line 266 (`out = raster.data.copy(); out[~cond_cp] = nodata`) and the dask+cupy `_apply_mask` around line 247.

**Environment**

xarray-spatial main, cupy backend, CUDA available.

Found by the metadata propagation sweep (Cat 4 dtype/nodata, Cat 5 backend-inconsistent).

brendancol:
Closes #3187

`clip_polygon` on an integer raster with the default `nodata=np.nan` worked on numpy and dask+numpy (where `.where()` promotes the result to float64) but crashed on cupy and dask+cupy with `ValueError: cannot convert float NaN to integer`. The GPU paths copied the input array and assigned `nodata` in place, keeping the integer dtype.

- Promote the GPU output dtype via `np.result_type(raster.dtype, nodata)`, which reproduces `DataArray.where()`'s promotion: the dtype stays the same when `nodata` fits the input dtype and widens otherwise.
- Applied in both the pure-cupy branch and the dask+cupy `_apply_mask` block.
- Updated the docstring's Returns note about dtype.

Backend coverage: numpy, dask+numpy, cupy, dask+cupy all verified on a CUDA host. int32+NaN promotes to float64 on every backend; int32 with an in-range integer nodata stays int32 on every backend; float inputs are unchanged.

Test plan:
- [x] `pytest xrspatial/tests/test_polygon_clip.py` (29 passed, includes cupy + dask+cupy)
- [x] New cupy regression test fails on the pre-fix code and passes with the fix
- [x] dtype and values match the numpy reference across all four backends

Found by the metadata propagation sweep (Cat 4 dtype/nodata, Cat 5 backend-inconsistent).


brendancol (commented):
## PR Review: Fix clip_polygon GPU dtype promotion for NaN nodata on integer rasters

None.

None.

- `xrspatial/polygon_clip.py`: the numpy and dask+numpy branches still get promotion from `.where()` while the GPU branches now route through `_masked_output_dtype`. The split is intentional and correct, but a one-line comment on the numpy branch pointing at the helper would help the next reader connect the two. Optional.

### What looks good
- `_masked_output_dtype` reproduces `DataArray.where()`'s promotion exactly. I checked int8/16/32, uint8, and float32/64 against NaN, in-range integers, out-of-range integers, and a float64 overflow value. Every case matches `.where()`.
- Both GPU branches use `astype(out_dtype, copy=True)`, so the input array is not mutated and the result is a fresh array. Confirmed on a CUDA host.
- `da.map_blocks` declares the promoted `dtype`, so the lazy metadata matches the computed result.
- Tests cover all four backends plus the integer-nodata-stays-int case. The new cupy test fails on the pre-fix code and passes with the fix, so it is a real regression guard.

### Checklist
- [x] Algorithm matches reference (`DataArray.where` promotion semantics)
- [x] All four backends produce consistent dtype and values
- [x] NaN handling correct
- [x] Edge cases covered (int+NaN, int+integer nodata, float inputs)
- [x] Dask chunk metadata (dtype) handled correctly
- [x] No premature materialization or input mutation
- [ ] Benchmark not needed (bug fix, no perf-sensitive change)
- [ ] README feature matrix unchanged (no new function, backend support unchanged)
- [x] Docstring updated to reflect dtype promotion

brendancol (commented):
## Follow-up review

The one optional nit from the previous pass is addressed: the numpy branch now carries a comment cross-referencing `_masked_output_dtype`, so the connection between the `.where()` promotion on CPU and the explicit promotion on the GPU branches is visible in place.

No Blockers or Suggestions were raised. All 29 tests in `test_polygon_clip.py` still pass, including the cupy and dask+cupy backends on a CUDA host. Nothing else outstanding.

brendancol:
Closing as a duplicate of #3208, which just merged. Both PRs fix the same int-raster + NaN nodata crash on the GPU backends via np.result_type, and #3208's tests also cover a float sentinel on an int raster. Issue #3187 is covered by the merged fix.

PR #3211, opened 2026-06-10
Validate raster inputs in hypsometric_integral()

### LINKED ISSUE #3200: hypsometric_integral() skips _validate_raster(), silently accepts complex dtype
## Description

`hypsometric_integral()` in `xrspatial/zonal.py` skips input validation. Every other public function in the module (`stats`, `crosstab`, `apply`, `regions`, `trim`, `crop`) calls `_validate_raster()` on its inputs, but `hypsometric_integral()` only calls `_maybe_rasterize_zones()` followed by `validate_arrays()`.

`validate_arrays()` checks that the two arrays have matching shape and backend. It does not check dtype, and it does not check that the inputs are actually `DataArray`s. So a complex-dtype `values` array slips through, and numpy silently drops the imaginary part.

## Reproduce

```python
from xrspatial.zonal import hypsometric_integral, stats

z = xr.DataArray(np.array([[1, 1], [2, 2]], dtype=np.int32), dims=['y', 'x'])
v = xr.DataArray(np.array([[1+2j, 3+1j], [5j, 2+0j]], dtype=np.complex128), dims=['y', 'x'])

# stats() rejects the complex input with a clear ValueError:
stats(z, v)
# ValueError: stats(): `values` must have a real numeric dtype (integer or float), got complex128

# hypsometric_integral() drops the imaginary part and returns finite output:
hypsometric_integral(z, v, nodata=None)
# Only a ComplexWarning, returns float64 output [0.5 0.5 0.5 0.5]
```

## Expected

`hypsometric_integral()` should reject complex and non-numeric inputs with the same `ValueError` the other zonal functions raise, instead of quietly returning wrong numbers.

## Fix

Add `_validate_raster()` calls for both `zones` and `values` (2D, real numeric) at the top of `hypsometric_integral()`, before `validate_arrays()`, matching the pattern in `crop()`.

Found during a security sweep of the zonal module (dtype-confusion category). A silent wrong-result path is worse than a loud error.

brendancol:
Closes #3200

`hypsometric_integral()` was the only public function in `zonal.py` that did not call `_validate_raster()` on its inputs. It went straight from `_maybe_rasterize_zones()` to `validate_arrays()`, which only checks shape and backend parity, not dtype. A complex-dtype `values` array slipped through and numpy silently discarded the imaginary part, returning finite wrong numbers with just a `ComplexWarning`.

- Add `_validate_raster()` for both `zones` and `values` (2D, real numeric) before `validate_arrays()`, matching the pattern in `crop()`.
- Complex / non-numeric `values` now raise a clear `ValueError` (same as `stats()`); a non-DataArray `values` raises a clear `TypeError` instead of an opaque `AttributeError`.

Found during a security sweep of the zonal module (dtype-confusion category).

Backend coverage: the validation runs before backend dispatch, so it applies to numpy, cupy, dask+numpy, and dask+cupy. No backend-specific code changed.

Test plan:
- [x] `test_hypsometric_integral_rejects_complex_values` (complex -> ValueError)
- [x] `test_hypsometric_integral_rejects_non_dataarray_values` (raw ndarray -> TypeError)
- [x] Full `test_hypsometric_integral.py` suite passes (38 passed), confirming valid float/int inputs across all backends are unaffected.

### DIFF
--- a/xrspatial/tests/test_hypsometric_integral.py
+++ b/xrspatial/tests/test_hypsometric_integral.py
@@ -405,3 +405,41 @@ def test_hi_preserves_backend_2525(backend):
         # confirm chunks are cupy
         sample = result.data.blocks[0, 0].compute()
         assert is_cupy_array(sample)
+
+
+def test_hypsometric_integral_rejects_complex_values():
+    """Regression #3200: complex/non-numeric inputs must raise, not silently
+    drop the imaginary part.
+
+    hypsometric_integral() used to skip _validate_raster(), so a complex
+    `values` array slipped through and numpy discarded its imaginary part
+    (only a ComplexWarning), returning finite wrong numbers. Every other
+    zonal function rejects complex input with a clear ValueError; this one
+    must too.
+    """
+    from xrspatial.zonal import hypsometric_integral
+
+    zones = xr.DataArray(
+        np.array([[1, 1], [2, 2]], dtype=np.int32), dims=['y', 'x']
+    )
+    values = xr.DataArray(
+        np.array([[1 + 2j, 3 + 1j], [5j, 2 + 0j]], dtype=np.complex128),
+        dims=['y', 'x'],
+    )
+    with pytest.raises(ValueError, match='real numeric dtype'):
+        hypsometric_integral(zones, values, nodata=None)
+
+
+def test_hypsometric_integral_rejects_non_dataarray_values():
+    """Regression #3200: a non-DataArray `values` must raise a clear
+    TypeError from _validate_raster, not an opaque AttributeError from
+    validate_arrays.
+    """
+    from xrspatial.zonal import hypsometric_integral
+
+    zones = xr.DataArray(
+        np.array([[1, 1], [2, 2]], dtype=np.int32), dims=['y', 'x']
+    )
+    raw_values = np.array([[1.0, 2.0], [3.0, 4.0]])
+    with pytest.raises(TypeError, match='must be an xarray.DataArray'):
+        hypsometric_integral(zones, raw_values, nodata=None)
--- a/xrspatial/tests/test_hypsometric_integral.py
+++ b/xrspatial/tests/test_hypsometric_integral.py
@@ -407,7 +407,8 @@ def test_hi_preserves_backend_2525(backend):
         assert is_cupy_array(sample)
 
 
-def test_hypsometric_integral_rejects_complex_values():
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Validate raster inputs in hypsometric_integral()

None.

- The two new tests run only on the default numpy path. The validation is a pre-dispatch metadata guard (dtype/ndim/type), so it behaves the same on every backend, and a numpy-only test is defensible. If you want belt-and-suspenders coverage, parametrizing `test_hypsometric_integral_rejects_complex_values` over the `backend` fixture (like the rest of the file) would confirm cupy and dask inputs hit the same guard. Optional.

- The new `ndim=2` guard now rejects a 3D `values` DataArray with a clear ValueError, which is a behavior improvement worth a one-line test, since `hypsometric_integral` is documented as 2D-only. Not required.

### What looks good
- The fix mirrors the existing `crop()` pattern: `_validate_raster` for both inputs, placed after `_maybe_rasterize_zones` (so rasterized vector zones get validated too) and before `validate_arrays`.
- `_validate_raster` only reads `.dtype` and `.ndim`, so there's no premature materialization on dask or cupy inputs.
- Leaving off `integer_only=True` for zones is the right call. The HI backends compare and unique float zone IDs fine, and forcing an integer dtype would break the existing float-zone tests.
- The regression tests assert both the ValueError (complex) and the clearer TypeError (raw ndarray), not just that the call runs.

### Checklist
- [x] Algorithm unchanged (validation-only)
- [x] All backends share the same pre-dispatch guard
- [x] NaN handling unchanged
- [x] Edge cases (complex, non-DataArray) covered by new tests
- [x] No dask chunk-boundary impact
- [x] Benchmark not needed (validation guard)
- [x] No README change needed (no new function, no backend change)
- [x] Docstring unchanged and still accurate

brendancol (commented):
## Follow-up review (after 4f2a1edc)

- Suggestion (cross-backend coverage): `test_hypsometric_integral_rejects_complex_values` now carries `@pytest.mark.parametrize("backend", ['numpy', 'dask+numpy', 'cupy', 'dask+cupy'])` and builds inputs with `create_test_raster`, skipping cupy/dask when the hardware or dask isn't present. On this CUDA host all four variants run and pass, so the guard is confirmed to fire before backend dispatch on real cupy and dask+cupy inputs, not just numpy.
- Nit (3D rejection): added `test_hypsometric_integral_rejects_3d_values`, which checks the new `ndim=2` guard raises a clear `ValueError` for a 3D `values` DataArray.

Full suite: 42 passed. No blockers. Nothing left open.

PR #3212, opened 2026-06-10
Vectorize regions() relabel loop (#3207)

### LINKED ISSUE #3207: regions() relabel loop is O(n_regions x n_cells) and slow on many-region inputs
## Summary

`regions()` relabels connected components with a per-region loop that scans the whole array once per region. That makes it O(n_regions * n_cells), which bites when the input has lots of small regions.

## Where

`xrspatial/zonal.py`, in `_regions_numpy` and `_regions_cupy`. After `scipy.ndimage.label` (or `cupyx.scipy.ndimage.label`) hands back a contiguous per-value labeling, the code does this:

```python
for region_id in range(1, n_features + 1):
    out[labeled == region_id] = uid
    uid += 1
```

Every iteration evaluates `labeled == region_id` over the full array, so a raster with N regions pays for N full-array scans.

## Impact

On a 512x512 raster of random 0/1 noise (about 35,000 connected regions), the relabel loop takes around 2.4s on numpy and around 24s on cupy. The dask and dask+cupy backends delegate to these two, so all four are affected.

## Proposed fix

Swap the per-region scan for one vectorized remap. Build a lookup array of length `n_features + 1` mapping each local label to its global id, then index it: `painted = remap[labeled]`, assign where `labeled > 0`, and bump the global id counter by `n_features`. The cupy backend uses the cupy equivalents.

A prototype gives byte-identical output to the current code (same labeling order) and runs about 180x faster on the 35k-region case (2503ms -> 14ms).

_Found by /sweep-performance (module: zonal)._

brendancol:
Closes #3207.

`regions()` relabeled connected components with a per-region loop that scanned the whole array once per region, so the cost grew as O(n_regions * n_cells). On a 512x512 raster of random 0/1 noise (about 35k regions) that was around 2.4s on numpy.

## Change

- Replace the per-region scan in `_regions_numpy` and `_regions_cupy` with a single fancy-index remap: build a lookup of length `n_features + 1` mapping each local label to its global id, then index `remap[labeled]` and assign where `labeled > 0`.
- Output is byte-identical to the old loop (same labeling order), so the docstring examples and existing tests are unchanged.

## Backends

numpy and cupy carry the fix directly. dask+numpy delegates to `_regions_numpy` and dask+cupy delegates to `_regions_cupy`, so all four benefit. No new host/device transfers: the cupy path stays on the GPU and returns a cupy array.

## Test plan

- [x] Existing `test_regions_*` pass unchanged (17 tests).
- [x] New `test_regions_many_regions_matches_reference_3207` checks the vectorized result against the original per-region semantics on a many-region input, across numpy, dask+numpy, cupy, and dask+cupy (CUDA available on the test host, so the GPU variants actually ran).
- [x] Full `test_zonal.py` suite green (187 passed).

_Found by /sweep-performance (module: zonal)._


brendancol (commented):
## PR Review: Vectorize regions() relabel loop (#3207)

None.

- There is no `regions` benchmark in `benchmarks/benchmarks/zonal.py` (only `zonal_stats` is covered). This PR is a performance change, so a small `time_regions` case on a many-region input would catch a future regression. Not blocking, since the new test already pins cross-backend correctness.

### Nits (optional)
- `xrspatial/zonal.py:2189` and `:2286`: `painted = remap[labeled]` builds a full-size temporary, then `out[sel] = painted[sel]` gathers from it. `out[sel] = remap[labeled[sel]]` gets the same result with one gather and no full-size temp. Marginal, and the current form reads fine.

### What looks good
- The remap keeps the exact global-id order of the old loop: local label k maps to uid+k-1, and index 0 is masked out by `labeled > 0`. So the change is behavior-preserving. I checked byte-identical output against the old per-region loop across neighborhoods 4 and 8, float and int dtypes, NaN inputs, and 1x1 / odd shapes.
- numpy and cupy get the same treatment, and the new test runs all four backends. cupy and dask+cupy actually executed on the CUDA host here rather than skipping.
- The cupy path stays on the GPU. remap, painted, sel, and out are all cupy, so there is no new host round-trip.
- The `n_features == 0` guard is a fair clarification, though the empty-arange path would have been harmless anyway.

### Checklist
- [x] Algorithm matches reference (byte-identical to prior per-region semantics)
- [x] NaN handling is correct (unchanged from prior code)
- [x] Dask chunk boundaries handled correctly (regions computes whole-array; delegation unchanged)
- [ ] Benchmark exists or is not needed (none for regions; suggested above)
- [x] README feature matrix updated (not applicable; no API change)

brendancol (commented):
## Follow-up review (after db783e80)

Both prior findings are addressed:

- Suggestion (benchmark): added a `Regions` asv benchmark in `benchmarks/benchmarks/zonal.py` over many-region 0/1 noise, parametrized across numpy, cupy, and dask. Smoke-tested on all three.
- Nit (extra temp): the relabel now gathers with `remap[labeled[sel]]` instead of building a full-size `painted` array first. Output is unchanged.

Re-ran the region tests (25 passed) and the full `test_zonal.py` suite (187 passed). No remaining findings.

Issue #1096, opened 2026-03-30
convolve_2d truncates float64 input to float32


brendancol:
**Describe the bug**

All four `convolve_2d` backends hardcode `.astype(np.float32)` on input data and `np.float32`/`f4` on output arrays. When float64 data is passed, half the mantissa bits are silently discarded.

Affected functions:
- `_convolve_2d_numpy` (line 290): `data = data.astype(np.float32)`
- `_convolve_2d_dask_numpy` (line 332): `data = data.astype(np.float32)`
- `_convolve_2d_cupy` (line 396): `data = data.astype(cupy.float32)`
- `_convolve_2d_dask_cupy` (line 405): `data = data.astype(cupy.float32)`

`convolve_2d` is used internally by `convolution_2d`, `focal.mean`, `focal.hotspots`, and other downstream functions.

**Expected behavior**

Integer inputs should be promoted to float32 (to avoid integer overflow in weighted sums). Float64 inputs should stay float64. Output dtype should match the promoted input dtype.

**Reproduction**

```python
import numpy as np
from xrspatial.convolution import convolve_2d

data = np.array([[1e7 + 0.01, 1e7 + 0.02],
                 [1e7 + 0.03, 1e7 + 0.04],
                 [1e7 + 0.05, 1e7 + 0.06],
                 [1e7 + 0.07, 1e7 + 0.08]], dtype=np.float64)
kernel = np.array([[0, 1, 0], [1, 1, 1], [0, 1, 0]], dtype=np.float64)

result = convolve_2d(data, kernel)
print(result.dtype)   # float32 -- should be float64
# The 0.0x differences are lost in float32
```

**Affected code**

`xrspatial/convolution.py` -- all four backend functions for `convolve_2d`

PR #3221, opened 2026-06-10
Preserve input float dtype in focal.mean() across all backends (#3214)

### LINKED ISSUE #3214: focal.mean() returns float32 on cupy backends and float64 on numpy, losing precision on large-offset rasters
**Describe the bug**

`focal.mean()` handles dtype differently on each backend:

- numpy and dask+numpy: `mean()` runs `agg.data.astype(float)`, so the output is always float64, even for float32 input.
- cupy and dask+cupy: `_mean_cupy()` and `_mean_dask_cupy()` force `cupy.float32`, so the output is always float32, even for float64 input.

Two consequences:

1. The same call returns a different dtype depending on the backend.
2. float64 rasters lose precision on GPU. For values with a large offset (UTM coordinates, epoch timestamps, elevations in mm), float32 resolution is larger than the focal-mean variation itself:

```python
from xrspatial import mean

rng = np.random.default_rng(1)
data = (1e7 + rng.random((8, 8))).astype(np.float64)

np_mean = mean(xr.DataArray(data))
cp_mean = mean(xr.DataArray(cupy.asarray(data)))

diff = np.abs(np_mean.data - cp_mean.data.get())
print(diff.max())                                   # 0.577
print(np_mean.data.max() - np_mean.data.min())      # 0.420  <- true spread
```

The GPU error (0.577) is larger than the entire spread of the true means (0.420), so the smoothing result on the GPU backends carries no signal for this kind of input. This is the same failure class as #2831 (GPU std/var collapsing on large-offset rasters).

**Expected behavior**

`mean()` follows the dtype contract the rest of the library already uses (`apply()` and `focal_stats()` via #2769, `convolve_2d()` via #1096) on all four backends:

- float64 in, float64 out
- float32 in, float32 out
- integer in, float32 out

**Additional context**

`apply()` and `focal_stats()` were fixed in #2769 with `_promote_float`; `mean()` was left out. The fix is to use `_promote_float` in `mean()` and drop the hardcoded `cupy.float32` casts in `_mean_cupy` / `_mean_dask_cupy`.

brendancol:
Closes #3214

- `mean()` now uses `_promote_float` on every backend: float64 stays float64, float32 stays float32, integers promote to float32. Previously numpy/dask+numpy always returned float64 and cupy/dask+cupy always returned float32, so the output dtype depended on the backend.
- Drops the hardcoded `cupy.float32` casts in `_mean_cupy` / `_mean_dask_cupy`. float64 rasters keep their precision on GPU; with values near 1e7 the old float32 path's error (0.58) was larger than the spread of the true focal means (0.42).
- `excludes` values are cast to the working dtype before matching, so exclusion behaves the same on CPU and GPU for float32 rasters.

This is the same contract `apply()` and `focal_stats()` got in #2769 and `convolve_2d()` got in #1096; `mean()` was the holdout.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy. All four ran locally on a CUDA host.

The diff also updates one row of `.claude/sweep-accuracy-state.csv` (accuracy-sweep bookkeeping for the focal module).

Test plan:
- [x] New dtype tests: float64 preserved, float32 preserved, int promotes to float32, parametrized across backends
- [x] New GPU large-offset parity test (numpy vs cupy at offset 1e7, rtol 1e-12)
- [x] New excludes-dtype matching test
- [x] Full `xrspatial/tests/test_focal.py`: 245 passed on a GPU host
- [x] `test_accessor.py` + `test_emerging_hotspots.py`: 113 passed


brendancol (commented):
## PR Review: Preserve input float dtype in focal.mean() across all backends (#3214)

None found.

- [ ] `test_mean_excludes_match_in_working_dtype_3214` (xrspatial/tests/test_focal.py) only runs on numpy. The cupy path has its own excludes cast (`cupy.asarray(excludes, dtype=fdtype)` in `_mean_cupy`, xrspatial/focal.py:262-264), so the float32 exclusion-match behavior should be asserted there too. Parametrizing the test over numpy and cupy covers it.
- [ ] This PR changes integer-input behavior on the CPU backends: int rasters used to come back float64 and now come back float32, so int values above 2^24 are no longer exact in the output. That matches the `apply()`/`focal_stats()`/`convolve_2d()` contract, but the `mean()` docstring (xrspatial/focal.py, Returns section) says nothing about output dtype. One sentence stating the dtype contract would prevent surprise.

- The two `F401 'cupy' imported but unused` warnings at test_focal.py:1688 and 1704 predate this PR (from the #2730 tests) and are correctly left alone here.

### What looks good
- All four backends now share one dtype rule, applied at the dispatch boundary (`mean()`) and inside the chunk function (`_mean_cupy`), so the dask+cupy path is covered both lazily and per chunk.
- Casting `excludes` to the working dtype closes a cross-backend mismatch that the float32 change would otherwise have introduced (CPU comparing float32 cells against float64 excludes).
- The large-offset GPU test pins parity at rtol 1e-12 and asserts the focal-mean spread survives, which is the actual failure mode from the issue, not just a dtype check.
- A `time_mean` benchmark already exists in benchmarks/benchmarks/focal.py, so no benchmark gap.

### Checklist
- [x] Algorithm matches reference (no algorithm change; dtype handling only)
- [x] All implemented backends produce consistent results (verified on a CUDA host)
- [x] NaN handling is correct (NaN excludes survive the dtype cast; `_equal_numpy` uses isnan)
- [x] Edge cases are covered by tests (float64/float32/int inputs, large offset, excludes matching)
- [x] Dask chunk boundaries handled correctly (no depth changes; astype is lazy)
- [x] Benchmark exists
- [x] README feature matrix unchanged (no backend support change)
- [ ] Docstring does not state the output dtype contract (see suggestion above)

brendancol (commented):
## Follow-up review (commit 83be5d1)

Both suggestions from the first pass are addressed:

- `test_mean_excludes_match_in_working_dtype_3214` is now parametrized over numpy and cupy and handles the device-array extraction with `.get()`. Verified both variants pass on a CUDA host.
- The `mean()` Returns section now states the dtype contract (float preserved, int promotes to float32, matching `apply` and `focal_stats`).

No new findings. The pre-existing F401 nits remain out of scope.

Disposition of first-pass findings: 2 suggestions fixed, 1 nit dismissed (pre-existing, untouched by this PR), 0 blockers.

PR #3224, opened 2026-06-10
Resolve focal.apply() default func per backend (#3215)

### LINKED ISSUE #3215: focal.apply() default func crashes on cupy and dask+cupy backends
**Describe the bug**

`focal.apply()` defaults to `func=_calc_mean`, an `@ngjit` CPU function. The cupy and dask+cupy paths launch `func` as a CUDA kernel (`func[griddim, blockdim](...)` in `_focal_stats_func_cupy`), so calling `apply()` without an explicit `func` breaks on GPU-backed rasters:

```python
import cupy, xarray as xr
from xrspatial.convolution import circle_kernel

agg = xr.DataArray(cupy.random.random((8, 10)))
apply(agg, circle_kernel(1, 1, 1))
# TypeError: 'CPUDispatcher' object is not subscriptable
```

The same call works on numpy and dask+numpy. On dask+cupy the graph builds fine and the error only shows up at compute time.

**Expected behavior**

`apply(agg, kernel)` works on all four backends. The default func should resolve per backend: `_calc_mean` on the CPU paths, `_focal_mean_cuda` on the GPU paths. Both compute a focal mean, so results stay consistent.

**Proposed fix**

- Change the signature default to `func=None` and resolve it per backend before dispatch.
- Passing `func` explicitly keeps current behavior. No parameter is renamed, so no deprecation shim is needed.
- Also document the `name` parameter in the `apply()` docstring; it is in the signature (`name='focal_apply'`) but missing from the Parameters section, while `mean`, `focal_stats`, and `hotspots` all document theirs.

Found by the API consistency sweep on the focal module (2026-06-10), verified on a CUDA host.

brendancol:
Closes #3215

- `apply()` defaulted to `func=_calc_mean`, an `@ngjit` CPU function, but the cupy and dask+cupy paths launch `func` as a CUDA kernel. Calling `apply(agg, kernel)` with no explicit `func` raised `TypeError: 'CPUDispatcher' object is not subscriptable` on GPU-backed rasters (at compute time for dask+cupy).
- The default is now a `None` sentinel resolved per backend before dispatch: `_calc_mean` on numpy/dask+numpy, `_focal_mean_cuda` on cupy/dask+cupy. Both compute a focal mean. Passing `func` explicitly behaves exactly as before, so nothing breaks.
- The `apply()` docstring gains the `name` parameter, which was in the signature but missing from the Parameters section (the other three focal functions document theirs), and the `func` entry now spells out the CPU-vs-GPU callable requirements.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy (verified on a CUDA host).

The branch also carries the api-consistency sweep state update for the focal module.

Test plan:
- [x] `test_apply_default_func_numpy_3215` / `_dask_numpy_3215`: default matches explicit `_calc_mean`
- [x] `test_apply_default_func_cupy_3215`: default no longer raises on cupy, matches explicit `_focal_mean_cuda` and the numpy default
- [x] `test_apply_default_func_dask_cupy_3215`: dask+cupy default computes, interior matches single-GPU cupy
- [x] Full `xrspatial/tests/test_focal.py`: 237 passed (GPU tests included, CUDA host)


brendancol (commented):
## PR Review: Resolve focal.apply() default func per backend (#3215)

None.

None.

- [ ] `inspect.signature(apply).parameters['func'].default` changes from `_calc_mean` to `None`. Nobody should depend on that, and `func=None` was never usable before (it crashed in the CPU path too), so this is safe; mentioning it here for the record.

### What looks good
- The resolution block (focal.py:699-708) runs after `_resolve_raster_alias` and `_validate_raster` but before the 3D per-band recursion, so a 3D input resolves the default once and each band call gets a concrete func.
- The `da is not None and is_dask_cupy(agg)` guard keeps the no-dask install working; `is_cupy_array` handles the no-cupy install.
- Parity between the two defaults holds because `apply()` enforces binary kernels via `_validate_binary_kernel`: with all-ones weights, `_focal_mean_cuda`'s weighted mean equals `_calc_mean`'s nanmean. All-NaN neighborhoods give NaN on both paths (nanmean of NaNs vs `w_sum == 0`).
- Tests cover all four backends and mirror the comparison conventions of the existing `test_apply_cupy` / `test_apply_dask_cupy` (interior-only compare for dask+cupy, rtol=1e-4 for float32 GPU output).
- Docstring now documents `name` and spells out the CPU vs GPU callable requirements.

### Checklist
- [x] Algorithm matches reference (focal mean, both default implementations agree on binary kernels)
- [x] All implemented backends produce consistent results (verified on a CUDA host, 237 tests pass)
- [x] NaN handling is correct (NaN result for empty neighborhoods on both paths)
- [x] Edge cases are covered by tests (int input promotion, dask chunk interiors)
- [x] Dask chunk boundaries handled correctly (no change to depth/boundary plumbing)
- [x] No premature materialization (dask graphs stay lazy; `.compute()` only in tests)
- [x] Benchmark not needed (no performance-relevant change; dispatch happens once per call)
- [x] README feature matrix unchanged on purpose (Apply already lists all four backends)

The state-CSV change in the diff is the api-consistency sweep bookkeeping for the focal module and rides along by design.

PR #3226, opened 2026-06-10
Make focal output dtype consistent across backends

### LINKED ISSUE #3217: focal: mean() dtype differs by backend; apply()/focal_stats() dask meta advertises float64 but computes float32
**Describe the bug**

Two related dtype problems in `xrspatial/focal.py`:

1. `mean()` returns a different dtype depending on the backend. The function casts input to float64 (`agg.data.astype(float)`) before dispatch, and the numpy and dask+numpy paths keep that. But `_mean_cupy` re-casts to float32 (`cupy.asarray(data, dtype=cupy.float32)`) and `_mean_dask_cupy` does `data.astype(cupy.float32)`, so the GPU backends silently downcast float64 input and lose precision. `_mean_dask_cupy` also builds its graph with an untyped `meta=cupy.array(())`, so the lazy DataArray advertises float64 while `.compute()` returns float32.

2. `apply()` and `focal_stats()` advertise the wrong dtype on the dask backends. #2805 made the chunk functions preserve the input float dtype via `_promote_float`, but the `map_overlap` calls in `_apply_dask_numpy`, `_apply_dask_cupy`, and `_focal_stats_dask_cupy` still pass an untyped `meta` (`np.array(())` / `cupy.array(())`), which defaults to float64. For float32 or integer input, the lazy result claims float64 but computes to float32.

Measured on this host (all four backends live):

```
mean   in=f8 numpy       advertised=float64 computed=float64
mean   in=f8 cupy        advertised=float32 computed=float32
mean   in=f8 dask+numpy  advertised=float64 computed=float64
mean   in=f8 dask+cupy   advertised=float64 computed=float32   <- mismatch
apply  in=f4 dask+numpy  advertised=float64 computed=float32   <- mismatch
apply  in=i4 dask+cupy   advertised=float64 computed=float32   <- mismatch
fstats in=f4 dask+numpy  advertised=float64 computed=float32   <- mismatch
```

**Expected behavior**

- `mean()` returns float64 on all four backends, matching the CPU contract and the docstring examples, with no precision loss on the GPU paths.
- The dask graphs for `apply()` and `focal_stats()` advertise the dtype they actually compute.

Same class of bug as #2682 (aspect) and #2723 (proximity), which were fixed by passing a typed `meta`/`dtype` to the dask graph constructors.

**Additional context**

Found by the metadata propagation sweep against `focal`. attrs, coords, dims, and `.name` checked clean across all four backends on the same run.

brendancol:
Closes #3217

- `mean()`: drop the hardcoded float32 cast in `_mean_cupy` and `_mean_dask_cupy`. The function already casts input to float64 before dispatch, so all four backends now return float64 and the GPU result matches the CPU result exactly (the float32 cast cost ~1e-4 relative error).
- `apply()` / `focal_stats()` / `mean()` dask paths: pass a typed `meta` (`dtype=data.dtype`) to every `map_overlap` call, so the lazy DataArray advertises the dtype the chunks actually compute. Before, float32 and integer input advertised float64 but computed float32. Same fix as #2682 (aspect) and #2723 (proximity).

Backend coverage: numpy, cupy, dask+numpy, dask+cupy all verified live on this host (CUDA available).

Test plan:
- [x] New parametrized tests: `test_mean_dtype_consistent_across_backends_3217` (4 backends x 3 input dtypes), `test_apply_dask_advertised_dtype_matches_computed_3217`, `test_focal_stats_dask_advertised_dtype_matches_computed_3217` (dask backends x 3 input dtypes), and `test_mean_gpu_matches_cpu_float64_3217` (exact CPU/GPU value parity).
- [x] Full `test_focal.py` suite: 258 passed.


brendancol (commented):
## PR Review: Make focal output dtype consistent across backends

None.

- [ ] `xrspatial/focal.py:385-402`: the `mean()` docstring's cupy example still shows output computed under the old float32 cast (`array([[0.47928995, ...]])` printed at float32 precision) and an outdated `cupy.core.core.ndarray` class path. With this PR the GPU path computes in float64, so the example no longer matches what a user would see. Worth refreshing while you are touching this contract.

- [ ] `xrspatial/tests/test_focal.py` (new `_computed_dtype` helper): the helper is defined mid-file between test sections. Fine as is, but the `general_checks.py` module is where the other cross-backend helpers live if it ever gets a second user.

### What looks good
- The fix targets exactly the five sites that produce or advertise the wrong dtype, and nothing else. `hotspots()` meta sites were already typed and are untouched.
- `_mean_cupy` keeps the device array and inherits the dispatched dtype instead of forcing float32; the new test asserts exact (not approximate) CPU/GPU equality in float64, which holds because both loops accumulate sequentially.
- Tests cover 4 backends x 3 input dtypes for `mean()` and both dask backends x 3 dtypes for `apply()`/`focal_stats()`, asserting advertised dtype == computed dtype, which is the actual regression.
- 4x4 input with (2, 2) chunks exercises multi-chunk map_overlap, so the typed meta is checked against real chunk boundaries.

### Checklist
- [x] Algorithm matches reference (no algorithm change; dtype plumbing only)
- [x] All implemented backends produce consistent results (verified live on CUDA host)
- [x] NaN handling unchanged and covered by existing suite (258 passed)
- [x] Edge cases covered (float64/float32/int32 inputs)
- [x] Dask chunk boundaries handled correctly (typed meta, depth unchanged)
- [x] No premature materialization (meta typing is graph-construction only)
- [x] Benchmark not needed (bug fix, no new function); note GPU mean now runs in float64, so a throughput drop on large GPU rasters is expected and intentional
- [x] Docstrings: see suggestion about the stale cupy example

brendancol (commented):
## Follow-up review (after 1f202245)

The suggestion from the first pass is addressed: the mean() cupy docstring example now shows the float64 output (0.47928994, verified by running the example on a CUDA host) and the current cupy.ndarray class path.

Disposition of first-pass findings:
- Suggestion (stale cupy docstring example): fixed in 1f202245.
- Nit (_computed_dtype helper location): left in test_focal.py. It has a single consumer; moving a 4-line helper to general_checks.py before a second user exists adds indirection for no benefit.

No new findings. Full focal suite still passes (258).

PR #3227, opened 2026-06-10
Fix hotspots() docstring: backend list and kernel weight semantics (#3216)

### LINKED ISSUE #3216: hotspots() docstring contradicts actual backend support and kernel semantics
**Describe the bug**

Two statements in the `hotspots()` docstring do not match what the function does:

1. The `agg` entry says the input "Can be a NumPy backed, CuPy backed, or Dask with NumPy backed DataArray". The dispatch table includes a dask+cupy path (`dask_cupy_func=partial(_hotspots_dask_cupy, ...)`) and it works.
2. The `kernel` entry says "2D array where values of 1 indicate the kernel", i.e. a binary mask. Unlike `apply()` and `focal_stats()`, `hotspots()` does not reject non-binary kernels, and the Gi* formula a few lines up in the same docstring uses kernel weights `w_ij`. A weighted kernel is accepted and used as weights, which is correct for Gi*; only the parameter description is wrong.

**Expected behavior**

The docstring matches behavior: it lists all four supported backends and describes the kernel as a weight array, with 1s giving a plain membership mask.

Docs-only fix, no code change.

Found by the API consistency sweep on the focal module (2026-06-10).

brendancol:
Closes #3216

Docs-only fix for two statements in the `hotspots()` docstring that contradict the function's behavior:

- The `agg` entry said input "Can be a NumPy backed, CuPy backed, or Dask with NumPy backed DataArray", but the dispatch table includes `dask_cupy_func=_hotspots_dask_cupy` and that path works. The entry now lists all four backends.
- The `kernel` entry said "2D array where values of 1 indicate the kernel" (a binary mask). `hotspots()` does not reject non-binary kernels (unlike `apply` and `focal_stats`), and the Gi* formula a few lines up in the same docstring uses kernel weights `w_ij`. The entry now describes the kernel as a weight array, with 0/1 kernels acting as a plain membership mask, and mentions the zero-sum rejection.

No code change. Weighted-kernel behavior was already pinned by `test_hotspots_gistar_weighted_kernel_matches_reference`, so no new test.

The branch also carries the api-consistency sweep state update for the focal module.

Test plan:
- [x] `flake8 xrspatial/focal.py` clean
- [x] Docstring sanity check (backend list and `w_ij` wording present)


brendancol (commented):
## PR Review: Fix hotspots() docstring: backend list and kernel weight semantics (#3216)

None.

- [ ] focal.py:1718-1722: the zero-sum kernel error message still says "The kernel is normalized by its sum, so a zero-sum kernel divides by zero." That rationale dates from the pre-Gi* implementation (#2771) and is stale since #2803: the current code passes the kernel to `convolve_2d` unnormalized and the z-score denominator is `S * sqrt((n*W2 - W^2)/(n-1))`. Since this PR exists to align hotspots' kernel wording with actual behavior, the error message is the same class of divergence and is one string away. `test_hotspots_kernel_zero_sum_2771` only matches the first sentence, so rewording is safe.

None.

### What looks good
- The new `agg` entry matches the dispatch table (all four backends, including `dask_cupy_func=_hotspots_dask_cupy` at focal.py:1731).
- The new `kernel` entry agrees with the Gi* formula in the same docstring (`w_ij` weights) and with the code: hotspots has no `_validate_binary_kernel` call, on purpose, while `apply`/`focal_stats` do. Weighted behavior is already pinned by `test_hotspots_gistar_weighted_kernel_matches_reference`.
- "Must not sum to zero" documents the existing guard at focal.py:1717.

### Checklist
- [x] Algorithm untouched (docs-only)
- [x] Backend list now matches `ArrayTypeFunctionMapping`
- [x] NaN handling unchanged
- [x] Edge cases unchanged; weighted-kernel behavior already covered by tests
- [x] No performance or dask implications
- [x] Docstring accurate after the change

The state-CSV change is the api-consistency sweep bookkeeping for the focal module.

brendancol (commented):
## Follow-up review after 592ccb4b

The one suggestion from the previous pass is addressed: the zero-sum kernel error at focal.py:1717-1722 no longer claims sum-normalization and now states the actual Gi* consequence. `test_hotspots_kernel_zero_sum_2771` still passes (it matches the unchanged first sentence), and the full hotspots test selection is green (55 passed). No new findings; docs and error text now agree with the implementation.

Issue #3225, opened 2026-06-10
focal: empty raster (0 rows) works on numpy but crashes on cupy and dask backends


brendancol:
**Describe the bug**

Backend behavior diverges for a degenerate empty raster (shape `(0, 5)`):

- numpy: `mean()`, `apply()`, and `focal_stats()` return an empty result with the input shape preserved.
- cupy: `mean()` fails with a raw `CudaAPIError: Call to cuLaunchKernel results in CUDA_ERROR_INVALID_VALUE` (the kernel launch gets a zero-sized grid from `cuda_args`).
- dask+numpy: `mean()` fails with `ValueError: The overlapping depth 1 is larger than your array 0.` from `map_overlap`.

```python
import numpy as np, xarray as xr, cupy
import dask.array as da
from xrspatial import mean

d = np.empty((0, 5))
mean(xr.DataArray(d, dims=['y', 'x']))                                # works, shape (0, 5)
mean(xr.DataArray(cupy.asarray(d), dims=['y', 'x']))                  # CudaAPIError
mean(xr.DataArray(da.from_array(d, chunks=(3, 3)), dims=['y', 'x']))  # ValueError from map_overlap
```

**Expected behavior**

Pick one of the two behaviors and apply it on every backend: either return an empty result of the input shape (matching numpy today), or reject empty rasters up front in `_validate_raster` with a clear message. Either is fine; a raw `cuLaunchKernel` error is not.

**Additional context**

Found by the test-coverage sweep (#3220) while probing geometric edge cases. The companion test PR adds a numpy-only empty-raster test and leaves the cupy/dask behavior unpinned until this is resolved.

Issue #3219, opened 2026-06-10
hotspots() silently returns all zeros when the raster contains Inf


brendancol:
**Describe the bug**

`hotspots()` returns an all-zeros classification when the input raster contains an Inf cell. No warning, no error. Same result on all four backends (numpy, cupy, dask+numpy, dask+cupy).

The cause: with Inf present, the global standard deviation (`np.nanstd` / `cupy.nanstd`) evaluates to NaN. `_gistar_global_stats` only guards `global_std == 0`, and `NaN == 0` is False, so the degenerate-input check passes. Every Gi* z-score then comes out NaN, and `_gistar_zscore` maps non-finite z-scores to 0.0. The output classifies every cell as "no significance".

```python
import numpy as np, xarray as xr
from xrspatial.focal import hotspots

data = np.array([[1., 2., 3.], [4., np.inf, 6.], [7., 8., 9.]])
kernel = np.array([[0, 1, 0], [1, 1, 1], [0, 1, 0]], dtype=np.float64)
hotspots(xr.DataArray(data, dims=['y', 'x']), kernel)
# -> all zeros, no warning, no error
```

**Expected behavior**

Other degenerate inputs raise: a constant raster raises `ZeroDivisionError`, an all-NaN raster and a single-valid-cell raster raise `ValueError` (#2843 made the dask paths raise these at compute time too). A raster whose global std is NaN because of Inf should be treated the same way: raise a clear error (eagerly on numpy/cupy, at compute time on the dask paths) instead of producing a meaningless all-zeros result.

**Additional context**

Found by the test-coverage sweep while probing Inf inputs (no focal test exercises Inf today). The companion test-coverage PR adds Inf tests for `mean`/`focal_stats`, where behavior is consistent across backends, but deliberately leaves the hotspots+Inf behavior unpinned so this bug can be fixed without rewriting a test.

PR #3230, opened 2026-06-10
Add focal edge-case and parameter tests from the test-coverage sweep

### LINKED ISSUE #3220: focal: missing tests for Inf inputs, mean() NaN/excludes/passes behavior, degenerate raster shapes, and dask+cupy boundary modes
The test-coverage sweep audited `xrspatial/focal.py` (60 commits, 1742 LOC) against `xrspatial/tests/test_focal.py` (1996 LOC). Backend coverage for the four public functions (`mean`, `apply`, `focal_stats`, `hotspots`) is solid, but several edge-case and parameter paths are never exercised.

**NaN / Inf edge cases (category 2)**
- Nothing in `test_focal.py` passes an Inf or -Inf input to any focal function. Probing shows consistent cross-backend behavior today (Inf propagates through mean/sum/max/range; std/var of a window containing Inf is NaN on both CPU and GPU), but no test pins that down.
- `mean()` is never tested with a NaN input on any backend, even though its default `excludes=[np.nan]` gives NaN cells specific semantics: a NaN center cell is left unchanged while neighbors use `nanmean`. The boundary-mode and parity tests for `mean` all use NaN-free random data.

**Geometric edge cases (category 3)**
- No test runs `mean`/`apply`/`focal_stats` on a 1x1 raster, an Nx1/1xN strip, or an empty (0-row) raster. Probing shows all backends handle these and agree, but kernel-bound bugs in this class of code regress exactly here.

**Parameter coverage (category 4)**
- `mean(passes=...)` is only ever tested at the default `passes=1`. The semantics (`passes=2` equals `mean(mean(x))`) are unasserted on every backend.
- `mean(excludes=...)` has no behavioral test. `test_mean_excludes_default_not_shared` calls `mean(agg, excludes=[0.0])` but discards the result; no test asserts that a sentinel-valued cell (e.g. -9999) is left unchanged. Parity verified across all four backends.

**Backend x parameter (categories 1/4)**
- The dask+cupy backend is never exercised with a non-default `boundary` mode for `mean`, `apply`, or `focal_stats`. The #2730 tests cover single-GPU cupy only, and the existing dask+cupy tests use the default `boundary='nan'`. Probing confirms parity for 'nearest'/'reflect'/'wrap' today.

All probes ran on a CUDA host against all four backends. Behavior agrees everywhere except hotspots with an Inf cell, which silently classifies everything to 0; that is filed separately as #3219. The fix for this issue is tests only.

brendancol:
Closes #3220

Test-only PR from the test-coverage sweep against `xrspatial/focal.py`. No source changes.

- Adds Inf-input tests for `mean()` and `focal_stats()` (no focal test exercised Inf before), a NaN-input test for `mean()` pinning the `excludes=[np.nan]` semantics, and behavioral tests for `mean(passes=2)` and `mean(excludes=[-9999])`.
- Adds degenerate-shape tests: 1x1 raster, 1x6 and 6x1 strips (all four backends), and a numpy-only empty-raster test.
- Adds dask+cupy parity tests for the 'nearest'/'reflect'/'wrap' boundary modes of `mean`/`apply`/`focal_stats`; that backend previously only ran with the default `boundary='nan'`.
- Updates the sweep state CSV (focal row only).

Two bugs surfaced while probing and are filed separately rather than fixed or pinned here:
- #3219: `hotspots()` silently classifies everything to 0 when the raster contains Inf (NaN global std slips past the `std == 0` guard).
- #3225: an empty raster works on numpy but crashes cupy (raw `CudaAPIError`) and dask (`map_overlap` depth error).

Backend coverage: numpy, cupy, dask+numpy, dask+cupy. All 36 new tests executed on a CUDA host; none were skipped.

Test plan:
- [x] `pytest xrspatial/tests/test_focal.py -k 3220` -> 36 passed (CUDA host, GPU backends executed)
- [x] `pytest xrspatial/tests/test_focal.py` -> 269 passed
- [x] `flake8 xrspatial/tests/test_focal.py` -> no new findings (two pre-existing F401s untouched)


brendancol (commented):
## PR Review: Add focal edge-case and parameter tests from the test-coverage sweep

Test-only PR; no source changes. Review covers correctness of the new assertions, backend coverage, and test quality.

None.

- [ ] `test_mean_nan_input_3220`, `test_mean_excludes_sentinel_3220`, `test_mean_inf_input_3220`, and `test_focal_stats_inf_input_3220` (xrspatial/tests/test_focal.py:2042-2155) all use 3x3 data with `chunks=(3, 3)`, so the dask and dask+cupy variants run as a single chunk and never cross a chunk seam. The whole point of running these under dask is the `map_overlap` halo behavior around the NaN/Inf/sentinel cell. Use `chunks=(2, 2)` so the special cell sits next to a real chunk boundary. (`test_mean_passes_3220` already gets this right with 5x5 data on (3, 3) chunks, as do the strip tests.)

- [ ] In `test_mean_excludes_sentinel_3220`, `test_mean_passes_3220`, and `test_focal_stats_inf_input_3220`, the parity assertion against a freshly computed numpy reference is a self-comparison when `backend='numpy'`. Harmless and deterministic, but the hardcoded-value asserts are what carry that param. Fine to leave.
- [ ] `apply()` and `focal_stats()` call `custom_kernel()` internally, so wrapping `_sweep_cross_kernel` in `custom_kernel(...)` at the call sites is redundant. It is idempotent and matches older tests in this file, so also fine to leave.

### What looks good
- The expected NaN-input matrix for `mean()` is hand-derived per window and matches the documented `excludes=[np.nan]` semantics (center preserved, neighbors via nanmean).
- The Inf tests pin only the behavior that is consistent across backends, and deliberately leave hotspots+Inf (#3219) and non-numpy empty rasters (#3225) unpinned instead of enshrining buggy output. Both bugs are filed with repros.
- All 36 tests executed on a CUDA host with zero skips, so the cupy and dask+cupy assertions were actually observed running.
- The dask+cupy boundary-mode test closes the last untested backend x boundary combination for mean/apply/focal_stats.
- State CSV diff is one row, one physical line, LF endings.

### Checklist
- [x] Algorithm matches reference (expected values hand-computed and probe-verified)
- [x] All implemented backends produce consistent results (asserted in the new tests)
- [x] NaN handling is correct (NaN-input test pins it)
- [x] Edge cases are covered by tests (1x1, strips, empty on numpy, all-Inf)
- [ ] Dask chunk boundaries handled correctly (see suggestion: four tests are single-chunk under dask)
- [x] No premature materialization concerns (test-only change)
- [x] Benchmark not needed (no source change)
- [x] Docstrings not applicable (no public API change)

brendancol (commented):
Follow-up review after 2fb101d6.

- Suggestion (single-chunk dask tests): fixed. `test_mean_nan_input_3220`, `test_mean_excludes_sentinel_3220`, `test_mean_inf_input_3220`, and `test_focal_stats_inf_input_3220` now use `chunks=(2, 2)`, so the NaN/Inf/sentinel cell sits next to a real chunk seam under dask and dask+cupy. All 36 tests re-ran green on the CUDA host. `test_mean_passes_3220` (5x5 on (3, 3) chunks, already multi-chunk) and the 1x1 test (single chunk is unavoidable) are unchanged.
- Nit (numpy self-comparison in the parity asserts): dismissed. The hardcoded-value asserts carry the numpy param; keeping the body uniform across params is worth the redundant comparison.
- Nit (redundant `custom_kernel` wrap): dismissed. The wrap is idempotent and matches how older tests in this file build kernels; removing it only here would make the file inconsistent.

No remaining findings.

PR #3232, opened 2026-06-10
Make focal kernel memory guard dtype-aware (#3223)

### LINKED ISSUE #3223: focal kernel memory guard still assumes float32 internals after #2805
## Description

`_check_kernel_vs_raster_memory()` in `xrspatial/focal.py` budgets 4 bytes per cell, with a comment saying "focal internals cast to float32". That was true when the guard was added, but #2805 changed `apply()` and `focal_stats()` to preserve float64 input, so for float64 rasters the padded raster and kernel scratch arrays now take 8 bytes per cell. The guard underestimates the real allocation by 2x.

The guard rejects combos above 50% of available memory. With the 2x underestimate, a float64 raster + kernel combo can pass the guard and then allocate up to ~100% of available memory, which is the OOM scenario the guard was added to stop (#1284).

Repro on a host with ~45 GB available:

```python
import numpy as np, math
from xrspatial.focal import _check_kernel_vs_raster_memory
from xrspatial.convolution import _available_memory_bytes

avail = _available_memory_bytes()
k = int(math.isqrt(int(0.30 * avail / 8))) | 1     # 41173 here
_check_kernel_vs_raster_memory(np.ones((k, k)), 101, 101, 'apply')
# passes; the actual float64 footprint is ~60% of available,
# and sized near the guard's threshold it reaches ~100%
```

## Expected behavior

The guard budgets bytes per cell from the dtype the focal internals will actually use (`_promote_float` of the input dtype): 8 bytes for float64 input, 4 otherwise.

brendancol:
Closes #3223

- `_check_kernel_vs_raster_memory` budgeted a flat 4 bytes per cell ("focal internals cast to float32"). That stopped being true when #2805 made `apply()` and `focal_stats()` preserve float64, so the guard underestimated float64 allocations by 2x and a kernel + raster combo could pass the check and then use up to ~100% of available memory, the OOM the guard exists to stop (#1284).
- The guard now takes an `itemsize` argument. `apply()` and `focal_stats()` pass `np.dtype(_promote_float(agg.dtype)).itemsize` (8 for float64 input, 4 otherwise). `hotspots()` keeps the default 4 since it computes in float32 on every backend.
- The state CSV commit is from the security sweep that found this.

Backend coverage: the guard runs at the public entry points before dispatch, so all four backends get the same check. No backend code changed.

Test plan:
- [x] `test_apply_oversize_kernel_accounts_for_float64_3223`: float64 combo sized to pass the old 4-byte budget now raises MemoryError; the same combo as float32 still runs
- [x] `test_focal_stats_oversize_kernel_accounts_for_float64_3223`
- [x] Existing #1284 guard tests unchanged and passing
- [x] Full `test_focal.py` suite: 235 passed


brendancol (commented):
## PR Review: Make focal kernel memory guard dtype-aware (#3223)

None.

None.

- [ ] A companion test asserting that `hotspots()` with float64 input is still budgeted at 4 bytes/cell (i.e. not over-rejected) would pin down the asymmetry the comment at focal.py:1758 describes. The existing #1284 hotspots test plus the comment already make the intent clear, so this is optional.

### What looks good
- The `itemsize` is derived from `_promote_float(agg.dtype)`, the same function the internals use to pick their compute dtype, so the budget and the actual allocations cannot drift apart again the way the hardcoded 4 did after #2805.
- `hotspots()` keeps the default 4 with a comment explaining why (it computes in float32 on every backend); passing the promoted itemsize there would have over-rejected float64 input.
- The new tests are deterministic (patched `_available_memory_bytes`, fixed 1 MB budget) and test both sides: float64 rejected at exactly the sizes that used to slip through, float32 still allowed.
- The guard runs at the public entry points before backend dispatch, so the 3D per-band recursion and all four backends inherit the corrected budget.
- `np.dtype(...)` works for numpy, cupy, and dask-backed DataArrays alike since `.dtype` is a numpy dtype in all cases.

### Checklist
- [x] Algorithm matches reference (budget formula unchanged; only bytes/cell corrected)
- [x] All implemented backends produce consistent results (guard is backend-independent)
- [x] Edge cases are covered by tests (pass/reject boundary on both dtypes)
- [x] No premature materialization or unnecessary copies (dtype lookup only, no data access)
- [x] Benchmark exists or is not needed (validation-path change, no kernel work)
- [x] Docstrings present and accurate (guard docstring updated with the #2805 history)

brendancol (commented):
Follow-up after the nit: added test_hotspots_float64_keeps_float32_budget_3223, which confirms float64 input to hotspots() stays on the 4 bytes/cell budget and is not over-rejected. Full focal suite: 236 passed. Nothing further from me.

PR #3233, opened 2026-06-10
Hoist input cast out of the cupy focal_stats loop (#3231)

### LINKED ISSUE #3231: focal_stats cupy backend re-casts the full raster once per stat
`_focal_stats_cupy()` in `xrspatial/focal.py` runs

```python
for stats in stats_funcs:
    data = agg.data.astype(_promote_float(agg.data.dtype))
    stats_data = _stats_cupy_mapper[stats](data, kernel)
```

The astype is inside the loop, so a `focal_stats()` call with the default 8 stats does 8 full-raster device casts. cupy's `astype` copies even when the dtype is unchanged (default `copy=True`), so each iteration allocates and fills a raster-sized device array that is identical to the previous one. Measured on this host: 8 casts of a 128 MB float64 raster cost 0.117s of pure copy bandwidth.

The cast result doesn't depend on the loop variable. Hoist it above the loop.

`_focal_stats_dask_cupy()` has the same textual pattern, but there the astype is lazy and the layers carry deterministic tokens, so the 8 graph layers dedupe when the per-stat results are concatenated. Only the eager cupy path pays for the repeat.

Found by the performance sweep over the focal module.

brendancol:
Closes #3231

- `_focal_stats_cupy()` cast `agg.data` with `astype(_promote_float(...))` inside the per-stat loop. cupy's `astype` copies even when the dtype is unchanged, so the default 8-stat call did 8 full-raster device casts where 1 is enough. The cast is now hoisted above the loop.
- The dask+cupy path is left alone: its astype layers are lazy and dedupe in the graph through deterministic tokens, so it never paid for the repeat.

Backend coverage: cupy only. numpy, dask+numpy, and dask+cupy paths are untouched.

Test plan:
- [x] New GPU regression test pins the call count (`_promote_float` runs once for the input plus once per stat for the output allocation); ran on this host with CUDA available
- [x] Full `xrspatial/tests/test_focal.py`: 234 passed, including the cupy/dask+cupy parity tests


brendancol (commented):
## PR Review: Hoist input cast out of the cupy focal_stats loop (#3231)

None.

None.

- [ ] `xrspatial/tests/test_focal.py:579-596`: the call-count assertion (`len(stats) + 1`) encodes the internal call pattern of `_focal_stats_func_cupy`, so an unrelated refactor that adds or removes a `_promote_float` call will fail this test with a confusing message. The in-test comment explains the formula, which is probably enough. No change requested.

### What looks good
- The hoist is behavior-preserving: `data` does not depend on the loop variable, and every `_focal_*_cuda` kernel only reads `data` and writes `out`, so sharing one cast across stats cannot leak state between them.
- The padded non-nan boundary path (`_focal_stats_cupy_boundary` -> `_focal_stats_cupy`) gets the same benefit for free.
- Leaving `_focal_stats_dask_cupy` alone is the right call; its lazy astype layers dedupe by token, so changing it would add diff for no runtime gain.
- The regression test is properly gated on CUDA+cupy and ran on this host. Existing parity tests (test_focal_stats_gpu, the 2730 boundary set, dask+cupy interior comparison) cover value correctness; 234 tests pass.

### Checklist
- [x] Algorithm matches reference (no numeric change; pure hoist)
- [x] All implemented backends produce consistent results (cupy parity tests pass; other backends untouched)
- [x] Edge cases are covered by tests (existing suite)
- [x] Dask chunk boundaries handled correctly (dask paths untouched)
- [x] No premature materialization or unnecessary copies (this PR removes 7 of them)
- [x] Benchmark exists or is not needed (micro-optimization; focal benchmarks unchanged)
- [x] Docstrings present and accurate (private helper, comment added)

PR #3236, opened 2026-06-11
Forward caller chunking to the coregister/auto_reproject output

### LINKED ISSUE #3234: coregister=True drops the caller's chunking; reproject output reverts to 512x512
**Describe the bug**

`.xrs.open_geotiff(..., coregister=True)` returns a dask array chunked at 512x512 no matter how the caller is chunked. The windowed read does pick up the caller's chunk size (`_infer_caller_y_chunk` feeds `kwargs.setdefault('chunks', ...)` in `accessor.py`), but the `reproject()` call that snaps the result onto the caller's grid is never given a `chunk_size`, so `_compute_chunk_layout` falls back to its 512 default.

```python
>>> study_area = open_geotiff("USGS_1_n35w112.tif", chunks=2048)
>>> study_area.data.chunksize
(2048, 2048)
>>> veg = study_area.xrs.open_geotiff("LF2024_EVC_CONUS.tif", unpack=True, coregister=True)
>>> veg.data.chunksize
(512, 512)
```

**Expected behavior**

Coregister puts the result on the caller's exact grid, so it should keep the caller's chunking too. A 2048-chunked study area should produce a 2048-chunked veg layer, and blockwise ops on the pair should line up chunk for chunk.

The `auto_reproject=True` path has the same gap: the windowed read uses the caller's chunks but the reprojected output reverts to 512.

**Additional context**

`reproject()` already accepts `chunk_size` (int or `(row, col)` tuple); `_open_geotiff_windowed` never passes it. Derive `chunk_size` from the caller's dask chunks (or from the explicit `chunks=` kwarg when the caller is in-memory) and forward it to both `reproject()` calls.

brendancol:
Closes #3234

- `.xrs.open_geotiff` now passes `chunk_size` to the `reproject()` call used by `coregister=True` and `auto_reproject=True`, so a dask result keeps the caller's chunk layout (or the explicit `chunks=` kwarg when one is given) instead of reverting to reproject's 512x512 default.
- Renamed the internal `_infer_caller_y_chunk` to `_infer_caller_chunk(obj, axis)` so the x-axis chunk can be inferred too. Single call site, no public API change.
- Documented the chunking behavior in the DataArray accessor docstring.

Backend coverage: the fix sits in the shared windowed-read path. `coregister` is CPU-only by design (numpy / dask+numpy); the `auto_reproject` chunk forwarding also covers dask+cupy callers.

Test plan:
- [x] `test_coregister_dask_template_keeps_caller_chunks`: dask template chunked (3, 2) gives output chunksize (3, 2), asymmetric to prove (y, x) ordering
- [x] `test_coregister_explicit_chunks_kwarg`: numpy template + `chunks=2` gives output chunksize (2, 2)
- [x] `test_auto_reproject_dask_template_keeps_caller_chunks`: auto_reproject output follows caller chunks
- [x] `test_open_geotiff_coregister.py`, `test_open_geotiff_resampling.py`, `test_accessor.py` all pass (90 tests)

### DIFF
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -237,7 +237,8 @@ def _open_geotiff_windowed(obj, source, *, auto_reproject=False,
     covers the full footprint even when the transform has curvature
     across the bbox.
     """
-    from .geotiff import open_geotiff, _read_geo_info, _extent_to_window
+    from .geotiff import (open_geotiff, _read_geo_info, _extent_to_window,
+                          _validate_chunks_arg)
     from .utils import _classify_backend
 
     if resampling not in _RESAMPLING_CHOICES:
@@ -320,7 +321,11 @@ def _open_geotiff_windowed(obj, source, *, auto_reproject=False,
 
     # Infer backend kwargs. Caller-supplied values always win.
     backend = _classify_backend(obj)
-    explicit_chunks = kwargs.get('chunks')
+    # Coerce here because reproject's chunk layout only unpacks builtin
+    # ints, while the read path accepts np.integer scalars.
+    explicit_chunks = _validate_chunks_arg(
+        kwargs.get('chunks'), allow_none=True
+    )
     if backend in ("cupy", "dask+cupy"):
         kwargs.setdefault('gpu', True)
     if backend in ("dask+numpy", "dask+cupy"):
@@ -977,10 +982,11 @@ class XrsSpatialDataArrayAccessor:
             The windowed portion of the GeoTIFF. In ``self``'s CRS when
             ``auto_reproject`` reprojection occurred, on ``self``'s exact
             grid when ``coregister`` is set, and otherwise in the file's
-            native CRS. A dask result chunks like an explicit ``chunks=``
-            kwarg when given, and otherwise like ``self``, including
-            through the ``auto_reproject`` / ``coregister`` reproject
-            step.
+            native CRS. The ``auto_reproject`` / ``coregister``
+            reproject step chunks its dask output like the explicit
+            ``chunks=`` kwarg when given, and otherwise like ``self``,
+            rather than reverting to
+            :func:`~xrspatial.reproject.reproject`'s own default.
         """
         return _open_geotiff_windowed(
             self._obj, source, auto_reproject=auto_reproject,
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -86,22 +86,22 @@ def _pick_representative_dataarray(ds, *, var=None):
     )
 
 
-def _infer_caller_y_chunk(obj):
-    """First y-axis chunk size of a dask-backed DataArray, or None."""
+def _infer_caller_chunk(obj, axis):
+    """First chunk size along *axis* of a dask-backed DataArray, or None."""
     try:
-        y_axis = obj.get_axis_num('y')
+        axis_num = obj.get_axis_num(axis)
     except (ValueError, KeyError):
         return None
     chunks_tuple = getattr(obj, 'chunks', None)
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Forward caller chunking to the coregister/auto_reproject output

- [ ] `xrspatial/accessor.py:323` — `explicit_chunks = kwargs.get('chunks')` captures the raw kwarg and forwards it to `reproject(chunk_size=...)` unconverted. `_validate_chunks_arg` deliberately accepts `np.integer` scalars for the read path and coerces them, but `_compute_chunk_layout` only special-cases builtin `int`, so `chunks=np.int64(2048)` with `coregister=True` now raises `TypeError: cannot unpack non-iterable numpy.int64 object`. Before this PR that call worked (the read coerced it; reproject never saw it). Coerce integer-likes with `int(...)` before forwarding, or run the captured value through `_validate_chunks_arg(..., allow_none=True)`.

- [ ] `xrspatial/tests/test_open_geotiff_coregister.py` — no test covers a tuple `chunks=(3, 2)` kwarg flowing through to the reproject output. It is the same forwarding path, but the tuple form is the one `_compute_chunk_layout` unpacks directly, so a one-line assertion would pin it down.

- [ ] `xrspatial/accessor.py:980-983` — the new Returns sentence says a dask result chunks "like ``self``", but the windowed read only infers the y-axis chunk (line 328), so a plain read against a template chunked (3, 2) comes back (3, 3). The exact (y, x) match only holds for the reproject output. Scoping the sentence to the reproject step would keep the docstring honest.

### What looks good
- Capturing `explicit_chunks` before the `setdefault` is the right move; reading `kwargs['chunks']` afterward would have conflated user intent with the inferred value.
- The asymmetric (3, 2) chunk test pins the (y, x) ordering, which is exactly the kind of thing that silently transposes.
- Forwarding into both the `coregister` and `auto_reproject` branches fixes the same gap in one pass instead of leaving the second branch for a follow-up.
- `None` falls through to reproject's existing defaults, so numpy callers without a `chunks=` kwarg see no behavior change.

### Checklist
- [x] Algorithm matches reference/paper (n/a — chunk plumbing only)
- [x] All implemented backends produce consistent results (CPU-only coregister guard unchanged; dask+cupy auto_reproject forwards the same value)
- [x] NaN handling is correct (untouched; nodata_arg path unchanged)
- [ ] Edge cases are covered by tests (np.integer and tuple chunks kwarg are not)
- [x] Dask chunk boundaries handled correctly (reproject computes per-chunk source windows; chunk size only changes tiling)
- [x] Docstrings present and accurate (modulo the nit above)

brendancol (commented):
## Follow-up review (after b4c66f4e)

- **Blocker (np.integer chunks crash): fixed.** The captured kwarg now goes through `_validate_chunks_arg(..., allow_none=True)` at `xrspatial/accessor.py:325`, which coerces `np.integer` scalars to builtin int before the value reaches `_compute_chunk_layout`. Verified `chunks=np.int64(2)` with `coregister=True` returns chunksize (2, 2) in the new test.
- **Suggestion (tuple chunks coverage): fixed.** `test_coregister_chunks_kwarg_tuple_and_np_integer` covers both the `(row, col)` tuple form and the np scalar form.
- **Nit (docstring overstatement): fixed.** The Returns sentence now scopes the chunk-matching claim to the reproject step.

One side effect worth noting, not a problem: invalid `chunks` values (e.g. a string) now raise from the accessor before the read starts, instead of from inside the read. The error message is identical because both call the same validator, so callers just see it earlier.

No new issues. 91 tests pass across the coregister, resampling, and accessor suites; flake8 is clean.

PR #3237, opened 2026-06-11
Defer the pack=True NaN guard into the dask graph so chunks compute once

### LINKED ISSUE #3235: to_geotiff(pack=True) computes the whole dask graph twice when no nodata sentinel is declared
**Describe the bug**
`_pack()` in `xrspatial/geotiff/_attrs.py` guards the integer restore with an eager NaN check:

```python
elif tgt.kind in ('i', 'u'):
    # ``isnull().any()`` forces a compute on dask; only reached on the
    # error path where no sentinel exists to fill an integer's holes.
    if bool(out.isnull().any()):
        raise ValueError(...)
```

The comment is wrong about when this runs: the branch fires whenever the packed target dtype is integer and no nodata sentinel is declared, including the normal success path. On dask-backed input, `bool(out.isnull().any())` executes the whole upstream graph at `to_geotiff(pack=True)` call time. The streaming writer then executes it again to produce the pixels.

Measured with a dask pretask counter on a 512x512 int16 source carrying SCALE/OFFSET but no GDAL_NODATA, read via `open_geotiff(src, unpack=True, chunks=128)` and written via `to_geotiff(out, pack=True)`: 16 chunk-decode tasks for the isnull/any check plus 16 for the write. Every source chunk decodes twice, so a computed pipeline feeding pack=True (say, slope over a mosaic) runs end to end twice.

The sentinel-present path doesn't have this problem: `out.fillna(nodata)` is lazy.

**Expected behavior**
Each source chunk computes once per pack round trip. For dask-backed data the NaN guard can run per chunk inside the graph and raise during the write's single compute. The numpy path can keep the eager call-time check.

**How to hit it**
Any CF-packed integer GeoTIFF with SCALE/OFFSET but no GDAL_NODATA tag, round-tripped via `open_geotiff(unpack=True, chunks=...)` then `to_geotiff(pack=True)`.

**Additional context**
Found in performance sweep pass 15 (2026-06-11); the sweep's module verdict is SAFE / IO-bound. Affected backend: the dask (CPU) write path with pack=True. numpy/cupy eager paths are unaffected, and dask+cupy crashes earlier (#3112) so it never reaches this code today.

brendancol:
Closes #3235.

`_pack`'s no-sentinel integer restore scanned for NaN with an eager `bool(out.isnull().any())`, which executed the entire upstream dask graph at `to_geotiff(pack=True)` call time. The streaming writer then executed the graph again for the pixels, so every source chunk computed twice. The guard now maps into the graph per chunk (`_pack_guard_no_nan`) for dask-backed data and raises from the write's single compute. numpy-backed data keeps the eager call-time check.

Measured on a 512x512 int16 source with SCALE/OFFSET and no GDAL_NODATA, read with `unpack=True, chunks=128`: decode-task executions drop from 32 to 16 (total task starts 71 to 33), round-trip parity unchanged.

Error timing for the dask NaN-with-no-sentinel case moves from the `to_geotiff` call to the write's compute; noted in `to_geotiff`'s `pack` docstring and in `_pack`. The write stays atomic (temp file plus rename), so a failed pack leaves no partial output at the destination.

Backend coverage: dask (CPU) is the affected path and gets the fix. numpy keeps its eager behavior. The guard itself is backend-agnostic (`np.isnan` dispatches on the array type, `meta=` preserves cupy backing) and has a cupy-chunk unit test; full dask+cupy pack round trips remain blocked upstream by #3112.

Also carries the geotiff performance-sweep state update for pass 15 (one CSV row).

Test plan:
- [x] 9 new tests in `tests/write/test_pack_lazy_nan_guard_3235.py`: fusion-proof per-chunk execution counter, eager/lazy round-trip parity, dask raise-at-compute, numpy raise-at-call, guard unit tests on numpy and cupy chunks
- [x] Existing pack suites pass (test_pack_3064, test_pack_band_subset_3161, test_pack_nodata_kwarg_3168, test_mask_and_scale_dtype_parity_3066, test_streaming): 111 passed, 1 skipped, 2 xfailed
- [x] Full write + attrs suites: 1148 passed, 1 skipped, 2 xfailed
- [x] GPU-marked tests executed on a CUDA host


brendancol (commented):
## PR Review: Defer the pack=True NaN guard into the dask graph so chunks compute once

- None found.

- [ ] `tests/write/test_pack_lazy_nan_guard_3235.py:50` -- the execution-counter test pins the tiled `.tif` path only. A `.vrt` destination consumes the packed graph through `_write_vrt_tiled`'s delayed tile writes instead of `_write_streaming`, and nothing covers that the deferred ValueError still surfaces (same message) through `_safe_write_tile`'s collect-and-reraise machinery. One small test writing the NaN no-sentinel dask array to a `.vrt` path would close that.

- [ ] `xrspatial/geotiff/_attrs.py:1823` -- the guard is mapped even when `out.dtype` is non-float, where it can only be a no-op (the `(data - offset) / scale` arithmetic promotes to float in every reachable case). The chunk-level `dtype.kind == 'f'` check already makes this safe; this is graph noise in an unreachable case at most. Fine to leave.
- [ ] `xrspatial/geotiff/_attrs.py:1707` -- `bool(np.isnan(chunk).any())` allocates a chunk-sized bool mask per chunk. `np.isnan(np.min(chunk))` would avoid it, but the mask is chunk/8 bytes and the `isnan(...).any()` form matches the rest of the module. Consistency wins; noting for completeness only.

### What looks good
- The fix lands exactly where the issue points: the eager `isnull().any()` is gone for dask input, and the measurement in the PR body (decode executions 32 to 16) matches an independent rerun during this review.
- `meta=out.data._meta` on the `map_blocks` call keeps cupy backing intact and avoids the meta-inference probe that would otherwise execute the guard once extra at graph build.
- The counter test threads a counting identity block through the graph rather than parsing fused task names, so it survives scheduler fusion changes.
- The error message is byte-identical between the eager and deferred paths, and `test_pack_dask_nan_no_sentinel_packed_graph_is_lazy` pins that `_pack` itself stays lazy even when NaN is present.
- The docstring claim about atomicity was verified against `_writer.py` (mkstemp + `os.replace` + unlink on failure) rather than asserted.
- The sweep CSV change is a single row with LF endings preserved, avoiding the known CRLF whole-file conflict hazard.

### Checklist
- [x] Algorithm matches reference (inverse-scale semantics unchanged; guard placement before round/astype is correct)
- [x] All implemented backends produce consistent results (numpy eager, dask deferred; cupy guard unit-tested, e2e blocked by #3112 upstream)
- [x] NaN handling is correct (`np.isnan`, not `== np.nan`; float-kind gate)
- [x] Edge cases covered (clean chunk, NaN chunk, integer chunk, cupy chunk, parity round trip)
- [x] Dask chunk boundaries handled correctly (elementwise map_blocks, no overlap needed)
- [x] No premature materialization or unnecessary copies (that is the point of the PR)
- [x] Benchmark not needed (regression pinned by execution-count test instead)
- [x] Docstrings present and accurate (error-timing change documented on both `to_geotiff` and `_pack`)

brendancol (commented):
## Follow-up review (after 5798562)

The suggestion from the first pass is addressed: `test_pack_dask_nan_no_sentinel_raises_through_vrt_write` now covers the `.vrt` destination, asserting the deferred ValueError comes back verbatim through `_write_vrt_tiled`'s collect-and-reraise path and that neither the `.vrt` file nor the `<stem>_tiles` dir is left behind (the assertion names match the writer's actual `stem + '_tiles'` construction at `_writers/eager.py:1264`). All 10 tests in the file pass.

The two nits stand as dismissed for the reasons given in the first review: the non-float mapping case is unreachable and already handled at chunk level, and the `isnan(...).any()` form matches the module's prevailing style.

PR #3238, opened 2026-06-11
hotspots() returned all zeros instead of raising when the raster contains Inf

### LINKED ISSUE #3219: hotspots() silently returns all zeros when the raster contains Inf

brendancol:
Closes #3219

With an Inf cell in the raster, `np.nanstd` / `cupy.nanstd` comes out NaN. `_gistar_global_stats` only guarded `global_std == 0`, and `NaN == 0` is False, so validation passed. Every Gi* z-score was then NaN, and `_gistar_zscore` maps non-finite z-scores to 0, so every cell got classified as "no significance".

- `_gistar_global_stats` now raises `ValueError` when `global_std` is not finite, with a message pointing at Inf values
- numpy and cupy raise eagerly; dask+numpy and dask+cupy pick the check up through `_gistar_validate_lazy` and raise at compute time, same as the degenerate inputs from #2843
- new `pos_inf` / `neg_inf` cases in the degenerate-input test matrix, which runs on all four backends

Test plan:
- [x] `pytest xrspatial/tests/test_focal.py -k "degenerate or hotspots"`: 64 passed, 0 skipped (GPU cases ran)
- [x] the repro from the issue raises `ValueError` instead of returning all zeros

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -1466,8 +1466,9 @@ def _gistar_global_stats(global_mean, global_std, n):
     if not np.isfinite(global_std):
         raise ValueError(
             "Standard deviation of the input raster values is not finite. "
-            "The raster contains Inf values; mask them (e.g. replace with "
-            "NaN) before calling hotspots()."
+            "The raster likely contains Inf values, or values too large "
+            "for float32; mask them (e.g. replace with NaN) or rescale "
+            "before calling hotspots()."
         )
     if global_std == 0:
         raise ZeroDivisionError(
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -1456,12 +1456,18 @@ def _gistar_global_stats(global_mean, global_std, n):
     ``global_mean`` is the mean of the valid (non-NaN) raster cells,
     ``global_std`` their population standard deviation, and ``n`` the
     count of valid cells. Gi* needs ``n > 1`` (the variance term divides
-    by ``n - 1``) and a non-zero spread.
+    by ``n - 1``) and a non-zero, finite spread.
     """
     if n < 2:
         raise ValueError(
             "hotspots() needs at least 2 valid (non-NaN) cells to "
             "compute the Getis-Ord Gi* statistic."
+        )
+    if not np.isfinite(global_std):
+        raise ValueError(
+            "Standard deviation of the input raster values is not finite. "
+            "The raster contains Inf values; mask them (e.g. replace with "
+            "NaN) before calling hotspots()."
         )
     if global_std == 0:
         raise ZeroDivisionError(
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: hotspots() returned all zeros instead of raising when the raster contains Inf

- None.

- [ ] `xrspatial/focal.py:1466` - the new message states "The raster contains Inf values" unconditionally. `np.isfinite` also fires when the std itself overflows to inf, e.g. a float32 raster full of finite values around 1e20 (verified locally: `np.nanstd` returns inf, no Inf in the input). Consider hedging the wording, e.g. "the raster likely contains Inf values, or values too large for float32", so the error doesn't assert something false in the overflow case.

- [ ] `xrspatial/tests/test_focal.py:1358-1362` - `pos_inf` and `neg_inf` are identical except for the sign of one cell. A short comment noting why both signs are pinned (both poison nanstd the same way) would make the duplication obviously intentional. Take it or leave it.

### What looks good
- One check in `_gistar_global_stats` covers all four backends: numpy/cupy validate eagerly, dask paths inherit it through `_gistar_validate_lazy` at compute time. No per-backend duplication.
- `np.isfinite` is the right predicate, not `np.isnan`: it also catches std overflow to inf.
- The check is ordered after the `n < 2` guard, so all-NaN rasters keep their existing "needs at least 2 valid" error (std is NaN there too).
- Tests reuse the #2843 degenerate-case matrix, so the Inf cases run on numpy, cupy, dask+numpy, and dask+cupy with both eager and compute-time raise semantics.

### Checklist
- [x] Algorithm matches reference (validation only; Gi* math untouched)
- [x] All implemented backends produce consistent results (same error, eager vs compute-time)
- [x] NaN handling is correct (NaN cells still excluded; Inf now rejected)
- [x] Edge cases are covered by tests (pos/neg Inf added to existing constant/all-NaN/single-valid matrix)
- [x] Dask chunk boundaries handled correctly (no convolution changes)
- [x] No premature materialization (dask globals stay lazy 0-d arrays)
- [x] Benchmark exists or is not needed (error path, no benchmark needed)
- [x] README feature matrix updated (n/a, no new function or backend change)
- [x] Docstrings present and accurate (helper docstrings updated)

brendancol (commented):
## Follow-up review (after 04164e7f)

- `xrspatial/focal.py:1467` - error message now says the raster "likely contains Inf values, or values too large for float32", so it no longer asserts something false in the float32 variance-overflow case. The first sentence (the part the tests match) is unchanged.
- `xrspatial/tests/test_focal.py:1358` - the pos_inf/neg_inf duplication is now explained by a comment.

No new findings. Degenerate-case suite still passes on all four backends (64 passed, 0 skipped).

PR #3239, opened 2026-06-11
Keep float32 sources float32 through to_geotiff(pack=True) (#3080)

### LINKED ISSUE #3080: to_geotiff(pack=True) widens a float32 source to float64
Spun off from the review on #3079.

When `pack=True` packs a float32 raster that was read with `unpack=True`, the output comes back float64 rather than float32. `_pack` (`xrspatial/geotiff/_attrs.py:1699`) infers the target dtype from the nodata scalar (`np.asarray(nodata).dtype.name`) when `mask_and_scale_dtype` is absent, and a float nodata resolves to float64. Float sources don't record `mask_and_scale_dtype` on read, so they always hit that path.

The nodata sentinel value is written correctly, so this is a dtype-width fidelity gap, not a wrong-value bug. Fixing it cleanly probably means recording the source dtype on an `unpack=True` read for float rasters too (the same way integer sources already get `mask_and_scale_dtype`), then having `_pack` honor it. That is a read-side change, which is why it is split out from the nodata fix in #3079.

brendancol:
Closes #3080.

- `unpack=True` reads of float sources now record `attrs['mask_and_scale_dtype']` when there is real unpack state to reverse (a declared nodata sentinel or a non-identity scale/offset). Both the eager and dask read paths record it; the cupy and dask+cupy backends share those code paths (`_finalize_eager_read` / `_read_geotiff_dask`), so the attr lands on all four backends.
- `_pack` no longer infers the target dtype from a float-typed nodata scalar. That fallback was written for pre-v5 integer reads, where the sentinel is stored in the source dtype; it now only applies to integer sentinels. Float buffers keep their own width, which also fixes the same widening when a plain `masked=True` read is packed back.
- No new attrs key and no shape change, so the attrs contract stays at v5. The contract docs in `_attrs.py` are updated.

Backend coverage: numpy and dask+numpy are tested directly. cupy and dask+cupy get the read-side attr through the shared paths, but their pack round-trips are still pinned on #3112; #3114 is adding that coverage separately.

Test plan:
- [x] New tests in `xrspatial/geotiff/tests/write/test_pack_float_width_3080.py`: float32 round-trip with and without scale/offset (numpy + dask), eager/dask attr parity, masked-read pack width, and a guard that a no-op unpack read still raises "no unpack state". 7 of the 9 fail without the fix.
- [x] Full `xrspatial/geotiff/tests` suite: 6385 passed, 68 skipped, 3 xfailed.

Heads up: #3114 (GPU/dask+GPU pack round-trip test coverage) is in flight on a separate branch from main. Both PRs touch the pack area, so whichever merges second may need a small reconcile. The new tests here live in their own file to keep the overlap small.

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -571,7 +571,10 @@ _TIFF_SHORT = 3
 # Version 5 adds ``attrs['mask_and_scale_dtype']``: the integer source
 # dtype recorded when ``mask_nodata`` / ``mask_and_scale`` promoted the
 # array to float on read. ``to_geotiff(pack=True)`` reads it to restore
-# the on-disk dtype. Existing keys keep their pre-v5 shape.
+# the on-disk dtype. Existing keys keep their pre-v5 shape. #3080 later
+# extended emission (same key, same shape, still v5) to float sources on
+# ``mask_and_scale`` reads that carry a sentinel or a non-identity
+# scale / offset, so ``pack`` keeps float32 sources float32.
 _ATTRS_CONTRACT_VERSION = 5
 
 
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -77,10 +77,13 @@ Canonical (xrspatial owns these; round-trip stable):
   happened so consumers can tell float-because-masked from
   float-because-promoted.
 - ``mask_and_scale_dtype`` (#3064, contract v5): string dtype name of
-  the integer source (e.g. ``"int8"``), only emitted when masking
-  and / or ``mask_and_scale`` promoted an integer array to float on
-  read. ``to_geotiff(pack=True)`` reads it to reverse the promotion and
-  restore the on-disk dtype.
+  the source (e.g. ``"int8"``). Emitted when masking and / or
+  ``mask_and_scale`` promoted an integer array to float on read, and
+  on ``mask_and_scale`` reads of float sources that carry a declared
+  sentinel or a non-identity scale / offset (#3080) -- float sources
+  are never promoted, but recording their width keeps
+  ``to_geotiff(pack=True)`` from widening float32 to float64.
+  ``pack`` reads it to restore the on-disk dtype.
 - ``raster_type``: ``'area'`` (implicit / RasterPixelIsArea) or ``'point'``
   (explicit / RasterPixelIsPoint).
 - ``georef_status``: one of ``'full'``, ``'transform_only'``, ``'crs_only'``,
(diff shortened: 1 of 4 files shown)


brendancol (commented):
## PR Review: Keep float32 sources float32 through to_geotiff(pack=True) (#3080)

None found.

- [ ] `xrspatial/geotiff/_writers/eager.py:365-377` -- the public `to_geotiff` docstring for `pack` still says "cast to the integer source dtype recorded on `attrs['mask_and_scale_dtype']`" and "The dtype falls back to the `attrs['nodata']` width for arrays read before contract v5." Both are stale now: float sources also get their dtype restored, and the fallback only applies to integer-typed sentinels. Update to match `_pack`'s revised docstring.
- [ ] `xrspatial/geotiff/_attrs.py:1796-1806` -- the integer-sentinel branch of the pre-v5 fallback has no test coverage (nothing pops `mask_and_scale_dtype` and packs with an integer-typed sentinel). This PR narrowed the gate on that branch, so a silent regression there would go unnoticed. Add a test: unpack-read an int source, `attrs.pop('mask_and_scale_dtype')`, set `attrs['nodata'] = np.int8(...)`, pack, assert the integer dtype comes back.
- [ ] `xrspatial/geotiff/tests/write/test_pack_float_width_3080.py` -- the PR newly records `mask_and_scale_dtype='float64'` on unpack reads of float64 sources, but no test pins that case. A float64 round-trip (or at least an attr assertion) would cover it.

- [ ] `xrspatial/geotiff/_attrs.py:571` -- the contract history comment ("Version 5 adds ... the integer source dtype") still describes the v5-era emission rule. Emission changed under the same version number, so a one-line "#3080 extends emission to float sources" note there would save the next reader some confusion.

### What looks good
- The fix lands in the two shared read paths (`_finalize_eager_read`, `_read_geotiff_dask`), so cupy and dask+cupy pick it up with no backend-specific code.
- New tests fail without the fix (7 of 9), use issue-numbered tmp file names, and cover numpy + dask, scale/offset and no-scale variants, the masked-read pack path, and a guard that the "no unpack state" error still fires for no-op float unpack reads.
- Restricting the `_pack` fallback to integer sentinels is right: float sentinels are stored as Python floats and carry no width signal, while the float buffer itself does.
- Tests live in a new file, which keeps merge overlap with the in-flight #3114 work small.

### Checklist
- [x] Algorithm matches reference (inverse of unpack read; checked against the read-side promotion rules)
- [x] All implemented backends produce consistent results (eager/dask attr parity test; GPU paths shared, GPU pack round-trip pinned on #3112)
- [x] NaN handling is correct (sentinel fillna unchanged; no-sentinel integer guard untouched)
- [ ] Edge cases are covered by tests (float64 source and pre-v5 integer fallback are not -- see suggestions)
- [x] Dask chunk boundaries handled correctly (attr-only change on the dask path; per-chunk masking untouched)
- [x] Benchmark exists or is not needed (metadata/dtype fix, no benchmark warranted)
- [x] README feature matrix updated (not applicable -- no new function, no backend support change)
- [x] Docstrings present and accurate (except the `to_geotiff` docstring noted above)

brendancol (commented):
## Follow-up review: c4c46ac9

- `to_geotiff` pack docstring (`_writers/eager.py`): fixed. It now says "source dtype", documents that float32 stays float32, and describes the narrowed fallback (integer sentinel width, buffer dtype for float sentinels).
- Pre-v5 integer fallback coverage: fixed. `test_pack_pre_v5_integer_sentinel_fallback` pops the attr, sets an `np.int8` sentinel, and asserts int8 comes back.
- float64 source coverage: fixed. `test_pack_keeps_float64_width` pins the attr value and the round-trip dtype on numpy and dask.
- Contract v5 history comment (`_attrs.py:571`): fixed, with a note that #3080 extended emission under the same version.

One observation that needs no action: the float64 test writes through the `_write_f32_tiff` helper, so the helper name is now slightly off. `test_pack_3064.py` has the same quirk (`_write_int_tiff` writes float32 data in its float test), so this matches the existing convention.

No new blockers, suggestions, or nits. The test file is at 12 tests, all passing locally; the write and attrs groups pass too (1161 passed, 1 skipped, 2 xfailed).

PR #3240, opened 2026-06-11
Fix to_geotiff(pack=True) crash on cupy and dask+cupy input

### LINKED ISSUE #3112: to_geotiff(pack=True) crashes on cupy and dask+cupy input

brendancol:
Closes #3112

- `_pack` used `DataArray.fillna` to put the nodata sentinel back, and that routes through xarray's `duck_array_ops.where`, which breaks on cupy-backed arrays. Eager GPU reads hit `AttributeError: module 'cupy' has no attribute 'astype'` (xarray 2025.12 + cupy 13.6); dask+GPU graphs raise `TypeError` at compute time when the host-side 0-d fill array reaches `cupy.where`.
- The fill is now a buffer-level helper, `_pack_fill_nan`: `np.isnan` dispatches to cupy via `__array_ufunc__` and the masked assignment takes a plain Python scalar. Dask backings map it per chunk with the same `map_blocks` shape as the existing no-sentinel guard, so the fill stays lazy.
- Removed the strict xfail from `test_pack_round_trip_gpu` (both legs pass now), added unit tests for the helper on numpy and cupy chunks, and fixed the stale "GPU pack crashes" notes in the release-contract docs and the `SUPPORTED_FEATURES` comments.

Backend coverage: numpy and dask+numpy keep the same fill semantics (full write suite passes); cupy and dask+cupy round-trip instead of crashing.

Test plan:
- [x] `pytest xrspatial/geotiff/tests/write/test_pack_3064.py`: 40 passed on a CUDA host with cupy 13.6, GPU legs un-xfailed
- [x] `pytest xrspatial/geotiff/tests/write/`: 1127 passed
- [x] `pytest xrspatial/geotiff/tests/test_round_trip.py xrspatial/geotiff/tests/read/`: 1308 passed
- [x] Issue repro script: eager GPU and dask+GPU writes both succeed

### DIFF
--- a/docs/source/reference/geotiff_release_contract.md
+++ b/docs/source/reference/geotiff_release_contract.md
@@ -93,7 +93,7 @@ category. The `Key` column matches the runtime key.
 | `reader.allow_rotated` | experimental | Opt-in `allow_rotated=True`; drops the axis-aligned `transform` attr in favour of `rotated_affine`. |
 | `reader.allow_unparseable_crs` | experimental | Opt-in escape hatch for CRS strings pyproj cannot parse. |
 | `reader.gpu` | experimental | GPU read path; no cross-backend numerical parity claim. |
-| `reader.unpack` | experimental | `unpack=True` scale/offset decode on `open_geotiff` (CF-packed integers to float, nodata sentinel to NaN). Experimental because the GPU path crashes (#3112) and the GPU / dask+GPU branches lack tests (#3114). |
+| `reader.unpack` | experimental | `unpack=True` scale/offset decode on `open_geotiff` (CF-packed integers to float, nodata sentinel to NaN). Experimental because the GPU / dask+GPU branches gained round-trip coverage only recently (#3112 pack crash fix, #3114); no behavioural promise yet. |
 
 ### Writers
 
--- a/docs/source/reference/release_gate_geotiff.rst
+++ b/docs/source/reference/release_gate_geotiff.rst
@@ -470,8 +470,9 @@ attrs contract
      - ``unpack=True`` applies GDAL ``SCALE`` / ``OFFSET`` and masks the
        nodata sentinel to NaN, recording ``scale_factor`` /
        ``add_offset`` / ``mask_and_scale_dtype`` on the attrs. CPU eager
-       and dask paths are tested; the GPU branches are not (#3112,
-       #3114), which is why the row sits at ``experimental``.
+       and dask paths are tested; the GPU branches gained round-trip
+       coverage only recently (#3112, #3114), which is why the row sits
+       at ``experimental``.
      - ``xrspatial/geotiff/tests/read/test_rioxarray_compat_2961.py``,
        ``xrspatial/geotiff/tests/read/test_mask_and_scale_dtype_parity_3066.py``
      - `#3163`_
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -333,9 +333,9 @@ SUPPORTED_FEATURES = {
     'reader.gpu': 'experimental',
     # ``unpack=True`` scale/offset decode on ``open_geotiff`` (CF-style
     # packed integers -> float, nodata sentinel -> NaN). Sits at
-    # ``experimental`` rather than ``advanced``: the GPU path crashes
-    # (#3112) and the GPU / dask+GPU branches have no test coverage
-    # (#3114), so no behavioural promise is made yet.
+    # ``experimental`` rather than ``advanced``: the GPU / dask+GPU
+    # branches gained round-trip coverage only recently (#3112 pack
+    # crash fix, #3114 coverage), so no behavioural promise is made yet.
     'reader.unpack': 'experimental',
     # Write paths.
     'writer.local_file': 'stable',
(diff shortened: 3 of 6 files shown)


brendancol (commented):
## PR Review: Fix to_geotiff(pack=True) crash on cupy and dask+cupy input (#3240)

None.

- [ ] `test_pack_restores_float_nodata_sentinel` (`xrspatial/geotiff/tests/write/test_pack_3064.py:95`) still runs only the numpy and dask legs. The changed fill is the path that restores a float sentinel, and the GPU round-trip test only covers the uint8 scale/offset case. Adding gpu / dask-gpu legs to the float-sentinel test would pin the float branch of the new fill on cupy too.

- [ ] Nothing pins the laziness of the new dask fill: `test_pack_lazy_nan_guard_3235.py` asserts the no-sentinel guard defers into the graph, but there's no equivalent for the sentinel fill path. `fillna` was lazy before and `map_blocks` is lazy now, so a test here would be pinning current behaviour, not fixing anything.
- [ ] The new `_pack_fill_nan` unit tests import the helper inside each test function (`test_pack_3064.py:197` etc.), while the sibling `test_pack_lazy_nan_guard_3235.py` imports `_pack_guard_no_nan` at module level. Harmless inconsistency.

### What looks good
- The fix is small and reuses the existing per-chunk pattern: `map_blocks` with `dtype=` / `meta=out.data._meta`, same shape in and out, matching `_pack_guard_no_nan`.
- `float(fill)` is the right device-agnostic scalar. It sidesteps both crash modes (xarray's `where` calling `cupy.astype`; a 0-d numpy fill array reaching `cupy.where`), and any float64 rounding it introduces matches what `fillna` into a float buffer already did.
- Input buffers are not mutated (`chunk.copy()` before the masked assignment), and the non-float early return mirrors `fillna`'s no-op on integer data.
- The strict xfail is removed in the same PR as the fix, so main never sees an XPASS break, and the stale "GPU pack crashes" claims in docs and `SUPPORTED_FEATURES` comments were corrected in the same change.

### Checklist
- [x] Algorithm matches reference/paper (n/a: fill semantics match the `fillna` it replaces)
- [x] All implemented backends produce consistent results (numpy/dask suites pass unchanged; gpu legs un-xfailed and passing on a CUDA host)
- [x] NaN handling is correct (`np.isnan` mask; a NaN sentinel degenerates to a no-op, same as before)
- [x] Edge cases are covered by tests (integer chunk no-op, input non-mutation, cupy chunk unit test)
- [x] Dask chunk boundaries handled correctly (elementwise `map_blocks`, no overlap needed)
- [x] No premature materialization or unnecessary copies (the copy in the helper replaces the copy `fillna` made; the fill stays lazy on dask)
- [x] Benchmark exists or is not needed (bugfix on an experimental-tier write path; not needed)
- [x] README feature matrix updated (n/a: GPU columns already marked experimental; tier unchanged)
- [x] Docstrings present and accurate (`_pack_fill_nan` documents the why, `_pack`'s comment updated)

brendancol (commented):
## Follow-up review (commit 399c4017)

- The float-sentinel test now runs gpu and dask-gpu legs (`test_pack_3064.py`, `chunks,gpu` parametrization), so the float branch of the fill is pinned on cupy. The new `_to_host` helper materialises dask- and cupy-backed buffers without `np.asarray`-on-cupy crashes.
- `test_pack_sentinel_fill_stays_lazy` pins the lazy shape of the sentinel fill with the same counting-block pattern as the #3235 guard tests, including the `meta=` detail that keeps dask's meta probe from inflating the counter.
- The `_pack_fill_nan` unit tests import the helper at module level, matching `test_pack_lazy_nan_guard_3235.py`.

No new issues in the delta. Pack suites pass locally: 43 passed across `test_pack_3064.py`, `test_pack_lazy_nan_guard_3235.py`, and `test_pack_float_width_3080.py` on a CUDA host.

PR #3243, opened 2026-06-11
Silence pyproj's lossy-PROJ-string warning in reproject param extractors (#3242)

### LINKED ISSUE #3242: Reprojection emits pyproj's lossy-PROJ-string UserWarning on every coregister/reproject of a projected raster
**Describe the bug**
Opening a projected raster with `open_geotiff(..., coregister=True)` (or calling `reproject`/`auto_reproject` directly) prints pyproj's UserWarning:

```
pyproj/crs/crs.py:1295: UserWarning: You will likely lose important projection information when converting to a PROJ string from another format. See: https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems
```

Repro:

```python
>>> study_area = open_geotiff("USGS_1_n35w112.tif", chunks=2048)
>>> myarr = study_area.xrs.open_geotiff("LF2024_EVC_CONUS.tif", unpack=True, coregister=True)
/home/.../pyproj/crs/crs.py:1295: UserWarning: You will likely lose important projection information ...
```

The parameter extractors in `xrspatial/reproject/_projections.py` read projection parameters via `crs.to_dict()`. pyproj implements `to_dict()` on top of `to_proj4()` and warns that the conversion is lossy. `_get_datum_params` already silences this warning and has a comment explaining why that's safe: it only reads the short datum/ellps names, never the lossy string itself. The other ten call sites call `to_dict()` raw: `_lcc_params`, `_aea_params`, `_cea_params`, `_sinu_params`, `_laea_params`, `_stere_params`, `_sterea_params`, `_omerc_params`, `_tmerc_params`, and `_is_wgs84_compatible_ellipsoid`. Any raster in one of those projections triggers the warning. The LANDFIRE repro above is CONUS Albers, so it hits `_aea_params`.

**Expected behavior**
No warning. What pyproj is warning about (datum and identifier metadata dropped from the PROJ string) isn't a problem for these extractors: they read numeric parameters and short names, and ellipsoid compatibility is checked separately. The suppression already written for `_get_datum_params` just needs to cover the other call sites too.

**Additional context**
Python 3.14, pyproj from conda-forge. Harmless, but it fires on the first reprojection of a session and makes it look like something is wrong with the CRS.

brendancol:
Closes #3242

- pyproj implements `CRS.to_dict()` on top of `to_proj4()` and warns that the conversion drops projection metadata. Ten parameter extractors in `xrspatial/reproject/_projections.py` called `to_dict()` raw, so reprojecting (or coregistering) any raster in one of those projections printed the warning.
- `_get_datum_params` already suppressed the warning with a comment on why that's safe (only short names and numeric parameters are read, never the lossy string). This PR moves that suppression into a `_crs_to_dict()` helper and uses it at all eleven call sites.
- No behavior change other than the silenced warning.

Backend coverage: the fix is in CRS parsing, which runs before backend dispatch, so it applies equally to numpy / cupy / dask+numpy / dask+cupy.

Test plan:
- [x] New `test_reproject_pyproj_warning_3242.py`: per-extractor no-warning checks across the 10 projection families, an end-to-end `reproject()` call with a `pyproj.CRS` source (the #3242 repro shape), and a check that `_crs_to_dict()` returns the same dict as raw `to_dict()`.
- [x] 12 of 13 new tests fail without the fix (the 13th covers `_get_datum_params`, which was already wrapped).
- [x] Full reproject suite: 500 passed, no lossy-PROJ warnings left.

### DIFF
--- a/xrspatial/reproject/_projections.py
+++ b/xrspatial/reproject/_projections.py
@@ -132,22 +132,30 @@ _DATUM_PARAMS = {
 }
 
 
+def _crs_to_dict(crs):
+    """``crs.to_dict()`` without pyproj's lossy-PROJ-string UserWarning.
+
+    crs.to_dict() goes through pyproj's to_proj4(), which warns that a
+    PROJ string drops detail. The param extractors here only read short
+    names and numeric projection parameters, never the lossy string, so
+    silence that one warning.
+    """
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            'ignore',
+            message='.*lose important projection information.*',
+            category=UserWarning,
+        )
+        return crs.to_dict()
+
+
 def _get_datum_params(crs):
     """Return (dx, dy, dz, rx, ry, rz, ds, a_src, f_src) for a non-WGS84 datum.
 
     Returns None for WGS84/NAD83/GRS80 (no shift needed).
     """
     try:
-        # crs.to_dict() goes through pyproj's to_proj4(), which warns that a
-        # PROJ string drops detail. We only read the datum/ellps short names
-        # here, never the lossy string, so silence that one warning.
-        with warnings.catch_warnings():
-            warnings.filterwarnings(
-                'ignore',
-                message='.*lose important projection information.*',
-                category=UserWarning,
-            )
-            d = crs.to_dict()
+        d = _crs_to_dict(crs)
     except Exception:
         return None
     datum = d.get('datum', '')
@@ -305,7 +313,7 @@ def _lcc_params(crs):
     Returns (lon0, lat0, n, c, rho0, k0) or None.
     """
     try:
-        d = crs.to_dict()
+        d = _crs_to_dict(crs)
     except Exception:
         return None
     if d.get('proj') != 'lcc':
@@ -450,7 +458,7 @@ def _aea_params(crs):
     Returns (lon0, n, c, dd, rho0, fe, fn) or None.
     """
     try:
-        d = crs.to_dict()
+        d = _crs_to_dict(crs)
     except Exception:
         return None
     if d.get('proj') != 'aea':
@@ -554,7 +562,7 @@ def _cea_params(crs):
     Returns (lon0, k0, fe, fn) or None.
     """
     try:
-        d = crs.to_dict()
+        d = _crs_to_dict(crs)
     except Exception:
         return None
     if d.get('proj') != 'cea':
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Silence pyproj's lossy-PROJ-string warning in reproject param extractors (#3242)

None.

- [ ] Add a dask-backed variant of the end-to-end test (`test_reproject_pyproj_warning_3242.py:75`). The extractors run per chunk inside `_try_numba_transform_locked`, which on the dask path executes inside worker tasks. A `raster.chunk(...)` + `.compute()` case would catch a warning leak that only shows up on that path.

- [ ] `_crs_to_dict` (`xrspatial/reproject/_projections.py:135`) uses `warnings.catch_warnings()`, which swaps process-global filter state. Every current caller runs under `_PARALLEL_KERNEL_LOCK`, so calls from dask worker threads are serialized, but that invariant is easy to lose in a future refactor. A sentence in the docstring noting that callers are expected to hold the lock would protect it. This pattern predates the PR (it comes from `_get_datum_params`), so this is informational.

### What looks good
- The change is minimal. The suppression that `_get_datum_params` already used, comment and all, becomes a shared helper, and all eleven `to_dict()` sites go through it. The try/except contract at each call site is unchanged.
- `_projections_cuda.py` imports its projection parameters from `_projections` (line 733), so the GPU fast path gets the fix too.
- The regression tests were verified to fail without the fix (12 of 13; the 13th covers the already-wrapped `_get_datum_params`). They pin every extractor family plus an end-to-end `reproject()` call with a `pyproj.CRS` source, the exact shape of the #3242 repro. Sinusoidal has no EPSG code and is built from a PROJ string instead.
- `test_crs_to_dict_matches_raw_to_dict` pins that the helper changes nothing except the warning.

### Checklist
- Algorithm matches reference: n/a, no numeric change
- All implemented backends consistent: yes; CRS parsing is host-side and the CUDA path shares the same extractors
- NaN handling: n/a
- Edge cases covered by tests: one test per projection family, plus the non-EPSG sinusoidal case
- Dask chunk boundaries: n/a
- No premature materialization or copies: n/a
- Benchmark: not needed for a warning fix
- README feature matrix: no new functions, no backend change
- Docstrings: the helper documents what it silences and why that is safe

brendancol (commented):
## Follow-up review (after 5acb1978)

- The dask-backed end-to-end test is in (`test_reproject_dask_pyproj_crs_no_lossy_proj_warning`). It chunks the EPSG:5070 raster 16x16 and computes inside the warning-recording block, so the per-chunk extractor calls on worker tasks are exercised. Guarded with `pytest.importorskip('dask.array')`.
- `_crs_to_dict` now documents that `catch_warnings` swaps process-global state and that callers are expected to hold `_PARALLEL_KERNEL_LOCK`.

No new issues in the delta. Full regression file passes locally (14 tests). Nothing blocking from my side.

PR #3244, opened 2026-06-11
Follow-up to #3243: dask warning regression test, lock note on _crs_to_dict (#3242)

brendancol:
Follow-up to #3243, which was merged while its review fixes were still in flight; the squash took the first commit only. This lands the two review items:

- A dask-backed variant of the end-to-end regression test. The param extractors also run per chunk inside worker tasks, so a warning leak there would not show up in the numpy-path test. Guarded with `pytest.importorskip('dask.array')`.
- A note on `_crs_to_dict` that `catch_warnings` swaps process-global filter state and callers are expected to hold `_PARALLEL_KERNEL_LOCK`.

Refs #3242.

Test plan:
- [x] `test_reproject_pyproj_warning_3242.py`: 14 passed locally on top of current main.

### DIFF
--- a/xrspatial/reproject/_projections.py
+++ b/xrspatial/reproject/_projections.py
@@ -139,6 +139,10 @@ def _crs_to_dict(crs):
     PROJ string drops detail. The param extractors here only read short
     names and numeric projection parameters, never the lossy string, so
     silence that one warning.
+
+    ``catch_warnings`` swaps process-global filter state and is not
+    thread-safe; callers are expected to hold ``_PARALLEL_KERNEL_LOCK``
+    (all current callers run inside the ``*_locked`` entry points).
     """
     with warnings.catch_warnings():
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: follow-up to #3243

This is the review-fix delta from #3243 (the squash merge there took only the first commit). Reviewed as part of the #3243 follow-up pass:

### Blockers
None.

### Suggestions / Nits
None. The dask test computes inside the warning-recording block so the per-chunk extractor calls on worker tasks are exercised, and the `pytest.importorskip('dask.array')` guard keeps it safe on minimal environments. The `_crs_to_dict` docstring note records the `_PARALLEL_KERNEL_LOCK` invariant that makes the process-global `catch_warnings` use safe.

### What looks good
Cherry-picked cleanly onto current main; the full regression file (14 tests) passes locally.

PR #3245, opened 2026-06-11
Replace the numpy proximity line-sweep with an exact cKDTree query

### LINKED ISSUE #3121: EUCLIDEAN proximity line-sweep on numpy misses the true nearest target for some pixels
**Describe the bug**

The numpy backend of `proximity()` uses the GDAL-ported line-sweep for EUCLIDEAN (the bounded dask+numpy path runs the same kernel per chunk). On some target layouts the sweep's neighbour propagation never delivers the true nearest target to a pixel, so it reports a distance that is too large. The cupy backend, which does an exact brute-force search, matches a float64 ground-truth check; the numpy result does not.

**Repro**

```python

data = np.zeros((8, 10))
for i, j in [(0, 4), (0, 8), (1, 2), (2, 1), (6, 6)]:
    data[i, j] = 1.0

r['y'] = np.linspace(0, 10, 101)[12:20]   # pitch 0.1
r['x'] = np.linspace(0, 10, 103)[28:38]   # pitch ~0.098

out = proximity(r)
print(out.data[3, 4])   # 0.3
# true nearest: target at (1, 2), distance 0.28008347
```

Pixel (3, 4) is 0.28008 away from the target at (1, 2), but the line-sweep reports 0.3, the distance to the target at (0, 4). About 0.2 pixel off, 7% relative error.

On a 101x103 random raster (seed 0, ~600 targets), 4 pixels differ from ground truth by more than 1e-4 (max error 0.0199). It happens with square pixels as well (2 pixels, max error 0.0252), and the bounded dask+numpy path reproduces the same 4 wrong pixels. MANHATTAN came back clean on the same input, but it runs the same sweep, so I would not assume it is immune.

ALLOCATION and DIRECTION are unaffected; they already route through the brute-force kernel (#2881). GREAT_CIRCLE was moved off the line-sweep in #2816. So the only modes still using the sweep are PROXIMITY with EUCLIDEAN or MANHATTAN.

This looks like an inherent limitation of the 4-pass nearest-point propagation (Danielsson-style sweeps have documented worst-case errors of a fraction of a pixel) carried over from GDAL, not a porting mistake.

**Expected behavior**

`proximity()` returns the exact nearest-target distance, and numpy, cupy, dask+numpy and dask+cupy agree. cupy is already exact.

**Additional context**

Found by the security sweep's CPU/GPU parity probe against the proximity module. One possible fix: route numpy PROXIMITY through cKDTree when scipy is available, mirroring the unbounded dask path (`distance_upper_bound` covers max_distance), with brute force as the fallback when scipy is missing. The dask chunk function should follow whatever the numpy backend does so the backends stay in sync.

brendancol:
Closes #3121

The numpy backend's GDAL-ported 4-pass line-sweep can report a distance that is too large for some pixels; the issue's repro is off by 7%. PROXIMITY with EUCLIDEAN or MANHATTAN now queries a cKDTree instead (p=2/p=1). The distance_upper_bound is widened by one ulp because scipy treats the bound as exclusive while the brute-force and CUDA kernels keep dist <= max_distance. When scipy is missing, the fallback is the brute-force kernel, which is also exact, rather than the sweep.

The bounded dask+numpy path runs the numpy chunk function inside map_overlap, so it gets the fix for free. NaN-padded halo pixels are dropped as targets (on integer rasters the NaN pad casts to a garbage finite value that would otherwise pass the target test) and skipped as query points, since cKDTree rejects non-finite coordinates.

The line-sweep kernels are now dead and removed, along with the #3103 recompile-guard test that pinned them.

Backend coverage: numpy and dask+numpy are fixed. cupy and dask+cupy were already exact and are untouched, but the new tests run all four backends for parity.

Test plan:
- [x] New regression tests: the issue's repro raster on all 4 backends (bounded and unbounded), a 101x103 random raster against a float64 ground truth (EUCLIDEAN and MANHATTAN, numpy and dask+numpy), and a no-scipy fallback check
- [x] Full xrspatial/tests/test_proximity.py passes (516 tests, CUDA host)
- [x] test_validation.py, test_accessor.py, test_surface_distance.py pass

### DIFF
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -734,8 +734,14 @@ def _process_dask_cupy(raster, x_coords, y_coords, target_values,
     )
 
 
-def _process_numpy_kdtree(img, xs, ys, target_values, max_distance, p):
+def _process_numpy_kdtree(img, xs, ys, target_values, max_distance, p,
+                          workers=1):
     """Exact nearest-target PROXIMITY on the CPU via scipy's cKDTree.
+
+    ``workers`` is forwarded to ``cKDTree.query``. The eager numpy backend
+    passes -1 (all cores) because it runs a single query over the whole
+    raster; the dask chunk path keeps the default 1 because concurrent
+    chunks already fill the cores and -1 would oversubscribe them.
 
     Replaces the GDAL-ported 4-pass line-sweep for EUCLIDEAN/MANHATTAN: the
     sweep propagated one nearest-target candidate between adjacent pixels,
@@ -769,7 +775,8 @@ def _process_numpy_kdtree(img, xs, ys, target_values, max_distance, p):
     upper = max_distance
     if np.isfinite(upper):
         upper = np.nextafter(upper, np.inf)
-    dists, _ = tree.query(query_pts, p=p, distance_upper_bound=upper)
+    dists, _ = tree.query(query_pts, p=p, distance_upper_bound=upper,
+                          workers=workers)
 
     dists = dists.astype(np.float32)
     dists[~np.isfinite(dists)] = np.nan
@@ -1333,7 +1340,7 @@ def _process(
         x_coords[0], x_coords[-1], y_coords[0], y_coords[-1], distance_metric
     )
 
-    def _process_numpy(img, x_coords, y_coords):
+    def _process_numpy(img, x_coords, y_coords, workers=1):
         # GREAT_CIRCLE is not a Minkowski metric, so the cKDTree query below
         # cannot answer it. Use an exact brute-force search instead (matches
         # the GPU kernel). The branch lives in plain Python rather than
@@ -1360,13 +1367,20 @@ def _process(
         # nearest target by a fraction of a pixel (issue #3121). Without
         # scipy, fall back to the brute-force kernel, which is also exact.
         if cKDTree is None:
+            warnings.warn(
+                "proximity: scipy is not installed; falling back to an "
+                "exact brute-force search, which is slower on large "
+                "rasters.",
+                UserWarning,
+                stacklevel=2,
+            )
             return _process_numpy_bruteforce(
                 img, x_coords, y_coords, target_values,
                 np.float32(max_distance), distance_metric, process_mode,
             )
         return _process_numpy_kdtree(
             img, x_coords, y_coords, target_values, max_distance,
-            2 if distance_metric == EUCLIDEAN else 1,
+            2 if distance_metric == EUCLIDEAN else 1, workers=workers,
         )
 
     def _process_dask(raster, xs, ys):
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Replace the numpy proximity line-sweep with an exact cKDTree query

None found.

- [ ] `xrspatial/proximity.py:772`: `tree.query(...)` runs single-threaded. On the eager numpy path it is one query over the whole raster, and scipy's `query` accepts `workers`, so `workers=-1` would parallelize the dominant cost on large rasters. The catch is that this function doubles as the dask chunk function, where concurrent chunks already keep the cores busy and `workers=-1` would oversubscribe. Threading a `workers` argument through and passing `-1` only from the eager numpy call gets the win without hurting dask.

- [ ] `xrspatial/proximity.py:1373` (dispatch in `_process_numpy`): without scipy, PROXIMITY now runs the O(N*T) brute-force kernel where the O(N) sweep used to run. Correct, but slower, and a scipy-less user has no way to know why. A one-time `warnings.warn` like the existing tiled-KDTree fallback warning would tell them.

### What looks good
- The kdtree path matches the brute-force kernel on random rasters for both metrics, bounded and unbounded, with explicit target_values and a descending axis (checked locally at rtol 1e-5).
- Widening `distance_upper_bound` by one ulp keeps the bound inclusive. A target at exactly `max_distance` still reports its distance instead of NaN, same as the brute-force and CUDA kernels.
- The chunk function handles both NaN problems: cKDTree refusing non-finite query points, and integer rasters whose NaN pad casts to a garbage finite value that passes the target test. The existing integer-raster tests cover that path and pass.
- The dead line-sweep kernels are gone rather than orphaned, along with the unused `math.sqrt` import and the #3103 recompile test that pinned the old kernel.
- The new tests compare against an independent float64 reference, not a snapshot of current output, and the repro test runs all four backends.

### Checklist
- [x] Algorithm matches reference (exact nearest-target; verified against brute force)
- [x] All implemented backends produce consistent results (new parity tests, 4 backends)
- [x] NaN handling is correct (halo pads, all-NaN raster, int-raster pad cast)
- [x] Edge cases are covered by tests (degenerate shapes, descending axes, exact-bound distance, no targets)
- [x] Dask chunk boundaries handled correctly (chunk fn keeps input shape; existing halo tests pass)
- [x] No premature materialization (no .compute() during graph construction)
- [x] Benchmark exists (benchmarks/benchmarks/proximity.py covers EUCLIDEAN and MANHATTAN proximity)
- [x] Docstrings updated (proximity/allocation/direction no longer describe the GDAL sweep)

brendancol (commented):
## Follow-up review (ea2e0fe5..0c2977ef)

- The eager numpy path passes workers=-1 to cKDTree.query and the dask chunk path keeps the single-worker default, which is the split the suggestion asked for. scipy has had the workers kwarg since 1.6, and this project's Python >= 3.12 floor implies scipy >= 1.11, so no compatibility worry.
- The no-scipy fallback emits a UserWarning in the same style as the tiled-KDTree warning, and the fallback test asserts it with pytest.warns.

test_proximity.py still passes in full (516 tests). Nothing new to flag.

PR #3247, opened 2026-06-11
Warn when coregister=True covers only a small fraction of the caller grid

### LINKED ISSUE #3246: Warn when coregister=True covers only a small fraction of the caller grid
## Reason or problem

`big.xrs.open_geotiff("small.tif", coregister=True)` resamples the file onto the caller's full grid. When the caller is a large (often dask-backed) template and the file only covers a small piece of it, the result is the full template size with NaN almost everywhere. Nothing tells the user this happened. They get a huge, mostly-empty array, and the better pattern (open the small raster first, crop the template to its bounds, then coregister) is easy to miss.

## Proposal

Emit a `UserWarning` from the accessor's coregister path when the file footprint covers less than 10% of the caller grid:

```
UserWarning: 'small.tif' covers only 0.4% of the caller grid; the coregistered
result will be mostly NaN. Consider opening the small raster first and cropping
the caller to its bounds before coregistering.
```

The coverage fraction can be computed from metadata already on hand in `_open_geotiff_windowed`: the caller bbox (already transformed into the file CRS when the two differ) versus the read window after clamping to the file bounds. No extra I/O.

`auto_reproject=True` is unaffected. Its output follows the file extent, so it is never sparse.

## Alternatives

A cropping API was considered (an `extent='intersection'` mode on the accessor, or a `crop_to()` helper) and set aside for now. A warning is enough to point users at the right pattern without growing the API.

brendancol:
Closes #3246

- `.xrs.open_geotiff(..., coregister=True)` now emits a `UserWarning` when the file footprint covers less than 10% of the caller grid, since the result will be mostly NaN. The message points at the better pattern: open the small raster first and crop the caller to its bounds.
- Coverage comes from metadata already computed in `_open_geotiff_windowed` (caller bbox in file CRS vs. the clamped read window), so there is no extra I/O. `auto_reproject` is untouched; its output follows the file extent and is never sparse.
- Docstrings for both accessor variants (DataArray and Dataset) mention the warning.

Backend coverage: the check runs before the read, on metadata only, so it applies identically to numpy, cupy, dask+numpy, and dask+cupy callers (coregister itself remains CPU-only, as before).

Test plan:
- [x] New: small file vs. large template warns (`covers only`), result still matches the template grid
- [x] New: full-coverage file does not warn
- [x] New: warning fires across a CRS mismatch (coverage computed after bbox transform)
- [x] `test_open_geotiff_coregister.py` 18 passed
- [x] `test_accessor.py` 64 passed

### DIFF
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -11,6 +11,8 @@ directly via tab-completion::
     nir.xrs.ndvi(red)
 """
 
+import warnings
+
 import xarray as xr
 
 
@@ -147,6 +149,12 @@ def _bbox_edge_samples(x_min, y_min, x_max, y_max, n_per_side=20):
 
 
 _RESAMPLING_CHOICES = ('auto', 'nearest', 'bilinear', 'cubic')
+
+# coregister=True paints the file onto the caller's full grid, NaN-filling
+# everything outside the file footprint. Below this coverage fraction the
+# result is mostly NaN, so warn and point at the crop-the-caller-first
+# pattern instead.
+_COREGISTER_COVERAGE_WARN_FRACTION = 0.10
 
 
 def _resolve_resampling(resampling, result):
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Warn when coregister=True covers only a small fraction of the caller grid

None.

- [ ] `xrspatial/tests/test_open_geotiff_coregister.py:288` - the three new tests all use numpy-backed templates, but the scenario that motivated #3246 is a large dask-backed template. The check runs on metadata before the read, so it should behave the same, and I confirmed manually that it does (warning fires, output stays lazy with the caller's chunks). Still worth pinning with a small dask-template test so a future reordering of the check relative to the read does not silently lose the dask case.

- [ ] `xrspatial/accessor.py:986` and `:1523` - both docstrings hard-code "10%" while the threshold lives in `_COREGISTER_COVERAGE_WARN_FRACTION`. If the constant ever changes, the docs drift. Fine to leave, just noting the coupling.

### What looks good
- The coverage computation reuses values already in scope (`window`, the transformed bbox, the file transform), so there is no extra I/O and no behavior change to the read or reproject path.
- Zero-overlap is deliberately excluded (`0 < coverage`), preserving the existing downstream error instead of warning before a failure.
- `stacklevel=3` is correct for both call sites; both accessor variants call `_open_geotiff_windowed` directly.
- The no-warning test records warnings instead of using `simplefilter("error")`, so unrelated pyproj warnings cannot break it.
- `auto_reproject` is correctly left out: its output follows the file extent and is never sparse.

### Checklist
- [x] Algorithm matches reference/paper (n/a, simple area ratio)
- [x] All implemented backends produce consistent results (metadata-only check, verified manually on dask)
- [x] Edge cases are covered by tests (small fraction, full coverage, CRS mismatch; dask case suggested above)
- [x] Dask chunk boundaries handled correctly (check runs before the read; chunk forwarding untouched)
- [x] Benchmark exists or is not needed (not needed, no hot-path change)
- [x] README feature matrix updated (n/a, no new function or backend change)
- [x] Docstrings present and accurate (both accessor variants updated)

brendancol (commented):
## Follow-up review after da9ef0f5

Disposition of the earlier findings:

- Suggestion (dask-template test): fixed in da9ef0f5. `test_coregister_warns_dask_template` pins the motivating case: the warning fires for a chunked template and the result keeps the caller's (50, 50) chunks. Full file passes (19 tests).
- Nit (docstrings hard-code "10%"): dismissed. Keeping the docstring in sync automatically would mean f-string docstrings or pointing public docs at a private constant, both worse than the drift risk. The constant's own comment documents the coupling.

No new findings in the follow-up diff. The added test reuses the existing `_file_4326` helper and the `.chunk()` pattern from `test_coregister_dask_template_keeps_caller_chunks`, with a #3246-suffixed temp filename consistent with the other new tests.

Issue #3249, opened 2026-06-12
Overhaul the getting-started docs: environment setup, optional deps, runnable quickstart


brendancol:
## Reason or Problem

docs/source/getting_started/ is about 70 lines across four pages and parts of it are out of date:

- usage.rst still says xarray-spatial "currently depends on Datashader, but will soon be updated to depend only on xarray and numba". Core install deps have been numpy, numba, scipy, xarray, urllib3, zstandard for a while now; datashader only appears in the `examples` extra.
- installation.rst documents 2 of the 10 extras (`plot`, `vector`). Nothing on `geotiff`, `dask`, `gpu`, `reproject`, `optional`, the libnvcomp/kvikio runtime caveat, or cloud-storage deps.
- No environment setup guidance and no mention that Python >= 3.12 is required.
- No quickstart that runs end to end, and no mention of the `xrspatial examples` / `fetch-data` download commands.
- No links to the user guide caveats or the stability policy, so a new user can miss the WGS84 / float32 / NaN-nodata assumptions and silently get wrong results.
- Sphinx only builds on Read the Docs; nothing checks that docs build on PRs.

## Proposal

Restructure getting_started/ into four pages (why_xarray_spatial, installation, quickstart, examples_and_data) in six small PRs:

1. Fix the stale Datashader claim and the hardcoded GitHub example links (usage.rst, installation.rst).
2. Add a docs-build CI job that runs sphinx-build on PRs touching docs/.
3. Rewrite installation.rst: environment setup (venv / conda / uv), a table covering every extra in setup.cfg, GPU setup notes (cupy/CUDA, libnvcomp + kvikio), cloud storage deps, and an install verification snippet.
4. Add quickstart.rst, replacing usage.rst: a generate_terrain-based first example that needs no data download, the .xrs accessor, a GeoTIFF read/write round trip, Dataset support, short dask and GPU pointers, and a link to the caveats page.
5. Add why_xarray_spatial.rst, absorbing raster_huh.rst and the GDAL note from usage.rst.
6. Add examples_and_data.rst covering the xrspatial CLI download commands, plus a landing-page rewrite for getting_started/index.rst.

Each PR should merge on its own, and every command shown in the docs gets run during review.

## Impacts

[...]

PR #3252, opened 2026-06-12
Add a docs-build CI job

brendancol:
Refs #3249 (part 2 of 6).

Sphinx currently only runs on Read the Docs, so a PR that breaks the docs build isn't caught until after merge. This adds a `docs` workflow that runs on PRs touching `docs/**`, `README.md`, `setup.cfg`, or `.readthedocs.yml`:

- installs `.[doc,tests]` on Python 3.12, mirroring the RTD pre_build step so the two can't drift
- installs the pandoc binary via apt (nbsphinx needs it; the `pandoc` pip package doesn't ship it)
- `sphinx-build -b html -j auto`, gating on errors only (no `-W`; the tree has a handful of pre-existing notebook/reference warnings). The read phase is dominated by plot-directive examples in the reference docstrings (sample data load + numba compile per page); every extension in conf.py declares parallel_read_safe, so -j auto cuts the build from ~8 minutes to ~3.5.
- linkcheck runs on a weekly cron (Mondays 06:00 UTC) and workflow_dispatch instead of per PR. It cost ~10 minutes per PR, never gated (continue-on-error), and was in fact already exiting 1 with nobody noticing. The cron job drops continue-on-error so a broken link actually fails the run. Expect the first scheduled run to be red: there are real broken links to fix.

The job id is `docs-build`, distinct from the pytest workflow's `run` job, to avoid the check-name dedup problem that hides failures.

Test plan:
- [x] YAML parses
- [x] `docs-build` passes on this PR in 4m51s (was 18m48s)
- [x] `linkcheck` is skipped on pull_request events

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,42 @@
+name: docs
+on:
+  pull_request:
+    paths:
+      - 'docs/**'
+      - 'README.md'
+      - 'setup.cfg'
+      - '.readthedocs.yml'
+      - '.github/workflows/docs.yml'
+  workflow_dispatch:
+
+jobs:
+  # Job id/name kept distinct from the pytest workflow's `run` job:
+  # duplicate check names across workflows get deduplicated in the PR
+  # checks UI and can hide failures.
+  docs-build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Setup Python
+      uses: actions/setup-python@v5
+      with:
+        # Match the Read the Docs build (.readthedocs.yml).
+        python-version: '3.12'
+    - name: Install pandoc
+      # nbsphinx needs the pandoc binary; the `pandoc` pip package in
+      # the doc extra does not ship it.
+      run: sudo apt-get update && sudo apt-get install -y pandoc
+    - name: Install dependencies
+      # Same extras as the Read the Docs pre_build step, so this job
+      # and RTD cannot drift apart.
+      run: |
+        python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: Add a docs-build CI job

### Blockers
None.

### Suggestions
- .github/workflows/docs.yml: the job has no `timeout-minutes`. `continue-on-error: true` keeps a linkcheck failure from gating, but a hung external link check still occupies the runner until the default 6-hour job timeout. A cap around 30 minutes covers a full build (RTD finishes in well under that) and bounds the damage.

### Nits
None.

### What looks good
- Action versions match test.yml (checkout@v4, setup-python@v5), and the install step mirrors .readthedocs.yml's `pip install '.[doc,tests]'` exactly, so CI and RTD build from the same dependency set.
- The paths filter includes the workflow file itself, which is why this check ran on this very PR.
- Job id `docs-build` avoids the duplicate-check-name dedup problem already seen between the pytest workflows.
- apt pandoc install is needed: the `pandoc` pip package in the doc extra does not provide the binary.

### Checklist
- Correctness/backend/test items: not applicable (CI config only)
- Verified the workflow triggers and the YAML parses

brendancol (commented):
Follow-up to the review: added `timeout-minutes: 30` to the docs-build job so a stalled linkcheck can't hold a runner for the default 6-hour timeout. No other findings to address.

brendancol:
Measured the first two runs: ~26-27 minutes each (HTML build ~11, linkcheck ~14), which left under 3 minutes of headroom against the 30-minute cap. Raised it to 45.

PR #3254, opened 2026-06-12
coregister=True reads COG overviews and fails fast on bad templates

### LINKED ISSUE #3253: open_geotiff(coregister=True): use COG overviews and fix overview_level windowing
## Reason or Problem

`.xrs.open_geotiff(path, coregister=True)` always decodes the file at full resolution. The output grid is fixed by the caller's template, so when the template is coarser than the file, most of the decoded pixels get thrown away by the resample. `open_geotiff` already takes `overview_level=`, but the coregister path never uses it. Locally, a 4096px COG coregistered onto a 64px grid takes ~109 ms at full resolution and ~5 ms reading the right overview. For remote COGs the gap is wider because the read also fetches fewer bytes.

Passing `overview_level=N` through the accessor by hand doesn't work either: the pixel window is computed against the full-resolution transform, but the reader interprets `window=` in the selected overview's pixel space. The read lands in the wrong region, or raises:

```
>>> template.xrs.open_geotiff('cog.tif', coregister=True, overview_level=1)
ValueError: window=(76, 76, 180, 180) is outside the source extent (128x128) ...
```

A few more cases fail silently:

- A template crossing the antimeridian transforms into file-CRS longitudes near +180 and -180. min/max windowing then reads nearly the full file width with no indication.
- `coregister=True` with `allow_rotated=True`: a rotated file reads as an ungeoreferenced pixel grid, and reproject treats its row/col indices as map coordinates. Every pixel lands in the wrong place.
- Irregular, x-descending, or y-ascending templates: `_caller_grid` assumes a regular x-ascending / y-descending grid and quietly misaligns otherwise.
- A file fully disjoint from the template fails with the same low-level window error as above instead of saying the file doesn't overlap.
- `resampling='auto'` picks `nearest` for any integer raster without a colormap. Right for class rasters, wrong for int16 DEMs, and nothing tells you the choice happened.

## Proposal

**Design:**

- When the template grid is coarser than the file and the file has overviews, read the coarsest overview that still meets the template resolution. Probing uses header-only reads (`_read_geo_info(source, overview_level=N)` already returns the overview's dimensions with the pixel size rescaled). An explicit `overview_level=` wins; `overview_level=0` forces the full-resolution read.
- Compute the pixel window against the selected overview's transform and dimensions. This fixes the explicit `overview_level=` case for all accessor windowed reads, not just coregister.
- Raise clear errors for: no overlap with the template, `allow_rotated=True`, irregular or flipped template grids.
- Warn on antimeridian-crossing templates (the full-width window is a correct superset, just slow) and skip overview auto-selection there, since the wrapped bbox span would overstate the template's cell size.
- Warn when `'auto'` picks `'nearest'` for integer data with no colormap, and say how to override.

**Usage:** no API change. `coregister=True` gets faster on COGs, and `overview_level=` becomes usable with windowed accessor reads.

**Value:** cuts the main cost of getting a file onto an analysis grid, and turns silent misalignment into errors the analyst can act on.

## Stakeholders and Impacts

Anyone coregistering COGs onto coarse analysis grids. Touches `xrspatial/accessor.py` only; the geotiff reader and reproject are unchanged.

## Drawbacks

Overview pixels are resampled at write time, so values differ slightly from a full-resolution read (within ~0.5% of the data range in the bench above; `overview_level=0` restores exact behaviour). Templates with flipped axes that previously appeared to work now raise.

## Alternatives

An `average` resampling mode in reproject would handle decimation aliasing for plain TIFFs without overviews. That is a bigger change and can be a separate issue.

brendancol:
Closes #3253

- `coregister=True` now reads the coarsest overview that still meets the caller grid's resolution instead of always decoding full-resolution pixels. A 4096px COG onto a 64px grid went from 109ms to 5ms locally; remote COGs gain more because fewer bytes are fetched. `overview_level=0` forces full resolution, an explicit level wins over the auto pick, and plain TIFFs without overviews behave as before. Windows are now computed in the selected overview's pixel space, which also fixes explicit `overview_level=` on every accessor windowed read (previously it raised "window is outside the source extent" or read the wrong region).
- New fail-fast guards: a disjoint file raises "does not overlap" instead of the low-level window error; `allow_rotated=True` is refused (a rotated file reads as an ungeoreferenced grid that reproject would mis-place); irregular or flipped template grids raise with a `sortby` hint; an antimeridian-crossing template warns and falls back to the full-width window (correct superset, just slower) while skipping overview auto-selection.
- `resampling='auto'` warns when it picks `nearest` for integer data with no colormap (the int16 DEM case), with the override named.

Backend coverage: numpy and dask+numpy, same as coregister today (GPU paths still raise by design). The reader and reproject are untouched; everything lives in `xrspatial/accessor.py`.

Test plan:
- [x] `test_open_geotiff_coregister.py` grew 19 -> 30 tests: explicit overview windowing, auto selection, explicit-wins/opt-out, plain-TIFF fallback, similar-resolution no-op, zero overlap, allow_rotated, irregular/flipped templates, antimeridian, integer auto-nearest warning
- [x] `test_open_geotiff_resampling.py` updated for the new advisory warning
- [x] Full geotiff + reproject suites: 6906 passed, 64 skipped, 1 xfailed, no new warnings

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -253,9 +253,29 @@ These combinations raise ``ValueError``:
 * A single-cell caller without ``attrs['res']`` or
   ``attrs['transform']`` -- resolution cannot be inferred from one
   coordinate. The caller must carry ``y``/``x`` coordinates either way.
+* ``allow_rotated=True``. A rotated file reads as an ungeoreferenced
+  pixel grid, which reproject would mis-place wholesale, so the
+  combination is refused up front (#3253).
+* A file that does not overlap the caller's grid at all -- the result
+  would be entirely NaN, so the read raises instead (#3253).
+* An irregular caller grid, or one with ``x`` descending or ``y``
+  ascending. The grid snap assumes a regular x-ascending /
+  y-descending template (reproject's output convention); anything else
+  would misalign silently, so it raises with a ``sortby`` hint
+  (#3253).
 
 What you should know about the output:
 
+* COG overviews are used automatically (#3253). When the caller's grid
+  is coarser than the file's full-resolution pixels and the file
+  carries overviews, the read probes the pyramid with header-only
+  reads and decodes the coarsest overview that still meets the
+  caller's resolution on both axes. Overview pixels are resampled at
+  write time, so values can differ slightly from a full-resolution
+  read; pass ``overview_level=0`` to force full resolution, or
+  ``overview_level=N`` to pick a level yourself (the window is
+  computed in that level's pixel space). Plain TIFFs without
+  overviews read at full resolution as before.
 * The caller's grid always wins. The output shape and coords exactly
   match the caller, and the grid snap (a resample) happens even when
   the CRS already matches. The file's native resolution is not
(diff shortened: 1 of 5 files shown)


brendancol (commented):
## PR Review: coregister=True reads COG overviews and fails fast on bad templates

None found.

- [ ] No test exercises auto overview selection across a CRS mismatch. Every overview test pairs a 4326 template with a 4326 file, so the caller-resolution-in-file-CRS estimate (`res_x_caller` / `res_y_caller`, xrspatial/accessor.py:432-433) only ever runs with an identity transform. A 3857 template over the 4326 COG that expects a specific level would lock the unit conversion in.

- [ ] `_select_overview_level` (xrspatial/accessor.py:253) re-parses the header once per probed level. Local mmap parses are sub-millisecond, but on fsspec/HTTP sources each probe is its own bounded range fetch. Still a good trade (a few small requests against potentially huge full-resolution tile fetches), but a comment would keep someone from "optimizing" the probing away without seeing the remote cost.
- [ ] The orientation error in `_validate_template_grid` (xrspatial/accessor.py:232-240) builds its sortby hint from a conditional string concat inside the raise. Precomputing the hint string would read better.

### Uncertain, flagged for awareness only
- Under strong projection curvature the transformed bbox overstates the caller's cell size, which can pick one overview level coarser than ideal. Levels step by 2x anyway, so the practical effect should be small. No action suggested.

### What looks good
- The overview windowing fix lands in the shared windowed path, so explicit `overview_level=` now works for plain windowed and auto_reproject reads too, not just coregister.
- All new guards fire before any pixel read, and the antimeridian case degrades to a correct superset window instead of failing.
- The overview tests assert both the mechanism (spying on the forwarded `overview_level` kwarg) and the outcome (value agreement with the full-resolution read), which keeps them honest.
- auto_reproject deliberately skips overview selection since it keeps the file's native resolution. Right call.

### Checklist
- [x] Algorithm matches reference (windowing math against the selected level's transform; verified by value-agreement tests)
- [x] All implemented backends produce consistent results (numpy + dask+numpy; GPU raises by design)
- [x] NaN handling is correct (nodata path unchanged; disjoint files now raise instead of returning all NaN)
- [x] Edge cases are covered by tests (zero overlap, rotated opt-in, flipped/irregular templates, antimeridian, no-overview fallback, explicit-wins)
- [x] Dask chunk boundaries handled correctly (chunk forwarding tests still pass; overview read respects caller chunks)
- [x] No premature materialization or unnecessary copies (probing is header-only)
- [x] Benchmark exists or is not needed (no open_geotiff benchmarks exist in benchmarks/; timing evidence is in the PR body)
- [x] README feature matrix updated (n/a: no new function, no backend change)
- [x] Docstrings present and accurate (DataArray + Dataset accessors, geotiff.rst, release contract row)

brendancol (commented):
## Follow-up review (after 9d5bfb77)

- Suggestion, CRS-mismatch auto-overview coverage: **fixed**. `test_coregister_auto_overview_crs_mismatch` runs a 3857 template over the 4326 COG and asserts level 2 is selected, so the file-CRS resolution estimate is now exercised with a real transform.
- Nit, remote probe cost: **fixed**. `_select_overview_level`'s docstring now spells out the per-level header parse cost for local and fsspec/HTTP sources and why the trade holds.
- Nit, orientation error readability: **fixed**. The sortby hint is precomputed instead of concatenated inside the raise.
- Awareness item (curvature can overstate the caller cell size by less than one 2x level step): no action, as noted.

No new findings in the follow-up commit. 43 tests pass across the coregister and resampling suites; flake8 is clean on the changed files.

PR #3255, opened 2026-06-12
Name dask compute tasks xrspatial.<tool> for readable task graphs

### LINKED ISSUE #3250: Name dask tasks after their xrspatial function so task graphs are readable
## Reason or Problem
Every dask backend in xrspatial passes a closure (usually named `_func`) to `map_overlap`/`map_blocks` without a `name=` argument. Dask derives graph key prefixes from the function's `__name__`, so `.visualize()` on a pipeline like `hillshade(slope(terrain(...)))` renders a wall of identical `_func-<hash>` nodes. You can't tell which node belongs to which tool.

## Proposal
Pass `name=f"xrspatial.<tool>"` at each `map_blocks`/`map_overlap` call site in the dask backends, and give the chunk closures real names.

**Design:**
- `da.map_blocks` and `Array.map_overlap` both take `name=`. Dask appends its deterministic token hash, so keys stay unique (checked on dask 2025.7.0). The older `token=` spelling is deprecated.
- The key prefix is what shows up as the node label in `visualize()` and as the layer name in the high-level graph.
- `visualize()` also draws the function name per task, so the `_func` closures should be renamed (e.g. `_slope_chunk`).
- Affected modules: slope.py, aspect.py, convolution.py, bilateral.py, terrain.py, preview.py, surface_distance.py, hydro/flow_direction_mfd.py, hydro/flow_length_mfd.py (~30 call sites).

**Usage:**
```python
import xrspatial
out = xrspatial.slope(big_dask_raster)
out.data.visualize("graph.png")   # nodes now read xrspatial.slope-<hash>
```

**Value:** task graphs and the distributed dashboard identify xrspatial operations by name instead of `_func`.

## Stakeholders and Impacts
Dask users of xrspatial. No API change; only dask graph key names change. Anything that depends on exact key names would notice, but nothing should.

## Drawbacks
A few extra kwargs at call sites.

## Alternatives
Renaming only the closures (no `name=`) also improves the default prefix, but `name=` gives a stable namespaced label that doesn't depend on internal function names.

## Unresolved Questions
Whether to name intermediate `map_blocks` calls (dtype conversions in surface_distance, lat/lon broadcast helpers) or only the main compute task per tool.

brendancol:
Closes #3250

- Pass `name='xrspatial.<tool>'` to the `map_blocks`/`map_overlap` calls in the dask backends, so `.visualize()` and the distributed dashboard label compute tasks by tool instead of `_cpu`/`_func`. `surface_distance` picks the name by mode (`surface_distance`/`surface_allocation`/`surface_direction`); `flow_direction_mfd` names each of its 8 band tasks.
- Did not rename any chunk closures: they all turned out to be `functools.partial` over named backend functions, which dask already unwraps for the function label, so the key prefix was the only missing piece.
- Helper `map_blocks` calls (dtype conversion, lat/lon broadcast, cpu<->gpu transfer) keep their default names on purpose; they read fine as plumbing.
- `name=` appends dask's deterministic hash on every dask version this repo can run (checked the wheels back to 2023.11.0, the oldest with python 3.12 support), so two calls of the same tool can't collide. `test_same_name_no_key_collision` guards that.

Backends: numpy unaffected; dask+numpy and dask+cupy call sites both named; cupy unaffected (no graph).

Test plan:
- [x] new `xrspatial/tests/test_dask_task_names.py` (14 tests: prefix per tool, 3 surface modes, 8 MFD bands, collision safety, value parity)
- [x] existing suites for touched modules: 822 passed (slope/aspect/convolution/bilateral/terrain/preview/surface_distance/focal) + 204 passed (hydro MFD, geodesic, dask laziness/fused overlap)

### DIFF
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -22,6 +22,7 @@ from xrspatial.utils import (Z_UNITS, ArrayTypeFunctionMapping, _boundary_to_das
                              _extract_latlon_coords, _pad_array, _validate_boundary,
                              _validate_raster, cuda_args, get_dataarray_resolution,
                              ngjit, warn_if_unit_mismatch)
+from xrspatial.utils import _dask_task_name_kwargs
 
 
 def _geodesic_cuda_dims(shape):
@@ -180,7 +181,7 @@ def _run_dask_numpy(data: da.Array,
                            depth=(1, 1),
                            boundary=_boundary_to_dask(boundary),
                            meta=np.array((), dtype=np.float32),
-                           name='xrspatial.aspect')
+                           **_dask_task_name_kwargs('xrspatial.aspect'))
     return out
 
 
@@ -194,7 +195,7 @@ def _run_dask_cupy(data: da.Array,
                            depth=(1, 1),
                            boundary=_boundary_to_dask(boundary, is_cupy=True),
                            meta=cupy.array((), dtype=cupy.float32),
-                           name='xrspatial.aspect')
+                           **_dask_task_name_kwargs('xrspatial.aspect'))
     return out
 
 
@@ -300,7 +301,7 @@ def _run_dask_numpy_geodesic(data, lat_2d, lon_2d, a2, b2, z_factor, boundary='n
         depth=(0, 1, 1),
         boundary={0: np.nan, 1: dask_bnd, 2: dask_bnd},
         meta=np.array((), dtype=np.float32),
-        name='xrspatial.aspect',
+        **_dask_task_name_kwargs('xrspatial.aspect'),
     )
     return out[0]
 
@@ -333,7 +334,7 @@ def _run_dask_cupy_geodesic(data, lat_2d, lon_2d, a2, b2, z_factor, boundary='na
         depth=(0, 1, 1),
         boundary={0: cupy.nan, 1: dask_bnd, 2: dask_bnd},
         meta=cupy.array((), dtype=cupy.float32),
-        name='xrspatial.aspect',
+        **_dask_task_name_kwargs('xrspatial.aspect'),
     )
     return out[0]
 
--- a/xrspatial/aspect.py
+++ b/xrspatial/aspect.py
@@ -179,7 +179,8 @@ def _run_dask_numpy(data: da.Array,
     out = data.map_overlap(_func,
                            depth=(1, 1),
                            boundary=_boundary_to_dask(boundary),
-                           meta=np.array((), dtype=np.float32))
+                           meta=np.array((), dtype=np.float32),
+                           name='xrspatial.aspect')
     return out
 
 
@@ -192,7 +193,8 @@ def _run_dask_cupy(data: da.Array,
     out = data.map_overlap(_func,
                            depth=(1, 1),
                            boundary=_boundary_to_dask(boundary, is_cupy=True),
-                           meta=cupy.array((), dtype=cupy.float32))
+                           meta=cupy.array((), dtype=cupy.float32),
+                           name='xrspatial.aspect')
     return out
 
 
@@ -298,6 +300,7 @@ def _run_dask_numpy_geodesic(data, lat_2d, lon_2d, a2, b2, z_factor, boundary='n
         depth=(0, 1, 1),
         boundary={0: np.nan, 1: dask_bnd, 2: dask_bnd},
         meta=np.array((), dtype=np.float32),
+        name='xrspatial.aspect',
     )
     return out[0]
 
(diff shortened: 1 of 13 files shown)


brendancol (commented):
## PR Review: Name dask compute tasks xrspatial.<tool> for readable task graphs

None. The change is graph metadata only; the collision test covers the one way this could have corrupted results.

- [ ] The issue's motivating example is `hillshade(slope(terrain(...)))`, but hillshade is not covered: `xrspatial/hillshade.py:87` and `:140` still produce default key names, so the example pipeline stays partly unlabeled. Same for `xrspatial/curvature.py:64`/`:117`, the other core surface tool. Add `name='xrspatial.hillshade'` / `name='xrspatial.curvature'` in this PR.
- [ ] The repo has roughly 280 `map_blocks`/`map_overlap` call sites across ~70 modules (flood, multispectral, fire, zonal, focal, classify, proximity, the rest of hydro, ...). The issue text lists only the 9 modules this PR touches, so the PR matches the issue as written, but the sweep is ~10x larger than the issue estimated. File a follow-up issue for the rest so the convention doesn't stall at 9 modules. `focal.py` (lines 210, 220, 565, 595, 1247, 1589-1631) is the highest-traffic omission after hillshade.
- [ ] `xrspatial/tests/test_dask_task_names.py:36` uses evenly dividing chunks (6, 6) on a 12x12 raster. One case with ragged chunks (say `chunks=(5, 7)`) would confirm naming doesn't interact with non-uniform chunk shapes.

- [ ] `xrspatial/hydro/flow_direction_mfd.py:324` names bands `xrspatial.flow_direction_mfd_band{k}`, departing from the plain `xrspatial.<tool>` pattern. The reason (8 distinct layers per call must not share a key) is worth a one-line comment at the call site.
- [ ] The dask+cupy call sites get no test coverage here (no GPU in CI). The kwarg is identical on both paths so risk is low; noting it for the record.

### What looks good
- `test_same_name_no_key_collision` directly guards the worst failure mode: older dask's verbatim `name=` semantics silently merging two computations. The PR also checked wheels back to 2023.11.0.
- `surface_distance` names tasks by mode, so `surface_allocation` graphs don't claim to be `surface_distance`.
- Leaving plumbing tasks (dtype conversion, lat/lon broadcast, cpu/gpu transfer) unnamed keeps the labeled nodes meaningful.
- No closure renames were needed: the chunk functions are partials over named backend functions, and the PR documents that instead of churning code.

### Checklist
- [x] Algorithm matches reference (n/a, metadata-only change)
- [x] All implemented backends produce consistent results (dask+numpy verified; dask+cupy symmetric, untested locally)
- [x] Edge cases are covered by tests (collision, all 3 surface modes, 8 MFD bands, unbounded path)
- [x] Dask chunk boundaries handled correctly (untouched; existing boundary tests green)
- [x] Benchmark exists or is not needed (not needed; tokenize already ran on the default path)

brendancol (commented):
## Follow-up review (a46af442)

- **hillshade/curvature unnamed** -- fixed. `name='xrspatial.hillshade'` / `'xrspatial.curvature'` added at all four call sites (both dask+numpy and dask+cupy), with prefix tests for each.
- **Repo-wide sweep (~260 remaining call sites)** -- deferred to #3256, which lists the remaining modules and the convention to follow. Doing it here would grow the diff by an order of magnitude.
- **Ragged-chunk coverage** -- fixed. `test_task_names_with_ragged_chunks` runs slope on (5, 7) chunks over a 12x12 raster and checks both the key prefix and value parity with numpy.
- **Band-suffix naming comment** -- fixed. Both `flow_direction_mfd` call sites now explain why the band index is in the name.
- **No dask+cupy test coverage** -- no action; CI has no GPU and the kwarg is identical on both paths.

Re-checked the delta: the four new `name=` kwargs match the pattern from the first commit, and the ragged test's smallest chunk (2) stays above the overlap depth (1). 203 tests pass across the new file plus the hillshade, curvature, and flow_direction_mfd suites. Nothing new to flag.

brendancol:
The first CI round failed (focal, emerging_hotspots, and the new collision test) and the failure was real, not flaky: CI installs dask 2026.6.0, where dask/dask#11952 reverted `map_blocks(name=...)` to verbatim-key semantics and un-deprecated `token=` as the hashed-prefix spelling. With a verbatim key, every `convolve_2d` in one graph shared the task `xrspatial.convolve_2d`, so dask merged them and Gi* computed the kernel convolution where it needed the kernel-squared one. The exact failure mode the collision test was written for, on a dask released after the version I verified against.

Fixed in 059e2086: `_dask_task_name_kwargs()` in utils.py probes once per process which spelling appends the hash and call sites pass `**_dask_task_name_kwargs('xrspatial.<tool>')`. Verified the full touched-module suites green on both dask 2025.7.0 (name= hashes) and 2026.6.0 (token= hashes): 863 and 654 tests respectively, including the suites that failed on CI.

PR #3258, opened 2026-06-12
Replace the usage page with a runnable quickstart

brendancol:
Refs #3249 (part 4 of 6).

usage.rst showed fragments (`xr.DataArray(...)`) that couldn't be run. The new quickstart.rst runs top to bottom on a base install with no data download:

- first analysis on `generate_terrain` synthetic elevation, with a plot line marked as needing the `plot` extra
- the `.xrs` accessor shown as equivalent to the functional style
- a GeoTIFF/COG round trip via `to_geotiff(..., crs=3857, cog=True)` and `open_geotiff`
- the Dataset-support section carried over from usage.rst (it was accurate)
- a dask teaser (`.chunk()` then the same call stays lazy) and a GPU pointer
- a warning box sending readers to the caveats page (WGS84 assumption, float32 outputs, NaN nodata) and the stability policy before production use

usage.rst is deleted; its GDAL section gets a permanent home in the why-xarray-spatial page (part 5 of the series, PR coming next). No other page references usage.rst, so nothing dangles in the meantime.

Test plan:
- [x] every code block executed end to end as written (hillshade/slope shapes, accessor matches functional result, round-tripped GeoTIFF reads back 2D with crs 3857, chunked slope stays dask-backed until compute)
- [x] `sphinx-build -b html` clean, pre-existing warnings only
- [x] grep for `getting_started/usage`, `getting_started.usage`, and `:doc:` usage refs across docs/, README, CONTRIBUTING: none


brendancol (commented):
## PR Review: Replace the usage page with a runnable quickstart

### Blockers
- quickstart.rst (GeoTIFF section): the page promises "everything on this page runs with just pip install xarray-spatial", but `to_geotiff(terrain, ..., crs=3857)` raises `UnknownCRSModelTypeError` on a base install. Without pyproj the writer only classifies EPSG codes in `_KNOWN_GEOGRAPHIC_EPSG_FALLBACK` (xrspatial/geotiff/_geotags.py:1464), which is geographic-only; 3857 is not in it. Reproduced by blocking the pyproj import: 3857 raises, 4326 round-trips fine. Switch the example to `crs=4326` with degree-scale ranges, and note that other EPSG codes want the `geotiff` extra.

### Suggestions
None.

### Nits
None.

### What looks good
- Every code block was executed end to end (terrain/hillshade/slope shapes, accessor parity with the functional call, dask laziness through compute).
- The Dataset section moves over verbatim and was accurate.
- No dangling references to usage.rst anywhere in docs/, README, or CONTRIBUTING after the deletion.

### Checklist
- Backend/NaN/dask code items: not applicable (docs-only)
- Docs build: clean, pre-existing warnings only

brendancol (commented):
Fixed: the quickstart now generates terrain over a 1-degree tile and writes with crs=4326, which the writer classifies without pyproj (it's in the geographic fallback allowlist). Added a sentence noting other EPSG codes need the geotiff extra. Re-verified with pyproj blocked from importing: the round trip runs on a simulated base install.

PR #3269, opened 2026-06-12
Correct the unpack docstring: masking still runs without scale/offset (#3263)

### LINKED ISSUE #3263: open_geotiff docstring calls unpack=True a "no-op" without scale/offset, but it still masks nodata and promotes to float64
**Describe the bug**

The `unpack` parameter doc in `open_geotiff` (xrspatial/geotiff/__init__.py:712-713) says:

> A source without scale / offset metadata is a no-op.

That's wrong for any source that declares a nodata sentinel. `unpack=True` folds into the masking gate (`_finalize_eager_read` in `_attrs.py`: "``mask_and_scale`` implies masking"), so even with no SCALE/OFFSET tags the read replaces the sentinel with NaN and promotes integer sources to float64. The behavior itself is correct and deliberate (it matches rioxarray's `mask_and_scale`); the docstring is what's broken.

**Repro**

```python

ints = np.array([[1, 2], [3, 65535]], dtype=np.uint16)
da = xr.DataArray(ints, dims=("y","x"), coords={"y":[2.0,1.0],"x":[0.0,1.0]},
                  attrs={"crs": 4326, "nodata": 65535})
to_geotiff(da, "t.tif")  # no SCALE/OFFSET written

open_geotiff("t.tif").dtype              # uint16, values [1 2 3 65535]
open_geotiff("t.tif", unpack=True).dtype # float64, values [1. 2. 3. nan]
```

Same result on all four backends (eager, dask, gpu, dask+gpu), so this is not a parity bug. A source with neither scale/offset nor a sentinel really is a no-op (uint16 comes back unchanged).

**Expected behavior**

The docstring should say the scaling step is skipped when there's no SCALE/OFFSET metadata, but the sentinel-to-NaN mask (and the float64 promotion it implies for integer sources) still runs. A user reading "no-op" today would expect their uint16 to survive `unpack=True` untouched.

**Fix**

Docs-only: rewrite the sentence in the `unpack` param doc. No signature or behavior change, so no deprecation impact. Add a regression test pinning the documented behavior (sentinel-only source: masked + promoted; bare source: unchanged).

Found by the 2026-06-12 api-consistency sweep (Cat 3, docstring/behavior divergence).

brendancol:
Closes #3263

- The `unpack` param doc on `open_geotiff` said a source without SCALE/OFFSET metadata "is a no-op". It isn't: `unpack=True` implies the sentinel-to-NaN mask (rioxarray `mask_and_scale` parity), so a uint16 source with a declared nodata comes back float64 with NaN holes. The doc now says the scaling step is skipped, the mask still runs, and only a source with neither scale/offset nor a sentinel reads unchanged.
- Docs-only change; no signature or behavior edits.
- New `test_unpack_noop_doc_3263.py` pins the corrected wording plus the behavior it describes (sentinel-only source masks and promotes; bare source reads unchanged).

Backends: behavior verified identical on numpy, dask, cupy, and dask+cupy before filing (the divergence was docs-only). Dtype parity across eager/dask was already pinned by `test_mask_and_scale_dtype_parity_3066.py`.

Test plan:
- [x] `pytest xrspatial/geotiff/tests/read/test_unpack_noop_doc_3263.py`
- [x] `pytest xrspatial/geotiff/tests/read/test_mask_and_scale_dtype_parity_3066.py xrspatial/geotiff/tests/parity/test_signature_contract.py`


brendancol (commented):
## PR Review: Correct the unpack docstring (#3263)

### Blockers
None.

### Suggestions
None.

### Nits
- [ ] `test_unpack_doc_does_not_claim_noop_3263` asserts `"is a no-op" not in doc` against the whole `open_geotiff` docstring. If a future parameter legitimately documents a no-op case, this test fails for an unrelated reason. Scoping the assertion to the `unpack` paragraph (slice between `unpack : bool` and `mask_and_scale :`) keeps the pin tight.

### What looks good
- The corrected wording matches verified behavior: sentinel-only sources mask + promote to float64 on all four backends (checked on numpy, dask, cupy, dask+cupy before the issue was filed), and a source with neither scale/offset nor a sentinel reads back as uint16 unchanged. The implementation comment in `_attrs.py` (`mask_and_scale implies masking`) confirms the behavior is the intended rioxarray parity, so a docs-only fix is the right scope.
- Behavior tests complement rather than duplicate `test_mask_and_scale_dtype_parity_3066.py`: 3066 pins eager/dask dtype parity, this PR pins values, NaN position, and the `mask_and_scale_dtype` attr for the sentinel-only case.
- No signature change, so the 218 signature-contract tests are unaffected (verified locally).

### Checklist
- Algorithm matches reference: yes (docs now match `_finalize_eager_read`)
- Backend consistency: verified manually, divergence was docs-only
- NaN handling: pinned by the new sentinel test
- Edge cases: bare source (true no-op) covered
- Dask chunks / materialization / benchmarks: n/a, docs-only
- README matrix: n/a
- Docstrings: this PR is the docstring fix

brendancol (commented):
Follow-up on the review above:

- Nit (whole-docstring "is a no-op" assertion): fixed in c14844a8. The pin now slices the `unpack : bool` ... `mask_and_scale : bool` paragraph and asserts inside it, so an unrelated parameter documenting a real no-op case cannot trip the test. 3/3 tests pass.

No open findings remain.

PR #3271, opened 2026-06-12
merge() keeps the shared integer input dtype instead of promoting to float64

### LINKED ISSUE #3262: merge() silently promotes integer inputs to float64; reproject() preserves integer dtype
**Describe the bug**

`merge()` always returns float64, even when every input raster is int16 or uint8. All three merge paths hardcode it: `_merge_inmemory` builds float64 arrays, `_merge_dask` advertises a float64 template, and `_merge_arrays_numpy` allocates `np.full(..., dtype=np.float64)`.

`reproject()` used to do the same thing. It got fixed path by path (#2185, #2505, #3093, #3096), and integer sources now round-trip back to their dtype after clamping on every backend. `merge()` never got that treatment, so a pipeline that reprojects int16 tiles (getting int16 back) and then mosaics them ends up with a float64 result. A uint8 mosaic takes 8x the memory, and writing it back to GeoTIFF changes the file's dtype.

Nothing pins the current behavior either: the public docstring doesn't mention output dtype, and no test asserts float64 output for integer inputs.

Reproducer:

```python
import numpy as np
import xarray as xr
from xrspatial.reproject import merge, reproject

tile = xr.DataArray(
    np.arange(64, dtype=np.int16).reshape(8, 8),
    dims=['y', 'x'],
    coords={'y': np.linspace(5, -5, 8), 'x': np.linspace(-5, 5, 8)},
    attrs={'crs': 'EPSG:4326', 'nodata': -32768},
)
print(reproject(tile, 'EPSG:4326', resolution=1.25).dtype)  # int16
print(merge([tile]).dtype)                                  # float64
```

**Expected behavior**

When every input shares one integer dtype, the mosaic comes back in that dtype, using the same round/clip/cast convention `reproject()` applies. The output nodata gets resolved with the dtype hint (`_detect_nodata(..., dtype=...)`) so NaN is swapped for a representable sentinel (#2185) and an explicit out-of-range nodata raises (#2572). Mixed dtypes or any float input keep returning float64.

**Additional context**

Found by the metadata propagation sweep. The dask template dtype has to match what the chunks actually return, or the graph advertises one dtype and computes another (the #3096 trap).

brendancol:
Closes #3262.

- When every input shares one integer dtype, `merge()` now casts the mosaic back to that dtype using the same round/clip/cast convention as `reproject()`'s integer round-trip (#2505/#3093). Mixed dtypes and float inputs still return float64, and the docstring now says so.
- The output nodata is resolved with the dtype hint (`_detect_nodata(..., dtype=out_dtype)`): an auto-detected NaN swaps for `dtype.min`/`dtype.max` (#2185) and an explicit out-of-range value raises ValueError (#2572), matching `reproject()`.
- The dask template, `meta`, and empty-chunk fills all carry the output dtype, so the lazy graph advertises the dtype the chunks actually return (the #3096 trap).

Backend coverage: numpy, cupy, dask+numpy, and dask+cupy, each exercised by a test. The cupy and dask+cupy paths go through the existing host merge and come back to the GPU with the dtype intact.

Test plan:
- [x] 14 new tests in `TestMergeIntegerDtype`: int16/uint8 eager, dask advertised-vs-computed dtype, no-overlap chunk fills, cupy and dask+cupy round-trips, default and declared sentinels, mean-strategy rounding, out-of-range nodata raise, mixed-dtype and float fallbacks, exact value preservation on same-CRS placement
- [x] Full reproject suite: 515 passed (CUDA host, all four backends live)

Also updates `.claude/sweep-metadata-state.csv` with the 2026-06-12 metadata sweep entry for the reproject module.


brendancol (commented):
## PR Review: merge() keeps the shared integer input dtype instead of promoting to float64

None found.

- [ ] `xrspatial/reproject/__init__.py:2331` -- routing the explicit `nodata` through `_detect_nodata(rasters[0], nodata, dtype=out_dtype)` is a behavior change beyond the dtype fix: `merge(..., nodata=np.inf)` previously passed the value through and now raises ValueError, and an explicit int sentinel now lands in `attrs['nodata']` as a float. Both align with `reproject()`, which is the right call, but neither is pinned. Add a small test asserting the inf rejection so the alignment doesn't regress silently.

- [ ] A fractional in-range sentinel on an integer mosaic (say `attrs['nodata'] = 0.5` on int16 tiles) gets rounded into the data range by `_cast_merged_dtype`, so the holes become 0 while `attrs['nodata']` says 0.5. `reproject()` has the same quirk for the same pathological input (the int data never matches 0.5 either way), so this is the existing convention rather than a new bug. Fine to leave; noting it for the record.

### What looks good
- The dtype rule mirrors the reproject contract exactly (shared integer dtype round-trips, mixed or float stays float64), and `_cast_merged_dtype` reuses the round/clip/cast convention from the chunk workers.
- The dask template, `meta`, and both empty-chunk fills (`_merge_block_adapter` no-overlap and no-arrays branches) all carry `out_dtype`, so the lazy graph can't advertise one dtype and compute another.
- Sentinel resolution reuses `_detect_nodata`'s dtype hint, so the NaN swap (#2185) and out-of-range rejection (#2572) behave the same as in `reproject()` with no duplicated logic.
- 14 tests cover the four backends, advertised-vs-computed dask dtype, no-overlap chunk fills, default and declared sentinels, mean rounding, and exact value preservation on same-CRS placement.
- The sweep state CSV rewrite was caught and corrected to LF-only with main's rows preserved byte for byte, avoiding the #2754 whole-file conflict mode.

### Checklist
- [x] Algorithm matches reference/paper (n/a -- dtype plumbing, convention matches reproject)
- [x] NaN handling is correct (NaN canonical sentinel converted before the integer cast; resolved sentinel is non-NaN by construction for integer output)
- [x] Edge cases are covered by tests (no-overlap chunks, gaps, mixed dtypes, out-of-range nodata)
- [x] Dask chunk boundaries handled correctly (per-block cast, template chunks unchanged)
- [x] No premature materialization or unnecessary copies (one extra full-array cast pass, unavoidable)
- [x] Benchmark exists or is not needed (no new function; cast cost is negligible next to the merge)
- [x] README feature matrix updated (n/a -- no new function, no backend change)
- [x] Docstrings present and accurate (Returns and nodata parameter document the new rule)

brendancol (commented):
Follow-up review after 973b882a: the one suggestion from the first pass is addressed. `test_merge_rejects_inf_nodata` pins the ValueError on `nodata=np.inf` and `test_merge_explicit_int_sentinel_lands_as_float_attr` pins the float coercion of an explicit integer sentinel, so both behavior alignments with reproject() are now under test. The nit about fractional in-range sentinels on integer mosaics stays as a recorded note (existing convention shared with reproject(), no change needed). Nothing further.

PR #3272, opened 2026-06-12
Refuse pack=True values the packed integer dtype cannot hold

### LINKED ISSUE #3260: to_geotiff(pack=True) silently wraps out-of-range values in the integer cast; Inf casts to garbage
**Describe the bug**

`_pack` (xrspatial/geotiff/_attrs.py) reverses the unpack transform with `(data - add_offset) / scale_factor`, rounds, and casts to the integer dtype recorded in `attrs['mask_and_scale_dtype']`. Nothing checks that the packed values actually fit that dtype, so two kinds of pixels get corrupted on disk with no error:

- Finite values whose packed form falls outside the dtype range wrap in the `astype`. An int16 source with SCALE=0.1, one pixel edited to 4000.0 after the unpack read, packs to 40000 and lands on disk as -25536. Reopening with `unpack=True` returns -2553.6.
- +/-Inf pixels pass the NaN guard (it checks `isnan` only) and cast to whatever the platform produces. On linux/x86 that is 0, with nothing but a numpy `RuntimeWarning: invalid value encountered in cast` from deep inside xarray.

All four backends are affected. The dask path defers the cast into the write's compute, so there is no warning at all at call time.

The function already fails closed for the neighbouring case: NaN pixels with no sentinel raise ValueError, and the comment there says the astype "would silently wrap it into a valid-looking integer" (_attrs.py around line 1875). Out-of-range finite values and Inf hit the same astype and wrap the same way, but have no guard.

Repro:

```python

data = np.array([[100, 200], [300, 32000]], dtype=np.int16)
da = xr.DataArray(data, dims=('y','x'), coords={'y':[1.5,0.5],'x':[0.5,1.5]})
da.attrs['gdal_metadata'] = {'SCALE': '0.1', 'OFFSET': '0.0'}
to_geotiff(da, 'src_3259.tif')

mod = open_geotiff('src_3259.tif', unpack=True)
mod.values[0, 0] = 4000.0          # packs to 40000 > int16 max
to_geotiff(mod, 'packed_3259.tif', pack=True)

print(open_geotiff('packed_3259.tif').values)
# [[-25536    200]
#  [   300  32000]]
```

**Expected behavior**

Refuse to write pixels that cannot be represented in the packed dtype, the same way pack already refuses NaN with no sentinel: ValueError naming the packed dtype and its range. Eager input should raise at call time; dask input should raise from the write's single compute (same shape as the #3235 NaN guard). A raster writer silently changing pixel values is the worst failure mode it has.

**Additional context**

Found by /sweep-accuracy against the geotiff module. The unpack -> process -> pack workflow is the whole point of the feature, and processing routinely pushes values outside the source's packed range.

brendancol:
Closes #3260

- `_pack` now range- and finiteness-checks the packed values after the round and before the integer cast. Finite values outside the target dtype's range used to wrap in the `astype` (40000 -> -25536 in int16) and +/-Inf cast to a platform-defined integer, all silently. Both now raise ValueError naming the dtype and its range.
- The exclusive upper bound is `iinfo.max + 1` as a float. `float(iinfo.max)` rounds up to `2**63` / `2**64` for the 64-bit dtypes, so a `>` test against it would let exactly-`2**63` through to wrap; `iinfo.max + 1` is a power of two and exact in float64.
- Error timing matches the existing NaN-no-sentinel guard (#3235): numpy and cupy raise at `_pack` call time, dask backings map the guard into the graph and raise from the write's single compute.
- Fixed en route: the eager NaN-no-sentinel scan used `bool(out.isnull().any())`, which materialises through `np.asarray` and raised TypeError on cupy-backed eager input before the scan could answer. It now goes through the cupy-safe `_pack_guard_no_nan`, same message.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy (one guard function; `np.isfinite` and the comparisons dispatch on the array type).

Test plan:
- [x] 15 new tests in `xrspatial/geotiff/tests/write/test_pack_range_guard_3260.py`: finite overflow and uint underflow rejected on all four backends (gpu legs run on CUDA), +/-Inf rejected, exact `iinfo.min`/`iinfo.max` round-trips, values that round back into range pass, float targets not range-guarded, `2**63` rejected for int64 while one ULP below passes
- [x] `xrspatial/geotiff/tests/write/` (1145 passed), `tests/attrs` + unpack parity + unit (1015 passed), `tests/gpu` (430 passed)

Found by /sweep-accuracy. The sweep state CSV update for the geotiff module rides along.


brendancol (commented):
## PR Review: Refuse pack=True values the packed integer dtype cannot hold

None.

- [ ] `xrspatial/geotiff/tests/write/test_pack_range_guard_3260.py:168` -- `test_pack_float_target_not_range_guarded` reopens `src` instead of `out`, so the assertion checks the source file, not the packed output the test is about. The write succeeding already covers the no-raise claim, but the dtype assertion should run against `out` (and could also check the packed value, 1e30 / 2 = 5e29, which fits float32).

- [ ] `xrspatial/geotiff/_attrs.py` (`_pack_guard_int_range`) -- the guard makes three passes over the chunk (`isfinite`, `< lo`, `>= hi_excl`) plus two `|=` merges. Elementwise and cheap relative to codec work, but a single fused expression would halve the temporaries if this ever shows up in a profile. Fine as is.
- [ ] `test_pack_range_guard_3260.py:44-46` -- `_unpacked` copies attrs explicitly after `da.copy()`, which already copies attrs. Harmless duplication.

### What looks good
- The exclusive upper bound `iinfo.max + 1` is the right call: `float(iinfo.max)` rounds up to `2**63` for int64, and the unit test pins exactly that boundary (rejects `2**63`, accepts one ULP below).
- Guard placement after the round means values like 32767.4 that round back into range are accepted; tested.
- Error timing mirrors the existing #3235 NaN guard (eager at call time, dask from the write's single compute), so pack keeps one consistent failure model.
- The `bool(out.isnull().any())` -> `_pack_guard_no_nan` change fixes a real eager-cupy TypeError on the no-sentinel integer path and removes a duplicated error-message literal at the same time.
- All four backends covered by the same guard function via ufunc dispatch; gpu and dask-gpu test legs execute on CUDA rather than being reasoned about.

### Checklist
- [x] Algorithm matches reference (cast-wrap semantics verified empirically pre-fix)
- [x] NaN handling is correct (NaN still routed to the fill / NaN guard before this one)
- [x] Edge cases are covered by tests (bounds, rounding, Inf, 64-bit float64 limits)
- [x] Dask chunk boundaries handled correctly (per-chunk elementwise guard)
- [x] No premature materialization (guard fuses into the write's compute)
- [x] Benchmark not needed (validation guard, no algorithmic change)
- [x] Docstrings updated (_pack and the to_geotiff pack kwarg)

brendancol (commented):
Follow-up review after f1e4fab0.

- Suggestion (test asserted on `src` instead of `out`): fixed. The float-target test now reopens the packed output and also checks the packed value (5e29 in float32).
- Nit (redundant attrs copy in `_unpacked`): fixed, helper is now a one-liner.
- Nit (three-pass guard scan): dismissed. The `|=` form already merges in place; a single fused expression would allocate the same temporaries and the guard is elementwise-cheap next to codec work. Revisit only if it shows in a profile.

15/15 tests pass after the changes, including the gpu and dask-gpu legs. No new findings.

PR #3273, opened 2026-06-12
Export VRTUnsupportedError, CloudSizeLimitError, PixelSafetyLimitError publicly (#3265)

### LINKED ISSUE #3265: Three exceptions raised on public open_geotiff paths are missing from xrspatial.geotiff's public surface
**Describe the bug**

`xrspatial/geotiff/__init__.py` exports 17 exception classes in `__all__`, but three more are raised on public `open_geotiff` paths and aren't exported. Users can only catch them via a base class or by importing from private modules:

| Exception | Defined in | Raised when | Importable from `xrspatial.geotiff`? | In `__all__`? |
|---|---|---|---|---|
| `VRTUnsupportedError` | `_errors.py:129` | a `.vrt` source declares a feature the read pipeline rejects (10+ raise sites in `_vrt_validation.py`) | no | no |
| `CloudSizeLimitError` | `_sources.py:79` | an eager fsspec read exceeds `max_cloud_bytes` | yes (bound in `__init__.py:76`) | no |
| `PixelSafetyLimitError` | `_layout.py:109` | a read exceeds `max_pixels` (a stable-tier kwarg) | no | no |

The inconsistencies that make this user-facing:

- `docs/source/user_guide/geotiff_safe_io.rst:232` documents `VRTUnsupportedError` in the error table, but unlike every other row it can't use the `:class:~xrspatial.geotiff.` role, and the page tells readers the class lives in `:mod:xrspatial.geotiff._errors`. Its companion in the same sentence, `UnsupportedGeoTIFFFeatureError`, also lives in `_errors` and IS publicly exported, so the doc steers users to a private import for one class and a public one for the other.
- `UnsafeURLError` and `CloudSizeLimitError` are siblings: both defined in `_sources.py`, both raised on the cloud-read path, both `ValueError` subclasses. One is in `__all__`, the other isn't. `CloudSizeLimitError` is also named in `docs/source/reference/geotiff_release_contract.md`.
- `max_pixels` is documented `[stable]` on `open_geotiff`, but the exception it raises is only importable from `_layout`, which is module-layout plumbing (`tests/integration/test_sidecar.py` and the vrt tests already import these from the private modules).

All three subclass `ValueError` (and `VRTUnsupportedError` subclasses the exported `GeoTIFFAmbiguousMetadataError`), so broad catches keep working. The gap is catching the specific class without reaching into private modules.

**Expected behavior**

`from xrspatial.geotiff import VRTUnsupportedError, CloudSizeLimitError, PixelSafetyLimitError` works, the names are in `__all__`, and the safe-io doc cross-references them like its other rows.

**Fix**

Additive, no rename and no deprecation impact: import the three classes in `xrspatial/geotiff/__init__.py`, add them to `__all__`, update the `geotiff_safe_io.rst` cross-reference, and extend the existing error-export test. Existing private import paths keep working.

Found by the 2026-06-12 api-consistency sweep (Cat 5, public API surface drift).

brendancol:
Closes #3265

- Imports `VRTUnsupportedError` (from `_errors`) and `PixelSafetyLimitError` (via `_reader`'s historical re-export of `_layout`) into `xrspatial/geotiff/__init__.py` and adds both, plus the already-bound `CloudSizeLimitError`, to `__all__`. All three are raised on public `open_geotiff` paths (VRT capability validation, `max_cloud_bytes`, `max_pixels`) and were the only such exceptions missing from the public surface; the other 17 error classes were already exported.
- Updates `docs/source/user_guide/geotiff_safe_io.rst`: the VRT error-table row and the `open_geotiff` overview now use the `:class:~xrspatial.geotiff.` role instead of steering readers to the private `_errors` module.
- Purely additive: no renames, no behavior change, private import paths (`_errors`, `_layout`, `_reader`) keep working and the new test pins object identity with them.

Backends: not applicable, no runtime dispatch touched. The change is namespace plus docs.

Test plan:
- [x] `pytest xrspatial/geotiff/tests/unit/test_exception_exports_3265.py` (export, `__all__`, identity, base-class catch, functional `max_pixels` raise)
- [x] `pytest xrspatial/geotiff/tests/parity/test_api_consolidation.py xrspatial/geotiff/tests/unit/test_metadata.py` (existing surface-pinning tests still pass)
- [x] isort / flake8 clean on the touched files

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -414,7 +414,9 @@ turn the process into a port scanner. The knobs are:
 * ``max_cloud_bytes`` (kwarg) / ``XRSPATIAL_GEOTIFF_MAX_CLOUD_BYTES``
   (env). Per-call total byte budget for a remote read. The kwarg wins
   over the env var; the env var wins over the built-in default. Pass
-  ``max_cloud_bytes=None`` to disable the cap on a single call. Locked
+  ``max_cloud_bytes=None`` to disable the cap on a single call. A
+  breach raises :class:`xrspatial.geotiff.CloudSizeLimitError` (a
+  ``ValueError`` subclass). Locked
   by ``xrspatial/geotiff/tests/integration/test_http_sources.py``
   (max_cloud_bytes_dispatcher and max_cloud_bytes_annot sections, plus
   the http_read_all_bounded section).
--- a/docs/source/user_guide/geotiff_safe_io.rst
+++ b/docs/source/user_guide/geotiff_safe_io.rst
@@ -41,10 +41,10 @@ the read and write paths:
        eager numpy reader; dask, GPU, VRT, and remote-URL paths require
        a string. A ``.vrt`` source reads a GDAL mosaic (tier:
        ``advanced``) over a documented subset of the GDAL VRT schema;
-       unsupported features raise ``VRTUnsupportedError`` or
+       unsupported features raise
+       :class:`xrspatial.geotiff.VRTUnsupportedError` or
        :class:`xrspatial.geotiff.UnsupportedGeoTIFFFeatureError` at
-       graph-build time rather than producing wrong pixels. Both error
-       classes live in :mod:`xrspatial.geotiff._errors`.
+       graph-build time rather than producing wrong pixels.
    * - :func:`xrspatial.geotiff.to_geotiff`
      - Write a DataArray to a local path. Pass ``cog=True`` for a
        Cloud-optimized GeoTIFF layout. Pass ``allow_experimental_codecs=True``
@@ -229,7 +229,7 @@ the ambiguous-metadata family at once.
      - ``attrs['nodata']`` and ``attrs['nodatavals']`` disagree on
        write.
      - Resolve in caller code; the writer will not pick one silently.
-   * - ``VRTUnsupportedError``
+   * - :class:`~xrspatial.geotiff.VRTUnsupportedError`
      - The parsed VRT declares a feature the read pipeline does not
        honour (CRS / dtype / band / nodata / transform / pixel-size /
        window / resampling mismatch).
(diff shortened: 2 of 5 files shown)


brendancol (commented):
## PR Review: Export VRTUnsupportedError, CloudSizeLimitError, PixelSafetyLimitError (#3265)

### Blockers
None.

### Suggestions
- [ ] The exceptions are now public but nothing in the `open_geotiff` docstring points at them. The `max_pixels` param doc (xrspatial/geotiff/__init__.py) describes the cap without naming the exception, and `max_cloud_bytes` likewise. Now that the classes are catchable from the package, naming them at the trigger points (`PixelSafetyLimitError` for `max_pixels`, `CloudSizeLimitError` for `max_cloud_bytes`) closes the discoverability loop this PR opens. Same applies to the "Remote-read safety limits" bullets in `docs/source/reference/geotiff.rst:413-419`, though the docstring is the higher-traffic spot.

### Nits
None.

### What looks good
- The export is additive and the identity test pins `g.X is _module.X`, so the private import paths used by `tests/integration/test_sidecar.py` and the vrt tests keep working.
- `__all__` insertions keep the list sorted, and the existing surface-pinning tests (`test_api_consolidation.py`: lowercase entries are exactly `open_geotiff`/`to_geotiff`; `test_metadata.py` error-export checks) still pass.
- The `geotiff_safe_io.rst` change removes the awkward "Both error classes live in `_errors`" steer, which was half-wrong anyway (`UnsupportedGeoTIFFFeatureError` was already exported).
- The functional `max_pixels` test exercises a real public read path rather than just constructing the exception.

### Checklist
- Algorithm / NaN / dask / GPU: n/a, namespace + docs change
- Backend consistency: n/a
- Edge cases: identity with private modules and base-class catch are pinned
- Benchmarks: n/a
- README matrix: n/a (no new functions)
- Docstrings: see suggestion above

brendancol (commented):
Follow-up on the review above:

- Suggestion (exceptions not named at their trigger points): fixed in af3c8a66. The `max_pixels` param doc now names `PixelSafetyLimitError` and the `max_cloud_bytes` doc names `CloudSizeLimitError` (both with the :class: role and the ValueError-subclass note), and the `max_cloud_bytes` bullet in `docs/source/reference/geotiff.rst` got the same cross-reference. Signature-contract and docstring-parity suites still pass (224 tests).

No open findings remain.

brendancol (commented):
CI follow-up: the first pytest run failed on the frozen-__all__ release gate (test_features.py::TestPublicAPI::test_all_lists_supported_functions), which pins the exact public surface so additions are deliberate. Extended the frozen set with the three exported exceptions in 991c6345, with comments matching the file's per-name style. Full geotiff suite passes locally (6407 tests) and all workflows are green on the new head.

PR #3278, opened 2026-06-12
Promote reproject() to the dask path when the output is large

### LINKED ISSUE #3267: reproject() in-memory path peaks at ~7x output size; no dask promotion for large outputs
Found during a performance sweep of `xrspatial/reproject`.

`reproject()` auto-promotes to the dask path based on the input size only (`xrspatial/reproject/__init__.py` around line 969: `nbytes = src_shape[0] * src_shape[1] * data.dtype.itemsize`). The output size never enters the decision. A small input upsampled to a large output stays on the in-memory numpy path, which holds about seven output-sized float64 arrays at once:

- `out_x_flat` / `out_y_flat` and `src_x_flat` / `src_y_flat` inside `try_numba_transform`
- `src_row_px` / `src_col_px` in `_reproject_chunk_numpy`
- `local_row` / `local_col`
- the result array

Measured: a 256x256 input (0.5 MB) reprojected to 12000x12000 (1.15 GB output array) peaks at 8.4 GB RSS, 7.3x the output array. An output around 2.5e8 pixels passes the `_MAX_OUTPUT_PIXELS` guard (1e9) but OOMs a 16 GB machine, even with dask installed and able to do the same job in O(chunk) memory.

`merge()` already handles this case: it computes the grid with `lazy_output=True`, promotes to the dask path when the output exceeds a 512 MB threshold, and re-applies the pixel guard only when staying in-memory (`__init__.py` ~2417-2436). `reproject()` should make the same decision from its output grid.

Repro:

```python
import numpy as np, xarray as xr, resource

r = xr.DataArray(
    np.random.random((256, 256)), dims=['y', 'x'],
    coords={'y': np.linspace(50, 40, 256), 'x': np.linspace(-10, 0, 256)},
    attrs={'crs': 'EPSG:4326'},
)
reproject(r, 'EPSG:3857', width=12000, height=12000)
print(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1e6, 'GB peak')  # 8.4
```

Sweep context: the dask path itself is fine (lazy, single blockwise layer, 2 tasks per output chunk, source windows capped at 64 Mpix). The hazard is only in the eager numpy path.

brendancol:
Closes #3267

- `reproject()` now computes the output grid before choosing an execution path and promotes numpy inputs to the lazy dask path when either the input or the output working set exceeds 512 MB. Previously only the input size counted, so a small input upsampled to a large output ran eagerly and peaked at 7.3x the output array (measured 8.4 GB RSS for a 1.15 GB output; 0.5 GB after this change).
- The `_MAX_OUTPUT_PIXELS` guard is re-applied only for paths that materialize the whole output (in-memory and streaming), matching the pattern `merge()` uses since #3048. A numpy input whose output exceeds 1e9 pixels now reprojects lazily instead of raising.
- Fixes a latent bug in the dask wrap: 3-D inputs were chunked with a 2-D chunks tuple, which raised ValueError whenever a multi-band raster crossed the promotion threshold.
- Includes the performance-sweep state CSV update for the reproject module.

Backends: behavior change is on the numpy input path only (result may now be dask-backed for large outputs). cupy, dask+numpy, and dask+cupy paths are unchanged.

Test plan:
- [x] New `TestReprojectOutputSizePromotion`: promotion threshold behavior, eager/lazy parity (2-D and 3-D), >1e9-pixel numpy input stays lazy, guard still raises with promotion disabled
- [x] Full reproject suite: 506 passed (includes streaming, coverage, cupy gate files)
- [x] Memory probe: 256x256 input to 12000x12000 output, peak RSS 8.4 GB before, 0.5 GB after


brendancol (commented):
## PR Review: Promote reproject() to the dask path when the output is large

None found.

- [ ] The promotion still excludes cupy inputs (`if not is_dask and not is_cupy`, xrspatial/reproject/__init__.py:998), so an in-memory cupy raster with a huge output keeps allocating the full working set on the GPU. That matches the old behavior, and `_reproject_dask_cupy` has its own VRAM check, but wrapping cupy inputs the same way would give them the chunked fallback for free. Worth a follow-up issue if not done here.

- [ ] `out_nbytes` counts one float64 output array while the comment above it talks about "several output-sized temporaries" (xrspatial/reproject/__init__.py:1001). The 512 MB budget effectively allows a ~3.5 GB working set. The threshold is a judgment call either way, but a sentence in the comment saying the budget is per-array (so peak is ~7x that) would stop the next reader from thinking the constant bounds total memory.

### What looks good
- The reorder (grid first, then path choice, then guard) mirrors what merge() already does for #3048, so the two entry points now make the same decision the same way.
- The 3-D chunks fix is real: the old input-size promotion would have crashed on any multi-band raster over 512 MB with the same ValueError the new test caught.
- Measured numbers in the PR body (8.4 GB -> 0.5 GB) match what the new tests codify, and the >1e9-pixel numpy case now behaves like the dask case instead of raising.
- Guard semantics are preserved for every materializing path: in-memory numpy/cupy and the streaming fallback still raise, and `test_inmemory_over_limit_still_raises` pins that with promotion disabled.

### Test coverage notes
- The streaming branch (dask unavailable) with an over-budget output is untested, but simulating a missing dask import is already handled in test_reproject_streaming_3101.py and the guard logic is shared, so the gap is acceptable.

### Checklist
- [x] Algorithm matches reference (merge() promotion pattern, #3048)
- [x] All implemented backends produce consistent results (eager/lazy parity tests, 2-D and 3-D)
- [x] NaN handling unchanged (equal_nan parity assertions)
- [x] Edge cases covered (3-D promotion, >1e9 px, promotion disabled)
- [x] No premature materialization introduced; result stays lazy
- [x] Benchmark not needed (no kernel change; path-selection only)
- [x] README matrix not applicable (no new function, no backend change)
- [x] Docstring updated with the promotion behavior

brendancol (commented):
## Follow-up review after c207be64

- Suggestion (cupy inputs skip output-size promotion): deferred to #3281. Promoting cupy inputs changes the result-type contract for in-memory cupy callers (cupy -> dask-of-cupy above the threshold) and needs GPU OOM testing, so it does not fit this PR.
- Nit (budget comment reads like a total-memory bound): fixed in c207be64. The constant's comment now states the comparison is per-array and that eager-path peak RSS can reach ~7x the budget.

Re-checked the diff after the follow-up commit: comment-only change, `TestReprojectOutputSizePromotion` still passes (5/5). No new findings.

PR #3279, opened 2026-06-12
Upload shared coordinate arrays once per chunk in multi-band cupy reproject

### LINKED ISSUE #3268: Multi-band cupy reproject re-uploads coordinate arrays once per band on the CPU-fallback transform path
Found during a performance sweep of `xrspatial/reproject` on a CUDA host.

When the cupy backend cannot use the on-device transform (CRS pair without a CUDA kernel, e.g. EPSG:28992, or `transform_precision=0`), `_reproject_chunk_cupy` computes `local_row` / `local_col` as numpy arrays and then loops over bands, calling `_resample_cupy_native` once per band (`xrspatial/reproject/__init__.py` ~633-667). `_resample_cupy_native` converts the coordinate arrays with `cp.asarray` on every call (`xrspatial/reproject/_interpolate.py` ~726-731), so the same two chunk-sized float64 arrays cross PCIe once per band.

Measured with `cp.asarray` instrumented: a 6-band 512x512 raster reprojected 4326 -> 28992 does 12 host-to-device coordinate uploads, 26.1 MB transferred where 4.4 MB is enough. The waste scales with band count times chunk area; an 8-band raster with 2048 chunks moves ~512 MB of redundant coordinate data per chunk.

Fix: convert `local_row` / `local_col` to cupy once before the band loop. `_resample_cupy_native` already accepts cupy inputs and skips the conversion (`cp.ascontiguousarray` on a contiguous device array is a no-op).

Not affected: the single-band path (one call per chunk) and the on-device CUDA transform path (coords are already cupy).

brendancol:
Closes #3268

- `_reproject_chunk_cupy` now converts `local_row` / `local_col` to the device once before the per-band resample loop. On the CPU transform fallback (CRS pair without a CUDA kernel, e.g. EPSG:28992, or `transform_precision=0`) these arrive as numpy arrays, and `_resample_cupy_native` re-uploaded both per band: 12 host-to-device transfers (26.1 MB) for a 6-band 512x512 chunk where 2 (4.4 MB) suffice. The waste scaled with band count times chunk area.
- No behavior change for single-band chunks or the on-device CUDA transform path (coordinates are already cupy there).

Backends: cupy and dask+cupy (multi-band, CPU-fallback transform). numpy and dask+numpy untouched.

Test plan:
- [x] New `TestCupyMultibandCoordUpload`: instrumented `cp.asarray` confirms 2 coordinate uploads for 6 bands (was 12), plus numpy/cupy parity on the 4326 -> 28992 multi-band fallback
- [x] Ran on a CUDA host: full test_reproject.py suite, 464 passed

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -632,6 +632,14 @@ def _reproject_chunk_cupy(
     # signature mismatch (#2027).
     if window.ndim == 3:
         n_bands = window.shape[2]
+        # The coordinate arrays are shared by every band. On the CPU
+        # transform fallback they arrive as numpy; convert them to the
+        # device once here, otherwise _resample_cupy_native re-uploads
+        # the same two chunk-sized arrays on every band iteration (#3268).
+        if not isinstance(local_row, cp.ndarray):
+            local_row = cp.asarray(local_row)
+        if not isinstance(local_col, cp.ndarray):
+            local_col = cp.asarray(local_col)
         bands = []
         for b in range(n_bands):
             band_data = window[:, :, b].astype(cp.float64)
--- a/xrspatial/tests/test_reproject.py
+++ b/xrspatial/tests/test_reproject.py
@@ -2207,7 +2207,12 @@ class TestCupyMultibandCoordUpload:
         coord_uploads = [s for s in uploads if s == out_shape]
         # One upload each for the shared row and column coordinate arrays,
         # regardless of the band count. Before #3268 this was 2 * n_bands.
-        assert len(coord_uploads) == 2
+        # All recorded 2-D upload shapes are included in the failure
+        # message so a future failure shows what extra upload appeared.
+        assert len(coord_uploads) == 2, (
+            f"expected 2 coordinate uploads of shape {out_shape}, got "
+            f"{len(coord_uploads)}; all 2-D ndarray uploads: {uploads}"
+        )
 
     def test_multiband_fallback_matches_numpy(self):
         from xrspatial.reproject import reproject
--- a/xrspatial/tests/test_reproject.py
+++ b/xrspatial/tests/test_reproject.py
@@ -2163,6 +2163,75 @@ class TestCupyPyprojFallbackParity:
                                    rtol=1e-6, atol=1e-6)
 
 
+@pytest.mark.skipif(not HAS_CUPY, reason="CuPy not installed")
+class TestCupyMultibandCoordUpload:
+    """Multi-band cupy reproject on the CPU-fallback transform path must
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Upload shared coordinate arrays once per chunk in multi-band cupy reproject

None found.

- [ ] `cp.ascontiguousarray(src_row_coords, dtype=cp.float64)` still runs inside `_resample_cupy_native` per band (xrspatial/reproject/_interpolate.py:730-731). For device arrays that is a no-op view check, so the cost is negligible, but if a caller ever hands in non-contiguous or float32 device coords it would still copy per band. Not worth changing in this PR; just noting the remaining per-band work is bounded and on-device.

- [ ] The upload-count test asserts an exact `== 2` (xrspatial/tests/test_reproject.py, TestCupyMultibandCoordUpload). If an unrelated change later adds a legitimate 2-D ndarray upload of the same shape, the test fails with no hint which upload is which. Recording `id(a)` or asserting `<= 2` with a comment would make a future failure easier to diagnose. Exact pinning is also defensible as a stricter regression lock, so this is a judgment call.

### What looks good
- The fix is the minimal correct hoist: two isinstance checks before the loop, no behavior change for the on-device transform path or single-band chunks.
- The regression test instruments the actual transfer mechanism instead of timing, so it is deterministic and cheap (2.3 s on a GPU host including the parity case).
- The parity test pins the 4326 -> 28992 fallback against the numpy backend with the same tolerance pattern the existing #2620 tests use.
- dask+cupy gets the fix for free since the chunked path goes through the same `_reproject_chunk_cupy`.

### Checklist
- [x] Algorithm unchanged (same kernels, same values; parity test confirms)
- [x] Backends consistent (numpy/cupy parity on the fallback path)
- [x] NaN handling untouched (per-band sentinel masking unchanged)
- [x] Edge cases: single-band and CUDA-transform paths intentionally untouched
- [x] No premature materialization; transfer count drops from 2N to 2
- [x] Benchmark not needed (covered by measured upload counts in the test)

brendancol (commented):
## Follow-up review after 7878f9f9

- Suggestion (per-band `cp.ascontiguousarray` remains): dismissed. On contiguous float64 device arrays, which is what the hoisted conversion produces, the call returns the same array without copying. The note stands as documentation of bounded on-device work, not a defect.
- Nit (bare `== 2` assertion hard to diagnose): fixed in 7878f9f9. The assertion keeps the exact count but now reports every recorded 2-D upload shape in the failure message.

Re-checked the diff after the follow-up commit: test-message-only change, `TestCupyMultibandCoordUpload` still passes on the GPU host (2/2). No new findings.

PR #3280, opened 2026-06-12
Fix LAEA inverse rq normalization and authalic inverse series

### LINKED ISSUE #3274: reproject: LAEA inverse off by up to 2.6 km, AEA/CEA inverse off by ~5 m vs PROJ
**Describe the bug**

Two numerical bugs in the inverse projection kernels in `xrspatial/reproject/_projections.py` (and their CUDA twins in `_projections_cuda.py`) make reprojections into LAEA, AEA, and CEA targets sample the wrong source location.

1. `_laea_inv_point` (oblique/equatorial modes) normalizes the input by `x / (a * xmf)` and `y / (a * ymf)`, which strips the `rq` factor that PROJ keeps in `rho`. The angular distance is then computed as `sce = 2 * asin(0.5 * rho / rq)`, dividing by `rq` a second time. The result is a radial error of about 0.11% of the distance from the projection origin. For EPSG:3035 over Europe that is up to **2.6 km** versus pyproj. The CUDA kernel `_d_laea_inv` has the same formula.

2. `_authalic_apa` claims sub-centimetre accuracy from a 6-term inverse authalic latitude series, but the coefficients do not invert `_authalic_q`. Measured against exact numerical inversion of `q(phi)`, the series has a max error of **4.79 m** (7.5e-7 rad). PROJ's standard 3-term series (517/5040, 23/360 + 251/3780, 761/45360) lands at 1.6 mm. The wrong coefficient is visible at leading order in term 2: e4*17/360 vs PROJ's e4*23/360. This series feeds the AEA and CEA inverses (`_aea_inv_point`, `_cea_inv_point`, plus the inline 3-term copy in `_laea_inv_point` and the CUDA `_d_authalic_inv` call sites).

Measured parity vs pyproj (random points in each CRS's region of use):

| pair | direction | max error |
|---|---|---|
| EPSG:3035 -> 4326 | laea inverse | 2643 m |
| EPSG:5070 -> 4326 | aea inverse | 4.86 m |
| EPSG:6933 -> 4326 | cea inverse | 4.81 m |
| 4326 -> each of the above | forward | <= 1.1e-5 m |

The inverse direction is what `reproject()` uses to build the sampling grid when the *target* CRS is LAEA/AEA/CEA, so `reproject(raster_4326, 'EPSG:3035')` georeferences the output wrong by the amounts above. With the LAEA normalization fixed and PROJ's series coefficients in place, the 3035 inverse parity drops to ~1 mm.

**Expected behavior**

Inverse fast-path kernels should match pyproj within the same tolerance the forward kernels already achieve (sub-mm to ~1e-5 m).

**Reproduction**

```python
import numpy as np, pyproj
from xrspatial.reproject._projections import transform_points

tr = pyproj.Transformer.from_crs(4326, 3035, always_xy=True)
lons = np.random.uniform(-10, 30, 500); lats = np.random.uniform(35, 70, 500)
px, py = tr.transform(lons, lats)
tx, ty = transform_points(pyproj.CRS(3035), pyproj.CRS(4326), np.asarray(px), np.asarray(py))
inv = pyproj.Transformer.from_crs(3035, 4326, always_xy=True)
rx, ry = inv.transform(px, py)
err = np.hypot((tx - rx) * np.cos(np.radians(ry)), ty - ry) * 111320
print(err.max())   # ~2600 m
```

**Additional context**

Found by /sweep-accuracy against the reproject module. End-to-end check: `reproject(..., 'EPSG:3035')` with the default fast path vs `transform_precision=0` (exact pyproj) differs accordingly; numpy and cupy backends agree with each other because both kernels carry the same bugs.

brendancol:
Closes #3274

- `_laea_inv_point` (oblique/equatorial) stripped the `rq` factor in its input normalization and then divided by `rq` again in the angular distance, inflating distances by ~0.11% of the distance from the projection origin. Keep `rq` in `rho` so `sce = 2*asin(0.5*rho/rq)` divides by it exactly once, matching PROJ's `laea.cpp`. Same change in the CUDA kernel `_d_laea_inv`.
- `_authalic_apa` coefficients did not invert `_authalic_q` (wrong at leading order in the second term, 17/360 vs 23/360). Replaced with PROJ's `pj_authset` / Snyder eq. 3-18 coefficients. This feeds the AEA, CEA, and LAEA inverses on CPU and CUDA; array shape and kernel signatures are unchanged (terms 4-6 are now zero).

Parity vs pyproj after the fix: EPSG:3035 inverse 1.0 mm (was 2.6 km), EPSG:6933 inverse 1.6 mm (was 4.8 m), WGS84 AEA inverse 1.5 mm (was ~4.9 m). Forward kernels were already at the 1e-5 m level and are unchanged. The residual ~2 m for EPSG:5070 is pyproj's WGS84->NAD83 step, which the fast path skips by design.

Backends: numpy and dask+numpy share the CPU kernels; cupy and dask+cupy use the fixed CUDA kernel. CUDA tests ran on a real GPU.

Test plan:
- [x] New `test_reproject_inverse_kernels_3274.py`: series round-trip vs `_authalic_q`, coefficient pin, inverse parity vs pyproj for LAEA/CEA/AEA (oblique and polar), forward regression pin, end-to-end fast path vs `transform_precision=0`, CUDA-vs-pyproj and CUDA-vs-CPU parity
- [x] Full reproject suite: 607 passed


brendancol (commented):
## PR Review: Fix LAEA inverse rq normalization and authalic inverse series

None.

- [ ] `_projections.py` `_laea_inv_point`: `sce = 2.0 * math.asin(0.5 * rho / rq)` has no clamp on the asin argument, unlike the `ratio` clamps elsewhere in the same function. With the corrected (larger by 1/rq) `rho`, inputs beyond the projection disc reach the asin domain edge slightly sooner and produce NaN instead of a clamped pole value. NaN for out-of-disc input is defensible (PROJ errors there) and the downstream finite-filters handle it, but a one-line clamp would match the file's own conventions. Same spot in `_d_laea_inv` (`_projections_cuda.py`).
- [ ] `test_reproject_inverse_kernels_3274.py` covers OBLIQ (3035) and N-polar LAEA but not the EQUIT branch (`lat_0=0`), which shares the fixed code path. I ran an EQUIT parity check locally (`+proj=laea +lat_0=0 +lon_0=20`, 500 points): max error 1.6 mm, so the fix holds there too; a pinned test would keep it that way.

- [ ] `_authalic_inv` still evaluates five series terms although terms 4-5 are now zero. The docstring explains the signature-stability tradeoff, so this is fine to leave; mentioning the dead terms could alternatively be dropped from the loop in a later cleanup.

### What looks good
- The rq fix matches PROJ's laea.cpp normalization exactly (`x /= dd; y *= dd` with rho keeping rq), and the comment documents why the factor must stay in rho.
- The replacement coefficients are pinned against the exact `_authalic_q` numeric inverse (round-trip test at <5e-9 rad) rather than against pyproj alone, so the test distinguishes a series bug from a datum-step difference.
- The AEA test deliberately uses a WGS84 proj-string CRS instead of EPSG:5070, with a comment explaining the ~1-2 m NAD83 step that pyproj inserts and the fast path skips. That avoids a flaky tolerance.
- CUDA kernels updated in lockstep and verified two ways: against pyproj per pixel and against the CPU fast path at 1e-9.
- Kernel signatures and the `_APA` array shape are unchanged, so no call-site churn on either backend.

### Checklist
- [x] Algorithm matches reference (PROJ laea.cpp, pj_authset/Snyder 3-18)
- [x] All implemented backends produce consistent results (CPU/CUDA parity test)
- [x] NaN handling unchanged; finite-filtering downstream unaffected
- [x] Edge cases: polar branch pinned, forward path pinned, end-to-end vs exact path
- [ ] EQUIT LAEA branch untested (verified manually, see suggestion)
- [x] No materialization or copy concerns (scalar kernel math only)
- [x] Benchmark not needed (no perf-relevant change)
- [x] README matrix not applicable (no new function)
- [x] Docstrings updated with measured error bounds and issue reference

brendancol (commented):
Follow-up review after the suggestion commit: both suggestions are addressed. The asin argument is now clamped in both the CPU kernel and the CUDA twin with a comment tying it to the existing ratio clamps, and the EQUIT-mode parity test pins the lat_0=0 branch at the same 5 cm tolerance (measured 1.6 mm). The nit about the two zero series terms stays as documented signature-stability. 11 regression tests and the full 460-test reproject suite pass. No remaining findings.

PR #3282, opened 2026-06-12
Reject non-WGS84 ellipsoids in the projection fast-path dispatch

### LINKED ISSUE #3275: reproject: spherical-ellipsoid CRSes (MODIS sinusoidal) take the WGS84 fast path, ~19-24 km error
**Describe the bug**

The numba/CUDA fast-path dispatch in `xrspatial/reproject/_projections.py` applies WGS84 ellipsoid constants to CRSes that are not on a WGS84-compatible ellipsoid. Two holes:

1. `_is_wgs84_compatible_ellipsoid` only looks at the `ellps` and `datum` keys of `crs.to_dict()`, treating empty values as WGS84. A spherical CRS defined by explicit axes is normalized by PROJ to an `R` key (no `ellps`), so it passes the check. The standard MODIS sinusoidal grid is exactly this case:

   ```python
   import pyproj
   from xrspatial.reproject._projections import try_numba_transform
   modis = pyproj.CRS("+proj=sinu +lon_0=0 +x_0=0 +y_0=0 +a=6371007.181 +b=6371007.181 +units=m +no_defs")
   # try_numba_transform(crs4326, modis, ...) takes the fast path
   ```

   The WGS84 meridional-arc kernel then runs against a sphere. Measured against pyproj: **18.9 km** max error at mid latitudes. Any `reproject()` between EPSG:4326 and a MODIS sinusoidal grid silently returns data shifted by that much. The CUDA dispatch (`try_cuda_transform`) shares the helper, so the GPU path is wrong the same way.

2. `_aea_params` and `_cea_params` have no ellipsoid check at all (every other param extractor calls `_is_wgs84_compatible_ellipsoid`). A spherical AEA or CEA definition runs through the WGS84 kernels:

   ```
   +proj=aea +lat_1=29.5 +lat_2=45.5 +lat_0=23 +lon_0=-96 +a=6370997 +b=6370997   -> up to 23.8 km error
   +proj=cea +lat_ts=30 +a=6371228 +b=6371228                                      -> up to 23.8 km error
   ```

   These slip past the datum-shift guard because `ellps=sphere` is not in `_DATUM_PARAMS`.

**Expected behavior**

CRS pairs the kernels cannot represent should return `None` from `try_numba_transform` / `transform_points` / `try_cuda_transform` and fall back to pyproj, the same way OSGB36/NAD27 pairs already do (#2651).

**Additional context**

Found by /sweep-accuracy against the reproject module. The fix needs `_is_wgs84_compatible_ellipsoid` to reject `R`-defined spheres and explicit `a`/`b` axes that differ from WGS84, and `_aea_params`/`_cea_params` to call the guard. WGS84/GRS80 pairs must keep the fast path (GRS80 axes differ from WGS84 by ~1e-4 m, so an exact-match check is too strict).

brendancol:
Closes #3275

- `_is_wgs84_compatible_ellipsoid` now rejects CRSes whose ellipsoid is given as an explicit sphere radius (`R` key) or as semi-axes that do not match WGS84/GRS80 (0.5 m tolerance, which keeps GRS80 in since its `b` differs from WGS84 by ~1e-4 m). PROJ normalizes `+a=R +b=R` definitions to a bare `R` key with no `ellps`, which is exactly how the MODIS sinusoidal grid slipped through the old empty-string match and ran the WGS84 kernels ~18.9 km off.
- `_aea_params` and `_cea_params` now call the guard like every other param extractor. Sphere-based AEA/CEA definitions previously ran the WGS84 kernels with up to ~24 km error.

Rejected pairs return `None` from `transform_points` / `try_numba_transform` / `try_cuda_transform` and fall back to pyproj, the same way non-WGS84 datum pairs already do (#2651). After the change, `reproject(raster_4326, MODIS_SINU)` matches the `transform_precision=0` exact path to <0.03 of a +/-100 data range (was ~150).

One deliberate side effect: the helper now reports EPSG:3857 (a sphere by definition) as not WGS84-compatible. Web Mercator is dispatched by EPSG code and never consults the helper, so its fast path is unchanged; a regression test pins that.

Backends: the CPU and CUDA dispatchers share the helper and param extractors, so numpy, cupy, dask+numpy, and dask+cupy all pick up the guard. CUDA tests ran on a real GPU.

Test plan:
- [x] New `test_reproject_sphere_ellipsoid_guard_3275.py`: helper accepts WGS84/GRS80 (named and explicit-axes) and Helmert-table datums, rejects R-spheres and named spheres; MODIS sinu / sphere AEA / sphere CEA bail to pyproj; WGS84 pairs keep the fast path; end-to-end MODIS reproject vs exact path; CUDA guard active and CUDA WGS84 path intact
- [x] Full reproject suite: 607 passed


brendancol (commented):
## PR Review: Reject non-WGS84 ellipsoids in the projection fast-path dispatch

None.

- [ ] `test_reproject_sphere_ellipsoid_guard_3275.py` only exercises the sphere CRSes as the *target* of `transform_points` (and both directions for MODIS via `try_numba_transform`). The sphere-as-source direction for AEA/CEA goes through the same `_aea_params`/`_cea_params` call on the other branch of the dispatcher; I verified locally that `transform_points(sphere_aea, 4326, ...)` returns None, but a one-line assertion in `test_sphere_aea_no_fast_path` would pin it.

- [ ] The 0.5 m axis tolerance and the 0.01 `rf` tolerance are magic numbers justified only in the comment. Fine as is; hoisting them to named module constants would help if a third place ever needs them.

### What looks good
- The guard sits in the shared helper plus the two param extractors that skipped it, so the CPU and CUDA dispatchers (and all four array backends behind them) are covered by one change.
- Helmert-table datums (e.g. EPSG:27700 / Airy, ellps=intl) still report compatible and still bail through the separate datum-params guard, so the #2651 behavior is preserved and pinned by a test.
- The EPSG:3857 side effect is the right call: Web Mercator really is a sphere, its dispatch never consults the helper, and the regression test documents both halves of that.
- The end-to-end MODIS test asserts against the `transform_precision=0` exact path with a tolerance three orders of magnitude under the pre-fix error, so it cannot pass accidentally.
- GRS80 is kept in via tolerance rather than name-matching, which also accepts WGS84 given as explicit axes (tested).

### Checklist
- [x] Guard semantics match the documented dispatch contract (fall back to pyproj, like #2651)
- [x] CPU and CUDA dispatchers behave identically (shared helper; CUDA tests ran on GPU)
- [x] NaN handling not applicable (dispatch-only change)
- [x] Edge cases: R-spheres, named spheres, explicit WGS84 axes, Helmert datums, all WGS84 EPSG pairs
- [x] No materialization or copy concerns
- [x] Benchmark not needed
- [x] Docstring documents the PROJ R-normalization trap and the issue number

brendancol (commented):
Follow-up review: the sphere-as-source assertion is added to test_sphere_aea_no_fast_path, covering the other dispatch branch. The magic-number nit stays as a comment-level justification, which is fine for two call-free constants. 12 regression tests pass. No remaining findings.

PR #3283, opened 2026-06-12
Read ITRF Helmert scale as ppm to match PROJ's data files

### LINKED ISSUE #3276: itrf_transform: Helmert scale read as ppb but PROJ files publish ppm, ~2 cm error
**Describe the bug**

`_helmert14_point` in `xrspatial/reproject/_itrf.py` converts the scale parameter with

```python
s_e = 1.0 + (s + ds * dt) * 1e-9  # ppb -> scale factor
```

but the parameters come from PROJ's ITRF data files, and PROJ's `helmert` operation defines `+s` / `+ds` in **ppm**, not ppb. The file values are ppb magnitudes written in ppm units, e.g. `ITRF2014 -> ITRF2000` carries `+s=0.00212` (ppm) = 2.12 ppb. Multiplying by 1e-9 makes the scale term 1000x too small, which effectively drops it.

Against a pyproj 4D helmert pipeline built from the same parameters, `itrf_transform(-74.0, 40.7, 100.0, src='ITRF2014', tgt='ITRF2000', epoch=2024.0)` is off by about **23 mm in height**. With the factor corrected to 1e-6 the pure-Python port matches pyproj to better than 0.01 mm:

```
pyproj 4D:  -73.99999996763594  40.69999965079892  99.98748955968767
ppm interp: -73.99999996763596  40.699999650798915 99.98748955968767
ppb (code): -73.99999996763594  40.69999965149493  99.96420177258551
```

A 2-3 cm error is negligible for raster reprojection but this function's stated purpose is mm-level frame transforms ("relevant for precision geodesy", module docstring), so the result is wrong at the scale the API advertises.

**Expected behavior**

`itrf_transform` matches a PROJ helmert pipeline built from the same parameter file entry at the same epoch, to sub-mm.

**Additional context**

Found by /sweep-accuracy against the reproject module. Note pyproj only applies the rate terms when given 4D coordinates (`transform(lon, lat, h, epoch)`); a 3D transform silently uses the static parameters, which is worth remembering when writing the regression test.

brendancol:
Closes #3276

`_helmert14_point` converted the Helmert scale term with `* 1e-9` (ppb), but the parameters come straight out of PROJ's ITRF data files and PROJ's `+proj=helmert` defines `+s` / `+ds` in ppm. The published ppb-magnitude ITRF scale terms are written in the files as ppm values (`ITRF2014 -> ITRF2000` carries `+s=0.00212` ppm = 2.12 ppb), so the old factor shrank the scale contribution 1000x, leaving ~23 mm of height error for that pair at epoch 2024.

One-character fix (`1e-9` -> `1e-6`) plus docstring correction. After the change, `itrf_transform` matches a pyproj 4D `cart -> helmert -> cart` pipeline built from the same parameter entry to better than 0.01 mm.

Test note: pyproj only applies the rate terms (`dx`, `ds`, ...) when given 4D coordinates, so the reference pipeline in the tests passes the epoch as the 4th value. A 3D transform through the same pipeline silently uses the static parameters only.

Backends: CPU-only code path (numba kernel shared by all input types).

Test plan:
- [x] New `test_reproject_itrf_scale_3276.py`: parity vs pyproj 4D pipeline for three direct frame pairs (<0.5 mm), scale-term magnitude check (~23 mm radial effect present, ~5% tolerance), array/scalar agreement
- [x] Existing ITRF tests (17) and full `test_reproject.py` (460) pass


brendancol (commented):
## PR Review: Read ITRF Helmert scale as ppm to match PROJ's data files

None.

- [ ] The new tests only cover direct file entries (`_direct_params` skips reversed lookups). The negated-parameter reverse path in `_find_transform` also carries the scale term; I checked locally that ITRF2000 -> ITRF2014 (a reversed lookup) matches a pyproj `+inv +proj=helmert` pipeline to <1e-5 mm, so the negation is sound, but a test pinning one reversed pair would protect it.

- [ ] `test_scale_term_not_dropped` hardcodes R = 6.37e6 m for the expected radial effect. The true geocentric radius at the test point is ~6368 km, comfortably inside the 5% tolerance, so this is fine; a comment noting the slack comes from the radius approximation would make the tolerance choice obvious.

### What looks good
- The fix is the minimal correct change (one factor plus an honest docstring), and the unit claim is verified against the authoritative source: a pyproj 4D `cart -> helmert -> cart` pipeline built from the same file entry agrees to <0.01 mm across three direct frame pairs.
- The test helper builds the reference pipeline from the module's own parsed parameters rather than hardcoded values, so it keeps working when PROJ data files update.
- `test_scale_term_not_dropped` is a sharp regression: with the old ppb factor the measured scale contribution collapses from ~23 mm to ~23 um and the approx(rel=0.05) assertion fails by three orders of magnitude.
- The 4D-vs-3D pyproj footgun (rates silently dropped without a coordinate epoch) is documented in both the PR body and the test docstring.

### Checklist
- [x] Units match the reference (PROJ +proj=helmert: +s/+ds in ppm)
- [x] Backends: single CPU numba kernel shared by all input types
- [x] NaN handling unchanged (epoch finiteness already validated upstream)
- [x] Edge cases: skip guards for missing PROJ data files and near-zero scale terms
- [ ] Reversed-lookup pair untested (verified manually, see suggestion)
- [x] No performance concerns
- [x] Docstring corrected alongside the code

brendancol (commented):
Follow-up review: the reversed-lookup pair (ITRF2000 -> ITRF2014, negated parameters) is now pinned against a pyproj inverse-helmert pipeline at <0.5 mm, and the radius approximation behind the 5% tolerance is documented at the assertion. 6 regression tests pass. No remaining findings.

PR #3289, opened 2026-06-12
Name kriging singular-fallback variance '{name}_variance'

### LINKED ISSUE #3285: kriging() singular-matrix fallback returns variance DataArray without the _variance name suffix
**Describe the bug**

`kriging(..., return_variance=True)` names the variance DataArray `f'{name}_variance'` on the normal path (`xrspatial/interpolate/_kriging.py` line 523). But when the kriging matrix is singular even after regularisation, `_build_kriging_matrix` returns `None` and the fallback returns `prediction, prediction.copy()` (line 499). `DataArray.copy()` keeps the prediction's name, so the variance comes back named `'krig'` instead of `'krig_variance'`.

Reproduce by forcing the singular path:

```python
import xrspatial.interpolate._kriging as kr

kr._build_kriging_matrix = lambda *a, **k: None  # force singular fallback

x = y = z = np.arange(5.0)
template = xr.DataArray(np.zeros((10, 10)), dims=['y', 'x'],
                        coords={'y': np.arange(10.0), 'x': np.arange(10.0)})
pred, var = kriging(x, y, z, template, return_variance=True, name='krig')
print(pred.name, var.name)  # krig krig  -- expected: krig krig_variance
```

**Expected behavior**

The variance DataArray is named `'{name}_variance'` on every code path. With both arrays sharing a name, anything that keys on `.name` (`xr.merge`, building a Dataset from the pair) silently overwrites the prediction with the variance.

**Additional context**

Found during an API consistency sweep of the interpolate subpackage. One-line fix in the fallback branch, plus a regression test covering the singular path.

brendancol:
Closes #3285

- `kriging(..., return_variance=True)` has a fallback for the case where the kriging matrix stays singular after regularisation: it returns NaN rasters via `prediction, prediction.copy()`. The copy kept the prediction's name, so both arrays came back named `'{name}'` while the normal path names the variance `'{name}_variance'`. Anything keying on `.name` (`xr.merge`, building a Dataset from the pair) collapsed the two into one variable.
- The fallback now renames the copy to `f'{name}_variance'`, matching the normal path.
- Also pins the names on the normal path in the existing `test_return_variance`.

Backend coverage: the fallback runs before backend dispatch, so one code path covers numpy, cupy, dask+numpy, and dask+cupy. Variance names on the normal path were verified on numpy and cupy.

Test plan:
- [x] New `test_return_variance_name_on_singular_fallback` forces the singular path via monkeypatch and checks both names plus NaN contents
- [x] `pytest xrspatial/tests/test_interpolation.py` -- 67 passed


brendancol (commented):
## PR Review: Name kriging singular-fallback variance '{name}_variance'

- [ ] `.claude/sweep-api-consistency-state.csv`: the rewrite flipped every line ending from LF to CRLF (`csv.DictWriter` defaults to `\r\n` when the file is opened with `newline=''`). The diff touches all 13 rows instead of adding one, and since the state file merges as plain text (#2754), this will whole-file-conflict with every other sweep branch that commits a row. Rewrite with `lineterminator='\n'` so the diff is just the new `interpolate` row.

- none

- none

### What looks good
- The fix is the minimal correct change: `_kriging.py:499` now renames the copied DataArray to `f'{name}_variance'`, matching the normal path at line 527.
- `test_return_variance_name_on_singular_fallback` forces the singular path by monkeypatching `_build_kriging_matrix` (which `kriging()` resolves as a module global, so the patch takes effect) and checks both names plus the NaN contents.
- The existing `test_return_variance` now pins the default names on the normal path, so a regression in either path gets caught.

### Checklist
- [x] Algorithm matches reference/paper (naming-only change, no numerics touched)
- [x] All implemented backends produce consistent results (fallback runs before backend dispatch)
- [x] NaN handling is correct (fallback still returns all-NaN rasters; test asserts it)
- [x] Edge cases are covered by tests (singular path was previously untested)
- [x] Benchmark exists or is not needed (n/a for a name fix)
- [x] Docstrings present and accurate (kriging docstring already describes the tuple return)
- [ ] State CSV line endings (see blocker)

brendancol (commented):
## Follow-up review after blocker fix

The CSV blocker from the first review is resolved: the state file is now main's bytes plus a single inserted `interpolate` row (1 insertion, 0 deletions against main), so other rows keep their existing line endings (two rows on main end with CRLF; they are untouched). No whole-file conflict risk for sibling sweep branches.

The code and test changes are unchanged from the first review and still look correct. `pytest xrspatial/tests/test_interpolation.py` passes (67) after the rebase.

No remaining findings.

Issue #3256, opened 2026-06-12
Extend xrspatial.<tool> dask task naming to the remaining modules


brendancol:
#3250 / PR #3255 added `name='xrspatial.<tool>'` to the dask `map_blocks`/`map_overlap` calls in slope, aspect, hillshade, curvature, convolution, bilateral, terrain, preview, surface_distance, flow_direction_mfd, and flow_length_mfd, so `.visualize()` and the distributed dashboard label tasks by tool.

The issue underestimated the sweep: about 260 call sites remain across roughly 60 modules. The largest by call-site count:

- flood.py (25), multispectral.py (22), fire.py (14), zonal.py (12), focal.py (9), classify.py (7)
- proximity.py, cost_distance.py, reproject/, interpolate/, mcda/, diffusion.py, dasymetric.py, emerging_hotspots.py
- the rest of hydro/ (d8/dinf variants, watershed, stream_*, hand, fill, sink, twi, basin, snap_pour_point)
- terrain_metrics.py, sky_view_factor.py, morphology.py, mahalanobis.py, glcm.py, resample.py, worley.py, perlin.py

focal.py is the highest-traffic gap (focal.py:210, 220, 565, 595, 1247, 1589-1631).

The change is mechanical: add `name='xrspatial.<public function>'` to each tool-level call, leave plumbing calls (dtype conversion, cpu/gpu transfer) unnamed, and extend `xrspatial/tests/test_dask_task_names.py` with one prefix assertion per tool. Where one public function builds several layers that must stay distinct (see the band suffix in flow_direction_mfd), suffix the name. `name=` appends dask's hash on every supported dask version (verified back to 2023.11.0 in #3255), so same-name calls cannot collide.

PR #3294, opened 2026-06-12
Keep template backend and variance name in kriging singular-matrix fallback

### LINKED ISSUE #3288: kriging() singular-matrix fallback ignores template backend and misnames the variance raster
**Describe the bug**

When the ordinary-kriging matrix is singular even after regularisation, `kriging()` takes a fallback path (`xrspatial/interpolate/_kriging.py`, the `K_inv is None` branch) that returns an all-NaN raster. Two problems with that branch:

1. The fallback builds its data with `np.full(template.shape, np.nan)`, so the result is numpy-backed no matter what backend the template uses. A cupy, dask+numpy, or dask+cupy template gets an eager numpy result. For a large chunked dask template this also materialises the full grid in host memory, which is exactly what the dask path of the memory guard promises not to do.

2. With `return_variance=True` the branch returns `prediction.copy()` as the variance, so the variance DataArray keeps the prediction's name (`'kriging'`). The normal path names it `f'{name}_variance'`.

Confirmed on all four backends by monkeypatching `_build_kriging_matrix` to return `None`:

```
numpy:     pred name='kriging' data=numpy.ndarray var name='kriging' var data=numpy.ndarray
dask:      pred name='kriging' data=numpy.ndarray var name='kriging' var data=numpy.ndarray
cupy:      pred name='kriging' data=numpy.ndarray var name='kriging' var data=numpy.ndarray
dask_cupy: pred name='kriging' data=numpy.ndarray var name='kriging' var data=numpy.ndarray
```

**Expected behavior**

The fallback should match the normal path: result data on the same backend as the template (lazy for dask templates, cupy-backed for GPU templates), and the variance raster named `'<name>_variance'`.

**Additional context**

Found by the metadata propagation sweep. The branch is rare in practice (`np.linalg.inv` has to fail twice), but it is reachable and warns the user, so its output should follow the same metadata contract as the normal path. Coords, dims, and attrs are already propagated correctly.

brendancol:
Closes #3288

- The `K_inv is None` fallback in `kriging()` now builds its all-NaN result on the template's backend: `map_blocks` + `np.full_like` for dask templates (stays lazy, keeps chunks and the cupy meta on dask+cupy) and `np.full_like` otherwise (dispatches to cupy for GPU templates). Previously the fallback always returned an eager numpy array, and for a large chunked dask template it materialised the whole grid in host memory.
- With `return_variance=True` the fallback's variance raster is now named `<name>_variance`, matching the normal path. It used to inherit the prediction's name.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy. All four verified locally on a GPU host.

Test plan:
- [x] 4 new tests in `TestKriging` force the fallback via monkeypatch and check name, backend type, chunks, dtype, attrs, and NaN content per backend
- [x] `pytest xrspatial/tests/test_interpolation.py`: 70 passed (GPU tests included)


brendancol (commented):
## PR Review: Keep template backend and variance name in kriging singular-matrix fallback

None.

None.

- [ ] `xrspatial/interpolate/_kriging.py` (new `map_blocks` call in the `K_inv is None` branch): the lambda gives the dask task a `lambda-<hash>` name instead of an `xrspatial.kriging`-style name per the #3255 convention. Every other `map_blocks` call in this module has the same gap, and #3256 already tracks the module-wide naming sweep, so fixing only this one call here would leave the module inconsistent either way. Fine to leave for #3256.

### What looks good
- The fix dispatches on the actual backend: `map_blocks` + `np.full_like` keeps dask templates lazy with their original chunks, and `np.full_like` dispatches to cupy for GPU templates. Verified in the new tests, including the dask+cupy meta check (`result.data._meta` is `cupy.ndarray`).
- The fallback was eager before this change even for chunked dask templates; now it never materialises the grid, which matches the memory-guard reasoning that skips the grid-sized term for dask (`_check_kriging_memory(is_dask=True)`).
- `variance` now carries `<name>_variance` like the normal path (line 523), and the test asserts it on all four backends.
- The monkeypatch in the tests targets the module global `_build_kriging_matrix`, which `kriging()` resolves at call time, so the forcing mechanism is sound.
- Coords, dims, attrs, and dtype assertions are explicit in `test_singular_fallback_metadata` rather than inferred.

### Checklist
- [x] Algorithm matches reference/paper (n/a, metadata-only fix)
- [x] NaN handling is correct (all-NaN fallback by design)
- [x] Edge cases are covered by tests (fallback forced per backend)
- [x] Dask chunk boundaries handled correctly (no overlap needed; chunks preserved)
- [x] No premature materialization (fallback is now lazy on dask; was eager before)
- [x] Benchmark exists or is not needed (degenerate path, not performance-relevant)
- [x] README feature matrix updated (n/a, no API or backend-support change)
- [x] Docstrings present and accurate (no public signature change)

PR #3297, opened 2026-06-12
Widen rasterize() dtype annotation to the accepted forms (#3291)

### LINKED ISSUE #3291: rasterize() dtype annotation rejects the dtype forms the function accepts
**Describe the bug**

`rasterize()` annotates `dtype` as `Optional[np.dtype]`, but the implementation accepts anything `np.dtype()` can parse: string names (`'int32'`), numpy scalar types (`np.int32`), and `np.dtype` instances. The value flows straight into `np.dtype(final_dtype)` and `astype(...)`, so all three forms work on every backend.

The annotation contradicts how the function is actually called. `dtype=np.int32` is the dominant form in the module's own tests (9 uses in `test_rasterize.py` alone, plus `np.float32`/`np.int64` elsewhere), and none of those pass a `np.dtype` instance. Under the current hint, mypy and pyright flag `rasterize(gdf, like=template, dtype=np.int32)` and `dtype='float32'` even though both work.

There is also sibling drift: `open_geotiff` hints the same parameter as `str | np.dtype | None`.

**Expected behavior**

The annotation covers the accepted forms, e.g. `Optional[Union[str, np.dtype, type]]`, and the docstring names them.

**Notes**

Annotation widening only, no runtime change, so no deprecation shim is needed. Found by the api-consistency sweep (Cat 3, docs/types disagree), 2026-06-12 pass on `xrspatial/rasterize.py:3155`.

brendancol:
Closes #3291.

- `dtype` was annotated `Optional[np.dtype]` but the implementation accepts anything `np.dtype()` parses: string names (`'int32'`), numpy scalar types (`np.int32`), and `np.dtype` instances. Widened to `Optional[Union[str, np.dtype, type]]`, matching how the function is actually called (`dtype=np.int32` is the dominant form in the existing tests) and lining up with `open_geotiff`'s `str | np.dtype | None`.
- Docstring now names the accepted forms.
- Annotation and docs only, no runtime change, so no deprecation shim.

Backends: the signature is shared across numpy / cupy / dask+numpy / dask+cupy; all four were smoke-tested with identical kwargs during the sweep and values matched.

Test plan:
- [x] `test_rasterize_dtype_annot_3291.py`: pins the hint args (str / np.dtype / type / None) and runtime acceptance of all three forms
- [x] `test_rasterize_signature_annot_2250.py` still passes (every param annotated)
- [x] `test_rasterize.py`: 215 passed, 2 skipped

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -3226,8 +3226,9 @@ def rasterize(
         Data type of the output array.  Accepts anything ``np.dtype()``
         can parse: an ``np.dtype`` instance, a dtype name string
         (``'int32'``), or a numpy scalar type (``np.int32``).  Defaults
-        to np.float64, or to the dtype of ``like`` if provided.  When this resolves to an
-        integer type, burn values are validated against the float64 safe
+        to np.float64, or to the dtype of ``like`` if provided.  When
+        this resolves to an integer type, burn values are validated
+        against the float64 safe
         integer range: the rasterizer computes in float64, so a value
         with magnitude above ``2**53 - 1`` cannot be cast back to an
         exact integer (e.g. ``2**53 + 1`` would land on ``2**53``).  Such
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -3152,7 +3152,7 @@ def rasterize(
     column: Optional[str] = None,
     columns: Optional[Sequence[str]] = None,
     fill: float = np.nan,
-    dtype: Optional[np.dtype] = None,
+    dtype: Optional[Union[str, np.dtype, type]] = None,
     all_touched: bool = False,
     gpu: bool = False,
     name: str = 'rasterize',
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Widen rasterize() dtype annotation to the accepted forms (#3291)

None. Annotation and docstring change only; no runtime behavior touched, and the existing 215-test rasterize suite plus the #2250 signature tests pass.

- [ ] xrspatial/rasterize.py:3229 -- the rewrapped docstring leaves an 89-char line ("...Defaults to np.float64, or to the dtype of ``like`` if provided.  When this resolves to an"). The rest of this docstring wraps at ~72. Rewrap so the paragraph reads cleanly and the diff doesn't introduce the longest line in the file.
- [ ] xrspatial/tests/test_rasterize_dtype_annot_3291.py:19 -- `pytest.importorskip("shapely")` sits at module level, so `test_dtype_annotation_accepts_str_dtype_and_type` gets skipped in a shapely-less env even though it only inspects type hints. That env is exactly where the annotation check earns its keep (rasterize imports fine without shapely). Move the importorskip/box import into the runtime test.

- [ ] The unused `shapely =` binding from importorskip can go once the import moves.

### What looks good
- `Optional[Union[str, np.dtype, type]]` covers the three forms the implementation parses via `np.dtype()`, and `type` is what actually admits `np.int32`-style scalar classes; the sibling `open_geotiff` hint (`str | np.dtype | None`) misses those, so this is strictly more accurate.
- The hint test asserts the Union args directly via `typing.get_type_hints`, so a future regression names the missing form.
- The runtime parametrization adds the first string-name dtype coverage in the rasterize tests.

### Checklist
- [x] Algorithm matches reference/paper (n/a -- annotation only)
- [x] All implemented backends produce consistent results (signature shared; smoke-tested numpy/cupy/dask/dask+cupy during the sweep)
- [x] Edge cases are covered by tests (three dtype forms pinned)
- [x] Benchmark exists or is not needed (no perf surface)
- [x] Docstrings present and accurate (modulo the long line above)

brendancol (commented):
Follow-up review after 5f394791: both suggestions and the nit are addressed. The docstring paragraph now wraps at the file's usual width (no line over 79 in the touched region), the annotation-only test runs without shapely, and the importorskip plus box import moved into the runtime test where they belong. Tests still pass (7 in the two annotation files). Nothing further from me.

PR #3300, opened 2026-06-12
Add interpolate edge-case and parameter test coverage

### LINKED ISSUE #3290: interpolate: untested code paths in idw, kriging, and spline
The test-coverage sweep audited `xrspatial/interpolate/` (test file `xrspatial/tests/test_interpolation.py`). Backend coverage is in good shape: all four backends are exercised for `idw`, `kriging`, and `spline`, with cross-backend equivalence asserted. But several code paths and edge cases have no test at all. I verified each one behaves correctly today by running it manually, so these are coverage gaps, not bugs. The point of adding tests is to keep them correct.

Untested paths:

1. `idw` `fill_value` zero-weight branch. The existing `test_fill_value` admits it only checks the parameter is accepted; the branch never runs. It can be triggered deterministically: a point at distance ~1e200 with `power=4` makes the weight underflow to 0, so the pixel gets `fill_value`. The branch exists in all four backends.
2. `idw` `power` is only ever tested at the default 2.0. Nothing asserts power actually changes the weighting. With points at distances 1 and 2 and z values 0 and 10, the expected value is `10 / (2**p + 1)`, so exact assertions are easy.
3. `spline` degenerate-geometry fallback (`_spline.py` lines 89-94): collinear points make the augmented system singular, `np.linalg.solve` raises, and the code falls back to `lstsq`. No test hits this. Three collinear points on the x axis recover the line exactly.
4. `kriging` duplicate input points: identical rows make K singular, triggering the `1e-10` regularisation retry in `_build_kriging_matrix` (lines 153-157). Untested.
5. `kriging` with all-equal z values: the experimental variogram is all zeros, which stresses the `curve_fit` initialisation. Output is the constant, but no test pins that.
6. Inf/-Inf inputs are never passed to any of the three functions. `validate_points` drops them via `isfinite`, but only the NaN case is tested.
7. `spline` and `kriging` never run end to end on a 1x1 template (`idw` does, incidentally). A 0-column template also works (returns an empty raster) but is untested.
8. Metadata: the `idw` metadata test skips attrs (the kriging and spline ones check them), and no test anywhere asserts non-default dim names like `lat`/`lon` propagate to the output.

Low-priority, no test planned: asv benchmark coverage and an explicit non-uniform cell spacing test.

Fix is test-only. If any added test exposes a source bug, that gets a separate issue.

brendancol:
Closes #3290

Test-only PR from the test-coverage sweep against `xrspatial/interpolate/`. No source changes; every path tested here was probed manually first and behaves correctly, so these tests pin current behavior rather than fix bugs.

- `idw`: the `fill_value` zero-weight branch now actually runs (weight underflow from a point at distance 1e200 with power=4), on numpy, dask+numpy, and cupy; non-default `power` values asserted against the exact closed form 10/(2^p+1); attrs preservation added to the metadata test.
- `spline`: collinear points (exactly singular TPS system -> lstsq fallback) recover the line; 1x1 template smoke test.
- `kriging`: duplicate input points, all-equal z (zero-variance variogram), a unit test that an exactly singular kriging matrix takes the 1e-10 regularisation retry in `_build_kriging_matrix`, and a 1x1 template.
- Validation: Inf/-Inf points are dropped like NaN (and all-Inf raises), `lat`/`lon` dim names propagate through all three functions, and a 0-column template returns an empty raster.

Backends: numpy throughout; the new fill_value test also covers dask+numpy and cupy. The cupy test ran and passed on a CUDA host (`pytest -k cupy`: 14 passed, 0 skipped).

Also updates `.claude/sweep-test-coverage-state.csv` with the interpolate row.

Test plan:
- [x] `pytest xrspatial/tests/test_interpolation.py`: 82 passed, 0 skipped on a CUDA host
- [x] flake8 on the test file introduces no new findings (7 pre-existing ones untouched)


brendancol (commented):
## PR Review: Add interpolate edge-case and parameter test coverage

Test-only PR; no source changes. Reviewed against the full file on the PR head branch.

None.

- [ ] `test_nondefault_dim_names_propagate[kriging]` (xrspatial/tests/test_interpolation.py, TestValidation) emits the "fewer than 3 non-empty lag bins" UserWarning: `_grid_points()` spans a 2x2 area, so with nlags=15 only one or two bins fill. Use wider-spread points for the kriging case (the `_spatial_data` layout works) or wrap that branch in `pytest.warns`. Right now it's just noise, but it becomes a failure if filterwarnings ever turns strict.

- [ ] The fill_value zero-weight tests cover numpy, dask+numpy, and cupy but not dask+cupy. That's fine since `_idw_dask_cupy` delegates per-chunk to `_idw_cupy`, but a one-line note in `test_cupy_fill_value_zero_weight`'s docstring would stop the next coverage sweep from flagging it.

### What looks good
- `test_power_changes_weighting` checks an exact closed form (10/(2^p+1)) instead of "result changed", and includes the default in the sweep so a regression in either direction shows up.
- The 1e200 underflow trick makes the previously unreachable fill_value branch deterministic, and the cupy variant ran on real hardware (verified, not just added).
- `test_singular_matrix_regularisation_retry` constructs exact singularity (all-zero variogram makes rows identical), so it pins the retry path without mocking.
- The collinear spline test asserts exact recovered values, not just "no exception".
- State CSV edit is a clean single-row append; other modules' rows untouched.

### Checklist
- [x] Algorithm matches reference (closed-form IDW oracle checked by hand)
- [x] All implemented backends produce consistent results (pre-existing parity tests; new branch coverage on 3 of 4 backends, 4th transitive)
- [x] NaN handling correct (Inf filtering now pinned alongside the existing NaN tests)
- [x] Edge cases covered (1x1, 0-column, collinear, duplicate, all-equal)
- [x] Dask chunk boundaries handled (single-chunk 1x1 case; multi-chunk parity pre-existing)
- [x] No premature materialization (test code only)
- [ ] Benchmark exists or is not needed -- no asv coverage for interpolate; documented as LOW in the issue, not added here
- [x] README feature matrix: N/A, no API change
- [x] Docstrings: every new test states what it pins and why

brendancol (commented):
Follow-up review of 4cf6d1f.

- The dim-name test now uses `TestKriging._spatial_data()`, and under the project's pytest config the kriging case runs warning-free (the remaining scipy OptimizeWarning is already filtered in setup.cfg). Verified: 82 passed, 2 warnings, both pre-existing.
- The cupy fill_value test docstring now records that dask+cupy is covered transitively via per-chunk delegation.

No new issues. Cross-class use of the `_spatial_data` staticmethod from TestValidation is a little unusual but fine for test code; not worth restructuring.

PR #3301, opened 2026-06-12
polygonize: raise on uint32 region overflow in cupy integer backend (#3292)

### LINKED ISSUE #3292: polygonize: cupy integer backend silently wraps region IDs past uint32 max
#### Description

The CPU connected-component pass in `_calculate_regions` raises a clear error when it runs out of uint32 region IDs (`xrspatial/polygonize.py`, ~line 749):

```
RuntimeError: polygonize generates too many polygons, split your raster into smaller chunks.
```

The GPU integer path has no equivalent check. `_calculate_regions_cupy` assigns per-value labels with

```python
regions[where] = labeled[where].astype(cp.uint32) + cp.uint32(uid - 1)
```

and keeps incrementing `uid` across unique values. Once the cumulative region count passes `np.iinfo(np.uint32).max`, the uint32 array addition wraps modulo 2**32. The first wrapped ID is 0, which is the masked-out sentinel, so those pixels drop out of the output entirely; later wraps collide with real low region IDs and merge unrelated polygons. No error, no warning. Once `uid - 1` itself exceeds uint32 max, `cp.uint32(uid - 1)` raises `OverflowError: Python integer ... out of bounds for uint32`, which is at least an error, but a confusing one.

Verified the wrap on a GPU host with cupy 13.x / numpy 2.3:

```python
>>> labeled = cp.array([1, 2, 3], dtype=cp.int32)
>>> labeled.astype(cp.uint32) + cp.uint32(2**32 - 3)
array([4294967294, 4294967295,          0], dtype=uint32)
```

#### Reachability

You need more than 2**32 - 1 total regions, which means a raster with over 4.29 billion pixels on a single GPU. A 65536x65536 int8 checkerboard does it (every pixel is its own region under 4-connectivity) and fits in roughly 45 GB of device memory, so an 80 GB card can hit this. The same input on the numpy backend raises the RuntimeError above. The dask+cupy path is unaffected: chunks go through the CPU CCL, which is guarded per chunk.

#### Expected behaviour

The cupy integer backend should raise the same RuntimeError as the CPU path when the cumulative region count would exceed the uint32 ceiling, instead of silently corrupting region labels.

Found by the 2026-06-12 security sweep (Cat 2, integer overflow in index math; severity medium).

brendancol:
Closes #3292.

The CPU CCL (`_calculate_regions`) raises `RuntimeError("polygonize generates too many polygons, split your raster into smaller chunks.")` when it runs out of uint32 region IDs. The cupy integer path had no such check: `_calculate_regions_cupy` keeps offsetting per-value labels by `uid - 1`, and past `np.iinfo(uint32).max` the uint32 array addition wraps modulo 2**32. The first wrapped ID is 0, the masked-out sentinel, so those pixels silently vanish from the output; later wraps collide with real low region IDs and merge unrelated polygons. Reachable with a >4.29-billion-pixel integer raster on an 80 GB GPU (a 65536x65536 int8 checkerboard fits).

- Add the same guard to `_calculate_regions_cupy`: raise the CPU path's `RuntimeError` before the wrapping assignment when `uid - 1 + n_features` would exceed the uint32 ceiling.
- Tests in `test_polygonize_issue_3292.py` trip the guard by patching `cupyx.scipy.ndimage.label` to report 2**31 features per value (so two values cross the ceiling without allocating billions of pixels), and check a normal small raster still polygonizes without tripping it.

Backend coverage: only the single-GPU cupy integer path changes. numpy already had the guard; float cupy routes through the CPU path; dask and dask+cupy run the guarded CPU CCL per chunk.

Test plan:
- [x] `test_polygonize_issue_3292.py` fails without the guard, passes with it (run on a CUDA host)
- [x] All 489 polygonize tests pass (`pytest -k polygonize`)
- [x] flake8 clean


brendancol (commented):
## PR Review: polygonize: raise on uint32 region overflow in cupy integer backend (#3292)

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The guard condition (`uid - 1 + int(n_features) > max_region`, polygonize.py:1218) matches the CPU semantics exactly: `_calculate_regions` assigns IDs up to and including `np.iinfo(uint32).max` and raises only when it needs one more, and this check permits a batch ending exactly at the ceiling. A single oversized batch also trips it on the first value since uid starts at 1.
- The error message is identical to the CPU one, so callers matching on the message see the same failure on both backends.
- The guard sits after the `n_features == 0` continue and before the wrapping assignment, so nothing is written for the offending batch.
- Test design is sound: patching `cupyx.scipy.ndimage.label` to report 2**31 features per value crosses the ceiling on the second value without large allocations, and the patch is read at call time by the production import. The companion test confirms small rasters do not trip the guard.
- Red/green was verified: the raise test fails with the guard removed.

One option I considered and rejected: a boundary test asserting that landing exactly on the ceiling does not raise. With the mock approach that drives region IDs to ~2**31, and `_renumber_regions` then allocates a remap array of `max_old + 1` uint32 (~8.6 GB) on the CPU, so it is not practical in CI. The `>` vs `>=` distinction is instead covered by reasoning against the CPU guard's semantics above.

### Checklist
- [x] Guard semantics match the CPU reference (`_calculate_regions`)
- [x] Backend coverage correct (numpy/dask paths already guarded; only single-GPU integer path changed)
- [x] NaN handling unaffected (float dtypes route through the CPU path)
- [x] Edge cases covered (oversized first batch, normal small raster)
- [x] No premature materialization or new allocations
- [x] Benchmark not needed (error path only)
- [x] No docs/README change needed (no public API change)

Issue #2064, opened 2026-05-18
rasterize: merge='first'/'last' ignores input order across geometry types


brendancol:
**Describe the bug**

`rasterize()` with `merge='first'` or `merge='last'` does not respect input order across geometry types. The pipeline classifies geometries into polygons, lines, and points, then burns them in that fixed order (polygons first, lines second, points last). The merge mode decides which write wins on a per-pixel basis, but it only sees the order within a single type bucket. Cross-type order is decided entirely by the burn phase.

Reproducer:

```python
from shapely.geometry import Point, Polygon
from xrspatial import rasterize

geoms = [(Point(0.5, 0.5), 9), (Polygon([(0,0),(1,0),(1,1),(0,1)]), 1)]
result = rasterize(geoms, width=2, height=2, bounds=(0,0,1,1), merge='last')
# Returns 9 at the covered pixel; should return 1 (the polygon is the last input).
```

Affects all four backends: numpy, cupy, dask+numpy, dask+cupy.

**Expected behavior**

`merge='last'` returns the last geometry in user input order. `merge='first'` returns the first. The burn pipeline order should not leak into public semantics.

**Suggested fix**

Track the global input index per geometry through classification. Carry an `order` array (int64, same shape as `out`) recording the input index of the current owner. For `'first'` and `'last'`, gate writes on the comparison between the new geometry's input index and the current owner index. Commutative merges (`max`, `min`, `sum`, `count`) and user-provided callables keep current behavior.

**Test gap**

Existing tests cover per-type correctness and backend parity but not mixed-type ordered burns. A regression test should construct inputs where the user order is not the type-bucket order and assert the merge result.

Files: `xrspatial/rasterize.py:1028`, `1032`, `1052`, `1059`

PR #3302, opened 2026-06-12
Add GPU tests for mixed-type first/last merge ordering (#3296)

### LINKED ISSUE #3296: rasterize: first/last merge with mixed geometry types has no GPU test
Test-coverage sweep pass 6 over `xrspatial/rasterize.py` (2026-06-12). One cross-backend gap, MEDIUM.

`merge='first'` / `merge='last'` pick the winner of an overlapping pixel by global input index across geometry types (issue #2064). The CPU backends do this with the `_should_write_first` / `_should_write_last` predicates inside the burn kernels. The GPU backends do it differently: pass 1 resolves the winning input index with `cuda.atomic.min`/`max` on the `order` array across the scanline, line, and point kernels, then pass 2 stamps the winner's value (`_apply_merge_gpu` / `_apply_merge_gpu_pass2` in `_ensure_gpu_kernels`).

What exists today:

- `TestMergeOrderAcrossTypes` (test_rasterize.py:672) pins the cross-type input-order semantics, on numpy and dask+numpy only.
- `test_rasterize_gpu_race_2167.py` pins first/last on cupy, but every scenario is single-type: coincident points, crossing lines, duplicate segments.
- `test_rasterize_coverage_2026_05_27.py` pins eager-cupy merge parity for polygons-only and points-only scenes.
- The GeometryCollection parity tests mix a polygon and a point on all four backends, but the geometries don't overlap and share one burn value, so they can't tell winners apart.

No test on cupy or dask+cupy ever has two geometries of different types competing for the same pixel under an ordered merge. A regression in the GPU cross-type order resolution (say, per-kernel order buffers, or a pass-2 kernel firing before all pass-1 kernels) would pass the whole suite.

Probed on a CUDA host on 2026-06-12: current output matches numpy on both GPU backends, so this is a coverage gap, not a source bug.

Plan: test-only PR pinning the #2064 mixed-type scenarios (Point + LineString + Polygon with distinct values and overlapping pixels, merge='first' and merge='last') on cupy and dask+cupy against the numpy reference, plus the same scene with all_touched=True so the supercover pass-2 kernel sees cross-type competition too.

brendancol:
Closes #3296. Test-only; no source changes.

- `merge='first'`/`'last'` resolve overlaps by global input index across geometry types (#2064), but that was pinned on numpy and dask+numpy only. The GPU backends use a separate two-pass atomic scheme (pass 1 resolves the winning index across the scanline, line, and point kernels; pass 2 stamps the winner), and no cupy or dask+cupy test ever had two geometry types competing for the same pixel under an ordered merge.
- New file `test_rasterize_mixed_type_ordered_merge_3296.py`: known-winner pins on all four backends (numpy / cupy / dask+numpy / dask+cupy), cupy and dask+cupy (chunks=2, four tiles) parity against numpy for both modes, an `all_touched=True` variant so the supercover pass-2 kernel sees cross-type competition, and a fixture non-degeneracy guard (first != last).
- Also commits the test-coverage sweep state CSV row for rasterize (pass 6).

Test plan:
- [x] 15 new tests pass on a CUDA host (cupy + dask+cupy executed, not skipped)
- [x] Mutation check: swapping `cuda.atomic.min` for `max` in the `'first'` branch of `_apply_merge_gpu` flips 5 tests red; clean restore verified by md5
- [x] Full rasterize suite: 757 passed, 1 xfailed (pre-existing #3102 pin), 2 skipped


brendancol (commented):
## PR Review: Add GPU tests for mixed-type first/last merge ordering (#3296)

Test-only PR; correctness/performance checks apply to the tests themselves.

None.

- [ ] `test_rasterize_mixed_type_ordered_merge_3296.py:154-161` -- the `all_touched=True` ordered-merge case runs only on eager cupy. The dask+numpy tile path is also ordering-relevant here: `_rasterize_tile_numpy` runs the supercover boundary burn against a tile-local `order` array with global indices sliced by `_slice_per_tile`, and no existing test combines all_touched with first/last on dask+numpy either. An extra parity case is a few lines.
- [ ] Same spot -- dask+cupy all_touched parity is (correctly) not asserted, since `test_rasterize_props_hoist_2506.py:179-190` documents a pre-existing boundary-segment-on-tile-border bug and only smoke-tests that combination. Worth a short comment in the new file saying the omission is deliberate, so the next coverage sweep doesn't re-flag it.

- [ ] `test_rasterize_mixed_type_ordered_merge_3296.py:105-113` and `123-131` -- the four-entry backend parametrize list is pasted twice. A module-level `_BACKENDS` constant keeps the two tests in sync if a backend id or marker changes.

### What looks good

- Known-winner pins on all four backends anchor the parity tests; the two GPU backends cannot pass by agreeing with a wrong numpy result.
- The mutation check in the PR body (atomic.min -> max in the `'first'` branch) shows the tests detect a real GPU order-resolution regression, and the fixture non-degeneracy guard rules out a both-modes-same-kernel regression slipping through.
- `chunks=2` on the 3x3 grid forces a 4-tile dask layout, so the point, line, and polygon actually compete inside shared tiles.
- `_as_numpy` materialises dask before transferring cupy chunks, matching the project convention.

### Checklist

- [x] Algorithm matches reference (#2064 semantics, mirrored from TestMergeOrderAcrossTypes)
- [x] All implemented backends produce consistent results (now asserted on cupy / dask+cupy)
- [x] NaN handling is correct (n/a -- fill=0 scene, NaN merge behaviour covered by #2255 tests)
- [x] Edge cases are covered by tests (mixed-type overlap, multi-tile, all_touched on eager cupy)
- [x] No premature materialization or unnecessary copies (tests only materialise final outputs)
- [x] Benchmark exists or is not needed (test-only)
- [x] Docstrings present and accurate (module docstring explains the gap and the GPU two-pass scheme)

brendancol (commented):
## Follow-up review (after 2cc889c3)

- dask+numpy all_touched ordered-merge parity added as `TestDaskNumpyAllTouchedOrderedMerge` -- correctly placed in a dask-only class so it still runs on CPU-only CI rather than hiding under the CUDA skip.
- The deliberate omission of dask+cupy all_touched parity is now documented inline with a pointer to the pre-existing tile-border bug noted in test_rasterize_props_hoist_2506.py.
- The duplicated backend parametrize list is hoisted to a module-level `_BACKENDS` constant.

17 tests pass on a CUDA host. No new findings.

PR #3305, opened 2026-06-12
polygonize: cover the dask mask-rechunk path (#3299)

### LINKED ISSUE #3299: polygonize: dask mask rechunk path has no test coverage
`_polygonize_dask` aligns the mask to the raster's chunk layout before building per-chunk tasks (xrspatial/polygonize.py):

```python
# Ensure mask chunks match raster chunks.
if mask_data is not None and mask_data.chunks != dask_data.chunks:
    mask_data = mask_data.rechunk(dask_data.chunks)
```

No test hits this branch. Every dask test that passes a mask builds it with the same chunks as the raster, so the condition is always false. The public API only validates the mask's type and shape, not its chunk layout, so users can reach this branch with any mismatched chunking.

If the rechunk were removed or broken, `mask_data.blocks[iy, ix]` would walk a different block grid than the raster. Some layouts fail loudly (IndexError, shape mismatch inside the chunk function). But with raster chunks (5, 6) and mask chunks (6, 7) on a (15, 18) raster, both block grids are 3x3, and an integer raster gets masked by misaligned, differently-shaped blocks with no error at all: `_calculate_regions` just indexes the flattened mask per pixel. Verified by mutation: deleting the rechunk flips every new test red.

Tests to add (no source change needed):

- masks with mismatched chunk layouts (same-grid-shape misalignment `(6, 7)`, single-chunk `(15, 18)`, more blocks than the raster `(4, 5)`) pinned against the same mask chunked to match the raster, on dask+numpy and dask+cupy, int and float rasters, connectivity 4 and 8
- an exact-geometry anchor: single-chunk mask over a multi-chunk raster excludes exactly the masked pixel

Found by the test-coverage sweep on 2026-06-12.

brendancol:
Closes #3299.

- Adds `test_polygonize_mask_chunk_mismatch_3299.py`: dask masks whose chunk layout differs from the raster's, pinned against the same mask chunked to match. Layouts: `(6, 7)` same-grid-shape misalignment (the one that would silently mis-mask an integer raster if the rechunk guard disappeared), single-chunk `(15, 18)`, and more-blocks `(4, 5)`. Int and float rasters, connectivity 4 and 8.
- Adds an exact-geometry anchor: a single-chunk mask over a multi-chunk raster carves a hole at exactly the masked pixel.
- Test-only; `xrspatial/polygonize.py` is untouched. Also carries this sweep's state-CSV row update.

Backends: dask+numpy and dask+cupy run the parametrized matrix (numpy and cupy have no chunk layout, so the branch doesn't exist there). All 25 tests executed locally on a CUDA host.

Test plan:
- [x] 25 new tests pass locally (dask+numpy and dask+cupy)
- [x] Mutation check: deleting the rechunk guard in `_polygonize_dask` flips all 25 red; clean restore verified by md5
- [x] Full polygonize suite afterwards: 486 passed, 16 skipped


brendancol (commented):
## PR Review: polygonize: cover the dask mask-rechunk path (#3299)

None.

- [ ] `test_mask_chunk_mismatch_excludes_correct_pixels` (xrspatial/tests/test_polygonize_mask_chunk_mismatch_3299.py:113) runs only on dask+numpy. The parametrized matrix already covers dask+cupy for the equivalence pin, but the exact-geometry anchor is cheap to parametrize over both dask backends, and the dask+cupy path has its own per-chunk block extraction (`_to_numpy` in `_polygonize_chunk`) worth anchoring with exact coordinates too.

- [ ] The float-raster cases use integer-valued floats (0.0/1.0/2.0), so the tolerance-chain machinery runs with trivially distinct values. That is fine for this file's purpose (the rechunk branch sits upstream of value semantics), but worth noting so nobody mistakes this file for tolerance coverage; that lives in test_polygonize_atol_rtol_backend_coverage_2026_05_27.py.

### What looks good
- The (6, 7) mask layout is the right adversarial choice: it produces the same 3x3 block grid as the raster, so a missing rechunk mis-masks an integer raster silently instead of raising. The other two layouts cover the loud-failure shapes.
- Mutation evidence in the PR body (all 25 tests red with the guard deleted, clean md5 restore) verifies the tests actually guard the branch.
- Comparing each backend against its own aligned-mask reference isolates chunk layout as the only variable and sidesteps the known numpy-vs-dask ring-representation differences.

### Checklist
- [x] Algorithm matches reference: N/A, test-only change
- [x] All implemented backends produce consistent results (dask+numpy, dask+cupy exercised; numpy/cupy have no chunk layout)
- [x] NaN handling: out of scope for this branch; covered elsewhere
- [x] Edge cases covered (same-grid misalignment, single-chunk mask, finer mask grid)
- [x] Dask chunk boundaries handled correctly (that is the subject under test)
- [x] Benchmark: not needed (test-only)
- [x] Docstrings: module docstring explains the gap and the silent-corruption layout

brendancol (commented):
## Follow-up review (after f072f31b)

The suggestion from the first pass is addressed: `test_mask_chunk_mismatch_excludes_correct_pixels` is now parametrized over both dask backends and reuses `_make_pair`, so the exact-geometry anchor also pins the dask+cupy per-chunk block extraction. All 26 tests pass on a CUDA host; flake8 clean.

Disposition of first-pass findings:
- Suggestion (anchor only on dask+numpy): fixed in f072f31b.
- Nit (integer-valued floats, not tolerance coverage): no change needed; the module docstring already scopes this file to the rechunk branch, and tolerance coverage lives in test_polygonize_atol_rtol_backend_coverage_2026_05_27.py.

PR #3308, opened 2026-06-12
polygonize: honor _xrspatial_no_georef in CRS detection (#3293)

### LINKED ISSUE #3293: polygonize: CRS attached to pixel-space geometries when raster carries the _xrspatial_no_georef marker
**Describe the bug**

`polygonize(return_type='geopandas')` gets its output CRS from `_detect_raster_crs`, which reads `attrs['crs']`, then `attrs['crs_wkt']`, then `rio.crs`. The transform counterpart, `_detect_raster_transform`, returns None when the raster carries `attrs['_xrspatial_no_georef'] = True` (the xrspatial.geotiff "no georeference" marker), so geometries stay in pixel space. The CRS detector ignores the marker.

A raster can carry both `crs` and the marker. `xrspatial.geotiff.open_geotiff` produces exactly that for a file with CRS geokeys but no geotransform tags (`georef_status == 'crs_only'`; `metadata_to_attrs` in `xrspatial/geotiff/_attrs.py` writes `crs` regardless of `has_georef`). Polygonizing such a raster gives a GeoDataFrame whose `.crs` says EPSG:4326 while the geometries are pixel indices. Same metadata/data mismatch that #2536 and #2607 fixed for the other channels.

`contours()` uses the same helper (`xrspatial/contour.py` imports `_detect_raster_crs`), so it has the same bug.

**Reproduction**

```python

data = np.array([[1, 1, 2], [1, 1, 2], [2, 2, 2]], dtype=np.int32)
raster = xr.DataArray(
    data, dims=('y', 'x'),
    attrs={'crs': 4326, '_xrspatial_no_georef': True},
)
df = polygonize(raster, return_type='geopandas')
print(df.crs)                    # EPSG:4326
print(df.geometry.total_bounds)  # [0. 0. 3. 3.]  (pixel space)
```

**Expected behavior**

The marker opts the raster out of georeferencing. `_detect_raster_crs` should return None when it is set, matching `_detect_raster_transform`, so the GeoDataFrame comes back with `crs=None` and pixel-space geometries that agree with each other.

**Affected backends**

All four (numpy, cupy, dask+numpy, dask+cupy). CRS detection runs in the public `polygonize()` before backend dispatch.

**Additional context**

Found by the metadata-propagation sweep. The rotated-read path (`allow_rotated=True`) is not affected; `geo_info_to_metadata` drops CRS attrs on that path.

brendancol:
Closes #3293

- `_detect_raster_crs` now returns None when the raster carries `attrs['_xrspatial_no_georef']=True`, matching `_detect_raster_transform`. Previously a geotiff-reader `crs_only` raster (CRS geokeys, no geotransform tags) came back as a GeoDataFrame whose `.crs` claimed CRS coordinates over pixel-space geometries.
- `contours()` resolves CRS through the same helper, so it picks up the fix without changes.
- Added a note to the `polygonize()` docstring's CRS paragraph.

Backend coverage: CRS detection runs in the public `polygonize()` before backend dispatch. Verified live that numpy, cupy, dask+numpy, and dask+cupy all return `crs=None` with pixel-space bounds for a marked raster (previously all four attached the CRS).

Test plan:
- [x] 2 new tests in `TestPolygonizeCRSPropagation` (crs attr + marker, crs_wkt + marker)
- [x] `pytest xrspatial/tests/test_polygonize.py xrspatial/tests/test_contour.py` -- 274 passed
- [x] 9 auxiliary polygonize test files -- 303 passed
- [x] Metadata-sweep state CSV row updated for polygonize


brendancol (commented):
## PR Review: polygonize: honor _xrspatial_no_georef in CRS detection (#3293)

- None.

- [ ] `contours()` is named in the issue and PR body as inheriting this fix (`xrspatial/contour.py:776,809` call `_detect_raster_crs`), but no contour-side test pins that. If someone later forks or inlines the helper in contour.py, the marker handling regresses there silently. A 5-line test in `xrspatial/tests/test_contour.py` (marker + `crs` attr, assert `df.crs is None`) would lock it in.
- [ ] The new docstring text (xrspatial/polygonize.py:2715-2719 and 909-917) says "the geometries stay in pixel space on that path", but that is only true when no explicit `transform=` is passed. A caller who passes `transform=` with a marked raster gets transformed geometries and, after this PR, `crs=None` (previously the CRS attached). That is a defensible reading of the marker as a hard opt-out, but the docstring should say the marker suppresses auto-detected CRS even when an explicit `transform=` is given, so callers know to `set_crs` themselves.

- [ ] The marker check is truthy (`attrs.get('_xrspatial_no_georef')`, xrspatial/polygonize.py:916) while geotiff's `_has_no_georef_marker` requires `is True` (`xrspatial/geotiff/_coords.py:155`). A third-party stamp like `'yes'` suppresses CRS here but is not a marker to geotiff. This mirrors the pre-existing transform-side check (line 970), and internal consistency within polygonize is the right priority, so noting it only.

### What looks good
- Fix sits at the single point both polygonize and contours flow through; one early return, no backend dispatch touched.
- Tests assert both halves of the invariant (crs is None AND bounds are pixel-space), not just the CRS.
- Reader-derived crs_only attrs shape verified live on all four backends per the PR body.
- State CSV diff is a single row.

### Checklist
- [x] Algorithm matches reference (mirrors `_detect_raster_transform` marker handling)
- [x] All implemented backends produce consistent results (detection precedes dispatch; verified live)
- [x] Edge cases covered (crs attr, crs_wkt fallback)
- [x] Benchmark not needed (metadata-only)
- [x] Docstrings updated (with the wording caveat above)

brendancol (commented):
## Follow-up review (after d326a9c5)

- **Fixed:** contour-side regression test added (`xrspatial/tests/test_contour.py::TestGeoDataFrame::test_geopandas_no_georef_marker_suppresses_crs`). It builds the marker+crs attrs shape directly and asserts a populated result with `gdf.crs is None`, so the shared-helper behavior is pinned on the consumer.
- **Fixed:** both docstrings (`_detect_raster_crs` and the `polygonize()` Notes section) now say the marker suppresses auto-detected CRS even when an explicit `transform=` is passed, and point callers at `set_crs` when deliberately overriding the opt-out.
- **Dismissed (nit):** truthy vs `is True` marker semantics. Tightening only the CRS check would desync it from the pre-existing truthy transform-side check; changing both is a separate cross-module decision. Internal consistency within polygonize is preserved, which was the point of the fix.

No new findings in the delta. polygonize + contour suites: 275 passed.

PR #3309, opened 2026-06-12
polygonize: document that column_name is also used for geojson output

### LINKED ISSUE #3306: polygonize: column_name docstring omits the geojson return type
The `column_name` docstring in `polygonize()` says:

> Name to use for column returned.  Only used if return_type is "geopandas" or "spatialpandas".

But the geojson path uses it too. `_to_geojson(column, polygon_points, column_name)` writes it as the property key on every feature:

```python
>>> gj = polygonize(raster, return_type="geojson", column_name="myval")
>>> gj["features"][0]["properties"]
{'myval': 1}
```

So the docstring claims a narrower scope than the code has. Someone reading the docs would assume geojson output always uses the default "DN" key, and code keying on `properties["DN"]` breaks as soon as a caller passes `column_name` for another reason.

Found by the api-consistency sweep (category 3, docstring/signature divergence). Fix is a one-line docstring correction in `xrspatial/polygonize.py` (around line 2630).

brendancol:
Closes #3306

- The `column_name` docstring said the parameter was only used when `return_type` is "geopandas" or "spatialpandas". The geojson path uses it too: `_to_geojson` writes it as the property key on every feature. The docstring now says so.
- Added `test_polygonize_geojson_column_name` to pin the geojson property naming, so the docs and the behavior can't drift apart again.

Docs-only change to the function body; no backend code touched (numpy / cupy / dask+numpy / dask+cupy all share the same return_type dispatch).

Test plan:
- [x] `pytest xrspatial/tests/test_polygonize.py -k "geojson or invalid_return_type"` (11 passed)

Found by the api-consistency sweep against the polygonize module.


brendancol (commented):
Review of the docstring fix and test.

Blockers: none.

Suggestions: none. The new wording matches what `_to_geojson` actually does (property key per feature), and the test pins both the key and the values so the docs can't silently drift again.

Nits:
- The Returns section's geojson bullet ("dict representing a GeoJSON FeatureCollection") could also name `column_name` as the property key, but the parameter doc now covers it, so this is optional.

Checked: `_to_geopandas` / `_to_spatialpandas` / `_to_geojson` call sites all consume `column_name`; "awkward" and "numpy" do not, consistent with the new wording. Test values [0, 1, 4] match the existing `test_polygonize_geojson` expectations for the same fixture.

PR #3310, opened 2026-06-12
Drop non-finite-coordinate geometries and clamp row casts in rasterize

### LINKED ISSUE #3295: rasterize: NaN/inf and int32-overflowing coordinates burn phantom pixels
**Describe the bug**

`rasterize()` produces silently wrong output when a geometry contains non-finite (NaN/inf) coordinates, or coordinates far enough outside the raster bounds that their pixel-space position overflows int32.

The edge/segment extraction helpers cast float pixel coordinates to int32 without checking finiteness or range first:

- `_extract_edges_vectorized` filters horizontal edges with `sr != er`, which is True for NaN, so NaN edges survive. `np.ceil(NaN).astype(np.int32)` then yields INT_MIN, and `INT_MIN - 1` wraps to INT_MAX, turning the edge into a phantom full-height edge (`ry_min=0`, `ry_max=height-1`).
- The same casts overflow for finite coordinates more than 2^31 pixels from the bounds. A polygon entirely below the raster at pixel row ~3e9 becomes a full-height edge pair instead of being filtered by the `ry_min <= ry_max` check.
- `_extract_lines_vectorized` Liang-Barsky clipping passes NaN endpoints through (`valid` stays True because every NaN comparison is False), and the subsequent `np.clip(INT_MIN, 0, ...)` lands the segment at column 0.

Observed on the numpy backend with a 10x10 raster, bounds (0, 0, 10, 10):

- Polygon with one vertex at y=-3e9, entirely off-raster: burns a 60-pixel phantom block. dask+numpy burns 0 for the same input, because the tile bbox filter correctly excludes it, so the two backends disagree.
- Polygon with one inf vertex: burns 12 pixels to the left of the polygon's actual x-extent.
- Polygon with one NaN vertex: silently dropped on the scanline path (0 pixels vs 36 expected), yet burns 13 boundary pixels with `all_touched=True`. Same backend, two different answers.
- LineString with a NaN endpoint: burns 5 spurious pixels in column 0.
- The GPU backend reproduces all of the above, since the extraction runs on the host either way.

The int32 overflow is reachable with real data. Pixel position is `(ymax - y) / py`, so a web-mercator coordinate (~2e7 m) burned onto a centimeter-resolution local grid, or any wrong-CRS mixup with a fine-resolution raster, crosses 2^31 pixels.

**Expected behavior**

Geometries with non-finite coordinates should be dropped with a warning, the same convention as the existing unsupported-geometry-type warning, and the drop should behave identically on the scanline, all_touched, line, and point paths across all four backends. Finite coordinates far outside the bounds should rasterize correctly: edges crossing the raster fill the right rows, fully off-raster edges get filtered, no phantom pixels at any magnitude.

**Reproduction**

```python
import numpy as np

poly_off = Polygon([(2, -2.5e9), (8, -2.5e9), (8, -3e9), (2, -3e9)])
res = rasterize([(poly_off, 9.0)], width=10, height=10, bounds=(0, 0, 10, 10), fill=0.0)
print(np.count_nonzero(res.values == 9.0))  # 60, should be 0
```

**Additional context**

Found by the security sweep (integer overflow in index math, NaN as logic error). The only symptom a user sees today is a numpy RuntimeWarning "invalid value encountered in cast" from inside the extraction helpers, which names neither the geometry nor the consequence.

brendancol:
Closes #3295.

- Geometries containing NaN or inf coordinates are now dropped with a warning in `_classify_geometries`, so the drop behaves the same on the scanline, all_touched, line, and point paths. The GeometryCollection slow path checks per leaf, so one bad member does not take down the whole collection. Detection inspects coordinates via `shapely.get_coordinates` because GEOS skips NaN when computing bounds, which makes a bbox check blind to exactly these inputs.
- `_extract_edges_vectorized` clamps pixel-space row values to a narrow band around the raster before the int32 casts. Coordinates beyond 2^31 pixels used to overflow the cast (INT_MIN, then `- 1` wraps to INT_MAX) and turn fully off-raster edges into phantom full-height edges; with the clamp they are filtered by the existing `ry_min <= ry_max` check. The `x_at_ymin` extrapolation still uses the unclamped values, so edges entering the raster from far away keep exact x intersections.

Backend coverage: the fix lives in host-side classification and edge extraction shared by numpy, cupy, dask+numpy, and dask+cupy. Before the fix, numpy burned a 60-pixel phantom block for an off-raster polygon while dask burned nothing; the backends now agree.

Probed but unchanged: x-direction analogs (coordinates up to 1e30, and past int64 pixel space at 1e19) burn nothing on x86_64, so the scanline kernels keep their current column handling.

Test plan:
- [x] New regression file `test_rasterize_nonfinite_coords_3295.py`: 17 tests covering NaN/inf polygons, lines, points, GeometryCollection leaves, the all_touched path, the off-raster overflow reproduction, a partially-on-raster polygon burning exactly its in-bounds region, no-warning behavior for clean and finite-far inputs, dask parity, and cupy / dask+cupy variants (run on a CUDA host, all passing)
- [x] Existing rasterize suite: 742 passed, 2 skipped, 1 xfailed
- [x] flake8 clean

Also carries the security-sweep state CSV row for the rasterize module (sweep convention).


brendancol (commented):
## PR Review: Drop non-finite-coordinate geometries and clamp row casts in rasterize

- None.

- [ ] `xrspatial/rasterize.py` (~line 318): the vectorized non-finite check runs before the fast/slow branch in `_classify_geometries`. When the input contains both a GeometryCollection and a top-level non-finite geometry, the vectorized check warns for the top-level geometry, then the slow path's per-leaf counter warns for it again, so the user sees two warnings for one drop. The drop itself happens exactly once (the slow path ignores `valid`), so this is cosmetic, but it's confusing. Run the vectorized check only when no GeometryCollection is present and let the slow path own the warning otherwise.

- [ ] No test covers the mixed input above (GeometryCollection alongside a top-level NaN geometry). Worth one test pinning a single warning and correct burns once the suggestion lands.
- [ ] `shapely.get_coordinates` over the full geometry array now runs on every `rasterize()` call, including clean inputs. It's one extra vectorized O(V) pass and the extractors build coordinate arrays of the same order anyway, so this seems acceptable, but it is per-call overhead on the hot path. No cheaper detection exists since GEOS bounds skip NaN.

### What looks good
- The clamp only feeds the int32 casts; `x_at_ymin` extrapolates from unclamped `top_r`, so edges entering the raster from far away keep exact x intersections. The `[-2, height+2]` band reproduces the original semantics for every in-range value (test `test_int32_overflow_partially_on_raster_polygon_exact` pins this).
- Detection via coordinates instead of bounds is correct and the comment explains why (GEOS skips NaN in bounds).
- Fix sits in shared host-side code, so all four backends agree; tests cover numpy, dask, cupy, and dask+cupy plus the all_touched path and GeometryCollection leaves.
- The x-direction analog was probed (1e30 and past-int64 1e19 pixel coords) and burns nothing, which justifies leaving the scanline kernels alone.

### Checklist
- [x] Algorithm matches reference behavior (off-raster edges filtered, in-bounds region exact)
- [x] NaN handling is correct
- [x] Edge cases are covered by tests (one gap noted above)
- [x] Dask chunk boundaries handled correctly (classification is driver-side)
- [x] No premature materialization; one extra vectorized pass noted above
- [x] Benchmark not needed (bug fix; no kernel changes)
- [x] Docstrings updated (drop behavior documented in `rasterize`)

brendancol (commented):
Follow-up review after 109061df.

- Double-warning suggestion: fixed. The vectorized check is now skipped whenever a GeometryCollection is present, and the slow path owns the per-leaf check and warning. `test_mixed_collection_and_toplevel_nonfinite_warns_once` pins exactly one warning plus correct burns for the mixed input.
- Missing-test nit: fixed by the same test.
- Per-call `get_coordinates` overhead nit: left as is, with reasoning on the PR. GEOS bounds skip NaN so there is no cheaper detection, and the pass is one vectorized O(V) sweep of the same order the extractors already do.

Test state: 18/18 in the new file, 760 passed / 2 skipped / 1 xfailed across the rasterize suite, flake8 clean. No remaining findings.

PR #3311, opened 2026-06-12
Cut redundant work in interpolate backends

### LINKED ISSUE #3298: interpolate: kriging and idw backends repeat avoidable work
## Reason or Problem

A performance audit of `xrspatial/interpolate/` found three places where the kriging and IDW backends do work they could skip. All numbers below were measured on this machine (CUDA available).

**1. `kriging()` builds the full weight matrix even when no variance is requested.**

`_kriging_predict()` (and `_kriging_predict_cupy()`) always compute `w = k0 @ K_inv`, a `(grid_pixels, N+1) x (N+1, N+1)` matmul, then dot `w` against `z`. When `return_variance=False` (the default), the same prediction comes from `k0 @ (K_inv @ z_aug)` with `z_aug = [z, 0]`: one `(N+1,)` matvec up front, then a matvec per pixel instead of a matrix product. This also drops the `(grid_pixels, N+1)` `w` temporary.

Measured (numpy backend): N=1000 and N=2000 points on a 200x200 grid both run 1.6x faster with the dual form; max difference in output is ~5e-9 absolute. The gap grows with N since the matmul term is O(N^2) per pixel.

**2. The dask backends compute the prediction twice when `return_variance=True`.**

`_kriging_dask_numpy()` and `_kriging_dask_cupy()` build two independent `map_blocks` graphs. The variance graph calls `_kriging_predict(..., True)` per chunk and throws the prediction away, so every chunk runs the dists/k0/matmul pipeline twice. Measured on an 800x800 grid with 800 points, chunks (200, 200): prediction alone 6.25 s, prediction plus variance 12.89 s (2.06x).

**3. Chunk-invariant data is re-prepared per chunk.**

- `_idw_dask_cupy()` re-uploads the `x`, `y`, `z` point arrays to the device for every chunk (measured 48 uploads for 16 chunks; 3 expected).
- `_kriging_dask_cupy()` re-uploads `x`, `y`, `z`, and `K_inv` for every chunk (64 uploads for 16 chunks; 4 expected). `K_inv` is `(N+1)^2` float64, so PCIe traffic scales with N^2 times the chunk count.
- `_idw_dask_numpy()` with `k` set rebuilds the same `cKDTree` in every chunk task.

The spline module already hoists its uploads out of the per-chunk closure (see `_spline_dask_cupy` and `test_dask_cupy_uploads_points_once`); idw and kriging never got the same treatment.

## Proposal

- In `_kriging_predict` / `_kriging_predict_cupy`, branch on `return_variance`: keep the `w`-based path when variance is needed, use `K_inv @ z_aug` otherwise.
- In the dask kriging backends, compute prediction and variance in a single `map_blocks` pass that returns a stacked `(2, ny, nx)` block, then slice channels, so chunks are processed once.
- Hoist the device uploads in `_idw_dask_cupy` and `_kriging_dask_cupy` out of the per-chunk closure (`cupy.asarray` is a no-op on arrays already on the device, so the inner helpers need no signature change). Build the `cKDTree` once in `_idw_dask_numpy`.

**Value:** faster default kriging on all four backends, half the chunk work for dask variance, and no repeated host-to-device transfers in the dask+cupy paths.

## Stakeholders and Impacts

Touches `_kriging.py` and `_idw.py` only. No API change; outputs differ from the current code by float round-off (~1e-10 relative) on the no-variance kriging path.

## Drawbacks

The dual-form prediction associates the float ops differently, so results shift within normal floating-point tolerance. Existing cross-backend tests compare at rtol=1e-10 and still pass.

Found by the 2026-06-12 performance sweep of the `interpolate` module. OOM verdict for the module's dask paths: SAFE (plain `map_blocks`, 2 tasks per chunk, per-chunk memory bounded by chunk size and N). Bottleneck class: compute-bound.

brendancol:
Closes #3298

Three fixes from the interpolate performance sweep, all behavior-preserving up to float round-off:

- `kriging()` with `return_variance=False` (the default) no longer computes the full `(grid_pixels, N+1) x (N+1, N+1)` weight matmul. The prediction comes from `k0 @ (K_inv @ z_aug)` instead, which is exactly equal because `z_aug[n] == 0`. Measured 1.6x on the numpy backend at N=1000-2000 and it drops the `(grid_pixels, N+1)` weight temporary. The variance path keeps the weight-based formulation since variance needs the full `w`.
- The dask kriging backends now compute prediction and variance in a single `map_blocks` pass that returns a stacked `(2, ny, nx)` block, then slice the channels. Before, the variance graph re-ran the whole dists/k0/matmul pipeline per chunk and discarded the prediction: a joint compute of both outputs cost 2.06x the prediction-only run; it now costs 1.23x.
- Chunk-invariant data is prepared once instead of per chunk: `_idw_dask_cupy` and `_kriging_dask_cupy` upload the point arrays (and `K_inv`, which is `(N+1)^2` floats) to the device before the closure, and `_idw_dask_numpy` builds the `cKDTree` once when `k` is set. Measured uploads on a 16-chunk grid went from 48 to 3 (idw) and 64 to 4 (kriging). `cupy.asarray` is a no-op on device-resident arrays, so the inner helpers keep their signatures.

Backends: numpy, cupy, dask+numpy, and dask+cupy all touched; cross-backend parity tests pass on all four (CUDA host).

Test plan:
- [x] `pytest xrspatial/tests/test_interpolation.py` (71 passed, GPU tests included)
- [x] New regression tests: no-variance prediction matches the weight-based path; `_kriging_predict` runs once per chunk for a joint pred+var compute; idw and kriging dask+cupy upload counts don't scale with chunk count; cKDTree built once
- [x] cupy and dask+cupy kriging/idw executed end-to-end on GPU, outputs finite and device-resident
- [x] flake8 on touched files adds no new findings (the 8 reported are pre-existing)

Also recorded in the sweep state CSV: one LOW finding left as documentation only (`_experimental_variogram` bins pairs with a per-lag boolean mask, nlags passes over the pair arrays where one `np.bincount` pass would do).


brendancol (commented):
## PR Review: Cut redundant work in interpolate backends

None found.

- [ ] No benchmark exists for the interpolate module under `benchmarks/benchmarks/`, so the speedups this PR claims (1.6x no-variance kriging, 2.06x -> 1.23x dask variance) have no regression tracking. A small ASV benchmark for `kriging` and `idw` would keep these wins from quietly eroding.
- [ ] PR is missing the `performance` label the repo uses for performance-sensitive changes.

- [ ] `xrspatial/interpolate/_kriging.py:205` - the dual weight vector `K_inv @ z_aug` is chunk-invariant but gets recomputed inside every chunk task on the dask paths (via `_chunk_pred`/`_chunk_both` -> `_kriging_predict`). It is an O(N^2) matvec against an O(chunk_pixels * N) pipeline, so the cost is small (~2% at N=2000 with 256x256 chunks), but it is the same category of waste this PR removes elsewhere. Hoisting it would need a new parameter through `_kriging_predict`, so it may not be worth the signature churn.
- [ ] `xrspatial/interpolate/_idw.py:228` and `_kriging.py:331` - the new hoisting comments say the upload happens once, but that is only true for the threaded/synchronous schedulers; a distributed scheduler re-serializes the captured device arrays per task. `_spline_dask_cupy` documents this caveat; the new comments should too so nobody assumes a guarantee the code does not make.

### What looks good
- The dual-form identity is justified in a comment at the call site (`z_aug[n] == 0`), and `test_prediction_same_with_and_without_variance` pins the two formulations to each other rather than to stored constants.
- The regression tests assert structural properties (upload counts, predict-calls-per-chunk, tree constructions) instead of timing, so they will not flake under load.
- The variance path keeps the weight-based formulation, which it genuinely needs; the fast path only triggers when `w` would be discarded.
- `cupy.asarray` no-op passthrough avoids signature changes in `_idw_cupy` / `_kriging_predict_cupy`, and both docstrings/comments state that host or device arrays are accepted.
- The singular-matrix (`K_inv is None`) and n<3 fallback paths are untouched.

### Checklist
- [x] Algorithm matches reference/paper (dual kriging form; algebraic identity verified to 5e-9 at N=2000)
- [x] All implemented backends produce consistent results (parity tests pass on numpy/cupy/dask/dask+cupy, CUDA host)
- [x] Edge cases are covered by tests (pre-existing single-point, all-NaN, fallback tests still pass)
- [x] Dask chunk boundaries handled correctly (pointwise map_blocks, no overlap needed; ragged 2/1 chunks exercised)
- [ ] Benchmark exists or is not needed (none exists; see suggestion)

brendancol (commented):
## Follow-up review after d68f2ce0

All four findings from the first review are addressed:

- **Benchmark coverage (suggestion): fixed.** `benchmarks/benchmarks/interpolate.py` adds ASV benchmarks for `idw`, `kriging` (with and without variance), and `spline` over numpy/cupy/dask, following the `flood.py` compute pattern. All nine class/backend combinations were smoke-run on this host, including the GPU ones. The Kriging class caps the grid at nx=300 with a comment explaining the per-chunk prediction-matrix memory.
- **`performance` label (suggestion): fixed.** `performance` and `gpu` labels added.
- **Per-chunk `K_inv @ z_aug` (nit): fixed.** New `_dual_z_aug` helper; both dask wrappers compute the dual weight vector once and pass it through the new `dual_w` parameter on the predict functions. The numpy/cupy in-memory paths keep the internal fallback, so call sites outside dask are unchanged.
- **Scheduler caveat comments (nit): fixed.** The hoisting comments in `_idw_dask_cupy` and `_kriging_dask_cupy` now carry the same distributed-scheduler note as `_spline_dask_cupy`.

Re-checked after the changes: 71 tests pass (GPU included), flake8 reports nothing new on the touched files (the `math` F401 in `_idw.py:5` predates this PR). No new findings.

PR #3312, opened 2026-06-12
polygonize: reject invalid return_type before running the computation

### LINKED ISSUE #3307: polygonize: invalid return_type rejected only after the full computation runs
`polygonize()` checks `return_type` at the very end, after the backend has already polygonized the whole raster:

```python
column, polygon_points = mapper(raster)(...)   # full computation
...
if return_type == "numpy":
    ...
else:
    raise ValueError(f"Invalid return_type '{return_type}'")
```

Verified with a spy on `_polygonize_numpy`: passing `return_type="bogus"` runs the backend once, then raises. On a large dask raster a typo costs the entire compute before the error shows up.

`contours()` is the sibling function with the same parameter and it validates up front, before touching the data, with an error that lists the allowed values:

```
Invalid return_type 'bogus'. Allowed values are 'numpy' and 'geopandas'.
```

Every other `polygonize` parameter (connectivity, transform, simplify_tolerance, simplify_method, atol, rtol) is validated before the computation. `return_type` is the only one checked after.

Fix: hoist the `return_type` check into the validation block at the top of `polygonize()` and list the allowed values in the message, matching `contours`. The existing test (`test_polygonize_invalid_return_type`) matches on the `Invalid return_type 'qwerty'` prefix, so the message can grow a suffix without breaking it. No behavior change for valid inputs.

Found by the api-consistency sweep comparing polygonize against contours.

brendancol:
Closes #3307

- `return_type` was the only `polygonize()` parameter checked after the backend dispatch. A typo like `return_type="geopanda"` ran the full polygonize computation, then raised. On a large dask raster that's the whole compute wasted before the error shows up.
- The check now sits in the validation block at the top of the function, next to the connectivity / transform / tolerance checks, and the message lists the allowed values the same way `contours()` does:

  `Invalid return_type 'qwerty'. Allowed values are 'numpy', 'awkward', 'geopandas', 'spatialpandas' and 'geojson'.`

- The dispatch tail's `else: raise` branch is unreachable now, so it returns the geojson conversion directly.

No behavior change for valid inputs on any backend (numpy / cupy / dask+numpy / dask+cupy share the same dispatch). The existing `test_polygonize_invalid_return_type` matches on the message prefix and still passes.

Test plan:
- [x] New `test_polygonize_invalid_return_type_rejected_before_compute`: spy on `_polygonize_numpy` confirms zero backend invocations, and the message names the allowed values
- [x] `pytest xrspatial/tests/test_polygonize.py` (184 passed, 13 skipped)
- [x] flake8 clean on `xrspatial/polygonize.py`

Found by the api-consistency sweep comparing polygonize against contours.


brendancol (commented):
Review of the early return_type validation.

Blockers: none.

Suggestions:
- In `test_polygonize_invalid_return_type_rejected_before_compute`, the second invocation is wrapped in a bare try/except, so if the call ever stopped raising, the message assertion would be skipped and the test would still pass. Use `pytest.raises(...) as excinfo` once and assert both the prefix and the "Allowed values are" suffix on `excinfo.value`.
- The allowed-values tuple now exists in two places (the early check and the dispatch tail). The tail no longer raises, so they can't disagree on the error path, but a module-level constant would keep a future return_type addition to a one-place change. Optional.

Checked: the check sits before the backend dispatch and before transform auto-detection; mapper construction still reads `_polygonize_numpy` at call time so the spy in the test is valid; the existing prefix-matching test passes unchanged; valid-input behavior is untouched on all four backends.

brendancol:
Applied the review suggestion in 272b1b5a: the message assertion now uses the excinfo from the existing pytest.raises, so it can't be skipped silently. Left the allowed-values tuple inline since the dispatch tail no longer raises; a shared constant can come with the next return_type addition if one ever lands.

PR #3313, opened 2026-06-12
Burn each geometry once per pixel under merge='sum'/'count'

### LINKED ISSUE #3304: rasterize: merge='sum'/'count' burns the same geometry into a pixel more than once
With `merge='count'`, a single geometry should produce 1 in every pixel it covers. It doesn't, in two cases:

**1. Polygons with `all_touched=True` (values up to 3x)**

```python

poly = Polygon([(2.2, 2.2), (7.4, 3.1), (5.0, 7.9)])
r = rasterize([(poly, 1.0)], width=10, height=10, bounds=(0, 0, 10, 10),
              fill=0, merge='count', all_touched=True)
print(r.data.max())  # 3.0, should be 1.0
```

Two overlapping write phases hit the same pixels for the same geometry: the scanline fill burns every cell whose center is inside, then the supercover boundary pass burns every cell the rings cross, including the ones scanline just burned. On top of that, consecutive ring segments share a vertex cell, so the boundary pass itself burns vertex cells once per adjoining segment. For thin polygons most covered cells are boundary cells, so `sum`/`count` results are roughly doubled.

rasterio with `merge_alg=MergeAlg.add, all_touched=True` returns 1.0 everywhere for the same input.

**2. Lines: connecting vertices counted once per segment**

```python
from shapely.geometry import LineString
line = LineString([(0.5, 0.5), (5.5, 3.5), (9.5, 0.5)])
r = rasterize([(line, 1.0)], width=10, height=10, bounds=(0, 0, 10, 10),
              fill=0, merge='count')
print(r.data.max())  # 2.0 at the interior vertex, should be 1.0
```

Each Bresenham segment burns both its endpoints, and consecutive segments of the same line share one. GDAL deduplicates within a geometry: rasterio's ADD gives 1.0 here, and 1.0 even for a self-crossing LineString.

All four backends (numpy, cupy, dask+numpy, dask+cupy) agree with each other, which is why the cross-backend parity probes never caught it - they all share the double burn.

Only `sum` and `count` are affected. `min`/`max` are idempotent, `first`/`last` already gate repeat writes through the per-pixel owner index, and user callables keep their documented last-call-wins behavior. Points are fine as-is: GDAL also adds once per point, so a MultiPoint with duplicated coordinates burning 2 matches rasterio.

The existing watertight-tiling tests in test_rasterize_accuracy.py already pin "every pixel exactly once" for `merge='count'`, but only run with the default `all_touched=False`, which is why this never tripped them.

Proposed fix: for `sum`/`count`, deduplicate per geometry. Enumerate boundary/line cells host side (the segment extraction already happens there), drop duplicates per `(geometry, row, col)`, drop boundary cells the geometry's own scanline pass already covers (parity test against the same edge table, same float arithmetic), and burn the survivors through the existing point kernels on all four backends.

Found by the accuracy sweep on 2026-06-12. xrspatial @ 652afdd0, shapely 2.x, rasterio for reference values.

brendancol:
Closes #3304

`merge='count'` reported up to 3 for a single polygon with `all_touched=True`: the scanline fill burned every cell whose center is inside, the supercover boundary pass burned the cells the rings cross (including ones scanline just burned), and consecutive ring segments re-burned their shared vertex cell. Lines had the same problem at connecting vertices, where each Bresenham segment burned both endpoints. rasterio's `MergeAlg.add` returns 1 per geometry per pixel for all of these, even for a self-crossing line.

- For `sum`/`count` only, line and `all_touched` boundary cells are now enumerated host side, deduplicated per (geometry, row, col), and burned through the existing point kernels. Boundary cells the geometry's own scanline fill covers are dropped with a crossing-parity test that replays the fill's float arithmetic on the same edge table, so the kept/dropped decision is bit-consistent with what scanline actually burned.
- Coverage is unchanged on every backend: the burned pixel set still equals `merge='last'`. Only the values change.
- Points are intentionally not deduplicated. GDAL adds once per point, so a MultiPoint with a repeated coordinate burning twice matches rasterio.
- `min`/`max`/`first`/`last` and user callables keep their existing code paths untouched.

Backends: numpy, cupy, dask+numpy, dask+cupy (cupy paths verified on a CUDA host; the dedup work happens host side and feeds the existing atomic point kernel).

Test plan:
- [x] New `test_rasterize_merge_dedup_3304.py`: single-geometry count==1 (polygon all_touched, line vertex, self-crossing line), sum==value, coverage equals `merge='last'`, overlap still accumulates per geometry, rasterio `MergeAlg.add` parity for polygon all_touched and for line values on common coverage, MultiPoint repeat still counts 2, other merges unchanged, chunked equals eager at several tile sizes. 48 tests, all four backends.
- [x] Full rasterize suite: 790 passed, 2 skipped, 1 xfailed.


brendancol (commented):
## PR Review: Burn each geometry once per pixel under merge='sum'/'count'

None found.

None. I went looking for the usual failure modes and came up empty:

- Tie-breaking: re-ran the four #2169 polygons (including `edges_on_pixel_centers`, where cell centers sit exactly on ring edges) with `merge='count'`, `all_touched=True`. Count is 1 everywhere and coverage equals `merge='last'` for all four. The parity filter replaying the fill's own float expression and ceil (`_cells_not_scanline_covered`, rasterize.py:1352) is what makes this hold; a geometric point-in-polygon test would have disagreed with the fill on these exact-tie cells.
- Degenerate inputs: zero-length leading segment, polygon fully outside bounds, and a 1x1 raster all behave.
- Cross-tile dedup: per-tile dedup is globally correct because tiles partition the output cells and the Bresenham/supercover walks are translation invariant. Verified chunked == eager at (3,3), (5,5), (4,7) in the new tests.
- GPU: the dedup'd cells feed the existing atomic point kernel, so overlap between different geometries stays deterministic. `dedup` never coincides with the two-pass first/last modes, so the pass-2 loop over `extra_point_launches` can't fire with stale launches.

- [ ] Memory tradeoff worth a line in the module comment: under sum/count, line cells are now materialized host-side as O(pixels traversed) int32 arrays instead of O(vertices) segment endpoints. For a few very long lines on a huge raster that's the same order as the work the kernel already did, but it's now allocated all at once. (rasterize.py:1110, `_bresenham_cells`)
- [ ] `_dedup_cells` packs (gid, row, col) into one int64; the key overflows only when n_geoms * height * width exceeds 2^63, at which point the cell arrays themselves would have OOM'd long before. Fine as is; a one-line comment on the bound would save the next reader the same arithmetic. (rasterize.py:1306)
- [ ] benchmarks/benchmarks/rasterize.py only exercises the default merge, so the new sum/count path has no benchmark. The hot default path is untouched, so nothing can regress silently there, but a `merge='count'` + `all_touched` case would catch future regressions in the dedup path.

### What looks good
- The parity test reuses the same edge table and the same `x_at_ymin + (row - y_min) * inv_slope` expression as the fill, so kept/dropped decisions are bit-consistent with what scanline actually burned. That's the right way to do this; anything else fights float ties forever.
- Coverage-invariance tests (`count != 0` equals `last != 0`) pin the property that dedup may only change values, never the burned set.
- Points intentionally left un-deduplicated, with a test pinning the rasterio-matching MultiPoint behavior instead of leaving it implicit.
- min/max/first/last and user callables keep their exact prior code paths; the diff is gated on merge identity (`_is_dedup_merge_fn`) which survives pickling into dask workers.

### Checklist
- [x] Algorithm matches reference (rasterio MergeAlg.add parity verified for polygon all_touched exactly; for lines, values on common coverage)
- [x] All implemented backends produce consistent results (numpy/cupy/dask+numpy/dask+cupy, asserted in tests on a CUDA host)
- [x] NaN handling is correct (untouched: NaN burns still poison sum via the existing atomic/CPU paths)
- [x] Dask chunk boundaries handled correctly (per-tile dedup, three chunkings tested)
- [x] No premature materialization on the default path; sum/count host-side cells noted above
- [x] Benchmark exists for rasterize; dedup-path case noted as nit
- [x] Docstrings updated (merge parameter documents the once-per-pixel contract)

brendancol (commented):
## Follow-up review (after 7a9de931)

All three nits from the first pass are addressed:

- `RasterizeMergeCountDedup` benchmark covers both dedup paths (`merge='count'` + `all_touched` polygons, `merge='sum'` lines) on numpy and cupy, matching the parameter grid of the surrounding classes. Setup and both timing methods smoke-tested at nx=100.
- The module comment now states the O(pixels traversed) line-cell materialization tradeoff and that it is per tile on dask.
- `_dedup_cells` documents the int64 key bound.

No behavioral changes in the follow-up commit; the new-test file (48) and the full rasterize suite still pass. Two flake8 F401s in benchmarks/benchmarks/rasterize.py predate this PR and were left alone.

Nothing further from me.

PR #3317, opened 2026-06-14
Serialize all reproject parallel=True kernel launches behind one lock

### LINKED ISSUE #3141: Streaming reproject thread pool aborts the process when numba parallel kernels run concurrently
`_process_tile_batch` (xrspatial/reproject/__init__.py) runs `_reproject_chunk_numpy` for several tiles at once in a ThreadPoolExecutor. Each call can enter the numba fast path (`try_numba_transform` in `xrspatial/reproject/_projections.py`), whose kernels are compiled with `@njit(nogil=True, cache=True, parallel=True)` (`merc_forward`/`merc_inverse` at lines 62/69, `tmerc_forward`/`tmerc_inverse` at 1595/1607, and the rest of the `parallel=True` set).

Calling a `parallel=True` numba function from multiple Python threads at the same time is unsupported on the workqueue threading layer: numba detects the concurrent access and calls `abort()`, killing the interpreter. macOS arm64 wheels use workqueue by default, so the crash shows up there; Linux CI typically has TBB or OpenMP available and doesn't abort.

First observed on the macos-latest / 3.14 job for PR #3118 (the first test to ever drive the streaming threaded branch):

```
Fatal Python error: Aborted

Current thread ... [ThreadPoolExecutor-73_2]:
  File "xrspatial/reproject/_projections.py", line 1829 in try_numba_transform   # tmerc_inverse(...)
  File "xrspatial/reproject/__init__.py", line 365 in _reproject_chunk_numpy
  File "xrspatial/reproject/__init__.py", line 1505 in _do_one
  ...
Thread ... [ThreadPoolExecutor-73_1]:
  File "xrspatial/reproject/_projections.py", line 1829 in try_numba_transform
  ...
```

(Two worker threads inside `tmerc_inverse` at once; C stack ends in numba `call_cfunc` -> abort. Job: https://github.com/xarray-contrib/xarray-spatial/actions/runs/27243353143/job/80451867583)

The comment in `_process_tile_batch` says "Uses ThreadPoolExecutor for intra-worker parallelism (Numba releases the GIL)", which is fine for plain `nogil` kernels but not for `parallel=True` ones. The same hazard applies to the dask.bag branch when a distributed worker uses a threaded executor.

Possible fixes (pick one):
- guard the `try_numba_transform` call with a module-level lock so only one thread runs a parallel kernel at a time (transform is a small share of tile time; resampling stays concurrent),
- compile non-parallel variants of the projection kernels for use from worker threads,
- or detect `numba.threading_layer()` and skip the fast path inside `_process_tile_batch` when it is `workqueue`.

Until then, the streaming tests added for #3101 route their threaded run through `transform_precision=0` (exact pyproj, one Transformer per thread) so CI doesn't hard-crash; the serial streaming tests still cover the numba fast path.

### LINKED ISSUE #3142: Geoid kernels in reproject/_vertical.py can hit numba's workqueue concurrency abort under dask
Follow-up to #3093 / PR #3111.

The macos-3.14 CI abort on PR #3111 came from launching numba `parallel=True` kernels concurrently from multiple host threads: numba's default workqueue threading layer is not threadsafe and terminates the process. PR #3111 serialized the launches in `_projections.py` (`try_numba_transform`, `transform_points`) behind a module lock.

`_vertical.py` has the same pattern and was left out of that PR to keep it focused:

- `_interp_geoid_batch` and `_interp_geoid_2d` are `@njit(parallel=True)`.
- `_apply_vertical_shift_dask` wraps `_apply_vertical_shift_numpy` in `map_blocks`, so dask's threaded scheduler can run two blocks at once, each launching `_interp_geoid_2d` concurrently.
- `geoid_height` / `geoid_height_raster` are public and can be called from user threads directly.

Nothing in the test suite currently triggers this (the vertical-shift dask tests appear to run with few enough concurrent blocks to dodge the race), but it is the same abort waiting to happen. Same fix shape as PR #3111: a module-level lock around the kernel launches, plus a subprocess regression test with `NUMBA_THREADING_LAYER=workqueue` forced.

brendancol:
Closes #3141
Closes #3142

The reproject numba kernels compiled with `parallel=True` (coordinate transforms in `_projections.py`, geoid interpolation in `_vertical.py`, ITRF in `_itrf.py`) abort the interpreter when two host threads launch them at once. Numba's default workqueue threading layer, which is the macOS arm64 wheel default, detects the concurrent parallel region and calls `abort()`. Both the streaming tile pool (`_process_tile_batch`) and dask's threaded scheduler call into these kernels from worker threads.

PR #3111 already serialized `try_numba_transform` and `transform_points` behind `_PARALLEL_KERNEL_LOCK`. This extends that same lock to the remaining `parallel=True` launches in the package and re-enables the streaming threaded numba path that #3101's tests had been routing around.

## Changes
- Wrap the geoid kernel launches (`_interp_geoid_batch`, `_interp_geoid_2d` in `_vertical.py`, and the per-chunk launch in `_apply_vertical_shift_numpy`) with `_PARALLEL_KERNEL_LOCK`. This is the fix for #3142.
- Wrap the `_itrf_batch` launch in `itrf_transform` with the same lock.
- The lock is shared from `_projections.py` because the hazard is any two parallel regions running at once, not just two launches of one kernel.
- Re-enable the streaming threaded test to run through the numba fast path (default precision) instead of `transform_precision=0`, so it covers the #3141 crash directly.

## Backend coverage
The fix is on the CPU numba path. cupy and dask+cupy do not use these host-thread numba kernels. dask+numpy benefits because dask's threaded scheduler is one of the concurrent callers.

## Test plan
- New `test_reproject_parallel_kernels_3141.py` hammers each entry point (transform, geoid, ITRF) from an 8-thread pool and checks results match the single-threaded baseline.
- Confirmed the regression suite passes under `NUMBA_THREADING_LAYER=workqueue`, the layer that aborts without the lock.
- Streaming and existing thread-safety tests pass.


brendancol (commented):
## PR Review: Serialize all reproject parallel=True kernel launches behind one lock

I read the full files, not just the diff, and traced the lock acquisition paths for deadlock and correctness.

### Blockers
None.

### Suggestions
None blocking. One thing I checked and want to record: the lock is now shared across `_projections.py`, `_vertical.py`, and `_itrf.py`. I confirmed there is no re-entrant acquisition. Each public entry point takes the lock, runs only the kernel inside the `with` block, and the kernels never call back into another locked entry point. `_apply_vertical_shift_numpy` calls `_interp_geoid_2d` directly rather than going through `geoid_height`, so the dask path does not nest the lock either. `threading.Lock` is non-reentrant, so this matters, and it is correct as written.

### Nits
- The three module-level comment blocks above the `_PARALLEL_KERNEL_LOCK` imports in `_vertical.py` and `_itrf.py` are near-identical copies of each other. Fine to leave; they each stand alone and explain the shared-lock rationale at the import site.

### What looks good
- The lock wraps only the kernel launch, not the surrounding coordinate setup, inverse projection, or accumulation. Transform and resampling stay concurrent, so the serialization cost is the small share of time the parallel region actually runs.
- Sharing one lock across all the parallel kernels is the right call. The workqueue hazard is two parallel regions running at once, not two launches of one kernel, so a per-kernel lock would not be enough.
- The streaming threaded test now runs the numba fast path at default precision instead of routing around it with `transform_precision=0`, so it exercises the actual #3141 crash path.
- The new test hammers each entry point from an 8-thread pool and compares against the single-threaded baseline, and it is guarded for missing pyproj and missing PROJ ITRF files. No temp files.

### Verification I ran
- `test_reproject_parallel_kernels_3141.py` and `TestParallelKernelThreadSafety` pass under `NUMBA_THREADING_LAYER=workqueue`, the layer that aborts without the lock.
- Streaming and existing thread-safety tests pass.

### Checklist
- [x] No deadlock or re-entrant lock acquisition
- [x] Lock held only around kernel launches, not heavy compute
- [x] Concurrent callers get correct independent results (tested)
- [x] Regression passes under the workqueue threading layer
- [x] Tests guarded for optional deps; no temp files
- [x] No CHANGELOG edits
- [ ] README / docs: not applicable (pure bug fix, no public API change)

PR #3318, opened 2026-06-14
Promote large-output in-memory cupy reproject inputs to dask (#3281)

### LINKED ISSUE #3281: reproject(): in-memory cupy inputs skip the output-size dask promotion
Follow-up from #3267 / PR #3278 review.

The output-size promotion added in #3278 applies to numpy inputs only (`if not is_dask and not is_cupy` in `reproject()`). An in-memory cupy raster with a very large output still allocates the full working set on the GPU and fails with a device OOM, same as before #3278.

Wrapping cupy inputs in dask the same way would route them through `_reproject_dask_cupy`, which already has a VRAM check and a chunked `map_blocks` fallback. The catch: the result type for in-memory cupy inputs would change from cupy to dask-of-cupy above the threshold, and the promotion needs `is_cupy_backed` detection after the wrap plus GPU OOM testing. Needs a design decision on the result-type contract before implementing.

brendancol:
Closes #3281

The output-size dask promotion from #3278 only covered numpy inputs. The promotion branch in `reproject()` was gated on `not is_dask and not is_cupy`, so an in-memory cupy raster with a very large output still allocated the full working set on the GPU and could hit a device OOM, same as before #3278.

This drops the `not is_cupy` part of the guard. A large-output cupy input now gets wrapped in a cupy-backed dask array, which keeps `is_cupy` True and routes it through `_reproject_dask_cupy`. That helper already has a GPU VRAM check and a chunked `map_blocks` fallback.

On the result-type contract (the design question the issue flagged), this matches what `merge()` already does for promoted cupy mosaics:
- Output still fits in VRAM: stays eager cupy (no change from today).
- Output exceeds VRAM: becomes a chunked dask result, computed block-by-block on the GPU.

The streaming fallback stays numpy-only, so a cupy input with dask unavailable keeps its eager GPU path instead of moving data off the device.

Scope is limited to the cupy output-size promotion path. It does not touch the numba-concurrency path under separate review in #3141/#3142.

Backends: cupy (in-memory promotion); numpy / dask+numpy / dask+cupy paths unchanged.

Test plan:
- [x] New regression tests in `test_reproject_cupy_promotion_3281.py` (skipped without cupy + CUDA): below-threshold stays eager cupy; above-threshold routes through `_reproject_dask_cupy`; VRAM-exceeds fallback yields a chunked dask result with numerical parity to the eager path.
- [x] Existing reproject suite green (508 passed).

### DIFF
--- a/xrspatial/reproject/__init__.py
+++ b/xrspatial/reproject/__init__.py
@@ -1008,7 +1008,7 @@ def reproject(
     # blockwise layer) so graph metadata is no longer a concern -- the
     # streaming fallback is only needed when dask itself is unavailable.
     _use_streaming = False
-    if not is_dask and not is_cupy:
+    if not is_dask:
         nbytes = src_shape[0] * src_shape[1] * data.dtype.itemsize
         out_nbytes = out_shape[0] * out_shape[1] * 8  # float64 working set
         if data.ndim == 3:
@@ -1025,6 +1025,13 @@ def reproject(
                 cs = (cs[0], cs[1], data.shape[2])
             try:
                 import dask.array as _da
+                # Wrapping an in-memory cupy array keeps the blocks
+                # cupy-backed, so is_cupy stays True and the promoted
+                # raster routes through _reproject_dask_cupy -- which has
+                # its own GPU VRAM check and chunked map_blocks fallback.
+                # Above the threshold an in-memory cupy input therefore
+                # yields a dask-of-cupy output, matching merge()'s
+                # promotion contract (#3281).
                 data = _da.from_array(data, chunks=cs)
                 raster = xr.DataArray(
                     data, dims=raster.dims, coords=raster.coords,
@@ -1032,8 +1039,12 @@ def reproject(
                 )
                 is_dask = True
             except ImportError:
-                # dask not available -- fall back to streaming
-                _use_streaming = True
+                # dask not available. The streaming fallback is a numpy
+                # CPU path, so it only applies to numpy inputs; a cupy
+                # input keeps the eager GPU path rather than silently
+                # moving its data off the device (#3281).
+                if not is_cupy:
+                    _use_streaming = True
 
     # Re-apply the output-size guard for paths that materialize the whole
     # output (in-memory numpy/cupy and streaming). A lazy dask output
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Promote large-output in-memory cupy reproject inputs to dask (#3281)

### Blockers
None.

### Suggestions
None blocking. The change is a one-condition relaxation of an existing guard, and the routing it enables (`_reproject_dask_cupy` with its VRAM check and chunked fallback) already existed and is tested.

### Nits
- `_REPROJECT_OOM_THRESHOLD` comment (`__init__.py:53`) still says "the eager numpy path holds ~7 output-sized float64 temporaries". The threshold now also gates the cupy promotion, so the comment is slightly narrower than the code. Not worth a reword on its own since the 7x reasoning is what justifies the budget and that reasoning is numpy-specific; leaving it as the rationale-of-record is defensible. Flagging only so it is a conscious choice.

### What looks good
- Minimal surface: drops `not is_cupy` from one branch and adds a guard so the numpy-only streaming fallback is not reused for cupy when dask is missing. No new code paths.
- Result-type contract is the right call. Reusing merge()'s precedent (eager cupy below VRAM, dask-of-cupy above) keeps the two functions consistent and is documented inline at the wrap site.
- Streaming guard at `__init__.py:1044` correctly keeps cupy data on the device rather than silently moving it to the numpy CPU path.
- Tests cover the three meaningful states: below-threshold stays eager cupy, above-threshold routes through the dask+cupy helper (verified by spying on both helpers), and the VRAM-exceeds fallback returns a chunked dask result with numerical parity to the eager path. All gated on `has_cuda_and_cupy()` so they skip cleanly without a GPU.

### Checklist
- [x] Algorithm unchanged (routing-only change); existing reproject math untouched
- [x] Implemented backend (cupy) routes correctly; numpy/dask paths unchanged
- [x] NaN handling unchanged (parity test uses equal_nan)
- [x] Edge cases: below/above threshold and VRAM fallback all covered
- [x] Dask chunk boundaries handled by the existing `_reproject_dask_cupy` path
- [x] No premature materialization; promotion keeps data lazy above VRAM
- [x] Benchmark not needed (no new function, routing change)
- [x] README feature matrix not applicable (cupy was already supported)
- [x] Docstrings/comments present at the wrap site

brendancol (commented):
## Follow-up review (after 9eb3aef8)

The one nit from the first pass is addressed: the `_REPROJECT_OOM_THRESHOLD` comment now states the budget gates both numpy and cupy promotion and references #3281, while keeping the numpy-specific 7x rationale as the justification for the budget size.

No remaining findings. New tests still pass (3 passed), routing and parity behavior unchanged.

PR #3319, opened 2026-06-14
focal: handle empty rasters consistently across backends (#3225)

### LINKED ISSUE #3225: focal: empty raster (0 rows) works on numpy but crashes on cupy and dask backends

brendancol:
Closes #3225

## Problem

A raster with a 0-length spatial axis (e.g. shape `(0, 5)`) behaved differently per backend. numpy returned an empty result with the input shape preserved, but cupy crashed with a raw `CudaAPIError` from a zero-sized kernel grid, and the dask backends raised `ValueError: The overlapping depth 1 is larger than your array 0` from `map_overlap`.

## Change

- `mean()`, `apply()`, and `focal_stats()` short circuit an empty raster and return an empty result of the input shape on every backend, matching the numpy path. The result keeps the input backend type (numpy stays numpy, dask stays lazy).
- `hotspots()` raises the same "needs at least 2 valid (non-NaN) cells" error numpy already raised, up front on every backend, instead of crashing in a zero-sized cupy reduction or a dask overlap.

## Backend coverage

## Test plan

- [x] New `test_focal_empty_raster_3220` covers mean/apply/focal_stats across all four backends and shapes `(0,5)`, `(5,0)`, `(0,0)`; checks shape and backend type are preserved.
- [x] New `test_hotspots_empty_raster_3225` covers hotspots raising the clear error across all four backends and the same shapes.
- [x] Full `test_focal.py` suite passes (351 passed). GPU tests run here and skip when cupy is unavailable.

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -33,6 +33,19 @@ from xrspatial.utils import (ArrayTypeFunctionMapping, _boundary_to_dask, _pad_a
                              is_cupy_array, is_dask_cupy, ngjit)
 
 
+def _is_empty_raster(agg):
+    """True if *agg* has a degenerate spatial axis (0 rows or 0 cols).
+
+    An empty raster has no cells to filter, but the GPU and dask paths
+    crash on it rather than no-op: ``cuda_args`` hands ``cuLaunchKernel``
+    a zero-sized grid, and ``map_overlap`` rejects a depth that exceeds
+    the (zero-length) axis. The numpy path already returns an empty
+    result with the input shape preserved, so the focal APIs short
+    circuit empty input to match it on every backend (issue #3225).
+    """
+    return agg.shape[-2] == 0 or agg.shape[-1] == 0
+
+
 def _validate_binary_kernel(kernel, func_name):
     """Reject non-binary kernels for the mask-based focal APIs.
 
@@ -441,6 +454,16 @@ def mean(agg, passes=1, excludes=None, name='mean', boundary='nan'):
         return _apply_per_band(mean, agg, passes=passes, excludes=excludes,
                                name=name, boundary=boundary)
 
+    if _is_empty_raster(agg):
+        # No cells to filter; return an empty result of the input shape
+        # and dtype contract, matching the numpy path on every backend
+        # (issue #3225).
+        return DataArray(agg.data.astype(_promote_float(agg.dtype)),
+                         name=name,
+                         dims=agg.dims,
+                         coords=agg.coords,
+                         attrs=agg.attrs)
+
     # Preserve the input float dtype, promoting ints to float32 -- the same
     # contract as apply() / focal_stats() (#2769) and convolve_2d() (#1096).
     # Cast excludes to the working dtype so value matching behaves the same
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: focal empty-raster consistency (#3225)

None.

None. The fix is small and the tests cover the matrix that matters.

- focal.py: `mean` and `apply` build the empty result with slightly different keyword order (`dims, coords` vs `coords, dims`). Each guard matches its own function's main return path, so this is cosmetic only.
- The two F401 "cupy imported but unused" flake8 warnings in test_focal.py (around lines 1894 and 1910) are pre-existing on main, not from this PR. Noting so they aren't mistaken for new.

### What looks good
- The empty-raster guard sits after kernel and memory validation in apply/focal_stats, so an empty raster with a bad kernel still raises the kernel error first, matching numpy's validate-before-compute order.
- `.astype` on a zero-element array is kernel-free on cupy and stays lazy on dask, so no GPU launch or compute is triggered. The input backend type is preserved (dask stays dask).
- focal_stats reuses apply's short circuit to build the stacked result, so the band count and stats index match the real path without duplicating the concat logic.
- hotspots reuses the exact error string numpy already raised, so all four backends fail the same way up front instead of crashing in a zero-sized reduction or overlap.
- Tests parametrize all four backends across (0,5), (5,0), (0,0) and assert both shape and backend type. GPU cases skip when cupy is absent.

### Checklist
- [x] Behavior matches the numpy reference on every backend
- [x] NaN handling unaffected (empty input has no cells)
- [x] Edge cases covered by tests (three empty shapes, four backends)
- [x] Dask handled correctly (stays lazy, no map_overlap on empty axis)
- [ ] Benchmark not needed for an edge-case guard
- [ ] README feature matrix not needed (no new function or backend change)
- [x] Docstrings/comments present and accurate

PR #3322, opened 2026-06-14
Burn all_touched lines with the supercover walk (#3102)

### LINKED ISSUE #3102: rasterize(all_touched=True) has no effect on LineString geometries

### LINKED ISSUE #3105: rasterize: no tests for all_touched with line input or the non-iterable input error path

brendancol:
## Summary

`rasterize(all_touched=True)` had no effect on LineString / MultiLineString input. Lines went through the Bresenham burner regardless of the flag, so `all_touched=True` returned the same pixels as `all_touched=False` on all four backends. The supercover machinery already existed for polygon boundaries, so this routes lines through it when `all_touched` is set.

- Added `_extract_line_segments_float`: float pixel-space line segments that feed the supercover (Amanatides & Woo) grid walk, the line counterpart to the polygon boundary extractor.
- Routed lines through the supercover walk under `all_touched=True` in the numpy, cupy, dask+numpy, and dask+cupy paths, including the `merge='sum'`/`'count'` per-geometry cell dedup.
- Updated the `all_touched` docstring to cover lines.

Closes #3102
Closes #3105

## Backend coverage

## Test plan

- [x] Off-corner single-segment lines match `rasterio.features.rasterize(all_touched=True)` pixel-for-pixel.
- [x] dask+numpy and cupy results equal eager numpy.
- [x] `merge='sum'` on a self-crossing line counts each pixel once per geometry.
- [x] The non-iterable input TypeError path stays covered (it lives in the sweep file).
- [x] Updated the test-coverage sweep file that pinned the old no-op behavior and its strict xfail.
- [x] Full rasterize suite passes (855 passed, 2 skipped).

Exact cell-corner crossings can still differ from rasterio by a cell or two. That is the same tie-breaking wiggle the polygon supercover parity tests already allow.

### DIFF
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -764,6 +764,98 @@ def _extract_lines_vectorized(geometries, bounds, height, width):
     return (r0[v], c0[v], r1[v], c1[v], seg_geom_idx[v])
 
 
+def _extract_line_segments_float(geometries, bounds, height, width):
+    """LineString/MultiLineString segments in float pixel space.
+
+    Variant of :func:`_extract_line_segments` that keeps sub-pixel
+    precision so the supercover grid walker can pick up every cell a
+    segment crosses (issue #3102, the line counterpart to the polygon
+    boundary path).  World-space clipping uses the same Liang-Barsky
+    pass; the result is converted to pixel coordinates
+    (``cx`` = x in pixels, ``cy`` = y in pixels, increasing downward)
+    without rounding.
+
+    Returns ``(cx0, cy0, cx1, cy1, geom_idx)`` as float64 arrays (and
+    int32 ids into the per-type props table).
+    """
+    if not geometries:
+        return _EMPTY_LINES_FLOAT
+
+    shapely = _require_shapely()
+    xmin, ymin, xmax, ymax = bounds
+    px = (xmax - xmin) / width
+    py = (ymax - ymin) / height
+
+    geom_arr = np.array(geometries, dtype=object)
+    idx_arr = np.arange(len(geometries), dtype=np.int32)
+
+    # Explode MultiLineStrings to individual LineStrings
+    parts, part_idx = shapely.get_parts(geom_arr, return_index=True)
+    part_geom_idx = idx_arr[part_idx]
+
+    if len(parts) == 0:
+        return _EMPTY_LINES_FLOAT
+
+    coords, coord_line_idx = shapely.get_coordinates(
+        parts, return_index=True)
+    n_coords = len(coords)
+    if n_coords < 2:
+        return _EMPTY_LINES_FLOAT
+
+    # Mark last coordinate of each line (don't form cross-line segments)
+    is_last = np.zeros(n_coords, dtype=bool)
+    changes = np.nonzero(np.diff(coord_line_idx))[0]
+    is_last[changes] = True
+    is_last[-1] = True
+
+    start_idx = np.nonzero(~is_last)[0]
+    end_idx = start_idx + 1
+    seg_geom_idx = part_geom_idx[coord_line_idx[start_idx]].astype(np.int32)
+
+    x0 = coords[start_idx, 0].astype(np.float64)
+    y0 = coords[start_idx, 1].astype(np.float64)
+    x1 = coords[end_idx, 0].astype(np.float64)
+    y1 = coords[end_idx, 1].astype(np.float64)
+
+    # Vectorized Liang-Barsky clip to raster world-space bounds.
+    dx = x1 - x0
+    dy = y1 - y0
+    p = np.array([-dx, dx, -dy, dy])
+    q = np.array([x0 - xmin, xmax - x0, y0 - ymin, ymax - y0])
+
+    t0 = np.zeros(len(x0))
+    t1 = np.ones(len(x0))
+    valid = np.ones(len(x0), dtype=bool)
+
+    for i in range(4):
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Burn all_touched lines with the supercover walk (#3102)

Reviewed the full rasterize.py diff and both test files against the existing polygon-boundary supercover path.

### Blockers
None.

### Suggestions
- `_float_segments_for_tile` (rasterize.py) reuses the integer `seg_bboxes` to pick which float segments belong to a tile. This is correct only because `_extract_line_segments` and `_extract_line_segments_float` share the same Liang-Barsky `valid` mask and so emit segments in the same order with the same `geom_idx`. That coupling is load-bearing and easy to break later. I verified it holds today (same count, same order, float-floor equals the int endpoints), and the comments call it out, so this is a watch-it note rather than a fix.

### Nits
- The float extractor duplicates the Liang-Barsky clip block already present in `_extract_lines_vectorized` and `_extract_polygon_boundary_segments_float`. Three copies now. Not worth refactoring in this PR (the polygon and line variants differ in their pixel-space conversion and a shared helper would need to thread several flags), but a future cleanup could factor the clip out.

### What looks good
- The fix reuses the existing supercover kernels (CPU `_burn_lines_supercover_cpu`, GPU boundary kernel, `_supercover_cells` for dedup) instead of adding new traversal code, so line and polygon all_touched share one walk.
- All four backends are covered, and the dask path keeps lines lazy by tiling float segments through the same mask as the integer segments rather than materializing.
- merge='sum'/'count' dedup is handled: a self-crossing line counts each pixel once per geometry, matching GDAL.
- Tests check rasterio parity on off-corner segments, cross-backend equality, and the corner tie-break wiggle is documented rather than asserted. The old no-op pins and the strict xfail in the sweep file were flipped, so CI reflects the new contract.

### Checklist
- [x] Algorithm matches the polygon supercover path it reuses
- [x] All four backends produce consistent results (verified numpy == dask+numpy pixel-for-pixel)
- [x] NaN/empty handled (empty-line and fully-clipped cases return early)
- [x] Edge cases covered (off-corner, corner-crossing, multi-segment, self-crossing)
- [x] Dask chunk boundaries handled (float segments filtered by the same per-tile mask)
- [x] No premature materialization in the dask path
- [x] Benchmark not needed (reuses existing kernels)
- [x] README/feature matrix unchanged (no new public function)
- [x] Docstring updated to cover lines

PR #3324, opened 2026-06-14
Type-check compression_level in to_geotiff

### LINKED ISSUE #3321: to_geotiff compression_level is not type-safe
**Describe the bug**

`to_geotiff` documents `compression_level` as `int | None`, but the validation in `xrspatial/geotiff/_writers/eager.py` (around lines 663-670) only does a range comparison. It never checks the type. Three things fall out of that:

- `compression_level=True` is accepted and used as level 1, because `bool` is a subclass of `int` in Python.
- A string or list value (e.g. `'9'` or `[9]`) produces a raw `TypeError` straight out of the `lo <= compression_level <= hi` comparison instead of a clear message that names the parameter.
- For codecs whose `_LEVEL_RANGES` entry is absent (anything other than deflate/zstd/lz4), the range block is skipped, so junk passes validation and is later ignored.

The rest of this module already guards its I/O-boundary parameters against this. `nodata` rejects `bool`/`np.bool_`, `_validate_overview_level_arg` rejects bool and non-int, and `_validate_chunks_arg` does the same. `compression_level` is the odd one out.

**Expected behavior**

Validate `compression_level` for type at the I/O boundary, the same way `tile_size`, `nodata`, `overview_level`, and `band` already are:

- Reject `bool`, so `compression_level=True` is not coerced to 1.
- Reject any non-`int` value with a `TypeError` that names the parameter, instead of a raw comparison `TypeError`.
- Run this type check for every codec, including ones with no level range, so non-int junk is rejected regardless of codec.
- Leave the existing out-of-range `ValueError` behavior unchanged for valid ints.

**Steps to reproduce**

```python
import numpy as np

arr = np.zeros((4, 4), dtype=np.float32)

# accepted as level 1
to_geotiff(arr, 'a.tif', compression='deflate', compression_level=True)

# raw, confusing TypeError from the range comparison
to_geotiff(arr, 'b.tif', compression='deflate', compression_level='9')

# accepted and ignored: lzw has no _LEVEL_RANGES entry
to_geotiff(arr, 'c.tif', compression='lzw', compression_level=[9])
```

**Additional context**

Validation runs in `to_geotiff` before backend dispatch, so the fix covers numpy, cupy, dask+numpy, and dask+cupy through one code path.

brendancol:
Closes #3321

`to_geotiff` documents `compression_level` as `int | None`, but the
validation only did a range comparison. That left three holes:

- `compression_level=True` was accepted as level 1 (`bool` is an int subclass).
- A string or list raised a raw `TypeError` from the `lo <= compression_level <= hi` comparison instead of a clear message.
- Codecs with no `_LEVEL_RANGES` entry (e.g. lzw) skipped the check, so junk passed through and was ignored.

This adds `_validate_compression_level_arg` in `_validation.py`, mirroring
the existing `_validate_overview_level_arg` guard, and calls it in
`to_geotiff` before the range check. It rejects `bool` and any non-int
with a `TypeError` that names the parameter, for every codec. Valid ints
keep the existing out-of-range `ValueError` behavior. Well-typed ints for
codecs without a level range are still silently ignored, matching how the
rest of the module treats inapplicable-but-well-typed params.

Backend coverage: the guard runs in `to_geotiff` before backend dispatch,
so it covers numpy, cupy, dask+numpy, and dask+cupy through one path.

Test plan:
- [x] `compression_level=True` raises `TypeError`
- [x] `'9'`, `[9]`, `9.0`, `object()` raise `TypeError`
- [x] non-int rejected for lzw (no level range)
- [x] valid int still writes
- [x] out-of-range int still raises `ValueError`
- [x] existing writer + codec-roundtrip suites pass (429 + streaming)

### DIFF
--- a/xrspatial/geotiff/_validation.py
+++ b/xrspatial/geotiff/_validation.py
@@ -375,6 +375,33 @@ def _validate_overview_level_arg(overview_level) -> None:
         )
 
 
+def _validate_compression_level_arg(compression_level) -> None:
+    """Validate the ``compression_level`` kwarg type.
+
+    Accepts ``None`` (default, codec picks its own level) and any
+    ``int`` / ``numpy.integer`` instance. ``bool`` is rejected
+    explicitly because Python treats it as a subclass of ``int``;
+    without the check, ``compression_level=True`` is coerced to ``1``
+    and silently used as level 1. Non-int types like ``str`` and
+    ``list`` would otherwise leak a raw ``TypeError`` from the
+    ``lo <= compression_level <= hi`` range comparison, or pass through
+    unchecked for codecs with no ``_LEVEL_RANGES`` entry.
+
+    Runs in ``to_geotiff`` before the codec-specific range check and
+    before backend dispatch, so every backend (numpy, cupy, dask) sees
+    the same rejection. The out-of-range ``ValueError`` for valid ints
+    is unchanged (#3321).
+    """
+    if compression_level is None:
+        return
+    if (not isinstance(compression_level, (int, np.integer))
+            or isinstance(compression_level, bool)):
+        raise TypeError(
+            f"compression_level must be an int or None, got "
+            f"{type(compression_level).__name__}: {compression_level!r}"
+        )
+
+
 def _validate_gpu_arg(gpu, *, allow_none: bool = False) -> None:
     """Validate the ``gpu`` dispatch flag type.
 
--- a/xrspatial/geotiff/_writers/eager.py
+++ b/xrspatial/geotiff/_writers/eager.py
@@ -34,7 +34,8 @@ from .._geotags import RASTER_PIXEL_IS_AREA, GeoTransform
 from .._nodata import NodataLifecycle as _NL
 from .._runtime import (GeoTIFFFallbackWarning, _geotiff_strict_mode, _gpu_fallback_warning_message,
                         _resolve_spatial_coords)
-from .._validation import (_validate_3d_writer_dims, _validate_gpu_arg, _validate_no_rotated_affine,
+from .._validation import (_validate_3d_writer_dims, _validate_compression_level_arg,
+                           _validate_gpu_arg, _validate_no_rotated_affine,
                            _validate_nodata_arg, _validate_tile_size_arg,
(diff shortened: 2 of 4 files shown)


brendancol (commented):
## PR Review: Type-check compression_level in to_geotiff

### Blockers
None.

### Suggestions
None.

### Nits
- `_validation.py` / `eager.py`: the new `_validate_compression_level_arg` is wired into `to_geotiff` only. The second range check in `_write_vrt_tiled` (eager.py ~1273) and the GPU writer's own range check (gpu.py ~395) still have no type guard. Both are internal and only reachable through `to_geotiff`, which now validates first, so there is no live bug here. A one-line comment at those two sites noting the type guard runs upstream would save a future direct caller some confusion. Not blocking.

### What looks good
- The new helper matches `_validate_overview_level_arg`: same None passthrough, same `(int, np.integer)` plus `not bool` check, same TypeError message shape. Reads cleanly against the existing precedent.
- Placement is right. It runs after compression-name validation (so `compression` is already a known-good string) and before the range check and backend dispatch, so one path covers numpy, cupy, and dask.
- Tests cover bool, str, list, float, and an arbitrary object, plus the no-level-range codec (lzw), a valid int that still writes, and an out-of-range int that still raises ValueError. Temp files carry the issue number.
- Keeping the silent-ignore behavior for well-typed-but-inapplicable ints (rather than rejecting them the way max_z_error does) is the minimal change and matches the rest of the module. The PR body says so.

### Checklist
- Algorithm matches reference: n/a (input validation)
- Backends consistent: yes, single pre-dispatch path
- NaN handling: n/a
- Edge cases covered: yes
- Dask boundaries: n/a
- Benchmark: not needed

brendancol (commented):
## PR Review (follow-up pass)

The one nit from the first pass is addressed: the internal range checks in `_write_vrt_tiled` (eager.py) and `_write_geotiff_gpu` (gpu.py) now carry a one-line note that `to_geotiff` rejects non-int / bool `compression_level` upstream (#3321), so those comparisons are type-safe on the normal dispatch path.

No new findings. Comment-only changes since the last pass; the write + codec-roundtrip suites still pass.

### Blockers
None.

### Suggestions
None.

### Nits
None remaining.

PR #3325, opened 2026-06-14
geotiff: don't flip masked_nodata True on a caller dtype= cast of an unmasked buffer (#3323)

### LINKED ISSUE #3323: geotiff: masked_nodata can be True after an unmaskable nodata sentinel plus an explicit float dtype cast
## Describe the bug

`open_geotiff(..., masked=True)` and the dask read paths can stamp `attrs['masked_nodata'] = True` even when the reader converted no sentinel pixels to NaN. The attr is meant to signal that the reader actually masked the declared sentinel to NaN. When it reads True without that having happened, writers that consult `masked_nodata` to decide whether NaNs should be restored to the sentinel can rewrite NaN pixels in downstream derived arrays back to the sentinel value.

The flag is computed from the final buffer dtype, taken after the caller's explicit `dtype=` cast. A caller cast to a float dtype flips `masked_nodata` to True regardless of whether masking ran.

## To Reproduce

1. A uint16 GeoTIFF with a fractional or out-of-range nodata declared, e.g. `30.5`. No uint16 pixel can equal it, so it is unmaskable.
2. Read it with `allow_invalid_nodata=True, masked=True, dtype='float32'`.
3. `_apply_eager_nodata_mask` leaves the buffer as uint16, since the sentinel can never match an integer pixel, so nothing becomes NaN.
4. The explicit `astype('float32')` then makes the final dtype float.
5. `masked` is computed as `effective_mask and final_dtype.kind == 'f'`, which comes out True even though no pixel was masked.

Expected: `attrs['masked_nodata']` is False, because no pixel was converted.

## Root cause

The float-ness that should drive `masked_nodata` is the float-ness from masking promotion (an integer source promoted to float64 when the sentinel is maskable), not the float-ness from an explicit caller `dtype=` cast. The eager finalizer `_finalize_eager_read` in `xrspatial/geotiff/_attrs.py` reads the post-cast dtype. The dask CPU backend (`_backends/dask.py`) and the dask GPU backend (`_backends/gpu.py`) pass the post-cast graph dtype to `_finalize_lazy_read_attrs`. The VRT path already passes the pre-cast dtype and gets this right.

## Fix direction

Capture the buffer's float-ness before the explicit `dtype=` cast and use that to compute `masked_nodata`, mirroring the VRT path. Keep the existing behavior intact: an int buffer with an unmaskable sentinel stays `masked_nodata=False`, and a maskable sentinel promoted to float stays `masked_nodata=True`.

## Environment

- xarray-spatial: main
- module: geotiff

brendancol:
Closes #3323

`masked_nodata` is meant to say the reader actually converted sentinel pixels to NaN. It was being computed from the final buffer dtype, taken after the caller's explicit `dtype=` cast, so a cast to a float dtype flipped `masked_nodata` to True even when masking ran on nothing.

The trigger: a uint16 file with an unmaskable sentinel (fractional, out-of-range, or non-finite) read with `masked=True, allow_invalid_nodata=True, dtype='float32'`. No pixel can match the sentinel, the buffer stays integer and unmasked, and then the caller cast makes it float. Downstream writers that consult `masked_nodata` to decide whether NaNs should be restored to the sentinel could rewrite NaN pixels in derived arrays back to the sentinel.

## What changed

The flag now reads the buffer's float-ness from before the caller cast, so it reflects masking-promotion only:

- Eager (`_finalize_eager_read` in `_attrs.py`): capture the pre-cast dtype kind and use it for `masked`.
- dask+numpy backend (`_backends/dask.py`): pass `effective_dtype` (pre-cast) as `graph_dtype`.
- dask+GPU backend (`_backends/gpu.py`): pass the pre-cast dtype as `graph_dtype`.

The VRT path already threaded its pre-cast dtype, so it was correct and is unchanged. `nodata_dtype_cast` still records the caller's cast as before.

## Backend coverage

numpy, dask+numpy, and dask+GPU all had the bug and are fixed. cupy eager runs through the same eager finalizer. VRT was already correct.

## Test plan

- [x] New tests assert `masked_nodata is False` for the unmaskable-sentinel + dtype-cast case on eager, dask+numpy, and dask+GPU.
- [x] Positive guards: maskable integer sentinel stays True through a cast; a genuinely-masked float source stays True through a widening cast.
- [x] Full geotiff suite passes locally (6453 passed, 68 skipped, 1 xfailed), GPU included.

### DIFF
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -2211,6 +2211,11 @@ def _finalize_eager_read(
     # ``_validation``; local import keeps ``_attrs`` free of a top-level
     # validation dependency for parity with ``_validate_read_geo_info``.
     dtype_cast_attr: str | None = None
+    # Capture float-ness BEFORE the caller's ``dtype=`` cast so ``masked``
+    # below reflects masking-promotion, not the cast (issue #3323). A
+    # caller casting an unmasked int buffer to float must not flip
+    # ``masked_nodata`` to True.
+    pre_cast_is_float = np.dtype(str(arr.dtype)).kind == 'f'
     if dtype is not None:
         from ._validation import _validate_dtype_cast
         target = np.dtype(dtype)
@@ -2219,16 +2224,18 @@ def _finalize_eager_read(
         dtype_cast_attr = target.name
 
     # Stamp the nodata lifecycle attrs. ``masked`` is True iff
-    # the caller opted into masking AND the final buffer dtype is float.
+    # the caller opted into masking AND the pre-cast buffer dtype is float.
     # ``_apply_eager_nodata_mask`` promotes a maskable integer source to
     # float64 whenever masking is on (issue #2990), so an ``int`` buffer +
     # ``mask_nodata=True`` here means the sentinel was unmaskable
     # (out-of-range / non-finite / fractional) and could never match, not
     # that masking was disabled. Either way ``masked`` is correctly False
-    # because the literal sentinel still occupies its integer slot.
+    # because the literal sentinel still occupies its integer slot. Using
+    # the pre-cast dtype keeps a caller ``dtype=<float>`` cast from
+    # flipping ``masked`` True on that unmasked int buffer (issue #3323).
     _set_nodata_attrs(
         attrs, nodata,
-        masked=(effective_mask and np.dtype(str(arr.dtype)).kind == 'f'),
+        masked=(effective_mask and pre_cast_is_float),
         pixels_present=nodata_pixels_present,
         dtype_cast=dtype_cast_attr,
     )
--- a/xrspatial/geotiff/_backends/dask.py
+++ b/xrspatial/geotiff/_backends/dask.py
@@ -521,10 +521,15 @@ def _read_geotiff_dask(source: str, *,
     # Share the validate-then-populate-then-stamp
     # block with the dask+GPU backend via ``_finalize_lazy_read_attrs``.
     #
-    # ``graph_dtype`` is the resolved dask graph dtype so
-    # ``masked_nodata`` reflects whether the per-chunk mask actually
-    # ran (masking on an int source auto-promotes to float64; an
-    # un-promoted int graph means masking didn't run).
+    # ``graph_dtype`` is ``effective_dtype`` -- the pre-cast graph dtype
+    # after the masking-induced int->float64 auto-promotion but BEFORE
+    # the caller's ``dtype=`` cast -- so ``masked_nodata`` reflects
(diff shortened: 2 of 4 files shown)


brendancol (commented):
## PR Review: don't flip masked_nodata True on a caller dtype= cast of an unmasked buffer (#3323)

None.

None.

- `test_dask_gpu_unmaskable_sentinel_plus_dtype_cast_masked_false` (test_nodata.py) checks only the attrs, not the pixel buffer. The eager and dask+numpy siblings also assert the computed values stay `[[10,20],[30,40]]` with no NaN. Adding the same value check on the GPU case would make it a full parity test rather than an attrs-only one. Left optional since the attr is the thing the fix actually changes.

### What looks good
- The fix is the same shape on all three affected paths: capture the buffer's float-ness from before the caller `dtype=` cast and feed that to `masked`. The eager path uses `pre_cast_is_float`; the dask+numpy path passes `effective_dtype`; the dask+GPU GDS path passes `pre_cast_dtype`. That matches what the VRT path was already doing, so the four read paths now agree.
- `effective_dtype` is the right pre-cast value for the dask+numpy path: it already carries the masking-induced int->float64 promotion and the `mask_and_scale` scale/offset promotion, so a genuine masking promotion still reports `masked=True` while a bare caller cast does not.
- `nodata_dtype_cast` still records the caller's cast, so the caller-intent signal is untouched.
- Tests cover the negative case (unmaskable sentinel + cast -> False) and two positive guards (maskable int sentinel stays True; genuinely-masked float source stays True through a widening cast), on eager, dask+numpy, and dask+GPU. Temp filenames carry the issue number.
- The dask+GPU `_read_geotiff_gpu_chunked` dispatcher reaches the fix through both branches: the GDS fast path (fixed directly) and the CPU-decode fallback (inherits attrs from the fixed `_read_geotiff_dask`).

### Checklist
- [x] Logic matches the stated fix on every path
- [x] All affected backends produce consistent masked_nodata
- [x] NaN handling unchanged
- [x] Edge cases covered by tests (unmaskable vs maskable sentinel, float source)
- [x] Benchmark not needed (attr-stamping fix, no perf surface)
- [x] Comments present and accurate

brendancol (commented):
## Follow-up review (#3323)

The one Nit from the first pass is addressed: `test_dask_gpu_unmaskable_sentinel_plus_dtype_cast_masked_false` now asserts the computed GPU buffer too (no NaN, values unchanged at `[[10,20],[30,40]]`), so it matches the eager and dask+numpy siblings as a full parity check rather than attrs-only.

No new findings. No Blockers, no Suggestions. The fix and tests pass locally on all four read paths.

PR #3327, opened 2026-06-14
Gate dict gdal_metadata behind the experimental rich-tag opt-in

### LINKED ISSUE #3320: geotiff: attrs['gdal_metadata'] bypasses the experimental rich-tag write gate
**Describe the bug**

`to_geotiff` gates writes carrying `attrs['gdal_metadata_xml']` or `attrs['extra_tags']` behind `allow_experimental_codecs=True`. Both tags sit in the `experimental` tier of `SUPPORTED_FEATURES` (`writer.gdal_metadata_xml`, `writer.extra_tags`). But `attrs['gdal_metadata']` is not gated, even though `_extract_rich_tags` converts it to XML via `_build_gdal_metadata_xml` and writes it to the on-disk TIFF when it's a dict and `gdal_metadata_xml` is None.

So a fresh DataArray (no `_xrspatial_geotiff_contract` marker) can write arbitrary GDAL metadata to the output file without opting in. That's a hole in the experimental rich-tag write gate.

The gate lives in `_validate_write_rich_tag_optin` (`xrspatial/geotiff/_attrs.py`, ~lines 531-539). The write trigger is in `_extract_rich_tags` (~lines 1469-1474).

**Expected behavior**

A fresh DataArray with `attrs['gdal_metadata']` set to a dict should raise `ValueError` on `to_geotiff` unless `allow_experimental_codecs=True` is passed, matching how `gdal_metadata_xml` and `extra_tags` behave. A round-tripped array (carrying the contract marker) should keep writing without the flag.

**Fix direction**

Add a `gdal_metadata` check to `_validate_write_rich_tag_optin`, triggering only when `attrs['gdal_metadata']` is a dict (a non-dict value is ignored by the writer, so it needs no gate). Name `attrs['gdal_metadata']` in the rejection message when it's the trigger. Keep the round-trip exemption intact and don't touch `_extract_rich_tags`.

**Additional context**

Part of the GeoTIFF experimental-tier contract work (epic #2340).

brendancol:
Closes #3320

## Summary

`to_geotiff` gated `attrs['gdal_metadata_xml']` and `attrs['extra_tags']` behind `allow_experimental_codecs=True`, but not `attrs['gdal_metadata']`. Since `_extract_rich_tags` builds GDAL XML from a dict `gdal_metadata` and writes it to disk, a fresh DataArray could write arbitrary GDAL metadata without opting in. This closes that hole.

- Add a `gdal_metadata` check to `_validate_write_rich_tag_optin`, firing only when the value is a dict (a non-dict value is ignored by the writer, so it stays ungated).
- The round-trip exemption for read-back arrays (carrying `_xrspatial_geotiff_contract`) is unchanged, so canonical round-trips stay flag-free.
- `_extract_rich_tags` is untouched.

## Backend coverage

The gate lives in `_validate_write_rich_tag_optin`, shared by the eager/dask CPU writer and the GPU writer, so all four backends enforce it.

## Test plan

- [x] Unit tests on `_validate_write_rich_tag_optin`: dict `gdal_metadata` rejected without the flag, non-dict ignored, accepted with the flag, round-trip marker still exempt.
- [x] End-to-end `to_geotiff`: fresh DataArray with dict `gdal_metadata` raises `ValueError` without the opt-in and writes with it; a read-back array writes flag-free.
- [x] Updated pack/scale-offset test setups that wrote a fresh `gdal_metadata` dict to pass the opt-in.
- [x] Full geotiff suite green (6451 passed).

### DIFF
--- a/docs/source/reference/geotiff_release_contract.md
+++ b/docs/source/reference/geotiff_release_contract.md
@@ -106,7 +106,7 @@ category. The `Key` column matches the runtime key.
 | `writer.bigtiff` | advanced | `bigtiff=True` (or auto-promotion above 4 GiB) writes BigTIFF magic, 8-byte offsets, and 20-byte IFD entries. |
 | `writer.bigtiff_cog` | advanced | BigTIFF plus COG. Tracked separately because the combination has its own external-interop surface. |
 | `writer.gpu` | experimental | GPU write path. |
-| `writer.gdal_metadata_xml` | experimental | `attrs['gdal_metadata_xml']` is escaped before serialisation. |
+| `writer.gdal_metadata_xml` | experimental | `attrs['gdal_metadata_xml']` is escaped before serialisation. A dict `attrs['gdal_metadata']` rides the same gate: writers build the XML from it, so a fresh DataArray carrying it also requires `allow_experimental_codecs=True` (#3320). |
 | `writer.extra_tags` | experimental | Pass-through of TIFF tags outside the structured set via `attrs['extra_tags']`. |
 | `writer.pack` | experimental | `pack=True` on `to_geotiff`; inverse of `reader.unpack`. Re-applies the recorded scale/offset, restores the integer source dtype, and fills NaN back to the nodata sentinel. Tracked at the same tier as `reader.unpack` (#3112, #3114). |
 
--- a/xrspatial/geotiff/_attrs.py
+++ b/xrspatial/geotiff/_attrs.py
@@ -531,6 +531,11 @@ def _validate_write_rich_tag_optin(
     triggered: list[str] = []
     if attrs.get('gdal_metadata_xml') is not None:
         triggered.append("attrs['gdal_metadata_xml']")
+    # ``_extract_rich_tags`` only builds GDAL XML from ``gdal_metadata``
+    # when it is a dict (and ``gdal_metadata_xml`` is absent); a non-dict
+    # value is ignored by the writer, so it does not need gating.
+    if isinstance(attrs.get('gdal_metadata'), dict):
+        triggered.append("attrs['gdal_metadata']")
     if attrs.get('extra_tags') is not None:
         triggered.append("attrs['extra_tags']")
     if gdal_metadata_xml_kwarg is not None:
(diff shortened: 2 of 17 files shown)


brendancol (commented):
## PR Review: Gate dict gdal_metadata behind the experimental rich-tag opt-in

None.

None.

- The gate fires on a dict `attrs['gdal_metadata']` even when `attrs['gdal_metadata_xml']` is also present. In that case `_extract_rich_tags` (`_attrs.py:1474-1478`) prefers the XML and never builds from the dict, so the dict alone wouldn't reach disk. This is harmless: with `gdal_metadata_xml` present the gate already fires for it, so the write is blocked either way, and a message listing both attrs is accurate. No change needed; noting it so the interaction is on record.
- A test for the "both `gdal_metadata_xml` and dict `gdal_metadata` present, no flag" case would pin the message wording, but the existing `gdal_metadata_xml` gate test already covers the reject path. Low value.

### What looks good
- The trigger `isinstance(attrs.get('gdal_metadata'), dict)` mirrors `_extract_rich_tags` exactly, so the gate fires for precisely the inputs that would otherwise write GDAL XML to disk. Non-dict values stay ungated, matching the writer.
- The round-trip exemption is untouched: the `_xrspatial_geotiff_contract` early return runs before the triggered checks, so read-back arrays still write flag-free, and there's a direct test for it.
- The gate is shared by the eager/dask CPU writer and the GPU writer, so all four backends enforce it.
- Test fallout was handled at the source: pack and scale-offset test helpers that build a fresh `gdal_metadata` dict now pass the opt-in, rather than weakening the gate. Full geotiff suite is green.

### Checklist
- [x] Matches the write trigger in `_extract_rich_tags`
- [x] All four backends enforce the gate (shared validator)
- [x] Round-trip exemption preserved and tested
- [x] Edge cases covered (non-dict ignored, accepted with flag)
- [x] No premature materialization or copies (validation-only change)
- [x] Benchmark not needed (input validation)
- [x] Docstrings/contract doc updated (release contract row)

brendancol (commented):
## Follow-up review

Addressed the review nits:

- Added `test_validate_write_rich_tag_optin_names_both_xml_and_gdal_metadata` to pin the rejection message when both `gdal_metadata_xml` and a dict `gdal_metadata` are present without the opt-in (the second nit).
- The first nit (gate also flags the dict when xml is present) is dismissed as a non-defect: `_extract_rich_tags` prefers the XML, the write is already blocked by the `gdal_metadata_xml` trigger, and listing both attrs in the message is accurate. No code change.

No remaining blockers or suggestions. The gate logic, round-trip exemption, and backend coverage are unchanged from the first pass.

PR #3334, opened 2026-06-15
contour: batch dask chunk compute to bound client memory

### LINKED ISSUE #3333: contour: dask backends materialize all chunk results in a single compute
## Summary
The dask backends in `xrspatial/contour.py` materialize every chunk's contour output in a single `dask.compute(*all_results)` call and build the per-chunk task graph with nested Python `for` loops over row/column chunks. For large dask-backed inputs this makes peak memory scale with the total contour complexity rather than with chunk size, and it puts heavy pressure on the dask scheduler.

## Affected code
- `_contours_dask` (`xrspatial/contour.py`, lines ~418-455)
- `_contours_dask_cupy` (`xrspatial/contour.py`, lines ~458-487)

### 1. Single `dask.compute` materializes all chunk results
Both `_contours_dask` and `_contours_dask_cupy` build a list of delayed `_process_chunk_numpy` / `_process_chunk_cupy` tasks, then call:

```python
chunk_results = dask.compute(*all_results)
```

This forces the scheduler to execute and return every chunk's contour geometry to the client process at once. The subsequent merge/deduplication step then builds numpy arrays holding every segment across the whole raster:

```python
seg_rows = np.array(all_segs_r, dtype=np.float64)
seg_cols = np.array(all_segs_c, dtype=np.float64)
```

So peak memory is proportional to the total number of contour segments, not to the per-chunk buffer size. The module's own docstring already notes this limitation, but it means the dask backend is not safe for very large (e.g. 30 TB-class) terrain workloads.

### 2. Python loops over chunk axes
`_contours_dask` / `_contours_dask_cupy` also iterate over `orig_row_chunks` and `orig_col_chunks` in Python to build `all_results`. For a raster with many small chunks this serializes graph construction and is slower than a vectorized approach.

## Impact
- **OOM risk:** dask inputs can run out of client memory during the merge step.
- **Scheduler pressure:** a single `dask.compute` call with one delayed object per chunk can create a very large task graph and a single synchronization point.

## Possible fixes
- Compute chunk results in batches instead of all at once, bounding the amount of intermediate geometry held in memory.
- Consider building the per-chunk task graph with vectorized indexing/offsets rather than nested Python loops.
- Optionally accumulate per-level segment signatures incrementally so the final global merge does not need to hold every segment simultaneously.

## Audit context
Found during the 2026-06-15 performance sweep against `xrspatial/contour.py`.

Melissari1997:
## Summary
Closes #3333.

The dask and dask+cupy `contours()` backends previously submitted every delayed chunk task in a single `dask.compute(*all_results)` call. For large dask-backed rasters this materializes all chunk contour results in the client process at once and creates a large single synchronization point for the scheduler. Peak memory therefore scaled with total contour complexity rather than with chunk size.

This change computes chunk results in batches of `_DASK_COMPUTE_BATCH_SIZE` (default 64). Batching bounds the amount of intermediate chunk geometry held in memory at one time and reduces scheduler pressure, while preserving the same final deduplication and stitching behavior.

## Changes
- Added a private `_DASK_COMPUTE_BATCH_SIZE` constant.
- Consolidated `_contours_dask` and `_contours_dask_cupy` into a shared `_contours_dask_generic` helper so the batching logic is not duplicated.
- Added `TestDaskBatchedCompute::test_dask_compute_called_in_batches` to verify that compute is called in multiple batches when the chunk count exceeds the batch size, and that the batched result still matches the numpy backend.

## Testing
- `python -m pytest xrspatial/tests/test_contour.py -q` passes (61 passed, 30 skipped).
- The new test specifically monkey-patches the batch size to 3 for a 16-chunk raster and confirms 6 `dask.compute` calls.

## Notes
This is a targeted fix for the HIGH-severity materialization finding from the 2026-06-15 performance sweep against `xrspatial/contour.py`. The MEDIUM Python-loop-over-chunks finding is related but left as-is to keep the change surgical; future work could vectorize the offset/graph construction if needed.

### DIFF
--- a/xrspatial/contour.py
+++ b/xrspatial/contour.py
@@ -36,6 +36,12 @@ except ImportError:
     cuda = None
 
 
+# Maximum number of delayed chunk tasks to submit in a single dask.compute
+# call.  Batching bounds peak client memory and reduces scheduler pressure
+# when a large dask-backed raster has many chunks.
+_DASK_COMPUTE_BATCH_SIZE = 64
+
+
 def _available_memory_bytes():
     """Best-effort estimate of available memory in bytes."""
     try:
@@ -421,7 +427,20 @@ def _contours_dask(data, levels):
     Uses ``dask.array.overlap.overlap`` to give each chunk a 1-cell halo
     so that 2x2 quads at chunk boundaries are processed by both neighbors.
     Duplicate segments are removed during the merge/stitch step.
+
+    Chunk results are computed in batches so peak client memory scales
+    with a batch of chunks rather than the total number of chunks.
     """
+    return _contours_dask_generic(data, levels, _process_chunk_numpy)
+
+
+def _contours_dask_cupy(data, levels):
+    """Dask+CuPy backend: overlap chunks, transfer each to CPU."""
+    return _contours_dask_generic(data, levels, _process_chunk_cupy)
+
+
+def _contours_dask_generic(data, levels, process_chunk):
+    """Shared dask backend implementation with batched chunk execution."""
     if da is None:
         raise ImportError("Dask is required for chunked contour extraction")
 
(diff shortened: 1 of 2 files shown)


Melissari1997:
## PR Review: contour: batch dask chunk compute to bound client memory

None found.

- `xrspatial/tests/test_contour.py:640` — The assertion `len(compute_calls) == 6` mixes two concerns: the numpy `contours()` call (which should make zero calls) and the dask call (which should make 6). This is correct now because the test passes explicit levels, skipping the auto-level `dask.compute`. However, if a future change adds a `dask.compute` call to the numpy path (e.g., in validation), this test would break in a confusing way. Consider either (a) resetting `compute_calls = []` between the two `contours()` calls, or (b) using separate counters for numpy vs dask paths.

- `benchmarks/benchmarks/` — No contour benchmark exists. This PR is a targeted performance fix, so a benchmark isn't strictly required. However, adding a simple contour benchmark (even a follow-up) would help prevent regressions in the batched-compute path.

- `xrspatial/tests/test_contour.py:643` — The batch size and chunk count are chosen so the math works out exactly (16 chunks / batch 3 = 6 calls). If anyone edits either constant without updating the other, the test breaks. Using `math.ceil(16 / 3)` for the expected value would make the intent clearer.

- `xrspatial/contour.py:472` — No test exercises the path where `_DASK_COMPUTE_BATCH_SIZE` > number of chunks (single compute call). This is trivially correct but easy to add.

### What looks good

- The batching logic correctly bounds `merged.extend()` to one batch at a time rather than holding the full `chunk_results` list.
- Consolidating `_contours_dask` and `_contours_dask_cupy` into `_contours_dask_generic` eliminates duplicated loop and overlap logic without changing the public API.
- The new test verifies both mechanism (counting compute calls) and correctness (segment equality with numpy), giving confidence that the optimization doesn't change results.
- The existing `TestBackendEquivalence` tests (numpy=dask, numpy=dask_cupy) now implicitly exercise the batched path as well.
- No changes to the public API, dispatch registration, or the marching squares kernel — the change is cleanly scoped.

### Checklist

- [x] Algorithm matches reference/paper (no algorithmic change)
- [x] Benchmark exists or is not needed (existing gap, PR is surgical)
- [x] README feature matrix updated (not applicable — no new public function)

Melissari1997:
## PR Review: contour: batch dask chunk compute to bound client memory

None found.

- `benchmarks/benchmarks/` — No contour benchmark exists. This PR is a targeted
  performance fix, so a benchmark isn't strictly required. Adding a simple
  contour benchmark (even a follow-up) would help prevent regressions in the
  batched-compute path.

None.

### What looks good

- The batching logic correctly bounds `merged.extend()` to one batch at a time
  rather than holding the full `chunk_results` list.
- Consolidating `_contours_dask` and `_contours_dask_cupy` into
  `_contours_dask_generic` eliminates duplicated loop and overlap logic without
  changing the public API.
- `test_dask_compute_called_in_batches` now resets `compute_calls` between
  numpy and dask calls, uses `math.ceil` for the expected count, and includes
  a correctness assertion against the numpy backend.
- New `test_single_batch_when_chunks_fit` covers the edge case where all chunks
  fit in one batch.
- Existing `TestBackendEquivalence` tests implicitly exercise the batched path.
- No changes to the public API, dispatch registration, or marching squares
  kernel — the change is cleanly scoped.

### Checklist

- [x] Algorithm matches reference/paper (no algorithmic change)
- [x] Benchmark exists or is not needed (existing gap, PR is surgical)
- [x] README feature matrix updated (not applicable — no new public function)

brendancol (approved):
@Melissari1997 Thanks for the fix and this should help with scaling

Issue #880, opened 2026-02-24
cost_distance: single-chunk fallback + full-size heap pre-allocation


brendancol:
## Summary

`cost_distance()` has two OOM paths: (1) a single-chunk rechunk fallback when `max_cost` is infinite (the default) or padding exceeds chunk dimensions, and (2) full-size heap pre-allocation inside the per-chunk function which, when applied to the now-single chunk, allocates multiple full-size arrays.

## OOM Code Paths

**Single-chunk fallback:**

| Line | Code | Memory (30 TB raster) |
|------|------|----------------------|
| `cost_distance.py:281` | `source_da = source_da.rechunk({0: height, 1: width})` | 30 TB |
| `cost_distance.py:282` | `friction_da = friction_da.rechuck({0: height, 1: width})` | 30 TB |

**Heap pre-allocation (inside per-chunk function, on the single chunk):**

| Line | Code | Memory |
|------|------|--------|
| `cost_distance.py:143` | `dist = np.full((height, width), np.inf)` | 30 TB |
| `cost_distance.py:145` | `h_keys = np.empty(height * width)` | 30 TB |
| `cost_distance.py:146` | `h_rows = np.empty(height * width)` | 30 TB |
| `cost_distance.py:147` | `h_cols = np.empty(height * width)` | 30 TB |
| `cost_distance.py:155` | `visited = np.zeros((height, width))` | 30 TB |

**Estimated peak RAM: ~150 TB** for a 30 TB raster with default parameters.

## Severity

**CRITICAL** — guaranteed OOM with default parameters on large dask-backed input.

## Suggested Fix

- Implement a tiled Dijkstra approach with halo exchange between chunks
- Avoid rechunking to a single chunk; use iterative relaxation across chunk boundaries
- When `max_cost` is finite, use it to bound the overlap/halo size for `map_overlap`

PR #3345, opened 2026-06-15
cost_distance: route bounded large-radius dask path to iterative Dijkstra (#3342)

### LINKED ISSUE #3342: cost_distance: bounded path rechunks toward single chunk when radius exceeds chunk size (OOM)
## Summary

`cost_distance` with a dask+numpy raster and a finite `max_cost` can rechunk
the input toward a single monolithic chunk, defeating out-of-core processing
and risking OOM. This is the same failure class as #880 (fixed in #888 for the
unbounded path), but it reappears on the bounded `map_overlap` path when the
search radius exceeds the chunk size.

## Cause

In `_cost_distance_dask` the branch that chooses between `map_overlap` and the
iterative tile Dijkstra compares the pad radius against the full array
dimensions:

```python
if not np.isfinite(max_radius) or pad >= height or pad >= width:
    # iterative path
```

`map_overlap`'s `depth` must not exceed the chunk size, not the full array. When
`chunk_size <= pad < full_dim`, this branch still takes `map_overlap` with
`depth > chunk_size`, and dask silently rechunks the inputs to larger blocks to
satisfy the depth. With a large radius it collapses to one chunk.

The GPU path already gets this right by comparing against chunk dimensions:

```python
chunks_y, chunks_x = source_da.chunks
if pad < max(chunks_y) and pad < max(chunks_x):
    use_map_overlap = True
```

## Reproduction

```python
import dask.array as da
import numpy as np
import xarray as xr

src = np.zeros((100, 100)); src[50, 50] = 1.0
fric = np.ones((100, 100))
r = xr.DataArray(da.from_array(src, chunks=(10, 10)), dims=['y', 'x'])
r['y'] = np.arange(100.0); r['x'] = np.arange(100.0)
f = xr.DataArray(da.from_array(fric, chunks=(10, 10)), dims=['y', 'x'])
f['y'] = np.arange(100.0); f['x'] = np.arange(100.0)

res = cost_distance(r, f, max_cost=95.0)  # pad=96, chunks=10
print(res.data.npartitions)   # -> 1  (collapsed from 100 chunks)
print(res.data.chunks)        # -> ((100,), (100,))
```

On a 30TB raster with a finite but generous `max_cost`, this branch builds a
single full-size chunk and runs the kernel against it.

## Expected

When `pad` reaches the chunk size, route to the iterative tile Dijkstra (which
keeps memory bounded), matching both the GPU path's guard and the existing
`test_iterative_no_rechunk_to_single` regression test.

## Severity

HIGH. OOM verdict for the bounded large-radius case: RISKY. Affected backend:
dask+numpy. The dask+cupy path is not affected (it already guards on chunk
size). Found by the performance sweep.

brendancol:
Closes #3342.

## Problem

`cost_distance` on a dask+numpy raster with a finite `max_cost` could rechunk
the input toward a single monolithic chunk and OOM. `_cost_distance_dask`
gated the `map_overlap` vs iterative-tile-Dijkstra branch on the pad radius
versus the full array dimensions:

```python
if not np.isfinite(max_radius) or pad >= height or pad >= width:
```

`map_overlap`'s `depth` must stay within the chunk size. When
`chunk_size <= pad < full_dim`, this still took the `map_overlap` branch with
`depth > chunk_size`, and dask silently rechunked toward larger blocks. With a
large radius it collapsed to one chunk, the same OOM class as #880 (fixed in
#888 for the unbounded path only).

The dask+cupy path already guarded correctly against chunk size.

## Fix

Compare `pad` against the max chunk dimension and route to the iterative tile
Dijkstra when it would overflow a chunk, matching the dask+cupy guard.

## Verification

- New regression `test_bounded_large_radius_no_chunk_collapse`: chunks (10,10),
  `max_cost=95` (pad 96). Asserts the result keeps >1 partition and matches the
  numpy result. Fails on `main` (collapses to 1 chunk, no iterative warning),
  passes with the fix.
- Full suite: 56 passed, including the GPU and dask+cupy paths (CUDA host).

OOM verdict for the bounded large-radius case: was RISKY, now bounded.
Bottleneck: memory-bound. Affected backend: dask+numpy.

### DIFF
--- a/.claude/sweep-performance-state.csv
+++ b/.claude/sweep-performance-state.csv
@@ -7,7 +7,7 @@ classify,2026-04-16T18:00:00Z,SAFE,compute-bound,0,fixed-in-tree,"Fixed-in-tree 
 contour,2026-03-31T18:00:00Z,SAFE,compute-bound,0,,
 convolution,2026-03-31T18:00:00Z,SAFE,compute-bound,0,,
 corridor,2026-03-31T18:00:00Z,SAFE,compute-bound,0,,
-cost_distance,2026-04-16T12:00:00Z,WILL OOM,memory-bound,4,1118,"Re-audit 2026-04-16 after PR 1192 Bellman-Ford fix. 4 HIGH re-surface in iterative tile_cache path (L645 full-dataset materialization, L1015 da.from_delayed wrapping computed tiles). Finite max_cost path remains SAFE. Unbounded path is fundamentally O(dataset) driver memory — covered by #1118."
+cost_distance,2026-06-15,RISKY,memory-bound,1,3342,"Perf sweep 2026-06-15. HIGH: bounded map_overlap branch in _cost_distance_dask gated on full dims (pad>=height/width) not chunk size; pad>chunk collapses to single chunk (#880-class OOM, verified npartitions=1 at chunks=10/pad=96). Fixed: compare pad vs max chunk dim, route to iterative when pad>=chunk (matches GPU path L484). dask+cupy path already correct. Register count 37 (no pressure). nanmin().compute() L478/L1149 intentional scalar. iterative tile_cache full-dataset materialization is documented MemoryError-guarded design (#1118). All 56 tests pass incl GPU."
 curvature,2026-03-31T18:00:00Z,SAFE,compute-bound,0,,
 dasymetric,2026-03-31T18:00:00Z,SAFE,memory-bound,0,1126,Memory guard added to validate_disaggregation. Core disaggregate uses map_blocks.
 diffusion,2026-03-31T18:00:00Z,WILL OOM,memory-bound,2,1116,Scalar diffusivity now passed as float to chunks. DataArray diffusivity passed as dask array via map_overlap.
--- a/benchmarks/benchmarks/cost_distance.py
+++ b/benchmarks/benchmarks/cost_distance.py
@@ -21,5 +21,16 @@ class CostDistance:
             friction.data = np.abs(friction.data) + 0.1
         self.friction = friction
 
+        # Pick a finite max_cost whose pixel radius stays well inside one
+        # chunk (chunks are ny//2 x nx//2), so the dask path exercises the
+        # bounded map_overlap branch rather than the unbounded iterative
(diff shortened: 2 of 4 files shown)


brendancol:
## Review: route bounded large-radius dask path to iterative Dijkstra

Reviewed from a worktree on the head branch. The fix is correct and narrowly scoped. No blockers.

### Blockers
None.

### Suggestions
- The ASV benchmark (`benchmarks/benchmarks/cost_distance.py`) only calls `cost_distance` with the default `max_cost=inf`, so it always runs the unbounded iterative path. The bounded `map_overlap` routing that this PR changes is never benchmarked, so a regression here would not show up in ASV. Worth adding a finite-`max_cost` case.

### Nits
- The guard compares `pad` against `max(chunks)` (`cost_distance.py:1167-1168`), which matches the dask+cupy guard at `cost_distance.py:484`. For ragged chunks, a `pad` that lands between the smallest and largest chunk still makes dask rechunk the smaller boundary chunk. That is a local merge, not the single-chunk collapse this PR fixes, so it is harmless. `min(chunks)` would avoid any rechunk at the cost of sending more cases to the iterative path. Matching the sibling guard is the reasonable call; noting it only so the choice is on the record.

### What looks good
- The OOM mechanism is diagnosed correctly: a `depth` larger than the chunk size makes `map_overlap` rechunk toward a single block.
- The new guard now agrees with the dask+cupy path (`cost_distance.py:484`) and with the module docstring (`cost_distance.py:24`), which already said the iterative path is used when "the implied radius exceeds chunk dimensions". The code now does what the docstring claimed.
- Single-chunk inputs are unaffected: `max(chunks)` equals the full dimension, so the condition reduces to the original `pad >= height or pad >= width`.
- The regression test fails on `main` and checks both `npartitions > 1` and value parity against the numpy result, not just that it runs.

### Checklist
- [x] Algorithm matches intent
- [x] Backends consistent (guard matches dask+cupy)
- [x] NaN handling unchanged
- [x] Single-chunk edge case preserved
- [ ] Benchmark covers the changed path (see suggestion)
- [x] Docstring accurate (now matches behavior)

PR #3346, opened 2026-06-15
Guard cost_distance numpy dask chunk path with _check_memory (#3343)

### LINKED ISSUE #3343: cost_distance: numpy dask map_overlap chunk path skips the _check_memory guard
## Description

The numpy dask path of `cost_distance` skips the working-memory guard that every other backend applies.

`_make_chunk_func` (`xrspatial/cost_distance.py:1122`) builds a closure that calls `_cost_distance_kernel` directly inside `da.map_overlap`. That kernel allocates `dist`, `h_keys`, `h_rows`, `h_cols`, `visited`, and `out`, each sized `height * width`, for the chunk it receives. The map_overlap chunk also includes the overlap depth, so the per-chunk footprint can exceed the nominal chunk size.

The other paths already guard this allocation:

- numpy: `_cost_distance_numpy` calls `_check_memory(height, width)` before the kernel.
- cupy and dask+cupy: `_cost_distance_cupy` calls `_check_gpu_memory(height, width)`.
- iterative tile Dijkstra: `_cost_distance_dask_iterative` estimates total RAM and raises `MemoryError`.

The numpy `map_overlap` path is the only one without a guard. A single oversized chunk allocates several `height * width` arrays and can exhaust a worker's memory. The other paths raise a `MemoryError` that points the user at `max_cost=` or smaller chunks; here the numpy allocator just fails with an opaque error instead.

## Expected behavior

The numpy `map_overlap` chunk function should call `_check_memory(h, w)` before running the Dijkstra kernel, matching `_cost_distance_numpy` and the other backends, so an oversized chunk raises a `MemoryError` that names the remedy.

## Severity

Medium. The chunk size is user-controlled, so this is a defense-in-depth consistency gap rather than a remote vulnerability. It is the only allocation path in the module without a guard.

Found during a security sweep of the cost_distance module.

brendancol:
Closes #3343

## What

- Call `_check_memory(h, w)` inside the `_make_chunk_func` closure before running `_cost_distance_kernel`, so the numpy `da.map_overlap` path raises a clean `MemoryError` (pointing at `max_cost=` or smaller chunks) instead of an opaque allocator failure on an oversized chunk.

This was the only allocation path in the module without a guard. The numpy (`_cost_distance_numpy`), cupy/dask+cupy (`_check_gpu_memory`), and iterative tile Dijkstra paths already had one.

## Backend coverage

- numpy: unchanged (already guarded).
- dask+numpy: now guarded per chunk (the fix).
- cupy / dask+cupy: unchanged (already guarded via `_check_gpu_memory`).

## Test plan

- [x] New `test_dask_map_overlap_chunk_memory_guard_raises` forces the map_overlap path with a finite `max_cost` and a chunk size larger than the overlap pad, mocks `_available_memory_bytes` low, and asserts a `MemoryError` mentioning `max_cost` at compute time.
- [x] Confirmed the test fails on `main` (DID NOT RAISE) and passes with the fix.
- [x] Full `test_cost_distance.py` suite passes (56 passed), covering numpy, dask+numpy, cupy, and dask+cupy.


brendancol (commented):
## PR Review: Guard cost_distance numpy dask chunk path with _check_memory (#3343)

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The fix is one line in the right place. `_make_chunk_func`'s closure receives `source_block`, whose shape already includes the map_overlap depth, so `_check_memory(h, w)` measures the actual per-chunk allocation the kernel is about to make.
- Backend parity is the point of the change: numpy direct (`_cost_distance_numpy`), cupy/dask+cupy (`_check_gpu_memory`), and the iterative tile path all guarded this already. This was the lone gap.
- `_check_memory` is a no-op for normal chunks (it only raises above 50% of available RAM), so well-sized dask graphs are unaffected.
- The new test forces the map_overlap path on purpose (finite `max_cost`, chunk size larger than the overlap pad), mocks memory low, and asserts the `MemoryError` surfaces at compute time. I confirmed it fails on `main` (DID NOT RAISE) and passes with the fix, so it actually pins the bug.

### Checklist
- [x] Algorithm unchanged; only a pre-allocation guard added
- [x] Backends consistent (numpy direct + dask numpy chunk now both guarded)
- [x] NaN handling unaffected
- [x] Edge case covered (oversized chunk raises a clear MemoryError)
- [x] Dask chunk path handled (guard runs inside the task, fires at compute)
- [x] Benchmark not needed (no perf-path change)
- [x] README/docs not applicable (no public API change)
- [x] Pure bug fix, no docstring changes needed

PR #3348, opened 2026-06-15
cost_distance: replace mutable default target_values=[] with None sentinel (#3340)

### LINKED ISSUE #3340: cost_distance(): replace mutable default target_values=[] with None sentinel
`cost_distance()` declares `target_values` with a mutable default argument:

```python
def cost_distance(
    raster, friction, x="x", y="y",
    target_values: list = [],   # mutable default
    max_cost=np.inf, connectivity=8,
):
```

A shared `[]` default is a Python anti-pattern: the same list instance is reused across calls. Here the value is rebound on the first line of the body (`target_values = np.asarray(target_values, ...)`), so there's no active mutation bug, but the declaration is still the unsafe form and it doesn't match how the sibling proximity functions are written.

`proximity()` and `allocation()` (proximity.py) use the `None` sentinel instead:

```python
def proximity(raster, x="x", y="y", target_values: list = None, ...):
    if target_values is None:
        target_values = []
```

Two conventions sit side by side for the same `target_values` parameter. The 2026-06-09 proximity api-consistency sweep flagged this and pointed the fix at the cost-surface modules (cost_distance, surface_distance, balanced_allocation), since proximity already uses the sentinel.

Fix, scoped to `cost_distance` only: change the default to `None` and substitute `[]` in the body, matching proximity/allocation. Non-breaking, no deprecation shim (nothing renamed, default behavior identical). surface_distance and balanced_allocation carry the same pattern and should be fixed in their own sweeps.

Found by /sweep-api-consistency (Cat 4, MEDIUM).

brendancol:
Closes #3340

## What

`cost_distance()` declared `target_values: list = []` (a mutable default). Switched it to the `None` sentinel with internal substitution, matching `proximity()` / `allocation()`.

- Signature: `target_values: list = None`
- Body: `if target_values is None: target_values = []` before the existing `np.asarray` rebind
- Regression tests pinning the `None` default and confirming `None` / `[]` / omitted are equivalent

## Why

A shared `[]` default is the unsafe form and drifted from the sibling proximity functions. The value was already rebound via `np.asarray` on entry, so there was no active mutation bug, but the declaration is still wrong and inconsistent. Found by /sweep-api-consistency (Cat 4, MEDIUM).

Non-breaking: callers passing a list are unaffected and the default behavior (all non-zero finite pixels are sources) is identical. No deprecation shim needed since nothing is renamed.

## Backends

Pure signature/default change in the public wrapper, shared by all four backends. Verified numpy, cupy, dask+numpy, dask+cupy still dispatch and produce identical results.

## Test plan

- [x] `pytest xrspatial/tests/test_cost_distance.py` (57 passed)
- [x] New: default is `None`, and `None` / `[]` / omitted give equal output
- [x] Existing `test_target_values` (explicit values still filter) passes on all parametrized backends


brendancol (commented):
## PR Review: cost_distance — replace mutable default target_values=[] with None sentinel

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/cost_distance.py:1230` — the docstring still reads "If empty, all non-zero finite pixels are sources." With the default now `None`, "if not provided" would be marginally more precise (proximity's docstring phrases it that way). Optional; the `optional` marker on the type line already covers the None default.

### What looks good
- Minimal, surgical change: signature default plus a two-line `None`->`[]` substitution placed before the existing `np.asarray` rebind. No behavior change.
- Matches the sibling convention exactly — proximity.py uses the same `target_values: list = None` plus internal substitution.
- Non-breaking, correctly identified as needing no deprecation shim (nothing renamed).
- Tests pin both the regression (default is `None`) and the behavioral equivalence of `None` / `[]` / omitted, with source-pixel assertions that confirm semantics rather than just "runs without error."
- Backend dispatch untouched; the change lives entirely in the shared public wrapper, so all four backends inherit it.

### Checklist
- [x] No algorithm change (signature/default only)
- [x] All backends inherit the change (shared wrapper)
- [x] NaN handling unchanged
- [x] Edge cases covered (None / [] / omitted equivalence)
- [x] No premature materialization or copies introduced
- [x] Benchmark not needed
- [x] Docstring present; minor wording nit noted

brendancol (commented):
Follow-up: applied the docstring nit (cost_distance.py:1230, "If empty" -> "If not provided") to match the proximity wording now that the default is the None sentinel. Tests still green. No blockers or suggestions outstanding.

PR #3349, opened 2026-06-15
Fix cost_distance leaking dask task name as output .name

### LINKED ISSUE #3344: cost_distance leaks dask task name as output .name on dask backends
**Describe the bug**

`cost_distance` returns a DataArray whose `.name` depends on which backend ran. The numpy and cupy backends return `name=None`, but the dask+numpy and dask+cupy backends leak an internal dask graph name as the user-visible `.name` (e.g. `_trim-052c0535...` from `map_overlap`, or `asarray-09cc...` from the dask+cupy path). The leaked name is a random per-run token, so it isn't even stable between runs.

The public `cost_distance()` ends with:

```python
return xr.DataArray(
    result_data,
    coords=raster.coords,
    dims=raster.dims,
    attrs=raster.attrs,
)
```

No `name=` is passed. On the numpy and cupy paths `result_data` is a plain array, so xarray sets `.name` to `None`. On the dask paths `result_data` carries the dask graph's internal name, and xarray surfaces that as `.name`.

**Expected behavior**

`.name` should be the same across all four backends. The closest peer function, `proximity`, returns `name=None` on every backend. `cost_distance` should match the input raster's name (which is `None` for an unnamed input) regardless of backend.

**Reproduce**

```python

data = np.zeros((6, 6)); data[2, 2] = 1
fric = np.ones((6, 6))
r = xr.DataArray(data, dims=['y', 'x'], attrs={'res': (1.0, 1.0)})
r['y'] = np.arange(6.0); r['x'] = np.arange(6.0)
f = xr.DataArray(fric, dims=['y', 'x']); f['y'] = r['y']; f['x'] = r['x']
r.data = da.from_array(r.data, chunks=(3, 3))
f.data = da.from_array(f.data, chunks=(3, 3))

out = cost_distance(r, f, max_cost=3.0)
print(out.name)   # '_trim-052c0535...' instead of None
```

**Impact**

Pipelines that key on `.name` (for example `.to_dataset()`, which uses `.name` as the variable key) get a different, nondeterministic result on the dask backends. That breaks reproducibility and any name-based logic downstream.

**Fix**

Force the result name to match the input by calling `.rename(raster.name)` on the returned DataArray. Passing `name=raster.name` to the constructor isn't enough: xarray ignores `name=None` and lets the dask token leak through, whereas `.rename(None)` does force the name to `None`.

Found by a metadata-propagation sweep (Cat 5: backend-inconsistent metadata).

brendancol:
Closes #3344

`cost_distance` returned a DataArray whose `.name` depended on the backend: numpy and cupy gave `name=None`, but dask+numpy and dask+cupy leaked the internal dask graph name (`_trim-...`, `asarray-...`) as the user-visible `.name`. That token is random per run, so `.to_dataset()` and any name-keyed logic became nondeterministic on the dask paths.

- Force the result name to the input raster's name with `.rename(raster.name)`. Passing `name=` to the constructor isn't enough because xarray treats `name=None` as "infer from data" and lets the dask token through.
- Add a parametrized regression test over all four backends and both named and unnamed input.

Backends: numpy, cupy, dask+numpy, dask+cupy all verified (CUDA host).

Test plan:
- [x] `test_result_name_matches_input` passes on all four backends (8 cases, no skips on this GPU host)
- [x] Full `test_cost_distance.py` suite passes (63 passed)


brendancol (commented):
## PR Review: Fix cost_distance leaking dask task name as output .name

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The fix is minimal and correct. `.rename(raster.name)` is the right tool: I confirmed that passing `name=None` to the `xr.DataArray` constructor does not override a dask array's internal name, but `.rename(None)` does force it, so the constructor-only approach would not have fixed the unnamed-input case.
- Only `.name` changes; data, coords, dims, and attrs are untouched.
- The regression test runs all four backends and both named and unnamed inputs (8 cases). On this CUDA host none of them skip, so the cupy and dask+cupy paths are genuinely exercised.
- The `supports_dataset` path is unaffected: a Dataset input calls the function per variable with `raster.name` already equal to the variable name, so the rename keeps the variable key. Verified by running a Dataset through the dask backend.
- The inline comment explains the non-obvious reason `.rename` is needed instead of a constructor `name=`, which will save the next reader a debugging session.

### Checklist
- [x] Behavior matches intent (name consistent across backends)
- [x] All four backends produce consistent .name
- [x] No data/coords/dims/attrs regression
- [x] Edge cases covered (named + unnamed input, all backends)
- [x] No premature materialization or extra copies (rename is metadata-only)
- [x] Benchmark not needed (metadata-only fix)
- [x] README / docs not applicable (no API change)

PR #3353, opened 2026-06-15
contour: test coverage for all-equal auto-levels and empty explicit levels

### LINKED ISSUE #3352: contour: test gaps for all-equal auto-levels and empty explicit levels
## Reason or Problem
Two edge cases in `contours()` are not tested:

1. **All-equal raster with auto-levels.**  `test_no_contours_flat` passes explicit `levels=[]`, so it never exercises the auto-level branch (`levels=None`) on a flat raster where `vmin == vmax`.  That branch calls `np.linspace(vmin, vmax, n_levels + 2)[1:-1]` which produces identical levels on equal min/max — the function returns an empty result, but no test asserts this.

2. **Empty `levels=[]`.**  Passing an empty list as `levels` returns `[]` silently.  No test documents that this is the intended behavior or that it does not raise.

## Proposal

**Design:** Add two test functions to `xrspatial/tests/test_contour.py`:

```python
def test_contours_flat_auto_levels():
    raster = xr.DataArray(
        np.ones((10, 10), dtype=np.float64), dims=["y", "x"]
    )
    result = contours(raster, levels=None, n_levels=10)
    assert result == []

def test_contours_empty_explicit_levels():
    raster = xr.DataArray(
        np.random.rand(10, 10), dims=["y", "x"]
    )
    result = contours(raster, levels=[], return_type="numpy")
    assert result == []

    result_gdf = contours(raster, levels=[], return_type="geopandas")
    assert len(result_gdf) == 0
```

**Value:** Covers two branches with zero test coverage, preventing regressions on the auto-level range computation and the empty-levels path that the geopandas empty-GeoDataFrame CRS fix (#2700) depends on.

## Stakeholders and Impacts
Test-only change.  No source modification needed.

## Drawbacks
Negligible — two small deterministic tests.

## Alternatives
- Rely on the accuracy and performance sweeps catching regressions here.
- File separate issues per gap (combined here for minimal overhead).

Melissari1997:
Closes #3352

## Summary
- Add `test_contours_flat_auto_levels` — asserts an all-equal raster with `levels=None` returns `[]` (covers the `vmin == vmax` branch in auto-level generation)
- Add `test_contours_empty_explicit_levels_numpy` — asserts `levels=[]` with numpy return type returns `[]`
- Add `test_contours_empty_explicit_levels_geopandas` — asserts `levels=[]` with geopandas return type returns an empty GeoDataFrame

## Backend coverage
Numpy only (both edge cases hit the early-return path before backend dispatch).

### DIFF
--- a/xrspatial/tests/test_contour.py
+++ b/xrspatial/tests/test_contour.py
@@ -1504,7 +1504,7 @@ class TestNaNBackendParity:
 # Coverage gaps: all-equal auto-levels and empty explicit levels (#3352)
 # ---------------------------------------------------------------------------
 
-class TestIssue3352Coverage:
+class TestFlatAutoLevelsAndEmptyLevels:
 
     def test_contours_flat_auto_levels(self):
         """An all-equal raster with auto-levels (levels=None) returns empty.
@@ -1512,11 +1512,12 @@ class TestIssue3352Coverage:
         When vmin == vmax, np.linspace produces identical levels and no
         contours are generated.  Regression guard for the auto-level branch.
         """
-        raster = xr.DataArray(
-            np.ones((10, 10), dtype=np.float64), dims=["y", "x"]
+        raster = create_test_raster(
+            np.ones((10, 10), dtype=np.float64), backend='numpy'
         )
         result = contours(raster, levels=None, n_levels=10)
-        assert result == []
+        assert isinstance(result, list)
+        assert len(result) == 0
 
     def test_contours_empty_explicit_levels_numpy(self):
         """Passing an empty list as levels with numpy return_type returns [].
@@ -1524,19 +1525,22 @@ class TestIssue3352Coverage:
         No crossing can exist with zero levels, so the function returns an
         empty list without raising.
         """
-        raster = xr.DataArray(
-            np.random.rand(10, 10), dims=["y", "x"]
+        raster = create_test_raster(
+            np.random.rand(10, 10), backend='numpy'
         )
-        result = contours(raster, levels=[], return_type="numpy")
+        result = contours(raster, levels=[], return_type='numpy')
+        assert isinstance(result, list)
         assert result == []
 
     def test_contours_empty_explicit_levels_geopandas(self):
         """Passing an empty list as levels with geopandas returns empty gdf."""
-        pytest.importorskip("geopandas")
+        pytest.importorskip('geopandas')
         import geopandas as gpd
-        raster = xr.DataArray(
-            np.random.rand(10, 10), dims=["y", "x"]
+        raster = create_test_raster(
+            np.random.rand(10, 10), backend='numpy'
         )
-        result_gdf = contours(raster, levels=[], return_type="geopandas")
+        result_gdf = contours(raster, levels=[], return_type='geopandas')
         assert isinstance(result_gdf, gpd.GeoDataFrame)
         assert len(result_gdf) == 0
+        assert 'level' in result_gdf.columns
+        assert 'geometry' in result_gdf.columns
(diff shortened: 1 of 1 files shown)


Melissari1997 (commented):
## PR Review: Test coverage for all-equal auto-levels and empty explicit levels

**Changed file:** `xrspatial/tests/test_contour.py`

None. All tests are logically correct and target real coverage gaps not exercised by existing tests.

**S1 — Use `create_test_raster` instead of raw `xr.DataArray`**

All three new tests construct the raster with raw `xr.DataArray(...)` rather than the project's `create_test_raster` helper. Every other test in this file that exercises contour behavior uses `create_test_raster`, which sets up dims, coords, units, and CRS attrs consistently. Using the raw constructor means:
- No coords are set
- The CRS attribute is absent

At minimum the geopandas test should test with a CRS-bearing raster to exercise the CRS-propagation path (consistent with other empty-result geopandas tests at `test_geopandas_empty_result_keeps_crs` and `test_geopandas_all_nan_keeps_crs`).

`xrspatial/tests/test_contour.py:1515, 1527, 1537`

**S2 — Class naming doesn't follow project convention**

The class is named `TestIssue3352Coverage`. The existing convention is descriptive class names like `TestEdgeCases`, `TestNaNBackendParity`, `TestGeoDataFrame`, etc. A descriptive name would be more maintainable.

`xrspatial/tests/test_contour.py:1507`

**S3 — `test_contours_empty_explicit_levels_geopandas` misses column assertions**

The existing test `test_geopandas_empty_result_keeps_crs` asserts `'level' in gdf.columns` and `'geometry' in gdf.columns` on an empty GeoDataFrame. The new geopandas test checks `isinstance` and `len == 0` but does not verify the expected schema.

`xrspatial/tests/test_contour.py:1540-1542`

**N1 — Double-quote style for dims strings**

The rest of the file consistently uses single quotes for string literals (`dims=['y', 'x']`). All three new tests use double quotes (`dims=["y", "x"]`).

`xrspatial/tests/test_contour.py:1515-1516, 1528, 1538`

**N2 — `assert result == []` vs explicit type + length checks**

`assert result == []` works for the numpy return case. However, other tests in this file (e.g. `test_no_contours_flat`) use `assert isinstance(result, list)` separately. An object of the wrong type (e.g. a tuple `()`) would slip past `== []`.

`xrspatial/tests/test_contour.py:1519, 1531`

### What looks good

1. Tests target genuine coverage gaps — `test_no_contours_flat` tests flat rasters with *explicit* levels, so the `levels=None` auto-level path was not exercised with `vmin == vmax`.
2. Both return types (numpy and geopandas) are covered for the empty-levels case.
3. `pytest.importorskip("geopandas")` is used for the geopandas test, consistent with the rest of the file.
4. Docstrings are clear and explain why these cases matter.
5. Minimal and focused — each test does exactly one thing.

### Checklist

- [x] Covers stated gap: all-equal auto-levels, empty explicit levels
- [x] Correct assertions
- [ ] Follows file conventions (see S1, S2, N1)
- [x] `pytest.importorskip` used appropriately

Melissari1997 (commented):
## Follow-up review (after fix commits)

### Disposition of previous findings

| Finding | Status | Notes |
|---|---|---|
| S1 — use `create_test_raster` | Fixed | All three tests now use `create_test_raster` |
| S2 — class naming convention | Fixed | Renamed to `TestFlatAutoLevelsAndEmptyLevels` |
| S3 — geopandas column assertions | Fixed | `level` and `geometry` columns now asserted |
| N1 — double-quote style | Fixed | Switched to single quotes throughout |
| N2 — bare `== []` assert style | Fixed | Both numpy tests now use `isinstance` + `len` |

No new issues found. All previous findings are resolved.

brendancol (approved):
@Melissari1997 LGTM...thanks again

PR #3357, opened 2026-06-15
Scope 'performance' labeler to benchmarked modules

### LINKED ISSUE #3356: labeler.yml applies 'performance' to nearly every PR, triggering benchmarks each time
**Describe the bug**

`.github/labeler.yml` adds the `performance` label to any PR that changes a non-test file under `xrspatial/`:

```yaml
performance:
  - changed-files:
    - any-glob-to-any-file:
      - 'xrspatial/*.py'
      - 'xrspatial/**/*.py'
      - '!xrspatial/tests/**'
```

That glob matches essentially every source file, so almost every code PR gets the label.

The label is not just noise. It is a trigger. `benchmarks.yml` runs the ASV suite whenever a PR carries it:

```yaml
if: >
  (github.event_name == 'push') ||
  (github.event_name == 'pull_request' && github.event.label.name == 'performance')
```

So the broad glob runs the benchmark job on nearly every PR, including changes that touch no benchmarked code.

**Expected behavior**

`performance` should land only on PRs that change code the benchmarks actually measure, so the benchmark job runs when it can move the numbers.

**Proposed fix**

Scope the rule to the modules that have PR-time benchmarks. The asv `continuous` filter in `benchmarks.yml` covers `Slope`, `Proximity`, `Zonal`, `CostDistance`, `Focal`, `Rescale`/`Standardize`, `Diffusion`, and `Dasymetric`, which live in:

`slope.py`, `proximity.py`, `zonal.py`, `cost_distance.py`, `focal.py`, `normalize.py`, `diffusion.py`, `dasymetric.py`

Removing the rule outright is the other option the title suggests, but that also drops the on-demand benchmark trigger, so scoping is the better choice.

**Additional context**

The over-broad glob has been in place since the labeler was added in #917.

brendancol:
Closes #3356

`.github/labeler.yml` matched `xrspatial/*.py` and `xrspatial/**/*.py`, so nearly every code PR picked up the `performance` label. That label is a trigger: `benchmarks.yml` runs the ASV suite whenever a PR carries it, so the benchmark job ran on almost every PR.

- Scope the `performance` rule to the eight modules the PR-time benchmark filter actually measures: `slope.py`, `proximity.py`, `zonal.py`, `cost_distance.py`, `focal.py`, `normalize.py` (Rescale/Standardize), `diffusion.py`, `dasymetric.py`.
- Add a comment tying the label to the asv `continuous` filter in `benchmarks.yml` so the list stays in sync as benchmarks are added.

Keeping the label (rather than removing it) preserves the on-demand benchmark trigger for the PRs that can move the numbers.

Not a code change, so no backend work (numpy / cupy / dask) applies.

Test plan:
- [x] `yaml.safe_load` parses the file and the structure matches the `actions/labeler@v5` schema
- [x] all eight globbed paths exist in the tree

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -3,8 +3,12 @@
 # source file, so the benchmarks ran on nearly every PR. Limit the label to
 # the modules the PR-time benchmarks actually measure (the asv `continuous`
 # filter in benchmarks.yml: Slope, Proximity, Zonal, CostDistance, Focal,
-# Rescale/Standardize, Diffusion, Dasymetric). Add a module here when it
-# gains a benchmark in that filter.
+# Rescale/Standardize, Diffusion, Dasymetric), plus the hot compute kernels
+# those benchmarks run through (convolution.py drives Focal; geodesic.py
+# drives Slope). utils.py is deliberately left out: it is shared infra
+# touched by almost every PR, so including it would re-broaden the label
+# back toward the original problem. Add a module here when it gains a
+# benchmark in that filter or becomes a kernel the benchmarks exercise.
 performance:
   - changed-files:
     - any-glob-to-any-file:
@@ -16,3 +20,5 @@ performance:
       - 'xrspatial/normalize.py'
       - 'xrspatial/diffusion.py'
       - 'xrspatial/dasymetric.py'
+      - 'xrspatial/convolution.py'
+      - 'xrspatial/geodesic.py'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,6 +1,18 @@
+# Applying the `performance` label triggers the ASV benchmark job in
+# benchmarks.yml. The old `xrspatial/**/*.py` glob matched almost every
+# source file, so the benchmarks ran on nearly every PR. Limit the label to
+# the modules the PR-time benchmarks actually measure (the asv `continuous`
+# filter in benchmarks.yml: Slope, Proximity, Zonal, CostDistance, Focal,
+# Rescale/Standardize, Diffusion, Dasymetric). Add a module here when it
+# gains a benchmark in that filter.
 performance:
   - changed-files:
     - any-glob-to-any-file:
-      - 'xrspatial/*.py'
-      - 'xrspatial/**/*.py'
-      - '!xrspatial/tests/**'
+      - 'xrspatial/slope.py'
+      - 'xrspatial/proximity.py'
+      - 'xrspatial/zonal.py'
+      - 'xrspatial/cost_distance.py'
+      - 'xrspatial/focal.py'
+      - 'xrspatial/normalize.py'
+      - 'xrspatial/diffusion.py'
+      - 'xrspatial/dasymetric.py'


brendancol (commented):
## PR Review: Scope 'performance' labeler to benchmarked modules

Config-only change, so the numerical/backend/NaN/dask/cupy sections of the review don't apply. I checked the parts that do: the glob list, the YAML structure, and whether the scoping lines up with what the benchmark job measures.

### Blockers
None.

### Suggestions
- The eight modules match the asv `continuous` filter one-to-one, but the benchmarks also run through shared helpers that aren't in the glob. `Focal` pulls in `convolution.py`, `Slope` pulls in `geodesic.py`, and most of them touch `utils.py`. A PR that changes only one of those shifts the benchmark numbers but won't get auto-labeled, so a regression could land without the PR-time benchmark running. Two ways to handle it: add the main shared helpers (`convolution.py`, `geodesic.py`) to the glob, or leave the comment as-is and rely on maintainers applying the label by hand for helper-only changes. I'd lean toward the comment route since pulling in `utils.py` would re-broaden the rule back toward the original problem, but it's worth a deliberate call rather than leaving the gap silent.

### Nits
None.

### What looks good
- Keeping the label instead of deleting it preserves the on-demand benchmark trigger, which is the better of the two options the issue raised.
- The comment ties the glob to the asv filter in `benchmarks.yml`, so the next person editing benchmarks knows to keep the two in sync.
- YAML parses and the structure matches the `actions/labeler@v5` schema; all eight paths exist in the tree.

### Checklist
- [x] YAML valid and matches actions/labeler@v5 schema
- [x] Globbed paths all exist
- [x] Scoping matches the benchmark filter
- [ ] Shared-helper coverage gap (see Suggestions) — intentional tradeoff, flagged for a deliberate decision
- n/a Algorithm / backends / NaN / dask / docstrings / README — no code or API change

brendancol (commented):
## Follow-up review

Re-reviewed after the follow-up commit (5c3dd7e).

The shared-helper coverage gap from the first pass is addressed: `convolution.py` (drives the Focal benchmarks) and `geodesic.py` (drives the Slope benchmarks) are now in the glob, so a regression in those kernels triggers the benchmark job. `utils.py` was deliberately left out and the reasoning is now in the file comment, which is the right call — it's touched by almost every PR and would undo the scoping.

No new findings. YAML still parses, the structure matches the `actions/labeler@v5` schema, and all ten paths exist in the tree. No remaining blockers, suggestions, or nits.

brendancol:
Root-cause note while verifying on this PR: the labeler tagged this PR `performance` even though it only touches `.github/labeler.yml`. That exposed a second cause beyond the broad glob.

The labeler workflow has no checkout step, so `actions/labeler@v5` fetches the config from the base branch via the API and matches it against the PR's changed files. In the old config the test exclusion `!xrspatial/tests/**` sits inside `any-glob-to-any-file`, which ORs its patterns. A negated glob matches every file that is *not* a test file, so any non-test path in the repo — docs, CI, this very file — satisfies it and earns the label. That, on top of `xrspatial/**/*.py`, is why `performance` landed on nearly everything.

The new config uses only explicit positive globs and no negation, so a PR is labeled only when it changes one of the listed modules. I removed the stray `performance` label from this PR.

PR #3360, opened 2026-06-16
Add free-threaded Python 3.14 (3.14t) to CI matrix

### LINKED ISSUE #3359: Add free-threaded Python 3.14 (3.14t) to the CI test matrix
## Reason or Problem
Python 3.14 ships a free-threaded (no-GIL) build, and as of 3.14 it's a supported build rather than experimental. Our matrix tests 3.12, 3.13, and 3.14, but only the default GIL-enabled interpreter. That leaves us with no signal on whether xarray-spatial works without the GIL, which is exactly where thread-safety bugs in our own code or in the numba/numpy layers would show up.

## Proposal
Add the free-threaded interpreter (`3.14t`) to the `python` axis in `.github/workflows/test.yml`, on both the PR fast lane and the full push/nightly lane. `actions/setup-python@v5` resolves `3.14t` to the free-threaded build.

numba and parts of the stack are still finishing their free-threaded support, so this job will likely break before it stays green. Mark `3.14t` as `continue-on-error` so a failed install or test reports without turning the PR check red or blocking a merge. Once the ecosystem settles we can promote it to a required job.

**Value:** we hear about no-GIL breakage before users do, and we get a single place to track the work toward full free-threaded support.

See the [free-threaded Python guide](https://py-free-threading.github.io/) for the ecosystem status and CI recommendations (allowed-failure job, generous `timeout-minutes` since deadlocks are more likely without the GIL).

### LINKED ISSUE #3361: Free-threaded (3.14t) data race: zlib decompress error in chunked COG sidecar read
## Describe the bug
Under free-threaded Python 3.14 (`3.14t`), `test_sidecar.py::test_fsspec_chunked_open_resolves_sidecar_overview` fails with a zlib decompression error when reading a COG sidecar overview through the chunked (dask) path:

```
zlib.error: Error -3 while decompressing data: incorrect header check
```

On Windows the same test also surfaces `ValueError: I/O operation on closed file.`

The failure showed up on the new `3.14t` CI lane (PR #3360). It reproduces on ubuntu-latest and windows-latest but not macos-latest, and never on the GIL-enabled builds (3.12, 3.13, 3.14 all pass). 9689+ other tests pass.

## Likely cause
The test opens a COG with `chunks=16` and then materializes `.values`, so tiles decompress in parallel on dask's threaded scheduler. The error pattern (corrupted zlib stream plus a closed-file error) points to a shared file handle and/or `zlib` decompressor used by concurrent tile reads without locking. `zlib` decompress objects are not thread-safe; the GIL was serializing these reads, so the race only appears once it is gone. macOS passing is consistent with a timing-dependent race rather than a platform-specific bug.

## Where to look
- `xrspatial/geotiff/tests/integration/test_sidecar.py::test_fsspec_chunked_open_resolves_sidecar_overview`
- the chunked read path: `xrspatial/geotiff/_backends/dask.py` and the tile decode helpers it calls (`_decode.py`, `_compression.py`, `_sources.py`)
- anything that holds a single open file object or a single decompressor across tiles

## Notes
This is pre-existing behavior, not caused by PR #3360, which only added `3.14t` to the matrix. The `3.14t` job is allowed-failure, so this does not block merges; it is the signal the lane was added to provide. This issue tracks the fix toward making the free-threaded lane green.

brendancol:
Closes #3359

Adds the free-threaded (no-GIL) build of Python 3.14 to the CI matrix.

- Add `3.14t` to the `python` axis on both the PR fast lane and the full push/nightly lane. `actions/setup-python@v5` resolves it to the free-threaded build.
- Mark the free-threaded job `continue-on-error`, keyed on `endsWith(matrix.python, 't')`, so a broken install or a thread-safety failure reports without blocking merges. numba and parts of the stack are still finishing their no-GIL support, so this job will likely go red before it stays green. Promote it to a required job once it stabilizes.

This is a CI-config-only change, so no library code, tests, docs, or README rows are touched.

Test plan:
- [ ] YAML parses (`yaml.safe_load`)
- [ ] CI shows `3.14t` jobs on this PR for all three OSes, reporting as allowed-failure

Also fixes the free-threaded race the new lane surfaced: Closes #3361

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,9 @@ jobs:
     # thread-safety failure reports signal without turning the PR check red or
     # blocking a merge. Promote it to a required job once it stays green.
     continue-on-error: ${{ endsWith(matrix.python, 't') }}
+    # Cap the free-threaded job so a no-GIL deadlock can't sit on a runner for
+    # the 360-minute default; other versions keep that default.
+    timeout-minutes: ${{ endsWith(matrix.python, 't') && 30 || 360 }}
     strategy:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,10 +23,15 @@ on:
 jobs:
   run:
     runs-on: ${{ matrix.os }}
+    # The free-threaded build (3.14t) is allowed to fail while numba and the
+    # rest of the stack finish their no-GIL support, so a broken install or a
+    # thread-safety failure reports signal without turning the PR check red or
+    # blocking a merge. Promote it to a required job once it stays green.
+    continue-on-error: ${{ endsWith(matrix.python, 't') }}
     strategy:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python: ${{ github.event_name == 'pull_request' && fromJson('["3.14"]') || fromJson('["3.12", "3.13", "3.14"]') }}
+        python: ${{ github.event_name == 'pull_request' && fromJson('["3.14", "3.14t"]') || fromJson('["3.12", "3.13", "3.14", "3.14t"]') }}
     env:
       OS: ${{ matrix.os }}
       PYTHON: ${{ matrix.python }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,14 +38,6 @@ jobs:
     env:
       OS: ${{ matrix.os }}
       PYTHON: ${{ matrix.python }}
-      # Force the GIL off on the free-threaded (3.14t) job so the tests
-      # exercise the real no-GIL path instead of letting CPython silently
-      # re-enable the GIL for a C extension that hasn't declared
-      # free-threaded support (which would hide thread-safety bugs behind
-      # a green run). Empty on the GIL builds, where it is ignored --
-      # PYTHON_GIL=0 there is a fatal error ("Disabling the GIL is not
-      # supported by this build").
-      PYTHON_GIL: ${{ endsWith(matrix.python, 't') && '0' || '' }}
     steps:
     - uses: actions/checkout@v4
     - name: Setup Python
@@ -62,7 +54,18 @@ jobs:
       # (today: the six compression fixtures). push-to-main and the
       # nightly schedule run the full set with no filter.
       if: github.event_name == 'pull_request'
+      # Force the GIL off only for the free-threaded (3.14t) test run so the
+      # suite exercises the real no-GIL path instead of letting CPython
+      # silently re-enable the GIL for a C extension that hasn't declared
+      # free-threaded support. Scoped to the pytest step (not the job) so it
+      # never reaches setup-python's bootstrap interpreter, where
+      # PYTHON_GIL=0 is a fatal error on a GIL build. Empty (ignored) on the
+      # GIL versions.
+      env:
+        PYTHON_GIL: ${{ endsWith(matrix.python, 't') && '0' || '' }}
       run: pytest -m "not slow"
     - name: Run pytest (full)
       if: github.event_name != 'pull_request'
+      env:
+        PYTHON_GIL: ${{ endsWith(matrix.python, 't') && '0' || '' }}
       run: pytest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,6 +38,14 @@ jobs:
     env:
       OS: ${{ matrix.os }}
       PYTHON: ${{ matrix.python }}
+      # Force the GIL off on the free-threaded (3.14t) job so the tests
+      # exercise the real no-GIL path instead of letting CPython silently
+      # re-enable the GIL for a C extension that hasn't declared
+      # free-threaded support (which would hide thread-safety bugs behind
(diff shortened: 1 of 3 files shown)


brendancol (commented):
## PR Review: Add free-threaded Python 3.14 (3.14t) to CI matrix

CI-config-only change, one file (`.github/workflows/test.yml`). The Actions expressions are valid and the rollout approach (allowed-failure experimental job) is the right call. No blockers.

- None.

- [ ] No `timeout-minutes` on the `run` job (`.github/workflows/test.yml:24`). The free-threading guide calls this out because deadlocks are more likely without the GIL, and this repo already sets timeouts in `docs.yml` and `test-cog-validator.yml`. With `continue-on-error`, a hung `3.14t` job has nothing to stop it before the 360-minute default, so it burns runner minutes without ever gating a merge. Cap the experimental entry, e.g. `timeout-minutes: ${{ endsWith(matrix.python, 't') && 30 || 360 }}`.

- [ ] The PR fast lane goes from 3 jobs to 6 (2 Python versions x 3 OSes). Intentional and cheap, noting it for visibility.
- [ ] Possible follow-up: set `PYTHON_GIL=0` for the `3.14t` job so it actually exercises the no-GIL path. By default CPython re-enables the GIL for C extensions that declare they need it, which can hide incompatibilities behind a green run. The tradeoff is that forcing it off can abort at import for extensions that have not declared support yet, so it is reasonable to leave for a later iteration once the job installs cleanly.

### What looks good
- `continue-on-error: ${{ endsWith(matrix.python, 't') }}` keys on the `t` suffix, so it covers any future free-threaded version without another edit, and reads clearly.
- Both lanes (PR and push/nightly) updated consistently.
- Matrix values stay quoted inside the JSON array, so `3.14`/`3.14t` are not coerced to YAML floats.
- `actions/setup-python@v5` is already used here and resolves `3.14t` to the free-threaded build.

### Checklist
- [x] Actions expressions valid (`endsWith`, job-level `continue-on-error`)
- [x] Both matrix lanes updated
- [x] Allowed-failure semantics correct (won't block merges)
- [ ] Timeout guard on the experimental job (suggested)
- [x] No library code, tests, docs, or README affected

brendancol (commented):
## Follow-up review (after timeout commit)

Re-reviewed after `93571b9`. Disposition of the earlier findings:

- **Suggestion (timeout guard): fixed.** Added `timeout-minutes: ${{ endsWith(matrix.python, 't') && 30 || 360 }}`, keyed the same way as `continue-on-error`. The 3.14t job is capped at 30 minutes; 3.12/3.13/3.14 keep the 360-minute default, so their behavior is unchanged.
- **Nit (3→6 PR jobs): dismissed.** Intended; exercising 3.14t on PRs is the point of the change.
- **Nit (`PYTHON_GIL=0`): deferred.** Whether to force the no-GIL path is a design choice. The first iteration is deliberately conservative so we can confirm the job installs and runs before tightening it. Worth revisiting once 3.14t is green.

No new findings. YAML parses; the timeout expression resolves to a number per matrix entry.

brendancol:
CI note: the `3.14t` jobs on ubuntu/windows are red, but the `pytest` workflow run still concluded **success** — `continue-on-error` kept the allowed-failure free-threaded job from gating the merge. All GIL builds (3.12/3.13/3.14) pass on every OS, and macOS 3.14t passes too.

The one failure is a pre-existing thread-safety bug surfaced by no-GIL, not introduced here: a zlib decompression race in the chunked COG sidecar read (`test_fsspec_chunked_open_resolves_sidecar_overview`). Tracked in #3361. This PR only adds the matrix entry, so it's good to merge independently of that fix.

brendancol:
## Fixed the free-threaded race (#3361) on this branch

**Root cause:** `_CloudSource.read_range`/`read_all` (`xrspatial/geotiff/_sources.py`) read via `fs.open(path, 'rb')` + `seek` + `read`. For fsspec's `MemoryFileSystem`, `open('rb')` returns the **same shared process-global handle** every call (verified: `fs.open(p) is fs.open(p)` -> `True`). Concurrent windowed reads — dask chunk tasks plus `read_ranges`' own worker pool — raced on that single cursor: one thread's `seek` landed under another's `read`, corrupting tile bytes (`zlib.error: incorrect header check`) or hitting a just-closed handle (`I/O operation on closed file`). The GIL serialized the reads enough to hide it; the new `3.14t` lane exposed it.

**Fix:** route both reads through the stateless `cat_file(path, start=, end=)` ranged API — a fresh byte slice per call, no shared cursor. `cat_file` matches `seek`+`read` semantics exactly, including the EOF clamp the COG header prefetch relies on (verified across `(0,5)`, `(0,100)`, `(3,100)`, `(2,0)`, `(5,10)`).

**Regression test** (`tests/read/test_cloud_source_concurrency_3361.py`): a plain thread stress test shows 0 mismatches under the GIL — the GIL fully masks this — so the test instead forces the shared-cursor interleaving with a `threading.Barrier` in `seek`. It fails deterministically on the old implementation (even under the GIL) and passes on the fix. Confirmed by reverting the fix: the test fails; restored: it passes.

**Verification:**
- `test_fsspec_chunked_open_resolves_sidecar_overview` (the original 3.14t failure): passes.
- geotiff read + integration + unit suites: 1624 + 1476 passed, no regressions.
- flake8/isort clean.

The PR now `Closes #3361` as well as #3359.

PR #3364, opened 2026-06-16
Make test_chunks_is_lazy counter thread-safe for the free-threaded CI lane

### LINKED ISSUE #3363: Free-threaded CI: test_chunks_is_lazy counter race miscounts decodes (15 != 16)
## Describe the bug
The free-threaded (`3.14t`) full-suite CI job on `main` fails on one test:

```
FAILED xrspatial/geotiff/tests/vrt/test_window.py::test_chunks_is_lazy_does_not_call_internal_reader
  AssertionError: expected 16 per-chunk decodes after compute, got 15
```

(run 27626148337, job 81688172532). Every other test passes (9696 passed).

## Root cause
The test instruments the internal VRT reader with a counting wrapper:

```python
counter = {'calls': 0}
def counting_read(*args, **kwargs):
    counter['calls'] += 1
    return real_read(*args, **kwargs)
```

`result.compute()` on a 4x4 = 16-chunk dask array runs the 16 decode tasks through dask's threaded scheduler. Under `PYTHON_GIL=0` (set on the `3.14t` pytest step) those tasks run truly in parallel, and `counter['calls'] += 1` is a non-atomic read-modify-write: two threads read the same value and write back the same increment, losing one. So 16 real decodes are counted as 15.

This is a test-instrumentation thread-safety bug, not a product bug. The chunked decode correctly runs 16 times; only the counter miscounts. The test runs in both the fast and full lanes (it is not slow-marked); the race is probabilistic, so it slipped through #3360's `3.14t` fast-lane run and first tripped on the `main` full run.

## Fix
Make the counter increment thread-safe (wrap it in a `threading.Lock`) so the count is exact under both the GIL and the free-threaded interpreter. Only this one test is affected; the other counter-based tests in `test_metadata.py` / `test_overview.py` / `test_visibility.py` assert a single non-concurrent call (`== 1`) and cannot lose an update.

## Notes
Surfaced by the `3.14t` lane added in #3360. The job is allowed-failure (`continue-on-error`), so the overall `main` run still concludes success, but the red job is noise until this is fixed.

brendancol:
Closes #3363

The `3.14t` full-suite job on `main` fails on one test:

```
test_chunks_is_lazy_does_not_call_internal_reader
  AssertionError: expected 16 per-chunk decodes after compute, got 15
```

`result.compute()` runs the 16 per-chunk reads through dask's threaded scheduler. Under `PYTHON_GIL=0` (the `3.14t` pytest step) those run truly in parallel, and the counting wrapper's `counter['calls'] += 1` is a non-atomic read-modify-write that loses an update, so 16 real decodes count as 15.

- Guard the increment with a `threading.Lock` so the count is exact under both the GIL and the free-threaded interpreter.
- Test-instrumentation fix only: the chunked decode already runs 16 times and returns correct data. No product code changes.
- Only this test is affected; the other counter-based tests assert a single non-concurrent call (`== 1`) and can't lose an update.

Surfaced by the `3.14t` lane (#3360). That job is allowed-failure, so `main` still concludes success, but this clears the red.

Test plan:
- [ ] `pytest xrspatial/geotiff/tests/vrt/test_window.py` passes under the GIL (no regression).
- [ ] `3.14t` full lane goes green.

### DIFF
--- a/xrspatial/geotiff/tests/vrt/test_window.py
+++ b/xrspatial/geotiff/tests/vrt/test_window.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import glob
 import os
 import tempfile
+import threading
 import uuid
 import warnings
 from pathlib import Path
@@ -872,10 +873,17 @@ def test_chunks_is_lazy_does_not_call_internal_reader(monkeypatch, lazy_chunks_t
     vrt_path, _ = lazy_chunks_two_by_two_vrt
     from xrspatial.geotiff import _vrt as vrt_module
     counter = {'calls': 0}
+    # ``compute()`` runs the 16 per-chunk reads through dask's threaded
+    # scheduler, so the counter is incremented from multiple worker threads
+    # concurrently. Guard the read-modify-write with a lock: a bare
+    # ``counter['calls'] += 1`` loses updates under the free-threaded
+    # interpreter (PYTHON_GIL=0), undercounting the decodes (issue #3363).
+    counter_lock = threading.Lock()
     real_read = vrt_module.read_vrt
 
     def counting_read(*args, **kwargs):
-        counter['calls'] += 1
+        with counter_lock:
+            counter['calls'] += 1
         return real_read(*args, **kwargs)
     monkeypatch.setattr(vrt_module, 'read_vrt', counting_read)
     result = _read_vrt(vrt_path, chunks=(64, 64))


brendancol (commented):
## PR Review: thread-safe counter for test_chunks_is_lazy

Test-only change, one file. The diagnosis and fix are right: `compute()` fans the 16 per-chunk reads across dask's threaded scheduler, and under `PYTHON_GIL=0` the bare `counter['calls'] += 1` is a racing read-modify-write that drops an update. A `threading.Lock` around the increment makes the count exact. No blockers.

### Blockers
- None.

### Suggestions
- None. The lock scope is minimal (wraps only the increment, not `real_read`), so it adds no contention on the actual decode and doesn't perturb timing.

### Nits
- None.

### What looks good
- The lazy assertion (`counter['calls'] == 0` before `compute()`) is unaffected — `counting_read` isn't called until compute, so the pre-compute count stays 0.
- Correctly scoped: confirmed the other counter-based tests (`test_metadata.py`, `test_overview.py`, `test_visibility.py`) assert a single non-concurrent call (`== 1`), which can't lose an update, so they don't need the same guard.
- Import added in stdlib alphabetical order; flake8 clean.
- Product code untouched — the chunked decode already runs 16 times and returns correct data; this only fixes the instrumentation.

### Checklist
- [x] Root cause correct (non-atomic counter under free-threading)
- [x] Fix is minimal and thread-safe
- [x] Pre-compute lazy assertion preserved
- [x] No product code touched
- [x] Scope verified against sibling counter tests

PR #3372, opened 2026-06-17
Reject predictor with jpeg2000/lerc/jpeg compression

### LINKED ISSUE #3371: GeoTIFF writer: predictor corrupts pixels with jpeg2000/lerc compression
## Describe the bug

The GeoTIFF writer applies the TIFF horizontal/floating-point differencing predictor (PREDICTOR=2 or 3) to pixel bytes before handing them to the JPEG2000 and LERC codecs, then stamps PREDICTOR=2 in the IFD. GDAL only honors the PREDICTOR tag for byte-oriented entropy codecs (deflate, lzw, zstd, lz4, packbits) and never for jpeg2000/lerc/jpeg. The result is a file whose on-disk pixel bytes are differenced but which no spec-compliant reader will un-difference, so the pixels read back corrupted everywhere except in our own reader (which symmetrically inverts the predictor).

## To Reproduce

```python
import numpy as np, xarray as xr, rasterio

data = (np.arange(64*64, dtype=np.uint16).reshape(64, 64) % 1000).astype(np.uint16)
da = xr.DataArray(data, dims=['y', 'x'],
                  coords={'y': np.arange(64., 0, -1), 'x': np.arange(64.)})

to_geotiff(da, "out.tif", compression='lerc', predictor=2,
           allow_experimental_codecs=True)

with rasterio.open("out.tif") as src:
    rt = src.read(1)
print(np.array_equal(rt, data))                        # False
print(np.abs(rt.astype(int) - data.astype(int)).max())  # 64537
```

For jpeg2000 + predictor=2, GDAL cannot open the file at all.

## Expected behavior

Combining predictor 2 or 3 with jpeg2000/lerc (or jpeg) should be rejected up front with a clear ValueError, matching GDAL, which only honors PREDICTOR for deflate/lzw/zstd/lz4/packbits. A predictor is meaningless for these codecs.

## Root cause

`normalize_predictor(predictor, dtype, compression)` in `xrspatial/geotiff/_encode.py:252` receives the compression tag but never uses it to force the predictor to 1 for codecs that do not support differencing. The three encode paths (`_prepare_strip` at `_encode.py:334`, the second strip path at `_encode.py:484`, the tiled path at `_encode.py:650`) all apply `_apply_predictor_encode` whenever `predictor != 1 and compression != COMPRESSION_NONE`, then pass the differenced bytes to `jpeg2000_compress` / `lerc_compress`. The GPU writer (`xrspatial/geotiff/_writers/gpu.py:696`) calls the same `normalize_predictor`, so it shares the bug. `_validation.py` has no codec-vs-predictor gate (only a predictor=3-requires-float check).

## Severity

High: silent data corruption and non-interoperable files. Requires the non-default `allow_experimental_codecs` opt-in plus an explicit `predictor=` argument, so it is not hit on default writes.

## Environment
- xarray-spatial: main (2026-06-16)
- Found by the accuracy deep-sweep on the geotiff module.

brendancol:
Closes #3371

The GeoTIFF writer applied the TIFF horizontal / floating-point differencing predictor (PREDICTOR=2/3) to pixel bytes before handing them to the jpeg2000 and lerc codecs, then stamped PREDICTOR=2 in the IFD. GDAL only honors the PREDICTOR tag for byte-oriented entropy codecs (deflate, lzw, zstd, lz4, packbits), so the result was a file whose pixels read back corrupted in any spec-compliant reader. GDAL read lerc+predictor=2 with a maxdiff of 64537; jpeg2000+predictor=2 would not open at all.

- `normalize_predictor` already received the `compression` tag but never used it. It now rejects predictor 2 or 3 combined with jpeg/jpeg2000/lerc with a clear `ValueError`. Every writer leg (eager strip + tiled, GPU) routes through this function, so the gate covers them all.
- Byte-oriented codecs keep honouring the predictor; the gate is codec-scoped, not a blanket ban.

Backend coverage: eager (numpy) and GPU writer (`_write_geotiff_gpu`). dask writes route through the same eager encode path.

### Test plan
- [x] `normalize_predictor` raises for predictor 2/3 + jpeg/jpeg2000/lerc; passes through for byte codecs
- [x] `to_geotiff` raises for predictor=2 + {jpeg2000, j2k, lerc} and predictor=3 + {jpeg2000, lerc}
- [x] `to_geotiff` predictor=2 + {deflate, lzw, zstd} still round-trips (regression guard)
- [x] GPU writer raises for predictor=2 + {jpeg2000, lerc} (CUDA-gated)
- [x] Full geotiff write suite green (1175 passed)

### Note on jpeg
`jpeg + predictor=2` previously hit the jpeg branch in `_prepare_strip` (`_encode.py:331`), which returns before the predictor branch, so that write silently succeeded with no predictor applied. It now raises, consistent with the jpeg2000/lerc rejection and with GDAL. This is a move from silently-ignore to hard-error for that one combination.


brendancol (commented):
## PR Review: Reject predictor with jpeg2000/lerc/jpeg compression

### Blockers
None.

### Suggestions
One behavior change worth a line in the PR description (not a code change): `jpeg + predictor=2` used to hit the jpeg branch in `_prepare_strip` (`_encode.py:331`), which returns before the predictor branch, so that write silently succeeded and produced a file with no predictor applied. It now raises. Raising is the correct, GDAL-consistent behavior, but it moves jpeg+predictor from silently-ignore to hard-error, so it is worth noting so it isn't a surprise.

### Nits
- `_PREDICTOR_INCOMPATIBLE_COMPRESSION` maps each tag to a lowercase name used only in the error message. Fine as is; a set plus a separate name lookup would be more code for no gain. No change needed.

### What looks good
- Single chokepoint. All three writer call sites (`_writer.py:477`, `_writer.py:911`, `_writers/gpu.py:696`) pass `comp_tag` from `_compression_tag`, so the guard in `normalize_predictor` covers eager strip, eager tiled, and the GPU writer at once. Confirmed by grep.
- The gate keys on the canonical integer compression tags, which are the only values those call sites ever pass, so it cannot misfire on an unrecognized compression.
- Codec-scoped, not a blanket ban: deflate/lzw/zstd/lz4/packbits + predictor still pass and round-trip (`test_to_geotiff_predictor2_byte_codec_round_trips`).
- Default writes use deflate, which is byte-oriented, so the common path is unaffected.
- Tests cover the unit level (`normalize_predictor`), the eager writer end-to-end for predictor 2 and 3, the positive round-trip regression guard, and the GPU writer (CUDA-gated). The predictor=3-requires-float check is preserved.
- The error message names the offending codec, lists the compatible codecs, and tells the caller how to proceed.

### Checklist
- [x] Behavior matches GDAL (PREDICTOR only honored for byte-oriented codecs)
- [x] All writer backends covered by one guard (eager strip/tiled + GPU)
- [x] No NaN/dask/compute path touched (validation-only change)
- [x] Edge cases covered: predictor 1/False pass-through, predictor 3 float check kept
- [x] Benchmark not needed (validation guard)
- [x] README matrix not applicable (no new function or backend change)
- [x] Docstring updated to note the new rejection

PR #3374, opened 2026-06-17
Read and parse a local file once on the chunked GPU read path (#3373)

### LINKED ISSUE #3373: open_geotiff(gpu=True, chunks=...) reads and parses a local file twice at graph build
## Problem

`_read_geotiff_gpu_chunked` in `xrspatial/geotiff/_backends/gpu.py` reads a local source file off disk twice and parses its IFDs twice, both at dask-graph-build time.

The per-tile byte-count safety-cap block (around lines 1367-1392) opens a `_FileSource(src_path)`, calls `read_all()`, then `parse_header(raw)` and `parse_all_ifds(raw, header)`. It uses the parsed IFDs only for the cap loop and discards them. The GDS qualification block right after (around lines 1398-1404) opens the same file again, calls `read_all()` again, and runs `parse_header` + `parse_all_ifds` a second time.

So every `open_geotiff(gpu=True, chunks=...)` on a local file pays one extra full-file read plus one extra IFD parse. IFD parsing is O(tile_count), so on a COG with 10k+ tiles the second parse costs real time. This is graph-construction overhead paid once per open call, not a per-tile or per-chunk decode cost.

It was recorded as a deferred low-severity item in an earlier performance sweep pass. Raising it now because the cost scales with tile count on large COGs.

## Proposed fix

Read the file once and parse the header and IFDs once, before the cap loop. Run the per-tile cap check on that single parse, then reuse the same `raw` / `header` / `ifds` for the GDS qualification probe (`select_overview_ifd`, `extract_geo_info_with_overview_inheritance`, `_gds_chunk_path_available`). No second `read_all` / `parse_header` / `parse_all_ifds`.

## Constraints to preserve

- The per-tile cap check must still raise `ValueError` (the denial-of-service guard) before any GDS qualification work runs.
- The qualification block currently lets any exception fall through to the CPU path (`_read_geotiff_dask`); that fall-through must stay. A metadata-parse failure must not raise out of this function, it must route to the CPU path exactly as today. A genuine over-cap tile must still raise.
- The local-file-only guard (`isinstance(src_path, str)` and not http/https) stays; the HTTP branch is separate and untouched.
- This block was hardened for a free-threaded (no-GIL) concurrency race; the `_FileSource` context-manager usage and scoping must not be weakened.

## Scope

Graph-build path only: one read and one parse saved per open call. Not a decode-throughput change. geotiff is not in the performance-labeler benchmarked set, so no ASV benchmark runs on this.

brendancol:
Closes #3373.

`_read_geotiff_gpu_chunked` read a local source file off disk twice and parsed its IFDs twice at dask-graph-build time. The per-tile byte-count cap block did `read_all` + `parse_header` + `parse_all_ifds` and discarded the result; the GDS qualification block right after repeated the same read and parse. IFD parsing is O(tile_count), so a COG with 10k+ tiles paid that cost twice (plus a second full-file read) on every `open_geotiff(gpu=True, chunks=...)` call.

## Change

- Read the bytes once and parse the header/IFDs once before the cap loop, run the cap check on that single parse, then reuse the same `raw` / `header` / `ifds` for the GDS qualification probe.

Contract preserved:
- The per-tile cap still raises `ValueError` (the denial-of-service guard) before any GDS work runs.
- A metadata-parse failure, an empty IFD list, or a bad overview level leaves `raw`/`header`/`ifds` as `None` and falls through to the CPU path (`_read_geotiff_dask`) instead of raising, exactly as before.
- The local-file-only guard and the `_FileSource` context-manager scoping (hardened for the free-threaded race) are unchanged.

This is graph-construction overhead only, not a decode-throughput change.

## Backends

GPU read path (`gpu=True, chunks=...`). The dask+cupy chunked read still returns a lazy cupy-backed array; numpy, cupy-eager, and dask+numpy paths are untouched.

## Tests

Added three regression tests in `xrspatial/geotiff/tests/gpu/test_reader.py`:

- [x] A `read_all` / `parse_header` / `parse_all_ifds` invocation counter proving the local chunked GPU read parses once (it counted 2 each before the fix).
- [x] The per-tile DoS cap still raises at graph build.
- [x] A synthetic metadata-parse failure routes to the CPU path without raising, and the result matches the source.
- [x] Full `tests/gpu/` (487 passed, 3 skipped) and `golden_corpus/test_dask_gpu.py` pass.

No ASV: geotiff is not in the performance-labeler benchmarked set, so the `performance` label does not trigger benchmarks here.

### DIFF
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -1413,6 +1413,10 @@ def _read_geotiff_gpu_chunked(source, *, dtype, chunks, overview_level,
     # through so we never lose the ability to return a result.
     try:
         if ifds is not None:
+            # An empty IFD list (non-None, falsy) raises here and the
+            # broad ``except`` below routes it to the CPU path, the same
+            # as a parse failure -- the chunked GPU read never qualifies
+            # a file with no IFDs.
             if not ifds:
                 raise ValueError("No IFDs found in TIFF file")
             ifd = select_overview_ifd(ifds, overview_level)
--- a/xrspatial/geotiff/_backends/gpu.py
+++ b/xrspatial/geotiff/_backends/gpu.py
@@ -1352,56 +1352,67 @@ def _read_geotiff_gpu_chunked(source, *, dtype, chunks, overview_level,
 
     src_path = _coerce_path(source)
 
-    # Per-tile compressed-byte cap, mirroring the eager GPU path and
+    # Local files: read the bytes once and parse the header/IFDs once,
+    # then reuse that single parse for both the per-tile cap check and
+    # the GDS qualification probe below. Parsing the IFDs is
+    # O(tile_count), so a 10k+ tile COG paid that cost twice (plus a
+    # second full-file read) at graph-build time before issue #3373.
+    #
+    # The per-tile compressed-byte cap mirrors the eager GPU path and
     # the CPU readers. The chunked dask + GPU path either qualifies for
-    # the GDS fast
-    # path (handled in ``_read_geotiff_gpu_chunked_gds`` which runs
-    # the same cap on its own metadata parse) or falls through to
-    # ``_read_geotiff_dask`` whose per-chunk ``read_to_array`` calls
-    # apply the cap inside the CPU reader. The check here closes the
-    # window between "qualification probe parses the IFDs" and "the
-    # dispatch decides which path to take" so a forged tile is
-    # rejected at graph-build time rather than at first ``.compute()``.
-    # Sparse tiles (``byte_count == 0``) pass under any positive cap
-    # by design.
+    # the GDS fast path (handled in ``_read_geotiff_gpu_chunked_gds``
+    # which runs the same cap on its own metadata parse) or falls
+    # through to ``_read_geotiff_dask`` whose per-chunk
+    # ``read_to_array`` calls apply the cap inside the CPU reader. The
+    # check here closes the window between "qualification probe parses
+    # the IFDs" and "the dispatch decides which path to take" so a
+    # forged tile is rejected at graph-build time rather than at first
+    # ``.compute()``. Sparse tiles (``byte_count == 0``) pass under any
+    # positive cap by design.
+    raw = header = ifds = None
     if isinstance(src_path, str) and not src_path.startswith(
             ('http://', 'https://')):
         try:
-            with _FileSource(src_path) as _cap_fs:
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Read and parse a local file once on the chunked GPU read path (#3373)

None.

None.

- `xrspatial/geotiff/_backends/gpu.py:1416`: the `if not ifds: raise ValueError("No IFDs found in TIFF file")` inside the qualification `try` is now only reachable when `parse_all_ifds` returns an empty list (a non-None, falsy value). That is the same path the old code took, and the broad `except` below routes it to the CPU fallback, so behavior is unchanged. A one-line comment noting that an empty list is intentionally routed to CPU would save the next reader the trace. Optional.

### What looks good
- The over-cap `ValueError` is raised in the cap loop (line ~1400), which sits before and outside the qualification `try/except`, so the denial-of-service guard still propagates. Verified.
- Parse failure, an empty IFD list, and a bad overview level all leave `raw`/`header`/`ifds` as `None` or fall through the broad `except`, routing to `_read_geotiff_dask` without raising, which matches the prior two-parse semantics.
- `header.byte_order` and `raw` are only dereferenced under `if ifds is not None:`, where a successful parse guarantees both are set.
- The `_FileSource` context-manager scoping (hardened for the free-threaded race) is preserved: the read still happens inside the `with` block.
- The invocation-counter test has teeth. It counts 2 reads/parses on the pre-fix code and 1 after, so the regression cannot silently return.

### Checklist
- [x] Behavior matches the prior implementation (cap raises, failures fall through)
- [x] GPU dask path still returns a lazy cupy-backed array
- [x] Cap / DoS guard preserved
- [x] Edge cases (empty IFDs, bad overview, parse failure) covered
- [x] No premature materialization or unnecessary copies (one fewer read, one fewer parse)
- [x] Benchmark not needed (graph-build path, geotiff not in the benchmarked set)
- [x] README feature matrix not applicable (no API change)
- [x] Tests present and meaningful (487 passed, 3 skipped in tests/gpu/)

brendancol (commented):
## PR Review (follow-up): #3373

The single optional nit from the first pass is addressed in d5c96b07: the empty-IFD-list path now carries a comment explaining it falls through to the CPU reader.

No remaining findings. The fix preserves the cap/DoS raise, the parse-failure and empty-IFD fall-through to the CPU path, and the `_FileSource` scoping. The 3373 regression tests still pass (3 passed), and the wider `tests/gpu/` suite was green (487 passed, 3 skipped).

PR #3377, opened 2026-06-17
Expose coregistered reads through the xarray engine via like= (#3376)

### LINKED ISSUE #3376: Expose coregistered reads through the xarray backend engine via a target-grid kwarg
## Summary

The xarray backend engine added in #3375 (`engine="xrspatial"`) forwards to the standalone `open_geotiff` function, so it deliberately does not expose the coregistered-read options (`coregister`, `auto_reproject`, `resampling`). Those live on the `.xrs.open_geotiff` accessor because they reproject and resample onto a target array's grid, and `xr.open_dataset` opens a single source from scratch with no target. This issue tracks whether to expose coregistered reads through the engine via a target-grid backend kwarg.

## Current behavior

```python
import xarray as xr
xr.open_dataset("scene.tif", engine="xrspatial",
                backend_kwargs={"coregister": True})
# TypeError: open_geotiff() got an unexpected keyword argument 'coregister'
```

Coregistration today goes through the accessor on an existing array:

```python
target.xrs.open_geotiff("scene.tif", coregister=True, auto_reproject=True)
```

This is documented in docs/source/reference/geotiff.rst (the backend-engine section notes coregister/auto_reproject are accessor-only).

## Proposed behavior

Add a way to pass a target grid through the engine, e.g. a `like=` (or `target=`) backend kwarg holding a DataArray/Dataset whose grid the read should be coregistered onto:

```python
xr.open_dataset("scene.tif", engine="xrspatial",
                backend_kwargs={"like": target, "coregister": True,
                                "auto_reproject": True})
```

The engine would route to the accessor path (`_open_geotiff_windowed(target, source, ...)`) instead of the standalone function when `like=` is supplied.

## Design questions

- Kwarg name: `like=` (matches `xr.zeros_like`/`ones_like` intuition) or `target=`?
- Should `like=` be the trigger, or should `coregister=True` without a `like=` raise a clear error pointing at it?
- The engine returns a Dataset; the accessor's coregister path returns a DataArray on the target grid. Variable naming would follow the same default_name/stem rule as the plain engine path.
- Multi-file: `open_mfdataset(..., backend_kwargs={"like": target})` would coregister every source onto the same grid. Worth verifying it composes.
- GPU/.vrt/allow_rotated combos already raise on the accessor coregister path; the engine wrapper would inherit those rejections for free.

## Why

Lets users coregister straight through the standard xarray API without dropping to the accessor, and makes `open_mfdataset` onto a fixed grid a one-liner. Out of scope for #3375, which only wraps the standalone reader.

brendancol:
Closes #3376.

Adds a `like=` backend kwarg to the `xrspatial` xarray engine. When you pass it, the engine routes the read through `like`'s `.xrs.open_geotiff` accessor instead of the standalone reader, so the coregistered-read options work through the standard `xr.open_dataset` / `open_mfdataset` API.

## What changed

- `like=` (a DataArray or Dataset) triggers the coregister path. `coregister`, `auto_reproject`, `resampling`, and `var` are forwarded to the accessor.
- A Dataset `like` dispatches to the Dataset accessor, so `var=` picks the variable used for backend/CRS inference.
- The returned variable is named by the same `default_name` / source-stem rule as the plain engine path.
- `open_mfdataset` with a shared `like=` coregisters every source onto one grid in a single call.
- The accessor's GPU / `.vrt` / `allow_rotated` rejections apply through the engine without extra code.
- Passing `coregister` / `auto_reproject` / `resampling` / `var` without `like=` raises a `ValueError` that names `like=`, instead of the opaque `TypeError` the standalone reader would throw for an unknown kwarg.

## Design decisions

The issue left two questions open. I went with `like=` for the kwarg name (the issue's first option, and it matches the `xr.zeros_like` / `ones_like` convention), and made `like=` the trigger: the coregister kwargs without it raise the pointed error rather than silently doing nothing.

## Backend coverage

The coregister path is CPU-only (numpy and dask+numpy). GPU and `.vrt` sources already raise on the accessor, and the engine inherits that. The plain read path's backend support is unchanged.

## Test plan

- [x] coregister via engine matches the accessor output (same CRS, and CRS-mismatch reproject)
- [x] auto_reproject via engine matches the accessor
- [x] Dataset `like` with `var=`
- [x] variable naming (stem and `default_name` override)
- [x] `open_mfdataset` coregisters two files onto one grid
- [x] guard raises `ValueError` for each coregister kwarg passed without `like=`
- [x] GPU rejection inherited through the engine
- [x] existing backend tests still pass

### DIFF
--- a/docs/source/reference/geotiff.rst
+++ b/docs/source/reference/geotiff.rst
@@ -270,13 +270,31 @@ ambiguous when another raster backend (e.g. rioxarray's ``rasterio``) is
 installed and also claims those extensions; xarray then raises and asks
 for an explicit ``engine=``.
 
-The engine forwards to the standalone ``open_geotiff`` function, so the
-coregistered-read options (``coregister``, ``auto_reproject``,
-``resampling``) are *not* available through it; they live on the
-``.xrs.open_geotiff`` accessor because they reproject and resample onto a
-target array's grid, and the engine opens a single source from scratch
-with no target. Passing them through ``backend_kwargs`` raises
-``TypeError``. Use the accessor on the target array instead, e.g.
+By default the engine forwards to the standalone ``open_geotiff``
+function, which opens a single source from scratch with no target grid.
+The coregistered-read options (``coregister``, ``auto_reproject``,
+``resampling``) reproject and resample onto a target array's grid, so
+they need that target. Pass it as a ``like=`` backend kwarg (a DataArray
+or Dataset); the engine then routes through ``like``'s
+``.xrs.open_geotiff`` accessor:
+
+.. code-block:: python
+
+    xr.open_dataset(
+        "scene.tif", engine="xrspatial",
+        backend_kwargs={"like": target, "coregister": True,
+                        "auto_reproject": True})
+
+When ``like`` is a Dataset, the ``var=`` backend kwarg picks the variable
+used for backend/CRS inference. ``open_mfdataset`` with a shared ``like=``
+coregisters every source onto the same grid in one call. The returned
+variable is named by the same ``default_name`` / source-stem rule as the
+plain engine path, and the accessor's GPU / ``.vrt`` / ``allow_rotated``
+rejections apply through the engine too.
+
+Passing ``coregister`` / ``auto_reproject`` / ``resampling`` / ``var``
+*without* a ``like=`` raises ``ValueError`` pointing at ``like=``. The
+accessor on the target array remains available directly, e.g.
 ``target.xrs.open_geotiff("scene.tif", coregister=True)``.
 
 Coregistered reads (experimental)
--- a/xrspatial/geotiff/_xarray_backend.py
+++ b/xrspatial/geotiff/_xarray_backend.py
@@ -29,6 +29,23 @@ argument to ``open_dataset``, so it cannot travel through
 a dask-backed dataset (xarray wraps the eager read)::
 
     xr.open_dataset("dem.tif", engine="xrspatial", chunks={})
+
+Coregistered reads (``coregister`` / ``auto_reproject`` / ``resampling``)
+reproject and resample a source onto an existing array's grid, so they
+need a target grid that the plain ``open_dataset`` path does not have.
+Pass that target as a ``like=`` backend kwarg (a DataArray or Dataset);
+the engine then routes to the ``.xrs.open_geotiff`` accessor on ``like``
+instead of the standalone reader::
+
+    xr.open_dataset(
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Expose coregistered reads through the xarray engine via `like=`

None.

- `_xarray_backend.py:106` — when `like` is not a DataArray or Dataset (say a path or a bare numpy array), `like.xrs` raises a generic `AttributeError`, which is harder to read than the pointed `ValueError` the no-`like` path gives. A short `isinstance(like, (xr.DataArray, xr.Dataset))` guard with a clear message would round out the error handling.

- `_xarray_backend.py:110-115` — when the only offending kwarg is `var`, the message ("reproject/resample a source onto a target grid") doesn't quite fit, since `var` just selects the inference variable on a Dataset target. The `like=` pointer is still correct, so this is cosmetic.

### What looks good
- The routing is a thin wrapper: `like` present → accessor, else standalone, with the variable-naming and `drop_variables` promotion shared across both paths. No duplicated read logic.
- Tests assert parity against the accessor output rather than just "it runs", and cover same-CRS, CRS-mismatch reproject, Dataset `var=`, naming, `open_mfdataset`, the per-kwarg guard, and the inherited GPU rejection.
- GPU / `.vrt` / `allow_rotated` rejections are inherited from the accessor instead of being re-implemented, so they can't drift out of sync.
- `open_dataset_parameters` is left untouched, so `like` travels through `backend_kwargs` like the other read options and doesn't trip xarray's CF decoder collision.

### Checklist
- [x] Routing matches the accessor (parity asserted in tests)
- [x] All implemented backends consistent (CPU-only path; GPU rejection inherited)
- [x] NaN handling unchanged (delegated to accessor)
- [x] Edge cases covered by tests (guards, naming, mfdataset)
- [x] Dask path exercised via `open_mfdataset`
- [x] Benchmark not needed (thin routing wrapper)
- [x] Docstrings / docs updated

brendancol (commented):
## Follow-up review (after 7f47195b)

- **Suggestion (like= type guard)** — fixed. `_xarray_backend.py:102-107` now raises `TypeError` naming the expected type when `like` is not a DataArray/Dataset, instead of falling through to an `AttributeError` on `.xrs`. Covered by `test_non_array_like_raises_typeerror`.
- **Nit (guard wording)** — fixed. `_xarray_backend.py:117-122` reads "only apply when reading onto a target grid", which fits `var` as well as the reproject/resample kwargs.

No new issues. Tests: 15 coregister cases plus the 17 existing backend tests pass locally.

PR #3382, opened 2026-06-17
Vectorize _sort_and_stride 3D reindex, drop deepcopy (#3381)

### LINKED ISSUE #3381: zonal: redundant deepcopy and Python loop in _sort_and_stride 3D branch
## Description

The 3D branch of `_sort_and_stride` in `xrspatial/zonal.py` builds the per-zone sorted values like this:

```python
values_by_zones = copy.deepcopy(values).reshape(
    values_shape[0], values_shape[1] * values_shape[2])
for i in range(values_shape[0]):
    values_by_zones[i] = values_by_zones[i][sorted_indices]
```

Two things are wasteful here. The `copy.deepcopy(values)` allocates a full duplicate of the entire values array before any reindexing happens, and a Python `for` loop then reindexes each row by `sorted_indices` one row at a time.

A single vectorized fancy-index does the same work:

```python
values_by_zones = values.reshape(values_shape[0], -1)[:, sorted_indices]
```

Fancy indexing returns a fresh array, so the input is never mutated and the explicit deepcopy is not needed. The Python loop goes away too.

## Impact

This 3D path runs in `crosstab()` on the numpy backend, and once per block on the dask+numpy backend, so the redundant full-array copy is paid per block at scale. The cupy backend rejects 3D input, so it is not affected.

## Backends affected

numpy, dask+numpy.

## Found by

`/sweep-performance` sweep of the zonal module. OOM verdict SAFE, bottleneck compute-bound.

brendancol:
Closes #3381

## What changed

- The 3D branch of `_sort_and_stride` deep-copied the entire values array and then reindexed each layer by `sorted_indices` in a Python loop. It now does the reindex with one vectorized fancy-index, `values.reshape(n, -1)[:, sorted_indices]`. Fancy indexing returns a fresh array, so the input is never mutated and the explicit copy is gone.
- Removed the now-unused `import copy`.

This path runs in `crosstab()` on the numpy backend and once per block on dask+numpy, so the redundant full-array copy was paid per block at scale. The cupy backend rejects 3D input, so it is unaffected.

## Backend coverage

numpy, dask+numpy (the only backends that reach the 3D `_sort_and_stride` path).

## Test plan

- [x] New `test_sort_and_stride_3d_no_mutation_and_parity` pins input non-mutation and layer-wise sort parity.
- [x] Existing 3D crosstab tests pass (`test_crosstab_3d_count`, `test_crosstab_3d_agg_method`, `test_nodata_values_crosstab_3d`), all of which already assert the input is unmodified.
- [x] Full `test_zonal.py` (194) and `test_zonal_backend_coverage_2026_05_27.py` (43) pass.


brendancol (commented):
## PR Review: Vectorize _sort_and_stride 3D reindex, drop deepcopy (#3381)

None.

None.

- `test_zonal.py` (new `test_sort_and_stride_3d_no_mutation_and_parity`): the parity assertion recomputes `expected` with the same expression the implementation now uses (`values.reshape(n, -1)[:, sorted_indices]`), so it is not an independent oracle. The non-mutation assertion also would not have failed against the old deepcopy code, since the old path mutated a copy rather than the input. The test still earns its place as a guard against a future in-place-mutation or shape regression, and `test_crosstab_3d_count` / `test_crosstab_3d_agg_method` already provide the independent reference-value check. No change needed, just noting the test's scope.

### What looks good
- The replacement is equivalent: reshape returns a view when the array is contiguous, and the fancy-index `[:, sorted_indices]` always allocates a fresh array, so the input is never touched (confirmed empirically).
- The deepcopy and the Python row loop both collapse into one vectorized op.
- Removing the now-unused `import copy` is the right cleanup.
- Backend scope is correct: only numpy and dask+numpy reach the 3D branch; cupy rejects 3D upstream, so nothing GPU-side is affected.
- Existing 3D crosstab tests already assert the input is unmodified and check against reference values.

### Checklist
- Algorithm matches reference: n/a (mechanical refactor, output identical)
- All implemented backends consistent: yes (numpy + dask+numpy share the helper; cupy unaffected)
- NaN handling correct: unchanged (NaN filtering happens downstream of this helper)
- Edge cases covered: yes (existing 3D tests + new unit test)
- Dask chunk boundaries handled: n/a (per-block helper, no chunk logic changed)
- No premature materialization or unnecessary copies: improved (one full-array copy removed)
- Benchmark exists or not needed: existing benchmarks/benchmarks/zonal.py covers stats and regions; crosstab 3D is unbenchmarked but this is a refactor, not a new function
- Docstrings present and accurate: n/a (private helper)

PR #3385, opened 2026-06-18
Add rasterize() validation-guard test coverage (#3383)

### LINKED ISSUE #3383: rasterize(): dtype/fill/burn-value validation guards and non-finite geometry drop have no tests
`rasterize()` has several input-validation guards that raise `ValueError` (or warn) before any backend allocation. Each guard has a detailed docstring and traces back to a specific issue, but none of them has an error-path test in `xrspatial/tests/test_rasterize.py`. A regression that weakened or removed a guard would not be caught.

Untested guards:

- Non-representable fill for an integer/bool dtype: NaN fill into an integer dtype, an out-of-range integer fill (`fill=-9999` into `uint8`), and any non-False fill into `bool`. Source `rasterize.py` ~line 4378, issues #2504 / #3054.
- Burn value above the float64 safe-integer range (`|value| > 2**53 - 1`) with an integer output dtype. Source ~line 4446, issue #3056.
- Non-finite burn value (NaN / inf) with an integer or bool dtype, plus the `merge='count'` exemption that lets NaN attributes through because count never reads property values. Source ~line 4474, issue #3085.
- Non-finite geometry coordinates (NaN / inf in a geometry's coords): the geometry is dropped with a `UserWarning`. Source `rasterize.py` lines 230-265 (`_warn_nonfinite_geometries`), issue #3295.

Found by `/sweep-test-coverage`.

Proposed fix: add error-path tests asserting each guard raises (or warns) with the documented behavior, plus valid-input and the count-exemption negative cases so the guards stay narrow. The guards all run on the CPU path before backend dispatch, so the tests need no GPU. Test-only, no source change.

brendancol:
Closes #3383.

Adds error-path tests for the `rasterize()` input-validation guards that had no coverage. The guards all run on the CPU path before backend dispatch, so the tests need no GPU.

- `TestFillRepresentableGuard`: NaN fill into an integer dtype, an out-of-range integer fill into `uint8`, and a non-False fill into `bool` are rejected (#2504 / #3054). Negative cases confirm valid integer/bool fills and a float NaN fill still pass.
- `TestBurnValueSafeIntegerGuard`: a burn value with `|value| > 2**53 - 1` into an integer dtype is rejected (#3056); the `2**53 - 1` boundary and float-dtype exemption pass.
- `TestNonFiniteBurnValueGuard`: NaN / +/-inf burn values into integer or bool dtypes are rejected (#3085); the `merge='count'` exemption and float-dtype path pass.
- `TestNonFiniteGeometryCoordsDropped`: a geometry with NaN / inf coordinates is dropped with a `UserWarning` (#3295), covering the list and GeoDataFrame paths.

Test-only. No source change.

Backend coverage: the guards are backend-independent (they fire before dispatch), so the tests exercise the shared CPU validation path. No new cupy / dask tests are needed.

Test plan:
- [x] 18 new tests pass locally on a CUDA host
- [x] full `test_rasterize.py` module passes (233 passed, 2 skipped)


brendancol (commented):
## PR Review: Add rasterize() validation-guard test coverage (#3385)

Test-only PR. It adds four test classes to `test_rasterize.py` covering input-validation guards in `rasterize()` that had no error-path tests. No source changes. I read the new tests and the four guards they exercise (`rasterize.py` lines 4378-4487 and 230-265).

None.

None.

- `test_nonfinite_geometry_in_gdf_dropped_with_warning` also triggers a shapely `RuntimeWarning` ("invalid value encountered in bounds") next to the expected `UserWarning`. `pytest.warns` matches on at least one warning so the test passes, but the stray warning lands in the pytest warnings summary. Could wrap the GeoDataFrame construction in `warnings.simplefilter` to keep the summary clean. Not worth blocking on.

### What looks good
- Each guard is paired with a negative case (valid integer/bool/float fill, the `2**53-1` boundary, the `merge='count'` exemption, float-dtype burn), so the tests confirm the guards are narrow rather than just that they reject.
- The `match=` regexes line up with the real error substrings in the source.
- The non-finite burn test parametrizes nan / +inf / -inf instead of testing one.
- The PR is right that the guards run on the CPU path before backend dispatch, so no GPU or cross-backend parity test is needed. All three validation blocks run before the `chunks` / `gpu` branch in `rasterize()`.

### Checklist
- NaN handling: tests assert both the reject paths and the float-dtype accept paths.
- Edge cases: this PR is the edge-case coverage.
- Docstrings: each test class has a docstring citing the guarded issue.
- Backend parity / dask / benchmark / README matrix: not applicable (test-only, guards are pre-dispatch).

brendancol (commented):
## Follow-up review (#3385)

Addressed the single nit from the first pass. `test_nonfinite_geometry_in_gdf_dropped_with_warning` now carries a `@pytest.mark.filterwarnings` for the incidental shapely "invalid value encountered in bounds" `RuntimeWarning`, so it no longer shows up in the pytest warnings summary. The expected `UserWarning` assertion is unchanged.

Re-ran the module: 233 passed, 2 skipped (the 2 remaining warnings are the pre-existing GPU custom-merge `UserWarning`s, not from this PR).

No outstanding findings. Disposition: 1 nit fixed, 0 dismissed, 0 deferred.

PR #3391, opened 2026-06-19
proximity: match GPU max_distance precision to the CPU brute-force (#3389)

### LINKED ISSUE #3389: proximity: max_distance comparison precision differs between CPU and GPU backends
**Describe the bug**

In `xrspatial/proximity.py`, the CPU brute-force distance helper `_distance` casts its result to `np.float32` before returning (around line 261). The brute-force search then compares that float32-rounded distance against `max_distance` via `best_dist <= max_distance`. The CUDA kernel `_proximity_cuda_kernel` and its device helpers compute and compare in float64.

When `max_distance` lands between the float32-rounded distance and the true float64 distance of a target, the backends disagree: the CPU path treats the target as in range and returns a finite value, while the GPU path treats it as out of range and returns NaN (or the reverse).

This hits every path that runs the brute-force search: GREAT_CIRCLE proximity/allocation/direction, and EUCLIDEAN/MANHATTAN allocation/direction. EUCLIDEAN/MANHATTAN proximity on numpy goes through the cKDTree path, which bounds differently.

**Reproducer**

A single-target lat/lon raster, GREAT_CIRCLE proximity, with `max_distance` set between the float32 and float64 distance of one column:

```python
from xrspatial.proximity import great_circle_distance

lon = np.linspace(0, 39, 40); lat = np.array([0.0])
data = np.zeros((1, 40)); data[0, 0] = 1.0
j = 20
d64 = great_circle_distance(lon[0], lon[j], lat[0], lat[0])
md = (float(np.float32(d64)) + d64) / 2.0   # between f32 and f64 distance

r_np = proximity(xr.DataArray(data, dims=['lat','lon'], coords={'lon':lon,'lat':lat}),
                 x='lon', y='lat', distance_metric='GREAT_CIRCLE', max_distance=md)
r_cp = proximity(xr.DataArray(cp.asarray(data), dims=['lat','lon'], coords={'lon':lon,'lat':lat}),
                 x='lon', y='lat', distance_metric='GREAT_CIRCLE', max_distance=md)

print(r_np.data[0, j])          # 2226389.75 (finite)
print(r_cp.data.get()[0, j])    # nan
```

numpy and dask+numpy return the finite distance; cupy and dask+cupy return NaN for the same input.

**Expected behavior**

All four backends should agree on whether a target is within `max_distance`. The comparison should use the same precision everywhere.

**Severity**

Medium. The divergence is silent, with no error raised, but it only surfaces when `max_distance` is set with sub-ulp precision relative to a target's distance. That is an adversarial input rather than typical usage.

**Found by** /sweep-accuracy on the proximity module.

brendancol:
Closes #3389

The CPU brute-force search compares a float32-rounded distance against `max_distance` (`_distance` returns `np.float32(d)`), while the CUDA kernel compared the float64 distance. A `max_distance` sitting in a target's float32 ulp gap then qualified the target on numpy/dask+numpy but rejected it on cupy/dask+cupy for the same input.

- Round `best_dist` to float32 in `_proximity_cuda_kernel` before the `best_dist <= max_distance` range test, so the in-range decision matches the CPU path.
- Add two regression tests pinning numpy == cupy at the float32 boundary: GREAT_CIRCLE proximity and EUCLIDEAN allocation (both run the brute-force on CPU and the kernel on GPU).

Backend coverage: the change is in the shared CUDA kernel, so it covers cupy and dask+cupy (which reuses the kernel per chunk). numpy and dask+numpy were already on the float32 side and are unchanged.

A separate, pre-existing edge is out of scope here: the bounded dask+numpy EUCLIDEAN allocation path can still return NaN for a target sitting exactly at a float32-ulp `max_distance`, via the halo-sizing path rather than the kernel. It predates this change and is noted in the sweep state for follow-up.

Test plan:
- [x] `pytest xrspatial/tests/test_proximity.py` (518 passed on a CUDA host, including cupy and dask+cupy)
- [x] New boundary tests fail before the fix (cupy returns NaN) and pass after


brendancol (commented):
## PR Review: proximity: match GPU max_distance precision to the CPU brute-force (#3389)

None.

None.

- The bounded-dask EUCLIDEAN-allocation float32-ulp edge is recorded in the PR body and the sweep state, but it is out of scope here. A one-line follow-up issue would make it easier to track than a buried note.

### What looks good
- Root cause is correct. `_distance` returns `np.float32(d)`, so the CPU compares a float32-rounded distance against `max_distance`, while `_proximity_cuda_kernel` compared the float64 distance. Rounding `best_dist` to float32 before the `<= max_distance` test makes the in-range decision identical on both.
- The stored PROXIMITY value does not change in practice: `out` is float32, so the previous float64 `best_dist` was already cast on store. The fix only moves the cast ahead of the comparison.
- The tests pin the real divergence. Each sets `max_distance` strictly between the float32 and float64 distance (asserted with `float(np.float32(d64)) < md < d64`) and checks numpy against cupy. They fail before the fix (cupy returns NaN) and pass after.
- Both brute-force output branches of the kernel are covered: GREAT_CIRCLE proximity and EUCLIDEAN allocation.
- No dispatch change. The fix lives in the shared kernel, so cupy and dask+cupy (which reuses the kernel per chunk) both pick it up.

### Checklist
- [x] Algorithm matches the CPU reference behaviour
- [x] All implemented backends produce consistent results at the boundary
- [x] NaN handling correct (in-range decision now matches across backends)
- [x] Edge case (float32 ulp boundary) covered by tests
- [x] Dask chunk boundaries unaffected (kernel-level change, dask+cupy reuses it)
- [x] Benchmark not needed (no performance-relevant change)
- [x] Docstrings unaffected (internal kernel comment added)

brendancol:
Filed the out-of-scope bounded-dask EUCLIDEAN-allocation float32-ulp edge as #3392 for tracking (review nit).

PR #3396, opened 2026-06-19
fire: declare float32 dtype on dask+numpy backend (#3394)

### LINKED ISSUE #3394: fire: dask+numpy backend reports float64 dtype but computes float32
## Description

The dask+numpy backend in `xrspatial/fire.py` reports the wrong output dtype. Six of the seven `map_blocks` wrappers pass `meta=np.array(())` with no `dtype=` argument. `np.array(())` is float64, so the dask-backed DataArray reports `dtype == float64`, but the `@ngjit` kernels all allocate and return float32 (`np.full(..., dtype=np.float32)`). numpy, cupy, and dask+cupy all report float32, so dask+numpy is the odd one out.

Affected: `dnbr`, `rdnbr`, `fireline_intensity`, `flame_length`, `rate_of_spread`, `kbdi`. The seventh wrapper, `burn_severity_class`, passes `dtype=np.int8` and is fine.

## Reproduce

```python
import numpy as np
from xrspatial.fire import dnbr

pre = np.array([[0.5, 0.6], [0.4, 0.2]], dtype=np.float32)
post = np.array([[0.1, 0.3], [0.35, 0.1]], dtype=np.float32)

np_r = dnbr(create_test_raster(pre, 'numpy'), create_test_raster(post, 'numpy'))
dk_r = dnbr(create_test_raster(pre, 'dask+numpy'), create_test_raster(post, 'dask+numpy'))

print(np_r.dtype)                 # float32
print(dk_r.dtype)                 # float64  <- wrong, lazy metadata
print(dk_r.data.compute().dtype)  # float32  <- actual
```

The declared dtype and the computed dtype disagree until the graph runs. Anything that reads `result.dtype` before `.compute()` (buffer sizing, dtype-based dispatch) gets the wrong answer, and only on dask+numpy.

## Fix

Pass `dtype=np.float32` to the six `map_blocks` calls, matching the `f4` the kernels produce.

Found during an accuracy sweep (category 5, backend inconsistency).

brendancol:
Closes #3394.

The dask+numpy `map_blocks` wrappers for six fire functions declared the wrong output dtype: they passed `meta=np.array(())` (float64) with no `dtype=`, so the lazy DataArray reported float64 while the `@ngjit` kernels return float32. numpy, cupy, and dask+cupy all reported float32, so dask+numpy was the odd one out.

- Pass `dtype=np.float32` to `_dnbr_dask`, `_rdnbr_dask`, `_fli_dask`, `_fl_dask`, `_ros_dask`, `_kbdi_dask`. `_bsc_dask` already passed `dtype=np.int8` and is unchanged.
- Add a parametrized regression test asserting the dask+numpy declared dtype (and the computed dtype) matches the numpy backend for every fire function.

Backends: only the dask+numpy declared-dtype metadata changes. Computed values are unaffected on all four backends. The fix matches the existing `_bsc_dask` pattern.

Test plan:
- [x] `pytest xrspatial/tests/test_fire.py` (61 passed, includes cupy + dask+cupy on a GPU host)
- [x] Confirmed the 6 new dtype assertions fail before the fix and pass after


brendancol (commented):
## PR Review: fire: declare float32 dtype on dask+numpy backend (#3394)

None.

None.

- Now that `dtype=np.float32` is passed explicitly, the trailing `meta=np.array(())` is redundant, since dask uses `dtype=` as the authoritative declared dtype. Keeping it matches the existing `_bsc_dask` style, so leaving it is fine. Dropping it would be marginally cleaner but not worth the churn.

### What looks good
- The declared dtype now lines up with the kernels' `np.full(..., dtype=np.float32)` allocation and with the other three backends. I confirmed `dtype=` drives the lazy declared dtype regardless of the float64 meta.
- The regression test covers all seven fire functions and checks both the lazy declared dtype and the computed dtype against the numpy backend. I ran it with the source change reverted: the six affected cases fail and `_bsc_call` passes (it was already correct).
- Small, repetitive change: six identical one-line kwarg additions, no algorithm or value change on any backend.

### Checklist
- [x] Algorithm matches reference (no algorithm change; metadata only)
- [x] All implemented backends produce consistent results (dtype now uniform float32)
- [x] Edge cases covered by tests (dtype parity parametrized over every function)
- [x] Dask chunk boundaries handled correctly (element-wise, no overlap)
- [x] Benchmark not needed (metadata fix)

PR #3397, opened 2026-06-19
Style: drop dead code + fix import ordering in fire.py

### LINKED ISSUE #3395: Style: dead code + import-ordering drift in xrspatial/fire.py
`xrspatial/fire.py` has a few dead-code leftovers from a refactor. The project's own linters catch them, but neither flake8 nor isort runs in CI, so the drift sits there unnoticed.

flake8 (setup.cfg, max-line-length=100):
- `fire.py:20:1 F401` `'math.log'` imported but unused
- `fire.py:105:5 F841` local variable `'S_T'` assigned but never used

isort (line_length=100):
- The `xrspatial.utils` import block uses the one-per-line layout; isort wants the grid layout. One reordering diff.

Both F-code findings are dead code. `log` is never called in the module. `S_T` (total mineral content, `row[5]` of the Anderson-13 table) gets read into a local in `_rothermel_fuel_constants` and then nothing uses it. The neighbouring `S_e` and `rho_p` locals are used; only `S_T` is dead. The column itself stays in `ANDERSON_13` and its documenting comment, so this only drops the unused binding.

Fix: drop `log` from the `math` import, delete the `S_T` local, and let isort reorder the utils import. No behaviour change, and nothing here touches runtime logic so it applies the same across the numpy/cupy/dask paths.

brendancol:
Closes #3395.

Style sweep on `xrspatial/fire.py`. The project's flake8 and isort configs (setup.cfg) flag dead code that nothing in CI checks for.

What changed:
- Drop unused `log` from the math import (flake8 F401)
- Drop unused `S_T` local in `_rothermel_fuel_constants` (flake8 F841); `S_e` and `rho_p` next to it are still used, and the `ANDERSON_13` table column + its comment are untouched
- isort reorders the `xrspatial.utils` import block to the grid layout (Cat 4)

Categories addressed: Cat 3 (F401, F841), Cat 4 (isort). No Cat 1/2/5 findings.

No behavioural change. Removing the `log` import and `S_T` local touches no runtime logic, so all four backends (numpy, cupy, dask+numpy, dask+cupy) are unaffected.

Test plan:
- [x] `flake8 xrspatial/fire.py` is clean
- [x] `isort --check-only xrspatial/fire.py` is clean
- [x] `pytest xrspatial/tests/test_fire.py`: 54 passed


brendancol (commented):
Domain-aware review (style sweep, fire.py)

Scope: 3-line dead-code/import cleanup plus a sweep-state CSV row. Verified the diff matches the stated intent.

Blockers: none.

Suggestions: none.

Nits: none.

Checks I ran:
- `log` is gone from the import and has no remaining call sites in the module.
- `S_T` local is removed; the only surviving `S_T` is in the ANDERSON_13 column-layout comment (line 47), which is correct to keep since the table column itself is unchanged. `S_e`/`rho_p` (the adjacent locals) are still read by the Rothermel math, so the deletion is surgical.
- isort reflow of the `xrspatial.utils` block is a pure reordering; the imported names are identical before and after.
- flake8 and isort are both clean on the file; `pytest xrspatial/tests/test_fire.py` is 54 passed. No runtime logic touched, so backend parity (numpy/cupy/dask) is unaffected by construction.

No behavioural change. Good to merge once CI is green.

PR #3409, opened 2026-06-20
Add missing morphology accessor methods

### LINKED ISSUE #3399: morph_gradient, white_tophat, black_tophat missing from .xrs accessor
## Reason or Problem

The `.xrs` accessor exposes only four of the seven public morphology functions. `morph_erode`, `morph_dilate`, `morph_opening`, and `morph_closing` have accessor methods on both `XrsSpatialDataArrayAccessor` and `XrsSpatialDatasetAccessor`. `morph_gradient`, `morph_white_tophat`, and `morph_black_tophat` do not.

All seven functions are public. They are re-exported from `xrspatial/__init__.py`, documented with full docstrings, and covered by tests (`test_morphology_derived.py`). Calling `da.xrs.morph_erode(...)` works, but `da.xrs.morph_gradient(...)` raises `AttributeError`. The accessor surface is inconsistent within one function family.

This looks like an oversight rather than a design choice. The four base ops landed in #949, their accessor methods in #1042. The three derived ops landed later in #1026, and the accessor was never updated to match.

## Proposal

Add accessor methods for the three derived functions to both accessor classes, following the existing forwarding pattern:

```python
def morph_gradient(self, **kwargs):
    from .morphology import morph_gradient
    return morph_gradient(self._obj, **kwargs)
```

**Value:** every public morphology function becomes reachable through `.xrs`.

## Drawbacks

None. The change is additive and does not touch existing methods.

## Stakeholders and Impacts

Users of the `.xrs` accessor for morphological operations. No impact on the direct function API.

brendancol:
Closes #3399

The `.xrs` accessor exposed only four of the seven public morphology functions. This adds the missing three to both accessor classes.

- Add `morph_gradient`, `morph_white_tophat`, and `morph_black_tophat` methods to `XrsSpatialDataArrayAccessor` and `XrsSpatialDatasetAccessor`, following the existing forwarding pattern.
- Add accessor tests for the morphology family (none existed before) and extend the expected-methods checks to guard all seven methods.

The change forwards to the existing morphology functions, so backend coverage matches those: numpy, cupy, dask+numpy, dask+cupy.

Test plan:
- [x] `pytest xrspatial/tests/test_accessor.py` (69 passed)
- [x] New tests assert accessor output is identical to the direct function call for erode, gradient, white_tophat, black_tophat (DataArray) and gradient (Dataset)


brendancol (commented):
## PR Review: Add missing morphology accessor methods

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The three new methods follow the existing forwarding pattern byte-for-byte (lazy import, `**kwargs` passthrough), so they inherit whatever backend support the underlying functions provide.
- Both accessor classes are updated, so DataArray and Dataset stay in sync.
- Tests assert the accessor output is identical to the direct function call, which is the right contract for a thin forwarder.
- The expected-methods checks now list all seven morph methods, so a future regression that drops one gets caught.

### Checklist
- [x] Forwards to existing, already-tested morphology functions; no new algorithm
- [x] Backend support inherited from the underlying functions (numpy / cupy / dask+numpy / dask+cupy)
- [x] NaN handling unchanged (delegated)
- [x] Accessor parity tested for DataArray and Dataset
- [x] No materialization or copies introduced
- [x] No benchmark needed (no compute path added)
- [x] No README/docs change needed (accessor methods are not separately documented)
- [x] Matches API conventions (mirrors the four existing morph accessor methods)

PR #3410, opened 2026-06-20
Skip morphology memory guard for lazy dask inputs (#3401)

### LINKED ISSUE #3401: morphology: memory guard raises false MemoryError on lazy dask rasters
## Description

`morph_erode`, `morph_dilate`, `morph_opening`, and `morph_closing` raise a spurious `MemoryError` on large lazy dask-backed rasters. The dask backend processes the array chunk by chunk and never materializes a full padded copy, so the guard is rejecting work that would run fine.

## Cause

`_dispatch()` in `xrspatial/morphology.py` calls `_check_kernel_memory(rows, cols, ky, kx, name)` with `rows, cols = agg.shape` (the full logical shape). The guard computes `required = padded_rows * padded_cols * 16` and raises `MemoryError` when that exceeds 50% of host RAM.

That is correct for the eager numpy and cupy backends, which allocate a full padded float64 copy of the input. It is wrong for the dask backends: `map_overlap` allocates only a padded copy of each chunk, so peak memory scales with chunk size, not the full array. The guard's own docstring says it budgets "a padded copy of the input" -- only the eager backends do that.

## Reproduction

```python
import numpy as np
import dask.array as da
import xarray as xr
from xrspatial.morphology import morph_erode

arr = da.zeros((200000, 200000), chunks=(2048, 2048), dtype="float64")
raster = xr.DataArray(arr, dims=["y", "x"])
morph_erode(raster, kernel=np.ones((3, 3), dtype=np.uint8), boundary="reflect")
```

Raises:

```
MemoryError: erode(): kernel shape (3, 3) on a (200000, 200000) raster needs ~640.0 GB of padded float64 memory, but only 45.9 GB is available. Use a smaller kernel.
```

## Fix

Skip the full-shape memory guard when the input is dask-backed. The per-chunk allocation is bounded by chunk size, so the full-array budget does not apply. Keep the guard for the eager numpy and cupy backends.

## Backends affected

dask+numpy and dask+cupy. Eager numpy and cupy behavior is preserved.

Found by /sweep-performance.

brendancol:
Closes #3401

## What changed

- `_dispatch()` in `xrspatial/morphology.py` no longer runs the full-shape memory guard when the input is dask-backed. The guard budgets a full padded float64 copy of the input, which only the eager numpy/cupy backends allocate. Dask processes the raster chunk-by-chunk via `map_overlap`, so peak memory scales with chunk size, not the full shape.
- Before this change, a large lazy dask raster (e.g. 200000x200000) raised a false `MemoryError` and rejected work that would have run fine.
- The guard still runs for the eager numpy and cupy backends, so their behavior is unchanged.

## Backend coverage

- dask+numpy: guard skipped (fixed)
- dask+cupy: guard skipped (fixed)
- numpy: guard preserved
- cupy: guard preserved

## Test plan

- [x] New regression test builds the graph for a 4096x4096 lazy dask raster under a 1 MB memory budget without raising (graph construction only, no `.compute()`).
- [x] Existing eager-backend guard tests still pass (oversized kernel raises, normal use allowed, all public APIs covered).
- [x] Full `test_morphology.py` suite passes (48 tests).


brendancol (commented):
## PR Review: Skip morphology memory guard for lazy dask inputs (#3401)

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/morphology.py:451` -- The skip is keyed on the input being any `dask.Array`, regardless of chunking. A user who passes a single-chunk dask array (`chunks=(-1, -1)`) now bypasses the guard even though that chunk function materializes a full padded copy per block. Unusual chunking choice, and the eager backends still guard, so not worth blocking on. A one-line note in the comment that the skip assumes reasonable chunking would help the next reader.

### What looks good
- The fix matches the cause in the issue: the guard budgets a full padded copy that only the eager backends allocate, so skipping it for dask is correct.
- The condition reuses `has_dask_array()` and `isinstance(agg.data, da.Array)`, the same pattern `ArrayTypeFunctionMapping` uses to detect the dask backends, so it stays consistent with dispatch.
- The eager numpy and cupy paths are untouched, so the existing guard tests still apply.
- The regression test builds the graph only and never calls `.compute()`, and uses a tiny monkeypatched memory budget so it actually exercises the skip rather than passing by luck.

### Checklist
- Algorithm unchanged; only the guard gate changed: ok
- Backends: numpy/cupy guard preserved, dask+numpy/dask+cupy guard skipped: ok
- NaN handling: unchanged
- Edge cases: new dask test added; eager guard tests intact
- Dask chunk boundaries: unchanged (map_overlap untouched)
- No premature materialization: confirmed, graph-only test
- Benchmark: not needed (gate change, not a hot path)
- Docstrings: comment added explaining the skip rationale

brendancol (commented):
## Follow-up review (after b1bafaae)

The one nit from the prior review is resolved: the guard-skip comment now notes that the skip assumes reasonable chunking and that a single giant chunk would still materialize a full padded copy per block. No code-path change, so the test results stand (16 memory/guard/dask tests pass; full file 48 pass).

PR #3415, opened 2026-06-20
Test coverage: morphology edge inputs (Inf, all-NaN, strip, integer)

### LINKED ISSUE #3404: Test coverage gaps in morphology: Inf, all-NaN, strip rasters, integer dtype
## Summary

The morphology module (erode, dilate, opening, closing, gradient, white/black top-hat) already has solid backend and edge-case coverage, but a few edge inputs are never exercised by the test suite. The source handles them correctly today, so these are regression guards rather than bug reports.

## Gaps found by /sweep-test-coverage

1. Infinite inputs (+Inf / -Inf). No test passes Inf to any morphology function. Inf interacts with the min/max kernel and with the NaN-propagation branch (the `v != v` check), so a future refactor of that branch could silently break Inf handling and no test would catch it.
2. All-NaN raster. The all-NaN case is not tested. Every cell sees only NaN neighbours and should return NaN.
3. Degenerate strip rasters (Nx1 and 1xN). The single-pixel (1x1) case is covered, but column and row strips are not. These exercise the dask map_overlap path where one dimension is narrower than the overlap depth.
4. Integer-dtype input. The numpy path casts to float64 via astype; no test confirms integer input is accepted and promoted.

## Proposed fix

Add tests only, no source changes. Cover the four gaps above on numpy and, where applicable, dask / cupy / dask+cupy using the cross-backend helpers in general_checks.py.

brendancol:
Closes #3404

Adds test-only coverage for four morphology edge inputs that the source already handles correctly but no test exercised. No source changes.

- Inf / -Inf inputs: confirm they flow through the min/max kernel and the NaN-propagation branch without turning into NaN
- All-NaN raster: confirm every op returns all NaN
- Nx1 / 1xN strip rasters: confirm shape is preserved on the dask map_overlap path where one axis is narrower than the overlap depth
- Integer-dtype input: confirm it is accepted and promoted to float64

Backends: numpy plus dask, cupy, and dask+cupy parity where applicable, using the cross-backend helpers in general_checks.py. The cupy and dask+cupy variants ran on a GPU host before this PR.

Test plan
- [x] new tests pass on numpy, dask, cupy, dask+cupy (GPU host)
- [x] full test_morphology.py and test_morphology_derived.py pass (86 tests)


brendancol (commented):
## PR Review: morphology edge-input coverage (Inf, all-NaN, strip, integer)

Test-only PR. No source files changed, so the correctness / backend / performance / docs categories are mostly N/A. I reviewed the four new test groups in test_morphology.py.

### Blockers
None.

### Suggestions
None blocking.

### Nits
- The all-NaN, strip, and integer-dtype tests cover numpy and dask but not cupy. That is fine here: the GPU min/max kernel logic is identical and is already cross-checked by the broad erode/dilate cupy parity tests plus the new Inf cupy parity test. Adding cupy variants would be belt-and-suspenders, not a real gap.

### What looks good
- Inf tests assert concrete min/max values and finite-vs-inf positions, not just "runs without error".
- Inf parity covered on numpy, dask, and cupy.
- Strip and integer tests are parametrized over both orientations and three integer dtypes.
- all-NaN uses boundary='nearest', so the assertion isolates the all-NaN input rather than edge padding.
- New tests pass alongside the existing 86-test suite.

### Checklist
- Source unchanged, no algorithm to verify
- Backends consistent: parity asserted for Inf (numpy/dask/cupy), all-NaN and strip (numpy/dask)
- NaN handling correct
- Edge cases covered (the point of the PR)
- Dask boundaries: strip tests hit the degenerate-axis map_overlap path
- No premature materialization: .compute() only in assertions
- Benchmark not needed, test-only
- README matrix not needed, no new function
- Docstrings present on each test

PR #3418, opened 2026-06-20
classify: align natural_breaks parameter order with sibling classifiers

### LINKED ISSUE #3398: classify: natural_breaks parameter order drifts from quantile/maximum_breaks
## Summary

`natural_breaks` orders its parameters differently from the other
classifiers that take the same arguments. Lining it up makes the
classify family predictable, but the reorder is a breaking change for
anyone passing `num_sample` positionally, so it needs a deprecation
shim rather than a hard swap.

## The drift

The three classifiers that accept `k`, `num_sample`, and `name`:

```
quantile(agg, k=4, num_sample=20000, name='quantile')
maximum_breaks(agg, k=5, num_sample=20000, name='maximum_breaks')
natural_breaks(agg, num_sample=20000, name='natural_breaks', k=5)   # odd one out
```

`quantile` and `maximum_breaks` use `(agg, k, num_sample, name)`.
`natural_breaks` puts `num_sample` second and `k` last. A user who
copies the positional style from one classifier to another gets a
surprise: `natural_breaks(raster, 5)` sets `num_sample=5`, not `k=5`.

## Proposed fix

Reorder `natural_breaks` to `(agg, k=5, num_sample=20000, name=...)` to
match its siblings, with a backward-compatible shim:

- New positional second argument is `k`.
- Legacy callers always passed `k` as a keyword (it was the last
  parameter), so when `k=` is given as a keyword AND a second
  positional is present, treat that positional as the old `num_sample`
  and emit a `DeprecationWarning`.

This keeps `natural_breaks(raster, 20000, k=4)` working (one example
notebook uses exactly that) while warning the caller to switch to
`num_sample=`.

## Also in scope (trivial)

`binary` is the only public classifier with no type hints on its
signature. Add `agg: xr.DataArray`, `name: Optional[str]` to match the
rest of the module.

## Deprecation impact

The parameter reorder is breaking for positional `num_sample` usage.
The shim covers the one documented pattern (`k` keyword + positional
`num_sample`) and warns rather than silently reinterpreting.

Found by the api-consistency deep-sweep against `classify`.

brendancol:
## What

`natural_breaks` ordered its parameters differently from the other
classifiers that take the same arguments. This lines it up and adds the
type hints `binary` was missing.

Closes #3398.

## The drift

```
quantile(agg, k=4, num_sample=20000, name='quantile')
maximum_breaks(agg, k=5, num_sample=20000, name='maximum_breaks')
natural_breaks(agg, num_sample=20000, name='natural_breaks', k=5)   # before
```

`quantile` and `maximum_breaks` put `k` first and `name` last.
`natural_breaks` did the opposite, so `natural_breaks(raster, 5)` set
`num_sample=5` instead of `k=5`.

## Change

- Reorder `natural_breaks` to `(agg, k=5, num_sample=20000, name=...)`.
- Add a shim for the old order. Pre-1.0 callers always passed `k` as a
  keyword (it was the last parameter), so a call that gives `k=` as a
  keyword and a second positional is using the old order: that
  positional is the old `num_sample`. Remap it and emit a
  `DeprecationWarning`. `natural_breaks(raster, 20000, k=4)` (used in
  one example notebook) keeps working.
- Add `agg: xr.DataArray`, `name: Optional[str]`, and the return hint to
  `binary`, the only public classifier without type hints.

## Deprecation impact

The reorder is breaking for code that passed `num_sample` positionally.
The shim covers the one documented pattern (`k` keyword plus positional
`num_sample`) and warns rather than silently reinterpreting the value.

## Tests

- `test_natural_breaks_positional_k_matches_siblings`: second positional
  arg is now `k`.
- `test_natural_breaks_legacy_positional_num_sample_warns`: the old
  order warns and produces the same result.

Verified on numpy and cupy entry points (CUDA available on this host),
plus the Dataset path. Full `test_classify.py` and `test_validation.py`
pass.

Found by the api-consistency deep-sweep against `classify`.


brendancol:

Issue #3406, opened 2026-06-20
pycnophylactic raises opaque ValueError on all-NaN zones or no matching zone id


brendancol:
## Description

`pycnophylactic` raises an opaque `ValueError` when no pixel in the raster is valid for smoothing. This happens in two reachable cases:

1. The `zones` raster is entirely NaN.
2. None of the zone ids in the raster appear in the `values` mapping.

In both cases no zone mask is built, so the `valid` mask is all-False and `surface[valid]` is empty. The convergence check then calls `np.nanmax` on a zero-size array:

```
File "xrspatial/dasymetric.py", line 630, in _pycnophylactic_numpy
    max_change = np.nanmax(np.abs(smoothed[valid] - surface[valid]))
ValueError: zero-size array to reduction operation fmax which has no identity
```

`disaggregate` handles the same inputs gracefully and returns an all-NaN raster. The two public functions should agree.

## Reproduce

```python
import numpy as np, xarray as xr
from xrspatial.dasymetric import pycnophylactic

# case 1: all-NaN zones
z = xr.DataArray(np.full((3, 3), np.nan), dims=['y', 'x'])
pycnophylactic(z, {1: 100.0})   # raises ValueError

# case 2: no matching zone id
z = xr.DataArray(np.array([[1, 1], [2, 2]], dtype=np.float64), dims=['y', 'x'])
pycnophylactic(z, {99: 100.0})  # raises ValueError
```

## Expected

Return an all-NaN raster (matching `disaggregate`), or raise a clear, documented error. The empty-array `ValueError` from numpy is neither.

## Suggested fix

Guard the convergence check for the empty-valid case in `_pycnophylactic_numpy`: if no pixel is valid, skip the smoothing loop and return the all-NaN surface immediately.

Found by the test-coverage sweep on `dasymetric`. A test pinning this (xfail-strict until fixed) is added in the linked test PR.

PR #3419, opened 2026-06-20
Add dasymetric test coverage: metadata, 1x1, Inf weight, 3-class limiting_variable

### LINKED ISSUE #3407: Test coverage gaps in dasymetric: metadata preservation, 1x1 raster, Inf weight, 3-class limiting_variable
The test-coverage sweep audited `xrspatial/dasymetric.py`. The module is well covered overall (cross-backend equivalence, conservation, NaN/nodata, memory guards). A few real gaps remain:

- **Metadata preservation not asserted.** `test_output_metadata` checks dims, name, and shape but never asserts that `attrs` (`res`, `crs`) and coords carry through from the input `zones` for either `disaggregate` or `pycnophylactic`. This is the smoke detector the metadata sweep relies on.
- **True 1x1 raster untested.** The closest case is a 1x2 strip. A 1x1 raster is the degenerate kernel case for `pycnophylactic` (both `nrows > 1` and `ncols > 1` guards are skipped, so no neighbour shifts run).
- **Inf weight untested.** A `+inf` weight collapses a zone total to 0 (silent conservation break). Worth pinning the current behaviour so a future change is caught.
- **3-class `limiting_variable` never exercised.** Only the 2-class `(0.0,)` break is tested, despite the docstring advertising three-class dasymetric. Multi-break + per-class density caps have no coverage.

Separately, the audit surfaced a crash bug in `pycnophylactic` on all-NaN / no-matching-zone input (filed as #3406). The test PR pins it with an xfail-strict test linked to that issue.

The fix here is test-only.

brendancol:
Closes #3407.

Test-only additions to `xrspatial/dasymetric.py` coverage, found by the test-coverage sweep. No source changes.

- `TestMetadataPreservation`: assert `attrs` (res/crs) and coords carry through `disaggregate` and `pycnophylactic` (the old metadata test only checked dims/name/shape). Covers numpy and dask.
- `TestSinglePixel`: true 1x1 raster for both functions (the existing case was a 1x2 strip). 1x1 is the degenerate kernel case for `pycnophylactic` where no neighbour shift runs.
- `TestInfWeight`: pin the current behaviour for a `+inf` weight, which collapses a zone total to 0 (silent conservation break).
- `TestLimitingVariableThreeClass`: three-class `limiting_variable` with per-class caps. Only the 2-class break was tested before, despite the docstring describing three-class dasymetric.
- `TestPycnophylacticEmptyValid`: xfail(strict) tests pinning a crash bug the sweep found, filed as #3406. `pycnophylactic` raises `ValueError` (np.nanmax on a zero-size array) when no pixel is valid for smoothing (all-NaN zones, or no zone id present in `values`); `disaggregate` returns all-NaN on the same input. The xfail flips red once #3406 is fixed.

Backends: numpy and dask+numpy run locally. cupy / dask+cupy backends for `disaggregate` are already covered by the existing `TestCrossBackend` equivalence tests, so no new GPU variants were added.

Test plan:
- [x] `pytest xrspatial/tests/test_dasymetric.py` on a CUDA host: 72 passed, 2 xfailed
- [x] New classes run: 11 passed, 2 xfailed
- [x] xfail(strict) tests confirmed to fail with the expected ValueError (the #3406 bug reproduces)


brendancol (commented):
## PR Review: Add dasymetric test coverage (metadata, 1x1, Inf weight, 3-class limiting_variable)

Test-only PR. No source files changed, so the correctness risk is low and the review focuses on whether the tests pin the right behaviour.

None.

- The two xfail(strict) tests in `TestPycnophylacticEmptyValid` depend on #3406 staying open. Once #3406 is fixed the strict marker flips them red (XPASS), which is the intended signal to remove the marker. The test PR is already cross-linked from the #3406 body, so whoever fixes the bug will know to drop the markers. Informational.

- `test_inf_weight_collapses_zone` emits a `RuntimeWarning: invalid value encountered in divide` from `dasymetric.py:206`. The module already raises this warning in normal operation and the existing tests don't suppress it, so leaving it matches the file's style. Wrapping the call in `np.errstate(invalid='ignore')` would give a clean pytest run, but it's not required.
- `TestMetadataPreservation` and `TestSinglePixel` cover numpy and dask but not cupy. The cupy dispatch is already exercised by `TestCrossBackend`, so this is fine; a cupy metadata assertion would be the only marginal gain and isn't worth a GPU-only test on its own.

### What looks good
- Reference values are concrete, not just "runs without crashing": 1x1 returns the full zone value, the 3-class case checks the 0/5/145 split and conservation, and the Inf case pins the exact NaN/0 collapse.
- The discovered crash bug was filed separately (#3406) and pinned with xfail(strict, raises=ValueError) rather than bundled into a source fix. That keeps the test PR honest and gives the fix a built-in regression guard.
- 1x1 dask-vs-numpy parity test covers the degenerate chunk case.

### Checklist
- All implemented backends produce consistent results: existing cross-backend tests cover cupy/dask+cupy; new tests add numpy/dask metadata + 1x1 parity
- NaN handling is correct: all-NaN disaggregate asserted all-NaN; pycnophylactic crash pinned via xfail
- Edge cases covered: 1x1, Inf weight, 3-class caps, empty-valid
- Dask chunk boundaries handled: 1x1 dask parity test
- No premature materialization: n/a (test-only)
- README feature matrix: not applicable (no new public API)
- Docstrings present and accurate: each new test class has a docstring explaining intent and issue link

PR #3420, opened 2026-06-20
Vectorize disaggregate() zone-membership loop (#3408)

### LINKED ISSUE #3408: Vectorize disaggregate() zone-membership loop (O(n_zones x n_pixels) -> O(n_pixels + n_zones))
## What

`disaggregate()` in `xrspatial/dasymetric.py` distributes zone values across
raster pixels by looping over zone IDs and rebuilding a full-shape boolean
mask for each zone. Three places do this:

- `_disaggregate_numpy` computes per-zone weight sums in a `for zid in
  zone_ids` loop, evaluating `(zones_arr == zid)` over the whole array each
  iteration.
- `_disaggregate_numpy_with_sums` distributes values in a second
  `for zid in values_dict` loop with the same per-zone full-array mask.
- `_chunk_sums` in the dask path repeats the per-zone loop for every chunk.

Peak work is O(n_zones x n_pixels): the array gets scanned once per zone.

## Impact

On a 1500x1500 raster the numpy weighted path takes 0.35s for 50 zones, 1.9s
for 500 zones, and 6.8s for 2000 zones. Runtime grows linearly with the zone
count instead of staying flat. The dask path pays the same per-zone cost per
chunk, and the cupy path inherits it through the CPU fallback.

## Proposed fix

Replace the per-zone loop with a single vectorized pass: `np.searchsorted` to
map each pixel to its value index, `np.add.at` (or `np.bincount`) to
accumulate zone weight sums in one pass, then a gather to distribute. That
brings it to O(n_pixels + n_zones).

Semantics that must stay identical (a proof-of-concept matched the current
output exactly on all of these for both the weighted and binary methods):

- NaN in zones or weight produces NaN output
- `nodata_zone` produces NaN output
- negative weights clamp to 0
- binary method binarizes nonzero weights to 1
- a zone present in the raster but absent from `values` produces NaN
- a zone whose total weight is 0 produces NaN

## Backends

numpy, cupy (CPU fallback), dask+numpy (per-chunk), dask+cupy. OOM verdict
SAFE, bottleneck compute-bound.

Found by /sweep-performance.

brendancol:
Closes #3408

## What changed

`disaggregate()` distributed zone values by looping over zone IDs and
rebuilding a full-shape boolean mask `(zones_arr == zid)` for each zone.
Three call sites did this: the numpy zone-sum computation, the numpy
distribute step, and the dask per-chunk reducer. Peak work scaled as
O(n_zones x n_pixels).

This replaces the per-zone loop with a single vectorized pass:

- `np.searchsorted` maps each valid pixel to its zone slot once
  (`_pixel_zone_index`).
- `np.add.at` accumulates per-zone weight sums in one pass
  (`_zone_weight_sums`).
- distribution gathers each pixel's zone value and weight sum, then divides.

A shared `_preprocess` helper does the float64 coercion, negative-weight
clamp, binary binarization, and invalid-mask construction so the numpy core
and the dask chunks stay identical.

## Performance

numpy weighted path on a 1500x1500 raster:

| n_zones | before | after |
|---------|--------|-------|
| 50      | 0.35s  | 0.18s |
| 500     | 1.9s   | 0.19s |
| 2000    | 6.8s   | 0.19s |

Runtime is now flat in zone count instead of linear.

## Backends

numpy, cupy (CPU fallback), dask+numpy (per-chunk), dask+cupy. All four
validated. OOM verdict SAFE, bottleneck compute-bound.

## Test plan

- [x] Existing 61 dasymetric tests pass (cross-backend parity + memory guards)
- [x] New test: many-zone (200) conservation, with missing-zone and
  zero-weight-sum zones staying NaN
- [x] New test: many-zone (150) dask matches numpy per-chunk
- [x] cupy and dask+cupy paths run end-to-end on a CUDA host and match numpy

Found by /sweep-performance.


brendancol (commented):
## PR Review: Vectorize disaggregate() zone-membership loop (#3408)

None.

None.

- `xrspatial/dasymetric.py`: in the pure-numpy path `_disaggregate_numpy` builds `zone_index` once for the sums, then `_disaggregate_numpy_with_sums` recomputes the same `zone_index` from scratch. That is two searchsorted passes where one would do. The dask path genuinely needs two passes (eager sums, then lazy distribute), so the split signature is right there, but the numpy path could thread the index through. Minor; the win over the old per-zone loop is already large.

### What looks good
- The vectorization holds across the edge cases I checked: NaN zones/weights, nodata_zone, negative-weight clamp, binary binarization, missing-zone -> NaN, and zero-weight-sum -> NaN. I confirmed numpy/dask parity on a 150-zone raster and ran cupy + dask+cupy end-to-end on a CUDA host.
- `searchsorted` needs sorted ids and the code passes `np.array(sorted(...))`, so the lookup is valid. Out-of-range positions get clipped and then rejected by the equality check, so a zone id absent from values cannot alias onto a real slot.
- `_preprocess` pulls together the clamp/binarize/invalid-mask logic that used to be duplicated across the numpy core, `_chunk_sums`, and `_distribute_chunk`, so the four backends share one definition.
- Empty values dict returns all-NaN; a single-pixel zone gets the full value. Both checked.
- The existing 61 tests stay green, and the 2 new tests cover the many-zone case the fix targets.

### Checklist
- [x] Algorithm matches the documented redistribution semantics
- [x] NaN handling is correct
- [x] Dask chunk boundaries handled correctly (no neighborhood access; per-chunk distribute is shape-preserving)
- [x] No premature materialization or unnecessary copies introduced (the eager zone-sum compute is pre-existing and algorithmically required)
- [x] Benchmark exists (benchmarks/benchmarks/dasymetric.py)

PR #3426, opened 2026-06-20
Add viewshed tests for 1x1, all-NaN, and Inf-terrain edge cases

### LINKED ISSUE #3425: viewshed: add test coverage for 1x1, all-NaN, and Inf-terrain edge cases
The test-coverage sweep went through `xrspatial/viewshed.py` and found three input shapes that nothing in `test_viewshed.py` touches. I checked all three by hand and each one already returns the right answer on numpy and dask Tier B, so this is a coverage gap rather than a bug. Adding the tests just pins the current behavior so a later refactor can't quietly break it.

The three gaps:

1x1 single-pixel raster. One cell, degenerate grid. The CPU sweep already guards against a zero-size resolution axis (#2744), but no test checks that 1x1 returns the observer angle of 180 instead of dividing by zero. Single-pixel and Nx1 rasters are a classic spot for off-by-one kernel bugs.

Inf as an actual terrain value. The tests only ever pass `inf` as a parameter (`observer_elev`, `target_elev`, `max_distance`), never as a cell in the raster. An infinitely tall cell currently comes back visible (180) and doesn't leak NaN into neighbors, but nothing pins that down.

all-NaN raster. Individual NaN cells are tested thoroughly; a fully-NaN raster isn't tested at all. Today the observer cell reads 180 and the rest is INVISIBLE, which is what you'd want, but no test asserts it.

Tests I'm adding to `xrspatial/tests/test_viewshed.py`, on numpy and dask Tier B (the CPU-backed backends; cupy/RTX is a separate implementation):
- `test_viewshed_single_pixel`: 1x1 returns `[[180.0]]` on numpy and dask
- `test_viewshed_inf_terrain_cell`: an inf elevation cell doesn't crash or produce NaN output
- `test_viewshed_all_nan_input`: all-NaN raster gives observer=180, rest INVISIBLE

No source change. If a test does surface a real bug, that goes in a separate accuracy issue.

brendancol:
Closes #3425

Test-only change from the test-coverage sweep. Three input shapes that nothing in `test_viewshed.py` exercised, all already behaving correctly:

- `test_viewshed_single_pixel`: a 1x1 raster returns `[[180.0]]` (the degenerate-axis resolution guard from #2744 keeps it from dividing by zero)
- `test_viewshed_inf_terrain_cell`: an `inf` elevation cell comes back visible and doesn't leak NaN into the rest of the grid
- `test_viewshed_all_nan_input`: an all-NaN raster gives observer=180 and INVISIBLE everywhere else

Each is parametrized over numpy and dask Tier B (the CPU-backed backends). The cupy/RTX path is a separate implementation and is out of scope here.

No source change. I probed all three by hand first to confirm the behavior was already correct, so these pin existing behavior rather than fixing a bug.

Test plan:
- [x] `pytest xrspatial/tests/test_viewshed.py` — 120 passed, 1 skipped (no-RTX fallback)
- [x] the 6 new test cases run and pass on a CUDA+RTX host
- [x] flake8 clean on the test file

### DIFF
--- a/xrspatial/tests/test_viewshed.py
+++ b/xrspatial/tests/test_viewshed.py
@@ -945,6 +945,91 @@ def test_viewshed_nan_input_across_backends(backend):
     assert vals[1, 1] == INVISIBLE
 
 
+@pytest.mark.parametrize("backend", ["numpy", "dask+numpy"])
+def test_viewshed_all_nan_input(backend):
+    """An all-NaN raster is the degenerate boundary of the algorithm.
+
+    Every cell is NODATA, so no sweep events are generated. The observer
+    cell is still marked 180 (the viewpoint is always visible to itself)
+    and every other cell stays at its INVISIBLE fill value. Pins the
+    current behaviour on the CPU-backed backends; the cupy/RTX path is a
+    separate implementation and is not covered here.
+    """
+    arr = np.full((5, 5), np.nan)
+    xs = np.arange(5.0)
+    ys = np.arange(5.0)
+
+    if backend == "numpy":
+        data = arr
+    else:
+        data = da.from_array(arr, chunks=(3, 2))
+    raster = xa.DataArray(data, coords=dict(x=xs, y=ys), dims=["y", "x"])
+
+    v = viewshed(raster, x=2.0, y=2.0, observer_elev=5)
+    vals = v.values
+
+    assert vals[2, 2] == 180.0
+    # Only the observer cell is non-INVISIBLE; everything else is NODATA.
+    others = np.delete(vals.ravel(), 2 * 5 + 2)
+    assert (others == INVISIBLE).all()
+
+
+@pytest.mark.parametrize("backend", ["numpy", "dask+numpy"])
+def test_viewshed_inf_terrain_cell(backend):
+    """An infinite terrain elevation value must not crash or poison output.
+
+    Tests pass inf only as observer_elev / target_elev / max_distance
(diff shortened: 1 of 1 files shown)


brendancol (commented):
## PR Review: viewshed edge-case test coverage

Test-only change, three new tests parametrized over numpy and dask Tier B. I read all three in full and ran the suite. No blockers, no suggestions.

### Blockers
None.

### Suggestions
None.

### Nits
- `test_viewshed_all_nan_input` hardcodes the observer's flat index as `2 * 5 + 2`. It's tied to the fixed 5x5 grid and the inline comment explains it, so this is fine as-is; just noting it for anyone who later changes the grid size.

### What looks good
- Each test checks against known values (observer cell = 180, INVISIBLE elsewhere), not just "runs without crashing".
- NaN checks use `np.isnan`, and the inf test explicitly asserts no NaN leaks into the rest of the grid.
- Backend scope (numpy + dask Tier B, cupy/RTX out) matches the existing NaN tests in this file and is called out in each docstring.
- 1x1 dask uses a `(1,1)` chunk, which exercises the degenerate-axis resolution guard on the dask path too.

### Checklist
- [x] Algorithm matches reference/paper (n/a — test-only)
- [x] All implemented backends produce consistent results (numpy + dask asserted; cupy/RTX scoped out)
- [x] NaN handling is correct
- [x] Edge cases are covered by tests (that's the PR)
- [x] Benchmark exists or is not needed (not needed — tests)

PR #3434, opened 2026-06-20
Propagate input dims/coords through true_color() (#3429)

### LINKED ISSUE #3429: true_color() hardcodes y/x dims and drops extra coords
**Describe the bug**

`true_color()` in `xrspatial/multispectral.py` builds its output metadata by hand instead of propagating it from the input. Every other index in the module passes `dims=in.dims, coords=in.coords, attrs=in.attrs` straight through. `true_color()` does this at the end:

```python
_dims = ['y', 'x', 'band']
_attrs = r.attrs
_coords = {'y': r['y'], 'x': r['x'], 'band': [0, 1, 2, 3]}
```

Two things break as a result:

1. The dim names and coord lookups are pinned to `y`/`x`. Feed it a raster with any other dim names (`lat`/`lon`, `row`/`col`) and `r['y']` throws `KeyError: 'y'`. The `ebbi` docstring in this same module uses `lat`/`lon` inputs, so that naming is supported elsewhere.

2. Non-spatial coords on the input get dropped. A `spatial_ref` scalar coord (which rioxarray attaches) disappears, because `_coords` is rebuilt from scratch with only `y`, `x`, and `band`.

The line above already carries a `# TODO: output metadata: coords, dims, attrs` comment, so the gap is known.

**Expected behavior**

`true_color()` should take the two spatial dims and their coords from the input DataArray, then append the `band` dim. Attrs and any extra coords should survive the round trip, same as `ndvi`/`evi`.

**Reproduction**

```python
from xrspatial.multispectral import true_color

d = np.random.default_rng(0).random((4, 4)).astype('f4')
r = xr.DataArray(d, dims=['lat', 'lon'])
r['lat'] = np.linspace(3, 0, 4)
r['lon'] = np.linspace(0, 3, 4)
true_color(r, r.copy(), r.copy())   # KeyError: 'y'
```

Found by /sweep-metadata (Cat 2 coords, Cat 3 dims).

brendancol:
Closes #3429.

`true_color()` hardcoded its output dims to `['y','x','band']` and rebuilt coords via `r['y']`/`r['x']`. That raised `KeyError: 'y'` on rasters with `lat`/`lon` (or `row`/`col`) dims and silently dropped non-spatial coords like `spatial_ref`. The other indices in the module already pass `dims`/`coords`/`attrs` straight through; this brings `true_color()` in line.

- Derive the two spatial dims and their coords from the input DataArray, then append the `band` dim.
- Pass input attrs and any extra coords through unchanged.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy (all verified locally on a GPU host).

Test plan:
- [x] New `test_true_color_preserves_non_yx_dims` runs on all four backends with `lat`/`lon` input.
- [x] New `test_true_color_preserves_extra_coords` checks a `spatial_ref` coord survives.
- [x] Full `test_multispectral.py` passes (159 tests).


brendancol (commented):
## PR Review: Propagate input dims/coords through true_color() (#3429)

None.

None.

- multispectral.py:1828 — if an input raster carried a coord literally named `band`, the `'band': [0, 1, 2, 3]` entry would overwrite it. Far-fetched (a 2D raster with a non-dim coord called `band`), and the old code had the same collision, so it's not a regression. Not worth guarding against unless it ever comes up.

### What looks good
- The fix mirrors how every other index in the module preserves metadata, so true_color is no longer the odd one out.
- Metadata is built in the public function after dispatch, so all four backends share one code path. The parametrized test runs numpy, cupy, dask+numpy, and dask+cupy with lat/lon input.
- `[*r.dims, 'band']` lines up with the `(H, W, 4)` output cube: `r` is validated 2D and the stack axis is appended last on every backend.
- Tests assert on dims, coord values, and attrs, not just that the call returns. The spatial_ref test covers the dropped-coord half of the bug.

### Checklist
- [x] Algorithm matches reference/paper (N/A, metadata-only change, pixel math untouched)
- [x] Edge cases are covered by tests (lat/lon dims, extra coord)
- [x] Dask chunk boundaries handled correctly (metadata built post-dispatch, no chunk interaction)
- [x] Benchmark exists or is not needed (not needed, no perf change)
- [x] README feature matrix updated (N/A, no new function or backend change)

PR #3436, opened 2026-06-20
Add multispectral test coverage for true_color edges and evi/savi validation

### LINKED ISSUE #3431: test coverage: untested branches in multispectral.py (true_color NaN/all-equal/params, evi & savi validation)
## Summary

The test-coverage sweep found several untested branches in `xrspatial/multispectral.py`. The index functions have solid backend and NaN coverage, but a few specific code paths are never exercised by the suite.

## Untested paths

1. `true_color` NaN handling. The function builds an alpha channel from `np.logical_or(np.isnan(r), r <= nodata)`. The existing tests feed it `random_data` (integers in [-100, 100], no NaN), so the NaN branch of the alpha mask is never hit on any of the four backends.

2. `true_color` all-equal input. When the input has zero range, `_normalize_data_cpu` guards a divide-by-zero with `if range_val != 0`. That guard's false branch is never tested.

3. `true_color` non-default `nodata` / `c` / `th`. All tests call `true_color` with default parameters only.

4. `evi` validation guards. `evi` raises `ValueError` for non-numeric `c1`/`c2`, `soil_factor` outside [-1, 1], and `gain < 0`. None of these error paths have a test.

5. `savi` `soil_factor` out-of-range. `savi` raises `ValueError` when `soil_factor` is outside [-1, 1]. Not tested.

## Proposed fix

Add test-only coverage for the above. No source changes. CUDA is available on the sweep host, so the `true_color` NaN tests run against cupy and dask+cupy as well as numpy and dask+numpy.

Found by /sweep-test-coverage.

brendancol:
Closes #3431

Test-only additions found by /sweep-test-coverage. No source changes.

- `true_color` NaN / nodata alpha mask: both legs of `isnan(r) OR r <= nodata` now checked on numpy, dask+numpy, cupy, and dask+cupy.
- `true_color` all-equal input: exercises the `range_val != 0` divide-by-zero guard in `_normalize_data_cpu`.
- `true_color` non-default `nodata` / `c` / `th`: confirms the parameters change the output.
- `evi` validation guards: non-numeric `c1`/`c2`, `soil_factor` outside [-1, 1], `gain < 0`.
- `savi` `soil_factor` out-of-range.

Backend coverage: the `true_color` NaN/alpha test runs on numpy, dask+numpy, cupy, and dask+cupy. The remaining new tests are backend-agnostic (validation guards) or CPU + dask (true_color edges).

Test plan
- [x] `pytest xrspatial/tests/test_multispectral.py` for the new tests passes locally
- [x] cupy and dask+cupy variants executed locally (CUDA available), not skipped


brendancol (commented):
## PR Review: Add multispectral test coverage for true_color edges and evi/savi validation

Test-only PR. No source changes. The new tests target branches the sweep flagged as never exercised, and I confirmed each one is real by running the paths locally.

### Blockers
None.

### Suggestions
None.

### Nits
- `test_true_color_all_equal_input` on the `dask+numpy` backend emits a `RuntimeWarning: invalid value encountered in cast` at compute time. The eager `true_color` path silences this with `warnings.catch_warnings()`, but the dask path casts lazily so the warning escapes to the test. It is benign (NaN to uint8 is exactly the behavior being asserted), but if you want a clean test log you could filter it with `pytest.warns` or `warnings.simplefilter('ignore')`. Not required.

### What looks good
- The NaN/alpha test drives both legs of `isnan(r) OR r <= nodata` with one fixture (a NaN cell and a `1.0 <= nodata` cell) and runs on all four backends, so the cupy and dask+cupy alpha paths are covered too.
- The all-equal test pins down the `range_val != 0` false branch, which was the one divide-by-zero guard with no coverage.
- The evi/savi validation tests match the exact error strings in the source, so a reworded guard will surface here rather than silently passing.
- Parametrizing `soil_factor` at both 1.5 and -1.5 covers both sides of the range check.

### Checklist
- All implemented backends consistent: true_color NaN test covers all four; validation tests are backend-agnostic
- NaN handling correct: yes, that is the point of the alpha test
- Edge cases covered: all-equal/zero-range and NaN both added
- Dask chunk boundaries: true_color is per-pixel, no neighborhood
- No premature materialization: compute only at assert time in tests
- README matrix: not needed (no new function)
- Docstrings present: n/a

brendancol (commented):
## Follow-up review

Re-reviewed after commit fae6e46.

- Nit (dask `invalid value encountered in cast` RuntimeWarning): **fixed.** Moved the filter into the shared `_true_color_to_numpy` helper so every true_color test silences the expected NaN->uint8 cast at compute time. Verified with `pytest -W error::RuntimeWarning` -- the new tests pass with RuntimeWarning promoted to an error.

No new findings. Full `test_multispectral.py` passes (169 tests, GPU variants included on this CUDA host).

PR #3443, opened 2026-06-22
Fix EUCLIDEAN proximity/allocation/direction dropping targets at exact max_distance

### LINKED ISSUE #3442: EUCLIDEAN proximity/allocation/direction: KDTree backends return NaN at exact max_distance (incl. max_distance=0)
## Description

For `proximity`, `allocation`, and `direction` with `distance_metric='EUCLIDEAN'`, the numpy and dask+numpy backends return NaN at a target pixel when `max_distance=0`, while cupy and dask+cupy return the correct value (0.0 for proximity, the target value for allocation, 0.0 for direction). The same divergence shows up whenever `max_distance` is set exactly to an achievable distance.

A target pixel is distance 0 from itself, so with `max_distance=0` it should qualify. The brute-force backends (cupy, dask+cupy) get this right. The cKDTree-based numpy/dask+numpy paths drop it to NaN.

## Root cause

The numpy/dask KDTree path converts the inclusive `max_distance` into cKDTree's exclusive `distance_upper_bound`. In `_process_numpy_kdtree` it widens the bound by one ulp: `upper = np.nextafter(max_distance, np.inf)`. For Euclidean queries (p=2) cKDTree compares squared distances internally, and `nextafter(0.0, inf)` is the smallest subnormal (5e-324); squaring it underflows to 0.0, so the exclusive bound collapses back to 0 and the distance-0 target is excluded. Manhattan (p=1) does not square, so it is unaffected.

Separately, `_kdtree_query_lowest_index` (used by the dask global and tiled KDTree paths, and by allocation/direction on dask) passes the raw `max_distance` as `distance_upper_bound` with no widening at all, so it stays exclusive and drops pixels sitting exactly at `max_distance`.

## Reproduce

```python

data = np.zeros((3, 3)); data[1, 1] = 1.0
r['y'] = np.arange(3)[::-1].astype(float)
r['x'] = np.arange(3).astype(float)

print(proximity(r, max_distance=0.0).values[1, 1])   # numpy -> nan, should be 0.0
```

cupy returns 0.0 at the target for the same input.

## Suggested fix

Route every cKDTree call through one inclusive-to-exclusive bound helper that holds up for both p=1 and p=2, including `max_distance=0`.

Found by /sweep-accuracy.

brendancol:
Closes #3442

## What changed

- For `distance_metric='EUCLIDEAN'`, the numpy and dask+numpy cKDTree paths dropped a target sitting exactly at `max_distance` to NaN, while cupy and dask+cupy (brute force) kept it. The clearest case is `max_distance=0`: a target pixel is distance 0 from itself and should qualify, but numpy/dask returned NaN there.
- Root cause: cKDTree's `distance_upper_bound` is exclusive, and for p=2 it compares squared distances. The old one-ulp widen of 0 underflows when squared, collapsing the bound back to `max_distance`. `_kdtree_query_lowest_index` did not widen at all.
- Routed every cKDTree call through one `_inclusive_upper_bound` helper that widens by a relative epsilon with an absolute floor, so the bound survives squaring for p=2 and stays inclusive for p=1.

## Backend coverage

numpy and dask+numpy were the affected backends and are fixed. cupy and dask+cupy already behaved correctly (brute force); the new tests pin all four together.

## Test plan

- [x] `max_distance=0` keeps the target pixel (proximity 0, allocation target value, direction 0) on numpy, dask+numpy, cupy, dask+cupy, for EUCLIDEAN and MANHATTAN
- [x] exact-distance boundary pixel qualifies on numpy and dask+numpy
- [x] numpy matches dask+numpy at `max_distance=0` for all three functions
- [x] full `test_proximity.py` suite passes (543 tests)


brendancol (commented):
## PR Review: Fix EUCLIDEAN proximity/allocation/direction dropping targets at exact max_distance

### Blockers
None.

### Suggestions
None blocking. One thing worth a sentence in the helper docstring: the widen is relative (`4*eps*max(|md|, 1.0)`), so at large `max_distance` it could in principle include a target whose true distance is within `~8*eps*md` of the bound. That gap is below float64 resolution at that magnitude, so the two distances are not distinguishable anyway and the behavior is correct. Calling that out would save the next reader the same check I just did.

### Nits
- The `np.isfinite ... else np.inf` guard in `_tiled_chunk_query` (proximity.py:1146) is now redundant since `_inclusive_upper_bound` handles inf, but it is pre-existing and harmless. Leave it.

### What looks good
- The root cause is correct: cKDTree squares `distance_upper_bound` for p=2, and `nextafter(0, inf)**2` underflows to 0, so the old widen collapsed at `max_distance=0`. Verified directly against scipy.
- One shared helper now feeds all three cKDTree call sites (`_process_numpy_kdtree`, both queries in `_kdtree_query_lowest_index`), so the numpy and dask paths can't drift apart again.
- Checked over-inclusion at md=0, md=1, and md=1000: includes exactly-at-bound, excludes just-below, never pulls in the next real distance.
- The tie-break in `_kdtree_query_lowest_index` is preserved: both k=2 neighbors use the same widened bound, so the equidistant test is unaffected.
- Tests cover all four backends for proximity/allocation/direction at max_distance=0 plus an exact-distance boundary pixel, and pin numpy against dask+numpy.

### Checklist
- [x] Algorithm matches reference (cKDTree exclusive-bound semantics verified against scipy)
- [x] All implemented backends produce consistent results (numpy/dask/cupy/dask+cupy)
- [x] NaN handling is correct (inf/unbounded returns inf; non-target stays NaN)
- [x] Edge cases covered (max_distance=0, exact-distance boundary)
- [x] Dask chunk boundaries handled (bounded map_overlap + global/tiled KDTree paths all routed through the helper)
- [x] Benchmark not needed (numerical bug fix, no perf-critical change)
- [x] README not applicable (no new function, no backend change)

brendancol (commented):
Follow-up review after addressing the prior pass.

Disposition of the earlier findings:
- Suggestion (document the relative-widen behavior in the helper docstring): fixed in 04c6c73f. Added a sentence noting that a target within ~8*eps*max_distance of the bound could qualify, but the two distances are indistinguishable in float64 at that magnitude, so the result is still correct.
- Nit (redundant `np.isfinite ... else np.inf` guard in `_tiled_chunk_query`): dismissed. It is pre-existing, harmless, and outside the scope of this fix; touching it would be an unrelated change.

No new findings. The follow-up was a docstring-only edit; the 25 regression cases still pass and flake8 is clean on the changed module.

PR #3444, opened 2026-06-22
test(fire): cover dask+cupy dispatch and metadata preservation

brendancol:
## What

Every function in the `fire` module registers a `dask+cupy` backend in its `ArrayTypeFunctionMapping`, but only `dnbr` had a test that actually ran that path. The other six (`rdnbr`, `burn_severity_class`, `fireline_intensity`, `flame_length`, `rate_of_spread`, `kbdi`) exercised their dask+cupy dispatch with zero coverage. Only `dnbr` also asserted that input `attrs`, `coords`, and `dims` survive the call.

This adds, for each of the six:
- `test_numpy_equals_dask_cupy`: numpy/dask+cupy parity, matching the existing `dnbr` pattern
- `test_output_preserves_metadata`: via `general_output_checks`

## Verification

All 73 `test_fire.py` tests pass on a CUDA host (cupy 13.6, numba CUDA available), including the 35 backend/metadata/dtype tests, none skipped. The branch is merged up to current `main`, so this composes with the recently merged #3396 (dask+numpy dtype fix) and #3397 (fire.py style cleanup).

Tests only, no source change. Found by the test-coverage sweep.

## Gap found, not fixed (out of scope)

Inf-input coverage is still missing across the module. Low severity: these are pure per-cell math ops, so an Inf flows through the arithmetic rather than hitting a kernel-boundary path. Left for a follow-up rather than bundled here.


brendancol:
Sweep audit summary (test-coverage, module `fire`):

- HIGH (Cat 1, backend coverage): `dask+cupy` dispatch registered for all 7 functions but invoked by a test only for `dnbr`. Added `test_numpy_equals_dask_cupy` for the other 6.
- MEDIUM (Cat 5, metadata preservation): only `dnbr` asserted attrs/coords/dims survive. Added `test_output_preserves_metadata` for the other 6.
- LOW (Cat 2, Inf inputs): untested, left out of scope (pure per-cell math, no kernel-boundary path).
- Not flagged: Cat 3 geometric degeneracies (no neighborhood/kernel, so 1x1/Nx1 add nothing); Cat 4 parameter coverage already strong (all 13 fuel models + invalid-range raises; heat_content/annual_precip covered).

All 73 `test_fire.py` tests green on a CUDA host, 35 backend/metadata/dtype tests run not skipped. No source change.

PR #3447, opened 2026-06-22
Validate matching shapes for precomputed corridor surfaces

### LINKED ISSUE #3445: least_cost_corridor(precomputed=True) silently truncates mismatched-shape surfaces
**Describe the bug**

`least_cost_corridor(..., precomputed=True)` sums two cost-distance surfaces with `corridor = cd_a + cd_b` (in `corridor.py`, `_compute_corridor`) without first checking that the surfaces share a shape and grid. The normal (non-precomputed) path can't hit this, because `cost_distance` rejects mismatched inputs with `raster.shape != friction.shape`. The precomputed path skips `cost_distance` entirely, so nothing guards the addition.

When the two precomputed surfaces have different shapes, xarray aligns them on the intersection of their coordinates instead of raising. You get a silently truncated corridor with wrong values.

**Reproduce**

```python
from xrspatial.corridor import least_cost_corridor

def mk(data):
    h, w = data.shape
    r = xr.DataArray(data.astype(float), dims=["y", "x"], attrs={"res": (1.0, 1.0)})
    r["y"] = np.arange(h, dtype=float)
    r["x"] = np.arange(w, dtype=float)
    return r

cd_a = mk(np.ones((4, 4)))
cd_b = mk(np.ones((3, 3)))
out = least_cost_corridor(mk(np.ones((4, 4))), cd_a, cd_b, precomputed=True)
print(out.shape)   # (3, 3) -- silently dropped a row and column
print(out.data)    # all zeros
```

A related variant with same-size but non-overlapping coordinates produces an empty intersection and raises a confusing internal `numpy.nanmin` error instead of a clear validation message.

**Expected behavior**

Mismatched precomputed surfaces should raise a clear `ValueError` naming the offending parameter and the shapes, the way `cost_distance` already does for mismatched raster/friction. They should never silently produce a truncated, wrong-valued corridor.

**Additional context**

Found during a security sweep of `corridor.py` (dtype/shape confusion, silent wrong results). The fix is to validate matching shapes across the surfaces before computing, reusing `_validate_matching_shape` from `xrspatial.utils`.

brendancol:
Closes #3445

`least_cost_corridor(..., precomputed=True)` summed two cost-distance surfaces (`cd_a + cd_b`) without checking they shared a shape. Mismatched surfaces were silently aligned by xarray on the intersection of their coordinates, producing a truncated, wrong-valued corridor (e.g. 4x4 + 3x3 returned an all-zero 3x3 result). The non-precomputed path was already safe because `cost_distance` rejects mismatched raster/friction shapes.

- Added a `_validate_matching_shape` check across the precomputed surfaces, the same guard `cost_distance` already enforces. Applies to both the two-source and pairwise paths.
- The check is shape-only, matching `cost_distance`'s own behavior (it validates `.shape`, not coordinate values).

Backends: pure xarray arithmetic; the validation runs identically across numpy, cupy, dask+numpy, and dask+cupy.

Test plan:
- [x] New test for mismatched precomputed shapes raising ValueError (two-source)
- [x] New test for mismatched precomputed shapes raising ValueError (pairwise)
- [x] Full corridor test suite passes (34 passed)


brendancol (commented):
## PR Review: Validate matching shapes for precomputed corridor surfaces

None.

- The guard is shape-only, so two same-shape surfaces with non-overlapping coordinates still fall through to the old failure mode: xarray aligns them to an empty intersection and `nanmin` raises an opaque error instead of a clear validation message. This matches `cost_distance`'s own shape-only check, so the scope boundary is reasonable, but a one-line comment or follow-up note saying coordinate mismatch is intentionally out of scope would help the next reader.

- The two new tests run numpy-only. The validation happens before backend dispatch, so parametrizing over backends adds little, but a single dask-backed case would confirm `.shape` access stays lazy (no implicit compute). Optional.

### What looks good
- The fix lands exactly on the gap: the precomputed path bypassed `cost_distance`, which is where the shape guard lived. Reusing `_validate_matching_shape` keeps the error message format consistent with the rest of the codebase.
- One loop covers both the two-source and pairwise paths.
- The comment explains why the check is needed (silent xarray alignment), which is the non-obvious part.
- Tests assert on the real failure mode and use the concrete 4x4 + 3x3 -> all-zero 3x3 example from the issue.

### Checklist
- [x] Algorithm matches reference/paper (N/A, validation only)
- [x] All implemented backends produce consistent results (validation is backend-agnostic)
- [x] Docstrings present and accurate (unchanged)

brendancol (commented):
## Follow-up review

Both items from the first pass are handled in a987a487:

- Suggestion (coordinate-mismatch scope): added an inline comment by the shape check stating that, like `cost_distance`, the guard is shape-only and same-shape surfaces with mismatched coordinates are out of scope.
- Nit (dask coverage): added `test_precomputed_mismatched_shape_dask_raises`, which confirms the check fires on dask-backed surfaces without forcing a compute.

Full corridor suite: 35 passed. No remaining findings.

PR #3449, opened 2026-06-22
Propagate friction geo-attrs through least_cost_corridor (#3446)

### LINKED ISSUE #3446: least_cost_corridor drops friction surface geo-attrs (res/crs/transform/nodatavals)
## Summary

`least_cost_corridor` in `xrspatial/corridor.py` drops the friction surface's geo-metadata (`res`, `crs`, `transform`, `nodatavals`) from its output. A downstream call that needs cell size or CRS, such as slope or clipping, silently sees nothing.

## Where it comes from

The corridor surface is `cd_a + cd_b`, where each `cd_*` is `cost_distance(source, friction)`. `cost_distance` copies its attrs from the *source* raster, not from `friction`. xarray's binary `+` then keeps only the attrs that are present and equal in both operands.

Source rasters are usually boolean marker masks built ad-hoc, so they often carry no geo-attrs at all. When that happens the corridor comes back with empty `attrs`, even though the `friction` surface that defines the grid had `res`, `crs`, and `nodatavals`.

## Reproduction

A 7x7 friction surface with `res`, `crs`, `nodatavals` and two source masks with no attrs. The returned corridor has `attrs == {}` on all four backends (numpy, cupy, dask+numpy, dask+cupy). Coords come through fine; only attrs are lost.

## Secondary issue: unstable result name

`cost_distance` renames each surface to its source's `.name`. Summing two differently-named surfaces gives `.name == None`. So the result name depends on whether the two sources happen to share a name.

## Proposed fix

When `precomputed` is False, re-emit the friction surface's attrs and a stable name on the corridor output, since friction is the raster that defines the grid. For the precomputed path there is no friction to draw from, so leave the current source-derived behavior alone. Check attrs/coords/dims/name on the actual returned object across all four backends.

Found by the metadata propagation sweep (`/sweep-metadata`).

brendancol:
Closes #3446

`least_cost_corridor` builds the corridor as `cd_a + cd_b`, where each cost-distance surface carries the attrs of its *source* raster. Source rasters are usually plain marker masks with no geo-attrs, so xarray's binary sum dropped `res`/`crs`/`transform`/`nodatavals` from the output even when the friction surface that defines the grid had them.

## Changes

- Add `_apply_geo_metadata` to re-emit the friction surface's attrs and name on the corridor output.
- Wire it into the non-precomputed path (single, threshold, unreachable, and pairwise outputs). The precomputed path keeps its source-derived behavior, since there is no friction surface to draw from.
- Only `.attrs` and `.name` are set; data, coords, dims, and dtype are untouched, and dask stays lazy.

## Backends

Verified live on numpy, cupy, dask+numpy, and dask+cupy (CUDA host).

## Test plan

- [x] `test_corridor_inherits_friction_geo_attrs` across all four backends
- [x] `test_corridor_threshold_keeps_geo_attrs` across all four backends
- [x] `test_corridor_unreachable_keeps_geo_attrs`
- [x] `test_pairwise_inherits_friction_geo_attrs`
- [x] `test_precomputed_keeps_source_attrs_not_friction`
- [x] Full corridor suite: 43 passed

Found by the metadata propagation sweep (`/sweep-metadata`).


brendancol (commented):
## PR Review: Propagate friction geo-attrs through least_cost_corridor (#3446)

None.

None.

- In the pairwise branch, `_apply_geo_metadata` calls `.rename(geo_source.name)` on each corridor, but the DataArray then goes into `xr.Dataset(corridors)` keyed by `corridor_i_j`, and the dict key wins as the variable name. So the rename is a no-op for the Dataset variables; only the attrs propagation does anything there. Harmless, just slightly redundant, not worth a change on its own.

### What looks good
- The fix goes at the root cause: the corridor is `cd_a + cd_b`, each surface carries its source raster's attrs, and the binary sum drops anything that differs between the two operands. Re-emitting friction's attrs and name is the right call, since friction is the raster that defines the grid.
- The precomputed path is left alone, which is correct. There is no friction surface to draw from there, and `test_precomputed_keeps_source_attrs_not_friction` pins that behavior.
- `_apply_geo_metadata` calls `result.copy()` before touching attrs, so it does not mutate the input and keeps dask arrays lazy (checked: no compute triggered). `.rename(None)` yields a None name when friction is unnamed.
- Test coverage is thorough: all four backends for the main and threshold paths, plus the unreachable all-NaN and pairwise paths. Full corridor suite passes (43 tests).

### Checklist
- [x] Algorithm matches reference/paper (N/A; metadata-only change, values untouched)
- [x] All implemented backends produce consistent results (numpy/cupy/dask+numpy/dask+cupy verified live)
- [x] NaN handling is correct (unreachable all-NaN path keeps attrs)
- [x] Edge cases are covered by tests (unreachable, pairwise, precomputed)
- [x] Dask chunk boundaries handled correctly (no neighborhood op; copy stays lazy)
- [x] No premature materialization or unnecessary copies (.copy() is a shallow metadata copy)
- [x] Benchmark exists or is not needed (not needed; no perf-sensitive change)
- [x] README feature matrix updated (N/A; no new function or backend change)
- [x] Docstrings present and accurate (helper docstring explains the root cause)

PR #3468, opened 2026-06-23
Document the `name` parameter in perlin() docstring

### LINKED ISSUE #3465: perlin(): docstring omits the `name` parameter present in the signature
## Description

`perlin()` in `xrspatial/perlin.py` takes a `name` parameter:

```python
def perlin(agg: xr.DataArray,
           freq: tuple = (1, 1),
           seed: int = 5,
           name: str = 'perlin') -> xr.DataArray:
```

`name` sets the `.name` attribute of the returned DataArray. The docstring's Parameters section documents `agg`, `freq`, and `seed`, but not `name`.

The sibling generator `generate_terrain()` in `xrspatial/terrain.py` documents the same parameter:

```
name : str, default='terrain'
```

An API-consistency audit of the perlin module flagged the mismatch.

## Expected

Add a `name` entry to the `perlin()` docstring Parameters section, matching the `generate_terrain` style, so the docs match the signature.

## Scope

Docstring only. No signature change, no behavior change, no deprecation. The parameter already works on all four backends.

## Related note (not part of this fix)

`bump()` has no `name` parameter, while `perlin()` and `generate_terrain()` both do. Adding `name` to `bump()` is a separate cross-module change, tracked elsewhere.

brendancol:
Closes #3465

`perlin()` accepts a `name` parameter that sets the output DataArray name, but the docstring's Parameters section never mentioned it. The sibling `generate_terrain()` documents the same parameter, so this aligns the two.

- Add a `name : str, default='perlin'` entry to the `perlin()` docstring, matching the `generate_terrain` style.
- Add tests: `name` default and custom value are honored, and the docstring documents the parameter.

Docstring only. No signature change, no behavior change. Backends unaffected (numpy / cupy / dask+numpy / dask+cupy).

Test plan:
- [x] `pytest xrspatial/tests/test_perlin.py` (14 passed, GPU paths included on a CUDA host)


brendancol (commented):
Domain-aware review (API-consistency sweep, perlin module).

Scope check: the diff is exactly the docstring entry plus two focused tests. No signature, behavior, or backend-dispatch changes. The `name` wording and `default='perlin'` match the sibling `generate_terrain()` convention, which is the consistency target.

Blockers: none.

Suggestions: none. `test_perlin_name_param` covers default and custom values on the numpy path; the `name` plumbing in `perlin()` is backend-agnostic (set on the wrapping DataArray after dispatch), so a numpy-only assertion is sufficient and a per-backend matrix would add no signal.

Nits: none.

Verified: `pytest xrspatial/tests/test_perlin.py` passes 14/14 including the GPU paths on a CUDA host.

PR #3470, opened 2026-06-23
perlin: preserve input x/y coordinates on the output

brendancol:
## Summary

- `perlin()` carried the input's attrs (including `crs`/`res`) but built the result with `dims=agg.dims` and no `coords`. A georeferenced input produced an output that kept its `crs` attribute while dropping the coordinate values, an internally inconsistent DataArray that breaks downstream coord-based ops (clip, interp, reproject).
- Fix: pass `coords=agg.coords` when constructing the result. The output has the same shape as the input, so coords align directly.

## Backend coverage

Verified on numpy, cupy, dask+numpy, and dask+cupy. The result is wrapped in one shared code path, so the fix applies to all four.

## Test plan

- [x] New `test_perlin_preserves_coords` asserts x/y coords and attrs pass through
- [x] `pytest xrspatial/tests/test_perlin.py` -- 13 passed (GPU and dask+GPU tests ran live)

Note: GitHub issues are disabled on the source fork, so this PR has no linked issue. Found by the metadata propagation sweep.


brendancol (commented):
## PR Review: perlin: preserve input x/y coordinates on the output

None.

- The new `test_perlin_preserves_coords` only runs on numpy. The GPU and dask tests (`test_perlin_gpu`, `test_perlin_dask_cpu`, `test_perlin_dask_gpu`) build coord-less rasters, so nothing asserts coord pass-through on cupy/dask. The fix is in one shared wrap path so all backends behave the same, and it was verified manually, but a coord-bearing assertion on at least the dask backend would lock it in. Low priority.

None.

### What looks good
- One-line fix at the source of the bug (xrspatial/perlin.py:354): coords were dropped while attrs were kept, so the output claimed a crs with no coordinates. `coords=agg.coords` matches how attrs are already handled.
- Output shape always equals input shape (perlin writes into a same-shape buffer or rebuilds it), so coords align directly with no off-by-one or half-pixel risk.
- The numpy in-place write touches only `.data`, not `.coords`, so reusing `agg.coords` is safe.
- Regression test uses `general_output_checks` plus explicit coord-value and attrs assertions.

### Checklist
- Algorithm unchanged; only result wrapping touched
- Four backends share the wrap path; coords verified on numpy/cupy/dask/dask+cupy
- NaN handling unchanged
- Coord pass-through covered on numpy; GPU/dask coverage relies on the shared path
- No benchmark needed (no perf change)
- No README change needed (no API change)
- No new public API; docstring unchanged

brendancol (commented):
## Follow-up review

The one suggestion from the prior pass is addressed: `test_perlin_preserves_coords_dask` now asserts coord pass-through on the dask backend with ragged chunks, alongside the numpy test. Full perlin suite passes (14 tests, GPU and dask+GPU ran live).

PR #3473, opened 2026-06-23
Fix perlin dask OOM by dropping redundant dask.persist (#3469)

### LINKED ISSUE #3469: perlin dask backends OOM: dask.persist materializes whole noise array before normalizing
## Description

The dask backends of `perlin` pull the whole noise array into memory before normalizing, so peak memory scales with the full array instead of one chunk.

Both `_perlin_dask_numpy` and `_perlin_dask_cupy` build the lazy noise array, then call `dask.persist(data)` to keep it in memory so the `min`/`ptp` reductions don't recompute it. That persist forces every chunk resident at once. On a 30TB input it tries to hold 30TB in memory and runs out.

The persist isn't needed. Both reductions already go out in a single `dask.compute(da.min(data), da.ptp(data))` call, and within one compute call dask shares the named noise subgraph between them, so each chunk is computed once and freed after both reductions read it. I checked: the `min` and `ptp` graphs reference the same 100 noise chunk keys. Drop the persist, compute the two reductions straight off the lazy array, and peak memory stays bounded by chunk size.

## Backends affected

dask+numpy and dask+cupy. The numpy and cupy single-GPU paths are unaffected.

## Fix

Remove `dask.persist(data)` from both dask paths and reduce directly off the lazy array.

## Steps to reproduce

Build a large dask-backed float32 DataArray with small chunks and call `perlin`; peak memory tracks the full array size, not the chunk size.

brendancol:
Closes #3469

## What changed

- Removed `dask.persist(data)` from `_perlin_dask_numpy` and `_perlin_dask_cupy`. Both paths now reduce straight off the lazy noise array.
- The `min`/`ptp` (and `min`/`max` on GPU) reductions already go out in one `dask.compute` call, which shares the noise subgraph between them, so each chunk is computed once and freed after both reductions read it. The persist was redundant and forced every chunk resident at once, which OOMs on large inputs.

## Backends

numpy and cupy single-GPU paths are untouched. dask+numpy and dask+cupy lose the persist; output is unchanged.

## Test plan

- [x] `xrspatial/tests/test_perlin.py` passes, including the cupy and dask+cupy GPU tests on a CUDA host (13 passed).
- [x] Added `test_perlin_dask_does_not_persist_whole_array`, which monkeypatches `dask.persist` to raise and confirms the dask path no longer calls it while still matching the numpy reference.


brendancol (commented):
## PR Review: Fix perlin dask OOM by dropping redundant dask.persist (#3469)

None.

- The fix swaps memory for compute, and that swap is worth spelling out in the PR so nobody "optimizes" the persist back in later. Without it, the noise subgraph runs twice: once for the `min`/`ptp` reductions in `dask.compute`, then again when the caller materializes the returned normalized array. For an OOM fix that's the right call (the noise is point-wise and deterministic, so the recompute is exact and cheap), but the code comment only covers the memory side. A half-sentence noting the recompute is intentional would save the next reader the trace.

- `test_perlin_dask_does_not_persist_whole_array` only covers the dask+numpy path. The dask+cupy path got the same edit but no persist guard. Folding a `monkeypatch.setattr(dask, "persist", ...)` assertion into `test_perlin_dask_gpu` would pin down both. Not blocking, since the edit is symmetric and the GPU path isn't always available in CI.

### What looks good
- Symmetric across both dask backends; numpy and cupy single-GPU are untouched.
- The premise checks out: the `min` and `ptp` graphs reference the same named noise keys, so one `dask.compute` over both shares the subgraph and the persist really is dead weight.
- Output is unchanged. The noise is deterministic, so recomputing yields identical values, and the regression test confirms the dask result still matches the numpy reference.
- The regression test fails loudly if `dask.persist` comes back, which is the right guard for this class of fix.

### Checklist
- Algorithm matches reference: not affected (normalization unchanged)
- All implemented backends consistent: yes, dask+numpy and dask+cupy tests pass
- NaN handling correct: not affected
- Edge cases covered: existing coverage retained
- Dask chunk boundaries: perlin is point-wise, no overlap
- No premature materialization: improved (persist removed)
- Benchmark exists or not needed: benchmarks/benchmarks/perlin.py exists
- Docstrings: unchanged

brendancol (commented):
## Follow-up review (after 67fa8d24)

- Suggestion (recompute not documented): the comment in both dask paths now states the returned lazy array recomputes the noise on materialization and that the extra pass is intentional, since the noise is point-wise and deterministic.
- Nit (dask+cupy not guarded): `test_perlin_dask_gpu` now monkeypatches `dask.persist` to raise, so the GPU path is covered by the same guard as the CPU path.

13 perlin tests pass locally, including the cupy and dask+cupy GPU tests on a CUDA host. No new findings.

PR #3475, opened 2026-06-23
Preserve caller coords/res/crs in generate_terrain (#3474)

### LINKED ISSUE #3474: generate_terrain() drops caller's coords, res, and crs
**Describe the bug**

`generate_terrain()` throws away the caller's spatial reference. When the input
DataArray already has coordinates and geo-attributes (`crs`, `res`, `units`), the
output drops them and builds a synthetic grid from `x_range`/`y_range` (default
`(0, 500)`) instead.

It's easiest to hit through the `.xrs` accessor, where you pass your own
georeferenced raster:

```python
my_attrs = {'res': (30, 30), 'crs': 'EPSG:5070', 'units': 'meters'}
xarr = xr.DataArray(arr, coords={'y': ys, 'x': xs}, dims=('y', 'x'),
                    name='study_area', attrs=my_attrs)

terrain = xarr.xrs.generate_terrain()
```

`xarr` has y/x coords on a 30 m grid, `res=(30, 30)`, and `crs='EPSG:5070'`. The
result comes back as:

```
Coordinates:
  * y        (y) float64  0.625 1.875 3.125 ... 498.1 499.4
  * x        (x) float64  0.8333 2.5 4.167 ... 497.5 499.2
Attributes:
    res:      (1.6666666666666667, 1.25)
```

The coords are rescaled to the `(0, 500)` default, `res` is overwritten, and
`crs`/`units` are gone.

**Expected behavior**

When the caller's array carries its own coords and geo-attrs, the output should
keep them: matching `coords`, `chunks` (dask), `attrs['res']`, and `attrs['crs']`.
A bare template with no coords/attrs should keep today's behavior of synthesizing
coords from `x_range`/`y_range`.

**Additional context**

In `xrspatial/terrain.py`, `generate_terrain()` builds fresh coords from
`x_range`/`y_range` and sets `attrs={'res': (dx, dy)}`, ignoring the caller's
coords and attrs. The accessor in `xrspatial/accessor.py` just forwards
`self._obj`, so the fix belongs in `generate_terrain`. The dask data chunks
already survive through `map_blocks` — coords and attrs are the gap.

brendancol:
Closes #3474

`generate_terrain()` was building fresh `y`/`x` coordinates from
`x_range`/`y_range` (default `(0, 500)`) and setting `attrs` to only
`{'res': (dx, dy)}`, so a georeferenced caller lost its coords, `res`, `crs`,
and `units`. This was easiest to hit through `xarr.xrs.generate_terrain()`.

Now, when the caller's array carries its own `y`/`x` coords, the output reuses
them and copies the caller's attrs (`crs`, `res`, `units`, ...). `x_range`/`y_range`
still drive the noise field. A bare template with no coords keeps the old
behavior: a synthetic grid plus a computed `res`.

When the caller has coords but no `res` attr, `res` is derived from the coord
spacing (which equals the old `(dx, dy)` for the synthetic case, so the
bare-template output is unchanged).

Backends: numpy, cupy, dask+numpy, dask+cupy. Dask data chunks were already
preserved through `map_blocks`; this restores the coords and attrs on top.

Test plan:
- [x] caller coords / res / crs / units preserved (numpy + accessor)
- [x] dask: chunks, coords, attrs preserved
- [x] bare template unchanged (synthetic grid, computed res, no crs)
- [x] coords-without-res derives res from spacing
- [x] full test_terrain.py suite passes (43 passed); cupy paths covered by guarded tests

### DIFF
--- a/xrspatial/terrain.py
+++ b/xrspatial/terrain.py
@@ -594,6 +594,10 @@ def generate_terrain(agg: xr.DataArray,
     ----------
     agg : xr.DataArray
         2D template array (determines height, width, and backend).
+        If it carries its own ``y``/``x`` coordinates and attrs (e.g.
+        ``crs``, ``res``, ``units``), they are preserved on the output and
+        ``x_range``/``y_range`` only drive the noise field. A bare template
+        with no coordinates gets a synthetic grid from ``x_range``/``y_range``.
     x_range : tuple, default=(0, 500)
         Range of x values.
     y_range : tuple, default=(0, 500)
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Preserve caller coords/res/crs in generate_terrain (#3474)

None.

- `xrspatial/tests/test_terrain.py`: there's no explicit dask+cupy parity test for the preservation path. The wrapper code is backend-agnostic so the risk is low, and the test would need a GPU runner, so this is fine to leave. Noting it for the record.

- `xrspatial/terrain.py:711-720`: the synthetic branch computes `dx`/`dy` for the coord `linspace`, then the `res` fallback re-derives spacing from the coords. That's intentional (single source of truth: the actual coord spacing) and produces the same `(dx, dy)` as before, so no change needed — just flagging the small redundancy.

### What looks good
- Backward compatibility is exact. `linspace(x0+dx/2, x1-dx/2, width)` has spacing exactly `dx`, so the bare-template `res` derived from coord spacing equals the old `(dx, dy)`. The `width==1`/`height==1` fallback to `x_range`/`y_range` span also matches the old `(x1-x0)/1`.
- The `erode=True` path is covered: `erode()` rebuilds the output with `coords=agg.coords, attrs=agg.attrs` (erosion.py:496-497), so crs/res/coords survive erosion too.
- Fix lives in the backend-agnostic public wrapper, so all four backends get the same treatment. Dask data chunks stay chunked, and the dask test asserts `terrain.data.chunks == src.data.chunks`.
- Caller attrs are copied with `dict(agg.attrs)` (shallow copy is fine for the str/tuple values here), and `res` is only synthesized when the caller didn't supply one.

### Checklist
- [x] Algorithm matches reference/paper (noise unchanged; only coord/attr wrapping)
- [x] NaN handling unchanged
- [x] Edge cases covered by tests (bare template, coords-without-res)
- [x] Dask chunk boundaries handled correctly (chunks asserted preserved)
- [x] Benchmark not needed (no perf-critical path changed)
- [x] Docstrings present and accurate (agg param documents the new behavior)

brendancol (commented):
## Follow-up review (after c0ede6be)

The one suggestion from the first pass is addressed: added
`test_generate_terrain_dask_cupy_preserves_coords_attrs`, double-guarded by
`@cuda_and_cupy_available` and `@dask_array_available`.

This box happens to have CUDA + cupy, so all four backends' preservation tests
ran for real and pass:

- test_generate_terrain_preserves_caller_coords_and_attrs (numpy)
- test_generate_terrain_accessor_preserves_georeference
- test_generate_terrain_dask_preserves_chunks_coords_attrs (dask+numpy)
- test_generate_terrain_cupy_preserves_coords_attrs (cupy)
- test_generate_terrain_dask_cupy_preserves_coords_attrs (dask+cupy)

Full `test_terrain.py`: 44 passed. No outstanding blockers, suggestions, or
nits. The earlier nit (synthetic-branch res re-derived from coord spacing) was
intentional and dismissed.

PR #3477, opened 2026-06-24
Add categorized repr to the .xrs accessor (#3476)

### LINKED ISSUE #3476: Rich `__repr__` for the .xrs accessor: list available tools by category
## Reason or Problem

The `.xrs` accessor on `xarray.DataArray` and `xarray.Dataset` exposes the
full xarray-spatial toolset (`slope`, `ndvi`, `watershed`, `zonal_stats`, and
dozens more). But there's no way to see what's available from inside a Python
session without tab-completion or an LSP. Type `da.xrs` in a REPL and you get
the default object repr:

```
<xrspatial.accessor.XrsSpatialDataArrayAccessor object at 0x7f...>
```

That says nothing about the operations you can call. On a bare REPL, in a
logged session, or reading code without an editor, there's no built-in way to
see the menu.

## Proposal

Give the accessor a `__repr__` (and a Jupyter `_repr_html_`) that lists the
available operations grouped by category, so `da.xrs` prints something like:

```
<XrsSpatialDataArrayAccessor> xarray-spatial tools, call as .xrs.<name>(...)

Surface: slope, aspect, hillshade, curvature
Hydrology: flow_direction, flow_accumulation, watershed, ...
Multispectral: ndvi, evi, savi, ...
...
```

**Design:**

The accessor source already groups its methods under `# ---- Category ----`
banner comments. A small helper parses those banners out of the class source
with `inspect.getsource` and builds an ordered list of `(category, [method
names])`. `__repr__` renders that as grouped, wrapped text; `_repr_html_`
renders it as a compact HTML list for notebooks. Both read from the same parsed
data. If source introspection isn't available, fall back to a flat alphabetical
listing of the public methods. The DataArray and Dataset accessors are parsed
independently so each shows its own method set.

**Usage:**

```python
>>> import xrspatial
>>> da.xrs
# prints the categorized list of available operations
```

**Value:** Discoverability without tooling. You can see the toolset from any
REPL or notebook.

## Drawbacks

Adds a bit of introspection code. The category banners become load-bearing
(they already exist in the source), but the flat fallback covers the case where
source isn't available.

## Alternatives

A hardcoded category-to-method mapping duplicates the banners and drifts out of
sync. A flat sorted list drops the category grouping that makes the output
scannable.

brendancol:
Closes #3476

Printing the `.xrs` accessor used to show the default object repr, which
tells you nothing about what you can call. Now it lists the available
operations grouped by category, so you can see the toolset from a plain
REPL or notebook without tab completion or an LSP.

- `__repr__` renders a grouped, wrapped text listing; `_repr_html_`
  renders a compact table in Jupyter.
- The grouping is parsed from the `# ---- Category ----` banners already
  in the accessor source, so it can't drift from the actual method set. A
  test asserts every public method lands in exactly one category.
- Falls back to a flat alphabetical method list if source introspection
  isn't available (e.g. a frozen build).

Example:

```
<XrsSpatialDataArrayAccessor> xarray-spatial tools for this DataArray
call as: .xrs.<name>(...)

Surface:
    slope, aspect, hillshade, curvature
Hydrology:
    flow_direction, flow_accumulation, watershed, ...
Multispectral (self = NIR band):
    ndvi, evi, arvi, savi, nbr, sipi
...
```

Backend coverage: not applicable. This is a pure-Python repr on the
accessor classes; it touches no array code, so numpy / cupy / dask paths
are unaffected.

Test plan:

- [x] `pytest xrspatial/tests/test_accessor.py` (79 passed)
- [x] New tests: category coverage of all public methods, text and HTML
      repr content, DataArray vs Dataset distinct listings, HTML escaping,
      source-unavailable fallback

### DIFF
--- a/docs/source/getting_started/quickstart.rst
+++ b/docs/source/getting_started/quickstart.rst
@@ -45,6 +45,14 @@ chaining. These two lines are equivalent:
    incline = slope(terrain)
    incline = terrain.xrs.slope()
 
+To see what's available without tab completion or an editor, print the
+accessor. Its repr lists the operations grouped by category, and renders
+as a table in a Jupyter notebook:
+
+.. code-block:: python
+
+   terrain.xrs        # prints slope, hillshade, ndvi, watershed, ... by category
+
 Reading and writing GeoTIFFs
 ============================
 
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -11,6 +11,11 @@ directly via tab-completion::
     nir.xrs.ndvi(red)
 """
 
+import functools
+import html
+import inspect
+import re
+import textwrap
 import warnings
 
 import xarray as xr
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Add categorized repr to the .xrs accessor (#3476)

None.

- The text and HTML fallbacks label the no-category case differently. The
  text path prints "(categories unavailable; flat method list)"; the HTML path
  uses an "Available tools" row header (accessor.py:638 vs accessor.py:655).
  Harmless, but matching the wording would make the two paths read the same.

- accessor.py:619: `_accessor_categories` is lru_cached and keyed on the class.
  Only two classes ever reach it, so the cache is a two-entry memo. Fine as-is;
  noting that `maxsize=None` is unbounded by design, not an oversight.

### What looks good
- The parser reads the `# ---- Category ----` banners already in the source, so
  the listing can't drift from the real method set. The coverage test
  (test_categories_cover_every_public_method) asserts an exact match against
  vars(cls), which also catches a future nested non-underscore def leaking into
  the listing.
- HTML output escapes method and category names, with a dedicated test that
  injects a `<script>` name to prove it.
- The source-unavailable fallback is tested with a type()-built class whose
  source inspect can't retrieve.
- Pure-Python change with no array code, so no backend / dask / cupy / NaN
  concerns apply.

### Checklist
- Algorithm matches reference/paper: N/A (no algorithm)
- All implemented backends produce consistent results: N/A (no array code)
- NaN handling is correct: N/A
- Edge cases are covered by tests: yes (fallback, HTML escaping, DataArray vs Dataset)
- Dask chunk boundaries handled correctly: N/A
- No premature materialization or unnecessary copies: N/A
- Benchmark exists or is not needed: not needed (repr is interactive, not a hot path)
- README feature matrix updated: N/A (no new function)
- Docstrings present and accurate: yes (helpers documented; quickstart updated)

brendancol (commented):
## Follow-up review (after e6d1be2)

The one Suggestion from the first pass is resolved. The text and HTML reprs
now both route through `_categories_or_fallback` (accessor.py:644), so the
no-banner case shows the same "Available tools" group in either path. The
fallback test was updated to assert the shared label.

No new findings. The Nit about `lru_cache(maxsize=None)` was a note, not a
defect, and stands as-is (two classes, intentional). Pure-Python repr change;
no backend, dask, cupy, or NaN concerns apply. Nothing blocking.

PR #3479, opened 2026-06-24
Forward per-tool repr to the accessed tool (#3478)

### LINKED ISSUE #3478: repr(da.xrs.slope) prints the whole accessor catalog instead of the tool's own info
## Description

Commit 8bb7c16c (#3476/#3477) gave the `.xrs` accessor a categorized `__repr__`, so `da.xrs` and `ds.xrs` list the available operations grouped by category. That part works.

The problem shows up one level down. Evaluating a single tool without calling it, e.g. `da.xrs.slope`, prints the entire accessor listing again instead of information about `slope`:

```
>>> da.xrs.slope
<bound method XrsSpatialDataArrayAccessor.slope of <XrsSpatialDataArrayAccessor> xarray-spatial tools for this DataArray
call as: .xrs.<name>(...)

Plot:
    plot
Surface:
    slope, aspect, hillshade, curvature
... (the whole catalog) ...>
```

`da.xrs.slope` is a bound method. Python builds a bound method's repr as `<bound method Cls.name of {repr(self)}>`, and `repr(self)` now routes through the accessor's custom `__repr__`, so the full catalog ends up embedded in the bound-method repr.

## Expected behavior

Evaluating `da.xrs.slope` should show information scoped to `slope` only: its call signature and docstring, the same documentation `help(da.xrs.slope)` already surfaces. The docstring is already attached to each accessor method (see `_attach_delegated_docs`), so the per-tool repr can reuse it.

## Steps to reproduce

```python
import numpy as np, xarray as xr, xrspatial
da = xr.DataArray(np.ones((4, 4)), dims=['y', 'x'])
da.xrs.slope   # prints the whole accessor catalog instead of slope's info
```

## Environment

- xarray-spatial: main (after #3477)

brendancol:
## Summary

Closes #3478.

After #3477 added a categorized `__repr__` to the `.xrs` accessor, evaluating a single tool without calling it (e.g. `da.xrs.slope`) printed the whole accessor catalog. `da.xrs.slope` is a bound method, and Python builds a bound method's repr as `<bound method Cls.name of {repr(self)}>`, so `repr(self)` routed through the accessor's catalog `__repr__` and embedded the entire listing.

- Accessor attribute access now wraps public bound methods in a small callable `_AccessorTool` proxy. Its `__repr__` / `_repr_html_` show that tool's own signature and docstring instead. Calls forward unchanged, and `__doc__` / `__wrapped__` are mirrored so `help()` and `inspect` keep seeing the delegated docstring.
- The accessor-level catalog repr (`da.xrs`) is untouched.

## Backend coverage

Pure accessor-layer change; no compute path is touched. Behaves the same across numpy / cupy / dask+numpy / dask+cupy.

## Test plan

- [x] `repr(da.xrs.slope)` shows slope's signature + docstring, not the catalog
- [x] tools stay callable and match the direct function call
- [x] `inspect.getdoc` / `help()` metadata preserved on the proxy
- [x] HTML repr scoped to the tool
- [x] Dataset accessor scoped the same way
- [x] accessor-level catalog repr still renders
- [x] full `test_accessor.py` suite passes (86 tests), flake8 clean

### DIFF
--- a/docs/source/getting_started/quickstart.rst
+++ b/docs/source/getting_started/quickstart.rst
@@ -53,6 +53,13 @@ as a table in a Jupyter notebook:
 
    terrain.xrs        # prints slope, hillshade, ndvi, watershed, ... by category
 
+Drop the call parentheses on a single tool to read its signature and
+docstring, the same text ``help(terrain.xrs.slope)`` shows:
+
+.. code-block:: python
+
+   terrain.xrs.slope  # prints the slope signature and docstring
+
 Reading and writing GeoTIFFs
 ============================
 
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -689,12 +689,87 @@ def _accessor_repr_html(cls):
     )
 
 
+# ---------------------------------------------------------------------------
+# Per-tool repr: evaluating a single tool (``da.xrs.slope``) without calling it
+# should show that tool's own signature and docstring, not the whole accessor
+# catalog. A bound method's repr is ``<bound method Cls.name of {repr(self)}>``,
+# and ``repr(self)`` routes through the accessor's catalog ``__repr__``, so
+# without intervention the full listing gets embedded (issue #3478). The
+# accessor wraps public methods in this proxy on attribute access instead.
+# ---------------------------------------------------------------------------
+
+
+def _accessor_tool_repr_text(method):
+    """Plain-text repr for a single accessor tool: signature + docstring."""
+    name = method.__name__
+    try:
+        sig = str(inspect.signature(method))
+    except (TypeError, ValueError):
+        sig = "(...)"
+    lines = [f".xrs.{name}{sig}"]
+    doc = inspect.getdoc(method)
+    if doc:
+        lines += ["", doc]
+    return "\n".join(lines)
+
+
+def _accessor_tool_repr_html(method):
+    """Jupyter HTML repr for a single tool: the text repr in a <pre> block."""
+    return (
+        "<pre style='white-space:pre-wrap'>"
+        f"{html.escape(_accessor_tool_repr_text(method))}"
+        "</pre>"
+    )
+
+
+class _AccessorTool:
+    """Callable proxy returned for a public accessor method on attribute access.
+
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Forward per-tool repr to the accessed tool (#3478)

I read accessor.py in full and exercised the proxy against the edge cases below. The fix is correct and targeted. No blockers.

### Blockers
None.

### Suggestions
None.

### Nits (optional)
- `_AccessorTool` returns a fresh proxy on every attribute access, so `da.xrs.slope == da.xrs.slope` is now False where a bound method compared equal. Harmless for normal use, but anything downstream comparing accessor methods by equality would see the change. Low risk; not worth caching unless it bites.
- `__getattribute__` now runs for every attribute access on the accessor, including `self._obj` inside each method body. The early `name.startswith("_")` return keeps the hot path cheap (one extra call plus a startswith), so this is fine, just noting the cost is non-zero.

### What looks good
- Root cause is correctly identified: a bound method's repr embeds `repr(self)`, which routed through the catalog `__repr__`. Wrapping at attribute access is the right layer.
- The proxy mirrors `__doc__` / `__wrapped__` / `__name__`, so `help()`, `inspect.signature`, and `inspect.getdoc` keep working. Verified `inspect.signature(da.xrs.slope)` returns `(**kwargs)` and `getdoc` matches the standalone function.
- Private attributes and the catalog `__repr__` are left untouched, so `da.xrs` still renders the full listing.
- Calls forward unchanged, positional args included (ndvi, rasterize are covered by the existing suite, which now all runs through the proxy).
- Missing attributes still raise AttributeError.
- Tests cover the scoped text repr, HTML repr, callable forwarding, help metadata, the Dataset accessor, and the unchanged catalog repr. Both accessors are handled.

### Checklist
- [x] Root cause correctly identified
- [x] help()/inspect metadata preserved
- [x] Catalog repr unchanged
- [x] Calls forward unchanged (incl. positional args)
- [x] Both DataArray and Dataset accessors covered
- [x] Tests present and passing (86 in test_accessor.py), flake8 clean
- [x] No compute path touched; backend-agnostic

PR #3481, opened 2026-06-24
Accept GeoDataFrames in interpolation + expose on .xrs accessor with coregister

### LINKED ISSUE #3480: Accept GeoDataFrames in the interpolation functions and expose them on the .xrs accessor with coregister
## Reason or Problem

The interpolation family — `kde`/`line_density` (`xrspatial/kde.py`) and `idw`/`kriging`/`spline` (`xrspatial/interpolate/`) — only accepts raw `x, y, z` array-likes. To produce a surface aligned to an existing raster, a user has to unpack a GeoDataFrame into x/y/z arrays by hand, reproject those points into the raster's CRS themselves, and pass the raster in as `template`. Meanwhile `rasterize` already accepts a GeoDataFrame directly and pulls its burn value from a `column`, and `.xrs.rasterize` aligns the result to the caller raster. The interpolation functions are inconsistent with that.

## Proposal

Let the interpolation functions consume geospatial vectors the same way `rasterize` does, and expose them on the `.xrs` accessor so they align to the caller raster automatically. The target call:

```python
surface = my_raster.xrs.kde(my_points, coregister=True)
```

`my_points` is a GeoDataFrame; the result is a DataArray on `my_raster`'s grid (bounds, resolution, chunks, CRS), with the points reprojected from the GeoDataFrame's CRS into the raster's CRS first.

**Design:**

Two layers, mirroring how `rasterize` is split between a standalone function and a thin accessor delegator.

1. *GeoDataFrame input on the standalone functions.* A shared helper `points_from_geodataframe(gdf, column=..., value_required=...)` returns `(x, y, value, crs)` from a GeoDataFrame of Point geometries. The numeric-column auto-pick mirrors `rasterize._parse_input` (first numeric column when `column` is omitted). Each function detects a GeoDataFrame in its first positional argument and extracts arrays from it; the existing `x, y, z, template` array path is unchanged. For `kde` the `column` maps to per-point `weights` (uniform if omitted); for `idw`/`kriging`/`spline` it maps to the interpolated value `z`.

2. *Accessor methods plus `coregister`.* Add `kde`/`idw`/`kriging`/`spline` to both `.xrs` accessor classes, passing the caller array as `template` so grid, chunks, CRS, and attrs are inherited. With `coregister=True`, the point coordinates are transformed from the GeoDataFrame's CRS into the caller's CRS before interpolation — the vector analog of `open_geotiff(coregister=True)`. Reprojection reuses the numba fast path in `reproject._projections.transform_points` and falls back to `pyproj.Transformer`. With `coregister=False`, a genuine CRS mismatch raises, matching `rasterize`'s `check_crs` default.

**Usage:**

```python
# Standalone, GeoDataFrame in:
kriging(points_gdf, template=dem, column="elevation")
kde(points_gdf, template=grid)            # uniform-weight density

# Accessor, aligned to the caller raster:
dem.xrs.spline(points_gdf, column="elevation")
grid.xrs.kde(points_gdf, coregister=True)  # reproject points into grid's CRS
```

**Value:** A GeoDataFrame of observations turns into a coregistered interpolated surface in one call, with no manual coordinate unpacking or hand-written reprojection. It also makes the interpolation functions behave like `rasterize`, which users already know.

## Stakeholders and Impacts

Users interpolating scattered geospatial point data onto a raster grid. Touches `xrspatial/kde.py`, `xrspatial/interpolate/`, and `xrspatial/accessor.py`. geopandas stays an optional import (lazy, guarded), so no new hard dependency. The existing array-based signatures keep working unchanged.

## Drawbacks

Adds a GeoDataFrame branch to four public functions and grows the accessor surface. The standalone GeoDataFrame form needs `template`/`column` passed as keywords because `template` is positional today.

## Alternatives

Leave reprojection to the caller and accept only same-CRS GeoDataFrames; or reproject the raster output instead of the points (less accurate for scattered data, and not what the request asks for).

## Unresolved Questions

None blocking. `line_density` is left out of scope (its four-endpoint input does not map to a single point column).

Scope confirmed to cover all four functions: `kde`, `idw`, `kriging`, `spline`.

brendancol:
Closes #3480

Lets the interpolation functions consume a GeoDataFrame the way `rasterize` does, and exposes them on the `.xrs` accessor so a surface lands on the caller raster's grid:

- `kde`, `idw`, `kriging`, and `spline` accept a GeoDataFrame of Point geometries as the first argument. `column` picks the value to interpolate (z for idw/kriging/spline, per-point weights for kde), defaulting to the first numeric column. The raw `x, y, z` array form is unchanged.
- The four operations are on both `.xrs` accessor classes. They pass the caller array as `template`, so grid, chunks, CRS, and attrs are inherited. The existing array-form accessor methods (`idw`/`kriging`/`spline`) keep working; `kde` is new on the accessor.
- `coregister=True` reprojects a GeoDataFrame's points from their CRS into the caller's CRS before interpolating, the vector analog of `open_geotiff(coregister=True)`. With `coregister=False` (default), a genuine CRS mismatch raises, matching `rasterize`'s `check_crs`.

So the headline call works:

```python
surface = raster.xrs.kde(points_gdf, coregister=True)
```

geopandas stays an optional, lazily-imported dependency.

### Note on reprojection

The plan floated reusing `reproject._projections.transform_points` for the point transform. The implementation uses `GeoDataFrame.to_crs(target)` instead: it is the idiomatic geopandas path, handles datum shifts and axis order through pyproj, and avoids reimplementing coordinate transforms. The coregistered result is bit-identical to interpolating the pre-projected points in the notebook check.

### Backends

numpy / dask+numpy verified directly; cupy / dask+cupy inherit from the unchanged `ArrayTypeFunctionMapping` dispatch (the GeoDataFrame branch only touches input parsing, which runs before backend dispatch). Output backend and chunks follow the caller raster's `template`.

### Test plan

- [x] GeoDataFrame form matches the array form for all four functions (numpy + dask templates)
- [x] Accessor methods match the standalone calls and inherit grid/chunks/CRS
- [x] `coregister=True` (EPSG:4326 points onto an EPSG:3857 grid) matches pre-projected points
- [x] CRS mismatch raises without coregister; missing CRS on either side is a no-op
- [x] Edge cases: non-Point geometry, missing/unknown column, positional y/z with a GeoDataFrame, kde weights+column conflict
- [x] Legacy array-form accessor signature still works
- [x] Existing `test_interpolation.py` / `test_kde.py` / `test_accessor.py` pass unchanged
- [x] User guide notebook executes end to end


brendancol (commented):
## PR Review: Accept GeoDataFrames in interpolation + expose on .xrs accessor with coregister

Reviewed the full changed files in a worktree on the PR branch. The design mirrors `rasterize` cleanly and keeps the array path intact. Findings are minor.

### Blockers
None.

### Suggestions
- `xrspatial/interpolate/_vector.py:38` (`_resolve_value_column`): an explicit `column` pointing at a non-numeric column surfaces a raw `ValueError: could not convert string to float: 'a'` from numpy rather than a function-prefixed message. The auto-pick path already gives a clean "no numeric column" error; consider catching the cast failure and re-raising as `f"{func_name}(): column {column!r} is not numeric"` so both paths read consistently. Low priority since it still raises and does not corrupt output.

### Nits
- Test coverage exercises numpy and dask templates only. cupy / dask+cupy are not asserted. This is acceptable because the GeoDataFrame branch runs entirely before `ArrayTypeFunctionMapping` dispatch, so it is backend-independent, but a single cupy smoke test (guarded by `cuda_and_cupy_available`) would document that explicitly.
- No benchmark added. Reasonable here: the change is input parsing plus an optional `to_crs`, not a new compute path, so the existing interpolation benchmarks still cover the hot loop.

### What looks good
- The `to_crs`-based coregistration is the right call over a hand-rolled transform: it goes through pyproj for datum shifts and axis order, and the notebook check shows the coregistered result is identical to interpolating pre-projected points.
- Backward compatibility is preserved: the legacy `da.xrs.idw(x, y, z)` array signature still works, and there is a test pinning it.
- CRS handling matches `rasterize` semantics (mismatch raises, missing CRS on either side is a no-op), so the behavior is predictable for anyone who already uses `rasterize`.
- Edge cases raise clear errors: MultiPoint, empty frame, missing column, positional y/z with a GeoDataFrame, and the kde weights+column conflict are all covered by tests.
- Caller chunks are inherited through `template`, verified for all four functions on a dask-backed grid.

### Checklist
- [x] Algorithm matches reference (no algorithm change; parsing only)
- [x] All implemented backends produce consistent results (parsing precedes dispatch)
- [x] NaN handling is correct (existing `validate_points` filtering unchanged)
- [x] Dask chunk boundaries handled correctly (inherited from template)
- [x] Benchmark not needed

brendancol (commented):
## Follow-up review (after 7c4333b4)

Re-reviewed the follow-up commit against the prior findings.

- **Suggestion (non-numeric column error)** — fixed. `_resolve_value_column` now catches the cast failure on an explicit `column` and re-raises `f"{func_name}(): column {column!r} is not numeric"`, matching the clean message the auto-pick path already gave. Covered by `test_non_numeric_column_raises`.
- **Nit (cupy coverage)** — addressed. `test_geodataframe_parity_cupy` (guarded by `cuda_and_cupy_available`) asserts the GeoDataFrame branch matches numpy on a cupy template. It runs and passes on a CUDA box.
- **Nit (benchmark)** — dismissed with reason: the change is input parsing plus an optional `to_crs`, not a new compute path, so the existing interpolation benchmarks still cover the hot loop.

No new findings. Remaining items are dismissed-with-reason only.

PR #3483, opened 2026-06-24
Support string/categorical columns in rasterize with QGIS-visible labels

### LINKED ISSUE #3482: Support string/categorical columns in rasterize with QGIS-visible category labels
## Reason or Problem

`rasterize` only accepts numeric columns. `_parse_input` calls `.astype(np.float64)` on the chosen column, so passing a string or categorical field (land-cover classes, zoning codes, and similar) raises `ValueError: could not convert string to float`. Users with text-keyed vector data have to hand-encode their categories to integers before rasterizing, then keep a separate lookup table to remember which integer meant what. That mapping is lost once the raster is written to disk, so opening the result in QGIS shows bare numbers instead of class names.

## Proposal

Accept a string or categorical column in `rasterize`, encode the labels to integer codes, and carry the value-to-label mapping in the output `attrs` so it can be written to a GeoTIFF and shown in QGIS.

**Design:**

- Encoding: when the selected `column` has a non-numeric dtype (object/string/`CategoricalDtype`), encode it with pandas Categorical. Already-categorical columns keep their declared category order; plain string/object columns get the pandas default (lexically sorted). Pixel values are the integer codes `0..N-1`; missing values use the pandas `-1` sentinel.
- Output: default to `int32` with nodata `-1` when a categorical column is detected and the caller did not pass `dtype`/`fill`. Numeric columns and `(geometry, value)` pairs keep their current float64/NaN behavior.
- Attrs: store `category_names` (ordered list of labels, index = pixel value) and an auto-generated `category_colors` (one distinct RGBA per category, from an evenly spaced HSV spread).
- GeoTIFF interop: `to_geotiff` writes a PAM `<file>.tif.aux.xml` sidecar containing `<CategoryNames>` and a thematic `<GDALRasterAttributeTable>` (Value/Class plus Red/Green/Blue/Alpha columns). This is the only mechanism GDAL/QGIS reads for category labels. An embedded RAT in the GeoTIFF `GDAL_METADATA` tag is ignored by GDAL (confirmed with gdalinfo 3.10.3), while the sidecar is read back as `Categories:` and a full RAT. `open_geotiff` parses the sidecar back into `attrs` for a round-trip.

**Usage:**

```python
import geopandas as gpd
from xrspatial import rasterize

gdf = gpd.GeoDataFrame(
    {'landcover': ['water', 'forest', 'urban']},
    geometry=[box(0, 0, 5, 5), box(5, 0, 10, 5), box(0, 5, 5, 10)],
    crs='EPSG:4326',
)
result = rasterize(gdf, column='landcover', width=100, height=100)
# result is int32, pixels are 0/1/2, nodata -1
# result.attrs['category_names'] == ['forest', 'urban', 'water']
to_geotiff(result, 'landcover.tif')   # also writes landcover.tif.aux.xml
```

Opening `landcover.tif` in QGIS shows the class names instead of the raw codes.

**Value:** Rasterizing thematic vector data becomes a single call with no manual encoding or external lookup table, and the labels travel with the file into any GDAL-based tool.

## Stakeholders and Impacts

Anyone rasterizing categorical vector data. Touches `rasterize` (`_parse_input`, dtype/fill defaults, attrs) and the GeoTIFF reader/writer (new sidecar helper, `to_geotiff` write hook, `open_geotiff` read, attrs contract). The `.xrs` accessor needs no change since it forwards kwargs. Encoding happens before backend dispatch, so all four backends are covered.

## Drawbacks

Writing a GeoTIFF now produces a second file (the `.aux.xml` sidecar) when categories are present. This is the standard GDAL PAM convention, but it is an extra artifact to keep alongside the `.tif`.

## Alternatives

- Embed the RAT/category names in the GeoTIFF `GDAL_METADATA` tag (no sidecar). Rejected: GDAL does not read it back, so QGIS would still show bare numbers.
- Write a TIFF ColorMap tag for colors. Not valid for an int32 band (that tag is limited to <=16-bit palette images), so colors go in the RAT instead.

GDAL RAT field usage codes used: Value=MinMax(5), Class=Name(2), Red/Green/Blue/Alpha=6/7/8/9. Field types: Integer=0, String=2.

brendancol:
Closes #3482.

## What this does

`rasterize` now accepts a string or categorical GeoDataFrame column as `column=`. The labels are encoded to integer codes and the value-to-label map rides along in the result attrs, so it can be written to a GeoTIFF and shown in QGIS.

- A non-numeric `column` is label-encoded with pandas Categorical: codes `0..N-1` on an `int32` band with `-1` nodata. Object/string columns sort lexically; an existing `Categorical` keeps its declared order. Explicit `dtype`/`fill` still win.
- The result carries `attrs['category_names']` (index == pixel code) and an auto-generated `attrs['category_colors']` (one RGBA per class).
- `to_geotiff` writes a PAM `<file>.tif.aux.xml` sidecar (`<CategoryNames>` plus a thematic RAT with Value/Class and RGBA columns) when those attrs are present. This is the only mechanism GDAL/QGIS read for category labels; an embedded RAT in the GDAL_METADATA tag is ignored (confirmed with gdalinfo 3.10.3).
- `open_geotiff` parses the sidecar back into attrs, so the round-trip preserves the labels and colors.

Numeric columns and `(geometry, value)` pairs are unchanged.

## Backend coverage

Encoding happens in `_parse_input` before backend dispatch, so all four backends (numpy / cupy / dask+numpy / dask+cupy) burn the same integer codes. The sidecar read and write are backend-independent.

## Test plan

- [x] Encoding: string column -> int32 codes, sorted `category_names`, codes match category order, default fill -1.
- [x] Ordered `Categorical` preserves declared order.
- [x] Missing category -> nodata; explicit fill remaps the missing code.
- [x] One color per category; RGBA in range.
- [x] Numeric column unchanged (float64, no category attrs).
- [x] GeoTIFF round-trip: sidecar written, `open_geotiff` reads names and colors back, no sidecar for numeric rasters.
- [x] `gdalinfo` shows the `Categories:` block (guarded by `shutil.which`).
- [x] PAM XML build/parse round-trip including special-character escaping.
- [x] GPU parity: codes and names match CPU (ran locally with CUDA).
- [x] Existing `test_rasterize.py` (233 passed) and geotiff writer/reader suites green.
- [x] User guide notebook executes clean.

### DIFF
--- a/docs/source/user_guide/rasterize.ipynb
+++ b/docs/source/user_guide/rasterize.ipynb
@@ -3,7 +3,7 @@
   {
    "cell_type": "markdown",
    "id": "jb1yj58wq4q",
-   "source": "## Rasterize\n\n`xrspatial.rasterize` converts vector geometries (polygons, lines, points) into a 2D `xr.DataArray`. No GDAL dependency required.\n\nThis guide covers:\n- [Basic rasterization](#Basic-rasterization) -- polygons, lines, and points\n- [Merge modes](#Merge-modes) -- controlling how overlapping geometries combine\n- [Custom merge functions](#Custom-merge-functions) -- user-defined numba-jitted merge logic\n- [Dask parallel rasterization](#Dask-parallel-rasterization) -- tile-based output chunking",
+   "source": "## Rasterize\n\n`xrspatial.rasterize` converts vector geometries (polygons, lines, points) into a 2D `xr.DataArray`. No GDAL dependency required.\n\nThis guide covers:\n- [Basic rasterization](#Basic-rasterization) -- polygons, lines, and points\n- [Categorical columns](#Categorical-columns) -- burn string/categorical labels, readable in QGIS\n- [Merge modes](#Merge-modes) -- controlling how overlapping geometries combine\n- [Custom merge functions](#Custom-merge-functions) -- user-defined numba-jitted merge logic\n- [Dask parallel rasterization](#Dask-parallel-rasterization) -- tile-based output chunking",
    "metadata": {}
   }
  ],
--- a/examples/user_guide/28_Rasterize.ipynb
+++ b/examples/user_guide/28_Rasterize.ipynb
@@ -18,20 +18,21 @@
     "### What you'll build\n",
     "\n",
     "1. Rasterize land-use zones with the `.xrs` accessor\n",
-    "2. Handle overlapping polygons and interior holes\n",
-    "3. Burn lines and points into a raster\n",
-    "4. Compare merge modes for overlapping features\n",
-    "5. Write a custom numba merge function\n",
-    "6. Use multi-column properties for density mapping\n",
-    "7. Run Dask parallel rasterization\n",
-    "8. Combine rasterization with zonal statistics\n",
-    "9. Compare default vs. `all_touched` rasterization\n",
-    "10. Use the standalone `rasterize()` function with geometry pairs\n",
+    "2. Burn a string or categorical column into a labeled raster\n",
+    "3. Handle overlapping polygons and interior holes\n",
+    "4. Burn lines and points into a raster\n",
+    "5. Compare merge modes for overlapping features\n",
+    "6. Write a custom numba merge function\n",
+    "7. Use multi-column properties for density mapping\n",
+    "8. Run Dask parallel rasterization\n",
+    "9. Combine rasterization with zonal statistics\n",
+    "10. Compare default vs. `all_touched` rasterization\n",
+    "11. Use the standalone `rasterize()` function with geometry pairs\n",
     "\n",
     "![Rasterize preview](images/rasterize_preview.png)\n",
     "\n",
     "**Jump to a section:**\n",
-    "[Basic rasterization](#Basic-rasterization) | [Overlapping polygons](#Overlapping-polygons) | [Lines and points](#Lines-and-points) | [Merge modes](#Merge-modes) | [Custom merge](#Custom-merge) | [Multi-column properties](#Multi-column-properties) | [Dask parallel](#Dask-parallel) | [Zonal statistics](#Zonal-statistics) | [All touched](#All-touched) | [Standalone function](#Standalone-function)"
+    "[Basic rasterization](#Basic-rasterization) | [Categorical columns](#Categorical-columns) | [Overlapping polygons](#Overlapping-polygons) | [Lines and points](#Lines-and-points) | [Merge modes](#Merge-modes) | [Custom merge](#Custom-merge) | [Multi-column properties](#Multi-column-properties) | [Dask parallel](#Dask-parallel) | [Zonal statistics](#Zonal-statistics) | [All touched](#All-touched) | [Standalone function](#Standalone-function)"
    ]
   },
   {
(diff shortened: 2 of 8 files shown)


brendancol (commented):
## PR Review: Support string/categorical columns in rasterize with QGIS-visible labels

Reviewed the diff in full from a worktree on the PR branch. The rasterize-side encoding and the write path look solid and well tested. The reader picked up two real problems that bite on ordinary files, not just categorical ones.

- [ ] `read_pam_sidecar` crashes on common GDAL statistics sidecars (`xrspatial/geotiff/_pam.py:135`, `:170`). `open_geotiff` now reads `<source>.aux.xml` for every local string source. GDAL routinely writes a `.aux.xml` holding band statistics with a histogram RAT whose Value column is `Real` (`<F>0.0</F>`). `_parse_rat` does `int(fields[value_col])`, which raises `ValueError: invalid literal for int() with base 10: '0.0'`. That exception propagates out through `_attach_category_attrs`, so opening a perfectly normal GeoTIFF that happens to have a stats sidecar would fail. Reproduced with an athematic histogram sidecar.
- [ ] Non-categorical RAT yields bogus `category_names` (`xrspatial/geotiff/_pam.py:159-188`). An athematic RAT with no Name column falls through to `name = ''` for every row, so a histogram/statistics RAT would attach `category_names = ['', '', ...]` to a continuous raster. Gate on `tableType == 'thematic'` and require a Name column; otherwise fall back to the `<CategoryNames>` element (which GDAL only writes for real categories).

- [ ] Make sidecar reads fully fail-closed (`xrspatial/geotiff/_pam.py:122-157`). The docstring promises an empty dict when the sidecar "cannot be parsed", but the `try` only covers the file read and `safe_fromstring`. Element lookups and `_parse_rat` run outside it, so a malformed external sidecar raises instead of returning `{}`. Widen the guard to cover parsing, and parse the value cell as `int(float(...))` for tolerance.

- [ ] Passing a non-numeric column via the plural `columns=` arg still raises the raw `could not convert string to float` from `.astype(np.float64)` (`xrspatial/rasterize.py:3801`). Categorical encoding only applies to the singular `column=`. Pre-existing behavior; a clearer message would help, but it is out of scope here.

### What looks good

- Encoding lives in `_parse_input` before backend dispatch, so all four backends burn identical codes; GPU parity is tested directly.
- The empirical call on storage is right: an embedded RAT is ignored by GDAL, the PAM sidecar is what QGIS reads. The round-trip test asserts it via `gdalinfo`.
- Missing categories map to the fill via the `-1` pandas sentinel, with an explicit-fill remap and a test for it.
- XML special characters are escaped and covered by a test.

### Checklist
- [x] Algorithm matches intent (pandas categorical codes, sorted vs declared order)
- [x] All implemented backends produce consistent results (GPU parity tested)
- [x] NaN / missing handling correct (-1 sentinel -> fill)
- [x] Edge cases covered (missing category, explicit fill, numeric unchanged)
- [ ] Reader robust on arbitrary real-world sidecars (see Blockers)
- [x] Benchmark not needed (no new perf-critical path)
- [x] README matrix unchanged (no new function, backend support unchanged)

brendancol (commented):
## Follow-up review (after 5c18598)

Re-reviewed the reader after the fixes. Both blockers are resolved and the follow-up has regression tests.

### Blockers -- resolved
- `read_pam_sidecar` no longer crashes on GDAL statistics sidecars. The whole parse runs inside a `try` that returns `{}` on `OSError`/`ValueError`/`TypeError`, and the Value cell is parsed as `int(float(...))` so a `Real`-typed cell like `0.0` is fine (`xrspatial/geotiff/_pam.py:118-150`, `:181`). Confirmed: opening a `.tif` with an athematic histogram sidecar returns `{}` and does not raise.
- Non-categorical RATs no longer leak bogus names. The reader only consults the RAT when `tableType == 'thematic'`, and `_parse_rat` returns `(None, None)` without a Name column, falling back to `<CategoryNames>` (`_pam.py:134`, `:169-171`).

### New tests
- `test_athematic_stats_sidecar_ignored` and `test_malformed_sidecar_returns_empty` lock in the fail-closed behavior. Full file: 18 passed.

### Disposition of the original review
- Blocker (stats-sidecar crash): fixed.
- Blocker (bogus names from athematic RAT): fixed.
- Suggestion (fully fail-closed reads + tolerant value parse): fixed.
- Nit (cryptic error on non-numeric `columns=`): left as-is. It is pre-existing behavior on the plural path and outside this PR's scope; the singular `column=` is the documented entry point for categorical data.

No new issues found.

PR #3486, opened 2026-06-24
Add .xrs.validate() to check a raster against the xarray-spatial contract

### LINKED ISSUE #3485: Add .xrs.validate() to check a raster against the xarray-spatial contract
## Reason or Problem

xrspatial functions all assume the same thing about their input DataArray, but none of them say so in one place. The array has to be an `xr.DataArray`; its last two dims have to be the spatial axes (2D, or 3D with a band/time axis in front); the dtype has to be real numeric; coordinates have to be numeric, monotonic, and evenly spaced; the cell size has to be positive and finite; lat/lon has to fall in geographic range; and geodesic methods and GeoTIFF I/O need a CRS.

Today those rules live scattered across the `utils.py` validators and the error messages of individual ops. If you load a raster that breaks one of them, you don't find out until some op raises partway through your pipeline, and the error only tells you about the first thing it tripped on. There's no way to check a raster against the whole contract before you start.

## Proposal

Add a `validate()` method to the `.xrs` accessor on both DataArray and Dataset. `da.xrs.validate()` checks the array against the contract and hands back a report of everything that's out of compliance, each with a suggestion for how to fix it.

**Design:**

A new `xrspatial/validate.py` holds the report classes and a top-level `validate()` function. Accessor methods already delegate to top-level functions, so this follows that pattern.

The report is a `ValidationReport` holding a list of `ValidationIssue` records (severity, check id, message, suggestion). There are two severities: `error` means the op will fail, `warning` means behavior degrades. `is_valid` and `bool(report)` look at errors only, so an array with nothing but warnings still reads as valid. Nothing raises by default; pass `raise_on_error=True` and it raises an aggregated `XrsContractError`. The report has a text repr and a Jupyter HTML repr matching the existing accessor repr style. On a Dataset it reports per variable.

The checks reuse what's already in `utils.py` instead of re-deriving it: dtype and ndim from `_validate_raster`, resolution from `get_dataarray_resolution`/`calc_res`, even-spacing from `_warn_if_irregular_spacing`, geographic bounds from `_validate_geographic_range`. Every check looks at structure only (dims, coords, dtype, attrs) and never reads the data buffer, so dask and cupy arrays validate without materializing and all four backends behave identically.

**Usage:**

```python
report = da.xrs.validate()
print(report)
# ValidationReport: 2 issues
# [error]   dtype is <U6 (string). Cast to float: da.astype('float64')
# [warning] 'x' coords not evenly spaced. Set da.attrs['res']=(dx, dy)
if not report:
    ...  # array is non-compliant

da.xrs.validate(raise_on_error=True)  # raises XrsContractError on any error
ds.xrs.validate()                     # per-variable report
```

**Value:**

You get every contract problem in one shot, with a fix for each, instead of peeling them off one op at a time. Since it doesn't raise by default, it's fine to poke at a raster in a notebook to see what's wrong with it.

## Stakeholders and Impacts

Mainly people loading rasters from outside sources (raw numpy arrays, third-party readers) that may not meet the contract. It's one new module plus a method on each accessor class. Nothing existing changes.

## Drawbacks

It's more surface to keep maintained. The contract is implicit, so the checks could drift from what individual ops actually require. Reusing the `utils.py` helpers is the hedge against that.

## Alternatives

Per-op targeting (`validate(for_op='slope')`) — deferred so the first cut stays focused on the general contract. Raising on non-compliance instead of returning a report — rejected, since it's worse for poking at things interactively.

## Unresolved Questions

Nothing blocking. Per-op targeting can be a follow-up.

brendancol:
Closes #3485

Adds `validate()` to the `.xrs` accessor on both DataArray and Dataset. It checks a raster against the input contract every xrspatial op assumes and hands back a report of every violation, each with a suggested fix, instead of letting ops raise one at a time partway through a pipeline.

- New `xrspatial/validate.py`: `validate()` / `validate_dataset()` plus `ValidationReport`, `DatasetValidationReport`, `ValidationIssue`, and `XrsContractError`. The report doesn't raise and is truthy when there are no error-level issues; pass `raise_on_error=True` to raise instead.
- Two severities. An `error` means a spatial op will fail (wrong type/ndim/dtype, non-numeric or non-monotonic coords, non-finite cell size). A `warning` means behavior degrades (unconventional dim names, missing coords, uneven spacing, projected lat/lon, missing CRS). `is_valid` keys off errors only.
- Checks reuse the existing `utils.py` validators and the polygonize CRS-resolution order. They look at structure only (dims, coords, dtype, attrs) and never read the data buffer, so dask stays lazy and cupy stays on device.
- Accessor methods delegate to the standalone functions and show up under a new Diagnostics banner in the categorized repr.

Backends: numpy, dask+numpy, cupy, dask+cupy. The check is structural metadata only, so it behaves the same on all four (the cupy and dask+cupy tests run on the dev box, not skipped).

Test plan:
- [x] New `xrspatial/tests/test_contract_validate.py` covering each error and warning check, severity semantics, `raise_on_error`, per-variable Dataset reports, repr, and all four backends
- [x] `test_accessor.py` expected-method lists updated; full accessor suite passes
- [x] `test_validation.py` still green
- [x] flake8 clean on new files


brendancol (commented):
## PR Review: Add `.xrs.validate()` to check a raster against the xarray-spatial contract

Reviewed in the worktree on the head branch. Two findings, both fixed in the follow-up commit on this branch.

### Blockers
None.

### Suggestions
- **`xrspatial/validate.py` `_check_geographic_range` — false warning on a bare lat/lon dim.** The check read coordinates with `coords.get(dim)`, which synthesizes a default integer index (`0..N-1`) for a dimension that has no real coordinate. A DataArray with a dim named `lat` but no coordinate and more than 91 rows would get synthesized indices whose max exceeds 90, producing a misleading "looks projected" warning on top of the correct `coords_present` warning. Fixed: the geographic check now runs only against a real coordinate (`dim in agg.coords`), with a regression test.

### Nits
- **`even_spacing` reference step.** The comment claimed it mirrors `utils._warn_if_irregular_spacing`, but that helper compares each step to the averaged `span/(n-1)` resolution while this used `diffs[0]`. Fixed to use the averaged step so behavior matches the cited helper.

### What looks good
- Checks are structural only (dims/coords/dtype/attrs) and never read the data buffer; verified dask stays lazy and cupy stays on device. Backend tests run unskipped on the dev box.
- CRS resolution reuses the polygonize order. Severity split is clean and `is_valid` keys off errors only.

### Checklist
- [x] All implemented backends produce consistent results (structural; tested across all four)
- [x] NaN handling correct (`isfinite` guards, nan-safe min/max)
- [x] Edge cases covered (1D/4D, warnings-only, non-finite cell size, bare dims)

brendancol (commented):

Re-reviewed the two changed helpers in `validate.py`:

- `_check_geographic_range` now gates on `dim in agg.coords`, so a bare lat/lon dimension no longer reads xarray's synthesized integer index. Covered by `test_bare_latlon_dim_does_not_trigger_geographic_warning`.
- `even_spacing` compares each step to the averaged `span/(n-1)` resolution, matching `utils._warn_if_irregular_spacing`.

No new findings. Full suite: 117 passed (accessor + contract-validate), flake8 clean. Backend tests (cupy, dask+cupy) ran unskipped.

brendancol:
Heads up on the red `run (..., 3.14)` checks: they are not caused by this PR. The only real failure is `xrspatial/tests/test_perlin.py::test_perlin_drops_input_coords`, which is broken on `main` (the macOS/Windows/3.14t jobs show "operation canceled" from matrix fail-fast). It is a merge race between perlin PRs #3470 and #3472, fixed in #3488. Every test in this PR passes on every platform it ran on (`test_contract_validate.py` is green on ubuntu and windows fast lanes). Once #3488 lands on `main`, I will re-merge `main` into this branch and the checks will go green.

PR #3487, opened 2026-06-24
Add from_template(): empty study-area DataArrays for common regions (#3484)

### LINKED ISSUE #3484: Add from_template(): empty study-area DataArrays for common regions
## Reason or Problem

Every analysis in xarray-spatial starts from an existing raster. You read one with `open_geotiff`, synthesize one with `generate_terrain`/`make_terrain`, or build a `DataArray` by hand. There is no fast way to get an empty grid for a well-known study area such as the contiguous United States in an Albers Equal-Area projection, a metro region, or a country. Someone who just wants a canvas for a new analysis has to look up the bounding box, pick a CRS, work out the cell count, and assemble the coordinates by hand.

## Proposal

Add a `templates` module with a `from_template(name)` helper, exported at the top level so users can write `from xrspatial import from_template`.

**Design:**

`from_template(name, resolution=None, backend="numpy")` resolves a string name to a study area and returns an empty (NaN-filled) `DataArray` that obeys the xarray-spatial array contract: 2-D with `dims=["y", "x"]`, pixel-center 1-D `y`/`x` coordinates (north-up, descending `y`), and `attrs` carrying `res` and `crs`. The grid covers the study area's rectangular bounding box.

Names resolve in two layers:

- A curated registry of regions, each with a projected CRS and default resolution. `conus` returns a grid in Albers Equal-Area (EPSG:5070); `nyc` returns a metro grid in a local projected CRS.
- An embedded static table mapping ISO-3166 / GADM alpha-3 country codes (e.g. `FRA`, `USA`) to a bounding box, returned in EPSG:4326. No network access and no geopandas dependency.

The caller controls grid size with `resolution` (cell size in CRS units); each template carries a default so a bare `from_template("conus")` works. Coordinate construction reuses the existing `_make_output_coords` helper in the reproject module. All four array backends are supported (numpy, dask+numpy, cupy, dask+cupy) since the data is just a NaN fill.

**Usage:**

```python

agg = from_template("conus")                    # CONUS in Albers, default resolution
agg = from_template("conus", resolution=1000)   # 1 km cells
agg = from_template("nyc", resolution=30)       # 30 m metro grid
agg = from_template("FRA")                       # France bbox in EPSG:4326
```

The result feeds straight into existing tools like `slope`, `hillshade`, and `rasterize`.

**Value:** Lowers the barrier to starting an analysis. You get a projected, contract-valid grid for a named area in one call instead of hand-rolling coordinates and CRS metadata.

## Stakeholders and Impacts

Anyone starting a new analysis from a known region. The change is additive: a new module, a new top-level export, and supporting data. No existing functions change.

## Drawbacks

The curated region list and country bbox table are static and need occasional curation. Bounding-box grids include area outside the true region boundary; masking to the actual shape is out of scope and would need vector geometries.

## Alternatives

Fetch real boundaries from Natural Earth or GADM at runtime via geopandas and mask to the region shape. Rejected for the first version because it adds a network call and an optional dependency. The embedded bbox approach is deterministic and dependency-free.

## Unresolved Questions

Which regions belong in the curated registry beyond `conus` and `nyc`.

brendancol:
Closes #3484

Adds a `templates` module with a `from_template(name)` helper, exported at the top level (`from xrspatial import from_template`). It returns an empty (NaN-filled, float32) DataArray for a named study area that follows the xarray-spatial array contract: 2-D `['y', 'x']`, pixel-center coordinates (north-up), and `attrs` carrying `res` and `crs`. The grid covers the area's rectangular bounding box and is a starting canvas for a new analysis.

- Curated regions resolve to a projected CRS: `conus` (Albers EPSG:5070), `nyc`, `alaska`, `hawaii`, `california`, `europe`, `world`.
- ISO-3166 / GADM alpha-3 country codes (240 of them) resolve to an EPSG:4326 bounding-box grid, e.g. `from_template("FRA")`. The country bounding boxes come from Natural Earth 50m admin_0 (public domain) and are committed as a static table, so there is no runtime network access or geopandas dependency. Countries that cross the antimeridian (Russia, Fiji) keep contiguous bounding boxes.
- `resolution` sets cell size in the template's CRS units, with a per-template default. An over-fine resolution is rejected before anything is allocated.

Coordinate construction reuses the existing `_make_output_coords` helper from the reproject module instead of re-deriving the math.

Backends: numpy, dask+numpy, cupy, dask+cupy. The data is a NaN fill, so each backend allocates its own array type and coords/attrs match across all four.

## Test plan
- [x] Contract: dims, coords, res tuple, crs int, units
- [x] Curated regions (conus Albers, nyc, case-insensitive) and country codes (EPSG:4326)
- [x] Resolution control, fill/dtype, error cases (unknown name, non-positive resolution, over-fine resolution, non-string, bad backend)
- [x] All four backends with coord/attr parity
- [x] Downstream smoke: `slope` accepts a template
- [x] User guide notebook executes end to end
- [x] flake8 clean on new files


brendancol (commented):
## PR Review: Add from_template() study-area grids

Reviewed from the PR head worktree. The change is additive (new module, data table, export, tests, docs, notebook) and does not touch existing code paths.

### Blockers
None.

### Suggestions
- [ ] `xrspatial/templates.py`, `_make_data` dask+cupy branch: `da.full(...).map_blocks(cupy.asarray)` has no `meta=`, so dask infers meta by running `cupy.asarray` on a probe block while building the graph (an eager GPU call at graph-build time). Pass `meta=cupy.empty((0, 0), dtype="float32")` to skip the probe.

### Nits
- [ ] `xrspatial/templates.py`, `_MAX_CELLS` guard: the cap is right for numpy and cupy, which allocate eagerly, but it also rejects large *lazy* dask grids that would never materialize all at once. Fine as a conservative default; a one-line note on the constant saying it is deliberately backend-agnostic would save the next reader a question.
- [ ] `xrspatial/tests/test_templates.py`, `test_downstream_slope_accepts_template`: running `slope` on an all-NaN grid prints `All-NaN slice` RuntimeWarnings. Harmless (it is slope's own behavior on empty input), but wrapping the call in `np.errstate`/`warnings.catch_warnings` would keep the test output clean.

### What looks good
- Antimeridian handling: country boxes for the wrap cases (USA, Russia, Fiji) stay contiguous instead of collapsing to a full -180..180 span. Verified against Natural Earth.
- Coordinate construction reuses `_make_output_coords` rather than re-deriving the linspace math, so pixel-center/north-up behavior matches the reproject path.
- Error coverage is thorough: unknown name, non-positive resolution, over-fine resolution, non-string name, and bad backend each have a test and a specific message.
- Cross-backend parity (numpy/dask/cupy/dask+cupy) checks coords and attrs, not just that the call returns.
- The notebook executes end to end and the country grids are honestly documented as full-territory bounding boxes, not masks.

### Checklist
- [x] Output obeys the array contract (dims, coords, res, crs)
- [x] All four backends produce consistent coords/attrs
- [x] NaN fill is correct; dtype is float32
- [x] Edge cases covered (errors, resolution bounds, fill)
- [x] No premature materialization; data stays lazy on dask
- [ ] No benchmark added (acceptable: the helper is a single allocation)

brendancol (commented):
## Follow-up review (commit f1e7deb0)

All three findings from the previous pass are addressed:

- Suggestion (dask+cupy meta): `_make_data` now passes `meta=cupy.empty((0, 0), dtype="float32")` to `map_blocks`, so the graph builds without the eager probe call. Verified `agg.data._meta` is a cupy array.
- Nit (_MAX_CELLS): added a comment noting the cap is applied uniformly across backends, dask included, on purpose.
- Nit (test warnings): `test_downstream_slope_accepts_template` now wraps the `slope` call in `np.errstate` + `catch_warnings`, so the suite runs clean (22 passed, no warnings).

No new issues. flake8 clean on the changed files. Nothing else outstanding.

brendancol (commented):
## Review: `preserve=` extension (EPSG-coded projection by property)

Reviewed the `preserve` commits on top of the base feature. Additive: a new keyword-only arg plus supporting data, no change to the default code path.

### Blockers
None.

### Suggestions
None outstanding. One issue found during review and fixed in 73624937: `from_template("world", preserve="shape")` produced a degenerate 80x970 grid at ~500 km cells, because World Mercator's y-extent diverges toward the poles and inflated the default resolution. Clamping the world template's preserve-path latitude band to +/-85 (the conventional Web Mercator limit) yields a sensible 797x802 grid; the default (non-preserve) world grid still spans the full +/-90 in EPSG:4326.

### Nits
- `xrspatial/templates.py` `_default_preserve_resolution`: snaps up to the next {1,2,5}x10^k, so a 5.86 Mm extent lands on 10 km cells (~586 cells) rather than ~1000. Intentional and documented as "~1000 cells"; fine to leave.
- Wide or overseas-spanning country codes (USA, RUS, FRA-with-overseas) get a single UTM zone for `shape` or a Greenwich-centered Equal Earth for `area`, both of which distort across the extent. This is inherent to EPSG-coded global/zonal projections and is called out in the notebook's info box; curated regions remain the better canvas.

### What looks good
- `attrs['crs']` stays an int EPSG for every `preserve` result, consistent with the base feature and the array contract.
- Reuses `_edge_samples` + `_transform_boundary` to project the lon/lat bbox and `pyproj.database.query_utm_crs_info` for the zone, rather than reimplementing projection math.
- Property correctness is asserted, not assumed: tests check `pyproj.CRS(...).to_dict()['proj']` is the equal-area / conformal family, plus exact codes (conus 5070, JPN UTM 32653, Europe LCC 3034, Equal Earth 8857 fallback).
- Hemisphere and polar handling verified: southern centroids -> 327xx, `|lat|>84` -> UPS 5041/5042.
- `distance`/`direction` are rejected with a message explaining no EPSG-coded equidistant/azimuthal projection exists, rather than silently synthesizing one.
- Antimeridian country codes (USA, RUS, FJI) build under both properties.
- Backends (numpy/dask/cupy/dask+cupy) keep coord/attr parity with `preserve` set; notebook re-executes end to end.

### Checklist
- [x] EPSG resolution correct per property (area/shape), verified by projection family + exact codes
- [x] CRS stored as int EPSG; units "m"
- [x] Hemisphere / polar / antimeridian paths exercised
- [x] Invalid values (distance/direction/bogus) raise a clear error
- [x] Backends consistent with preserve set
- [ ] No benchmark (acceptable: a single allocation + one pyproj bbox transform, not a hot loop)
- [x] README + docstring + notebook updated

PR #3488, opened 2026-06-24
Fix stale perlin test that pins dropped-coords behavior (main is red)

brendancol:
`main` is currently red on the `run (..., 3.14)` checks. The matrix runs fail-fast, so when the ubuntu 3.14 job fails the macOS/Windows/3.14t jobs report as canceled, but there is only one real failure:

```
FAILED xrspatial/tests/test_perlin.py::test_perlin_drops_input_coords
AssertionError: assert 'lat' not in Coordinates: ...
```

This is a merge race between two perlin PRs:

- #3472 added `test_perlin_drops_input_coords`, pinning the old behavior where `perlin()` dropped input coordinates. Its comment says to "Flip this to assert the coords ARE preserved once that is fixed."
- #3470 then changed `perlin()` to preserve the caller's x/y coordinates, but did not update the pinning test.

They merged in an order that left the test asserting behavior the code no longer has.

This flips the test to match the shipped behavior: it now asserts the coordinates are preserved and equal to the input, and is renamed `test_perlin_preserves_input_coords`. No source changes; `perlin()` already does the right thing.

Test plan:
- [x] `pytest xrspatial/tests/test_perlin.py` — 31 passed

### DIFF
--- a/xrspatial/tests/test_perlin.py
+++ b/xrspatial/tests/test_perlin.py
@@ -319,17 +319,16 @@ def test_perlin_preserves_dims_and_attrs():
     assert out.attrs == src.attrs
 
 
-def test_perlin_drops_input_coords():
-    # Pins current behavior: perlin() rebuilds the output from dims + attrs but
-    # not coords, so input coordinate variables are dropped.  This is a known
-    # bug -- see the perlin "drops input coordinates from output" issue.  Flip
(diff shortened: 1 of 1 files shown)


brendancol:
Superseded by #3489, which already fixed the same stale perlin coords test on main (parallel duplicate). Closing as redundant.

PR #3493, opened 2026-06-25
Add coregister=True to .xrs.rasterize accessor (#3492)

### LINKED ISSUE #3492: Add coregister=True to .xrs.rasterize to reproject GeoDataFrame geometries onto the caller grid
## Reason or Problem

The interpolation accessors `.xrs.kde`/`idw`/`kriging`/`spline` gained `coregister=True` in #3481. When the input is a GeoDataFrame, that flag reprojects its geometries from their own CRS into the caller raster's CRS before computing, so the surface lands on the caller grid even when the vector data is in a different projection.

`.xrs.rasterize` has no equivalent. Rasterizing a GeoDataFrame whose CRS differs from the caller raster either raises (the default `check_crs=True`), or, with `check_crs=False`, burns the geometries onto the grid without reprojecting them, so the burned pixels land at the wrong coordinates. Either way the user has to reproject the GeoDataFrame by hand before calling `.xrs.rasterize`. That is the manual step #3481 removed for interpolation.

## Proposal

Add `coregister=False` to `.xrs.rasterize` on both accessor classes (DataArray and Dataset), mirroring the interpolation accessor. With `coregister=True` and a GeoDataFrame input, reproject the geometries into the caller raster's CRS via `GeoDataFrame.to_crs` before rasterizing. It needs a CRS on both sides (the GeoDataFrame's `.crs` and the caller's detected CRS) and raises a clear error otherwise. The default keeps current behavior, and the standalone `rasterize()` function stays unchanged. `coregister` is accessor-only, the same split #3481 used for interpolation.

Target call:

```python
out = raster.xrs.rasterize(points_gdf, coregister=True)
```

brendancol:
Closes #3492

Gives `.xrs.rasterize` the same `coregister` flag the interpolation accessors got in #3481.

- `coregister=False` (default) keeps current behavior: a GeoDataFrame whose CRS differs from the caller raster still raises through rasterize's own `check_crs`.
- `coregister=True` reprojects a GeoDataFrame's geometries from their CRS into the caller raster's CRS (via `GeoDataFrame.to_crs`) before rasterizing, so the burn lands on the caller grid. It needs a CRS on both the geometries (`.crs`) and the caller (detected the same way `check_crs` detects the template CRS), and raises a clear error otherwise.
- Added on both accessor classes (DataArray and Dataset) through a shared `_rasterize_on_accessor` helper that mirrors `_interp_on_accessor`.
- The standalone `rasterize()` function is untouched; `coregister` is accessor-only, the same split #3481 used.

### Backends

Input-parsing only: the coregister branch runs before backend dispatch and just reprojects the GeoDataFrame, then hands off to the unchanged `rasterize`. numpy and dask+numpy verified directly; cupy / dask+cupy inherit the unchanged rasterize dispatch.

### Test plan

- [x] coregister round-trip (4326 points onto a 3857 grid) burns the same pixels as rasterizing the native-CRS points (numpy + dask)
- [x] coregister output inherits the caller grid; five points burn five finite cells
- [x] Dataset accessor coregister matches the DataArray path
- [x] CRS mismatch raises without coregister; matching CRS passes
- [x] coregister with a missing CRS on either side raises
- [x] coregister with a non-GeoDataFrame input raises
- [x] existing `test_accessor.py` / `test_rasterize.py` / `test_interpolate_vector.py` pass unchanged

### DIFF
--- a/docs/source/reference/utilities.rst
+++ b/docs/source/reference/utilities.rst
@@ -32,6 +32,17 @@ Rasterize
 
     xrspatial.rasterize.rasterize
 
+.. note::
+
+   ``rasterize`` is also on the ``.xrs`` accessor, where the caller raster
+   supplies the output grid, chunks, and CRS (the ``like`` template)::
+
+       dem.xrs.rasterize(geometries_gdf, column="value")
+       dem.xrs.rasterize(geometries_gdf, column="value", coregister=True)
+
+   ``coregister=True`` reprojects a GeoDataFrame's geometries from their CRS
+   into the caller's CRS before rasterizing.
+
 Contours
 ========
 .. autosummary::
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -829,16 +829,22 @@ def _rasterize_on_accessor(obj, geometries, *, coregister, **kwargs):
     from xrspatial.interpolate._vector import is_geodataframe
 
     if coregister:
+        # is_geodataframe matches the interpolation accessor (#3481): plain
+        # geopandas only, not dask_geopandas, even though standalone
+        # rasterize._parse_input accepts dask_geopandas.  Reprojecting a
+        # dask_geopandas frame under coregister is left for a follow-up.
         if not is_geodataframe(geometries):
             raise ValueError(
                 "coregister=True requires a GeoDataFrame input carrying a CRS"
             )
         src = _to_pyproj_crs(geometries.crs)
+        # _like_crs detects the caller CRS the same way rasterize's check_crs
+        # does, so the reprojected frame compares equal to the template.
         target = _to_pyproj_crs(_like_crs(obj))
         if src is None or target is None:
             raise ValueError(
                 "coregister=True needs a CRS on both the geometries "
-                "(GeoDataFrame.crs) and the caller (attrs['crs'])"
+                "(GeoDataFrame.crs) and the caller (its detected raster CRS)"
             )
         geometries = geometries.to_crs(target)
     return rasterize(geometries, like=obj, **kwargs)
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -814,6 +814,36 @@ def _interp_on_accessor(obj, func, points, rest, *, coregister, **kwargs):
     return func(points, *rest, template=obj, **kwargs)
 
 
+def _rasterize_on_accessor(obj, geometries, *, coregister, **kwargs):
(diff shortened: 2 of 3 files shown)


brendancol (commented):
## PR Review: Add coregister=True to .xrs.rasterize accessor

None.

- `dask_geopandas` gap: `is_geodataframe` recognizes plain geopandas only, so a `dask_geopandas` frame with `coregister=True` raises "requires a GeoDataFrame", even though standalone `rasterize._parse_input` accepts one. This mirrors the interpolation accessor (#3481) on purpose. Addressed by a code comment so the gap is intentional and discoverable; reprojecting a dask_geopandas frame under coregister can be a follow-up.

- The missing-CRS error message said `(attrs['crs'])`, but `_like_crs` also reads `crs_wkt`, a `spatial_ref` coord, or `rio.crs`. Reworded to "(its detected raster CRS)".

### What looks good
- Using `_like_crs(obj)` for the reprojection target (rather than only `attrs['crs']`) matches how rasterize's own `check_crs` finds the template CRS, so the reprojected frame compares equal and the default guard passes cleanly.
- `coregister` is accessor-only; the standalone `rasterize()` signature is untouched, the same split #3481 used for interpolation.
- Shared `_rasterize_on_accessor` helper keeps the DataArray and Dataset accessors in sync.

### Checklist
- [x] Reprojection path defers to the unchanged rasterize (algorithm unaffected)
- [x] Backend dispatch unchanged; coregister runs before dispatch (numpy/dask verified, cupy inherits)
- [x] NaN/fill handled by rasterize as before
- [x] Edge cases covered: missing CRS, mismatch, non-GeoDataFrame, Dataset accessor
- [x] No premature materialization; dask template stays lazy until output
- [x] Benchmark not needed (input-parsing glue, no new kernel)
- [x] Docstrings present on both accessor methods and the helper

brendancol (commented):
## Follow-up review (after 3a6ae84c)

Re-read the full changed files at head. The two findings from the first pass are resolved:

- **Suggestion (dask_geopandas gap)** — fixed: added a code comment in `_rasterize_on_accessor` marking the geopandas-only gate as intentional parity with #3481, with the dask_geopandas case noted as a follow-up.
- **Nit (error message)** — fixed: the missing-CRS message now reads "(its detected raster CRS)" instead of "(attrs['crs'])", matching `_like_crs`'s broader detection.

No new issues. Tests still green (8 passed locally). Nothing blocking.

PR #3495, opened 2026-06-25
Honor requested resolution exactly in from_template() (#3494)

### LINKED ISSUE #3494: from_template() should honor the requested resolution exactly
## Reason or Problem
`from_template(name, resolution=r)` does not honor the requested resolution exactly. It keeps the template's bounding box fixed, divides it by the resolution, rounds the cell count to an integer, then recomputes the realized cell size from the fixed box. When the box extent is not an exact multiple of the resolution, the returned `res` drifts.

Example:

    >>> from xrspatial import from_template
    >>> from_template('nyc', resolution=10).attrs['res']
    (10.0, 10.000757432304487)

The NYC bbox is 55510 m wide and 52814 m tall. 55510 divides evenly by 10, so res_x stays 10. 52814 does not, so res_y comes back as 10.000757. A user who asked for 10 m cells expects (10.0, 10.0).

## Proposal
Honor the requested resolution exactly. Hold the lower-left corner fixed and nudge the far edges (right, top) out to the nearest exact multiple of the cell size, the same way the reproject grid path already does in `_compute_output_grid`. The grid extent shifts by less than half a cell while `res` matches the request.

**Design:**
In `from_template`, after computing width/height with round(), set `right = left + width * res_x` and `top = bottom + height * res_y`, then build the coordinates and the `res` attr from the requested resolution. This mirrors `xrspatial/reproject/_grid.py` lines 457-459.

**Usage:**

    >>> from_template('nyc', resolution=10).attrs['res']
    (10.0, 10.0)

**Value:** Resolution becomes a guarantee rather than advisory. Grids from `from_template` line up with other rasters at the same resolution, which matters for stacking, alignment, and tiling.

## Drawbacks
The grid no longer reaches the registry bbox's exact stated far edge; right/top move by up to half a cell. Pixel centers still fall inside the original bbox.

## Alternatives
Leave the behavior as is and document the drift. Rejected because it makes resolution advisory rather than a guarantee.

brendancol:
## Summary

`from_template(name, resolution=r)` now honors the requested resolution exactly. The previous code held the template's bbox fixed and recomputed the realized cell size from it, so `res` drifted when the extent was not a whole number of cells (e.g. `from_template('nyc', resolution=10)` returned `res=(10.0, 10.000757)`).

- Anchor the lower-left corner and nudge the far edges (right, top) to an exact multiple of the cell size, so `attrs['res']` matches the requested resolution.
- This matches the reproject grid path (`_compute_output_grid` in `reproject/_grid.py`).
- Pixel centers still fall inside the original registry bbox; the far edges move by less than half a cell.

Closes #3494

## Backends

Backend-agnostic: the change is in the shared grid math before the array is allocated, so numpy, cupy, dask+numpy, and dask+cupy all get the same exact `res`.

## Test plan

- [x] `from_template('nyc', resolution=10).attrs['res'] == (10.0, 10.0)`
- [x] tuple resolution honored exactly
- [x] coord spacing equals the requested resolution
- [x] pixel centers stay within the registry bbox
- [x] preserve path honors resolution exactly
- [x] full test_templates.py suite passes (44 tests)

### DIFF
--- a/xrspatial/templates.py
+++ b/xrspatial/templates.py
@@ -182,6 +182,12 @@ def from_template(name, resolution=None, *, preserve=None, backend="numpy",
     (north-up, descending ``y``) and ``res``/``crs`` attributes. It covers the
     study area's rectangular bounding box and is meant as a starting canvas.
 
+    The requested ``resolution`` is honored exactly. The study-area box is
+    rarely a whole number of cells wide, so the far edges (right, top) are
+    nudged out by up to half a cell to land on an exact multiple of the cell
+    size, anchoring the lower-left corner. ``attrs['res']`` therefore matches
+    the ``resolution`` you pass.
+
     Parameters
     ----------
     name : str
@@ -268,10 +274,14 @@ def from_template(name, resolution=None, *, preserve=None, backend="numpy",
             f"Use a coarser resolution."
         )
 
-    ys, xs = _make_output_coords(bounds, (height, width))
-    # Realized cell size from the integer grid (may differ slightly from input).
-    actual_res_x = (right - left) / width
-    actual_res_y = (top - bottom) / height
+    # Honor the requested resolution exactly: anchor the lower-left corner and
+    # nudge the far edges to an exact multiple of the cell size, so res comes
+    # back as (res_x, res_y) instead of drifting when the bbox extent isn't a
+    # whole number of cells. Mirrors the reproject grid path
+    # (_compute_output_grid in reproject/_grid.py).
+    right = left + width * res_x
+    top = bottom + height * res_y
+    ys, xs = _make_output_coords((left, bottom, right, top), (height, width))
 
     data = _make_data((height, width), fill, backend, chunks)
     unit = "degree" if crs == 4326 else "m"
@@ -281,7 +291,7 @@ def from_template(name, resolution=None, *, preserve=None, backend="numpy",
         name=key,
         coords={"y": ys, "x": xs},
         dims=["y", "x"],
-        attrs={"res": (actual_res_x, actual_res_y), "crs": crs},
+        attrs={"res": (res_x, res_y), "crs": crs},
     )
     template["x"].attrs["units"] = unit
     template["y"].attrs["units"] = unit
--- a/xrspatial/tests/test_templates.py
+++ b/xrspatial/tests/test_templates.py
@@ -82,6 +82,34 @@ def test_resolution_tuple():
     assert rx > ry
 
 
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Honor requested resolution exactly in from_template() (#3494)

None.

- The new exact-resolution tests only cover projected-CRS regions (nyc, conus) and the preserve path. Country codes come back in EPSG:4326 and run through the same nudge math, but nothing pins exact `res` for a 4326 template. A one-liner like `from_template('FRA', resolution=0.25).attrs['res'] == (0.25, 0.25)` would close that gap. (xrspatial/tests/test_templates.py)

- The dask backend's exact `res` is covered indirectly by `test_dask_numpy_backend` (it asserts `attrs == numpy ref`). An explicit assertion would document the intent, but it's optional. (xrspatial/tests/test_templates.py:139)

### What looks good
- The fix mirrors the existing reproject grid path (`_compute_output_grid` in `reproject/_grid.py` lines 457-459), so the two paths now agree on how resolution is honored.
- The change sits in the shared grid math before `_make_data`, so all four backends get identical coords and `res` with no per-backend code.
- Pixel centers stay inside the original registry bbox: the center sits res/2 inward, which absorbs the half-cell nudge. Checked both axes.
- I checked the EPSG:4326 country path for latitude/longitude overflow from the nudge. Latitude never gets near 90 (the northernmost bbox top is ~83.6 for GRL), and the lon>180 values come from pre-existing antimeridian-wrapped bboxes, not from the nudge. No new invalid-coordinate risk.
- Docstring updated to describe the nudge and the lower-left anchor; autosummary picks it up.

### Checklist
- [x] Algorithm matches the reproject grid convention
- [x] All four backends produce consistent results (shared pre-allocation math)
- [x] NaN handling unchanged (fill path untouched)
- [x] Edge cases covered (tuple res, centers within bbox)
- [x] No dask chunk concern (change is before array allocation)
- [x] No benchmark needed (constant-time arithmetic)
- [x] No README change needed (no new function, no backend change)

brendancol (commented):
## Follow-up review (after 0dc8a4ef)

- **Suggestion (4326 exact-res test):** fixed. Added `test_country_resolution_honored_exactly`, asserting `from_template('FRA', resolution=0.25).attrs['res'] == (0.25, 0.25)`.
- **Nit (explicit dask exact-res):** fixed. `test_dask_numpy_backend` now asserts the dask+numpy path returns `res == (10.0, 10.0)`.

Full `test_templates.py` suite passes (45 tests). No new findings.

PR #3505, opened 2026-06-25
Document num_sample=None in natural_breaks docstring

### LINKED ISSUE #3501: natural_breaks num_sample docstring omits None, unlike the other classifiers
## Summary

In `xrspatial/classify.py`, `natural_breaks` documents `num_sample` as:

```
num_sample : int, default=20000
```

The signature is `num_sample: Optional[int] = 20000`, and `None` is a real value that means "use all data". The other three samplers in the module document this:

- `quantile`: `int or None, default=20000`, with a note that `None` means use all data
- `maximum_breaks`: same
- `percentiles`: same

`natural_breaks` doesn't mention `None` at all, so someone reading its docstring wouldn't know they can pass it to fit on the full array.

## Verification

`natural_breaks(agg, k=5, num_sample=None)` runs on both the numpy and cupy backends and classifies using all finite data.

## Scope

Docstring only. No signature or behavior change, not breaking. Match the `num_sample` wording to the other three classifiers (`int or None, default=20000`, plus the note that `None` uses all data).

Found by the api-consistency sweep (Cat 3, MEDIUM).

brendancol:
Closes #3501.

`natural_breaks` already accepts `num_sample=None` (use all data) the same
way `quantile`, `maximum_breaks`, and `percentiles` do, but its docstring
only described `num_sample` as `int` and never mentioned `None`. This brings
the wording in line with the other three classifiers.

Docstring only. No signature or behavior change.

Verified `natural_breaks(agg, k=5, num_sample=None)` on both the numpy and
cupy backends. `pytest xrspatial/tests/test_classify.py -k natural_breaks`
passes (22 passed).

### DIFF
--- a/xrspatial/classify.py
+++ b/xrspatial/classify.py
@@ -848,8 +848,11 @@ def natural_breaks(agg: xr.DataArray,
         where n is the total number of data points, i.e: `agg.size`
         When n is large, we should fit the model on a small sub-sample
         of the data instead of using the whole dataset.
-        ``None`` means use all data (safe for numpy/cupy,
-        automatically capped for dask).
+        ``None`` means fit on all data instead of a sub-sample. That is
+        the full O(n²) case described above, so it may be slow and raises
+        ``MemoryError`` if the Jenks matrices would exceed half of the
+        available RAM. For dask the full sample is drawn lazily via
+        indexed access.
     name : str, default='natural_breaks'
         Name of output aggregate.
 
--- a/xrspatial/classify.py
+++ b/xrspatial/classify.py
@@ -842,12 +842,14 @@ def natural_breaks(agg: xr.DataArray,
         of values to be reclassified.
     k : int, default=5
         Number of classes to be produced.
-    num_sample : int, default=20000
+    num_sample : int or None, default=20000
         Number of sample data points used to fit the model.
         Natural Breaks (Jenks) classification is indeed O(n²) complexity,
         where n is the total number of data points, i.e: `agg.size`
         When n is large, we should fit the model on a small sub-sample
         of the data instead of using the whole dataset.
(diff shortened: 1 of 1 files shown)


brendancol:
## Review: document num_sample=None in natural_breaks

Read the full `natural_breaks` / `_compute_natural_break_bins` / `_run_dask_natural_break` paths, not just the diff. Documenting that `None` is accepted is the right call and matches `quantile`, `maximum_breaks`, and `percentiles`. One wording issue.

### Suggestion

`xrspatial/classify.py:851-852` -- the new parenthetical "(safe for numpy/cupy, automatically capped for dask)" reads like it was copied from `quantile`/`percentiles`, where `None` just runs an exact percentile over the finite values. That clause doesn't describe `natural_breaks`:

- On numpy/cupy, `None` makes `_compute_natural_break_bins` fit Jenks on the whole array. That's the O(n^2) path this same docstring warns about two lines up ("When n is large, we should fit the model on a small sub-sample").
- `_compute_natural_break_bins` (line 684) raises `MemoryError` when the two Jenks matrices would exceed half of available RAM. So `None` on a large raster is the opposite of "safe" -- it's the case most likely to blow up.
- For dask it isn't capped to a small sample either. `_run_dask_natural_break` sets `num_sample = num_data` and fits Jenks on all the data via indexed access.

Suggest rewording so it says `None` fits on all data, and that this is the expensive O(n^2) case guarded by a memory check, instead of calling it "safe".

### What looks good

- Documents the `None` option, in line with the other three samplers.
- Docstring-only, non-breaking. `num_sample=None` verified on numpy and cupy.

### Checklist
- [x] No code change; behavior unchanged
- [ ] Docstring accurately describes `None` for the Jenks case (see suggestion)
- [x] API consistency improved
- [x] No tests needed (docstring only)

Issue #3508, opened 2026-06-25
binary(): numpy/dask return input dtype while cupy returns float32


brendancol:
`binary()` returns a different output dtype depending on the array backend. The numpy and dask+numpy paths return the input dtype; the cupy and dask+cupy paths always return `float32`. Switching backends in a pipeline silently changes the container dtype for the same values.

## Reproduce

```python
import numpy as np, xarray as xr, dask.array as da, cupy
from xrspatial.classify import binary

data = np.array([[np.nan, 1., 2.], [3., 4., np.inf]], dtype=np.float64)

print(binary(xr.DataArray(data), [1, 2]).dtype)                       # float64
print(binary(xr.DataArray(da.from_array(data)), [1, 2]).dtype)        # float64
print(binary(xr.DataArray(cupy.asarray(data)), [1, 2]).data.dtype)    # float32
```

## Cause

`_cpu_binary` allocates its output with `dtype=data.dtype`, while `_run_cupy_binary` uses `dtype='f4'`.

## Why it matters

The docstring example shows `float32` output, and every other classifier in `classify.py` (`reclassify`, `quantile`, `natural_breaks`, and so on) emits `float32` on all four backends. `binary` is the only one that doesn't. For integer input the numpy path returns an integer dtype, which can't hold the `NaN` sentinel the docstring promises for non-finite cells.

## Fix

Allocate the numpy output as `float32` so the four backends agree and the NaN sentinel is always representable.

Found by the metadata-propagation sweep on `classify`.

brendancol:
Resolved on `main`. The float32 fix landed across two merged PRs:

- #3514 — `_cpu_binary` now allocates `dtype=np.float32`, so the numpy and dask+numpy paths match the cupy backends and the NaN sentinel is always representable.
- #3513 — types the dask+cupy `map_blocks` meta to `f4` so the lazy dtype matches the computed result.

All four backends now return `float32`. Covered by `test_binary_output_dtype_float32` and `test_binary_dask_output_dtype_float32` in `xrspatial/tests/test_classify.py`, both passing.

Closing as completed.

PR #3513, opened 2026-06-25
binary(): type the dask+cupy meta so the lazy dtype matches float32 (#3508)

brendancol:
Part of #3508. Rebased onto current main after the duplicate accuracy-sweep PR #3514 merged.

The `_cpu_binary` float32 fix and the `verify_dtype=True` backend tests are already on main via #3514, so this PR is now scoped to the one piece that didn't land there: the dask+cupy lazy dtype.

## What this fixes

`_run_dask_cupy_binary` passed `meta=cupy.array(())`, which defaults to float64. So the lazy dask+cupy array advertised float64 while `_run_cupy_binary` computes float32. The merged tests only check the computed dtype because `general_output_checks` computes the array before asserting, so nothing caught the lazy mismatch.

Probe before the fix:

```
dask+cupy : lazy dtype = float64 | computed dtype = float32
```

This is the same advertised-vs-computed mismatch class as aspect #2682 and focal #3217.

## Change

- `_run_dask_cupy_binary` now passes `meta=cupy.array((), dtype='f4')`.
- `test_binary_dask_cupy` asserts the lazy dtype is float32 so it stays pinned (the existing dtype checks all compute first).

After the fix, dask+cupy reports float32 both lazy and computed. Full `test_classify.py` suite passes on a CUDA host with the GPU backends running live.

## Follow-up (out of scope)

The sibling classifiers' dask+cupy helpers (`_run_dask_cupy_bin` and friends) use the same untyped `meta=cupy.array(())` and likely share the same latent lazy-dtype mismatch. Left for a separate change to keep this PR focused on `binary`.


brendancol:
## Self-review (domain-aware)

I ran the review-pr checks over this branch and found one real gap, now fixed in 2db65f88.

### Fixed during review

The eager dtype fix was correct, but it left the dask+cupy path half done. `_run_dask_cupy_binary` passed `meta=cupy.array(())`, which defaults to float64, so the lazy dask array advertised float64 while `_run_cupy_binary` computes float32. The four backend tests didn't catch it because `general_output_checks` computes the array before asserting the dtype, so it only ever checked the computed dtype.

Probe before the fix:

```
dask+numpy : lazy dtype = float32 | computed dtype = float32
dask+cupy  : lazy dtype = float64 | computed dtype = float32
```

This is the same advertised-vs-computed mismatch class as aspect #2682 and focal #3217. Fix: pass `meta=cupy.array((), dtype='f4')`, and assert the lazy dtype in `test_binary_dask_cupy` so it stays pinned. After the fix both paths report float32 lazy and computed. Full classify suite: 103 passed, GPU backends run live.

The dask+numpy path was already fine because `map_blocks` runs the chunk function to infer its meta, and that now returns float32.

### What looks good

- Root cause is right: `_cpu_binary` allocated `dtype=data.dtype`. float32 matches the docstring and every sibling classifier (`reclassify`, `quantile`, and the rest go through `_cpu_bin`, which is already float32).
- `verify_dtype=True` added to all four binary backend tests.
- The integer-input regression test covers the case that previously returned int32 and couldn't hold the NaN sentinel.

### Out of scope (follow-up)

The sibling classifiers' dask+cupy helpers (`_run_dask_cupy_bin` and friends) use the same untyped `meta=cupy.array(())` and likely share the same lazy-dtype mismatch. I left them alone to keep this PR focused on `binary`, matching how the aspect audit deferred slope/curvature.

### Note on the duplicate

PR #3514 (accuracy sweep) fixes the same `binary()` dtype bug. I'm not touching either PR; flagging it here so a maintainer can reconcile the two. This PR also carries the dask+cupy lazy-meta fix and the lazy-dtype assertion, which #3514 may not.

brendancol:
Rebased onto current main now that the duplicate accuracy-sweep PR #3514 has merged. Main already carries the `_cpu_binary` float32 fix and the `verify_dtype=True` backend tests, so I dropped those from this branch (they were identical to what merged) and dropped my integer-input test since #3514's `test_binary_output_dtype_float32` already covers float64/float32/int32.

This PR is now scoped to the only piece that wasn't in #3514: typing the dask+cupy `map_blocks` meta as `cupy.array((), dtype='f4')` so the lazy array advertises float32 instead of float64, plus a lazy-dtype assertion in `test_binary_dask_cupy`. Net diff vs main is just those two spots (and the classify metadata-sweep state row).

New head: 1faa5c6caf47797fb11558fe8b7b3f96b44488aa

PR #3514, opened 2026-06-25
Make binary() output float32 on all backends

### LINKED ISSUE #3511: binary() output dtype differs across backends (float64 on numpy/dask vs float32 on cupy)
`binary()` returns a different result dtype depending on the array backend.

On numpy and dask+numpy the output keeps the input dtype, because `_cpu_binary` allocates `np.empty(data.shape, dtype=data.dtype)`. On cupy and dask+cupy the output is always float32, because `_run_cupy_binary` allocates `cupy.empty(..., dtype='f4')`.

Reproducer with a float64 input:

```
numpy      -> float64
dask       -> float64
cupy       -> float32
dask+cupy  -> float32
```

Every other classifier in `classify.py` routes through `_cpu_bin`, which fixes the output to float32, so `binary` is the only classifier whose result dtype depends on the backend. Preserving the input dtype also cannot hold the NaN sentinel that `binary` writes for non-finite cells when the input is an integer array.

Proposed fix: allocate float32 in `_cpu_binary` so all four backends agree.

Found by the accuracy sweep (Cat 5, backend inconsistency).

brendancol:
## Summary
`binary()` returned the input dtype on the numpy and dask+numpy backends but always float32 on cupy and dask+cupy. This allocates float32 in `_cpu_binary` so the result dtype matches across all four backends, which is what the cupy path and every other classifier already do.

## Why
`_cpu_binary` allocated `np.empty(data.shape, dtype=data.dtype)`. For a float64 input the numpy and dask results were float64 while the cupy results were float32. Integer input was worse: the array cannot store the NaN that `binary` writes for non-finite cells.

## Changes
- `_cpu_binary` allocates a float32 output.
- Added `test_binary_output_dtype_float32` and `test_binary_dask_output_dtype_float32`.

## Testing
- `pytest xrspatial/tests/test_classify.py` passes (104 tests, numpy/dask/cupy/dask+cupy).

Closes #3511.

### DIFF
--- a/xrspatial/classify.py
+++ b/xrspatial/classify.py
@@ -48,7 +48,13 @@ def _available_memory_bytes():
 
 @ngjit
 def _cpu_binary(data, values):
-    out = np.empty(data.shape, dtype=data.dtype)
+    # Output float32 to match the cupy/dask+cupy backends (which always
+    # allocate 'f4') and the other classifiers, which route through _cpu_bin
+    # and likewise emit float32. Preserving the input dtype here made binary()
+    # the lone op whose result dtype diverged across backends (float64 on
+    # numpy/dask vs float32 on cupy) and could not hold the NaN sentinel for
+    # integer input.
+    out = np.empty(data.shape, dtype=np.float32)
     out[:] = np.nan
     rows, cols = data.shape
     for y in range(0, rows):
--- a/xrspatial/tests/test_classify.py
+++ b/xrspatial/tests/test_classify.py
@@ -44,7 +44,7 @@ def test_binary_numpy(result_binary):
     values, expected_result = result_binary
     numpy_agg = input_data()
     numpy_result = binary(numpy_agg, values)
-    general_output_checks(numpy_agg, numpy_result, expected_result)
+    general_output_checks(numpy_agg, numpy_result, expected_result, verify_dtype=True)
 
 
 @dask_array_available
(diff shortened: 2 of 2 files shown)


brendancol:
## PR Review: Make binary() output float32 on all backends

### Blockers
None.

### Suggestions
- The four existing backend tests (`test_binary_numpy`, `test_binary_dask_numpy`, `test_binary_cupy`, `test_binary_dask_cupy` in `xrspatial/tests/test_classify.py`) call `general_output_checks` without `verify_dtype=True`. The two new tests assert float32 only on numpy and dask+numpy, so the cupy and dask+cupy output dtype is not checked anywhere. Passing `verify_dtype=True` to those four tests locks the float32 contract on all four backends without adding new test functions, which is what the reclassify tests already do.

### Nits
- The new comment says the input dtype "could not hold the NaN sentinel for integer input." That holds in principle, but an integer raster has no non-finite cells, so `binary` never actually wrote NaN into an int output. The cross-backend consistency point alone justifies the change.
- `test_binary_dask_output_dtype_float32` uses one chunk layout. Output dtype is chunk-independent here, so a second layout adds little.

### What looks good
- One-line change plus a comment. Values are unchanged, only the output dtype.
- Brings numpy/dask in line with the cupy 'f4' allocation and with the docstring example, which already prints `dtype=float32` (classify.py line 161).
- Regression tests cover numpy (three input dtypes) and dask.
- Full classify suite passes: 104 tests across numpy, dask, cupy, dask+cupy.

### Duplicate
PR #3513 (metadata sweep) fixes the same `binary()` dtype bug. I am not merging or closing either here; flagging for maintainer reconciliation. This PR keeps its fix self-consistent.

### Checklist
- [x] Classification logic untouched (dtype-only change)
- [x] Backends produce consistent dtype
- [x] NaN handling correct
- [x] Edge cases: integer/float32/float64 input covered
- [x] Dask chunk boundaries: dtype is chunk-independent
- [x] No benchmark needed (dtype-only)
- [x] Docstring documents float32

Issue #3510, opened 2026-06-25
classify: head_tail_breaks, percentiles, box_plot crash on all-NaN / all-inf input


brendancol:
When every pixel is non-finite (all NaN, or all +/-inf, which gets mapped to NaN before binning), three classifiers raise an opaque reduction error instead of returning an all-NaN result like the others do.

Affected:
- `head_tail_breaks` -> `ValueError: zero-size array to reduction operation fmax which has no identity`
- `percentiles` -> `IndexError: index -1 is out of bounds for axis 0 with size 0`
- `box_plot` -> `ValueError: zero-size array to reduction operation fmax which has no identity`

Repro:
```python
import numpy as np, xarray as xr
from xrspatial.classify import box_plot
box_plot(xr.DataArray(np.full((4, 5), np.nan)))
```

Each one takes a max or percentile over the finite subset without checking for the empty case, and the finite mask drops every element. `equal_interval`, `std_mean`, and `maximum_breaks` already handle this and return an all-NaN array; these three should match.

Found by the test-coverage sweep. All-NaN regression tests are added and xfail-marked against this issue.

PR #3515, opened 2026-06-25
test(classify): cover all-NaN degenerate input across classifiers

brendancol:
Adds all-NaN regression tests for the classifiers that lacked them.

`std_mean` and `maximum_breaks` already return an all-NaN result on
all-NaN input, so they get plain passing tests. `head_tail_breaks`,
`percentiles`, and `box_plot` currently crash on all-NaN input, so their
tests are xfail (non-strict) and reference the source bug in #3510. Flip
them to plain assertions once #3510 is fixed.

`equal_interval` and `natural_breaks` already had all-NaN coverage.

Test-only change, no source modified. Ran locally on a CUDA host:
4 passed, 3 xfailed.

Found by the test-coverage sweep. Fixes the coverage half of #3510.

### DIFF
--- a/xrspatial/tests/test_classify.py
+++ b/xrspatial/tests/test_classify.py
@@ -1252,3 +1252,47 @@ def test_generate_sample_indices_large_is_deterministic():
     a = _generate_sample_indices(20_000_000, 20_000)
     b = _generate_sample_indices(20_000_000, 20_000)
     np.testing.assert_array_equal(a, b)
+
+
+# ===================================================================
+# Degenerate all-NaN input for the remaining classifiers
+# ===================================================================
+# equal_interval and natural_breaks already cover all-NaN. The other
+# classifiers were not exercised on an all-non-finite raster, where the
+# finite mask removes every element. std_mean and maximum_breaks degrade
+# to an all-NaN result; head_tail_breaks, percentiles, and box_plot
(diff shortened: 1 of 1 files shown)


brendancol:
## Review: degenerate-input regression tests for classify

Blockers: none.

Suggestion (addressed in a8cbf28f): the all-NaN tests ran on numpy only, but the degenerate path diverges by backend. Running all five on dask:

- `std_mean` and `maximum_breaks` return all-NaN on both eager and dask
- `head_tail_breaks` crashes eager, but its dask path returns all-NaN cleanly because of a `total_count == 0` guard the eager path lacks
- `percentiles` and `box_plot` crash on dask too

Added five `@dask_array_available` all-NaN tests: passing for `std_mean`, `maximum_breaks`, and `head_tail_breaks` (dask); `xfail(#3510)` for `percentiles` and `box_plot` (dask). Also wrapped the `std_mean` all-NaN tests in `warnings.catch_warnings()` to drop three empty-slice RuntimeWarnings.

Note for #3510: the `head_tail_breaks` crash is eager-only (numpy/cupy). The source fix should bring the eager paths in line with the existing dask guard.

What looks good: `xfail(strict=False)` is the right call with parallel sweeps in flight, since the tests assert the real all-NaN result and the xfails flip to pass exactly when #3510 lands. Full `test_classify.py`: 107 passed, 5 xfailed.

Issue #3506, opened 2026-06-25
classify: stale docstring example outputs (reclassify, equal_interval) and five classifiers with no example


brendancol:
Some docstring examples in `xrspatial/classify.py` print results that do not match what the functions return, and five public classifiers have no Examples section.

### Wrong example output (printed array differs from the actual result for the given input)

- `reclassify`: the numpy, dask, and cupy blocks all print
  `[[nan,1,1,1,1],[1,2,2,2,2],[2,2,2,2,2],[2,3,3,3,3]]`
  but running the example produces
  `[[nan,1,1,1,1],[1,1,1,1,1],[1,2,2,2,2],[2,3,3,3,nan]]`.
- `equal_interval`: the block prints
  `[[nan,0,0,0,0],[0,0,0,0,1],[1,1,1,1,1],[1,2,2,2,2],[2,2,2,2,nan]]`
  but running it produces
  `[[nan,0,0,0,0],[0,1,1,1,1],[2,2,2,2,2],[3,3,3,3,4],[4,4,4,4,nan]]`.
- `binary`: the printed array uses `np.nan` where numpy reprs plain `nan`.

The `test_classify.py` fixtures confirm the functions behave correctly, so the docstring outputs are stale, not the code.

### Missing Examples section

`std_mean`, `head_tail_breaks`, `percentiles`, `maximum_breaks`, and `box_plot` have Parameters, Returns, and References but no Examples block.

### Scope

Documentation only (docstrings). No behavior change. Found via the documentation sweep on the `classify` module.

brendancol:
Already resolved by #3516 (merged 2026-06-26). That PR fixed the stale `reclassify`/`equal_interval` outputs and the `binary` `nan` repr, and added Examples sections to `std_mean`, `head_tail_breaks`, `percentiles`, `maximum_breaks`, and `box_plot`.

The issue stayed open because #3516 referenced it as `(#3506)` in parentheses rather than with a `Fixes #3506` keyword, so it never auto-closed.

Verified on current `main` by running each example: every docstring output now matches the function's actual return value, and all five previously-missing Examples sections are present. Closing as completed.

PR #3516, opened 2026-06-25
docs(classify): fix stale example outputs and add missing examples

brendancol:
Fixes the documentation issues found by the documentation sweep on the `classify` module (#3506).

### Stale / wrong example output
- `reclassify`: the numpy, dask, and cupy example blocks all printed an array that does not match the function's result for the example input. Corrected all three to the actual output.
- `equal_interval`: the example printed a stale array. Corrected to the actual output.
- `binary`: the example used `np.nan` in the array repr where numpy prints `nan`. Fixed.

`test_classify.py` fixtures confirm the functions return the corrected values, so the code was right and only the docstrings were stale.

### Missing Examples
Added an Examples section to `std_mean`, `head_tail_breaks`, `percentiles`, `maximum_breaks`, and `box_plot`. Each output was generated by running the function and pasted verbatim.

### Scope
Docs only. No behavior change. `test_classify.py` passes (102) and flake8 is clean.

The `natural_breaks` `num_sample`/`None` type omission (already tracked in #3501) is intentionally left out of this PR.


brendancol:
## Review: classify docstring examples

Re-ran every docstring example on a CUDA host and diffed the printed output against actual output.

Blockers: none.

Suggestion (fixed in ef98faa9): the `reclassify` docstring had its three result arrays corrected by the PR, but the surrounding lazy reprs in the same docstring were still stale:

- the `agg_da` repr showed `dtype=float32`, but the float64 input gives `dtype=float64`
- the lazy result repr showed `dask.array<_run_numpy_bin, ...>`; the current task name is `dask.array<xrspatial.reclassify, ...>` after the `_dask_task_name_kwargs` rename
- the cupy block printed `<class 'cupy.core.core.ndarray'>`; current cupy returns `<class 'cupy.ndarray'>`

All three corrected against live output.

Nit (left as-is): the example headers omit the `Size: NNB` suffix modern xarray prints. Kept consistent with the other examples in this module (`quantile`, `natural_breaks`); these are `.. sourcecode::` blocks, not doctests.

What looks good: the two genuinely wrong outputs (`reclassify`, `equal_interval`) now match actual output; the five new Examples print exactly what the functions produce; docs-only with no behavior change; 102 tests pass; flake8 clean.

PR #3519, opened 2026-06-25
geotiff: cover categorical PAM sidecar round-trip on dask and GPU write paths

### LINKED ISSUE #3518: geotiff: categorical PAM sidecar round-trip tested only on the eager numpy write path
The #3483 categorical PAM sidecar feature wires `_write_category_sidecar()` into every `to_geotiff` write path. The eager numpy writer, the dask streaming writer, and the GPU (nvCOMP) writer each call it after writing pixels (`_writers/eager.py` lines 831, 880, 1063, 1159).

The feature's tests (`xrspatial/tests/test_rasterize_categorical_3482.py`) only round-trip the eager numpy path. The dask and GPU branches each have their own emit call that no test touches, so dropping the sidecar emit on one of those branches in a refactor would lose category labels and colors with no failing test.

I probed both on a CUDA host: a categorical int32 raster written through the dask streaming writer and through the GPU writer both emit the sidecar and read back the names and RGBA colors correctly today. So this is a backend coverage gap, not a source bug.

Also untested: the `category_colors=None` branch of `build_pam_xml` (no Red/Green/Blue/Alpha columns) and the names-only read path. The existing suite always supplies colors.

Plan, test-only, in `xrspatial/geotiff/tests/write/`: dask write round-trip of names + colors; GPU write round-trip (requires_gpu); names-only round-trip.

brendancol:
Closes #3518.

The #3483 categorical PAM sidecar feature emits a `<file>.aux.xml` sidecar on the eager numpy, dask streaming, and GPU (nvCOMP) write paths, but only the eager numpy round-trip was tested. The dask and GPU branches each have their own `_write_category_sidecar()` call, so dropping the emit on one of them in a refactor would lose category labels with no failing test.

New file `xrspatial/geotiff/tests/write/test_category_sidecar_backends_3483.py`:
- `test_dask_write_round_trips_categories`: dask streaming writer emits and round-trips names + colors
- `test_gpu_write_round_trips_categories`: GPU writer round-trips names + colors (requires_gpu)
- `test_round_trips_names_without_colors`: covers the `category_colors=None` build branch and the names-only read path, which the existing suite never hits

Test-only, no source edits. The paths already work today; this locks them in. All three pass on a CUDA host.


brendancol:
review-pr: approve with one suggestion, now applied.

The new tests reach the branches they claim. The dask test hits the streaming writer + sidecar emit at _writers/eager.py:1063 (not the eager-compute fallback at 1159), and the GPU test reaches _write_geotiff_gpu + sidecar emit at line 880. The names-only test covers the category_colors=None build path.

Applied: the tests asserted sidecar metadata but not the pixel codes, so a corrupted int32 band would have passed. Added assert_array_equal(back.values, _CODES) to all three. 3/3 pass on the CUDA host, including the requires_gpu GPU-write test.

PR #3521, opened 2026-06-25
Add Examples section to to_geotiff docstring

brendancol:
## What

Adds an Examples section to the `to_geotiff` docstring.

## Why

`open_geotiff` has an Examples section with copy-pasteable usage. `to_geotiff`, the matching write entry point, documented Parameters, Returns, and Raises but had no Examples. That was the one structural docstring gap a documentation sweep of the geotiff module turned up; everything else (signature/docstring parity, reference-page coverage, tier claims) is already locked by tests.

## What changed

A numpydoc Examples block with three `>>> ... # doctest: +SKIP` snippets, following the `open_geotiff` block and the writing examples already in `docs/source/reference/geotiff.rst`:

- plain GeoTIFF write then read back
- `cog=True` Cloud Optimized GeoTIFF write
- `.vrt` mosaic write

Docs only. No behavior change, no signature change. `test_polish.py` and `test_to_geotiff_parameters_match_signature_1683` pass.


brendancol:
## PR Review: Add Examples section to to_geotiff docstring

Reviewed from the head-branch worktree. Docs-only change: an Examples block on the `to_geotiff` docstring plus one row in the sweep state CSV. No code paths, signatures, or behavior touched.

### Blockers
None.

### Suggestions
None.

### Nits
- The import line `>>> from xrspatial.geotiff import open_geotiff, to_geotiff` has no `# doctest: +SKIP` while the calls under it do. That matches the existing `open_geotiff` block exactly and the import is valid, so it's fine. Worth knowing for later: if the repo ever turns on `--doctest-modules`, the import runs (and passes) and the `+SKIP` calls are skipped. Doctests don't run on modules today.

### What looks good
- All four example calls hit the real API: `to_geotiff(data, path)`, `to_geotiff(data, path, cog=True)`, `to_geotiff(data, 'mosaic.vrt')`, `open_geotiff(path)`. Both functions are in `__all__`.
- The `cog=True` note ("tiled, with internal overviews") and the `.vrt` note ("tiles the array and emits the index that references the tiles") agree with the rest of the docstring and with docs/source/reference/geotiff.rst.
- `+SKIP` usage and style copy the `open_geotiff` block, which was the stated model. `data` is an undefined placeholder, but every line that uses it is `+SKIP`, so doctest collection is unaffected, same as the `open_geotiff` block.

### Checklist
- Algorithm / backend parity / NaN / dask boundaries / materialization: n/a (docs only)
- Benchmark: not needed
- README feature matrix: n/a (function already present)

Verdict: no actionable changes. Accurate and consistent with the existing `open_geotiff` docstring. `test_polish.py` and `test_to_geotiff_parameters_match_signature_1683` pass.

PR #3522, opened 2026-06-25
Stop a short-row PAM RAT from crashing open_geotiff

### LINKED ISSUE #3520: GeoTIFF: a crafted .aux.xml sidecar crashes open_geotiff with IndexError
`open_geotiff()` reads a PAM `.aux.xml` sidecar for any local string source through `read_pam_sidecar()` (`xrspatial/geotiff/__init__.py:920`). That function's docstring promises a missing, malformed, or foreign sidecar is non-fatal auxiliary metadata and returns `{}` instead of raising.

`_parse_rat()` in `xrspatial/geotiff/_pam.py` reads RAT cells by FieldDefn column index: `fields[value_col]`, `fields[name_col]`, and the RGBA cells `fields[c]`. A `<Row>` can carry fewer `<F>` cells than the highest column index. Put the Name column at index 1 (behind a Value column at index 0) and give a row a single cell, and `fields[name_col]` raises `IndexError`. `read_pam_sidecar()` only catches `(OSError, ValueError, TypeError)`, so the `IndexError` escapes and crashes the `open_geotiff` call.

A PAM sidecar travels next to the TIFF it describes, so a crafted or corrupt `.aux.xml` shipped in a dataset crashes any consumer that opens the `.tif`. Other malformed RATs (non-numeric value cell, missing index attribute) already degrade to `{}`. The short-row case is the one path that escapes.

Repro: write `<evil.tif>.aux.xml` with a thematic `GDALRasterAttributeTable` holding a Value FieldDefn at index 0 (Usage 5), a Class FieldDefn at index 1 (Usage 2), and one `<Row index="0"><F>0</F></Row>`. `read_pam_sidecar("evil.tif")` raises `IndexError` instead of returning `{}`.

Severity: MEDIUM. It's a denial-of-service/robustness break reachable from attacker-controlled auxiliary input, with no memory corruption or code execution. Found by the geotiff security sweep (2026-06-25).

Fix: add `IndexError` to the exceptions `read_pam_sidecar()` swallows so a short-row RAT degrades to `{}` like every other malformed sidecar.

brendancol:
`read_pam_sidecar()` is supposed to never raise on a bad sidecar. `open_geotiff()` reads the PAM `.aux.xml` for any local string source, and a malformed one should degrade to `{}` rather than break the read.

`_parse_rat()` indexes RAT cells by column index (`fields[name_col]` and friends). A `<Row>` with fewer `<F>` cells than the highest column index raises `IndexError`, which the `read_pam_sidecar()` except clause didn't cover (it caught `OSError`, `ValueError`, `TypeError`). The `IndexError` escaped and crashed `open_geotiff` for any TIFF carrying such a sidecar.

This adds `IndexError` to the caught exceptions so a short-row RAT falls back to `{}` like every other malformed case.

Test: `test_short_row_thematic_rat_returns_empty` builds a thematic RAT with the Name column at index 1 and a single-cell row, then asserts `read_pam_sidecar` returns `{}`. It fails with `IndexError` before the fix and passes after.

Closes #3520.

### DIFF
--- a/xrspatial/geotiff/_pam.py
+++ b/xrspatial/geotiff/_pam.py
@@ -144,9 +144,14 @@ def read_pam_sidecar(path):
         if colors is not None:
             out['category_colors'] = colors
         return out
-    except (OSError, ValueError, TypeError):
+    except (OSError, ValueError, TypeError, IndexError):
         # A missing, malformed, or foreign sidecar is non-fatal auxiliary
         # metadata, not a read error -- never let it break open_geotiff.
+        # IndexError covers a thematic RAT whose <Row> carries fewer <F>
+        # cells than its highest column index (e.g. a Name column at index
+        # 1 paired with a single-cell row), which would otherwise escape
+        # _parse_rat and crash the open_geotiff call that reads the
+        # adjacent sidecar.
         return {}
 
 
--- a/xrspatial/geotiff/_pam.py
+++ b/xrspatial/geotiff/_pam.py
@@ -15,6 +15,7 @@ This module builds that sidecar from ``attrs['category_names']`` /
 from __future__ import annotations
 
 import os
+from xml.etree.ElementTree import ParseError
 from xml.sax.saxutils import escape
 
 from ._safe_xml import safe_fromstring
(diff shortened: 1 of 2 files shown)


brendancol:
review-pr: one blocker, now fixed.

The original IndexError fix is correct but incomplete. read_pam_sidecar still let a parse error escape: safe_fromstring() raises xml.etree.ElementTree.ParseError on any non-well-formed sidecar, and ParseError subclasses SyntaxError, which was not in the caught tuple (OSError, ValueError, TypeError, IndexError). So a truncated or corrupt .aux.xml, which is more common than a crafted short-RAT row, still escaped and crashed open_geotiff: the same contract violation this PR set out to fix, reached through a different exception. defusedxml 0.7.1 raises the same ET.ParseError, so catching it covers both parse paths.

Fix: added ParseError to the caught tuple plus a regression test (truncated XML returns empty). TestPamHelpers now 7 passed, flake8 clean. The test fails with ParseError before the fix and passes after.

Non-blocking: whole-sidecar discard vs per-row skip is consistent with existing behavior, left as-is.

PR #3524, opened 2026-06-25
Add crs_units and crs_name to from_template() result.attrs

### LINKED ISSUE #3523: from_template(): add crs_units and crs_name to result.attrs
## Description

`from_template()` returns a DataArray whose `attrs` carry `res` and `crs` (the EPSG int). The CRS units and the projection's human-readable name both follow from the EPSG code, but neither is surfaced, so a caller inspecting the result has to look them up separately.

## Proposed change

Add two keys to the returned `attrs`:

- `crs_units`: the unit of the CRS (`"m"` for projected templates, `"degree"` for country codes in EPSG:4326). This is the same value already attached to the `x`/`y` coordinate `units`.
- `crs_name`: the projection's human-readable name, for example `"NAD83 / Conus Albers"` for EPSG:5070 or `"WGS 84"` for EPSG:4326.

The name is resolved through the existing two-tier CRS path (the `LiteCRS` built-in table first, then a pyproj fallback). For EPSG codes outside the built-in table when pyproj isn't installed, `crs_name` is left off rather than raising, so the default non-reproject path stays dependency-free.

## Why

It makes the template result self-describing. Downstream code and notebooks can print or assert on the projection name and units without a separate pyproj lookup.

brendancol:
Closes #3523

## What

`from_template()` now adds two keys to the returned `attrs`:

- `crs_units`: `"m"` for projected templates, `"degree"` for country codes in EPSG:4326 (the same value already on the `x`/`y` coordinate `units`).
- `crs_name`: the projection's human-readable name, e.g. `"NAD83 / Conus Albers"` (5070), `"WGS 84"` (4326).

The name resolves through the existing two-tier CRS path (`LiteCRS` table, then pyproj). For a code outside the built-in table with pyproj absent, `crs_name` is left off rather than raised, so the default path stays dependency-free. A `name` property was added to `LiteCRS` to mirror `pyproj.CRS.name`.

## Backends

Pure metadata attached after array creation, so all four backends (numpy, cupy, dask+numpy, dask+cupy) carry identical attrs. The existing `agg.attrs == ref.attrs` cross-backend assertions still hold.

## Test plan

- [x] `crs_units` matches coordinate units for projected and geographic templates
- [x] `crs_name` deterministic for built-in-table codes (5070, 4326)
- [x] `crs_name` present on the preserve path
- [x] `crs_name` omitted (not raised) when `_resolve_crs` can't import pyproj
- [x] `LiteCRS.name` property
- [x] Full `test_templates.py` (44) and `reproject` suite (633) green


brendancol (commented):
## PR Review: Add crs_units and crs_name to from_template() result.attrs

None.

None.

- `templates.py` — `crs_units` is derived by `"degree" if crs == 4326 else "m"`. That's correct for every code the templates emit today (country codes are 4326, all projected/preserve codes are metre-based), but it's a heuristic rather than an axis-unit lookup. Worth a one-line comment if a future template ever adds a non-metre projected CRS or a non-4326 geographic one.
- `crs_name` can differ by resolver: a UTM code in the built-in `LiteCRS` table returns `"WGS84 / UTM zone 18N"`, while the pyproj fallback returns `"WGS 84 / UTM zone 18N"`. Descriptive metadata only; the tests assert just on 5070/4326 (identical in both), so it's not a correctness problem, just a seam to be aware of.

### What looks good
- Metadata is attached after array creation, so all four backends carry identical attrs. The existing `agg.attrs == ref.attrs` cross-backend assertions confirm parity and still pass.
- `crs_name` is best-effort: the `(ImportError, ValueError, TypeError)` catch keeps the default non-reproject path dependency-free, and the omission branch is tested via monkeypatch.
- `LiteCRS.name` mirrors `pyproj.CRS.name` with an `EPSG:<code>` fallback, giving one way to read the name across both resolver tiers.

### Checklist
- [x] No algorithm/paper involved (pure metadata)
- [x] All implemented backends produce consistent attrs
- [x] No NaN/float concerns (no numeric path touched)
- [x] Edge case covered: pyproj-absent omission branch tested
- [x] No dask chunk concerns (metadata only)
- [x] No benchmark needed
- [x] No README matrix change needed
- [x] Docstring Returns section updated

brendancol (commented):
## Follow-up review (commit 54ab7e0e)

Nit 1 is addressed: `templates.py` now carries a one-line comment noting that every template emits either EPSG:4326 (degrees) or a metre-based projected CRS, which is what the `crs_units` heuristic relies on.

Nit 2 (the `LiteCRS` vs pyproj UTM name spacing seam) is left as-is. It reflects an existing naming convention in `LiteCRS`'s UTM table; normalizing it would change `LiteCRS.name`/`to_wkt` output for every UTM code, which is outside this PR's scope. The new attrs are descriptive metadata and the tests assert only on codes whose names match across both resolvers.

No new findings. `test_templates.py` (44) green after the follow-up.

PR #3532, opened 2026-06-26
from_template: use CF Conventions metadata instead of crs_units/crs_name

### LINKED ISSUE #3531: from_template: use CF Conventions metadata instead of crs_units/crs_name
## What's wrong

`from_template` tags its output `DataArray` with two attribute keys that don't follow any recognized metadata standard:

- `crs_units` (`'m'` or `'degree'`)
- `crs_name` (the human-readable projection name, e.g. `'NAD83 / Conus Albers'`)

`crs_units` also duplicates information that belongs on the coordinate variables, and the coordinate `units` it mirrors (`'degree'`) isn't CF-valid for lon/lat axes.

## Proposed change

Align the template metadata with the CF (Climate and Forecast) Conventions, the metadata standard xarray and the wider geoscience stack already lean on.

- Drop `crs_units` from the DataArray attrs. Units belong on the coordinates, where `x`/`y` already carry them.
- Make the coordinate metadata CF-correct:
  - EPSG:4326: `units` of `degrees_east`/`degrees_north` with `standard_name` `longitude`/`latitude` (today it emits `'degree'`).
  - Projected: `units` of `m` with `standard_name` `projection_x_coordinate`/`projection_y_coordinate`.
- Replace `crs_name` with the CF grid-mapping keys `grid_mapping_name` (the projection token, e.g. `albers_conical_equal_area`) and `crs_wkt` (full WKT, which carries the human-readable name). This matches how the rest of the library identifies a CRS: `crs`/`crs_wkt`, with everything else derived via pyproj.

## Notes

- The CF keys come from `pyproj.CRS.to_cf()`. Same as the current `crs_name`, they stay best-effort: without pyproj the default path is dependency-free and the keys are omitted.
- Equal Earth (the `preserve='area'` fallback) has no CF grid mapping, so `grid_mapping_name` is left off there and only `crs_wkt` is emitted.
- This changes the attribute contract of `from_template`. The library is pre-1.0, so no deprecation shim.

brendancol:
Closes #3531.

`from_template` was tagging its output with two non-standard attrs, `crs_units` and `crs_name`. This swaps them for CF (Climate and Forecast) Conventions metadata, the standard xarray and the geoscience stack already use.

## Changes

- Drop `crs_units`. Units now live on the `x`/`y` coordinates where they belong. EPSG:4326 axes get `degrees_east`/`degrees_north` (was a non-CF `'degree'`); projected axes get `m`, each with the matching CF `standard_name` (`longitude`/`latitude`, `projection_x_coordinate`/`projection_y_coordinate`).
- Replace `crs_name` with the CF grid-mapping keys `grid_mapping_name` (e.g. `albers_conical_equal_area`) and `crs_wkt` (full WKT, which carries the human-readable name). This lines up with the rest of the library, which keys off `crs`/`crs_wkt`.
- Source the CF keys from `pyproj.CRS.to_cf()`, kept best-effort: without pyproj the default path stays dependency-free and the keys are omitted, same as `crs_name` before. Equal Earth has no CF grid mapping, so `grid_mapping_name` is left off there and `crs_wkt` stands alone.

## Backends

Metadata-only change; behaviour is identical across numpy / cupy / dask+numpy / dask+cupy.

## Test plan

- [x] `test_cf_coordinate_units` — CF units + `standard_name` on x/y, no `crs_units`
- [x] `test_cf_grid_mapping_attrs` — `grid_mapping_name` + `crs_wkt`, no `crs_name`
- [x] `test_cf_grid_mapping_preserve_path` — preserve path keys present
- [x] `test_grid_mapping_omitted_for_equal_earth` — `crs_wkt` only for 8857
- [x] `test_cf_attrs_omitted_without_pyproj` — graceful dependency-free fallback
- [x] Full `test_templates.py` suite green (51 passed)

### DIFF
--- a/xrspatial/templates.py
+++ b/xrspatial/templates.py
@@ -173,18 +173,29 @@ def _make_data(shape, fill, backend, chunks):
     )
 
 
-def _crs_name(crs):
-    """Best-effort human-readable name for an EPSG code.
+def _cf_crs_attrs(crs):
+    """CF Conventions grid-mapping attributes for an EPSG code.
 
-    Resolves through the same two-tier path as the rest of the library
-    (built-in ``LiteCRS`` table first, pyproj fallback). Returns ``None``
-    when the code is outside the built-in table and pyproj is not
-    installed, so the default (non-reproject) path stays dependency-free.
+    Emits the two canonical CF-1.7 identifiers via :meth:`pyproj.CRS.to_cf`:
+    ``grid_mapping_name`` (the controlled-vocabulary projection token, e.g.
+    ``'albers_conical_equal_area'``) and ``crs_wkt`` (the full WKT, which
+    carries the human-readable CRS name). This mirrors the rest of the
+    library, which identifies a CRS by ``crs`` / ``crs_wkt`` and derives
+    anything else with pyproj on demand.
+
+    Best-effort: without pyproj installed the mapping is empty, so the
+    default (non-reproject) path stays dependency-free, same as the old
+    ``crs_name`` behaviour. ``grid_mapping_name`` is also left out for
+    projections CF does not define (e.g. Equal Earth), where ``to_cf``
+    returns ``crs_wkt`` alone.
     """
     try:
-        return _resolve_crs(crs).name
-    except (ImportError, ValueError, TypeError):
-        return None
+        import pyproj
+    except ImportError:
+        return {}
+    cf = pyproj.CRS.from_epsg(int(crs)).to_cf()
+    return {k: cf[k] for k in ("grid_mapping_name", "crs_wkt")
+            if cf.get(k) is not None}
 
 
 def from_template(name, resolution=None, *, preserve=None, backend="numpy",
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: from_template — CF Conventions metadata

### Blockers
None.

### Suggestions
- **Flat attrs vs. a real CF grid mapping** (`templates.py:325`, `:338-345`). Strict CF puts `grid_mapping_name` on a separate scalar grid-mapping variable that the data variable points to via a `grid_mapping` attribute; here both `grid_mapping_name` and the coordinate `standard_name`s sit directly on the data variable. That matches how this library already models CRS (flat attrs, `crs`/`crs_wkt`), so it's a reasonable call — but a strict CF reader like cf_xarray won't auto-detect it as a grid mapping. Worth a one-line docstring caveat, or a follow-up if fuller CF encoding is ever wanted. Not blocking.
- **Narrower exception catch than before** (`templates.py:194-200`). `_cf_crs_attrs` catches only `ImportError`; the old `_crs_name` also swallowed `ValueError`/`TypeError`. Every caller passes a curated/queried EPSG int, so `from_epsg().to_cf()` won't fail in practice — but if an unknown code ever reached here it would now raise instead of degrading to "no CF keys." Low risk given current call sites; flagging for awareness.

### Nits
- **Inconsistent wrapping** (`templates.py:339-345`). The 4326 branch keeps `.attrs.update(...)` on one line (83 chars) while the projected branch wraps onto two. Both fit the 100-char limit; pick one style for symmetry.

### What looks good
- Clean swap: no code outside the tests still reads `crs_units`/`crs_name` from a template, so nothing downstream breaks.
- `crs_wkt` lines up with the geotiff layer, which already documents `crs`/`crs_wkt` as the canonical, additive CRS keys.
- Test coverage is solid — both coordinate flavors, both grid-mapping branches, plus the two real edge cases (Equal Earth has no CF mapping, and the pyproj-absent fallback).

brendancol (commented):

Re-reviewed the two follow-up commits. Disposition of the first-pass findings:

- **Nit — inconsistent wrapping:** fixed. Both branches of the coordinate-units block now use single-line `.attrs.update(...)` (all within the 100-char limit).
- **Suggestion — flat attrs vs. CF grid-mapping variable:** addressed by docstring. The `Returns` section now states the keys sit on `attrs` (the library's flat CRS convention), not a separate CF grid-mapping variable, so a strict CF reader won't auto-detect them.
- **Suggestion — narrower exception catch in `_cf_crs_attrs`:** dismissed. Every call site passes a curated or pyproj-queried EPSG int, so `from_epsg().to_cf()` can't fail in practice. CLAUDE.md asks for no error handling on unreachable paths, so I left the catch scoped to `ImportError` (the only real branch — the dependency-free fallback).

`flake8` clean; `test_templates.py` green (51 passed). No blockers outstanding.

PR #3534, opened 2026-06-26
Add major world cities as bounding-box templates in from_template

### LINKED ISSUE #3533: Add major world cities as bounding-box templates in from_template
## Reason or Problem

`from_template()` already gives you an empty study-area grid from a region name or country code. The catch: of the curated regions, only `nyc` is a city. If you want to start an analysis over Tokyo, London, or São Paulo, you're back to looking up a bounding box, picking a sensible projection, and assembling coordinates by hand — the exact chore `from_template` exists to remove.

## Proposal

Add a set of world cities (national capitals plus major regional metros) as templates, so `from_template("tokyo")` returns an empty grid over greater Tokyo in a sensible projection.

**Design:**
Each city follows the existing `nyc` pattern: a lon/lat metro bounding box, a UTM-zone CRS picked from the city's centroid (EPSG:326xx north / 327xx south), projected bounds computed once at authoring time, and a metre-based default resolution. UTM keeps every city on a known EPSG code — no synthesized PROJ4/WKT.

The city table is generated from Natural Earth populated places at authoring time and committed as static data next to the country bboxes, so there's no runtime network or geopandas dependency. Box size scales with population: bigger for megacities, tighter for regional centres. Cities slot into `_resolve` between the curated regions and the country codes, so `preserve='area'/'shape'` keeps working the same way it does for `nyc`.

**Usage:**

```python
agg = from_template("tokyo")          # empty UTM-54N grid over greater Tokyo
agg.attrs["crs"]                      # 32654
from_template("london").attrs["crs"]  # 32630
```

**Value:** Drops a common first step — find a bbox, choose a CRS, build the coordinates — for the places people most often demo and prototype on.

## Stakeholders and Impacts

I'll implement it. It only extends the template data and `_resolve`; no change to the backend dispatch or any existing region.

## Drawbacks

The registry grows by ~500 entries. City names collide (Paris, France vs Paris, Texas), so a disambiguation rule is needed. UTM is a reasonable default but isn't always the locally preferred grid.

## Alternatives

Use a national or city grid per city instead of UTM. Rejected for now: it needs per-city curation and judgment, and UTM already satisfies "standard EPSG projection" everywhere.

## Unresolved Questions

Final city count and the population cutoff for "regional" cities.

brendancol:
Closes #3533

Adds world cities to `from_template`, following the existing `nyc` pattern.

- 558 cities, generated from Natural Earth populated places and committed as static `_CITIES` data next to the country bboxes. No runtime network or geopandas dependency. Selection = every national capital + metros with POP_MAX >= 1.2M + a curated set of recognizable US secondary cities (Austin, New Orleans, Las Vegas, Portland, ...), since Natural Earth's POP_MAX/SCALERANK underrate US metros.
- Each city resolves to its UTM zone (EPSG:326xx north / 327xx south), picked from the centroid — a standard EPSG code, never a synthesized projection. Bounding box is a metro-scale lon/lat box (half-width scales with population) projected into that zone.
- Wired into `_resolve` between the curated regions and the country codes, so `preserve='area'/'shape'` work exactly as they do for `nyc`. No backend-dispatch changes.
- Name collisions: the larger-population city keeps the bare slug, the others take an `_<iso2>` suffix (e.g. `hyderabad` vs `hyderabad_pk`, `birmingham` UK vs `birmingham_us`).

```python
from_template("tokyo").attrs["crs"]        # 32654
from_template("austin").attrs["crs"]       # 32614
from_template("new_orleans").attrs["crs"]  # 32615
```

Backend coverage: numpy / cupy / dask+numpy / dask+cupy — cities reuse the same `_make_data` paths as every other template (verified locally on all four).

### Test plan
- [x] `pytest xrspatial/tests/test_templates.py` (56 passed)
- [x] Integrity loop over all cities (keys, UTM-range CRS, ordered lonlat/bounds, ascii lowercase keys)
- [x] Sample builds obey the array contract; pixel centers stay inside the bbox; southern-hemisphere build exercised
- [x] UTM spot-checks (london 32630, tokyo 32654, southern -> 327xx), case-insensitivity, collision disambiguation
- [x] Backend parity on dask+numpy / cupy / dask+cupy
- [x] flake8 + isort clean on touched files
- [x] Reconciled with main #3532 (CF grid-mapping metadata)


brendancol (commented):
## Review: world-city bounding-box templates

### Blockers
None.

### Suggestions
- Discoverability. The unknown-name error points to "the templates reference" (`templates.py:69-70`), but neither `templates.rst` nor any public API lists the 519 city names — `_CITIES` is private. Country codes have the same gap, except ISO-3166 is a discoverable standard and an ad-hoc city list isn't. A follow-up that exposes a names listing (or a reference note on how to find them) would help. Not blocking.

### Nits
- `test_city_sample_builds` (`test_templates.py:187`) builds only `sorted(_CITIES)[:20]`, which is alphabetical and northern-heavy. A southern-hemisphere build is covered in practice because `test_city_utm_spot_checks` calls `from_template("sao_paulo")`, but adding one southern city to the sample loop would make that explicit.

### What looks good
- Cities slot into `_resolve` between regions and countries (`templates.py:49-55`) and reuse `_make_data` unchanged, so all four backends work with no dispatch edits (checked locally on dask+numpy / cupy / dask+cupy).
- UTM-zone-from-centroid keeps every city on a standard EPSG code, and `preserve='area'/'shape'` fall through to the same `nyc` behavior for free.
- Data is committed static (no runtime geopandas or network), and the `_CITIES` header documents the source, the UTM rule, the buffer sizing, and the collision rule.
- `test_city_registry_integrity` checks the shape of all 519 entries, so a bad regeneration fails in CI instead of at call time.

### Checklist
- [x] CRS is a standard EPSG code (UTM), not synthesized
- [x] All four backends consistent
- [x] crs_units correct (metres for UTM cities)
- [x] Integrity covered for every entry
- [x] No dask/dispatch changes; reuses existing paths
- [x] No premature materialization (static dict)
- [x] No benchmark needed (allocation-only, matches existing from_template)
- [x] README + reference + notebook updated
- [x] Docstring updated with city example and collision rule

brendancol (commented):
## Follow-up review

Changes since the first pass:

- Nit fixed: `test_city_sample_builds` now includes `sao_paulo` and `sydney`, so the southern-hemisphere (327xx) build path is exercised directly.
- Discoverability suggestion deferred to #3535 — a public listing API is a separate design decision (name, return shape, whether to include the 240 country codes).
- Merged `origin/main` and reconciled with #3532, which replaced `crs_units`/`crs_name` with CF grid-mapping attributes. The city test now asserts CF coord units (`x.attrs["units"] == "m"`, `standard_name == "projection_x_coordinate"`) and the user-guide cell prints `grid_mapping_name` instead of `crs_name`.

No new blockers. Local: 56 passed in `test_templates.py`, flake8 + isort clean, backend parity holds on dask+numpy / cupy / dask+cupy.

PR #3536, opened 2026-06-26
Add list_templates() to discover from_template names

### LINKED ISSUE #3535: Expose a public way to list available from_template names (regions + cities)
## Reason or Problem

`from_template` now resolves 7 curated regions, 519 world cities, and 240 country codes (#3533). The city and region names live in private dicts (`_REGIONS`, `_CITIES`), so there's no public way to discover what's available. When a name misses, the error points to "the templates reference," but the reference doesn't enumerate the cities either. Country codes are at least an external standard (ISO-3166); the city list isn't, so it's the hardest to discover.

## Proposal

Add a small public helper that returns the available template names, e.g. `from xrspatial.templates import list_templates`, returning regions, cities, and country codes (probably grouped, or filterable by kind). Then point the unknown-name error and the reference doc at it.

Open design questions: name (`list_templates` vs `available_templates`), return shape (flat list vs grouped dict), whether to include the ~240 country codes or just say "ISO-3166 alpha-3."

## Stakeholders and Impacts

Follow-up to #3533. Pure addition; no change to existing behavior.

brendancol:
## Summary

`from_template` accepts curated regions, world cities, and country codes, but the names live in private dicts (`_REGIONS`, `_CITIES`, `_COUNTRY_BBOXES`). Short of reading the source, there was no supported way to find out what's available. This adds `list_templates()` to fill that gap.

- `list_templates()` returns the accepted names grouped into `regions`, `cities`, and `countries`, each a sorted list.
- `list_templates(kind)` returns one group's list; a bad `kind` raises `ValueError`.
- Every name it returns is a valid `from_template` argument (regions/cities lowercase, countries uppercase ISO-3166 / GADM alpha-3).
- The unknown-name error and the templates reference doc now point at the helper.

Available as both `xrspatial.list_templates` (next to `from_template`) and `xrspatial.templates.list_templates`.

## Backends

No backend dimension. `list_templates` returns plain lists and dicts of strings, not arrays. `from_template` is unchanged.

## Test plan

- [x] `list_templates()` keys are exactly `regions`/`cities`/`countries`, each equal to the sorted registry keys
- [x] `kind` filter returns the right sorted list; invalid `kind` raises
- [x] every listed name builds via `from_template`
- [x] unknown-name error mentions `list_templates`
- [x] full `test_templates.py` passes (63 tests), flake8 clean

Closes #3535

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ Built-in Numba JIT and CUDA projection kernels bypass pyproj for per-pixel coord
 
 | Name | Description | Source | NumPy xr.DataArray | Dask xr.DataArray | CuPy GPU xr.DataArray | Dask GPU xr.DataArray |
 |:----------:|:------------|:------:|:----------------------:|:--------------------:|:-------------------:|:------:|
-| [from_template](xrspatial/templates.py) | Empty study-area grid for a named region (CONUS, NYC, ...), a world city (London, Tokyo, ... in its UTM zone) or country code; `preserve='area'/'shape'` picks an EPSG projection by property | Custom | ✅ | 🔼 | 🔼 | 🔼 |
+| [from_template](xrspatial/templates.py) | Empty study-area grid for a named region (CONUS, NYC, ...), a world city (London, Tokyo, ... in its UTM zone) or country code; `preserve='area'/'shape'` picks an EPSG projection by property; `list_templates()` lists every accepted name | Custom | ✅ | 🔼 | 🔼 | 🔼 |
 
 -----------
 
--- a/docs/source/reference/templates.rst
+++ b/docs/source/reference/templates.rst
@@ -11,9 +11,13 @@ feeds straight into the rest of the library. Cities (national capitals, major
 regional metros, and recognizable US secondary cities) come back as a metro
 bounding box in their UTM zone.
 
+Call :func:`~xrspatial.templates.list_templates` to discover every name
+``from_template`` accepts (curated regions, world cities, and country codes).
+
 From Template
 =============
 .. autosummary::
     :toctree: _autosummary
 
     xrspatial.templates.from_template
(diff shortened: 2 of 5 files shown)


brendancol (commented):
## PR Review: Add list_templates() to discover from_template names

### Blockers
None.

### Suggestions
None.

### Nits
- `xrspatial/templates.py:202` and `:238-242` keep the kind names in two places: the `_TEMPLATE_KINDS` tuple feeds the error message, while the validation checks `kind not in groups`. Add a fourth group to `groups` and forget the tuple, and the error message goes stale. Building the message from `tuple(groups)` and dropping `_TEMPLATE_KINDS` removes the drift.

### What looks good
- Every name the helper returns is a real `from_template` argument, and the tests check that directly: `test_list_templates_grouped` compares against the sorted registry keys, and `test_list_templates_names_resolve` builds one from each group. That's the contract that matters.
- The unknown-name error and the reference doc both point at the helper now, which is what the issue asked for.
- Leaving `list_templates` out of the README backend matrix is right. It returns plain strings, no backend behavior to track.

### Checklist
- Algorithm matches reference: N/A (registry listing)
- Backend parity: N/A (no backend dimension)
- NaN handling: N/A
- Edge cases covered by tests: yes (invalid kind, every group resolves)
- Dask chunk boundaries: N/A
- Benchmark: not needed (dict lookup)
- README feature matrix updated: yes (folded into the from_template row)

brendancol (commented):
## Follow-up review

The single nit from the first pass is resolved in 2b9c6a44: dropped the `_TEMPLATE_KINDS` constant and the error message now reads its options from `tuple(groups)`, so the message can't drift from the actual groups.

No new issues. Tests still pass (9 list_templates/error tests green, flake8 clean). Nothing else outstanding.

PR #3538, opened 2026-06-26
Write continuous-raster symbology sidecars from to_geotiff (#3537)

### LINKED ISSUE #3537: to_geotiff: write best-practice symbology sidecars for continuous rasters (QGIS color ramp)
## Reason or Problem
When you write a categorical raster with `arr.xrs.to_geotiff()`, xarray-spatial drops a PAM `.aux.xml` sidecar holding the class names and colors, so QGIS opens the file already showing labelled, coloured classes. Continuous rasters (DEMs, slope, NDVI, and so on) get no such treatment. They open in QGIS as a flat grayscale stretch with whatever default min/max GDAL guesses, and the user has to set up a colour ramp by hand every time.

## Proposal
Add an opt-in `color_ramp` parameter to `to_geotiff` that, for continuous single-band rasters, writes best-practice symbology sidecars so QGIS applies a sensible colour ramp on open. This is the continuous-data counterpart to the categorical RAT sidecar we already write.

**Design:**
Two sidecars are written when `color_ramp` is set and the raster is continuous (no `category_names` attr) and single-band:
1. A QGIS `.qml` style file (`dem.qml`, replacing the extension, which is the name QGIS auto-loads) with a `singlebandpseudocolor` renderer and an interpolated colour ramp. This is what gives QGIS real colours.
2. `STATISTICS_MINIMUM/MAXIMUM/MEAN/STDDEV` in the existing PAM `.aux.xml` (`dem.tif.aux.xml`), which GDAL tools and QGIS use for a min/max stretch even if the `.qml` is ignored.

Colour ramps ship as hardcoded control points (viridis default, plus a small curated set) so there is no matplotlib runtime dependency. Statistics are computed with a single backend-aware reduction pass covering numpy, cupy, dask+numpy and dask+cupy. A `color_ramp_range=(min, max)` escape hatch skips the reduction for large dask graphs.

**Usage:**
```python
dem.xrs.to_geotiff('dem.tif', color_ramp='viridis')
# writes dem.tif, dem.qml, dem.tif.aux.xml
```
Open `dem.tif` in QGIS and it renders with the viridis ramp stretched across the data range.

**Value:** Continuous rasters written by xarray-spatial open in QGIS looking right without manual styling, matching the experience categorical rasters already get.

## Stakeholders and Impacts
Anyone writing continuous rasters for use in QGIS. Touches the geotiff writer (`to_geotiff` wrapper) and a new `_symbology` module plus the existing `_pam` sidecar module. Default behaviour is unchanged because the feature is opt-in.

## Drawbacks
Writes extra sidecar files. The `.qml` format is QGIS-specific (other GIS tools ignore it, though they still read the GDAL statistics). For the dask streaming write path, computing statistics adds a separate reduction pass unless `color_ramp_range` is supplied.

## Alternatives
- Statistics only, no `.qml`: works in all GDAL tools but only gives a grayscale stretch, no colour.
- Automatic for every continuous write: rejected to avoid surprise side-files and a forced statistics pass on writes that do not want symbology.
- A GDAL colour table embedded in the TIFF: discrete/paletted only, not suited to a continuous ramp.

## Unresolved Questions
On a constant raster (min == max) the statistics sidecar is still written but the `.qml` is skipped, since a single-stop ramp is degenerate.

Follows the categorical sidecar work (#3482 / #3483).

brendancol:
Closes #3537.

Adds an opt-in `color_ramp` parameter to `to_geotiff`. For a continuous single-band raster it writes two sidecars next to the file so QGIS opens it with a color ramp instead of a flat grayscale stretch:

- `<base>.qml` — a QGIS `singlebandpseudocolor` style (extension-replacing name, which is what QGIS auto-loads).
- `STATISTICS_MINIMUM/MAXIMUM/MEAN/STDDEV` in the PAM `<file>.aux.xml`, so GDAL tools and QGIS get a min/max stretch even if the `.qml` is ignored.

This is the continuous-data counterpart to the categorical RAT sidecar (#3482 / #3483). Categorical rasters keep the RAT sidecar; the two are mutually exclusive and never collide.

## Usage
```python
dem.xrs.to_geotiff('dem.tif', color_ramp='viridis')
# writes dem.tif, dem.qml, dem.tif.aux.xml
```
Ramps ship as hardcoded control points (viridis default, plus plasma, magma, inferno, cividis, greys, spectral, terrain), so there is no matplotlib runtime dependency. `color_ramp_range=(min, max)` sets the stretch bounds and skips the statistics pass for large dask graphs.

## Backend coverage
Statistics come from a single backend-aware reduction pass; the dask path fuses the reductions into one `dask.compute` so the graph is read once. Covered across numpy, cupy, dask+numpy, dask+cupy.

## Gating
- Opt-in only (no `color_ramp` → no sidecars, behaviour unchanged).
- Continuous only (skips a raster carrying `category_names`).
- Single-band only; file-like destinations skipped; no-op when there is no finite data.
- Constant raster (min == max): writes the stats sidecar, skips the degenerate `.qml`.
- Unknown ramp name raises `ValueError` before any bytes are written.

## Test plan
- [x] `pytest xrspatial/geotiff/tests/write/test_symbology_sidecar_3537.py` (19 passed, GPU + dask+cupy included locally)
- [x] Categorical regression: `test_category_sidecar_backends_3483.py`, `test_rasterize_categorical_3482.py` (21 passed)
- [x] Manual: wrote a DEM, confirmed `dem.tif` / `dem.qml` / `dem.tif.aux.xml`; `.qml` parses with correct classificationMin/Max; stats exclude the NaN hole

### DIFF
--- a/xrspatial/geotiff/_pam.py
+++ b/xrspatial/geotiff/_pam.py
@@ -104,6 +104,33 @@ def write_pam_sidecar(path, category_names, category_colors=None):
     return out
 
 
+def build_stats_pam_xml(stats):
+    """Build a PAM ``.aux.xml`` document carrying band statistics.
+
+    *stats* maps GDAL ``STATISTICS_*`` keys to numeric values. GDAL tools and
+    QGIS read these from the band ``<Metadata>`` to drive a default min/max
+    stretch on a continuous raster, the way :func:`build_pam_xml` drives class
+    coloring on a categorical one.
+    """
+    lines = ['<PAMDataset>', '  <PAMRasterBand band="1">', '    <Metadata>']
+    for key, value in stats.items():
+        text = escape('{:.10g}'.format(float(value)))
+        lines.append(f'      <MDI key="{escape(str(key))}">{text}</MDI>')
+    lines.append('    </Metadata>')
+    lines.append('  </PAMRasterBand>')
+    lines.append('</PAMDataset>')
+    return '\n'.join(lines) + '\n'
+
+
+def write_stats_pam_sidecar(path, stats):
+    """Write the statistics PAM sidecar for *path* and return its path."""
+    xml = build_stats_pam_xml(stats)
+    out = sidecar_path(path)
+    with open(out, 'w', encoding='utf-8') as fh:
+        fh.write(xml)
+    return out
+
+
 def read_pam_sidecar(path):
     """Read ``category_names`` / ``category_colors`` from *path*'s sidecar.
 
--- a/xrspatial/geotiff/_symbology.py
+++ b/xrspatial/geotiff/_symbology.py
@@ -1,0 +1,272 @@
+"""Symbology sidecars for continuous (non-categorical) rasters.
+
+A continuous single-band GeoTIFF opens in QGIS as a flat grayscale stretch
+unless QGIS finds styling next to the file. Two sidecars give it a sensible
+default color ramp on open:
+
+* a QGIS ``.qml`` style file (``<base>.qml``) with a ``singlebandpseudocolor``
+  renderer -- this is what makes QGIS draw real colors, and
+* band statistics (min/max/mean/stddev) in the PAM ``<file>.aux.xml`` sidecar
+  (written via :mod:`._pam`), which GDAL tools and QGIS use for a min/max
+  stretch even when the ``.qml`` is ignored.
+
+This is the continuous-raster counterpart to the categorical RAT sidecar in
+:mod:`._pam`. Color ramps are hardcoded control points (no matplotlib runtime
+dependency); ``viridis`` is the default.
+"""
+from __future__ import annotations
+
+import math
+import os
+
+import numpy as np
+
+from .. import utils
+
+# Named color ramps, each sampled at 9 evenly spaced control points
+# (matplotlib colormaps, committed as static constants so we carry no
+# matplotlib runtime dependency). Each entry is (position 0..1, (r, g, b)).
+_COLOR_RAMPS = {
+    'viridis': [(0.0, (68, 1, 84)), (0.125, (71, 45, 123)), (0.25, (59, 82, 139)),
+                (0.375, (44, 114, 142)), (0.5, (33, 145, 140)),
+                (0.625, (40, 174, 128)), (0.75, (94, 201, 98)),
+                (0.875, (173, 220, 48)), (1.0, (253, 231, 37))],
+    'plasma': [(0.0, (13, 8, 135)), (0.125, (76, 2, 161)), (0.25, (126, 3, 168)),
+               (0.375, (170, 35, 149)), (0.5, (204, 71, 120)),
+               (0.625, (230, 108, 92)), (0.75, (248, 149, 64)),
+               (0.875, (253, 197, 39)), (1.0, (240, 249, 33))],
+    'magma': [(0.0, (0, 0, 4)), (0.125, (29, 17, 71)), (0.25, (81, 18, 124)),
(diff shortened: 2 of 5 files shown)


brendancol (commented):
## PR Review: Write continuous-raster symbology sidecars from to_geotiff (#3537)

Reviewed the four changed files in full on the PR branch. The feature is well-scoped and mirrors the categorical sidecar pattern cleanly.

None.

- `xrspatial/geotiff/_writers/eager.py:481` — `_sym_data = data` captures the original DataArray, but `_pack` rewrites `data` later at line 544. When `pack=True` and `color_ramp` are combined, the statistics and the `.qml` `classificationMin/Max` are computed from the logical (pre-pack) values, while the TIFF stores the packed integers. QGIS renders against the raw stored pixels, so the ramp would be stretched over the wrong range. `pack` is experimental and the combination is unusual, but it produces a silently wrong-looking render. Worth either skipping symbology when `pack=True` (with a one-line note) or documenting the interaction in the `color_ramp` docstring.

- `xrspatial/geotiff/_symbology.py` — the `greys` ramp follows matplotlib's orientation (0 -> white, 1 -> black), so low elevations render light and high render dark. That is the matplotlib convention, but it is the opposite of what many DEM users expect. Not wrong, just worth a mention in the docstring if it is easy.
- No end-to-end test exercises the `.vrt` write path with `color_ramp` (the closure is shared, so the unit + numpy/dask/gpu coverage already protects it, but a one-liner would close the gap).

### What looks good
- `_finite_stats` fuses the dask reductions into a single `dask.compute`, so the graph is read once rather than four times. Backend parity is verified by an actual test across numpy/dask/cupy/dask+cupy.
- Gating is thorough and tested: categorical wins, multiband skipped, file-like ignored, unknown ramp raises before any bytes, constant raster writes stats but skips the degenerate QML.
- The `qml_path` extension-replace vs PAM append distinction is called out with a comment explaining why a wrong name is silently ignored by QGIS — the exact trap to flag.
- Ramps are static constants, no matplotlib runtime dependency.

### Checklist
- [x] Algorithm/format matches QGIS singlebandpseudocolor + GDAL STATISTICS_*
- [x] All four backends produce consistent statistics
- [x] NaN handling correct (finite mask excludes NaN and the nodata sentinel)
- [x] Edge cases covered (all-NaN, constant, nodata exclusion, multiband, file-like)
- [x] Dask path reads the graph once, no premature materialization
- [x] Benchmark not needed (opt-in sidecar I/O)
- [x] README matrix not applicable (parameter on existing function)
- [x] Docstrings present for the new params and helpers

brendancol (commented):
## Follow-up review (after 7ceb4b3e)

Re-reviewed against the follow-up commit. All findings from the first pass are resolved.

- **Suggestion — pack/color_ramp range mismatch** — fixed. `_writers/eager.py` now gates symbology on `not pack` (with a comment), so a packed write never emits sidecars whose range disagrees with the stored pixels. Documented in the `color_ramp` docstring. Covered by `test_pack_skips_symbology`.
- **Nit — greys orientation** — fixed. The `color_ramp` docstring now notes greys follows matplotlib's light-to-dark orientation.
- **Nit — no VRT end-to-end test** — fixed. Added `test_vrt_write_emits_sidecars`, which writes a dask-backed array to a `.vrt` path and asserts both `mosaic.qml` and `mosaic.vrt.aux.xml`.

### Verification
- `pytest xrspatial/geotiff/tests/write/test_symbology_sidecar_3537.py` — 21 passed (numpy, dask, GPU, dask+cupy parity).
- Categorical regression suites still green.
- flake8 clean on all four changed files.

PR #3548, opened 2026-06-27
classify: return all-NaN instead of crashing on all-non-finite input (#3510)

### LINKED ISSUE #3510: classify: head_tail_breaks, percentiles, box_plot crash on all-NaN / all-inf input

brendancol:
Closes #3510

## What

`head_tail_breaks`, `percentiles`, and `box_plot` raised an opaque reduction error when every pixel was non-finite (all NaN, or all +/-inf, which is mapped to NaN before binning). Each one took a max or percentile over the finite subset of the data, and on this input that subset is empty.

- `head_tail_breaks` -> `ValueError: zero-size array to reduction operation fmax`
- `percentiles` -> `IndexError: index -1 is out of bounds for axis 0 with size 0`
- `box_plot` -> `ValueError: zero-size array to reduction operation fmax`

They now guard the empty-finite case and return an all-NaN array, matching `equal_interval`, `std_mean`, and `maximum_breaks`.

## Backends

numpy, cupy, dask+numpy, dask+cupy. The guard sits in the shared bin/percentile helpers and the eager box_plot branch, so it covers all four.

## Test plan

- [x] Flipped the xfail regression tests (added by the test-coverage sweep) to plain assertions
- [x] Added all-inf coverage for the three functions
- [x] Full test_classify.py passes (117)


brendancol (commented):
## PR Review: classify all-non-finite handling (#3510)

Read the full classify.py context, not just the diff. The fix is correct and the backend coverage holds up.

### Blockers
None.

### Suggestions
- `percentiles` now reaches `np.nanmax(clean)` in the wrapper on all-NaN input (the crash that used to preempt it is gone), so it emits a benign `RuntimeWarning: All-NaN slice encountered`. `std_mean` has the same behaviour and its test suppresses it with `warnings.catch_warnings()`. For consistency, `test_percentiles_all_nan` / `test_percentiles_all_inf` could do the same (xrspatial/tests/test_classify.py). It does not fail CI -- RuntimeWarning is not promoted to error in setup.cfg -- so this is housekeeping, not a correctness issue. `head_tail_breaks` and `box_plot` short-circuit before any nan-reduction and stay warning-free.

### Nits
None.

### What looks good
- The guard sits in the shared helpers (`_compute_head_tail_bins`, `_run_percentiles`, `_run_dask_percentiles`, `_box_plot_bins_from_sample`) plus the eager `_run_box_plot` branch, so a single guard per code path covers all four backends without duplicating logic. Verified the box_plot mapper routes numpy/cupy through `_run_box_plot` and both dask paths through `_box_plot_bins_from_sample`.
- Returning a `[nan]` bin (or `_bin(agg, [nan], [0])`) is the same degradation `equal_interval`/`std_mean`/`maximum_breaks` already use, so behaviour is uniform across classifiers.
- all-inf coverage uses both `+inf` and `-inf`, exercising the `isinf -> nan` mapping.
- Full test_classify.py passes (117), including cupy and dask+cupy on a CUDA box.

### Checklist
- [x] Degradation matches sibling classifiers
- [x] All four backends produce all-NaN consistently
- [x] NaN/inf handling correct (isinf catches both signs)
- [x] Edge case (all-non-finite) covered by tests
- [x] Dask sample paths guarded
- [x] No new public API (no README/docstring change needed)

brendancol (commented):
## Follow-up review (#3510)

Re-reviewed after the follow-up commit and the clean merge of origin/main.

- Suggestion (percentiles all-NaN RuntimeWarning): **fixed** in c94d36e5. `test_percentiles_all_nan` and `test_percentiles_all_inf` now wrap the call in `warnings.catch_warnings()` the same way `test_std_mean_all_nan` does. Confirmed by running the all-NaN/all-inf subset with `-W error::RuntimeWarning`: 16 passed, no unsuppressed warnings.

### New state
- Merged origin/main (geotiff symbology sidecar commit, unrelated to classify). No conflicts. Full test_classify.py still passes (117).

No remaining findings.

PR #3550, opened 2026-06-27
Add global-projection templates to from_template (web_mercator, wgs84/latlon, equal_earth)

### LINKED ISSUE #3549: Add global-projection templates to from_template (web_mercator, wgs84/latlon, equal_earth)
## Summary

`from_template` covers curated regions, world cities, and country codes, but there's no built-in way to get an empty global grid in a chosen world projection. To start from a whole-world canvas in Web Mercator or plain lat/lon today, you have to look up the projected bounds and build the coordinates yourself.

Add named global-projection templates:

- `web_mercator`: global grid in EPSG:3857 (WGS 84 / Pseudo-Mercator)
- `wgs84`: global grid in EPSG:4326 (the same grid the existing `world` returns)
- `latlon`: alias for `wgs84`
- `equal_earth`: global equal-area grid in EPSG:8857 (Equal Earth)

## Motivation

Web Mercator is what web tiles and slippy maps use, so an empty 3857 canvas is a common place to start. `wgs84`/`latlon` are friendlier spellings for the plain geographic grid. Equal Earth is an equal-area projection meant for thematic global maps (population, climate) where the relative size of regions matters more than their shape.

## On Robinson, Gall-Peters, and Mollweide

These were requested too, but none of them has an EPSG code; they exist only as ESRI codes (Robinson ESRI:54030, Mollweide ESRI:54009, Gall-Peters / Gall Stereographic ESRI:54016). This library's convention is that `attrs['crs']` is an EPSG integer. Storing an ESRI code as a bare int would be ambiguous with EPSG and would break the CF metadata path, which calls `pyproj.CRS.from_epsg`. So they're left out. `equal_earth` covers the equal-area case those projections are usually reached for, and it has a real EPSG code.

## Proposed behavior

```python

from_template("web_mercator").attrs["crs"]   # 3857
from_template("wgs84").attrs["crs"]           # 4326
from_template("latlon").attrs["crs"]          # 4326
from_template("equal_earth").attrs["crs"]     # 8857
```

Each returns an empty NaN-filled global grid that follows the array contract, same as the rest of `from_template`. Projected templates carry metre coordinates; `wgs84`/`latlon` carry degree coordinates. All four backends work (numpy, cupy, dask+numpy, dask+cupy), same as the existing templates.

brendancol:
Closes #3549

Adds whole-world templates to `from_template`, each resolving to a real EPSG code:

- `web_mercator` -> EPSG:3857 (WGS 84 / Pseudo-Mercator)
- `equal_earth` -> EPSG:8857 (Equal Earth, the equal-area option)
- `wgs84` and `latlon` -> aliases for the existing EPSG:4326 `world` grid (single source of truth via a small `_REGION_ALIASES` map)

Robinson, Gall-Peters, and Mollweide were requested but only have ESRI codes, not EPSG. Storing those as a bare int would be ambiguous with EPSG and would break the CF metadata path, so they are left out; `equal_earth` covers the equal-area case with a real EPSG code. The reasoning is written up in the issue.

The new names go through the same backend dispatch as the rest of `from_template`, so numpy, cupy, dask+numpy, and dask+cupy all work with no new backend-specific code. The dask path stays lazy (`da.full`).

## Test plan

- [x] `pytest xrspatial/tests/test_templates.py` (80 passed, including the cupy and dask backends that run on this box)
- [x] Contract checks for all four names (dims, all-NaN, float32, crs int, north-up coords)
- [x] `wgs84`/`latlon` produce the same grid and attrs as `world`
- [x] Case-insensitive lookup, exact-resolution honoring, metre coords within bounds
- [x] CF grid-mapping: 3857 and 8857 have no `grid_mapping_name`, so `crs_wkt` carries the name
- [x] Verified the web_mercator dask grid stays lazy


brendancol (commented):
## PR Review: Add global-projection templates to from_template

None.

- `xrspatial/_template_data.py`: the `web_mercator` and `equal_earth` entries carry `area_epsg=8857` / `shape_epsg=3395`, so `from_template('web_mercator', preserve='shape')` will reproject a global template into yet another projection. That combination works but is odd, and nothing tests it. Either add a one-line build test or drop the keys if preserve-on-global is out of scope.

- `xrspatial/_template_data.py`: the Web Mercator latitude limit is written as `85.051129`, but the exact value is `85.0511287798`. It only matters on the rarely-used preserve path, so it is cosmetic. The nearby comment cites the precise definition, which makes the rounded literal look slightly off.

### What looks good
- `wgs84`/`latlon` reuse the `world` grid through `_REGION_ALIASES` rather than copying the bounds, so there is one source of truth.
- All four projections resolve to real EPSG ints, matching the project convention. The ESRI-only projections (Robinson, Gall-Peters, Mollweide) are documented as deliberately left out.
- No new backend code: coverage comes from the existing `_make_data` dispatch, and the dask path stays lazy via `da.full`.
- Tests cover the array contract, the alias-equals-world check, case-insensitive lookup, exact resolution, bounds containment, and the CF grid-mapping behavior for 3857/8857.

### Checklist
- [x] Algorithm matches reference (EPSG codes and projected bounds verified with pyproj)
- [x] All implemented backends produce consistent results (shared dispatch)
- [x] NaN handling is correct (NaN-filled, asserted)
- [x] Edge cases covered (case-insensitive, exact resolution, bounds)
- [x] Dask chunk boundaries handled (lazy da.full, no overlap needed)
- [x] Benchmark not needed (no compute kernel)

brendancol (commented):
## Follow-up review (after commit b1719cbf)

Both items from the first pass are resolved.

- Suggestion (preserve on global templates): kept the `area_epsg`/`shape_epsg` hints, since they match the `world` entry and steer `preserve='shape'` to World Mercator (EPSG:3395) instead of a stray UTM zone for the (0, 0) centroid. Added `test_global_preserve_picks_world_projection` to lock that in for both `web_mercator` and `equal_earth`.
- Nit (latitude limit): the Web Mercator lonlat band now uses the exact `85.0511287798`, matching the comment.

No new findings. Tests: 82 passed locally (numpy, cupy, dask backends).

PR #3551, opened 2026-06-27
fix: return all-NaN pycnophylactic output for empty valid masks

### LINKED ISSUE #3406: pycnophylactic raises opaque ValueError on all-NaN zones or no matching zone id

gaoflow:
## Summary
- return the initialized all-NaN surface when no pixels match any pycnophylactic zone
- convert the existing #3406 strict-xfail regressions into passing tests

Fixes #3406.

## Tests
- `uv run --with '.[tests]' pytest xrspatial/tests/test_dasymetric.py::TestPycnophylacticEmptyValid -q`\n- `uv run --with '.[tests]' pytest xrspatial/tests/test_dasymetric.py::TestPycnophylactic xrspatial/tests/test_dasymetric.py::TestPycnophylacticEmptyValid -q`

### DIFF
--- a/xrspatial/dasymetric.py
+++ b/xrspatial/dasymetric.py
@@ -644,6 +644,8 @@ def _pycnophylactic_numpy(zones_arr, values_dict, nodata_zone,
 
     # pixels that belong to some zone (valid for smoothing)
     valid = ~np.isnan(surface)
+    if not np.any(valid):
+        return surface
 
     for _ in range(max_iterations):
         # Laplacian smoothing: mean of 4-connected neighbours
--- a/xrspatial/tests/test_dasymetric.py
+++ b/xrspatial/tests/test_dasymetric.py
@@ -1070,13 +1070,7 @@ class TestLimitingVariableThreeClass:
 # ---------------------------------------------------------------------------
 
 class TestPycnophylacticEmptyValid:
-    """pycnophylactic crashes when no pixel is valid for smoothing (#3406).
-
-    disaggregate handles the same inputs gracefully (all-NaN output); these
-    are xfail(strict) until #3406 makes pycnophylactic agree.  When the
-    source fix lands, the tests start XPASSing and strict mode flips them
-    red, prompting removal of the marker.
-    """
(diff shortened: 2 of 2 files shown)


brendancol:
hey @gaoflow... thanks for the PR and interest in xarray-spatial.  I love to learn more about you, the stack you are using and your background in computers and geography.

brendancol:
@gaoflow thanks for the interest, but closing this PR due to lack to contributor responsiveness.

PR #3555, opened 2026-06-27
Match cKDTree max_distance check to the brute-force float32 convention (#3392)

### LINKED ISSUE #3392: proximity: bounded dask EUCLIDEAN allocation can NaN a target at a float32-ulp max_distance
**Describe the bug**

Split off from #3389 / #3391, which fixed the CPU-vs-GPU brute-force divergence. A separate, pre-existing edge remains on the bounded dask+numpy path.

For EUCLIDEAN (or MANHATTAN) allocation/direction on a dask array with a finite `max_distance`, a target whose distance lands exactly in the float32 ulp gap of `max_distance` can come back NaN under dask, while the eager numpy backend returns the target. The divergence comes from the bounded `map_overlap` halo-sizing path (`_halo_depth` / chunk handling), not from the per-pixel comparison that #3389 fixed.

**Reproducer**

```python
import dask.array as da
from xrspatial import allocation
from xrspatial.proximity import euclidean_distance

lon = np.array([0.0, 0.1, 0.7]); lat = np.array([0.0])
data = np.zeros((1, 3)); data[0, 0] = 9.0
j = 2
d64 = euclidean_distance(lon[0], lon[j], lat[0], lat[0])
md = (float(np.float32(d64)) + d64) / 2.0   # in the float32 ulp gap

r = xr.DataArray(data.copy(), dims=['lat','lon'], coords={'lon':lon,'lat':lat})
r.data = da.from_array(r.data, chunks=(1, 2))
print(allocation(r, x='lon', y='lat', max_distance=md).data.compute()[0, j])  # nan
# eager numpy / cupy return 9.0 (after #3389)
```

**Expected behavior**

Bounded dask EUCLIDEAN/MANHATTAN allocation/direction should agree with the eager backends at a float32-ulp `max_distance`.

**Severity**

Low. Surfaces only at sub-ulp `max_distance` precision, an adversarial input. Filed for tracking so it is not lost.

**Found by** /sweep-accuracy on the proximity module; split out of #3389.

brendancol:
Closes #3392

Bounded EUCLIDEAN/MANHATTAN allocation (and proximity/direction) on a dask
array could return NaN for a target whose distance sits one float32 ulp above
`max_distance`, while the eager numpy/cupy backends returned the target.

## Root cause

The cKDTree backends compared the float64 distance against `max_distance`, but
the brute-force and CUDA kernels round the distance to float32 before the
`dist <= max_distance` check (the convention #3389 established). When
`max_distance` lands in the float32 ulp gap just below a target's distance, the
two conventions disagree.

The dask reproducer reaches the cKDTree path because `max_possible_distance` is
itself float32-rounded, so a sub-ulp bounded request gets routed to the
global-tree path rather than the halo'd `map_overlap` path. The same
float64/float32 mismatch affects every cKDTree consumer, including eager numpy
proximity, so the fix lives in the shared bound helper instead of the routing.

## Fix

`_inclusive_upper_bound` now widens the query bound to the midpoint between the
largest float32 `<= max_distance` and the next float32 up. That midpoint is the
supremum of float64 distances that still round to a float32 value within range,
so the cKDTree decision matches the kernels and nothing the kernels exclude is
admitted. The absolute floor that keeps the p=2 boundary inclusive at small
`max_distance` (#3442) is preserved.

## Backend coverage

- numpy, dask+numpy: fixed (these use cKDTree).
- cupy, dask+cupy: already used the float32 brute-force convention, unchanged.
- Verified all four backends return the target at the boundary.

## Test plan

- [x] New regression tests: dask+numpy proximity/allocation/direction match
  eager numpy at a float32-ulp `max_distance`, for EUCLIDEAN and MANHATTAN
  (they fail without the fix).
- [x] Mirror tests: a `max_distance` one float32 ulp below the target's
  distance still excludes it on both backends.
- [x] Full `xrspatial/tests/test_proximity.py` passes (552 tests).

### DIFF
--- a/xrspatial/proximity.py
+++ b/xrspatial/proximity.py
@@ -750,33 +750,45 @@ def _process_dask_cupy(raster, x_coords, y_coords, target_values,
 
 
 def _inclusive_upper_bound(max_distance):
-    """Exclusive ``distance_upper_bound`` that cKDTree treats as inclusive.
+    """Exclusive cKDTree ``distance_upper_bound`` matching the float32 keep test.
 
-    ``cKDTree.query``'s ``distance_upper_bound`` is exclusive, while the
-    brute-force and CUDA kernels keep ``dist <= max_distance``. To turn the
-    exclusive check into the inclusive one the bound has to be widened just
-    past ``max_distance``.
+    The brute-force and CUDA kernels round each candidate distance to float32
+    and keep it when ``np.float32(dist) <= max_distance`` (see
+    ``_process_numpy_bruteforce`` and ``_proximity_cuda_kernel``). cKDTree
+    instead compares *float64* distances against an *exclusive* bound, so to
+    reproduce the kernels' keep decision the bound has to account for both
+    differences.
 
-    A one-ulp widen (``np.nextafter``) is not enough for the EUCLIDEAN query
-    (``p=2``): cKDTree compares *squared* distances internally, and the square
-    of ``nextafter(0.0, inf)`` (the smallest subnormal) underflows back to
-    0.0, so the bound collapses to exactly ``max_distance`` again and a
-    target sitting exactly at the bound -- most visibly a target pixel itself
-    at ``max_distance=0`` -- is dropped to NaN. numpy/dask then disagreed with
-    the cupy/dask+cupy brute-force backends, which include it.
+    * Float32 rounding: a target whose true float64 distance rounds *down*
+      across a float32 step to a value ``<= max_distance`` is kept by the
+      kernels but, against a float64-exact bound, dropped by cKDTree to NaN.
+      This surfaces when ``max_distance`` sits in the float32 ulp gap just
+      below a target's distance (issue #3392). The bound is therefore widened
+      to the midpoint between the largest float32 ``<= max_distance`` and the
+      next float32 up: the supremum of float64 distances that still round to a
+      float32 value ``<= max_distance``. Nothing beyond that midpoint is pulled
+      in, so no target the kernels exclude is admitted.
 
-    Widen by a relative epsilon with an absolute floor instead. The bump stays
-    large enough to survive squaring for ``p=2`` (and is harmless for ``p=1``)
-    while staying far below the next representable distance, so nothing beyond
-    ``max_distance`` is pulled in. Because the bump is relative, a target at a
-    true distance within roughly ``8 * eps * max_distance`` of the bound could
-    qualify, but at that magnitude the two distances are not distinguishable in
-    float64 anyway, so the result is still correct.
+    * Inclusive boundary for ``p=2``: cKDTree compares *squared* distances
+      internally, and the square of ``nextafter(0.0, inf)`` (the smallest
+      subnormal) underflows back to 0.0, collapsing the bound to exactly
+      ``max_distance`` and dropping a target sitting on the bound -- most
+      visibly a target pixel itself at ``max_distance=0``. A relative bump with
+      an absolute floor stays large enough to survive squaring for ``p=2`` (and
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Match cKDTree max_distance check to the brute-force float32 convention (#3392)

Reviewed the full `_inclusive_upper_bound` change and the new tests against the
surrounding proximity code. The root-cause analysis holds up: I traced the
reproducer and it does reach `_process_dask_kdtree` (not the `map_overlap` halo
path the issue text guessed) because `max_possible_distance` is computed through
`_distance`, which returns float32. Putting the fix in the shared bound helper is
the right call since it covers every cKDTree consumer at once.

None.

- The new tests pin numpy against dask+numpy only. The same boundary is worth
  asserting on cupy / dask+cupy so the four backends can't drift apart again
  (the older #3389 `*_gpu_matches_cpu` tests cover GPU allocation, and the #3442
  block parametrizes all four backends via `create_test_raster`). Not blocking
  because the fix is unreachable from the brute-force GPU paths and the helper
  is backend-independent, but a 4-backend parametrization would close the gap.

- `xrspatial/tests/test_proximity.py:2670` uses
  `np.linalg.norm([lon[j] - lon[0]], ord=p)` to get the distance. For a
  one-element vector that is just `abs(lon[j] - lon[0])` for both p=1 and p=2;
  a plain `abs(...)` reads more directly. The current form is correct, only
  slightly indirect.

### Out of scope (mention only)
- `max_possible_distance` is still float32 (via `_distance` at
  `xrspatial/proximity.py:1388`), so a sub-ulp bounded request is routed to the
  global-tree path instead of `map_overlap`. With this fix the result is correct
  either way, so it is a routing/perf nuance rather than a bug. Fine to leave;
  flagging so it is on record.
- Pre-existing flake8 E127 at `xrspatial/tests/test_proximity.py:1134` and
  `:1160`, unrelated to this PR.

### What looks good
- The midpoint bound is the exact supremum of float64 distances that round to a
  float32 value within range, so it matches the kernels without admitting
  anything they exclude. Verified the widen is half a float32 ulp across
  magnitudes and the absolute floor still keeps p=2 inclusive at small
  max_distance (md=0 -> 8.88e-16, squaring-safe).
- Regression tests fail without the change (6 of them) and the mirror
  "just below" tests confirm a genuinely out-of-range target still NaNs out.
- Manually confirmed all four backends return the target at the boundary.

### Checklist
- [x] Algorithm matches the established float32 keep convention (#3389)
- [x] NaN handling is correct (inclusion widened, exclusion preserved)
- [x] Edge cases covered (md=0 floor, inf, just-below boundary)
- [x] Dask paths handled (shared helper covers global/tiled and chunk fns)
- [x] Benchmark not needed (scalar helper, no perf-critical change)
- [x] Docstring updated to explain the float32 reasoning

brendancol (commented):
## Follow-up review (after 917dc90)

Re-reviewed the diff after the follow-up commit.

- Suggestion (4-backend coverage): **fixed.** The keep/exclude tests now
  parametrize over numpy, dask+numpy, cupy, and dask+cupy via the `_ulp_raster`
  helper, so the brute-force GPU backends are pinned to the cKDTree backends at
  the boundary. 36 boundary cases pass locally (CUDA backends included).
- Nit (`np.linalg.norm` on a one-element vector): **fixed.** Replaced with the
  plain longitude difference, which is the distance for both p=1 and p=2 on the
  constant-latitude row.
- `max_possible_distance` float32 routing and the pre-existing flake8 E127s:
  **left as-is**, both noted in the first review as out of scope (correct either
  way / not introduced here).

### Status
No blockers, no remaining actionable findings. Full `test_proximity.py` passes
(579 tests across all four backends).

PR #3556, opened 2026-06-27
from_template: lazy dask grids skip the cell cap when chunks is supplied

### LINKED ISSUE #3554: from_template: lazy dask grids should skip the cell-count cap when chunks is supplied
**Is your feature request related to a problem? Please describe.**

`from_template` enforces a hard 500-million-cell cap on every backend, dask included. Asking for a fine-resolution grid over a large region with a chunked backend trips the cap even though a dask grid is lazy and never gets built:

```python
>>> study_area = xrspatial.from_template('new_england', chunks=512, resolution=10)
ValueError: resolution (10.0, 10.0) produces a 89218 x 77280 grid
(6,894,767,040 cells), exceeding the 500,000,000-cell limit. Use a coarser resolution.
```

The cap is there to stop an eager `np.full`/`cupy.full` from eating all the memory, but it runs before the backend is even looked at. `da.full` only builds a chunk graph, so a 6.9-billion-cell NaN grid costs next to nothing until you compute it. Separately, `chunks=512` was passed but `backend` defaults to `numpy`, so the chunk spec is silently dropped -- even past the cap, the call would try to allocate ~27 GB.

**Describe the solution you'd like**

Passing `chunks` should give back a lazy dask grid. Concretely: when `chunks` is supplied, promote an eager backend to its dask variant (`numpy` to `dask+numpy`, `cupy` to `dask+cupy`), and apply the cell-count cap only to the eager backends.

**Describe alternatives you've considered**

Keep the uniform cap and tell the user to coarsen the resolution. That defeats the point of a chunked backend, where a large lazy grid is the whole workflow.

**Additional context**

The eager cap is unchanged: `from_template('conus', resolution=1)` on the default numpy backend still raises.

brendancol:
Closes #3554

Supplying `chunks` now gives back a lazy dask grid instead of silently dropping the chunk spec, and the 500M-cell cap only guards the eager backends.

- When `chunks` is passed, an eager backend is promoted to its dask variant (`numpy` to `dask+numpy`, `cupy` to `dask+cupy`), so the result is lazy and chunked.
- The cell cap applies only to numpy and cupy, which materialize the whole grid. A dask grid is a lazy `da.full` chunk graph, so it skips the cap. The eager cap is unchanged: `from_template('conus', resolution=1)` on numpy still raises.
- The `chunks` default changes from `'auto'` to `None` as a "not supplied" sentinel; the dask paths still use `'auto'` when it is omitted.

Backend coverage: numpy and cupy stay eager and capped; dask+numpy and dask+cupy are lazy and uncapped. The promotion paths exercise both dask variants.

Test plan:

- [ ] `from_template('new_england', chunks=512, resolution=10)` returns a lazy dask array with chunksize (512, 512) and no cap error
- [ ] an explicit `backend='dask+numpy'` over the cap is lazy without chunks
- [ ] `chunks` promotes cupy to dask+cupy on GPU
- [ ] eager cap unchanged: `from_template('conus', resolution=1)` still raises

### DIFF
--- a/xrspatial/templates.py
+++ b/xrspatial/templates.py
@@ -313,7 +313,11 @@ def from_template(name: str,
         Dask chunk specification. Supplying it returns a lazy, chunked grid:
         an eager backend is promoted to its dask variant (``'numpy'`` to
         ``'dask+numpy'``, ``'cupy'`` to ``'dask+cupy'``), and the cell cap no
-        longer applies. When omitted, the dask backends use ``'auto'``.
+        longer applies. When omitted, the dask backends use ``'auto'``. The
+        data stays lazy, but a very fine resolution still builds one task per
+        chunk, so an extreme shape with small chunks can make a task graph
+        large enough to bog down the client; coarsen the resolution or use
+        larger chunks if that happens.
 
     Returns
     -------
@@ -359,6 +363,10 @@ def from_template(name: str,
         32630
         >>> from_template("FRA", preserve="area").attrs["crs"]    # Equal Earth
         8857
+        >>> # passing chunks returns a lazy dask grid, exempt from the cell cap
+        >>> agg = from_template("new_england", resolution=10, chunks=512)
+        >>> type(agg.data).__name__
+        'Array'
     """
     spec = _resolve(name)
     key = spec["key"]
--- a/xrspatial/templates.py
+++ b/xrspatial/templates.py
@@ -22,8 +22,10 @@ from xrspatial.reproject._grid import _edge_samples, _make_output_coords, _trans
 _PRESERVE_OPTIONS = ("area", "shape")
 
 # Guard against a stray fine resolution allocating an enormous array. Applied to
-# every backend, including dask: a lazy grid this large is almost always a typo,
-# and the uniform cap keeps the error identical across backends.
+# the eager backends (numpy, cupy) only: those materialize the whole grid up
+# front, so a huge shape is an immediate out-of-memory risk. A dask grid is
+# lazy (da.full builds a chunk graph, never the array), so the cap is skipped
+# there -- a large chunked study area is a legitimate workflow.
 _MAX_CELLS = 500_000_000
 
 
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: lazy dask grids skip the cell cap when chunks is supplied

### Blockers
None.

### Suggestions
- Dropping the cap for dask leaves that path with no guard at all, and the new footgun is chunk count rather than memory. With a fixed `chunks=512`, the old 500M-cell cap also bounded the task graph at ~1900 chunks. An over-fine resolution now builds an unbounded graph: `from_template('conus', resolution=1, chunks=512)` is ~50 million chunks, and graph construction alone hangs the client even though no data is materialized. The lazy grid is cheap; building tens of millions of tasks is not. Worth a sanity check on chunk count (not cells), or at least a docstring note that an over-fine resolution on a dask backend can build a huge task graph. (templates.py:390-402)

### Nits
- The Examples block in the `from_template` docstring does not show the chunks/lazy path. A one-line example would make the new behavior discoverable. (templates.py Examples)

### What looks good
- Minimal and surgical. The cap is now backend-aware, the stale comment that claimed it applied to dask is fixed, and the error message points users at `chunks=`.
- Tests cover the three new paths (numpy promotion, explicit dask without chunks, cupy promotion) plus the eager cap staying intact. The over-cap region is picked to keep the chunk count small so the tests run fast.

### Checklist
- Algorithm matches reference/paper: n/a (grid construction)
- All implemented backends consistent: yes, promotion verified for numpy and cupy
- NaN handling correct: yes, fill unchanged
- Edge cases covered: eager cap retained; lazy paths tested
- Dask chunk boundaries: n/a (empty grid)
- No premature materialization: yes, da.full stays lazy
- Benchmark: not needed
- README matrix: not needed (no backend support change)

brendancol (commented):

- Nit (docstring example): fixed in 587dfe2 -- added an Examples entry for the lazy chunked path.
- Suggestion (dask task-graph footgun): split. The documentation half is fixed in 587dfe2 (the `chunks` param now warns that a very fine resolution builds a large task graph and suggests coarser resolution or larger chunks). The actual chunk-count guard is a separate design decision (threshold, `chunks='auto'` handling, raise vs warn) and reintroduces a gate the issue asked to remove for dask, so it is deferred to #3557.

The follow-up commit is docstring-only; no logic change. The new Examples line evaluates (`type(agg.data).__name__ == 'Array'`) and the cap tests still pass.

Remaining open item is the deferred guard (#3557). Nothing blocking on this PR.

PR #3560, opened 2026-06-27
Guard the datashader import in the rasterizer benchmark

### LINKED ISSUE #3452: Guard the datashader import in the rasterizer benchmark
## Reason or Problem
`benchmarks/rasterizer_benchmarks.py` compares xarray-spatial against datashader, geocube, and rasterio. The datashader comparison is the point of the benchmark, so datashader stays referenced here. But once datashader is no longer installed by default, the unconditional `import datashader as ds` makes the whole script fail to run.

## Proposal
- Wrap the datashader import in a try/except so the benchmark runs without it, skipping the datashader column when it is absent and including it when it is present.

## Value
Keeps the A/B benchmark working both with and without datashader installed.

## Done when
- The benchmark runs in an environment without datashader (datashader column skipped) and in one with datashader (full comparison).

Part of removing datashader as a dependency; this file keeps a deliberate datashader reference.

brendancol:
Closes #3452.

`benchmarks/rasterizer_benchmarks.py` keeps datashader as a deliberate A/B comparison, so it stays referenced. But once datashader is no longer installed by default, the unconditional `import datashader as ds` made the whole script fail to run. This wraps that import (plus spatialpandas, which only feeds the datashader path) in a try/except.

- With datashader present, `HAS_DATASHADER` is True and the full comparison runs unchanged.
- With it absent, `HAS_DATASHADER` is False, the polygon path skips the datashader timing run, and the datashader cells show `--` (same as how the cupy column already handles a missing backend).

Backend coverage: not applicable, this is a benchmark script rather than a spatial op.

Test plan:
- With datashader installed: confirmed `HAS_DATASHADER` is True and datashader takes part in both the timing and consistency runs.
- With datashader and spatialpandas blocked at import: confirmed `HAS_DATASHADER` is False, the script imports, the xrspatial/geocube/rasterio runners and consistency checks still work, and the markdown report builds.
- `py_compile` passes.
- No packaged unit test: the standalone benchmark scripts have no pytest coverage, and importing this one pulls in benchmark-only deps (geocube, spatialpandas) that the CI test env does not have.

https://claude.ai/code/session_01TF3CsJ7nee9agjajEWLw5d

### DIFF
--- a/benchmarks/rasterizer_benchmarks.py
+++ b/benchmarks/rasterizer_benchmarks.py
@@ -21,7 +21,6 @@ from datetime import datetime, timezone
 
 import numpy as np
 import geopandas as gpd
-import spatialpandas
 from rasterio.features import rasterize as rio_rasterize
 from rasterio.transform import from_bounds
 from shapely.geometry import (
@@ -29,9 +28,19 @@ from shapely.geometry import (
     box,
 )
 
-import datashader as ds
 from geocube.api.core import make_geocube
 from xrspatial.rasterize import rasterize as xrs_rasterize
+
+# datashader is a deliberate A/B comparison here but is no longer a default
+# dependency. Guard the import (and spatialpandas, which only feeds the
+# datashader path) so the benchmark still runs when it is absent; the
+# datashader column is skipped in that case.
+try:
+    import datashader as ds
+    import spatialpandas
+    HAS_DATASHADER = True
+except ImportError:
+    HAS_DATASHADER = False
 
 
 # ---------------------------------------------------------------------------
@@ -598,7 +607,7 @@ def main():
     for gen_name, gen_cfg in GENERATORS.items():
         gen_fn = gen_cfg["fn"]
         kind = gen_cfg["kind"]
-        use_datashader = kind == "polygon"
+        use_datashader = kind == "polygon" and HAS_DATASHADER
 
         for n in feature_counts:
             geoms, vals = gen_fn(n, rng)


brendancol (commented):
## PR Review: Guard the datashader import in the rasterizer benchmark

None.

None.

- `benchmarks/rasterizer_benchmarks.py` (generate_markdown): when datashader is absent, the polygon timing and ratio tables still print a "datashader" column header filled with `--`. The issue wording says "skip the datashader column". Dropping the column outright for polygons would mean threading `HAS_DATASHADER` into `generate_markdown`. The `--` fallback matches how the cupy column already handles a missing backend, so I read this as a reasonable consistency choice rather than a defect. Flagging it so the maintainer can decide.
- `benchmarks/rasterizer_benchmarks.py:39`: the guard catches only `ImportError`. If datashader is installed but breaks on import for some other reason, that will propagate. Standard practice and fine; noting for completeness.

### What looks good
- spatialpandas and datashader are folded into one guard. That is correct, since spatialpandas only feeds the datashader path, so there is no half-guarded state.
- Every reference to `ds.` and `spatialpandas` (the `run_datashader` body, the `spd_gdf` build) sits behind `use_datashader`, which now requires `HAS_DATASHADER`. No NameError path when datashader is missing.
- `use_datashader = kind == "polygon" and HAS_DATASHADER` is the minimal correct gate.

### Checklist
- Algorithm matches reference/paper: n/a (benchmark infra)
- All implemented backends produce consistent results: n/a
- Edge cases covered: both datashader-present and datashader-absent paths verified
- Benchmark exists or not needed: this is the benchmark
- Docstrings present and accurate: the guard carries an explanatory comment

PR #3562, opened 2026-06-27
from_template ergonomics: coregister your data onto the grid + neighborhood-friendly dask chunks

### LINKED ISSUE #3561: from_template: coregister your own data onto the grid + neighborhood-friendly dask chunks
## Reason or Problem

`from_template` gets you an empty study-area grid, but two common next steps are still awkward:

1. **Coregistering your own raster onto the grid.** Vectors, points, and on-disk GeoTIFFs already have a one-call path (`template.xrs.rasterize(gdf, coregister=True)`, the interpolators with `coregister=True`, `template.xrs.open_geotiff(path, coregister=True)`). But if you already hold a raster `DataArray` in another CRS or grid, there is no equivalent. `template.xrs.reproject(crs)` reprojects the template, not your data, and you would have to hand-thread `bounds`/`width`/`height`/`resolution` into `reproject()` to land on the template grid.

2. **Dask templates do not chunk for neighborhood ops.** `from_template('conus', resolution=1000, backend='dask')` comes back as a single 3105x5865 chunk (no parallelism), and `conus@250m` tiles into 5792-px blocks with ragged 291-px edge slivers. `chunks='auto'` is byte-based, so most templates are either one giant block or have awkward edges. Downstream `slope`/`focal`/etc. inherit that graph.

## Proposal

- Add `template.xrs.coregister(data)`: reproject a raster `DataArray` onto the template's exact grid (CRS + bounds + shape), or rasterize a `GeoDataFrame` onto it. One obvious method for "put my data on this grid".
- Make `chunks='auto'` (the default for dask templates) tile into even, square, map_overlap-friendly blocks instead of one giant chunk or ragged slivers. Explicit `chunks=` is still honored verbatim, and the existing chunk-count guard still applies.
- Document the get-set-up-and-go path in the `from_template` docstring (pick a template, coregister your data, run a function).

## Done when

- `template.xrs.coregister(raster_or_gdf)` returns data on the template grid for all four backends where the underlying reproject/rasterize support them.
- Dask templates come back tiled into balanced square blocks; small templates stay single-chunk.
- Docstrings show the recipe.

brendancol:
## Summary

Two ergonomic additions to `from_template`, so users go from a region name to an analysis-ready grid with less friction.

- **`template.xrs.coregister(data)`**: one call to put your own data on the grid. A raster `DataArray` is reprojected onto the template's exact CRS, bounds, and shape; a `GeoDataFrame` is rasterized onto it (the `rasterize(coregister=True)` path). The result shares the template's `y`/`x` coords, so layers stack cell-for-cell. This fills the gap left by the existing coregister paths, which covered vectors, points, and on-disk GeoTIFFs but not in-memory rasters.
- **Neighborhood-friendly dask chunks**: `chunks='auto'` (the default for dask templates) now tiles into even, square blocks (~2048 cells per side) instead of one giant byte-sized block or thin ragged edge slivers. Downstream `slope`/`focal`/etc. inherit a parallel, overlap-friendly graph. Small grids stay a single chunk; the block grows for very large grids so the default never trips the existing chunk-count cap (#3557).
- Docstring "Recipes" section and a templates reference note showing the pick-a-grid, coregister, run-a-tool path.

Measured before/after for the chunking: `from_template('conus', resolution=1000, backend='dask')` went from a single 3105x5865 chunk to 6 balanced ~1553x1955 blocks; `conus@250m` from 15 blocks with 291-px slivers to 66 even ~2070x2132 blocks.

## Backend coverage

`coregister` runs on whatever backends the underlying reproject/rasterize support (numpy and dask verified; a dask source returns a dask result). The chunk change is dask-only. No new compute kernels.

## Test plan

- [x] coregister: raster lands on grid (exact coords, constant field preserved), vector path equals `rasterize(coregister=True)`, missing-CRS and bad-type errors, nearest resampling, dask source stays dask
- [x] chunks: default tiling balanced and square, small template stays one chunk, block grows for huge grids (stays under the cap), explicit chunks bypass the tiling
- [x] full test_templates.py and test_accessor.py suites pass; reproject and rasterize suites pass
- [x] integration: dask template -> coregister dask raster -> slope stays lazy and computes
- [x] flake8 clean

Closes #3561

https://claude.ai/code/session_01TEXdRkRqYJKvATjquz26Pa

### DIFF
--- a/docs/source/reference/templates.rst
+++ b/docs/source/reference/templates.rst
@@ -25,3 +25,28 @@ From Template
 
     xrspatial.templates.from_template
     xrspatial.templates.list_templates
+
+Putting your data on a template
+===============================
+
+A template is an empty canvas; ``coregister`` fills it with your own data. It
+reprojects a raster :class:`~xarray.DataArray` onto the template's grid, or
+rasterizes a ``GeoDataFrame`` onto it, so every layer lines up cell-for-cell.
+
+.. sourcecode:: python
+
+    from xrspatial import from_template
+
+    grid = from_template("conus", resolution=1000)   # empty Albers grid
+    elevation = grid.xrs.coregister(my_dem)           # raster -> grid
+    roads = grid.xrs.coregister(my_roads_gdf)         # vectors -> grid
+    slope = elevation.xrs.slope()
+
+For an out-of-core run, ask for a dask backend. The template tiles into even,
+square blocks tuned for the neighborhood ops that follow, so the downstream
+task graph stays parallel and overlap-friendly:
+
+.. sourcecode:: python
+
+    grid = from_template("conus", resolution=250, backend="dask")
+    slope = grid.xrs.coregister(my_dem).xrs.slope()   # stays lazy
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -850,6 +850,72 @@ def _rasterize_on_accessor(obj, geometries, *, coregister, **kwargs):
     return rasterize(geometries, like=obj, **kwargs)
 
 
+def _coregister_on_accessor(obj, data, *, resampling, **kwargs):
+    """Put *data* on caller raster *obj*'s grid (CRS + bounds + shape).
+
+    Dispatches on the input type:
+
+    - A GeoDataFrame is rasterized onto the grid, reprojected into the caller
+      CRS first -- the same path as ``rasterize(coregister=True)``. The
+      ``resampling`` argument does not apply to a burn and is ignored.
+    - A raster ``DataArray`` is reprojected and resampled onto the grid.
+
+    Either way the result shares *obj*'s ``y``/``x`` coordinates exactly, so it
+    lines up cell-for-cell with the template and any other layer coregistered
+    the same way.
+    """
+    from xrspatial.interpolate._vector import is_geodataframe
+
+    if is_geodataframe(data):
+        return _rasterize_on_accessor(obj, data, coregister=True, **kwargs)
+    if isinstance(data, xr.DataArray):
+        return _reproject_onto_accessor(obj, data, resampling=resampling,
+                                        **kwargs)
+    raise TypeError(
+        "coregister expects a raster xarray.DataArray or a GeoDataFrame, "
+        f"got {type(data).__name__}"
+    )
+
+
+def _reproject_onto_accessor(target, source, *, resampling, **kwargs):
+    """Reproject raster *source* onto *target*'s exact grid."""
+    import numpy as np
+
+    from .reproject import reproject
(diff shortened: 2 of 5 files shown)


brendancol (commented):
## PR Review: from_template ergonomics (coregister + neighborhood dask chunks)

I read the full diff from the PR worktree and exercised both features, including a few cases the tests do not cover (non-constant gradient orientation, ascending-y target, huge-grid chunk cap).

None.

None.

- `accessor.py` `_reproject_onto_accessor`: `**kwargs` is forwarded to `reproject`, which already gets `bounds`, `width`, `height`, and `resampling` as explicit args. A caller passing any of those through `coregister(..., bounds=...)` would hit a "multiple values for keyword argument" TypeError instead of a friendly message. Low odds, since those are exactly what coregister computes for you. Not worth guarding unless it comes up.
- `accessor.py` ~line 895: the 2x2-cell minimum raises even when `attrs['res']` is present and could anchor a 1x1 grid. Coregistering onto a single pixel is degenerate, so the clear error is fine, just slightly stricter than necessary.

### What looks good
- Orientation is correct, which was the real risk. reproject always emits north-up (descending y, ascending x) no matter the source order, and the follow-up commit flips to the caller's axis directions before the coordinate snap, so an ascending-y target no longer flips silently. Verified with a north-south gradient: the north row stays north.
- Raster path data is byte-identical to a raw `reproject` with the same derived bounds; the coordinate snap only removes float drift, it does not move data.
- Vector path is exactly `rasterize(coregister=True)`, so it inherits that path's CRS handling and tests.
- The chunk default is self-limiting: the block grows for very large grids so `chunks='auto'` stays under the existing 1e6 cap (the #3557 contract holds). conus@1m lands at ~200k blocks instead of erroring.
- Small grids still come back as a single chunk, so the change is invisible for the common small-template case.
- CRS attrs (crs int, crs_wkt, grid_mapping_name) are carried from the template onto the result, so it is a real drop-in rather than keeping reproject's WKT-string crs.

### Checklist
- [x] coregister output lands on the template grid (exact coords, CRS carried over)
- [x] No coordinate flip (orientation verified with a gradient and an ascending-y target)
- [x] Raster data identical to raw reproject; vector path identical to rasterize(coregister=True)
- [x] Dask source returns a dask result; default tiling balanced and square
- [x] chunks='auto' stays under the chunk cap at extreme resolution; explicit chunks bypass the tiling
- [x] Docstrings present (coregister, from_template Recipes) and a templates reference note
- [ ] README feature matrix: N/A (no new top-level function)
- [ ] Benchmark: N/A (thin wrapper over already-benchmarked reproject/rasterize; chunk change is in grid construction, not a hot path)

PR #3564, opened 2026-06-27
Expose multi_stop_search on the .xrs DataArray and Dataset accessors

### LINKED ISSUE #3563: Expose multi_stop_search on the .xrs DataArray accessor
## Description

`xrspatial.multi_stop_search` routes a least-cost path through a sequence of waypoints, but it is not reachable from the `.xrs` DataArray accessor. The `# ---- Pathfinding ----` section of the accessor delegates only `a_star_search`. So `raster.xrs.a_star_search(...)` works, but `raster.xrs.multi_stop_search(...)` does not, and callers have to drop back to the module-level function.

## Proposed solution

Add a `multi_stop_search` delegation method to `XrsSpatialDataArrayAccessor`, mirroring the existing `a_star_search` method. The first positional argument is `waypoints`; everything else forwards as keyword arguments.

```python
raster.xrs.multi_stop_search(waypoints, optimize_order=True)
```

DataArray-only, matching `a_star_search` (the Dataset accessor does not expose pathfinding).

## Scope

- Accessor delegation method under the Pathfinding banner.
- Presence and functional-parity test in `test_accessor.py`.
- Add `multi_stop_search` to the pathfinding reference docs, which currently list only `a_star_search`.

The `multi_stop_search` implementation does not change. It is already public and in the README feature matrix.

brendancol:
Closes #3563

Exposes `multi_stop_search` on the `.xrs` accessor, alongside the existing `a_star_search` method, so multi-waypoint routing is reachable as `raster.xrs.multi_stop_search(waypoints, ...)` without dropping to the module-level function.

- DataArray accessor: new delegation method under the Pathfinding banner in `XrsSpatialDataArrayAccessor`. `waypoints` is positional, everything else forwards as keyword args.
- Dataset accessor: `multi_stop_search` is decorated with `@supports_dataset`, matching `cost_distance` and `proximity`, so `ds.xrs.multi_stop_search(...)` routes each data variable through the same waypoints and returns a Dataset of per-variable results.
- Added `multi_stop_search` to the pathfinding reference docs, which previously listed only `a_star_search`.

`a_star_search` stays DataArray-only (it is not `@supports_dataset` and was not requested here).

Backend coverage: the accessor is a thin delegate, so backend support is whatever `multi_stop_search` already provides (numpy, cupy, dask+numpy, dask+cupy). No change to the routing implementation.

Test plan:
- [x] `test_da_multi_stop_search`: DataArray accessor matches the direct call
- [x] `test_da_multi_stop_search_kwargs`: keyword forwarding (`optimize_order=True`)
- [x] `test_ds_multi_stop_search`: Dataset accessor matches the direct call, per-variable
- [x] `test_da_a_star_search`: functional parity test for the sibling method
- [x] presence checks updated for both accessors
- [x] `test_accessor.py` (90) and `test_pathfinding.py` (51) pass

### DIFF
--- a/docs/source/reference/pathfinding.rst
+++ b/docs/source/reference/pathfinding.rst
@@ -21,3 +21,10 @@ A* Pathfinding
     :toctree: _autosummary
 
     xrspatial.pathfinding.a_star_search
+
+Multi-Stop Routing
+==================
+.. autosummary::
+    :toctree: _autosummary
+
+    xrspatial.pathfinding.multi_stop_search
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -1947,6 +1947,12 @@ class XrsSpatialDatasetAccessor:
         from .surface_distance import surface_direction
         return surface_direction(self._obj, elevation, **kwargs)
 
+    # ---- Pathfinding ----
+
+    def multi_stop_search(self, waypoints, **kwargs):
+        from .pathfinding import multi_stop_search
+        return multi_stop_search(self._obj, waypoints, **kwargs)
+
     # ---- Preview ----
 
     def preview(self, **kwargs):
--- a/xrspatial/accessor.py
+++ b/xrspatial/accessor.py
@@ -1210,6 +1210,10 @@ class XrsSpatialDataArrayAccessor:
         from .pathfinding import a_star_search
         return a_star_search(self._obj, start, goal, **kwargs)
 
+    def multi_stop_search(self, waypoints, **kwargs):
+        from .pathfinding import multi_stop_search
+        return multi_stop_search(self._obj, waypoints, **kwargs)
+
     # ---- Zonal ----
 
     def zonal_stats(self, zones, **kwargs):
--- a/xrspatial/pathfinding.py
+++ b/xrspatial/pathfinding.py
@@ -15,6 +15,7 @@ except ImportError:
     dask = None
 
 from xrspatial.cost_distance import _heap_push, _heap_pop
+from xrspatial.dataset_support import supports_dataset
 from xrspatial.utils import (
     _validate_raster,
     get_dataarray_resolution, ngjit,
@@ -1324,6 +1325,7 @@ def _optimize_waypoint_order(surface, waypoints, barriers, x, y,
     return [waypoints[i] for i in order]
 
 
+@supports_dataset
 def multi_stop_search(surface: xr.DataArray,
                       waypoints: list,
                       barriers: list = [],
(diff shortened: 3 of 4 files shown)


brendancol (commented):
## PR Review: Expose multi_stop_search on the .xrs DataArray accessor

None.

None.

None.

### What looks good
- The accessor method (`xrspatial/accessor.py:1213`) mirrors the sibling `a_star_search` delegate: `waypoints` positional, the rest forwarded as `**kwargs`. The lazy `from .pathfinding import` matches every other method in the class, so it costs nothing at import time.
- DataArray-only placement is right. `a_star_search` is also absent from the Dataset accessor, and `multi_stop_search` is a DataArray operation (it builds a single cost surface).
- The tests check the real contract: accessor output is `assert_identical` to the direct call, including the `optimize_order=True` keyword path. The new `test_da_a_star_search` closes a gap, since the sibling previously had only a presence check.
- The method sits under the existing `# ---- Pathfinding ----` banner, so it appears in the accessor repr and passes the category-banner guard test.
- The docs RST gets a matching entry, and the section underline length is correct, so Sphinx stays quiet.

### Notes
- No README change needed. `multi_stop_search` is already in the feature matrix and exported from `__init__`. No new algorithm, so no benchmark or user-guide notebook.
- Backend coverage is inherited from `multi_stop_search` unchanged (numpy, cupy, dask+numpy, dask+cupy).

### Checklist
- [x] Algorithm matches reference/paper (no algorithm change; pure delegation)
- [x] All implemented backends produce consistent results (inherited)
- [x] Edge cases are covered by tests (delegation parity tested)
- [x] README feature matrix updated (already present)
- [x] Docstrings present and accurate (delegate inherits documentation)

brendancol (commented):
## PR Review (follow-up): Dataset accessor support for multi_stop_search

None.

None.

None.

### What looks good
- `@supports_dataset` on `multi_stop_search` (`xrspatial/pathfinding.py:1328`) is the same mechanism `cost_distance` and `proximity` already use, so the Dataset accessor delegate (`xrspatial/accessor.py:1952`) forwards directly with no special-casing.
- The decorator only injects `name` when the wrapped function has a `name` parameter. `multi_stop_search` has none, so the wrapper calls `func(da, waypoints, **kwargs)` per variable without passing an unsupported kwarg. Confirmed `name` is absent from the signature.
- The DataArray path keeps its old behavior: the wrapper passes a DataArray straight through. The 51 pathfinding tests still pass, so the decorator did not regress single-array routing.
- `test_ds_multi_stop_search` checks the real contract: `assert_identical` against the direct `multi_stop_search(ds, ...)` call, plus a per-variable `data_vars` assertion. Both accessor presence lists are updated.
- Docstring extended to `xr.DataArray or xr.Dataset` in Parameters and Returns, matching how `cost_distance` documents the same decorator.

### Notes
- `a_star_search` stays DataArray-only by design. It is not `@supports_dataset` and was out of scope here. The asymmetry is intentional and called out in the PR body.
- The unused import `has_dask_array` (`pathfinding.py:21`) is already on main, unrelated to this change, left as-is.
- Dataset application iterates every data variable, so a Dataset carrying a non-2D variable would raise on that variable. That matches the existing behavior of `cost_distance`/`proximity` on Datasets, so it is not a new sharp edge.

### Checklist
- [x] Algorithm matches reference/paper (no algorithm change; delegation + decorator)
- [x] All implemented backends produce consistent results (inherited)
- [x] Edge cases are covered by tests (DataArray and Dataset parity tested)
- [x] Benchmark exists or is not needed (thin delegation; not needed)
- [x] README feature matrix updated (already present)
- [x] Docstrings present and accurate (updated for Dataset input)

PR #3566, opened 2026-06-27
Speed up generate_terrain numpy backend with a fused parallel kernel

### LINKED ISSUE #3565: Speed up generate_terrain on the numpy backend with a fused parallel kernel
## Reason or Problem
`generate_terrain` is slow on the numpy backend. The octave loop in `_gen_terrain` (xrspatial/terrain.py) calls `_perlin` once per octave, and `_perlin` (xrspatial/perlin.py) is plain numpy: floor, astype, `& 255`, four fancy-index gathers (`p[p[xi] + yi]`), fade polynomials, lerps. Every one of those allocates a full-size temporary, and it all repeats for each octave (16 by default). The dask+numpy backend calls the same `_gen_terrain` per chunk, so it's slow too.

Measured here (numpy, default fbm, 16 octaves):
- 512x512: 244 ms
- 1024x1024: 1506 ms
- 2048x2048: 7368 ms

About 6.3s of the 2048x2048 run is the perlin octave loop. The GPU backend is already fast (28 ms at 2048x2048), so it's not the target.

## Proposal
Replace the per-octave numpy work with one fused `@njit(parallel=True)` kernel that computes the whole field per pixel -- domain warp, the fbm/ridged octave loop, and the final cube -- with `prange` over rows for threading. No per-octave temporaries; permutation tables are built once and passed in as a stacked int32 array.

A prototype of just the octave loop does the 2048x2048 16-octave fbm in 27.7 ms versus 6258 ms for the current 16x `_perlin`, and agrees with the current output to 8e-7 (float32 epsilon).

**Design:**
- Add a fused scalar kernel in terrain.py with parallel jit. Per pixel: optionally compute the warp displacement (warp_octaves perlin sum, normalized, scaled by warp_strength), then run the fbm or ridged octave loop, divide by the amplitude norm, cube.
- `_gen_terrain` takes the fast path when `worley_blend == 0` (the default). When worley blending is on, keep the existing numpy code, since worley needs a global min/max pass and is off by default.
- numpy and dask+numpy both go through `_gen_terrain`, so both benefit. cupy and dask+cupy are unchanged.

**Value:** Default CPU `generate_terrain` gets much faster with output unchanged within float32 epsilon; dask+numpy speeds up for free.

## Drawbacks
The worley path (worley_blend > 0) isn't sped up here.

## Alternatives
Fuse the GPU kernels too, but the GPU is already fast and its parity tests are tight, so leave it for a follow-up.

brendancol:
Closes #3565

Speeds up `generate_terrain` on the numpy backend (and dask+numpy, which routes through the same `_gen_terrain`).

## What changed
- The numpy octave loop called `_perlin` once per octave, and each call allocated ~20 full-size numpy temporaries (floor, fancy-index gather, fade, lerp). At 16 octaves that's hundreds of allocations.
- Replaced it with one fused `@njit(parallel=True)` kernel that computes warp + fbm/ridged + cube per pixel, `prange` over rows. No per-octave temporaries.
- `_gen_terrain` takes the fast path when `worley_blend == 0` (the default). The worley path keeps the existing numpy code, since it needs a global min/max pass.

## Results (numpy, default fbm, 16 octaves)
- 512x512: 244 ms -> 7.8 ms
- 1024x1024: 1506 ms -> 21 ms
- 2048x2048: 7368 ms -> 76 ms (~97x)

dask+numpy at 2048x2048 drops to 78 ms.

## Correctness
Output matches the previous implementation to float32 epsilon: fbm ~1.2e-7 relative, ridged ~1.6e-6 relative, worley path unchanged exactly.

## Backends
- numpy: fused kernel (faster)
- dask+numpy: faster for free (same `_gen_terrain`)
- cupy, dask+cupy: unchanged (GPU was already ~28 ms at 2048x2048)

## Test plan
- [x] `pytest xrspatial/tests/test_terrain.py` (52 passed, includes GPU + dask parity)
- [x] Added a reference-based test built from the un-fused `_perlin` so the kernel has an independent anchor on CPU-only CI (fbm, ridged, non-default params)
- [x] perlin / validation / templates / dask-task-name suites green

https://claude.ai/code/session_012qsDtsjQFiiNQ3MZjW3HBR

### DIFF
--- a/xrspatial/terrain.py
+++ b/xrspatial/terrain.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 # std lib
 import math
+import threading
 from functools import partial
 from typing import Dict, List, Optional, Tuple, Union
 
@@ -109,6 +110,14 @@ def _scale(value, old_range, new_range):
 # over rows with prange, so there are no per-octave temporaries.  Output matches
 # the numpy path to float32 epsilon.  Worley blending is not handled here (it
 # needs a global min/max pass); _gen_terrain keeps the numpy path for that.
+
+# Numba parallel=True kernels must not be launched concurrently from multiple
+# Python threads: the default 'workqueue' threading layer is not threadsafe and
+# aborts the process (SIGABRT on macOS) when two host threads enter a parallel
+# region at once.  _terrain_dask_numpy calls _gen_terrain_fast per chunk under
+# dask's threaded scheduler, so the kernel launch is serialized behind this
+# lock.  Same hazard and fix as the reproject kernels (#3141).
+_PARALLEL_KERNEL_LOCK = threading.Lock()
 
 
 @njit(nogil=True, inline='always')
@@ -231,12 +240,13 @@ def _gen_terrain_fast(height, width, seed, x_range, y_range, octaves,
         warp_norm = 1.0
 
     out = np.empty((height, width), dtype=np.float32)
-    _fused_terrain_numpy(
-        out, linx, liny, perms, octaves,
-        np.float32(persistence), np.float32(lacunarity), np.float32(norm),
-        noise_mode == 'ridged', do_warp, warp_px, warp_py,
-        warp_octaves, np.float32(warp_norm), np.float32(warp_strength),
-    )
+    with _PARALLEL_KERNEL_LOCK:
+        _fused_terrain_numpy(
+            out, linx, liny, perms, octaves,
+            np.float32(persistence), np.float32(lacunarity), np.float32(norm),
+            noise_mode == 'ridged', do_warp, warp_px, warp_py,
+            warp_octaves, np.float32(warp_norm), np.float32(warp_strength),
+        )
     return out
 
 
--- a/xrspatial/terrain.py
+++ b/xrspatial/terrain.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
 # std lib
+import math
 from functools import partial
 from typing import Dict, List, Optional, Tuple, Union
 
 # 3rd-party
 import numpy as np
 import xarray as xr
+from numba import njit, prange
 
 try:
     import cupy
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Speed up generate_terrain numpy backend with a fused parallel kernel

None.

- terrain.py:61 `_check_memory` still budgets `_SCRATCH_BYTES_PER_PIXEL = 24` (the old 5-7 same-shape arrays). The fused fast path allocates far less per pixel: linx/liny plus one output array, no per-octave temporaries. So for worley-off calls the guard now over-estimates and can raise MemoryError on grids that would actually fit. Not a correctness problem (it errs toward refusing), but the budget could be revisited for the fast path. Fine as a follow-up.

- test_terrain.py: the reference test uses `atol=2e-2`, which is loose in absolute terms on zfactor-scaled values. The `rtol=1e-4` is what actually guards against drift; the atol is only there to absorb the ridged feedback path's ~6e-3 absolute gap. A one-line comment would stop a future reader from tightening atol and hitting a flaky failure.
- terrain.py: the dask path rebuilds the permutation tables per chunk via `_make_perm_table`. Cheap (256-entry shuffle) and it matches the pre-existing per-chunk structure, so no action needed. Noting it for the record.

### What looks good
- Output verified against the pre-change implementation to float32 epsilon: fbm 1.2e-7 rel, ridged 1.6e-6 rel, worley path byte-identical.
- Edge cases confirmed: 1x1, single-row, single-column, 2x2, and strong warp pushing coordinates negative (the `& 255` two's-complement wrap matches the numpy path).
- prange writes distinct rows and acc/amp/freq/weight are loop-local, so there's no data race.
- Worley stays on the numpy path because it needs a global min/max pass; the fast-path gate is `worley_blend <= 0`.
- The new reference test gives the kernel an independent anchor on CPU-only CI, where dask-vs-numpy parity can't catch a shared-kernel bug and the GPU tests are skipped.
- Memory footprint is lower than before (no meshgrid, no per-octave temporaries).

### Checklist
- [x] Algorithm matches reference (perlin octave loop preserved, verified numerically)
- [x] All implemented backends produce consistent results (numpy/dask parity tests pass; GPU unchanged)
- [x] NaN handling is correct (all-NaN template overwritten, output finite)
- [x] Edge cases are covered (verified 1x1/row/col/warp; existing all-NaN tests pass)
- [x] Dask chunk boundaries handled correctly (per-chunk block_info coords unchanged)
- [x] Benchmark exists (benchmarks/benchmarks/terrain.py)
- [x] Docstrings present and accurate (public signature unchanged)

brendancol (commented):
## Follow-up review

Pushed one follow-up commit for the review findings.

- Nit (reference-test tolerance): fixed. The rtol/atol are now named constants with a comment explaining that rtol is the real drift guard and atol only absorbs the ridged path's ~6e-3 gap.
- Suggestion (`_check_memory` 24 B/pixel budget): not changing it. The guard is a deliberately conservative memory safety check (caps scratch at 50% of RAM) and erring toward refusing is the safe direction. Loosening it to admit a few larger worley-off grids trades real OOM protection for a niche win, so I'd rather leave it.
- Nit (per-chunk perm rebuild on dask): no action, it was informational and matches the existing per-chunk structure.

Terrain suite still green (52 passed). No blockers outstanding.

PR #3570, opened 2026-06-29
Recompute res attribute for downsampled preview() output

### LINKED ISSUE #3569: preview() copies stale res attribute to the downsampled output
## Description

`.xrs.preview()` downsamples a raster and rebuilds the `x`/`y` coordinates for the smaller grid, but copies the input's `attrs` unchanged. So `attrs['res']` keeps describing the original full-resolution grid instead of the downsampled one. Any downstream code that reads `res` for pixel-size math or georeferencing gets the wrong value.

## To Reproduce

```python
study_area.xrs.generate_terrain().xrs.preview()
```

```
<xarray.DataArray 'preview' (y: 988, x: 1000)>
Coordinates:
  * y        (y) float64 8kB 3.37e+06 3.369e+06 ... 3.329e+06 3.329e+06
  * x        (x) float64 8kB 6.005e+05 6.005e+05 ... 6.414e+05 6.415e+05
Attributes:
    res:                (0.25, 0.25)
    crs:                32614
    ...
```

The `x` coordinates span about 41000 map units across 1000 pixels, so the real resolution is roughly 41 units/pixel. But `res` still reads `(0.25, 0.25)` from the source grid.

## Expected behavior

`res` should match the grid that comes back. Recompute it from the output coordinate spacing so it lines up with the returned array's `x`/`y` coordinates.

## Notes

The stale copy is in `xrspatial/preview.py`. Every output path rebuilds coords with `_interpolate_coords` but passes `attrs=agg.attrs`, and the main `preview()` re-assigns `result.attrs = agg.attrs` after the second-pass refinement. The fix recomputes `res` from the final coordinates. It affects every method (mean, median, max, min, nearest, bilinear) and all backends.

brendancol:
Closes #3569.

`preview()` downsamples a raster and rebuilds the `x`/`y` coordinates for the smaller grid, but it copied the input's `attrs` verbatim. So `attrs['res']` kept describing the original full-resolution grid, and any downstream code reading `res` for pixel-size math or georeferencing got the wrong value.

## Change

- Add `_res_from_coords()` in `xrspatial/preview.py` to read the pixel size off the output coordinate spacing (positive magnitude, matching how templates and terrain store `res`).
- Recompute `res` once, right before `preview()` returns, after the second-pass refinement has settled the final coordinates. This covers every method (mean, median, max, min, nearest, bilinear) and all four backends in one place.
- Only touch `res` when the input already had one, and only when both axes have at least two coordinates. A raster without `res` stays without it; a no-op passthrough keeps its original.

## Backends

numpy, cupy, dask+numpy, dask+cupy — verified on all four. The one combination not covered is `dask+cupy` + `bilinear`, which already fails before this change in `_zoom_block` (scipy `zoom` on a cupy block); that is a separate pre-existing bug, untouched here.

## Test plan

- [x] `res` matches the returned array's coordinate spacing for all six methods
- [x] source-grid `res` does not survive a downsample
- [x] input `attrs` is not mutated
- [x] passthrough (no downsample) keeps the original `res`
- [x] input without `res` stays without `res`
- [x] dask path through snap + refinement
- [x] cupy path
- [x] full `test_preview.py` suite: 59 passed

### DIFF
--- a/xrspatial/preview.py
+++ b/xrspatial/preview.py
@@ -368,7 +368,8 @@ def preview(agg, width=1000, height=None, method='mean', name='preview'):
     Returns
     -------
     xr.DataArray
-        Downsampled raster with updated coordinates.
+        Downsampled raster with updated coordinates.  ``attrs['res']`` is
+        recomputed from the output spacing when the input carried a ``res``.
     """
     _validate_raster(agg, func_name='preview', ndim=2)
 
--- a/xrspatial/preview.py
+++ b/xrspatial/preview.py
@@ -282,6 +282,24 @@ def _interpolate_coords(coords, n_out):
     return np.interp(indices, np.arange(len(vals)), vals.astype(np.float64))
 
 
+def _res_from_coords(result, y_dim, x_dim):
+    """Pixel size ``(res_x, res_y)`` from the output coordinate spacing.
+
+    Returns ``None`` when either axis has fewer than two coordinates, so
+    the caller can leave any existing ``res`` untouched.  Spacing is taken
+    as a positive magnitude to match how ``res`` is stored elsewhere
+    (templates, terrain), regardless of whether the axis ascends or
+    descends.
+    """
+    if x_dim not in result.coords or y_dim not in result.coords:
+        return None
+    xs = np.asarray(result.coords[x_dim].values)
+    ys = np.asarray(result.coords[y_dim].values)
+    if xs.size < 2 or ys.size < 2:
+        return None
+    return (abs(float(xs[1] - xs[0])), abs(float(ys[1] - ys[0])))
+
+
 # ---------------------------------------------------------------------------
 # Second-pass refinement
 # ---------------------------------------------------------------------------
@@ -443,5 +461,13 @@ def preview(agg, width=1000, height=None, method='mean', name='preview'):
             result, target_h, target_w, y_dim, x_dim,
         )
         result.name = name
+
+    # The output grid is coarser than the input, so the copied 'res' is
+    # stale.  Recompute it from the downsampled coordinates so it matches
+    # the returned array's x/y spacing.
+    if 'res' in result.attrs:
+        new_res = _res_from_coords(result, y_dim, x_dim)
+        if new_res is not None:
+            result.attrs = {**result.attrs, 'res': new_res}
 
     return result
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: Recompute res attribute for downsampled preview() output

### Blockers
None.

### Suggestions
- `preview()` docstring (`xrspatial/preview.py:350-353`) still says the output has "updated coordinates" but not updated resolution. Worth a half-sentence so readers know `res` now tracks the downsampled grid.

### Nits
- `_res_from_coords` (`xrspatial/preview.py:285-300`) reads spacing from the first two coordinates only. That is exact for the uniform grids `preview` produces (interp over a linspace, or a strided isel), so it is fine here — just noting it assumes uniform spacing.

### Observations (not action items)
- `nearest` reports a slightly different `res` than `mean`/`bilinear` (5.0 vs 5.115 in my checks). That traces to how each path builds coords: `nearest` strides (spacing = factor x original) while the reduce/bilinear paths interpolate across the full original extent. The PR faithfully reflects whatever coords each path emits; the coord-construction difference predates this change.
- If an input carries `res` but no x/y coords, the recompute is skipped and `res` stays stale. That input is self-contradictory (res describes coord spacing) and there is no coordinate to measure, so leaving it alone is the right call.

### What looks good
- One recompute point after the second-pass refinement covers all six methods and all four backends, instead of patching each helper.
- Builds a fresh attrs dict, so the input is not mutated (covered by a test).
- Only touches `res` when it was present and measurable; absent-res inputs and no-op passthroughs keep their original attrs, both tested.
- Tests assert against the returned array's real coordinate spacing rather than a hardcoded number, and confirm the stale source value does not survive.

### Checklist
- [x] Result reflects the actual output coordinates
- [x] All implemented backends produce consistent results (numpy/cupy/dask+numpy/dask+cupy; bilinear+dask+cupy excluded by a pre-existing unrelated bug)
- [x] No input mutation
- [x] Edge cases covered (passthrough, absent res, dask refinement)
- [x] Docstrings present (minor wording suggestion above)
- [x] No README/benchmark change needed (bug fix, no API or perf change)

brendancol (commented):
## Follow-up review (after docstring commit 1227f85e)

Re-checked the diff after the follow-up commit.

- Suggestion from the first pass (document the `res` recompute) — **fixed**. The `preview()` Returns section now states `attrs['res']` is recomputed from the output spacing when the input carried a `res`.
- Nit (first-two-coords spacing) and the two observations — left as written; they describe correct, intended behavior, not defects.

No new findings. `flake8` clean on `preview.py`; the 17 res tests still pass. No blockers.

PR #3572, opened 2026-06-29
rasterize/kde/line_density: set output attrs['res'] when grid resolution changes (#3571)

### LINKED ISSUE #3571: rasterize/kde/line_density: output attrs['res'] not updated when output grid resolution differs from input
**Describe the bug**

Some tools build an output grid at a different resolution than their input but don't write the correct `res` into the output `attrs`. `get_dataarray_resolution()` reads `attrs['res']` in preference to deriving cell size from the coordinates, so a stale or missing `res` feeds the wrong cell size to downstream distance- and area-based ops like slope, aspect, and proximity.

Affected tools:

- `rasterize(..., like=template)` when the caller overrides the grid (explicit `width`/`height`/`bounds`/`resolution`) so the output no longer matches the template. The reshape path drops the template's stale `res` (correct) but never writes the new `res` it computes from the output grid, so the result carries no `res` at all.
- `kde(..., template=None)` and `line_density(..., template=None)` build a fresh grid with a known cell spacing `dx`/`dy`, then return a DataArray with no `res` attr.

`resample()` already does the right thing and is the model to follow: it recomputes `res` (and `transform`) from the rebuilt coordinates on every path, including downsample, upsample, 3D per-band, and `target_resolution`.

**Expected behavior**

Each of these outputs should carry `attrs['res'] = (abs(dx), abs(dy))` matching its actual output cell size, so downstream tools read the true resolution.

**Additional context**

`preview()` has the same class of issue but is being fixed in a separate PR, so it's out of scope here.

brendancol:
Closes #3571

Tools that build an output grid at a different resolution than their input now write the correct `res` into the output `attrs`. `get_dataarray_resolution()` reads `attrs['res']` before deriving cell size from coordinates, so a missing or stale `res` was handing the wrong cellsize to downstream distance/area ops like slope, aspect, and proximity.

- `rasterize()`: the reshape path (caller overrides `bounds`/`width`/`height`/`resolution`, or rasterizes with no `like` at all) now records the freshly-built grid's cell size as `attrs['res']` instead of leaving it absent. The reuse-`like`-coords path still keeps the template's `res`/`transform` unchanged. A stale inherited `transform` is still dropped on reshape.
- `kde()` / `line_density()`: when `template` is omitted, the output now carries `attrs['res'] = (abs(dx), abs(dy))` for the grid built from `x_range`/`y_range`.

`resample()` already did this on every path; this aligns the remaining resolution-changing tools with it. `preview()` has the same class of issue and is handled in a separate PR.

Backend coverage: the rasterize change covers numpy, cupy, dask+numpy, and dask+cupy (attrs are built backend-independently, and the #2251 test class exercises all four). The kde/line_density change is on the eager `template=None` construction path.

Test plan:
- Updated the issue #2251 rasterize tests (and the no-`like` rectangular-pixel test) to assert the refreshed `res` value instead of asserting `res` is absent.
- Added `TestKDEResolutionAttr` covering kde/line_density `res` from range, res-matches-coords, and template-supplied res.
- `test_kde.py`, `test_rasterize.py`, and `test_rasterize_coverage_2026_05_21.py` pass (340 passed, 2 skipped).

### DIFF
--- a/xrspatial/kde.py
+++ b/xrspatial/kde.py
@@ -546,7 +546,9 @@ def kde(
     Returns
     -------
     xr.DataArray
-        2-D density surface.
+        2-D density surface.  When *template* is omitted, ``attrs['res']``
+        carries the ``(x, y)`` cell size of the grid built from the data
+        extent; when *template* is given, the output inherits its attrs.
     """
     # -- Resolve GeoDataFrame input ----------------------------------------
     if is_geodataframe(x):
@@ -669,10 +671,15 @@ def kde(
             coords=template.coords, dims=template.dims,
             attrs=template.attrs,
         )
+    # No template: the grid was built from x_range/y_range above, so the
+    # cell spacing `dx`/`dy` is the output resolution.  Record it as
+    # `res` so downstream tools (which prefer `attrs['res']` over deriving
+    # cellsize from coords) read the true spacing.
     return xr.DataArray(
         data, name=name,
         dims=['y', 'x'],
         coords={'y': y_coords, 'x': x_coords},
+        attrs={'res': (abs(dx), abs(dy))},
     )
 
 
@@ -725,7 +732,10 @@ def line_density(
     Returns
     -------
     xr.DataArray
-        2-D line-density surface.
+        2-D line-density surface.  When *template* is omitted,
+        ``attrs['res']`` carries the ``(x, y)`` cell size of the grid built
+        from the data extent; when *template* is given, the output inherits
+        its attrs.
     """
     x1a = np.asarray(x1, dtype=np.float64).ravel()
     y1a = np.asarray(y1, dtype=np.float64).ravel()
@@ -799,10 +809,13 @@ def line_density(
             coords=template.coords, dims=template.dims,
             attrs=template.attrs,
         )
+    # No template: `dx`/`dy` is the output cell spacing, so record it as
+    # `res` for downstream tools that prefer `attrs['res']` over coords.
     return xr.DataArray(
         out, name=name,
         dims=['y', 'x'],
         coords={'y': y_coords, 'x': x_coords},
+        attrs={'res': (abs(dx), abs(dy))},
     )
 
 
--- a/xrspatial/rasterize.py
+++ b/xrspatial/rasterize.py
@@ -4354,7 +4354,11 @@ def rasterize(
         GeoDataFrame with a ``.crs`` and neither ``like`` nor its
         ``spatial_ref`` coord supplies a CRS, the geometry CRS is
         emitted on the output (``attrs['crs']`` as an EPSG int when one
-        resolves, else ``attrs['crs_wkt']`` as WKT).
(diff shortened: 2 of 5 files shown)


brendancol (commented):
## PR Review: rasterize/kde/line_density: set output attrs['res'] when grid resolution changes (#3571)

None.

None.

- `xrspatial/kde.py:676, 805` — For a single-column or single-row output (`width=1` or `height=1`), `dx`/`dy` comes out as `span / max(cols-1, 1) = span`, so `res` reports the full extent on that axis (e.g. `width=1` over `x_range=(0,10)` gives `res_x=10` against a single x coord at 0). That's a degenerate grid and the value just echoes the existing `dx`/`dy` binning, so it isn't wrong, just meaningless. Not worth special-casing unless single-pixel KDE grids are a real use case.
- `test_kde.py` — `test_kde_res_matches_coords` checks res-vs-coords and `get_dataarray_resolution` for kde; `line_density` only gets the from-range assertion. A matching res-vs-coords check for `line_density` would be symmetric, even though the construction path is the same.

### What looks good
- The reuse-`like`-coords path is untouched, so the bit-identical-template contract and the existing res/transform preservation still hold. Only the genuinely-reshaped path changes.
- `abs()` on px/py/dx/dy keeps `res` positive when a caller passes a descending `x_range`/`y_range` (checked: `x_range=(10,0)` yields `res (1.0, 1.0)`, not negative).
- `res` matches the coordinate-derived cellsize under both conventions: rasterize uses cell-center coords, kde uses an endpoint-inclusive linspace, and `res` equals the coord spacing in each.
- Stale `transform` is still dropped on reshape, so `res` and `transform` can't disagree.
- The #2251 test class was updated rather than deleted, so the regression coverage survives and now pins the stronger contract: `res` is present and correct instead of just absent.

### Checklist
- [x] res order is (x_res, y_res), consistent with resample() and get_dataarray_resolution()
- [x] All four backends covered for rasterize (attrs built post-dispatch, backend-independent; #2251 dask/cupy tests assert res)
- [x] NaN handling unaffected (metadata-only change)
- [x] Edge cases covered (rectangular pixels, no-like, like-reshape, template-supplied)
- [x] Dask chunk boundaries unaffected (no compute path change)
- [x] No premature materialization or copies introduced
- [x] No benchmark needed (metadata-only)
- [x] Docstrings updated for kde, line_density, and rasterize Returns

brendancol (commented):
## Follow-up review (#3571)

Re-reviewed after the follow-up commit (`939d358a`). The commit only adds a test, no production change, so the prior review's findings still stand. Disposition of the two nits:

- Nit 2 (line_density res-vs-coords test) — fixed in `939d358a` via `test_line_density_res_matches_coords`, mirroring the kde check.
- Nit 1 (single-cell `res` reports the full extent) — dismissed. For a 1-cell axis, `res` mirrors the `dx`/`dy` the kernel already uses for binning, so it stays consistent with how the grid was built. There's no sensible alternative value, and special-casing a degenerate config would add code for no real gain.

No blockers or suggestions. Good to merge once CI is green.

PR #3575, opened 2026-06-29
generate_terrain: map dask backends over an empty skeleton instead of the template

### LINKED ISSUE #3574: generate_terrain dask backends materialize the template's NaN fill before discarding it
## Summary
The dask backends of `generate_terrain` regenerate every cell from coordinates and never read the values of the template array passed in. They still map the per-chunk generator over that array, so the template's fill gets computed and then thrown away. For a `from_template()` grid that fill is a `da.full(shape, nan)`, a full-grid NaN memset that is dead work in the graph.

## Example
```python
import xrspatial
study_area = xrspatial.from_template('austin', chunks='auto', resolution=.5)  # 81920x81920, backed by da.full(nan)
terrain = study_area.xrs.generate_terrain()  # the NaN fill is materialized per chunk, then overwritten
```

## Proposed change
Map the per-chunk terrain generator over a `da.empty_like(data)` skeleton instead of the template array. The skeleton keeps the template's shape, dtype and chunking, so the per-chunk `block_info` coordinates stay identical, but it carries no fill and the template's graph drops out. While we're here, remove the leftover `data * 0` in the dask+cupy path, which builds a NaN array that the per-chunk `cupy.zeros` immediately discards.

Output values are unchanged. This only removes the wasted allocation and fill from the dask graph.

## Backends
dask+numpy and dask+cupy. The eager numpy/cupy paths already allocate their own output and aren't affected.

brendancol:
## Summary
- The dask backends of `generate_terrain` regenerate every cell from coordinates and never read the template's values. They now map the per-chunk generator over a `da.empty_like(data)` skeleton instead of the template array, so a NaN-filled template (e.g. `from_template()`'s `da.full(shape, nan)`) drops out of the graph instead of being memset and discarded.
- Removes the leftover `data * 0` in the dask+cupy path, which built a NaN array the per-chunk `cupy.zeros` immediately discarded.

Output values are unchanged; this only removes wasted allocation and fill from the dask graph. Verified that the `from_template -> generate_terrain` graph no longer carries a `full`/NaN layer (only `empty_like` + `xrspatial.terrain` remain).

Closes #3574

## Backends
- numpy: unchanged (eager path allocates its own output)
- cupy: unchanged
- dask+numpy: skeleton swap
- dask+cupy: skeleton swap + `data * 0` removed

## Test plan
- [x] `test_terrain_dask_does_not_materialize_template` (worley on and off): a poisoned template whose blocks raise if computed must still produce finite terrain
- [x] `test_terrain_dask_skeleton_matches_numpy`: skeleton output matches the numpy reference
- [x] `test_terrain_dask_cupy_does_not_materialize_template`
- [x] Full `test_terrain.py` suite green (56 passed, all four backends on the CUDA box)

### DIFF
--- a/xrspatial/terrain.py
+++ b/xrspatial/terrain.py
@@ -378,6 +378,13 @@ def _terrain_dask_numpy(data, seed, x_range_scaled, y_range_scaled, zfactor,
     """
     height, width = data.shape
 
+    # The chunk functions below regenerate every cell from coordinates and
+    # never read the incoming block values.  Map over an uninitialised
+    # skeleton with the same shape/chunking so the (NaN-filled) template the
+    # caller passed in — e.g. from_template()'s da.full — drops out of the
+    # graph instead of being memset and then thrown away.
+    skeleton = da.empty_like(data, dtype=np.float32)
+
     # --- worley pre-pass: get global min/max to avoid per-chunk seams ---
     w_norm_range = None
     if worley_blend > 0:
@@ -417,7 +424,7 @@ def _terrain_dask_numpy(data, seed, x_range_scaled, y_range_scaled, zfactor,
             return _worley_numpy_xy(w_p, x_arr, y_arr)
 
         raw_worley = da.map_blocks(
-            _chunk_worley_raw, data, dtype=np.float32,
+            _chunk_worley_raw, skeleton, dtype=np.float32,
             meta=np.array((), dtype=np.float32),
         )
         (raw_worley,) = dask.persist(raw_worley)
@@ -452,7 +459,7 @@ def _terrain_dask_numpy(data, seed, x_range_scaled, y_range_scaled, zfactor,
         out *= zfactor
         return out
 
-    return da.map_blocks(_chunk_terrain, data, dtype=np.float32,
+    return da.map_blocks(_chunk_terrain, skeleton, dtype=np.float32,
                          meta=np.array((), dtype=np.float32),
                          **_dask_task_name_kwargs('xrspatial.terrain'))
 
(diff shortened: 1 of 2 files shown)


brendancol (commented):
## PR Review: generate_terrain dask backends map over an empty skeleton

### Blockers
None.

### Suggestions
- The new tests cover worley on and off for the dask+numpy non-materialization case, but `test_terrain_dask_cupy_does_not_materialize_template` only runs with worley off. The worley pre-pass `map_blocks` was also switched to the skeleton, so a worley-on cupy case would exercise that second call site on GPU too. Low priority, since the numpy parametrize already covers both sites and the existing `test_terrain_all_nan_template_dask_cupy_matches_numpy` covers the default path. (xrspatial/tests/test_terrain.py)

### Nits
- The skeleton still produces one uninitialized `np.empty`/`cupy.empty` allocation per chunk feeding `map_blocks`. That is the expected pattern and there is no memset, so the real win is dropping the `da.full(nan)` fill, not the allocation. No code change needed.

### What looks good
- Verified the `from_template -> generate_terrain` graph drops the full/NaN layer (only `empty_like` + `xrspatial.terrain` remain) and the output stays finite and matches numpy.
- Removes the leftover `data * 0` in the dask+cupy path, the same NaN*0 pattern #3525 fixed on the eager backends.
- The poison-template tests are a direct proof that the template is never read: blocks that raise on compute, yet the result computes fine.
- `empty_like` preserves shape, dtype, and chunking, so `block_info` coordinates are unchanged and output values are identical (parity test passes).

### Checklist
- [x] No algorithm change (output verified identical)
- [x] NaN handling correct (template NaN dropped, output finite)
- [x] Edge cases covered (poison template, worley on/off)
- [x] Dask chunk boundaries handled correctly (empty_like preserves chunking)
- [x] Removes a premature materialization rather than adding one
- [x] Benchmark exists (benchmarks/benchmarks/terrain.py)
- [x] README feature matrix update not needed (no new function or backend)
- [x] Comments added; no public API change

brendancol (commented):
## Follow-up review (after 8a9b1ebc)

The earlier Suggestion is addressed: `test_terrain_dask_cupy_does_not_materialize_template` is now parametrized over `worley_blend` [0.0, 0.2], so the worley pre-pass `map_blocks` call site is exercised on the dask+cupy path as well.

- Blockers: none
- Suggestions: none outstanding
- Nits: the earlier allocation note stands as documentation only; no code change warranted.

Local run: `test_terrain.py` template/materialize/skeleton tests pass (10 passed), full file green earlier (56 passed) on the CUDA box. Nothing new introduced by the follow-up.

PR #3586, opened 2026-07-01
Fix pycnophylactic crash on all-NaN zones or unmatched zone ids (#3406)

### LINKED ISSUE #3406: pycnophylactic raises opaque ValueError on all-NaN zones or no matching zone id

brendancol:
## Summary

- `pycnophylactic` raised an opaque numpy `ValueError` ("zero-size array to reduction operation fmax which has no identity") when no pixel was valid for smoothing. Two reachable cases: the `zones` raster is entirely NaN, or none of the zone ids match the `values` mapping. In both, the `valid` mask is all-False and the convergence check runs `np.nanmax` over an empty slice.
- The guard returns the all-NaN surface before the smoothing loop, so `pycnophylactic` behaves like `disaggregate` on the same inputs instead of crashing.
- Removed the `xfail(strict)` markers that were pinning the crash, and added a cupy-fallback case.

Closes #3406

## Backend coverage

`pycnophylactic` supports numpy and cupy (CPU fallback); dask raises `NotImplementedError` by design. The fix is in `_pycnophylactic_numpy`, which both supported paths share.

## Test plan

- [x] `test_pycnophylactic_all_nan_zones` (numpy)
- [x] `test_pycnophylactic_no_matching_zone` (numpy)
- [x] `test_pycnophylactic_all_nan_zones_cupy` (CUDA box)
- [x] Full `test_dasymetric.py`: 78 passed


brendancol (commented):
## PR Review: Fix pycnophylactic crash on all-NaN zones or unmatched zone ids (#3406)

None.

None.

- `xrspatial/tests/test_dasymetric.py:1073` -- the class docstring's first line still reads "pycnophylactic crashes when no pixel is valid for smoothing", which is stale now that the fix removes the crash. The rest of the docstring already describes the fixed behaviour, so only the opening clause is out of date.

### What looks good
- The guard sits right after `valid` is computed and before the loop, so `np.nanmax` never sees the empty slice. `not valid.any()` means every pixel is NaN, so returning `surface` hands back an all-NaN raster, which is what `disaggregate` does on the same inputs.
- Shape, dtype, coords and attrs match the normal return path.
- Coverage hits both reachable cases (all-NaN zones, no matching zone id) on numpy, plus the cupy fallback that routes through the same `_pycnophylactic_numpy`. dask already raises `NotImplementedError` and is covered elsewhere.

### Checklist
- [x] Behaviour now matches disaggregate
- [x] Implemented backends consistent (numpy + cupy fallback)
- [x] NaN handling correct
- [x] Dask path unsupported, raises (tested elsewhere)
- [ ] Docstrings accurate (see stale class-docstring nit)

brendancol (commented):
## Follow-up review (after 3e94d091)

The one nit from the previous pass is resolved: the `TestPycnophylacticEmptyValid` class docstring no longer claims pycnophylactic "crashes", so it now matches the fixed behaviour.

No new findings. No open Blockers, Suggestions, or Nits. Tests green locally (4 passed in the class, 78 in the full dasymetric suite).

PR #3594, opened 2026-07-02
Stop a non-finite RAT value in a PAM sidecar from crashing open_geotiff

### LINKED ISSUE #3590: Non-finite RAT value in a PAM sidecar crashes open_geotiff with OverflowError
#### Description

`read_pam_sidecar()` must never raise on a bad sidecar: `open_geotiff()` reads the PAM `.aux.xml` for any local string source, and a malformed sidecar is supposed to degrade to `{}` rather than break the read (#3520, #3522).

There's still a gap. `_parse_rat()` converts a thematic RAT's Value cell with `int(float(fields[value_col]))`. When the cell text parses to infinity (`1e400`, `inf`, `-inf`), `int()` raises `OverflowError`, which subclasses `ArithmeticError`, not `ValueError`, so the `(OSError, ValueError, TypeError, IndexError, ParseError)` tuple in `read_pam_sidecar()` doesn't cover it. The exception escapes and crashes `open_geotiff()`.

Same contract violation as #3520, different exception. A corrupt or hostile `.aux.xml` next to an otherwise valid raster makes that raster unopenable.

#### Steps to reproduce

```python
import numpy as np
import xarray as xr

da = xr.DataArray(np.arange(12, dtype='float32').reshape(3, 4), dims=('y', 'x'),
                  coords={'y': [2.5, 1.5, 0.5], 'x': [0.5, 1.5, 2.5, 3.5]},
                  attrs={'crs': 32633})
to_geotiff(da, 'dem_3xxx.tif')

with open('dem_3xxx.tif.aux.xml', 'w') as fh:
    fh.write('''<PAMDataset>
  <PAMRasterBand band="1">
    <GDALRasterAttributeTable tableType="thematic">
      <FieldDefn index="0"><Name>Value</Name><Type>1</Type><Usage>5</Usage></FieldDefn>
      <FieldDefn index="1"><Name>Class</Name><Type>2</Type><Usage>2</Usage></FieldDefn>
      <Row index="0"><F>1e400</F><F>water</F></Row>
    </GDALRasterAttributeTable>
  </PAMRasterBand>
</PAMDataset>
''')

open_geotiff('dem_3xxx.tif')
# OverflowError: cannot convert float infinity to integer
```

#### Expected behaviour

The raster opens and the sidecar is ignored (no `category_names` attr), like every other malformed-sidecar case.

#### Actual behaviour

`OverflowError: cannot convert float infinity to integer` from inside `_parse_rat`, crashing the read.

#### Fix sketch

Add `OverflowError` to the caught exception tuple in `read_pam_sidecar()`, the same way #3522 added `IndexError` and `ParseError`.

brendancol:
Closes #3590

A thematic RAT `<F>` Value cell whose text parses to infinity (`1e400`, `inf`, `-inf`) made `int(float(...))` in `_parse_rat` raise `OverflowError`. That subclasses `ArithmeticError`, not `ValueError`, so it slipped past `read_pam_sidecar()`'s fail-closed except tuple and crashed the `open_geotiff()` call for any local source with such a sidecar next to it. Same contract violation as #3520/#3522, reached through a different exception.

- Add `OverflowError` to the caught exceptions in `read_pam_sidecar()` so an infinite Value cell degrades to `{}` like every other malformed-sidecar case.
- Add `test_non_finite_rat_value_returns_empty` alongside the #3520 short-row test; it fails with `OverflowError` before the fix.
- Also carries this run's `.claude/sweep-security-state.csv` update (found by `/sweep-security`).

Backend coverage: the sidecar read happens in `open_geotiff` after backend dispatch on all four read paths (eager, gpu, dask+numpy, dask+gpu), so the fix covers them all; no kernel or graph code touched.

Test plan:
- [x] `pytest xrspatial/tests/test_rasterize_categorical_3482.py` (21 passed)
- [x] End-to-end repro from #3590: `open_geotiff` on a TIFF with the adversarial sidecar now opens and ignores it


brendancol (commented):
Domain-aware review (rockout step 9, /sweep-security run).

**Scope**: one-line exception-tuple widening in `read_pam_sidecar()` plus a regression test and the sweep state CSV row. Matches the precedent set by #3522, which added `IndexError` and `ParseError` to the same tuple for the same fail-closed contract.

**Blockers**: none.

**Suggestions**: none. I checked the other conversions in `_parse_rat` for sibling gaps: `int(usage_el.text)`, `int(fd.get('index'))`, `int(row.get('index'))`, and the RGBA `int(fields[c])` all raise `ValueError`/`TypeError` on bad text (including `'1e400'`, which `int()` refuses directly), and `int(float('nan'))` raises `ValueError`. Only the Value cell's `int(float(...))` path can produce `OverflowError`, so the one added type closes the gap. Catching it in `read_pam_sidecar` rather than clamping inside `_parse_rat` keeps behaviour consistent with the short-row case (whole sidecar degrades to `{}`).

**Nits**: none. Test filename embeds the issue number per project convention, and the sidecar XML mirrors the neighbouring #3520 test.

Verified locally: the new test fails with `OverflowError` on the parent commit and passes with the fix; the end-to-end repro from #3590 (`open_geotiff` on a TIFF beside the adversarial sidecar) now opens and ignores the sidecar. Full test file: 21 passed.

PR #3596, opened 2026-07-02
Fix to_geotiff docstring tier for extra_tags pass-through

### LINKED ISSUE #3593: to_geotiff docstring lists extra_tags pass-through as advanced, but the write gate is experimental
**Describe the bug**

The release-contract tier list in `to_geotiff`'s docstring (`xrspatial/geotiff/_writers/eager.py`) puts "`extra_tags` pass-through" in the `[advanced]` bullet. Advanced-tier features carry no kwarg gate. The tier map disagrees: `SUPPORTED_FEATURES['writer.extra_tags']` is `'experimental'`, and `_validate_write_rich_tag_optin` rejects a fresh DataArray carrying `attrs['extra_tags']` unless the caller passes `allow_experimental_codecs=True`:

```python
da.attrs["extra_tags"] = [(305, 2, None, "sweep")]
to_geotiff(da, "out.tif")
# ValueError: to_geotiff: attrs['extra_tags'] pass-through is experimental: ...
```

A user following the docstring expects a no-flag path and gets a ValueError. (Attrs that came from an xrspatial read are exempt from the gate; the tier list doesn't mention that either.)

**Expected behavior**

The tier list should match `SUPPORTED_FEATURES` and the gate: `extra_tags` (and `gdal_metadata_xml`) pass-through belongs under `[experimental]`, with a note about the `allow_experimental_codecs=True` opt-in and the round-trip exemption.

Doc-only fix, no behavior change. Found during an API-consistency sweep of the geotiff module; the gate behavior above was verified against current main.

brendancol:
Closes #3593

- Moves "`extra_tags` pass-through" from the `[advanced]` bullet of `to_geotiff`'s release-contract tier list to `[experimental]`, matching `SUPPORTED_FEATURES['writer.extra_tags']` and the `_validate_write_rich_tag_optin` gate (which raises `ValueError` on a fresh DataArray with `attrs['extra_tags']` unless `allow_experimental_codecs=True`). The bullet now names the opt-in and the round-trip exemption, and mentions `gdal_metadata_xml` which shares the gate.
- Clarifies the `photometric` parameter doc: the override rides `attrs['extra_tags']` (not a kwarg) and sits at the experimental tier.
- Adds `test_to_geotiff_docstring_extra_tags_tier_matches_supported_features` to `tests/unit/test_signatures.py` (Section 3, next to the existing rich-tag gate tests) so the docstring cannot drift from the tier map again.
- Updates the api-consistency sweep state CSV row for `geotiff`.

Doc-only change, no behavior touched, so backend coverage is unaffected (the gate itself already fires identically on the CPU and GPU write paths).

Test plan:
- [x] `pytest xrspatial/geotiff/tests/unit/test_signatures.py` (197 passed, includes the new test)
- [x] `pytest xrspatial/geotiff/tests/unit/test_photometric.py` (37 passed)
- [x] flake8 clean on both touched source files
- [x] Empirical check on a CUDA host: fresh `attrs['extra_tags']` write raises without the flag, succeeds with it, and the cupy write round-trip still works


brendancol (commented):
Doc-only change; I checked the claims against the code rather than just the prose.

- The new [experimental] bullet matches `_validate_write_rich_tag_optin`: fresh `attrs['extra_tags']` / `attrs['gdal_metadata_xml']` raise without `allow_experimental_codecs=True`, and the `_xrspatial_geotiff_contract` marker exempts round-tripped attrs. `SUPPORTED_FEATURES` carries both keys at `'experimental'`, so the docstring and tier map now agree.
- The drift-guard test scopes its string checks to the region before "Parameters", which keeps the `photometric` parameter doc (which legitimately mentions the extra_tags override) out of the assertion. The bullet slicing on `'* ['` holds because `[internal-only]` follows the experimental bullet.

Nits, not blocking:

- The gate also fires on a dict `attrs['gdal_metadata']` (#3320); the revised bullet doesn't name that third attr. The runtime error does, so this is fine to leave.
- The test pins the two SUPPORTED_FEATURES tiers directly, which overlaps with the release-gate suite. Harmless redundancy that makes the test self-contained.

No blockers.

PR #3602, opened 2026-07-02
Remove unused xrspatial.analytics module

### LINKED ISSUE #3601: Remove the xrspatial.analytics module
## Reason or Problem
`xrspatial/analytics.py` holds one function, `summarize_terrain`. It runs `slope`, `aspect`, and `curvature` on a DataArray and packs the three results into a Dataset named after the input. That's the whole module.

Nothing in the project points at it:

- It isn't exported from `xrspatial/__init__.py`, so it's not part of the top-level public API.
- It's absent from the README feature matrix and the `docs/` API reference.
- A repo-wide grep finds no internal callers. The only references are its own test file (`xrspatial/tests/test_analytics.py`) and its own docstring.

## Proposal
Delete `xrspatial/analytics.py` and its test file `xrspatial/tests/test_analytics.py`.

**Impact:** You can still reach the function today via `from xrspatial.analytics import summarize_terrain`. That path is undocumented, but removing the module still breaks anyone who found it. The replacement is to call `slope`, `aspect`, and `curvature` yourself and build the Dataset — which is all the wrapper does.

**Value:** One less undocumented entry point to keep working as the `slope`/`aspect`/`curvature` signatures change.

## Drawbacks
Someone might depend on the undocumented import. The wrapper is a few lines, so reconstructing it is cheap.

## Alternatives
- Deprecate with a warning for one release first. Since the function is already off the public API and out of the docs, a straight removal seems proportionate, but I'm open to the softer path.
- Go the other way: export it, document it, give it backend coverage. That works against the goal of a smaller surface.

There's one other "analytics" match in the tree — a use of the word in a `geotiff/_writers/eager.py` docstring. Unrelated to this module.

brendancol:
Closes #3601

## What

- Delete `xrspatial/analytics.py` (its only function, `summarize_terrain`, is a thin wrapper over `slope`/`aspect`/`curvature`).
- Delete its test file `xrspatial/tests/test_analytics.py`.

## Why

The function was never exported from `xrspatial/__init__.py`, never appeared in the README feature matrix or the `docs/` API reference, and has no internal callers (repo-wide grep). Callers who want the same result can call `slope`, `aspect`, and `curvature` directly and assemble the Dataset — that's all the wrapper did.

Removing the module breaks the undocumented `from xrspatial.analytics import summarize_terrain` import path. See #3601 for the discussion of a deprecation-first alternative.

## Backend coverage

Not applicable — this only removes code.

## Test plan

- [x] `python -c "import xrspatial"` still works
- [x] `pytest xrspatial/tests/ --co` collects clean (7616 tests, no import errors)
- [x] No remaining references to `summarize_terrain` or the `analytics` module (the one `analytics` hit left is an unrelated word in a `geotiff/_writers/eager.py` docstring)


brendancol (commented):
## PR Review: Remove unused xrspatial.analytics module

This PR deletes `xrspatial/analytics.py` and its test file, nothing else. The usual domain checks (algorithm accuracy, backend dispatch, dask/cupy kernels, performance) don't apply to a straight removal, so I looked at whether anything still depends on the module and whether the removal leaves the tree consistent.

### Blockers
None.

### Suggestions
The deprecation-first path raised in #3601 is still open. If a maintainer would rather ship a `DeprecationWarning` for one release than remove outright, that's their call. The PR takes the hard-removal route, which holds up given the function was never public.

### Nits
None.

### What looks good
- The diff is exactly two file deletions, no stray edits. (I caught and fixed a base-branch issue first: the branch had been cut from a feature branch and was dragging unrelated template work into the diff. A rebase onto origin/main cleaned it up.)
- No remaining importers. A repo-wide grep for `summarize_terrain` and `from xrspatial.analytics` comes back empty. The one surviving `analytics` hit is an unrelated word in a geotiff docstring.
- `import xrspatial` still works, and `pytest --co` collects 7616 tests with no import errors from the removed test file.
- Nothing in `__init__.py`, the README matrix, or `docs/` pointed at the module, so there are no dangling doc entries.

### Checklist
- Removal leaves no dangling imports or references: verified
- Package still imports: verified
- Test suite collects clean: verified
- No public API or docs entry pointed at the module: verified
- Breaking change (undocumented import path) noted in the PR body and issue: yes

PR #3607, opened 2026-07-02
docs(bump): correct height_func contract and count default in docstring

### LINKED ISSUE #3606: bump() docstring misdescribes the height_func contract
The `bump()` docstring in `xrspatial/bump.py` describes `height_func` as a "function which takes x, y and returns a height value". That is not how `bump()` calls it. It invokes `height_func(locations)` once, passing a single `(count, 2)` array of integer `(x, y)` coordinates, and expects a 1D array of `count` heights back. Following the documented per-point `f(x, y)` contract literally raises a TypeError.

```python
import numpy as np
from xrspatial import bump

def hf_xy(x, y):   # the documented contract
    return 1.0

bump(width=10, height=10, count=5, height_func=hf_xy, spread=0)
```

Output:

```
TypeError: hf_xy() missing 1 required positional argument: 'y'
```

The real contract works:

```python
def hf_arr(locations):
    return np.ones(locations.shape[0])

bump(width=10, height=10, count=5, height_func=hf_arr, spread=0)   # OK
```

Secondary: `count` is documented as `int` with no default, but the signature is `Optional[int] = None`. When omitted it defaults to `min(width * height // 10, 10_000_000)`, and that default is undocumented.

This is documentation-only. The behavior is correct and stays unchanged; the runnable example already uses the array-based convention.

Found during a documentation sweep of `bump`.

brendancol:
Fixes two accuracy problems in the `bump()` docstring, found during a documentation sweep of the module.

- `height_func` was described as taking `x, y` per point. `bump()` actually calls it once with a `(count, 2)` locations array and expects a length-`count` heights array back. The per-point form raises `TypeError` when used literally.
- `count` was typed `int` with no default. It is optional and defaults to `min(width * height // 10, 10_000_000)`.

Docstring-only, no behavior change. The runnable example already used the array-based `height_func` convention, so it needed no edit.

Closes #3606


brendancol (commented):
## PR Review: docs(bump): correct height_func contract and count default in docstring

Documentation-only change to `bump()` in `xrspatial/bump.py`. I verified the corrected claims by running the code against the worktree build.

None.

None.

- `xrspatial/bump.py:184-185`: the `count` default is written as `min(width * height // 10, 10_000_000)`. When `agg` is passed, `width`/`height` are ignored and the dimensions come from `agg.shape` (lines 337-338, 344). The formula is still correct in terms of the raster size, but a reader who only passes `agg` might wonder where `width`/`height` come from. Optional: phrase it in terms of the output raster dimensions.

### What looks good

- The old `height_func` description (`function which takes x, y`) was wrong: the code calls `height_func(locs)` once with a single `(count, 2)` int32 array (line 394), so a literal `f(x, y)` callable raises `TypeError`. I confirmed both the new contract and the `TypeError` on the old one.
- The new `count` doc matches `count = min(w * h // 10, _MAX_DEFAULT_COUNT)` with `_MAX_DEFAULT_COUNT = 10_000_000` (lines 349-350, 29).
- The `height_func=None` default is documented as height 1, matching `lambda bumps: np.ones(len(bumps))` (line 387).
- The docstring's own plot example already uses the `(count, 2)` contract (`heights(locations, ...)` reading `locations.shape[0]` / `locations[r]`), so example and prose are now consistent.

### Checklist

- [x] Docstring matches actual behavior (verified by execution)
- [x] Parameter types and defaults documented
- [n/a] Algorithm/backend/NaN/dask correctness — no code changed
- [n/a] Benchmark / README feature matrix — docs-only, no API surface change

brendancol (commented):
## PR Review (follow-up): docs(bump): correct height_func contract and count default

Re-review after merging `origin/main` and applying the one nit from the first pass.

- Merged `origin/main` (was 19 commits behind, PR showed CONFLICTING). The only conflict was in the `bump()` docstring, where main's #3610 had rewritten the `count` line and added the `name=` parameter while still carrying the old, wrong `height_func` text. Resolved by keeping this PR's corrected `height_func` contract and main's `name=` parameter (signature `name: str = 'bump'`, doc entry, and `name=name` on both `DataArray` returns are all present). PR is now MERGEABLE.
- Applied the nit I raised: the `count` default now reads `min(w * h // 10, 10_000_000)`, where `w`/`h` are the output raster dimensions (taken from `agg` when given), instead of naming `width`/`height` which are ignored on the `agg` path.

### Blockers / Suggestions

None.

### Nits

None outstanding.

### Verification

- `pytest xrspatial/tests/test_bump.py` — 20 passed against the worktree build.
- Post-merge diff vs base is confined to the `bump()` docstring and the sweep state CSV. No code paths changed.

### Checklist

- [x] Merge conflict resolved preserving both main's `name=` and this PR's corrections
- [x] Docstring matches actual behavior (verified by execution earlier)
- [x] Tests pass post-merge
- [n/a] Algorithm/backend/NaN/dask correctness — no code changed

Issue #3612, opened 2026-07-02
bump(): dask graph construction is O(n_chunks x count) in _partition_bumps


brendancol:
## Summary

`_partition_bumps` in `xrspatial/bump.py` assigns each bump to the dask chunk that contains its centre pixel. It does that by looping over every chunk and computing a fresh boolean mask over the full `locs` array:

```python
for yi in range(ny):
    for xi in range(nx):
        mask = ((locs[:, 0] >= x0) & (locs[:, 0] < x1) &
                (locs[:, 1] >= y0) & (locs[:, 1] < y1))
        ...
```

That is O(n_chunks x n_bumps) work, and it runs eagerly during dask graph construction, before any `.compute()`. For a finely chunked raster the runtime is dominated by rescanning `locs`, not by rendering bumps.

## Reproduction

Graph construction only, no compute:

```python
import time
import numpy as np
import dask.array as da
from xrspatial.bump import _partition_bumps

count = 100_000
locs = np.empty((count, 2), dtype=np.int32)
for side in (10, 40, 80, 160):
    n = side * 250
    np.random.seed(0)
    locs[:, 0] = np.random.randint(0, n, count)
    locs[:, 1] = np.random.randint(0, n, count)
    heights = np.ones(count)
    arr = da.zeros((n, n), chunks=(250, 250), dtype="float64")
    t0 = time.time()
    _partition_bumps(arr, locs, heights, 1)
    print(side ** 2, "chunks", round(time.time() - t0, 3), "s")
```

Observed:

| n_chunks | _partition_bumps time (count=100k) |
|---|---|
| 100 | 0.03 s |
| 1,600 | 0.48 s |
| 6,400 | 1.93 s |
| 25,600 | 7.77 s |

The default count heuristic caps at 10M bumps, which is 100x these times. A 30 TB dask raster has millions of chunks, so graph construction alone runs for hours before compute starts. The memory guard does not apply to a dask-backed `agg` by design, so nothing bounds this cost.

## Proposed fix

Bucket every bump into its chunk in a single pass with `np.searchsorted` on the chunk offset arrays, then group by chunk. This is O(n_bumps log n_chunks) and returns byte-identical partitions, verified across 30 seeds with irregular chunk layouts and preserving intra-chunk order. Measured 85x faster at 25,600 chunks, and the gap widens as chunk count grows:

| n_chunks | old | new | speedup |
|---|---|---|---|
| 100 | 0.036 s | 0.010 s | 3.4x |
[...]

PR #3611, opened 2026-07-02
Fix isort import-ordering drift in bump.py

brendancol:
Closes #3609

## What

Run isort on `xrspatial/bump.py` to fix a Cat 4 import-ordering finding. The `from xrspatial.utils import (...)` block was laid out one name per line; the configured isort (`line_length=100`) wraps it as a two-line grid instead.

## Categories addressed

- Cat 4 (isort import ordering). No behavior change intended.
- Cat 1/2/3/5 were clean, so nothing else in scope: no flake8 E/W codes (baseline 0), no unused imports or dead locals, no bare `except`, no mutable default args, no shadowed builtins.

## Test plan

- [x] `flake8 xrspatial/bump.py` clean
- [x] `isort --check-only xrspatial/bump.py` clean
- [x] `xrspatial/tests/test_bump.py` passes (18 tests, CUDA available so cupy and dask+cupy paths ran)
- [x] Module imports and all five `xrspatial.utils` names resolve

Found by the /sweep-style deep sweep (2026-07-02).

PR #3614, opened 2026-07-02
bump: validate agg, count, and spread inputs

brendancol:
## What

`bump()` validated only `width` and `height`. This adds the same style of validation to `agg`, `count`, and `spread`, using the existing `_validate_raster` and `_validate_scalar` helpers.

Before, bad inputs surfaced errors from deep in the call or passed silently:

| input | old behavior |
|-------|--------------|
| `agg=np.zeros((10,10))` | `TypeError: Unsupported Array Type` from the backend dispatcher |
| `agg` 3D/1D DataArray | `ValueError: too many values to unpack` from `h, w = agg.shape` |
| `count=-5` / `count=5.0` | raw `np.empty` errors |
| `spread=-3` / `spread=2.5` | accepted silently, no spreading applied |

After, each raises a message that names the parameter, e.g. `bump(): \`agg\` must be 2D, got 3D`.

## Notes

- `bump` only reads `agg`'s shape and backend, so I left the dtype unchecked (`numeric=False`).
- `spread=0` and `count=0` stay valid.
- Rejecting negative/non-int `spread` changes behavior for inputs that were previously accepted silently. Those values are out of domain per the docstring (`spread : int`), and no test relied on them.

## Tests

Added error-path tests asserting the new exception type and message for each case, plus a check that `spread=0` still works. Full `test_bump.py` passes (23 tests) on numpy, cupy, dask, and dask+cupy.

---
error-handling sweep, 2026-07-02


brendancol (commented):
## PR Review: bump: validate agg, count, and spread inputs

Reviewed against the worktree at the PR head. The change adds input
validation to `bump()`: `agg` must be a 2D `DataArray`, and `count`/`spread`
go through `_validate_scalar` with `dtype=int, min_val=0`. It's a small
change with good test coverage.

None.

None.

- `xrspatial/bump.py:184` — the `count` docstring still reads `count : int`
  even though it defaults to `None` and is now validated only when passed.
  Adding `, optional` would match the signature. Pre-existing, and outside
  the diff, so purely optional.

### What looks good
- `numeric=False` on the `agg` check (`bump.py:342`) is the right call: `agg`
  is only a shape/coords/backend template and its dtype never reaches the
  math, so requiring a numeric dtype would reject valid templates.
- Validation order is correct — `spread` is checked unconditionally (it is
  always used), `count` only when non-None (the computed `w*h//10` default is
  always valid so there is nothing to check), and all of it runs before the
  memory guard.
- `test_bump_dask_bypasses_raster_guard` still passes, so the new 2D check
  does not disturb the lazy dask path that skips the materialization guard.
- Error messages name the parameter and the requirement, replacing the
  dispatcher's "Unsupported Array Type" and the "too many values to unpack"
  unpack failure with something a caller can act on.
- `count=0` and `spread=0` remain valid (empty bump map / single-pixel
  bumps); `test_bump_spread_zero_still_allowed` locks the latter in.

### Test coverage
Regression tests cover ndarray `agg`, 3D and 1D `agg`, negative and non-int
`count`, negative and non-int `spread`, and the `spread=0` happy path. Ran
`pytest xrspatial/tests/test_bump.py` in the worktree: 23 passed. The inline
`import pytest` per test matches the existing style in this file.

### Checklist
- [x] All implemented backends produce consistent results (validation is
      backend-agnostic; dask lazy path unaffected)
- [x] Edge cases are covered by tests (count=0, spread=0, 1D/3D agg)
- [x] Benchmark exists or is not needed (no performance change)
- [x] Docstrings present and accurate (one optional `count` nit above)

brendancol (commented):
## PR Review (follow-up): bump: validate agg, count, and spread inputs

Follow-up after two commits since the first review: `0121506f` documents the
`count` default, and `9e180e38` merges `origin/main`.

None.

None.

None outstanding. The `count : int` docstring nit from the first review is
resolved (`bump.py:202`), and main's #3610 wording won the merge, so it now
reads `count : int, optional` with the `width * height // 10` (capped at
10,000,000) default spelled out.

### Merge notes
Main advanced 19 commits, three of them touching `bump.py`: #3610 (`name=`
parameter), #3611 (isort fix), and #3612 (vectorised `_partition_bumps`). Two
conflicts, both resolved by hand:
- The `xrspatial.utils` import block: kept main's isort-compact two-line form
  and added `_validate_raster` in order (`bump.py:18`). `isort --check-only`
  and `flake8` both pass.
- The `count` docstring: took main's more specific wording.

The validation block (`bump.py:347-361`) and main's `name=name` on both
`DataArray` return paths (`bump.py:426`, `430`) coexist cleanly. #3612's
vectorised `_partition_bumps` is untouched by this branch, so the validation
sits in front of it without interfering.

### Tests
`pytest xrspatial/tests/test_bump.py` in the worktree: 25 passed (up from 23,
since main's merge added the `name=` tests). Import verified to resolve into
the worktree, not the main editable install.

### Checklist
- [x] All implemented backends produce consistent results (validation is
      backend-agnostic)
- [x] Edge cases are covered by tests (count=0, spread=0, 1D/3D agg)
- [x] Dask chunk boundaries handled correctly (#3612 path unchanged)
- [x] Benchmark exists or is not needed (main added benchmarks/benchmarks/bump.py)

PR #3620, opened 2026-07-02
docs(convolution): fill docstring gaps and fix stale examples

brendancol:
Documentation-only cleanup for `xrspatial/convolution.py`, from the deep-sweep documentation pass. No behavior changes; `test_convolution.py` still passes and flake8 is unchanged (one pre-existing F401 baseline).

## What changed

- `custom_kernel`: replaced the one-line stub with a full numpydoc docstring (Parameters, Returns, Examples). The example output was run and matches.
- `convolve_2d`: added a docstring. It had none, even though it is importable and used by focal/edge_detection/morphology.
- `convolution_2d`: the `agg` parameter listed only NumPy, CuPy, and Dask+NumPy, but the dispatch also provides Dask+CuPy (`_convolve_2d_dask_cupy`). Now names all four. Also fixed the typos "values to processed" and "CuPybacked".
- `annulus_kernel`: the two `print(kernel)` example blocks showed comma-separated arrays with a stray trailing `)`. Real `print` output has no commas or paren. Corrected to the measured output (values were already right).
- `convolution_2d` Dask example: the `.compute()` output claimed `dtype=float32`, but the `np.ones` input is float64 and stays float64. Dropped the wrong label.
- `convolution_2d` CuPy example: `type(...)` showed `cupy.core.core.ndarray`; modern CuPy reports `cupy.ndarray`. Updated.

## Not fixed (LOW, left as-is)

`calc_cellsize` km example writes its output line as `>>> (1000.0, 1000.0)` (a stray prompt on an output line). Harmless on copy-paste, so left alone and noted in the sweep state.

## Verification

- numpy, dask, and cupy docstring examples executed on a CUDA host; outputs match.
- `pytest xrspatial/tests/test_convolution.py` -> 6 passed.


brendancol (commented):
## PR Review: docs(convolution): fill docstring gaps and fix stale examples

Documentation-only change to `xrspatial/convolution.py`. I read the full file from the PR branch and executed every touched example on a CUDA host to confirm the outputs match what the docstrings now claim.

None.

None.

- `calc_cellsize` km example (convolution.py:130) still writes its output line as `>>> (1000.0, 1000.0)`, a stray prompt on an output line. The PR body already flags this as a deliberate LOW deferral. It sits in the same file and category as the other example fixes, so folding it in here would be reasonable, but leaving it is fine too since it does not error on copy-paste.

### What looks good

- Every changed example reproduces exactly. I ran them from the PR branch (import resolved into the worktree, not the main editable install):
  - `custom_kernel` repr returns the int array unchanged.
  - Both `annulus_kernel` prints match the space-separated, no-comma, no-trailing-paren `print()` output.
  - The dask `.compute()` example is float64, so dropping the `dtype=float32` label was correct (`np.ones` input stays float64 through `_promote_float`).
  - The CuPy example reports `<class 'cupy.ndarray'>` and the same values.
- The `convolution_2d` `agg` backend list now names all four backends, which matches the dispatch table (`_convolve_2d_dask_cupy` is registered as `dask_cupy_func`). The `to processed` / `CuPybacked` typos are gone.
- The new `convolve_2d` docstring's `boundary` options (`'nan'`, `'nearest'`, `'reflect'`, `'wrap'`) match `VALID_BOUNDARY_MODES` in utils.py, and its "integer inputs promoted to at least float32" note matches `_promote_float`.

### Checklist

- [x] Docstring examples match live output (executed, not eyeballed)
- [x] Backend list matches the actual dispatch registration
- [x] boundary options match the validator
- [x] No behavior change (docs only; test_convolution.py unaffected)
- [x] No stale dtype/type reprs left in the touched examples

brendancol (commented):
## PR Review (follow-up): docs(convolution)

Re-reviewing the two commits added since the first review.

### Commit 1a5987ec — drop stray prompt on calc_cellsize output line

This is the one nit from my first review. The km example output line went from `>>> (1000.0, 1000.0)` to `(1000.0, 1000.0)`, which now matches the other two output lines in the same block (`(0.5, 0.5)` and `(1.0, 1.0)`). Docs only, correct.

### Merge 9fa9e03c — origin/main

Clean merge, no conflicts. Main's #3613 (`style(convolution): drop unused import`) removed the `not_implemented_func` import that was the pre-existing F401 baseline, so flake8 on `convolution.py` is now fully clean rather than the one documented warning.

### Verification on the merged tree

- `flake8 xrspatial/convolution.py`: clean.
- `pytest xrspatial/tests/test_convolution.py`: 6 passed.
- The calc_cellsize fix survived the merge (convolution.py:129).

No new findings. Still no blockers.

PR #3622, opened 2026-07-02
Align convolution dask output dtype with numpy/cupy (#3618)

### LINKED ISSUE #3618: convolution_2d dask backend reports float64 dtype while numpy/cupy return float32
## Description

`convolution_2d` / `convolve_2d` promote an integer or float32 raster to float32 on the eager numpy and cupy backends, but the dask backends advertise `float64`. The declared dask dtype also does not match the float32 chunks that dask actually computes, because the `meta` array passed to `map_overlap` is hardcoded to the numpy default float64.

float32 rasters are common (most GeoTIFF elevation data), so a pipeline that switches from a numpy DataArray to a dask-backed one sees the output dtype change from float32 to float64 for the same input, and `result.dtype` disagrees with the values produced on compute.

## Reproduction

```python
import numpy as np
import dask.array as da
from xrspatial.convolution import convolve_2d, circle_kernel

K = circle_kernel(1, 1, 1)
data = np.arange(64, dtype=np.float32).reshape(8, 8)

eager = convolve_2d(data, K)
lazy = convolve_2d(da.from_array(data, chunks=(4, 4)), K)

print("eager :", eager.dtype)                  # float32
print("dask  :", lazy.dtype)                   # float64  (declared)
print("dask computed:", lazy.compute().dtype)  # float32  (actual)
```

Observed:

```
eager : float32
dask  : float64
dask computed: float32
```

The same mismatch happens for int32 input. For float64 input all backends agree (float64), so the bug only shows up when the input is promoted.

## Cause

In `xrspatial/convolution.py`, `_convolve_2d_dask_numpy` and `_convolve_2d_dask_cupy` cast the data with `_promote_float(data.dtype)` but pass `meta=np.array(())` / `meta=cupy.array(())`, which is float64. `map_overlap` takes the declared dtype from `meta`, so the array claims float64 even though every chunk is float32.

## Fix

Pass the promoted float dtype into `meta` so the declared dtype matches both the eager backends and the computed chunks.

brendancol:
Closes #3618.

The dask backends of `convolution_2d` / `convolve_2d` declared `float64`
via `meta=np.array(())` even when the eager numpy and cupy backends promote
an int or float32 raster to float32. So `result.dtype` changed depending on
the backend, and the declared dask dtype did not even match the float32
chunks dask computes.

## Change

- `_convolve_2d_dask_numpy` and `_convolve_2d_dask_cupy` now pass the
  promoted float dtype into the `map_overlap` `meta`, so the declared dtype
  matches both the eager backends and the computed chunks.

## Backends

numpy, cupy, dask+numpy, dask+cupy all verified on a CUDA host: for a
float32 input every backend now reports float32 (was float64 on the two
dask paths). float64 input is unchanged.

## Test plan

- [x] `pytest xrspatial/tests/test_convolution.py` (9 passed)
- [x] New parametrized test asserts the dask declared dtype and the
      computed chunk dtype both equal the eager dtype for int32, float32,
      and float64 input.


brendancol (commented):
## PR Review: Align convolution dask output dtype with numpy/cupy (#3618)

Domain-aware review (metadata sweep).

- None.

- None.

- The regression test in `xrspatial/tests/test_convolution.py:12` covers only the numpy and dask+numpy backends. The same untyped-meta bug lived on `_convolve_2d_dask_cupy` (`xrspatial/convolution.py:477`) and is fixed the same way, but there's no GPU assertion, so the cupy half of the fix rides on inspection. A `@cuda_and_cupy_available` dtype check would exercise the cupy and dask+cupy paths on hosts with a GPU. Not blocking, since the CI matrix skips GPU tests anyway.

### What looks good
- The change stays scoped to the reported defect. Both dask backends hoist `_promote_float(data.dtype)` into a local and pass it as the `meta` dtype, so the lazy DataArray advertises the same dtype the chunks compute (`np.empty_like(data)` on the already-promoted input in `_convolve_2d_numpy`). Nothing else in the two functions moves.
- Verified on a CUDA host across all four backends: int32/float32 input reports float32 everywhere, and the two dask paths used to advertise float64. float64 input is unchanged. `pytest xrspatial/tests/test_convolution.py` passes.
- The wider metadata audit came back clean. `convolution_2d` keeps name, dims, coords (passthrough coords included), and attrs (crs, res, transform, nodatavals, _FillValue) on every backend. dtype was the only cross-backend divergence.
- The pre-existing isort drift in `convolution.py` was left alone (it already fails isort on `main` and belongs in a style pass), and the second commit only fixes the isort grouping the new test imports introduced.

### Checklist
- [x] Algorithm matches reference/paper (no algorithm change; dtype metadata only)
- [x] All implemented backends produce consistent results (float32 for int/float32, float64 preserved)
- [x] Edge cases are covered by tests (int32/float32/float64 parametrized)
- [x] Dask chunk boundaries handled correctly (depth/boundary unchanged)
- [x] Benchmark exists or is not needed (metadata-only fix, no perf change)

brendancol (commented):
## Follow-up review (#3622)

Follows up on my earlier review's one nit (no GPU dtype assertion) and the sync with `main`.

- New commit `26c904e8` adds `test_convolve_2d_cupy_dtype_matches_numpy` in `xrspatial/tests/test_convolution.py:26`, a `@cuda_and_cupy_available` parametrization over int32/float32/float64 that asserts the cupy and dask+cupy outputs declare (and compute) the same promoted dtype the eager numpy backend returns. That closes the gap the nit called out, so the `meta=cupy.array((), dtype=fdtype)` half of the fix is regression-tested rather than eyeballed.
- Merged `origin/main` (19 commits). `main` had touched `convolution.py` only in a style commit (#3613); the dtype fix is unchanged and there were no conflicts.

### Verification
- Ran the new GPU test live on a CUDA host (not skipped): all three dtype cases pass on both cupy and dask+cupy. The full `xrspatial/tests/test_convolution.py` is 12 passed. isort and flake8 are clean on the changed files after the merge.

No remaining findings.

Issue #3628, opened 2026-07-03
kde/line_density NaN inputs: backends disagree and the auto extent collapses to zeros


brendancol:
**Describe the bug**

`kde()` and `line_density()` have no policy for non-finite input coordinates or weights, and the four backends disagree about what happens. One NaN row in a point table (a common artifact of empty geometries or bad CSV parses) produces three different outcomes depending on backend and call style:

1. Auto-extent path (no `template`): a single NaN coordinate makes `x_arr.min()/max()` NaN, so `x_range`/`y_range` are NaN, the coordinate vectors are NaN, and the output is a silent all-zero grid. Same for `line_density` endpoints. All backends.
2. Template path, numpy / dask+numpy / dask+cupy: the NaN point is silently dropped (the pixel-range math produces an empty loop range, and the dask tile filter's NaN comparisons are False). Reasonable outcome, but undocumented and accidental.
3. Template path, eager cupy with the gaussian kernel: the GPU kernel has no cutoff, so `exp(-0.5 * nan)` reaches every pixel and the whole grid comes back NaN. Compact kernels drop the point instead (`nan <= 1.0` is False).

A NaN weight also poisons every pixel within the cutoff, which for gaussian on any backend means the whole grid.

**Expected behavior**

One documented policy, the same on every backend. `xrspatial.interpolate` already has one (`interpolate/_validation.py`): drop non-finite points up front and raise when a non-empty input has no valid points left. kde and line_density should do the same, for coordinates and weights (and all four segment endpoints for line_density).

**Reproduction** (this host, worktree at b9ba8a85, CUDA available)

```python
import numpy as np
import xarray as xr
import cupy
import dask.array as dask_array
from xrspatial.kde import kde, line_density

t = xr.DataArray(np.zeros((16, 16)), dims=['y', 'x'],
                 coords={'y': np.linspace(-4, 4, 16), 'x': np.linspace(-4, 4, 16)})
x = np.array([0.0, 1.0, np.nan])
y = np.array([0.0, 1.0, 0.5])

for name, tt in [('numpy', t),
                 ('cupy', t.copy(data=cupy.asarray(t.values))),
                 ('dask', t.copy(data=dask_array.from_array(t.values, chunks=(8, 8)))),
                 ('dask_cupy', t.copy(data=dask_array.from_array(cupy.asarray(t.values), chunks=(8, 8))))]:
    r = kde(x, y, bandwidth=1.0, kernel='gaussian', template=tt).data
[...]

PR #3633, opened 2026-07-03
Fix kde dask backends dropping points on descending-coordinate templates (#3627)

### LINKED ISSUE #3627: kde returns all-zero or wrong output on dask backends with descending-coordinate templates
**Describe the bug**

`kde()` with a dask-backed template whose y or x coordinates descend (row 0 = north, the usual layout for geospatial rasters and the default in our own `create_test_raster` helper) returns all zeros for the compact kernels (epanechnikov, quartic). The gaussian kernel returns nonzero but wrong values, which is worse: it looks plausible. dask+cupy fails the same way. Eager numpy and cupy are fine.

This is #1198 again, one layer up. #1199 fixed the sign handling in the CPU kernels, but the dask paths pre-filter points per tile in `_filter_points_to_tile` (xrspatial/kde.py:384) and that filter still assumes positive pixel spacing:

```python
tile_x1 = tile_x0 + tile_cols * dx
tile_y1 = tile_y0 + tile_rows * dy
mask = ((xs >= tile_x0 - cutoff) & (xs <= tile_x1 + cutoff) &
        (ys >= tile_y0 - cutoff) & (ys <= tile_y1 + cutoff))
```

With negative dx/dy, `tile_x1 < tile_x0`, so the interval is inverted and most or all points are dropped before the kernel runs. The gaussian kernel's 4*bw cutoff is wide enough for a few points to survive, so its output is partially populated rather than empty.

**Expected behavior**

Dask results match numpy for any coordinate orientation, same as the eager path already guarantees via the #1199 regression tests.

**Reproduction** (this host, worktree at b9ba8a85)

```python
import numpy as np
import xarray as xr
import dask.array as dask_array
from xrspatial.kde import kde

rng = np.random.default_rng(42)
x = rng.uniform(-3, 3, 50)
y = rng.uniform(-3, 3, 50)

def template(desc_y, desc_x, dask=False):
    ys = np.linspace(4, -4, 16) if desc_y else np.linspace(-4, 4, 16)
    xs = np.linspace(4, -4, 16) if desc_x else np.linspace(-4, 4, 16)
    t = xr.DataArray(np.zeros((16, 16)), dims=['y', 'x'], coords={'y': ys, 'x': xs})
    if dask:
        t = t.copy(data=dask_array.from_array(t.values, chunks=(8, 8)))
    return t

for kernel in ['quartic', 'gaussian']:
    for desc_y, desc_x, label in [(False, False, 'ascending'), (True, False, 'desc-y'),
                                  (False, True, 'desc-x'), (True, True, 'desc-both')]:
        np_res = kde(x, y, bandwidth=1.0, kernel=kernel, template=template(desc_y, desc_x)).values
        dk_res = kde(x, y, bandwidth=1.0, kernel=kernel, template=template(desc_y, desc_x, dask=True)).values
        print(kernel, label, 'numpy.sum=%.4f' % np_res.sum(), 'dask.sum=%.4f' % dk_res.sum(),
              'max|diff|=%.3e' % np.abs(np_res - dk_res).max())
```

Output:

```
quartic ascending numpy.sum=175.9665 dask.sum=175.9665 max|diff|=0.000e+00
quartic desc-y    numpy.sum=175.9665 dask.sum=0.0000   max|diff|=3.475e+00
quartic desc-x    numpy.sum=175.9665 dask.sum=0.0000   max|diff|=3.475e+00
quartic desc-both numpy.sum=175.9665 dask.sum=0.0000   max|diff|=3.475e+00
gaussian ascending numpy.sum=171.5630 dask.sum=171.5630 max|diff|=0.000e+00
gaussian desc-y    numpy.sum=171.5640 dask.sum=134.8469 max|diff|=1.409e+00
gaussian desc-x    numpy.sum=171.5631 dask.sum=135.9322 max|diff|=9.542e-01
gaussian desc-both numpy.sum=171.5640 dask.sum=107.2278 max|diff|=1.588e+00
```

dask+cupy gives the same zeros (verified on this host with CUDA).

**Additional context**

The existing `TestDaskParity` and `TestCuPyParity` tests only use an ascending template, which is how this slipped past the #1199 regression tests. Found by the 2026-07-03 accuracy sweep of the kde module.

brendancol:
Closes #3627

- `_filter_points_to_tile` computed the tile extent as `tile_x0 .. tile_x0 + cols*dx`, which inverts when dx/dy are negative. Descending-coordinate templates (the normal north-up raster layout) lost most or all of their points before the kernel ran: compact kernels returned all zeros, gaussian returned partially wrong values. The fix orders the tile edges with min/max before widening by the cutoff.
- Affected dask+numpy and dask+cupy only. The eager paths were already fixed for this in #1199; the tile filter sits above them and was missed.
- The sweep state CSV update for the kde accuracy audit rides along in this commit.

Backends: numpy and cupy unchanged (already correct); dask+numpy and dask+cupy fixed and now match numpy exactly on descending-y, descending-x, and both-descending templates.

Test plan:
- [x] New `test_dask_matches_numpy_descending_coords` (3 orientations x 3 kernels) fails on main, passes with the fix
- [x] New `TestDaskCupyDescending` runs the dask+cupy path on a GPU host (CUDA verified locally)
- [x] Full `test_kde.py` suite: 65 passed
- [x] Repro script from #3627 shows max|diff|=0 across all orientations after the fix


brendancol (commented):
## PR Review: Fix kde dask backends dropping points on descending-coordinate templates (#3627)

- none

- [ ] xrspatial/tests/test_kde.py (TestDaskCupyDescending): the dask+cupy regression test covers only the descending-y quartic case. The filter is shared code, so this is enough to pin the bug, but a desc-x case would be cheap and would catch any future divergence between the two dask wrappers (they duplicate the tile loop).

- [ ] xrspatial/kde.py:391-401: `tile_x1 = tile_x0 + tile_cols * dx` overshoots the last pixel centre by one spacing, so the filter keeps a slightly wider band than strictly needed. That is the safe direction (never drops a contributing point) and matches the pre-existing behavior on ascending grids; worth a half-sentence in the comment if anyone touches this again.

### What looks good
- The fix is minimal: order the tile edges with min/max before widening by the cutoff, exactly parallel to what #1199 did inside the kernels.
- Verified locally on a CUDA host: dask+numpy and dask+cupy now match numpy exactly (max abs diff 0 for quartic, fringe-only for gaussian) on desc-y, desc-x, and desc-both templates. Before the fix, compact kernels returned all zeros and gaussian returned partially wrong values.
- The new parametrized test (3 orientations x 3 kernels) fails on main and passes here; tolerance choices mirror the existing #1199 regression tests (exact for compact kernels, atol=1e-5 for the gaussian 4*bw box-cutoff fringe).
- The sweep-accuracy state CSV update rides along per the sweep contract; it does not touch code.

### Checklist
- [x] Algorithm matches reference/paper (filter now consistent with the kernels' own cutoff logic)
- [x] All implemented backends produce consistent results (verified numpy / cupy / dask+numpy / dask+cupy on this host)
- [x] NaN handling is correct (unchanged by this PR; tracked separately in #3628)
- [x] Edge cases are covered by tests (3 orientations x 3 kernels + dask+cupy)
- [x] Dask chunk boundaries handled correctly (per-tile origins already sign-correct; only the filter was wrong)
- [x] Benchmark exists or is not needed (bug fix, no perf-relevant change: the filter does the same amount of work)
- [x] Docstrings present and accurate (helper docstring unchanged, still correct)

brendancol (commented):
## PR Review follow-up (#3627)

Disposition of the earlier findings, after ae16e5e8:

- Suggestion (dask+cupy desc-x coverage): fixed. `TestDaskCupyDescending` is now parametrized over desc-y and desc-x; 66 tests pass locally, both GPU variants executed on this host.
- Nit (filter overshoot comment): fixed. The comment now says the tile edge overshoots the last pixel centre by one spacing and why that is the safe direction.

No new findings. No blockers.

brendancol (commented):
## PR Review: Fix kde dask backends dropping points on descending-coordinate templates (#3627)

Re-review after merging current `main` (SHA `228fadf5`). This pulls in the libjxl=0.11.* CI pin (#3639), which was the only reason CI was red. The merge was clean with no conflicts; kde.py was untouched by main.

### Correctness
The fix in `_filter_points_to_tile` (xrspatial/kde.py:396-402) is right. With descending coordinates `dx`/`dy` go negative, so `tile_x1 = tile_x0 + tile_cols*dx` falls below `tile_x0`. The previous bounds `xs >= tile_x0 - cutoff` and `xs <= tile_x1 + cutoff` then described an inverted interval and masked out every contributing point. Ordering the edges with `min`/`max` before widening by `cutoff` restores a valid interval. The `tile_x1`/`tile_y1` one-spacing overshoot past the last pixel centre keeps the filter conservative, so it never drops a real contributor.

Both dask backends route through this one helper (xrspatial/kde.py:436 for numpy, :475 for cupy), so the fix covers dask+numpy and dask+cupy together.

### Tests
`test_dask_matches_numpy_descending_coords` covers all three descending combinations across gaussian/epanechnikov/quartic, with a `sum() > 0` guard plus per-kernel tolerances. `TestDaskCupyDescending` adds the same check on the GPU path. Ran locally against the worktree:

```
xrspatial/tests/test_kde.py .................................................. 66 passed in 1.29s
```

The dask+cupy cases execute on this box (CUDA present), not skipped.

### What looks good
- Change is scoped to the bug, nothing else touched.
- The comment explains why the ordering matters and cites the issue.
- Tests reproduce the original failure across kernels and both dask backends.

No blockers. Ready to merge once CI goes green with the main merge.

PR #3634, opened 2026-07-03
Aim sky_view_factor rays in ground azimuth space (#3626)

### LINKED ISSUE #3626: sky_view_factor: ray azimuths uniform in cell-index space, biasing SVF on rectangular cells
**Describe the bug**

`sky_view_factor()` spaces its `n_directions` rays evenly in cell-index angle: `angle = 2 * pi * d / n_directions`, then `dx = cos(angle)`, `dy = sin(angle)` in units of cells (CPU kernel around line 156, GPU kernel around line 204). Since #1410 the horizon angle along each ray uses true ground distance, but the ray directions are still cell-index directions. With square cells the two agree. With rectangular cells (`cellsize_x != cellsize_y`) a direction set that is uniform in cell space is not uniform on the ground, so the equal-weight mean over directions oversamples some ground azimuths and undersamples others. The SVF comes out biased.

The docstring says rays are cast at "evenly spaced azimuths", which a user will read as ground azimuths. Anisotropic cell sizes are a supported input; `test_anisotropic_cellsize` exercises them.

**Reproduction**

Same physical scene, different grids: a planar ramp rising east at 45 degrees (analytic SVF = 0.75 away from edges), sampled at several cell aspect ratios.

```python
import numpy as np
import xarray as xr
from xrspatial import sky_view_factor

def svf_center(cx, cy):
    ny = nx = 121
    cols = np.arange(nx, dtype=np.float64)
    z = np.broadcast_to(cols * cx, (ny, nx)).copy()  # 45-deg ramp on the ground
    agg = xr.DataArray(z, dims=['y', 'x'])
    out = sky_view_factor(agg, max_radius=40, n_directions=64,
                          cellsize_x=cx, cellsize_y=cy)
    return float(out.data[60, 60])

for cx, cy in [(1.0, 1.0), (1.0, 2.0), (2.0, 1.0), (1.0, 4.0)]:
    print((cx, cy), round(svf_center(cx, cy), 4))
```

Output on main (b9ba8a85):

```
(1.0, 1.0) 0.7284
(1.0, 2.0) 0.7731
(2.0, 1.0) 0.6927
(1.0, 4.0) 0.8147
```

Cell aspect alone moves the result by up to 0.12 (16% relative) on an identical surface. The square-cell value sits below 0.75 too, but that offset comes from nearest-cell ray sampling (the max along a rounded ray picks up azimuth jitter) and applies equally at every aspect ratio. The spread between the rows is the azimuth-weighting bias. The weighting effect can also be computed without any raster, from exact horizon angles: equal weights at ground azimuths `atan2(cy*sin(t), cx*cos(t))` give +0.052 error at 2:1 and +0.105 at 4:1.

**Expected behavior**

The same terrain gives roughly the same SVF whatever the grid's cell aspect ratio. Rays should be spaced evenly in ground azimuth: pick `phi = 2 * pi * d / n_directions`, convert to cell space as `(cos(phi) / cellsize_x, sin(phi) / cellsize_y)`, and renormalize to a unit cell-space step. For square cells this reduces to the current directions, so results on square grids do not change.

**Environment**
- xarray-spatial main (b9ba8a85), Linux, Python 3.14, numba CPU + CUDA

**Additional context**

Found by the accuracy sweep. All four backends share the direction code, so all four are affected the same way. Follow-up to #1407, which fixed the ground-distance half of anisotropy handling but not the direction spacing.

brendancol:
Closes #3626

- `_svf_cpu` and `_svf_gpu` now start each ray from a ground azimuth and convert it to a unit step in cell-index space (`cos(phi)/cellsize_x`, `sin(phi)/cellsize_y`, renormalized). With square cells this reduces to the old directions; a 40x40 random grid produces bit-identical output before and after.
- On the issue's repro (planar 45-degree ramp, analytic SVF 0.75), cell aspect 1:1 / 1:2 / 2:1 / 1:4 gave 0.728 / 0.773 / 0.693 / 0.815 before and 0.728 / 0.723 / 0.723 / 0.709 after. The residual spread comes from nearest-cell ray sampling, which the issue documents as a separate, aspect-independent effect.
- Docstrings now say azimuth spacing is uniform on the ground.

Backend coverage: the CPU and CUDA kernels carry the same change, so numpy, cupy, dask+numpy, and dask+cupy all pick it up. Verified on a CUDA host.

Test plan:
- [x] `pytest xrspatial/tests/test_sky_view_factor.py` -- 38 passed on a CUDA host (cupy and dask+cupy paths ran, not skipped)
- [x] new `test_anisotropic_cellsize_ground_truth_consistency` fails on main (diffs 0.046 / 0.035 vs 0.02 tolerance), passes here
- [x] new anisotropic numpy==dask and numpy==cupy parity tests
- [x] flat=1.0, flip/rot90/translate invariants, and NaN-holes-at-chunk-boundary parity probes all unchanged at machine epsilon

Also carries the accuracy-sweep state CSV row for this module (2026-07-03 re-audit).


brendancol (commented):
## PR Review: Aim sky_view_factor rays in ground azimuth space (#3626)

None.

- [ ] `xrspatial/sky_view_factor.py:159-166` (and the GPU twin at 213-220): the direction table (`dx`, `dy` per azimuth) is recomputed for every pixel. That was true before this PR too, and the inner radius loop dominates runtime, so this is not a merge concern. If the kernels get touched again, precomputing the `n_directions` unit vectors once per array would drop two divides and a sqrt per pixel-direction.

- [ ] `xrspatial/sky_view_factor.py:326-329`: the new docstring text explains the behavior well. A sentence noting the residual limit would round it out: at extreme aspect ratios (say 10:1) the integer-cell ray marching still quantizes the sampled azimuths, so the evenly spaced azimuths are nominal, met to within one cell of rounding. The new consistency test only covers 2:1, which is the realistic range.
- [ ] No asv benchmark exists for `sky_view_factor` (pre-existing gap, and this PR adds no new function). Flagging for the benchmark sweep rather than for this PR.

### What looks good
- The fix is minimal and mathematically right: dividing the ground-azimuth unit vector by cell size and renormalizing gives a unit cell-index step, so `map_overlap` depth semantics (`depth == max_radius`) are untouched.
- Square-cell output is bit-identical to main (verified on a 40x40 random grid), so no existing user results change.
- `test_anisotropic_cellsize_ground_truth_consistency` fails on main (diffs 0.046/0.035 vs the 0.02 tolerance) and passes here with 3x headroom (0.0065/0.0061). It encodes a physical invariant (grid aspect cannot change the SVF of the same terrain), not a snapshot value.
- The changed CUDA path gets its own parity tests (`test_anisotropic_numpy_equals_dask`, `test_anisotropic_numpy_equals_cupy`); the full suite ran on a CUDA host with cupy and dask+cupy executing, 38 passed.
- `dnorm` cannot be zero: cos and sin cannot both vanish and the public wrapper floors non-positive cell sizes to 1.0 before the kernels run.

### Checklist
- [x] Algorithm matches reference/paper (Zakšek et al. 2011 formula unchanged; only azimuth spacing corrected)
- [x] All implemented backends produce consistent results (CPU and CUDA kernels carry the same change; parity tests added)
- [x] NaN handling is correct (untouched by this PR; center-NaN and ray-truncation semantics unchanged)
- [x] Edge cases are covered by tests (flat surface, cell-size overrides, anisotropic aspect ratios)
- [x] Dask chunk boundaries handled correctly (direction depends only on scalar cell sizes, identical per chunk; depth unchanged)
- [ ] Benchmark exists or is not needed (no sky_view_factor benchmark; pre-existing gap, see nit)
- [x] README feature matrix updated (not applicable, no new function or backend change)

brendancol (commented):
## PR Review: follow-up after 66d805db

Disposition of the first review's findings:

- Suggestion (per-pixel direction table): fixed for the CPU kernel. `_svf_cpu` now fills `dxs`/`dys` once before the pixel loops (`xrspatial/sky_view_factor.py:148-159`) and output is bit-identical to the previous commit on a 40x40 random grid. The CUDA kernel keeps per-thread computation: `cuda.local.array` needs a compile-time size and `n_directions` is a runtime argument, so a table there means a kernel signature change threaded through `_run_cupy` and `_run_dask_cupy`. Out of scope for an accuracy fix; per-thread trig is normal CUDA style anyway.
- Nit (docstring): added. The public docstring now says azimuth spacing is nominal to within one cell of rounding and quantizes toward the grid axes beyond roughly 4:1 aspect (`xrspatial/sky_view_factor.py:326-332`).
- Nit (missing asv benchmark): not addressed here, routed to the benchmark sweep. Pre-existing gap, no new function in this PR.

Checks on the delta itself: the hoisted table is filled with the same expressions in the same order, `np.empty` inside `@ngjit` is fine, and the full test file passes (38 tests, CUDA host). No new findings.

brendancol (commented):
## PR Review: Aim sky_view_factor rays in ground azimuth space (#3626) — follow-up

Follow-up after updating the branch with `main`. Confirms the CI fix and that the merge left the ground-azimuth change intact.

### Branch update
- Merged `origin/main` at `c4c3ec3e`. That pulls in the libjxl 0.11.* pin from #3639, which the stale cog-validator and geotiff-corpus jobs were tripping over. The red CI here was environmental, not caused by the PR.
- Clean merge, no conflicts. Main's changes were in unrelated modules (bilateral, bump, convolution, sweep-state CSVs); sky_view_factor was untouched.

### Re-check of the core change
- `_svf_cpu` and `_svf_gpu` both build each ray direction from a ground azimuth, divide by `cellsize_x`/`cellsize_y`, then normalize to a unit step in cell-index space. The horizon angle still uses true ground distance (`gx = (sx - x) * cellsize_x`, `gy = (sy - y) * cellsize_y`), so step length stays in ground units. Both paths compute the direction identically.
- The docstring is honest about the limit: spacing is nominal to within one cell of rounding, and extreme aspect ratios quantize toward the grid axes.

### Tests
- `test_sky_view_factor.py`: 38 passed.

### Remaining blockers
- None from this pass. CI should go green once it reruns against the merged tip.

Issue #3631, opened 2026-07-03
HPA* fallback in a_star_search returns a partial cost trail when no path exists


brendancol:
**Describe the bug**

When `a_star_search` falls back to hierarchical pathfinding (auto search radius under memory pressure with a long start-goal distance), `_hpa_star_search` can return a partially written cost surface. If a refinement segment fails to connect after the radius escalation loop, the function bails with

```python
return path_img  # partial result
```

leaving the pixels of already-refined segments holding finite costs even though no start-to-goal path was found. Everywhere else in the module, "no path" means an all-NaN output (the plain kernel, the bounded search, and `test_search_radius_too_small` all follow that contract), so a caller checking `np.isfinite(result).any()` for reachability gets a false positive, and the finite trail dead-ends partway across the grid.

Reproduction (run on this host):

```python
import numpy as np
from xrspatial.pathfinding import _hpa_star_search, _neighborhood_structure

# grid fully split by a 1-px NaN wall: no path exists
H = W = 200
data = np.ones((H, W))
data[:, 100] = np.nan
dy, dx, dd = _neighborhood_structure(1.0, 1.0, 8)
barriers = np.array([], dtype=np.float64)

path = _hpa_star_search(data, None, 0, 0, 199, 199, barriers, dy, dx, dd,
                        1.0, False, 1.0, 1.0, H, W)
print(np.isfinite(path[199, 199]))   # False - goal not reached
print(np.isfinite(path).sum())       # 89   - but 89 pixels carry finite costs
```

The coarse grid sees the wall's blocks as passable (each block contains non-NaN cells), routes across, and the fine refinement then fails at the wall after writing the earlier segments.

**Expected behavior**

An all-NaN surface when the goal cannot be reached, consistent with the other search paths in the module.

**Additional context**

Found by the accuracy sweep (Cat 2/5, rated MEDIUM). Reachable through the public API only in the auto-radius HPA* regime (grid too large for RAM and manhattan distance > 1000), which is also exactly the setting where a user cannot easily cross-check against a full-grid run.

Issue #3630, opened 2026-07-03
multi_stop_search crashes on dask+cupy input and returns numpy-backed output for cupy input


brendancol:
**Describe the bug**

`multi_stop_search()` mishandles GPU-backed inputs in its segment-stitching loop:

```python
seg_data = seg.data
if hasattr(seg_data, 'get'):
    seg_vals = seg_data.get()  # cupy -> numpy
else:
    seg_vals = np.asarray(seg.values)
```

1. dask+cupy input crashes. For a dask-of-cupy array, `seg.data` is a `dask.array.Array` (no `.get`), so the code falls into `np.asarray(seg.values)`. Computing `.values` yields a cupy array and numpy raises `TypeError: Implicit conversion to a NumPy array is not allowed`. `a_star_search` itself handles dask+cupy, so the wrapper breaks a backend the underlying function supports. `_optimize_waypoint_order` has the same pattern, so `optimize_order=True` hits it too.

2. cupy input returns a numpy-backed result. The stitched output is assembled in `np.full(...)` and never converted back, so a cupy (or dask) input comes back as a numpy-backed DataArray. `a_star_search` documents "DataArray of the same type as `surface`" and honors it; `multi_stop_search` quietly downgrades the backend. The existing `test_multi_stop_cupy_matches_numpy` doesn't catch this because its conversion expression is a no-op (`x if not hasattr(x, 'get') else x`).

Reproduction (run on this host, CUDA available):

```python
import numpy as np, xarray as xr
import dask.array as da, cupy
from xrspatial import multi_stop_search

def make(backend):
    h = w = 8
    r = xr.DataArray(np.ones((h, w)), dims=['y', 'x'], attrs={'res': (1.0, 1.0)})
    r['y'] = np.linspace(h - 1, 0, h); r['x'] = np.linspace(0, w - 1, w)
    if 'dask' in backend:
        r.data = da.from_array(r.data, chunks=(4, 4))
    if 'cupy' in backend:
        r.data = r.data.map_blocks(cupy.asarray) if isinstance(r.data, da.Array) else cupy.asarray(r.data)
    return r

wps = [(7.0, 0.0), (4.0, 3.0), (0.0, 7.0)]

print(type(multi_stop_search(make('cupy'), wps).data))
# -> <class 'numpy.ndarray'>   (input was cupy)

multi_stop_search(make('dask+cupy'), wps)
# -> TypeError: Implicit conversion to a NumPy array is not allowed.
```

[...]

PR #3635, opened 2026-07-03
Nearest-center point mapping and out-of-bounds rejection in pathfinding _get_pixel_id

### LINKED ISSUE #3629: a_star_search maps out-of-bounds points to interior pixels and truncates instead of rounding in _get_pixel_id
**Describe the bug**

`_get_pixel_id()` in `xrspatial/pathfinding.py` converts a `(y, x)` point to a pixel index with

```python
py = int(abs(point[0] - y_coords[0]) / cellsize_y)
px = int(abs(point[1] - x_coords[0]) / cellsize_x)
```

Two problems come out of this:

1. `abs()` folds out-of-bounds points back inside the raster. A point past the edge on the `coords[0]` side gets a positive offset and maps to a mirrored interior pixel. `a_star_search` then runs from the wrong start pixel instead of raising the "start location outside the surface graph" error it raises for points past the opposite edge. `multi_stop_search`'s waypoint bounds check has the same blind spot.

2. `int()` truncates instead of rounding to the nearest cell center. Any point in the upper half of a cell is assigned to the neighboring lower-index cell (a half-cell bias), and a point that should land exactly on a cell center can flip to the previous cell through float noise: on a grid with 0.1 resolution, `x=0.3` maps to pixel 2 because the pixel-3 center is `0.30000000000000004`.

```python
import numpy as np
import xarray as xr
from xrspatial.pathfinding import _get_pixel_id

h = w = 5
agg = xr.DataArray(np.ones((h, w)), dims=['y', 'x'], attrs={'res': (1.0, 1.0)})
agg['y'] = np.linspace(h - 1, 0, h)   # [4, 3, 2, 1, 0]
agg['x'] = np.linspace(0, w - 1, w)

# 1) start y=5.0 is OUTSIDE the raster (top edge is y=4.0)
path = a_star_search(agg, (5.0, 0.0), (0.0, 4.0))
print(np.argwhere(path.values == 0))   # -> [[1 0]]  path starts at row 1, no error

# 2) truncation
agg2 = xr.DataArray(np.ones((2, 10)), dims=['y', 'x'])
agg2['y'] = [0.1, 0.0]
agg2['x'] = np.arange(10) * 0.1
print(_get_pixel_id((0.0, 0.3), agg2, 'x', 'y'))   # -> (0, 2), pixel-3 center is 0.30000000000000004
print(_get_pixel_id((0.0, 2.6), agg, 'x', 'y'))    # -> (?, 2) on a unit grid, nearest center is 3
```

Observed output:

```
[[1 0]]
(0, 2)
(2, 2)
```

**Expected behavior**

- A point outside the raster bounds raises `ValueError` ("start location outside the surface graph") regardless of which side it falls on.
- A point maps to the cell whose center is nearest, so `x=0.3` on a 0.1-resolution grid resolves to pixel 3 and `x=2.6` on a unit grid resolves to pixel 3.

The fix is to compute a signed offset using the coordinate direction and round to the nearest index; negative indices then fail the existing `_is_inside` check and raise as intended.

**Additional context**

Found by the accuracy sweep (Cat 1/Cat 3). The folding half is rated HIGH (silent wrong path from a mis-mapped start/goal), the truncation half MEDIUM. Both `a_star_search` and `multi_stop_search` route through `_get_pixel_id`. Existing tests only use points that sit exactly on cell centers, so they pass under either convention.

brendancol:
Closes #3629

- `_get_pixel_id` now computes a signed offset from `coords[0]` (handles ascending and descending coords) instead of `abs()`, so a point outside the raster gets an out-of-range index and `a_star_search` / `multi_stop_search` raise their existing "outside the surface" errors on every side instead of silently pathing from a mirrored interior pixel.
- The index is rounded to the nearest cell center instead of truncated. A point in the upper half of a cell no longer shifts down one pixel, and a user-typed `0.3` on a 0.1-resolution grid (pixel-3 center `0.30000000000000004`) resolves to pixel 3 rather than pixel 2.
- Also carries the accuracy-sweep state CSV row for the pathfinding module (findings #3629, #3630, #3631).

Backend coverage: `_get_pixel_id` runs on the host for every backend, so the change applies uniformly to numpy, cupy, dask+numpy, and dask+cupy inputs.

Test plan:
- [x] New tests: nearest-center mapping (descending and ascending coords, float-noise case), out-of-bounds raises on all four sides for start and goal, out-of-bounds waypoint raises in `multi_stop_search`
- [x] `pytest xrspatial/tests/test_pathfinding.py` (55 passed, includes cupy and dask+cupy on a CUDA host)
- [x] `pytest xrspatial/tests/test_cost_distance.py` (89 passed)


brendancol (commented):
## PR Review: Nearest-center point mapping and out-of-bounds rejection in pathfinding _get_pixel_id

- none

- [ ] `xrspatial/pathfinding.py` `a_star_search` docstring (`start`/`goal` params): the point-to-pixel convention is now nearest-cell-center with out-of-bounds rejection on every side. Worth a sentence in the docstring, since this PR also tightens behavior for points more than half a cell past the far edge (previously tolerated up to a full cell via truncation, now a `ValueError`).

- [ ] `xrspatial/pathfinding.py:46` sign detection comes from the coords while the cell size still comes from `get_dataarray_resolution` (which prefers `attrs['res']`). If `attrs['res']` disagrees with the actual coord spacing the mapping is off, but that trade-off predates this PR and is consistent with the rest of the module.
- [ ] Python `round()` uses half-to-even, so a point exactly halfway between two centers snaps to the even index. Deterministic and either neighbor is defensible; noting it so nobody reads it as a bug later.

### What looks good
- The signed-step change makes the existing `_is_inside` validation actually fire on the `coords[0]` side; the fix reuses the module's own error paths instead of adding new checks.
- Tests cover descending and ascending coords, the float-noise case at an exact center (`0.3` on a 0.1-res grid), all four out-of-bounds sides for both start and goal, and the `multi_stop_search` waypoint check.
- Existing suite still passes (55 tests including cupy and dask+cupy on a CUDA host), and the fixture points in older tests sit exactly on cell centers, so no codified behavior changed.

### Checklist
- [x] Algorithm matches reference/paper (nearest-center mapping, verified by hand cases in tests)
- [x] All implemented backends produce consistent results (mapping runs host-side for all four)
- [x] Benchmark exists or is not needed (scalar point mapping, not benchmark-relevant)
- [x] Docstrings present and accurate (see suggestion about documenting the convention)

brendancol (commented):
## PR Review follow-up

The one suggestion from the first pass is addressed in 481b69ef: the `start`/`goal` parameter docs now state the nearest-cell-center mapping and that out-of-bounds points raise `ValueError`. Both nits were informational (pre-existing `attrs['res']` precedence, half-to-even rounding at exact midpoints) and need no code change. Tests re-run after the docstring commit: 55 passed. No open findings.

brendancol (commented):
## PR Review: Nearest-center point mapping and out-of-bounds rejection in pathfinding `_get_pixel_id`

Re-review after merging `origin/main` (SHA 07b28b7b). The merge was clean, no
conflicts; it pulls in the libjxl=0.11.* CI pin (#3639) that was the only reason
CI showed red. `xrspatial/pathfinding.py`, the tests, and the accuracy-sweep CSV
row are untouched by the merge.

None.

None.

- `_get_pixel_id` (pathfinding.py:59-62): a point sitting exactly on a cell
  boundary produces a `.5` offset, so `round()` uses banker's rounding (half to
  even) and the tie can land on either neighbor. That is an inherent ambiguity of
  nearest-center for a boundary point and does not affect the tested cases, so no
  change is needed. Noting it only so a future reader is not surprised.

### What looks good
- Direction and magnitude are now separated cleanly: magnitude from
  `get_dataarray_resolution` (taken as `abs`), sign derived from the coord order
  (pathfinding.py:47-54). This is what makes an out-of-bounds point on the
  `coords[0]` side produce an out-of-range index instead of folding back inside
  through `abs()`.
- Out-of-bounds indices flow to the existing `_is_inside` guards
  (pathfinding.py:957-961 for a_star, 1430-1433 for multi_stop), so the fix reuses
  the current error contract rather than adding a new one.
- Tests exercise the real failure modes: nearest-center on descending and
  ascending coords, the `0.3` on a `0.1`-res grid float-noise case, out-of-bounds
  rejection on all four sides for both start and goal, and an out-of-bounds
  multi_stop waypoint. All 55 tests in test_pathfinding.py pass on the merged tree.

### Backend note
`_get_pixel_id` runs on host coords for every backend, so numpy, cupy, dask+numpy,
and dask+cupy all share the corrected mapping. No backend-specific path to split.

### Checklist
- [x] Algorithm matches intent (nearest cell center, signed offset)
- [x] All implemented backends produce consistent results (host-side mapping)
- [x] NaN handling unchanged (not in scope for this function)
- [x] Edge cases covered by tests (ascending/descending, float noise, all four out-of-bounds sides)
- [x] No dask chunk-boundary concern (point-to-index mapping, not a neighborhood op)
- [x] Benchmark not needed (index arithmetic, not a hot path)
- [x] README feature matrix unaffected (no new public function)
- [x] Docstrings updated for start/goal out-of-bounds behavior

PR #3636, opened 2026-07-03
Drop non-finite points and segments in kde/line_density (#3628)

### LINKED ISSUE #3628: kde/line_density NaN inputs: backends disagree and the auto extent collapses to zeros

brendancol:
Closes #3628

- `kde()` now drops points with non-finite coordinates or weights before doing anything else, and `line_density()` does the same for segments with non-finite endpoints or weights. This is the policy `xrspatial.interpolate` already uses (`interpolate/_validation.py`).
- Fixes three inconsistent failure modes from one NaN row: the auto-computed extent going NaN (all-zero grid with NaN coords, every backend), eager cupy gaussian returning an all-NaN grid, and the other backends silently skipping the point.
- Raises `ValueError` when a non-empty input has no finite entries left. Truly empty input behaves as before.
- Both docstrings document the behavior.

Backends: the filter runs before dispatch, so numpy, cupy, dask+numpy, and dask+cupy now agree. Verified on this host with CUDA: all four backends return identical finite output for a NaN-containing input.

Test plan:
- [x] New `TestNonFiniteInputs` class: NaN/Inf points, NaN weights, NaN auto-extent, all-NaN raise, cupy gaussian parity, line_density variants
- [x] The cupy gaussian test fails on main (all-NaN grid), passes with the fix
- [x] Full `test_kde.py` suite: 64 passed


brendancol (commented):
## PR Review: Drop non-finite points and segments in kde/line_density (#3628)

- none

- [ ] xrspatial/tests/test_kde.py (test_nan_point_auto_extent_not_poisoned): the assertion is only "coords finite and sum > 0". Since the filter runs before the extent is derived, the NaN-containing call should produce a grid identical to the clean two-point call; asserting that equality pins the behavior much harder than sum > 0. Same for the line_density auto-extent test.
- [ ] xrspatial/tests/test_kde.py: no test combines a NaN point with bandwidth='silverman'. Before this PR the NaN reached np.std/np.percentile and made the bandwidth NaN; the filter fixes that path too, but nothing guards it against regression.

- [ ] The GeoDataFrame input path inherits the filter (it runs after points_from_geodataframe), but no test feeds a GeoDataFrame with a NaN-coordinate point through kde. Optional since geopandas empty/invalid geometries may be caught earlier; fine to leave if that turns out to be dead territory.

### What looks good
- The filter placement is right: after the length checks (so mismatched inputs still raise on original lengths), before bandwidth selection and extent derivation (so both see only finite data), and before backend dispatch (so all four backends behave identically).
- The empty-input case is preserved: `valid.all()` on an empty array is True, so `kde([], [], template=...)` behaves exactly as before, and the new ValueError only fires when a non-empty input loses everything.
- Verified on this CUDA host: numpy, cupy, dask+numpy, and dask+cupy now return identical finite output for a NaN-containing input; before the fix eager cupy gaussian returned a 256-pixel all-NaN grid.
- Policy matches interpolate/_validation.py rather than inventing a new convention, and both docstrings say what happens.

### Checklist
- [x] Algorithm matches reference/paper (n/a, input validation change)
- [x] All implemented backends produce consistent results (verified all four on this host)
- [x] NaN handling is correct (that is the PR)
- [x] Edge cases are covered by tests (NaN/Inf coords, NaN weights, all-NaN raise, auto extent, cupy parity, line_density variants)
- [x] Dask chunk boundaries handled correctly (filter is pre-dispatch, no chunk interaction)
- [x] No premature materialization or unnecessary copies (one boolean mask; fancy-index only when something is dropped)
- [x] Benchmark exists or is not needed (validation-only, negligible cost)
- [x] Docstrings present and accurate (both updated)

brendancol (commented):
## PR Review follow-up (#3628)

Disposition of the earlier findings, after 4ab2aaee:

- Suggestion (exact-equality auto-extent assertions): fixed for both kde and line_density. The NaN-containing calls are now asserted byte-identical to the finite-only calls, coordinates included.
- Suggestion (silverman + NaN coverage): fixed. New test compares bandwidth='silverman' with and without a NaN point; results must be identical.
- Nit (GeoDataFrame NaN point): fixed. New test feeds a GeoDataFrame containing Point(nan, 0.5) and asserts the output matches the finite-only array-input call. geopandas is importorskip-gated like the interpolate tests.

66 tests pass locally, cupy paths included. No new findings. No blockers.

brendancol (commented):
## PR Review: Drop non-finite points and segments in kde/line_density (#3628)

Re-review after merging current `main` into the branch. The merge was clean (no conflicts; `main` did not touch `kde.py`) and pulls in the libjxl=0.11.* CI pin (#3639) that was the only cause of the red cog-validator/geotiff-corpus checks. Full `test_kde.py` passes locally: 66 passed.

- None.

- None.

- `kde.py:599` and `kde.py:778`: the drop happens on the host arrays before dispatch, so every backend gets pre-filtered input and the numpy-only tests are sufficient. Worth a one-line note in the test class that the filter is intentionally backend-independent, so a future reader does not add redundant dask/cupy drop tests. Optional.

### What looks good
- The filter sits before extent computation, silverman bandwidth, and dispatch in both `kde()` and `line_density()`, so the auto-extent, bandwidth, and all four backends see the same finite subset. That is the right place to fix the divergence described in #3628.
- `n` is only read for weight-length validation ahead of the filter, so leaving it un-updated after the drop is fine (no stale-length bug downstream).
- The empty-after-filter `ValueError` is gated on `not valid.all()`, so truly empty input keeps its prior behavior rather than newly raising.
- Tests cover the reported failure modes directly: NaN/Inf point drop matching a clean call, NaN weight, un-poisoned auto-extent and silverman bandwidth, GeoDataFrame path, all-NaN raise, the eager cupy gaussian all-NaN regression, and the line_density variants.
- The cupy parity test's `rtol=1e-2, atol=1e-6` and the comment about the GPU kernel lacking the CPU's 4*bw cutoff match the tolerances already used in `TestCuPyParity`.

### Checklist
- [x] Algorithm matches reference/paper (mirrors `interpolate/_validation.py` policy)
- [x] All implemented backends produce consistent results (filter is pre-dispatch)
- [x] NaN handling is correct (`np.isfinite`, drops NaN and Inf)
- [x] Edge cases are covered by tests (all-NaN raise, Inf, NaN weights, auto-extent)
- [x] Dask chunk boundaries handled correctly (unchanged; filter precedes dask path)
- [x] Benchmark exists or is not needed (behavior fix, no perf-critical change)
- [ ] README feature matrix updated (not applicable; no new function)
- [x] Docstrings present and accurate (both functions document the drop and the raise)

PR #3637, opened 2026-07-03
Fix multi_stop_search on cupy and dask+cupy backends

### LINKED ISSUE #3630: multi_stop_search crashes on dask+cupy input and returns numpy-backed output for cupy input

brendancol:
Closes #3630

- Segment results are now materialized through a `_segment_to_numpy` helper that computes dask arrays first and then `.get()`s cupy blocks, in both the stitching loop and `_optimize_waypoint_order`. This fixes the `TypeError: Implicit conversion to a NumPy array is not allowed` crash on dask+cupy input.
- The stitched output is converted back to the input's array type (cupy for cupy input, dask with cupy blocks for dask+cupy, dask for dask+numpy), matching what `a_star_search` already does.
- `test_multi_stop_cupy_matches_numpy` had a no-op conversion expression (`x if not hasattr(x, 'get') else x`) that hid the numpy-backed output; it now asserts the output is cupy-backed and converts with `.get()`. Adds `test_multi_stop_dask_cupy_matches_numpy` covering the previously untested backend, including `optimize_order=True`.

Backend coverage: numpy unchanged, cupy and dask+numpy get output-type fixes, dask+cupy goes from crashing to working.

Test plan:
- [x] `pytest xrspatial/tests/test_pathfinding.py` (52 passed on a CUDA host, cupy and dask+cupy tests executed)
- [x] Repro from #3630 rerun: dask+cupy returns a dask array of cupy blocks with the correct goal cost, cupy input returns a cupy-backed DataArray


brendancol (commented):
## PR Review: Fix multi_stop_search on cupy and dask+cupy backends

- none

- [ ] `xrspatial/pathfinding.py` `multi_stop_search` docstring, Returns section: now that the output matches the input's array type, saying so would make the contract explicit (a_star_search's docstring already does).

- [ ] `xrspatial/pathfinding.py:1495` the dask branch wraps a fully materialized numpy array back into a dask array, so the graph carries the whole stitched surface. That is inherent to the dense stitching design (each segment was already computed to numpy before this PR too), not a regression; worth remembering if multi_stop ever needs the sparse treatment a_star_search's dask path has.
- [ ] `_segment_to_numpy` computes one full-size array per segment on dask inputs. Same pre-existing behavior as `np.asarray(seg.values)`, just centralized; noting it for a future out-of-core pass.

### What looks good
- The helper fixes the actual failure mode: `hasattr(seg_data, 'get')` was false for dask-of-cupy, and `np.asarray(seg.values)` then hit cupy's implicit-conversion TypeError. Compute-then-get covers all four backends in one place, and `_optimize_waypoint_order` gets the same treatment so `optimize_order=True` works too.
- The no-op expression in `test_multi_stop_cupy_matches_numpy` (`x if not hasattr(x, 'get') else x`) is gone; the test now asserts the output is cupy-backed, which is what let the downgrade slip through before.
- New dask+cupy test executes on this host (CUDA available) and covers both the plain call and `optimize_order=True`; 52 tests pass.

### Checklist
- [x] Algorithm matches reference/paper (no algorithmic change; extraction and output typing only)
- [x] All implemented backends produce consistent results (numpy/cupy/dask+numpy/dask+cupy all tested)
- [x] Dask chunk boundaries handled correctly (output rechunked to the input's chunks)
- [x] No premature materialization or unnecessary copies beyond the pre-existing dense stitching (see nits)
- [x] Benchmark exists or is not needed (bug fix, no perf-critical change)
- [x] README feature matrix updated (n/a, no new function; dask+cupy was already claimed via a_star_search)
- [x] Docstrings present and accurate (see suggestion)

brendancol (commented):
## PR Review follow-up

The docstring suggestion from the first pass is addressed in f9894956: the Returns section now states the output is backed by the same array type as the input. The two nits document pre-existing dense-stitching behavior and stay as notes for a future out-of-core pass rather than changes in this PR. Tests re-run after the commit: 52 passed. No open findings.

brendancol (commented):
## PR Review: Fix multi_stop_search on cupy and dask+cupy backends

Re-reviewed after the branch was updated with `origin/main` (clean merge, no conflicts; pulls the libjxl=0.11.* CI pin from #3639 that was the only reason CI was red). Pathfinding suite runs green on this CUDA host: `52 passed` for `xrspatial/tests/test_pathfinding.py`, cupy and dask+cupy tests actually executed. Pushed as beeacae0.

None.

None.

- `pathfinding.py:1501` and `:1504` repeat `import cupy` in both branches. This matches the local-import style already used in `a_star_search`, so it is fine to leave as is.

### What looks good
- `_segment_to_numpy` (pathfinding.py:1287) is the right consolidation. It handles all four backends in one place: dask computes first, cupy (including the cupy blocks that fall out of a dask+cupy `.compute()`) goes through `.get()`, numpy passes through. Both call sites (the stitching loop at :1466 and `_optimize_waypoint_order` at :1336) now share it, which removes the earlier `seg.values` path that triggered the implicit-conversion crash.
- The output-type restoration at :1497-1505 mirrors what `a_star_search` already does: dask input wrapped back with `da.from_array` using the surface chunks, dask+cupy blocks pushed to cupy via `map_blocks(cupy.asarray)`, plain cupy input via `cupy.asarray`. The `is_dask_cupy(surface)` / `is_cupy_array(surface_data)` guards match the argument conventions used elsewhere in the file (:923-926).
- Test coverage closes the real gaps: the old `test_multi_stop_cupy_matches_numpy` had a no-op `x if not hasattr(x,'get') else x` that masked a numpy-backed result; it now asserts a cupy-backed output. `test_multi_stop_dask_cupy_matches_numpy` is new and exercises the previously-crashing backend, including the `optimize_order=True` path through `_optimize_waypoint_order`. The dask test now asserts the output stays dask-backed.

### Checklist
- [x] Algorithm unchanged; fix is backend plumbing only
- [x] All four backends produce consistent results (verified against numpy on a CUDA host)
- [x] NaN handling unchanged (finite-mask stitching untouched)
- [x] Edge cases: backend coverage tested; no new numerical edge cases introduced
- [x] Dask chunk boundaries: output rewrapped with the surface's own chunks
- [x] No premature materialization beyond what the segment loop already did
- [x] Benchmark not needed (bugfix, no new public function)
- [x] README feature matrix: not applicable (no new function)
- [x] Docstring updated to note the output is backed by the input's array type

PR #3638, opened 2026-07-03
Return all-NaN from HPA* fallback when the goal cannot be reached

### LINKED ISSUE #3631: HPA* fallback in a_star_search returns a partial cost trail when no path exists

brendancol:
Closes #3631

- `_hpa_star_search` now returns a fresh all-NaN surface when a segment cannot be refined at the largest retry radius, instead of the partially written path image. An unreachable goal no longer produces a finite cost trail that dead-ends mid-grid, so `np.isfinite(result).any()` is a reliable reachability check in the HPA* regime, same as in the plain and bounded searches.

Backend coverage: HPA* runs on the numpy path (and the cupy CPU-fallback path); the dask backends use the sparse A* and are unaffected.

Test plan:
- [x] New test: wall-split 200x200 grid where the coarse route crosses the wall but refinement cannot; asserts the output has no finite pixels (previously 89)
- [x] `pytest xrspatial/tests/test_pathfinding.py` (52 passed on a CUDA host)


brendancol (commented):
## PR Review: Return all-NaN from HPA* fallback when the goal cannot be reached

- none

- none

- [ ] `xrspatial/pathfinding.py:520` a refinement failure now discards segments that were already connected, so a caller gets no hint of how far the search got. That matches the bounded-search behavior codified in `test_search_radius_too_small` (all-NaN when the search bounds exclude the only route), and the partial trail was worse: it made `np.isfinite(result).any()` a false reachability signal. If partial progress ever matters, a warning would be the way to surface it, not finite pixels.
- [ ] The test drives `_hpa_star_search` directly rather than through `a_star_search` with a mocked `_available_memory_bytes` and a >1000-pixel manhattan distance. Direct invocation keeps the test fast and deterministic; the public-API route to this code path is already covered by `test_auto_radius_selection`.

### What looks good
- One-line semantic fix with the reasoning in a comment at the return site.
- The new test constructs the exact failure geometry (thin NaN wall that coarse blocks see as passable) and asserts zero finite pixels, which fails on the old code with 89.
- No behavior change for reachable goals: the escalation loop and stitching are untouched, and the existing `test_hpa_star_correctness` still passes.

### Checklist
- [x] Algorithm matches reference/paper (no-path contract now uniform across search paths)
- [x] All implemented backends produce consistent results (HPA* runs on the numpy path and cupy CPU-fallback; dask uses sparse A*, unaffected)
- [x] NaN handling is correct
- [x] Benchmark exists or is not needed (failure-path fix)
- [x] Docstrings present and accurate (internal function; public docstring already describes HPA* only at a high level)

brendancol (commented):
## PR Review: Return all-NaN from HPA* fallback when the goal cannot be reached

Re-review after the branch was updated with `main` (picks up the libjxl=0.11.* CI pin from #3639, which clears the earlier cog-validator/geotiff-corpus red). The merge was clean, no conflicts, and it did not touch `pathfinding.py`.

### Blockers
None.

### Suggestions
None.

### Nits
None.

### What looks good
- The fallback at `xrspatial/pathfinding.py:527` now returns `np.full((h, w), np.nan, dtype=np.float64)`, the same value and dtype as the existing no-coarse-path return at line 476 and the `path_img` allocation at line 492. The two "no path" exits are now identical, so `np.isfinite(result).any()` reads reachability the same way across the plain, bounded, and HPA* searches.
- Fallback only fires when the segment goal stays non-finite after the radius sweep (1, 2, 4, 8), i.e. the segment genuinely cannot be connected. A partial cost trail there is misleading, so dropping it is the right call.
- Backend scope checks out: HPA* is numpy-only (with the cupy CPU fallback); the dask paths go through sparse A* and never hit this branch. No cross-backend parity concern.
- `test_hpa_star_unreachable_goal_all_nan` builds a real unreachable case: a full NaN wall at column 100 with a factor-16 coarsening, so the coarse blocks straddling the wall stay passable and the coarse route crosses while refinement cannot. Asserting no finite pixels pins the exact regression (89 finite pixels before).

### Verification
- `PYTHONPATH=<worktree> python -m pytest xrspatial/tests/test_pathfinding.py -x -q` → 52 passed.
- Pushed as `a2e43fd4e8de44ac467c4b0c5c7bf34865ac62b7`.

### Checklist
- [x] Algorithm/contract matches the other search paths
- [x] NaN handling correct (all-NaN no-path result)
- [x] Edge case (unreachable goal) covered by a new test
- [x] Dask paths unaffected (numpy-only branch)
- [x] Benchmark not needed (edge-case return fix)
- [x] Docstrings: private helper, no public API change

PR #3643, opened 2026-07-07
Fix _parse_rat to pad category_names to pixel indices for sparse RATs

### LINKED ISSUE #3591: read_pam_sidecar misaligns category_names when the RAT Value column is sparse or not zero-based
`_pam._parse_rat` sorts thematic RAT rows by their Value column and returns the names densely packed. The attrs contract in `_attrs.py` says `category_names` is an "ordered list of class label strings (index == pixel value)". Any sidecar whose Value column isn't exactly 0..N-1 comes back misaligned.

Repro:

```python
from xrspatial.geotiff._pam import _parse_rat
from xrspatial.geotiff._safe_xml import safe_fromstring

xml = """<PAMDataset><PAMRasterBand band="1">
  <GDALRasterAttributeTable tableType="thematic">
    <FieldDefn index="0"><Name>Value</Name><Type>0</Type><Usage>5</Usage></FieldDefn>
    <FieldDefn index="1"><Name>Class</Name><Type>2</Type><Usage>2</Usage></FieldDefn>
    <Row index="0"><F>1</F><F>water</F></Row>
    <Row index="1"><F>2</F><F>forest</F></Row>
    <Row index="2"><F>5</F><F>urban</F></Row>
  </GDALRasterAttributeTable>
</PAMRasterBand></PAMDataset>"""
names, colors = _parse_rat(safe_fromstring(xml).find('.//GDALRasterAttributeTable'))
print(names)  # ['water', 'forest', 'urban']
```

Under index==value, pixel 1 now reads as "forest" (should be "water") and pixel 5 has no entry. `category_colors` shifts the same way, so QGIS colors land on the wrong classes.

Our own writer (`build_pam_xml`) always emits a dense 0-based Value column, so xrspatial-to-xrspatial round trips are fine. This only bites when the `.aux.xml` was authored by GDAL or QGIS, and those commonly write 1-based or sparse Value columns (classified rasters where 0 is nodata, NLCD's 11/21/22/...).

Possible fix: pad gaps with empty strings (matching GDAL's CategoryNames semantics) and (0,0,0,0) colors. Should also fail closed on negative values or a huge max value so a weird RAT can't allocate an enormous list.

Found during the 2026-07-01 accuracy sweep. Filing rather than fixing since this is label metadata, not pixel values.

Melissari1997:
Closes #3591.

When a PAM sidecar's thematic RAT has sparse or non-zero-based values (e.g. `[1, 2, 5]` from QGIS/GDAL classified rasters), `_parse_rat` was sorting rows and returning a dense list, misaligning category names against their pixel values. The attrs contract promises index == pixel value.

- Pad `category_names` with empty strings and `category_colors` with `(0,0,0,0)` for gaps between sparse RAT rows so list index == pixel value.
- Fail closed on negative pixel values or a max value exceeding 1,000,000 to prevent OOM from pathological RATs.

Backend coverage: backend-agnostic — runs during `open_geotiff` metadata attach on all backends (numpy, dask, cupy, dask+cupy).

## Test plan

- [x] Existing dense 0-based RAT tests pass unchanged.
- [x] Sparse 1-based RAT produces padded, index-aligned names.
- [x] Sparse RAT with RGBA columns pads both names and colors.
- [x] Negative values fail closed (return empty dict from `read_pam_sidecar`).
- [x] Values >= 1M fail closed.
- [x] No Value column uses Row index attributes and pads.
- [x] `test_field_defn_without_usage_is_skipped` still passes.

### DIFF
--- a/xrspatial/geotiff/_pam.py
+++ b/xrspatial/geotiff/_pam.py
@@ -191,13 +191,21 @@ def read_pam_sidecar(path):
         return {}
 
 
+_MAX_CATEGORIES = 1_000_000
+
+
 def _parse_rat(rat):
     """Return (names, colors) from a thematic ``<GDALRasterAttributeTable>``.
 
     Returns ``(None, None)`` when the table has no Name column, i.e. it does
-    not describe categories. ``names`` is ordered by the row's Value column;
-    ``colors`` is the RGBA list when the RAT defines color columns, else
-    ``None``.
+    not describe categories. ``names`` is index-aligned to pixel values
+    (index == pixel value), padded with empty strings for gaps between
+    sparse RAT rows. ``colors`` is the RGBA list when the RAT defines color
+    columns, else ``None``; gaps are padded with ``(0, 0, 0, 0)``.
+
+    Fails closed (returns ``(None, None)``) on negative pixel values or a
+    maximum value exceeding ``_MAX_CATEGORIES``, which would otherwise
+    allocate a list large enough to OOM.
     """
     usage_to_col = {}
     for fd in rat.findall('FieldDefn'):
@@ -230,7 +238,19 @@ def _parse_rat(rat):
     if not rows:
         return None, None
 
-    rows.sort(key=lambda r: r[0])
-    names = [name for _, name, _ in rows]
-    colors = [color for _, _, color in rows] if have_colors else None
+    max_val = max(r[0] for r in rows)
+    min_val = min(r[0] for r in rows)
+    if min_val < 0 or max_val >= _MAX_CATEGORIES:
+        return None, None
+
+    names = [''] * (max_val + 1)
+    if have_colors:
+        colors = [(0, 0, 0, 0)] * (max_val + 1)
+    else:
+        colors = None
+    for value, name, color in rows:
+        names[value] = name
+        if have_colors:
+            colors[value] = color
+
     return names, colors
--- a/xrspatial/geotiff/_pam.py
+++ b/xrspatial/geotiff/_pam.py
@@ -238,8 +238,13 @@ def _parse_rat(rat):
     if not rows:
         return None, None
 
-    max_val = max(r[0] for r in rows)
-    min_val = min(r[0] for r in rows)
+    min_val = max_val = rows[0][0]
+    for r in rows:
+        v = r[0]
+        if v < min_val:
+            min_val = v
+        if v > max_val:
+            max_val = v
     if min_val < 0 or max_val >= _MAX_CATEGORIES:
         return None, None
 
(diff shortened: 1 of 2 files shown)


Melissari1997 (commented):
## PR Review: Fix _parse_rat to pad category_names to pixel indices for sparse RATs

_None._

- [ ] **`_pam.py:241-242`** -- `max(r[0] for r in rows)` / `min(r[0] for r in rows)` iterate `rows` twice. `rows` is typically small (a handful of RAT classes), so this is negligible in practice, but computing both in a single pass over `rows` (e.g. manual loop or tee) would be trivially cleaner. Not blocking.

- [ ] **`_pam.py:194`** -- `_MAX_CATEGORIES = 1_000_000` is module-level but `_parse_rat` is the only consumer. If you prefer keeping the guard close to the code it protects, it could be a local constant inside `_parse_rat`. Style preference, no functional difference.

### What looks good

- The fix correctly aligns the output with the documented `category_names` contract ("index == pixel value").
- Fails closed on the two known OOM hazards (negative values, pathological max) rather than silently allocating.
- The `_MAX_CATEGORIES` guard (1M) gives ample headroom for realistic classified rasters while preventing genuine OOM.
- Six well-structured tests cover: the exact repro from #3591, color padding, negative guard, max guard, no-Value-column path with sparse Row indices, and regression on dense 0-based RATs.
- All 10 tests in the file pass, including the 4 pre-existing reader-edge tests.
- The change is surgical -- only `_parse_rat` is modified; `build_pam_xml`, `read_pam_sidecar`, and callers are untouched.
- Backend coverage is correctly described as backend-agnostic (runs during `open_geotiff` metadata attach on all backends).

### Checklist

- [x] Algorithm matches reference/paper -- N/A (metadata parser fix, not a numerical algorithm)
- [x] All implemented backends produce consistent results -- N/A (backend-agnostic metadata reader)
- [x] NaN handling is correct -- N/A (no numerical computation)
- [x] Edge cases are covered by tests -- negative values, huge values, sparse values, no Value column, dense 0-based regression all tested
- [x] Dask chunk boundaries handled correctly -- N/A
- [x] No premature materialization or unnecessary copies -- no `.values`/`.compute()` calls
- [x] Benchmark exists or is not needed -- not needed for a metadata parser
- [x] README feature matrix updated -- N/A (no new public API)
- [x] Docstrings present and accurate -- `_parse_rat` docstring updated to reflect new padding behavior

Melissari1997 (commented):
## Follow-up review: disposition of findings

### Suggestion: single-pass min/max (`_pam.py:241-242`)
- **Fixed** in 53d4a2f. Replaced `max()`/`min()` with a single `for r in rows:` loop computing both in one pass.

### Nit: `_MAX_CATEGORIES` module-level vs function-local (`_pam.py:194`)
- **Dismissed.** Module-level constants are the established convention in this file: `_USAGE_MINMAX`, `_USAGE_NAME`, `_USAGE_RED`, `_USAGE_GREEN`, `_USAGE_BLUE`, `_USAGE_ALPHA`, `_TYPE_INT`, `_TYPE_STRING` are all at module level (lines 24-32). Keeping `_MAX_CATEGORIES` alongside them is consistent with existing style.

All 10 tests continue to pass.

brendancol:
@Melissari1997 wooohooo

brendancol (approved):
LGTM

PR #3648, opened 2026-07-08
Extend pathfinding asv benchmarks to dask/cupy backends and multi_stop_search

### LINKED ISSUE #3645: pathfinding asv benchmark covers only numpy a_star_search; dask/cupy backends and multi_stop_search unbenchmarked
## Problem

`benchmarks/benchmarks/pathfinding.py` measures `a_star_search` on the numpy backend only, at grids up to 300x150. The module has four backend paths and a second public function, and most of that surface has no benchmark:

| Function | numpy | cupy | dask | dask+cupy |
|---|---|---|---|---|
| `a_star_search` | nx=10/100/300, conn 4/8 | none | none | none |
| `multi_stop_search` | none | none | none | none |

Gaps:

1. Backend gap (HIGH). The dask backend is not the numpy kernel behind a wrapper: `_a_star_dask` is a separate sparse Python A* with an LRU chunk cache, and `_path_to_dask_array` builds the lazy output block by block. A regression there (say, the chunk cache silently reloading blocks) would not show up in any benchmark. The cupy path (device-to-host fallback around the numba kernel) is also unmeasured.
2. Missing benchmark (HIGH). `multi_stop_search` routes N waypoints through repeated A* calls, stitches the segments, and can reorder waypoints via Held-Karp / 2-opt. None of it is benchmarked, directly or indirectly.
3. Small inputs (MEDIUM). The largest input is 300x150 (~45k cells, below 256x256), and nx=10 mostly measures setup overhead.
4. Degenerate input (LOW, note only). The input grid has no barriers and no friction, which is the best case for A*. Not fixing here, but worth remembering when reading absolute numbers.

## Evidence (executed on this host, CUDA available)

The current benchmark imports and runs (so this is a coverage gap, not a broken benchmark). Candidate timings for the missing paths, per call after warmup:

```
dask  a_star nx=100   0.039 s
dask  a_star nx=300   0.304 s
dask  a_star nx=1000  3.857 s
cupy  a_star nx=100   0.0017 s
cupy  a_star nx=300   0.0045 s
cupy  a_star nx=1000  0.0486 s
numpy a_star nx=1000  0.0462 s
numpy multi_stop nx=300 ordered    0.0089 s
dask  multi_stop nx=300 ordered    0.2407 s
cupy  multi_stop nx=300 ordered    0.0131 s
numpy multi_stop nx=100 optimize   0.0249 s
dask  multi_stop nx=100 optimize   0.2514 s
```

## Proposed fix (benchmark-only)

Extend `benchmarks/benchmarks/pathfinding.py`:

- `AStarSearch`: parameterize `type` over numpy / cupy / dask, sizes `[100, 300, 1000]` (dask capped at 300 via `NotImplementedError` in setup; the pure-Python dask A* takes ~4 s at nx=1000 and would dominate suite runtime). Drop nx=10. `snap_start`/`snap_goal` stay on for numpy/cupy but off for dask, where they raise by design.
- New `MultiStopSearch` class: 4 waypoints on numpy / cupy / dask at `[100, 300]`, timing both the ordered path and `optimize_order=True`.

dask+cupy stays uncovered: the suite's `common.get_xr_dataarray` has no dask+cupy type, and no benchmark in the suite parameterizes it. Noting it here rather than growing shared benchmark infra in this change.

No source changes to `xrspatial/pathfinding.py`.

brendancol:
Closes #3645

- Parameterize `AStarSearch` over numpy / cupy / dask and raise the size ladder to `[100, 300, 1000]` (was numpy-only at `[10, 100, 300]`). The dask backend is a separate implementation (`_a_star_dask`, sparse Python A* with an LRU chunk cache) that previously had no benchmark at all.
- Skip dask at nx=1000 via `NotImplementedError` in setup: a single call takes ~4 s and would dominate suite runtime. dask still runs chunked (2x2 chunks) at 100 and 300.
- Add `MultiStopSearch`: 4 waypoints across the grid, timing both ordered routing and `optimize_order=True` (Held-Karp path), on numpy / cupy / dask at `[100, 300]`.
- `snap_start`/`snap_goal` stay on for numpy/cupy (matching the old benchmark) and off for dask, where they raise by design.

Benchmark-only change; `xrspatial/pathfinding.py` is untouched. dask+cupy is still not parameterized because `common.get_xr_dataarray` has no such type anywhere in the suite; noted in #3645.

Backend coverage: numpy, cupy, dask+numpy benchmarked; dask+cupy not (suite-wide gap).

Test plan:
- [x] Ran every parameter combination directly (28 combos, 2 intentional dask skips) on this host, including cupy on GPU; timings in #3645. Slowest single call is 1.4 s (dask `optimize_order` at nx=300).
- [x] Existing benchmark import path unchanged (`benchmarks.pathfinding`), flake8 and isort clean.


brendancol (commented):
Domain-aware review of the benchmark changes.

**Blockers**

None. The change is benchmark-only, `xrspatial/pathfinding.py` is untouched, and every runnable parameter combination was executed on a GPU host per the PR body.

**Suggestions**

None.

**Nits**

1. Changing `AStarSearch.params` (new sizes and backend axis) resets the asv history for this class, so old numpy timings won't line up with new ones in the dashboard. Unavoidable when adding a backend axis, just something to know when reading trends across this commit.
2. `MultiStopSearch.setup` builds waypoints from `self.agg.y.data[...]` numpy scalars while `AStarSearch` uses `self.agg.y[0]` DataArray scalars. Both are accepted by the API (`_get_pixel_id` floats them), so this is style only. The numpy-scalar form is arguably better in setup code since it skips xarray indexing overhead.

Details checked:

- The dask skip at nx=1000 uses `NotImplementedError` in `setup`, the same skip mechanism `common.get_xr_dataarray` already uses for missing cupy/rtxpy, and asv marks those combos n/a rather than failed.
- `snap_start`/`snap_goal` stay on for numpy/cupy, matching the old benchmark's call signature; they are off for dask where the API raises `ValueError` by design. Snap on a valid start pixel is O(1) (`_find_nearest_pixel` returns immediately), so it doesn't distort timings.
- `get_xr_dataarray` dask chunks are `(ny//2, nx//2)`, so the dask combos run 2x2 chunked and exercise `_a_star_dask`'s chunk cache for real, not a single-chunk degenerate case.
- With 4 waypoints, `optimize_order=True` takes the Held-Karp branch (N <= 12) and issues 12 matrix A* calls plus 3 segment calls, which is the cost profile users of that flag actually hit.
- Slowest single call is ~1.4 s (dask optimize_order at nx=300), so the added suite runtime is bounded.

PR #3654, opened 2026-07-08
Raise on infeasible tours and honor snap in multi_stop_search optimize_order

### LINKED ISSUE #3646: multi_stop_search(optimize_order=True) silently drops waypoints when the distance matrix has no finite tour
**Describe the bug**

`multi_stop_search(optimize_order=True)` can return a finite path that silently skips waypoints instead of raising `ValueError`. Two holes in `_optimize_waypoint_order` / `_held_karp` (xrspatial/pathfinding.py) combine here:

1. When no finite tour through all waypoints exists, every `dp[full][j] + dist[mid[j]][end]` in `_held_karp` is `inf`, so `best_last` stays `-1` and the reconstruction loop exits immediately. The function returns `[start] + [] + [end]`, dropping every interior waypoint. `multi_stop_search` then routes start -> end directly and returns a normal-looking result whose `waypoint_order` attr has fewer entries than the input. With `optimize_order=False` the same input raises `no path between waypoints 0 and 1`.

2. With `snap=True`, the distance matrix reads each segment cost at the *unsnapped* goal pixel (`_get_pixel_id(waypoints[j], ...)`). If waypoint `j` sits on a NaN/barrier cell, `a_star_search` snaps the goal and finds a path, but the unsnapped pixel is NaN in the segment image, so `dist[i][j]` stays `inf` for every `i`. That feeds hole 1: a waypoint that is fully reachable after snapping gets dropped. The main segment loop already handles this case (it relocates the goal to the max-finite-cost pixel); the ordering pass doesn't.

The non-exact path (`_nearest_neighbor_2opt`, N > 12) is not affected by the dropping: it always returns a complete tour, and the segment loop raises later if a leg is unreachable.

**Reproduction** (current main, numpy backend)

```python
import numpy as np
import xarray as xr

def make_raster(data):
    h, w = data.shape
    r = xr.DataArray(data.astype(np.float64), dims=['y', 'x'],
                     attrs={'res': (1.0, 1.0)})
    r['y'] = np.linspace(h - 1, 0, h)
    r['x'] = np.linspace(0, w - 1, w)
    return r

# Case 1: unreachable interior waypoint
data = np.ones((8, 8))
data[4, :] = np.nan                      # wall
agg = make_raster(data)
wps = [(7.0, 0.0), (0.0, 0.0), (7.0, 7.0)]   # middle one is below the wall

# multi_stop_search(agg, wps)  -> ValueError: no path between waypoints 0 and 1
r = multi_stop_search(agg, wps, optimize_order=True)
print(r.attrs['waypoint_order'])   # [(7.0, 0.0), (7.0, 7.0)]  <- middle dropped
print(r.attrs['total_cost'])       # 7.0 (finite, looks fine)

# Case 2: snap=True with a reachable (snappable) waypoint
data2 = np.ones((8, 8))
data2[3, 3] = np.nan
agg2 = make_raster(data2)
wps2 = [(7.0, 0.0), (4.0, 3.0), (0.0, 7.0)]  # middle sits on the NaN cell

r0 = multi_stop_search(agg2, wps2, snap=True)
print(len(r0.attrs['waypoint_order']))   # 3 (correct)
r1 = multi_stop_search(agg2, wps2, snap=True, optimize_order=True)
print(len(r1.attrs['waypoint_order']))   # 2  <- reachable waypoint dropped
```

Observed output:

```
[(7.0, 0.0), (7.0, 7.0)]
7.0
3
2
```

**Expected behavior**

- Case 1: raise `ValueError`, matching the `optimize_order=False` contract documented in the docstring ("Raises ValueError ... or a segment is unreachable").
- Case 2: keep all three waypoints; the ordering pass should use the same snapped-goal cost lookup as the segment loop.

**Additional context**

Found by the accuracy sweep. `waypoint_order`, `segment_costs`, and `total_cost` all report the truncated route, so downstream code has no signal that stops were skipped.

brendancol:
Closes #3646

- `_optimize_waypoint_order` now raises `ValueError` when Held-Karp's total cost is infinite. Previously `best_last` stayed `-1`, the reconstruction returned `[start, end]`, and `multi_stop_search(optimize_order=True)` silently dropped every interior waypoint while returning a finite route. The `optimize_order=False` path already raises on the same input.
- The pairwise distance matrix now reads each segment's cost at the true (snapped) goal pixel when `snap=True`, using the same max-finite-cost lookup as the segment loop. Before, a waypoint on an invalid cell got `inf` distance in every row even though snapping made it reachable, which fed the hole above.
- The heuristic solver (`_nearest_neighbor_2opt`, N > 12) is untouched: it always returns a complete tour, and an infinite heuristic cost is not proof that no feasible tour exists, so the segment loop keeps raising `no path between waypoints` for that case.

Backends: the fix is in the ordering pass shared by all four backends (numpy, cupy, dask+numpy, dask+cupy); no backend-specific code changed.

Test plan:
- [x] `test_optimize_order_unreachable_waypoint_raises` (fails on main: returns a finite 2-waypoint route)
- [x] `test_optimize_order_with_snap_keeps_waypoints` (fails on main: drops the snapped waypoint)
- [x] `pytest xrspatial/tests/test_pathfinding.py` — 59 passed, including cupy and dask+cupy tests on a CUDA box
- [x] flake8 clean on changed lines

Also includes this sweep's state-CSV update.


brendancol (commented):
Domain-aware review of the optimize_order fix.

**Blockers**

None.

**Suggestions**

- The snap-relocation block (max-finite-cost pixel as the true goal) now exists twice: here and in the segment loop of `multi_stop_search`. Both copies are four lines and the loop variant also persists `waypoint_pixels`, so a shared helper is borderline; if a third caller shows up, extract one.
- The raise is gated on the exact solver only. For N > 12 the heuristic tour can carry infinite legs and the failure surfaces later as `no path between waypoints i and i+1` from the segment loop. That asymmetry is deliberate (an infinite heuristic cost does not prove infeasibility) and the PR body says so; worth keeping in mind if anyone later tightens the heuristic.

**Nits**

- `np.unravel_index` returns numpy scalars for `goal_py, goal_px`; the subsequent `dist[i][j]` assignment and comparisons handle them fine, matching what the segment loop already does.

Verified locally: both new tests fail on main for the stated reasons (finite 2-waypoint route; dropped snapped waypoint) and pass with the fix; the full pathfinding suite (59) passes including cupy and dask+cupy on a CUDA box. The Held-Karp reconstruction path for feasible tours is untouched, and `test_optimize_order_finds_better_route` / `test_optimize_order_preserves_endpoints` still pass.

PR #3656, opened 2026-07-08
Reset dask graph-token .name on pathfinding outputs (#3652)

### LINKED ISSUE #3652: pathfinding: dask-backed outputs adopt the dask graph token as DataArray.name
`a_star_search` and `multi_stop_search` return unnamed DataArrays on the numpy and cupy backends, but on dask and dask+cupy the result's `.name` is the internal dask graph token. `xr.DataArray(...)` falls back to `getattr(data, 'name', None)` when no name is given, and dask arrays carry their task-key prefix there.

Observed on main (2026-07-08):

```python
import dask.array as da
from xrspatial.pathfinding import multi_stop_search

agg = xr.DataArray(np.ones((8, 8)), dims=['y', 'x'], attrs={'res': (1.0, 1.0)})
agg['y'] = np.linspace(7, 0, 8); agg['x'] = np.linspace(0, 7, 8)
agg_dask = agg.copy(); agg_dask.data = da.from_array(agg.data, chunks=(4, 4))

a_star_search(agg, (7.0, 0.0), (0.0, 7.0)).name        # None
a_star_search(agg_dask, (7.0, 0.0), (0.0, 7.0)).name   # 'concatenate-bf94f35e9f109e8253ed25eaf6425599'
multi_stop_search(agg_dask, [(7.0, 0.0), (0.0, 7.0)]).name  # 'array-16f7995fd6d22ee4b08829515ec36e6b'
```

cupy behaves like numpy (returns None); dask+cupy leaks the token like dask ('asarray-<hash>' for multi_stop_search).

The hash ends up anywhere the variable name matters: `to_dataset()`, netcdf output, plot titles. And it changes on every call, since the token is content-addressed.

This bug class was already fixed in zonal (#2611) and focal (#2733); slope carries the post-construction reset with a comment pointing at those. pathfinding never got the fix.

Fix: reset `.name = None` after constructing the output DataArray in both functions, so all four backends match, and pin it with a cross-backend test.

Found by the metadata propagation sweep. Not the same issue as #3644, which covers the attrs drop in multi_stop_search.

brendancol:
Closes #3652

On dask and dask+cupy backends, `a_star_search` and `multi_stop_search` returned DataArrays named after the internal dask graph token ('concatenate-<hash>', 'array-<hash>') while numpy/cupy returned unnamed results. `xr.DataArray` picks up `data.name` when no name is passed, and dask arrays expose their task key there.

- Reset `.name = None` post-construction in both functions (same fix as zonal #2611 and focal #2733; slope has the same reset).
- Add `test_output_name_consistent_across_backends`, parametrized over numpy / dask+numpy / cupy / dask+cupy (GPU cases skip when CUDA is absent).

All four backends verified locally (CUDA available on this box).

Test plan:
- [x] `pytest xrspatial/tests/test_pathfinding.py` (61 passed)
- [x] probe script confirms `.name is None` on all four backends for both functions


brendancol (commented):
## PR Review: Reset dask graph-token .name on pathfinding outputs (#3652)

None.

None.

- [ ] `xrspatial/tests/test_pathfinding.py:1150` — `_name_check_backends()` overlaps with the module-level `_backends` list defined at line 273. Reasonable as-is, since `_backends` deliberately excludes the GPU backends (those tests use the `@cuda_and_cupy_available` decorator instead), but a shared "all runnable backends" helper in `general_checks.py` would serve future cross-backend metadata tests too.

### What looks good
- Both public functions have exactly one DataArray construction site, so the post-construction reset covers every backend path (numpy, cupy, dask, dask+cupy, bounded, HPA*).
- Post-construction assignment is the correct mechanism here: passing `name=None` to the constructor does not help because xarray falls back to the dask array's `.name` attribute. Same approach as slope, zonal (#2611), focal (#2733), and proximity (#2728).
- `@supports_dataset` on `multi_stop_search` re-keys per-variable results in a dict, so the `None` name cannot leak into Dataset variable names.
- The regression test asserts the actual property (`.name is None`) for both functions and parametrizes over all four backends with runtime gating, so it degrades to numpy+dask on CPU-only hosts instead of skipping wholesale.

### Checklist
- Algorithm matches reference: n/a (metadata-only change)
- All implemented backends produce consistent results: yes, verified by the new test on a CUDA host
- NaN handling: untouched
- Edge cases covered: single construction site per function, so no uncovered paths
- Dask chunk boundaries: untouched
- No premature materialization: attribute assignment only
- Benchmark: not needed
- README feature matrix: no API change
- Docstrings: no signature change; existing docstrings make no claims about `.name`

PR #3659, opened 2026-07-08
Route the HPA* coarse grid without value barriers

### LINKED ISSUE #3647: HPA* coarse routing falsely blocks passable blocks whose mean equals a barrier value
**Describe the bug**

`_hpa_star_search` (xrspatial/pathfinding.py) routes the coarsened grid with the caller's `barriers` list. Coarse cell values are block *means* of the non-barrier fine cells (`_coarsen_surface`), so a passable block whose values average to exactly a barrier value is treated as impassable at the coarse level. If those blocks separate start from goal on the coarse grid, the coarse route fails and `a_star_search` returns all-NaN ("no path") for a grid where the full-resolution search finds a path.

`_coarsen_surface` already encodes impassability: a block with no crossable fine cells stays NaN. Re-applying the exact-value barrier test to averaged data is what creates the false blocks.

This is reachable from the public API: `a_star_search` uses HPA* when `search_radius=None`, the grid exceeds 50% of available RAM, and the start/goal manhattan distance is over 1000 pixels.

**Reproduction** (current main)

```python
import numpy as np
from xrspatial.pathfinding import (
    _hpa_star_search, _a_star_search, _neighborhood_structure,
)

H = W = 200
data = np.full((H, W), 1.0)
data[:, ::2] = -1.0            # alternating -1/+1 columns; every 16x16 block mean == 0.0
barriers = np.array([0.0])     # no fine cell equals 0.0

dy, dx, dd = _neighborhood_structure(1.0, 1.0, 8)
path = _hpa_star_search(data, None, 0, 0, H - 1, W - 1,
                        barriers, dy, dx, dd, 1.0, False, 1.0, 1.0, H, W)
print(np.isfinite(path).sum())          # 0  <- "no path"

ctrl = np.full((H, W), np.nan)
_a_star_search(data, ctrl, 0, 0, H - 1, W - 1, barriers, dy, dx, dd,
               np.ones((1, 1)), 1.0, False, 1.0, 1.0)
print(np.isfinite(ctrl[H - 1, W - 1]))  # True <- full-resolution A* finds it
```

End-to-end through `a_star_search` (auto-radius + HPA* path), same grid at 600x600 with `_available_memory_bytes` patched to 40 MB and `barriers=[0]`: 0 finite pixels; the same call without the memory cap reaches the goal.

**Expected behavior**

HPA* finds a path on this grid, like the full-resolution search. The coarse route should be computed with an empty barrier list and rely on NaN coarse cells to mark impassable blocks.

**Additional context**

Found by the accuracy sweep. Needs a value distribution whose block means hit a barrier value exactly, so it mostly bites categorical-style rasters with signed values and a 0 barrier.

brendancol:
Closes #3647

- `_hpa_star_search` now routes the coarsened grid with an empty barrier list. Coarse cell values are block means of the non-barrier fine cells, so testing them against the exact barrier values could falsely block a passable block whose values average to a barrier value (alternating -1/+1 data with `barriers=[0]` makes every block mean exactly 0.0, blocking the whole coarse grid and returning all-NaN "no path" on a fully passable surface).
- `_coarsen_surface` already marks fully-barrier blocks as NaN, and NaN cells stay impassable in the coarse search, so no blocking information is lost. Refinement still runs on the fine grid with the caller's barriers.

Backends: HPA* runs on the numpy and cupy (CPU-fallback) paths only; the dask sparse A* never coarsens.

Test plan:
- [x] `test_hpa_star_barrier_valued_block_means` (fails on main: 0 finite pixels)
- [x] Re-ran the issue's end-to-end repro (600x600, mocked RAM, auto-radius HPA*): goal reached after the fix
- [x] `pytest xrspatial/tests/test_pathfinding.py` — 58 passed, including the existing HPA* correctness and all-NaN no-path tests
- [x] flake8 clean on changed lines

Also includes this sweep's state-CSV update.


brendancol (commented):
Domain-aware review of the HPA* coarse-barrier fix.

**Blockers**

None.

**Suggestions**

- The empty array reuses `barriers.dtype`, which keeps the numba signature of `_a_star_search` stable across the coarse and fine calls when callers pass integer barrier lists. Good; a bare `np.array([])` would have forced a second float64 specialization but also worked. No change needed.

**Nits**

- The fix only touches the coarse route. Refinement (`_bounded_a_star_sub`) still runs on the fine grid with the caller's barriers, so true barriers keep constraining the actual path; the change removes only the mean-aliasing false blocks. The pre-existing semantics that a block with any single crossable cell is coarse-passable are unchanged, and the radius-escalation loop plus the #3638 all-NaN contract still cover refinement dead-ends.

Verified locally: the new test returns 0 finite pixels on main and a connected start-to-goal path with the fix; the issue's end-to-end repro (600x600, mocked RAM so auto-radius HPA* engages, barriers=[0]) reaches the goal after the fix; `test_hpa_star_correctness` and `test_hpa_star_unreachable_goal_all_nan` still pass (58 total).

Issue #3655, opened 2026-07-08
Add golden-value reference tests for a_star_search (validated against scipy dijkstra and skimage MCP_Geometric)


brendancol:
## Summary

`a_star_search` has no test that pins its results against an independent reference implementation. The existing suite checks backend parity, an internal cross-check against `cost_distance`, and one hand-computed 3x3 case, but nothing compares the algorithm with an external tool. A change that broke the cost model (edge weights, heuristic admissibility, cellsize scaling) could pass the parity tests while giving users different answers than scipy or scikit-image on the same data.

## Validation performed

A reference-validation sweep (2026-07-08) compared `a_star_search` against:

- `scipy.sparse.csgraph.dijkstra` (1.16.1) on an explicitly built grid graph with the same edge model (`geometric_distance`, or `geometric_distance * mean(endpoint_friction)` with a friction surface)
- `skimage.graph.MCP_Geometric` (0.26.0), whose cost model is identical to the friction model
- analytic ground truth on open uniform grids

Inputs: 64x64 open grid (30 m cells), seeded barrier maze, seeded random friction surface, friction with NaN holes, anisotropic cells (cx=30, cy=10), both 4- and 8-connectivity. Every case matched to machine precision (max rel diff about 2e-15), path chains were adjacency- and cost-increment-consistent, and all four backends (numpy, dask+numpy, cupy, dask+cupy) agreed on this host.

## Proposal

Add a golden-value regression test (test-only, no source changes) that hardcodes goal costs on small inputs:

- 6x6 open grid, 8- and 4-connectivity (also matches the closed-form value)
- 8x8 seeded barrier maze, including a start/goal pair with no path
- 6x6 random friction surface
- anisotropic cellsize case

Expected values hardcoded from the verified comparison above, with a comment naming the reference tools and versions. Runs on numpy and dask+numpy in CI; the existing cupy parity tests carry the result to GPU backends.

## Comparison script

<details>
<summary>refval_pathfinding_compare.py (executed 2026-07-08, all checks passed)</summary>

```python
"""Reference validation for xrspatial.pathfinding.a_star_search.

References:
  - scipy.sparse.csgraph.dijkstra on an explicitly-built grid graph
    (exact optimal least-cost, same edge model)
  - skimage.graph.MCP_Geometric (edge cost = distance * mean endpoint cost,
    identical to xrspatial's friction cost model; sampling= handles cellsize)
  - analytic ground truth on an open uniform grid
"""
import sys
[...]

PR #3664, opened 2026-07-08
Validate barriers, search_radius, start/goal points, and dims in pathfinding

### LINKED ISSUE #3649: pathfinding: string barriers silently ignored, search_radius unvalidated, inconsistent start/goal and dims errors
**Describe the bug**

Error-handling audit of `xrspatial/pathfinding.py`, with the bad-input battery executed on numpy, dask+numpy, cupy, and dask+cupy. Four validation gaps in `a_star_search` / `multi_stop_search`.

**1. String-typed `barriers` values are silently ignored (severity: critical)**

Numba compares the float cell value against a unicode element, which is always False, so every barrier is dropped. No error, no warning, and the returned path crosses the cells the caller asked to block:

```python

data = np.ones((5, 9))
data[:, 4] = 0.0                      # wall of zeros
r = xr.DataArray(data, dims=('y', 'x'), attrs={'res': (1.0, 1.0)})
r['y'] = np.linspace(4, 0, 5); r['x'] = np.linspace(0, 8, 9)

a_star_search(r, (2.0, 0.0), (2.0, 8.0), barriers=[0])    # all-NaN: wall blocks the path (correct)
a_star_search(r, (2.0, 0.0), (2.0, 8.0), barriers=['0'])  # routes straight through the wall
```

Barrier values that arrive as strings (argparse, CSV, JSON config) give a wrong path that looks valid. A scalar (`barriers=0` instead of `[0]`) dies instead with a numba `TypingError: iteration over a 0-d array` on the numpy/cupy backends and a bare `TypeError` on dask. Same on all four backends.

**2. `search_radius` is never validated (severity: high)**

- `a_star_search(agg, start, goal, search_radius=-1)` returns all-NaN ("no path") on numpy and cupy even though a path exists. Through `multi_stop_search` the same input surfaces as the misleading `ValueError: no path between waypoints 0 and 1`.
- Other geometries crash deep in the slicing code instead: `ValueError: negative dimensions are not allowed` (radius -3, interior points) or `TypeError: slice indices must be integers or None or have an __index__ method` (radius 2.5). Whether a float radius crashes or works depends on where start/goal sit relative to the grid edges.

**3. `multi_stop_search` with mismatched dim names raises bare `KeyError: 'y'` (severity: medium)**

`a_star_search` checks `surface.dims != (y, x)` and raises `ValueError`; `multi_stop_search` skips that check, so the same mistake dies inside xarray internals:

```python
multi_stop_search(raster_with_lat_lon_dims, [start, goal])
# KeyError: 'y'   (from xarray's dataset_utils, no mention of the x=/y= parameters)
```

The `a_star_search` message has its own problems: `` `surface.coords` should be named as coordinates:(y, x)`` says "coords" where it means dims, is missing a space, and never shows the dims the raster actually has.

**4. Scalar `start`/`goal` in `a_star_search` (severity: medium)**

`a_star_search(agg, 3.0, goal)` raises `TypeError: 'float' object is not subscriptable` from inside `_get_pixel_id`. `multi_stop_search` already checks each waypoint has exactly 2 elements; `a_star_search` accepts anything, and a scalar waypoint in `multi_stop_search` itself hits `TypeError: object of type 'float' has no len()` at the check site.

**Expected behavior**

- Non-numeric or non-1-D `barriers` raise a clear error naming the parameter and the offending dtype, before any kernel runs.
- `search_radius` other than None or a non-negative integer is rejected up front.
- `multi_stop_search` raises the same dims `ValueError` as `a_star_search`, and the message names the actual dims plus the `x=`/`y=` parameters.
- `start`/`goal` get the same 2-element check that `multi_stop_search` applies to waypoints.

Found by the error-handling sweep on 2026-07-08. Exception types stay as they are for currently-raising paths; the two silent cases (1 and 2) start raising, which only affects calls that today return wrong output.

brendancol:
Closes #3649

- Reject non-numeric and non-1-D `barriers` before any kernel runs. `barriers=['0']` used to silently disable every barrier (numba compares float cells to unicode, always False) and route straight through walls; `barriers=0` died with a numba TypingError. Numeric values are now coerced to float64.
- Reject `search_radius` values that are not None or a non-negative integer. Negative radii used to return a silent all-NaN "no path" on numpy/cupy (a misleading "no path between waypoints" through `multi_stop_search`), or crash with "negative dimensions are not allowed" / "slice indices must be integers" depending on where start and goal sit.
- Add the dims check to `multi_stop_search` (was a bare `KeyError: 'y'` from xarray internals) and rewrite the shared message to name the actual dims and the `x=`/`y=` parameters.
- Validate `start`/`goal` are 2-element pairs in `a_star_search`, matching the existing waypoint check, and make that check survive scalar waypoints in `multi_stop_search`.

Backend coverage: the new checks run before backend dispatch, so behavior is identical on numpy, cupy, dask+numpy, and dask+cupy. Verified by executing the bad-input battery on all four (CUDA host).

Exception types on already-raising paths are unchanged. The two silent cases (string barriers, negative radius) now raise, which only affects calls that returned wrong output.

Test plan:
- [x] `pytest xrspatial/tests/test_pathfinding.py` (71 passed, includes cupy and dask+cupy on this host)
- [x] New `TestPathfindingErrorHandling` class: string/scalar barriers, negative/float/zero radius, scalar start/goal/waypoint, dims mismatch in both functions, custom dims still work, numeric barriers still block
- [x] flake8 diff vs main: no new findings

State-CSV row for the error-handling sweep is included.


brendancol (commented):
## PR Review: Validate barriers, search_radius, start/goal points, and dims in pathfinding

None found. The silent-barrier repro from #3649 (wall of zeros, `barriers=['0']`) now raises TypeError on all four backends, and the previously-crashing inputs get errors that name the parameter.

- [ ] `multi_stop_search` forwards `barriers` unvalidated and relies on the first `a_star_search` segment call to reject it (xrspatial/pathfinding.py:1553). The error still fires before any real work, but with `optimize_order=True` the traceback points into `_optimize_waypoint_order` instead of the user's call. One `_validate_barriers(barriers)` line next to the other multi_stop checks would keep the failure at the API boundary.

- [ ] The `a_star_search` docstring still describes `search_radius` as plain `int, optional` and `barriers` as `array like object` (xrspatial/pathfinding.py:925, 932); neither mentions the new constraints (non-negative integer; 1-D numeric). Same for the `Raises` section of `multi_stop_search` (xrspatial/pathfinding.py:1479), which lists only ValueError while the barrier/radius type checks raise TypeError.
- [ ] `_validate_barriers` rejects boolean arrays (`barriers=[True]`) since `np.bool_` is neither integer nor floating. Almost certainly nobody does this, but it did compare successfully before. Worth a conscious decision rather than an accident.
- [ ] proximity.py:1346 carries its own copy of the old malformed dims message ("raster.coords should be named as coordinates:"). Out of scope here; noting it so the wording fix can be replicated there in a follow-up.

### What looks good

- The fix targets the failure at the dispatch boundary, so numpy, cupy, dask+numpy, and dask+cupy all get identical error behavior, and the PR states the battery was executed on all four.
- Exception types on already-raising paths are preserved; only the two silent-wrong-output cases start raising.
- Tests assert message content, not just exception type, and include positives (numeric barriers still block, `search_radius=0` accepted, custom dims still work) so the new validation can't over-reject.
- The scalar-barriers hint ("Did you mean barriers=[0]?") is the kind of message that saves a user a trip to the docs.

### Checklist

- [x] Algorithm matches reference/paper (no algorithm change; validation only)
- [x] All implemented backends produce consistent results (checks run pre-dispatch)
- [x] Edge cases are covered by tests (scalar/string/3-element/negative/float inputs)
- [x] No premature materialization or unnecessary copies (barriers copy is a handful of floats)
- [x] Benchmark exists or is not needed (validation-only change; existing pathfinding benchmark unaffected)
- [ ] Docstrings present and accurate (see nit on parameter constraint docs)

brendancol (commented):
## PR Review follow-up (after 0c8c5179)

- Suggestion, barriers validated only inside the first segment call: fixed. `multi_stop_search` now calls `_validate_barriers` next to its other boundary checks (xrspatial/pathfinding.py:1496), with a test asserting the TypeError fires from the multi_stop call (`test_string_barriers_raise_in_multi_stop`). Re-validation in `a_star_search` is idempotent on the coerced float64 array.
- Nit, docstring constraints: fixed. `barriers` and `search_radius` parameter docs in `a_star_search` state the 1-D-numeric and non-negative-integer requirements, and `multi_stop_search`'s Raises section now lists the TypeError paths.
- Nit, boolean barriers rejected: kept as-is, now with a code comment recording the decision. `[True]` as a barrier list is more plausibly a bug than a request for "cells equal to 1".
- Nit, proximity.py's copy of the old dims message: not touched. This sweep is scoped to the pathfinding module; the wording fix can ride along whenever proximity gets its pass.

72 tests pass locally including the cupy and dask+cupy lanes; flake8 output is byte-identical to main for both touched files. Nothing further from me.

brendancol (commented):
## PR Review: Validate barriers, search_radius, start/goal points, and dims in pathfinding

Re-review after resolving the merge conflict with `main`. The branch was 4 commits
behind and conflicted with the sibling pathfinding PRs (#3653 API consistency, #3667
optimize_order, #3652/#3656 dask token, #3655/#3663 golden tests). Merged
`origin/main` at `9f448802`; the net PR diff keeps its original scope (94 lines in
`pathfinding.py`, 123 in the tests, plus one state-CSV row).

### Conflict resolution

- **barriers docstring** (`a_star_search`): kept main's "Default is no barriers" note
  alongside this branch's "must be a 1-D sequence of numeric values" note, in one
  sentence.
- **barriers handling** (`a_star_search`, ~line 1043): #3653 changed the default from
  `[]` to `None` and added an `if barriers is None: barriers = []` guard. Kept that
  guard, then called `_validate_barriers(barriers)`, which is how the merge had already
  resolved the same pattern in `multi_stop_search`.
- **test_pathfinding.py** (~line 1463): both sides only add tests. Kept both this
  branch's `TestPathfindingErrorHandling` class and main's module-level API-consistency
  tests.
- The state CSV auto-merged to a single pathfinding row.

### Integration check against #3653's new signatures

The main risk was validation still attaching to the pre-merge signatures. What I checked:

- `barriers: Optional[list] = None` and `search_radius: Optional[int] = None` are the
  new defaults. `_validate_search_radius(None)` returns early, and `_validate_barriers`
  runs after the `None -> []` guard, so both defaults pass.
- #3653 added Dataset routing through the `@supports_dataset` decorator on both
  functions. That decorator unwraps a Dataset into per-variable DataArrays and calls the
  wrapped body on each one, so `_validate_surface_dims` (which does
  `surface.dims != (y, x)`) never sees a Dataset, where `.dims` is a mapping rather than
  a tuple. Dataset input is not falsely rejected.
- main's new `test_barriers_default_none_matches_empty_list` checks that `barriers=None`,
  `barriers=[]`, and the default all agree. It passes under this branch's validation.

### Tests

`PYTHONPATH=<worktree> python -m pytest xrspatial/tests/test_pathfinding.py`: 99 passed
in 3.53s. CUDA + cupy are present here, so the GPU and dask+cupy paths ran locally
rather than being skipped.

### What looks good

- Validation sits at the API boundary, before backend dispatch, so all four backends
  get the same errors.
- Error messages name the offending parameter and value and point at the fix (the
  scalar-barriers hint, the list-wrapping suggestion).
- Rejecting bool barrier arrays on purpose is a defensible call, and it's commented.

No blockers. The resolution keeps both this branch's validation and the merged sibling
work.

PR #3663, opened 2026-07-08
Add golden-value reference tests for a_star_search (#3655)

brendancol:
Closes #3655

- Adds golden-value regression tests for `a_star_search`: goal costs on a 6x6 open grid (8- and 4-connectivity, matches the closed form), an 8x8 barrier maze (including a no-path pair), a 6x6 friction surface, and an anisotropic-cellsize grid.
- Expected values were generated on 2026-07-08 and verified to machine precision against scipy 1.16.1 `sparse.csgraph.dijkstra` and scikit-image 0.26.0 `MCP_Geometric` (same edge model: geometric distance, or distance times mean endpoint friction). The values are hardcoded, so neither reference library is needed at test time.
- Test-only; no source changes. The full comparison harness (64x64 inputs, NaN friction holes, path-chain consistency, all four backends) is in #3655.

Backend coverage: the new tests run on numpy and dask+numpy. The existing `test_cupy_matches_numpy` / `test_dask_cupy_matches_numpy` parity tests carry the pinned values to the GPU backends.

Also updates the reference-validation sweep state CSV row for `pathfinding` (verdict MATCHES).

Test plan:
- [x] `pytest xrspatial/tests/test_pathfinding.py -k golden` (18 passed)
- [x] `pytest xrspatial/tests/test_pathfinding.py` (75 passed, includes cupy and dask+cupy on a CUDA host)
- [x] flake8 on the test file: same 14 pre-existing findings before and after

https://claude.ai/code/session_0155N4QGamQVxgpAAPbpQNq4

PR #3666, opened 2026-07-08
Stitch multi_stop_search segments lazily on dask backends

### LINKED ISSUE #3660: multi_stop_search materializes the full grid on dask backends, defeating the sparse A* design
`a_star_search` has a memory-safe dask path: sparse Python A* with an LRU chunk cache, and a lazy output assembled from per-chunk delayed blocks. There is even a regression test (`test_dask_no_large_numpy_arrays`) pinning that no full-size numpy arrays get allocated.

`multi_stop_search` undoes that. On a dask-backed surface it:

- allocates `path_data = np.full(surface.shape, np.nan, dtype=np.float64)` eagerly (`xrspatial/pathfinding.py` line 1473), whatever the backend
- calls `.compute()` on each segment's full-size lazy result inside the segment loop, via `_segment_to_numpy` (line 1488)
- wraps the finished numpy array back into dask with `da.from_array` (line 1521), so the "dask" output is a wrapper around an array that already lived in RAM

`_optimize_waypoint_order` has the same problem: it materializes a full h x w numpy array per waypoint pair (N(N-1) of them) just to read one pixel.

Peak memory therefore scales with the full grid, not the chunk size. The point of the dask backend is out-of-core pathfinding on grids that don't fit in RAM; routing 3 waypoints across such a grid OOMs even though each individual `a_star_search` call is fine.

Repro, executed on this host (2000x2000 float64 = 32 MB, chunks 250x250 = 0.5 MB):

```python
import tracemalloc
from unittest.mock import patch
import numpy as np, dask.array as da, xarray as xr
from xrspatial import a_star_search, multi_stop_search

H = W = 2000
raster = xr.DataArray(da.from_array(np.ones((H, W)), chunks=(250, 250)),
                      dims=['y', 'x'], attrs={'res': (1.0, 1.0)})
raster['y'] = np.linspace(H - 1, 0, H)
raster['x'] = np.linspace(0, W - 1, W)
wps = [(float(H - 1), 0.0), (float(H // 2), float(W // 2)), (0.0, float(W - 1))]

# np.full tracking patch (same approach as test_dask_no_large_numpy_arrays)
# omitted here for brevity; full script attached to the PR.
result = multi_stop_search(raster, wps)
```

Output:

```
full array size          : 32.0 MB
chunk size               : 0.50 MB
a_star_search  peak      : 30.1 MB, full-size np.full allocs: []
multi_stop_search peak   : 133.8 MB, full-size np.full allocs: [('np.full', (2000, 2000))]
multi peak / full array  : 4.18x
multi peak / chunk       : 268x
```

(The 30 MB `a_star_search` peak is frontier-dict overhead, which scales with the explored corridor, not the grid.)

Suggested fix: stitch segments lazily. Each segment's lazy array already carries the surface's chunking, so the cumulative-cost overlay can be `da.where(da.isfinite(seg + offset), seg + offset, acc)`, and the per-segment goal cost only needs the block containing the goal pixel. Junction pixels need no special casing on this path: the overwriting value is identical (segment-start cost 0 plus the cumulative offset), and snap is already rejected for dask inputs. The same single-pixel read fixes `_optimize_waypoint_order`.

Related, not in scope here: with a dask friction surface, every `a_star_search` call recomputes `f_min` with a full nanmin pass over friction, so an N-waypoint route does N-1 full scans of the friction array (N(N-1) with `optimize_order=True`).

Backends affected: dask+numpy and dask+cupy. The numpy and cupy paths are in-memory by design and unaffected.

Found by the performance sweep (deep-sweep, 2026-07-08). OOM verdict for this path at the 30 TB target workload: WILL OOM.

brendancol:
Closes #3660.

`multi_stop_search` on a dask-backed surface allocated a full-size `np.full` path array and computed each segment's full-size lazy result to numpy inside the segment loop, so peak memory scaled with the grid rather than the chunks. The sparse design of `a_star_search`'s dask path (and its `test_dask_no_large_numpy_arrays` guard) was being thrown away one call up the stack.

Changes:

- Dask segments are now combined lazily with `da.where`, keeping the surface's chunking; dask+cupy keeps cupy blocks throughout.
- Per-segment goal costs are read via a new `_cost_at_pixel` helper that computes only the block containing the goal pixel. `_optimize_waypoint_order` uses the same helper, so `optimize_order=True` no longer materializes a full h x w array per waypoint pair.
- Junction pixels need no masking on the dask path: the overwriting value (segment-start cost 0 plus the cumulative offset) equals what the previous segment wrote, and snap is already rejected for dask inputs.
- numpy and cupy paths are unchanged.

Measured on a 2000x2000 dask raster with 250x250 chunks, 3 waypoints (repro script from the issue):

```
before: peak 133.8 MB (4.18x full array), np.full((2000, 2000)) allocated
after : peak  11.0 MB (0.34x full array), no full-size allocations
```

The remaining 11 MB is A* frontier-dict overhead, which scales with the explored corridor.

Backend coverage: numpy and cupy untouched; dask+numpy and dask+cupy verified against numpy results locally (GPU host, cupy tests executed, not skipped).

Tests and benchmarks:

- `test_multi_stop_dask_no_large_numpy_arrays` and `test_multi_stop_optimize_order_dask_no_large_numpy_arrays`: same `np.full`-tracking guard as the existing a_star test, plus parity checks against the numpy backend.
- `MultiStopSearchDaskMemory` asv benchmark (`peakmem_`): large grid (128 MB) with nearby waypoints, so a regression back to eager stitching shows as a step change in peak memory while the benchmark stays under a second.

Test plan:

- [x] `pytest xrspatial/tests/test_pathfinding.py` (59 passed, includes cupy and dask+cupy on this host)
- [x] Issue repro re-run against the fix (output above)
- [x] New benchmark smoke-tested (setup + run: 0.13 s)
- [x] flake8 finding count unchanged vs base

Note: #3648 (benchmarks sweep) touches the same benchmark file; if it merges first this branch needs a trivial conflict resolution in the import block.


brendancol (commented):
## PR Review: Stitch multi_stop_search segments lazily on dask backends

None found. The claimed memory behavior was verified by running the issue repro against this branch (peak 133.8 MB to 11.0 MB on a 2000x2000 / 250x250-chunk raster, no full-size `np.full` calls), and the full pathfinding suite passes on numpy, dask+numpy, cupy, and dask+cupy on a GPU host.

- [ ] `xrspatial/pathfinding.py` (dask branch of the segment loop): each segment adds a `da.where` layer over every chunk, so a route with hundreds of waypoints on a finely chunked grid builds a task graph of roughly `n_waypoints x n_chunks` elementwise tasks. `_MAX_WAYPOINTS` caps this at 1000 and the eager alternative OOMs, so it is a sane tradeoff, but a short comment in the code (or a note in the docstring) would save the next person from rediscovering it.
- [ ] `multi_stop_search` docstring: `a_star_search` documents its dask strategy in a backend table, but the multi-stop docstring says nothing about dask inputs now staying lazy. One sentence ("dask inputs are routed sparsely and stitched lazily; the full grid is never materialized") would document the new contract this PR's tests enforce.

- [ ] `xrspatial/tests/test_pathfinding.py`: the `tracking_full` helper is now copy-pasted three times (a_star test plus the two new multi-stop tests). A small module-level helper or fixture would keep the three guards in sync if the byte threshold ever changes.

### What looks good

- The junction-pixel reasoning (no masking needed because the overwriting value is identical, and snap is rejected for dask) is written down where the code does it, not just in the PR description.
- `_cost_at_pixel` handles the 0-d cupy case for dask+cupy, and `_optimize_waypoint_order` reuses it, so `optimize_order=True` is covered by the same fix rather than left as a known hole.
- The peakmem benchmark uses a large grid with nearby waypoints, so it stays fast while making an eager-stitching regression show up as a step change.
- Known conflict with #3648 (benchmarks sweep touching the same file) is called out in the PR body instead of left for the merger to find.

### Remaining known cost (tracked, not for this PR)

With a dask friction surface, every `a_star_search` call still does a full nanmin pass to compute `f_min`, so an N-waypoint route scans the friction array N-1 times. Issue #3660 records this as out of scope here.

### Checklist

- [x] Algorithm matches the documented sparse-dask design
- [x] All implemented backends produce consistent results (parity tests executed, including GPU)
- [x] NaN handling is correct (NaN + offset stays NaN; `isfinite` gates the overlay)
- [x] Edge cases are covered by tests (optimize_order path has its own no-materialization guard)
- [x] Dask chunk boundaries handled correctly (segments inherit the surface's chunking)
- [x] No premature materialization in the changed paths; single-pixel reads compute one block
- [x] Benchmark exists (`MultiStopSearchDaskMemory.peakmem_multi_stop_search`)
- [x] README feature matrix: no new function, no backend support change, not needed
- [x] Docstrings: no public signature change; see suggestion about documenting the dask contract

brendancol (commented):
## PR Review follow-up: Stitch multi_stop_search segments lazily on dask backends

Re-review after commit 8e22f154, which addressed the earlier pass:

- Graph-growth suggestion: fixed. The dask stitch loop now carries a comment explaining the `n_waypoints x n_chunks` task growth and why `_MAX_WAYPOINTS` bounds it.
- Docstring suggestion: fixed. `multi_stop_search` now states that dask surfaces are routed sparsely and stitched lazily, with peak memory scaling with chunk size and explored corridor.
- Test-helper nit: fixed. The three `np.full` guards now share `_tracking_np_full`, and the pre-existing a_star test was switched to it as well, so the guards stay in sync.

Suite re-run after the changes: 59 passed (numpy, dask+numpy, cupy, dask+cupy on a GPU host). flake8 finding count unchanged from base. No new findings; nothing blocking from my side.

brendancol (commented):
## PR Review: Stitch multi_stop_search segments lazily on dask backends (follow-up: conflict resolution)

This is a follow-up after merging `origin/main` to clear the CONFLICTING state. Pushed as `5f0d0639`.

### Conflict resolution

Only `xrspatial/pathfinding.py` had a real content conflict, inside the TSP cost-matrix loop in `_optimize_waypoint_order`. Two sibling changes overlapped there:

- This PR's lazy read: `_cost_at_pixel(seg.data, goal_py, goal_px)`, which computes only the block holding the goal pixel instead of materializing the whole segment.
- `#3667` on main: reuse of symmetric costs so the reverse A* search is skipped when `snap` is off, computed through a `_pair_cost` helper that still materialized the segment via `_segment_to_numpy`.

I kept main's symmetric-reuse loop and moved the lazy single-pixel read into `_pair_cost` (pathfinding.py:1360-1388). The two optimizations now compose: on dask, the cost matrix is built without materializing any segment, and the reverse direction is reused when `snap` is off. Behavior on numpy is unchanged since both sides read the same requested goal pixel.

`test_pathfinding.py` auto-merged (golden-value tests from `#3655`/`#3663` and the cupy/dask fix from `#3637` came in without conflict). The three sweep-state CSVs text-merged; the performance CSV was untouched by this branch, so no row surgery was needed.

### Lazy-stitch behavior

The core change of this PR (lines 1538-1638) merged with no conflict. The dask path still builds `path_data` as a chunked `da.where` overlay, reads segment goal costs one pixel at a time, and never allocates the full grid. The graph-token reset (`path_agg.name = None`) from `#3652`/`#3656` is intact.

### Test result

`PYTHONPATH=<wt> python -m pytest xrspatial/tests/test_pathfinding.py -x -q` -> 86 passed in 4.74s (CUDA+cupy present, so GPU and dask+cupy lanes ran).

### Checklist
- [x] Merge conflict resolved integrating both sibling changes, not one side
- [x] Lazy dask stitching preserved
- [x] Symmetric-cost reuse from #3667 preserved
- [x] Dask graph-token reset preserved
- [x] All pathfinding tests pass across backends

PR #3667, opened 2026-07-08
Reuse symmetric costs in optimize_order instead of running reverse A* searches

### LINKED ISSUE #3661: optimize_order builds its symmetric cost matrix with N(N-1) A* searches instead of N(N-1)/2
The comment at the top of `xrspatial/pathfinding.py` (line 28) says `optimize_order` "builds an N x N distance matrix and runs N(N-1)/2 A* calls". The code in `_optimize_waypoint_order` actually loops `for i in range(n): for j in range(n)` and runs a full `a_star_search` for every ordered pair, so N(N-1) calls.

The matrix is symmetric: the pixel graph is undirected and edge costs are `geometric_distance * mean(friction at both endpoints)`, so cost(i -> j) == cost(j -> i). Half the searches recompute values already sitting in the matrix, and those searches are the dominant cost of `multi_stop_search(..., optimize_order=True)`.

Measured on this host, 400x400 surface, 6 waypoints:

```
A* calls made for the cost matrix : 30
N(N-1)   (current)                : 30
N(N-1)/2 (sufficient, symmetric)  : 15
total _optimize_waypoint_order time: 0.95s (0.95s of it inside a_star_search)
time wasted on mirror pairs       : ~0.47s
cost(a->b) = 453.800433, cost(b->a) = 453.800433, symmetric: True
```

(The symmetry check above ran on a random non-uniform friction surface, not just uniform costs.)

One caveat: with `snap=True` the goal cost is read at the requested (unsnapped) goal pixel, so if snapping moves a waypoint the two directions can genuinely disagree; one direction reads a pixel the path hit, the other may not. Keeping the double computation for the snap case and mirroring only when snap is off preserves current behavior exactly.

On dask backends the waste is worse than CPU time: each of those calls also materializes a full-size array today (tracked separately in the memory-safety issue).

Found by the performance sweep (deep-sweep, 2026-07-08).

brendancol:
Closes #3661.

`_optimize_waypoint_order` ran `a_star_search` for every ordered waypoint pair, N(N-1) searches, even though the cost matrix is symmetric: the pixel graph is undirected and edge costs are `geometric_distance * mean(friction at both endpoints)`, so cost(i -> j) == cost(j -> i). The comment at the top of `pathfinding.py` already described the intended N(N-1)/2 behavior; the code just never matched it.

Changes:

- The matrix build runs only the upper triangle and mirrors the result, halving the dominant cost of `multi_stop_search(..., optimize_order=True)`. Measured: 6 waypoints on a 400x400 grid went from 30 A* runs to 15, with identical costs on a random non-uniform friction surface.
- With `snap=True` both directions still run. Snapping can move the requested start and goal pixels, and the goal cost is read at the requested pixel, so the two directions are not guaranteed to agree in that case. Behavior there is unchanged.

Backend coverage: the change is in the pure-Python matrix build, upstream of backend dispatch; all four backends benefit equally. Full pathfinding suite (including cupy and dask+cupy tests, executed on this GPU host) passes.

Tests: call-count regressions for both cases (`test_optimize_order_symmetric_matrix_call_count` expects N(N-1)/2 with snap off, `test_optimize_order_snap_keeps_both_directions` expects N(N-1) with snap on), plus the existing `test_optimize_order_*` behavior tests.

Test plan:

- [x] `pytest xrspatial/tests/test_pathfinding.py` (59 passed)
- [x] Issue repro re-run: 15 matrix searches instead of 30, same optimized order
- [x] flake8 finding count unchanged vs base

Note: #3666 touches the same function for the dask memory fix (#3660); whichever merges second needs a small conflict resolution in `_optimize_waypoint_order`.

### DIFF
--- a/xrspatial/pathfinding.py
+++ b/xrspatial/pathfinding.py
@@ -1330,23 +1330,34 @@ def _optimize_waypoint_order(surface, waypoints, barriers, x, y,
     INF = float('inf')
     dist = [[INF] * n for _ in range(n)]
 
+    def _pair_cost(a, b):
+        seg = a_star_search(
+            surface, waypoints[a], waypoints[b],
+            barriers=barriers, x=x, y=y,
+            connectivity=connectivity,
+            snap_start=snap, snap_goal=snap,
+            friction=friction, search_radius=search_radius,
+        )
+        seg_vals = _segment_to_numpy(seg.data)
+        goal_py, goal_px = _get_pixel_id(waypoints[b], surface, x, y)
+        goal_cost = seg_vals[goal_py, goal_px]
+        return float(goal_cost) if np.isfinite(goal_cost) else INF
+
     for i in range(n):
-        for j in range(n):
-            if i == j:
-                dist[i][j] = 0.0
-                continue
-            seg = a_star_search(
-                surface, waypoints[i], waypoints[j],
-                barriers=barriers, x=x, y=y,
-                connectivity=connectivity,
-                snap_start=snap, snap_goal=snap,
-                friction=friction, search_radius=search_radius,
-            )
-            seg_vals = _segment_to_numpy(seg.data)
-            goal_py, goal_px = _get_pixel_id(waypoints[j], surface, x, y)
-            goal_cost = seg_vals[goal_py, goal_px]
-            if np.isfinite(goal_cost):
-                dist[i][j] = goal_cost
+        dist[i][i] = 0.0
+        for j in range(i + 1, n):
+            dist[i][j] = _pair_cost(i, j)
+            if snap:
+                # Snapping can move the requested start and goal pixels,
+                # so the reverse direction may read its cost at a
+                # different pixel; compute it explicitly.
+                dist[j][i] = _pair_cost(j, i)
+            else:
+                # The pixel graph is undirected and edge costs use the
+                # mean friction of both endpoints, so the cost matrix is
+                # symmetric: reuse cost(i -> j) as cost(j -> i) instead
+                # of running the reverse A* search.
+                dist[j][i] = dist[i][j]
 
     # Fixed endpoints: first=0, last=n-1
     if n <= 12:
--- a/xrspatial/tests/test_pathfinding.py
+++ b/xrspatial/tests/test_pathfinding.py
@@ -1062,6 +1062,66 @@ def test_optimize_order_preserves_endpoints():
     assert tuple(order[-1]) == wp3
 
 
+def test_optimize_order_symmetric_matrix_call_count():
+    """Cost matrix reuses symmetric costs: N(N-1)/2 A* runs, not N(N-1).
+
+    Issue #3661: the pixel graph is undirected and edge costs use the
+    mean friction of both endpoints, so cost(i->j) == cost(j->i) and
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Reuse symmetric costs in optimize_order instead of running reverse A* searches

None found. The symmetry argument holds: the pixel graph is undirected, edge cost is `dd[i] * (f_u + f_v) * 0.5` which is invariant under direction, and barriers apply per cell, so shortest-path cost between two pixels is the same both ways. The repro confirmed identical forward and reverse costs on a random non-uniform friction surface, and the matrix build dropped from 30 to 15 searches for 6 waypoints with the same optimized order.

- [ ] `test_optimize_order_snap_keeps_both_directions` runs on an all-ones grid, so snapping never actually moves a pixel; the test pins the call-count contract but not the asymmetry it exists to protect. A variant where a waypoint sits on a NaN cell (so snap relocates it) would exercise the real case. Reasonable to leave for a follow-up since the current test still catches an accidental removal of the snap branch.

- [ ] `_pair_cost` converts the finite cost through `float(...)` where the old code stored `np.float64`. Downstream (`_held_karp`, `_nearest_neighbor_2opt`) only does arithmetic and comparisons, so this is behavior-neutral, but worth knowing it changed.

### What looks good

- The snap caveat is handled by keeping both directions rather than assuming symmetry everywhere; the asymmetry mechanism (goal cost read at the requested, unsnapped pixel) is spelled out in the comment.
- Two call-count regression tests pin both sides of the contract: N(N-1)/2 with snap off, N(N-1) with snap on.
- The module-header comment describing N(N-1)/2 behavior now matches the code instead of the code quietly doing double the work.

### Interaction with other open PRs

#3666 rewrites the same loop's cost extraction for the dask memory fix (#3660). Whichever merges second needs a small manual resolution in `_optimize_waypoint_order`; both PR bodies flag it.

### Checklist

- [x] Algorithm correctness: symmetry holds for the undirected mean-friction cost model; snap exception preserved
- [x] All backends: change is upstream of backend dispatch; full suite (including GPU tests) passes
- [x] NaN handling: unreachable pairs stay INF, mirrored consistently
- [x] Edge cases: snap-on path keeps old behavior and has a call-count test
- [x] No premature materialization introduced (cost extraction unchanged on this branch)
- [x] Benchmark: existing AStarSearch benchmark covers the underlying searches; matrix build has no dedicated benchmark, acceptable for a call-count reduction with test-pinned counts
- [x] README feature matrix: no API change, not needed
- [x] Docstrings: no public signature change

brendancol (commented):
## PR Review follow-up: Reuse symmetric costs in optimize_order

Re-review after commit 6dd3ea71:

- Snap-test suggestion: fixed. The middle waypoint now sits on a NaN cell, so snapping genuinely relocates it and the direction asymmetry the double computation protects is exercised, not just the call count. Running the variant surfaced pre-existing behavior where the unreachable waypoint is dropped from the returned order; that is issue #3646 (filed separately today) and the test notes it rather than papering over it or fixing it out of scope here.
- float vs np.float64 nit: acknowledged in the first pass as behavior-neutral; no change needed.

Suite re-run: 59 passed. flake8 finding count unchanged. Nothing further from my side.

PR #3673, opened 2026-07-18
Add asv benchmarks for edge_detection (#3672)

### LINKED ISSUE #3672: edge_detection has no asv benchmarks
None of the five public functions in `xrspatial/edge_detection.py` (`sobel_x`, `sobel_y`, `prewitt_x`, `prewitt_y`, `laplacian`) appear anywhere in the asv suite:

```
$ grep -rn "sobel\|prewitt\|laplacian\|edge_detection" benchmarks/
(no matches)
```

The compute path delegates to `convolve_2d`, and `benchmarks/benchmarks/convolution.py` does time that, but only at 5x5 and 25x25 kernel sizes with the default boundary mode. The 3x3 case the edge detectors use never runs, and nothing calls the public wrappers, so a regression in either would not show up in benchmark results.

Fix: add `benchmarks/benchmarks/edge_detection.py` following the existing `Convolve2d` pattern (numpy/cupy/dask via `get_xr_dataarray`, one small and one realistic grid size, one timing method per function). Benchmark-only change, no source edits.

Found by the benchmark coverage sweep. Severity per the sweep rubric: medium. The convolution engine is covered; the public functions and the 3x3 kernel case are not.

brendancol:
Closes #3672.

- Adds `benchmarks/benchmarks/edge_detection.py` with an `EdgeDetection` class covering all five public functions (`sobel_x`, `sobel_y`, `prewitt_x`, `prewitt_y`, `laplacian`), parameterized over nx in [300, 3000] and numpy/cupy/dask via the suite's `get_xr_dataarray`.
- Benchmark-only change, no source edits. Also carries the benchmark sweep's state CSV row.

Backend coverage: numpy, cupy, dask+numpy. dask+cupy is not parameterizable with the current `get_xr_dataarray` (common.py has no such type), same as the rest of the suite.

Test plan:
- [x] All 30 function/size/backend combinations executed locally by direct class invocation, including cupy on a GPU host. Timings range from sub-ms (cupy) to ~60 ms (dask at nx=3000), so added CI runtime is small.
- [x] flake8 and isort clean on the new file.


brendancol (commented):
## PR Review: Add asv benchmarks for edge_detection (#3672)

- None.

- None.

- [ ] `benchmarks/benchmarks/edge_detection.py:14-19`: the cupy timings don't synchronize the device after the call, so for async numba kernel launches `time_*` can measure launch overhead rather than kernel time. Every other benchmark in the suite (`common.Benchmarking`, `Convolve2d`) does the same, so fixing it only here would make the numbers incomparable. Fine to leave; a suite-wide sync pass would be its own change.

### What looks good
- `setup` goes through `get_xr_dataarray`, so a missing cupy raises `NotImplementedError` and asv records a skip instead of an error.
- The dask result is forced with `.compute()` (checked via `hasattr` on `result.data`), so the benchmark times the kernel, not graph construction. numpy and cupy results pass through untouched.
- Sizes ([300, 3000]) follow the "one small, one realistic" convention and keep added CI runtime low; the PR body reports all 30 combos executed locally, including cupy on a GPU host.
- The state CSV row is one physical line and matches the sweep schema.

### Checklist
- [x] Algorithm matches reference/paper (n/a, benchmark-only)
- [x] All implemented backends produce consistent results (n/a, no source change)
- [x] Edge cases are covered by tests (n/a, benchmark-only)
- [x] Dask chunk boundaries handled correctly (default `get_xr_dataarray` 2x2 chunking, real chunked execution)
- [x] No premature materialization (`.compute()` only where the timing requires it)
- [x] Benchmark exists or is not needed (this PR is the benchmark)
- [x] README feature matrix updated (n/a, no new functions)
- [x] Docstrings present and accurate (n/a, benchmark classes follow suite style)

PR #3678, opened 2026-07-18
Document NaN propagation in edge_detection docstrings

### LINKED ISSUE #3676: edge_detection docstrings do not mention NaN propagation
**Describe the issue**

The five edge detection functions (`sobel_x`, `sobel_y`, `prewitt_x`, `prewitt_y`, `laplacian`) propagate input NaNs: any output cell whose 3x3 neighborhood contains a NaN becomes NaN. This is intentional (`TestNanHandling::test_nan_in_input_propagates` pins it down) and behaves the same on all four backends, but none of the five docstrings mention it. Run edge detection on a raster with sparse nodata and you get a 3x3 NaN blot around every nodata cell, with nothing in the docs explaining where it came from.

**Reproduction**

```python
import numpy as np
import xarray as xr
from xrspatial.edge_detection import sobel_x

data = np.ones((5, 6), dtype=np.float64)
data[2, 3] = np.nan
result = sobel_x(xr.DataArray(data, dims=['y', 'x']), boundary='reflect')
print(np.isnan(result.data).sum())  # 9 -- one NaN in, nine NaN out
```

Same 3x3 footprint on numpy, dask+numpy, cupy, and dask+cupy.

**Expected behavior**

The docstrings should say that NaN cells propagate to their 3x3 neighborhood, and that with the default `boundary='nan'` the outer one-cell ring of the output is also NaN.

Found during an error-handling audit of `xrspatial/edge_detection.py`.

brendancol:
Closes #3676

- Adds a Notes section to the `sobel_x`, `sobel_y`, `prewitt_x`, `prewitt_y`, and `laplacian` docstrings stating that NaN input cells propagate to their 3x3 neighborhood, and that the default `boundary='nan'` leaves the outer one-cell ring NaN.
- Tightens `test_nan_in_input_propagates` to assert the exact documented footprint: the 3x3 block around the NaN cell is NaN and nothing else is.
- Also commits the error-handling sweep's state CSV row for this module.

Behavior is unchanged; this documents what the code already does. The footprint is identical on numpy, cupy, dask+numpy, and dask+cupy (checked on a GPU host).

Test plan:
- [x] `pytest xrspatial/tests/test_edge_detection.py` - 87 passed, including the cupy and dask+cupy classes


brendancol (commented):
## PR Review: Document NaN propagation in edge_detection docstrings

- none

- none

- [ ] xrspatial/edge_detection.py (Notes sections, all five functions): "every output cell whose 3x3 neighborhood contains a NaN becomes NaN" is exact for interior NaNs, but under `boundary='wrap'` a NaN on one edge also poisons cells on the opposite edge (checked: NaN at (2,0) on a 5x6 raster with `boundary='wrap'` makes rows 1-3 of the last column NaN, since the wrapped neighborhood crosses the seam). Wording like "3x3 neighborhood (as extended by the boundary mode)" would cover that case.

### What looks good
- The documented claim was verified against observed behavior, not just read from the code: one interior NaN produces exactly a 3x3 NaN block, identical on numpy, cupy, dask+numpy, and dask+cupy.
- `test_nan_in_input_propagates` now asserts the full footprint (NaN block and everything else non-NaN) instead of a single cell, so the docstring claim is pinned by a test.
- Behavior is unchanged; the diff is docstrings, one test, and the sweep's state CSV row, which belongs in this PR per the sweep contract.

### Checklist
- [x] Algorithm matches reference/paper (n/a, docs-only)
- [x] All implemented backends produce consistent results (footprint checked on all four)
- [x] NaN handling is correct (and now documented)
- [x] Dask chunk boundaries handled correctly (n/a, no code change)
- [x] Benchmark exists or is not needed (docs-only)
- [x] Docstrings present and accurate (see nit on wrap wording)

brendancol (commented):
## PR Review: follow-up after 907f8ed6

The single nit from the first pass is addressed: all five Notes sections now read "3x3 neighborhood (as extended by the boundary mode)", which covers the wrap-seam case observed during review. Re-ran `pytest xrspatial/tests/test_edge_detection.py`: 87 passed.

No blockers, no suggestions, no nits remaining.

PR #3683, opened 2026-07-18
Add Examples sections and fix cross-correlation wording in edge_detection docstrings

### LINKED ISSUE #3674: edge_detection docstrings: no examples, convolution wording implies the wrong sign, NaN propagation undocumented
A documentation sweep of `xrspatial/edge_detection.py` turned up three docstring problems in the five public functions (`sobel_x`, `sobel_y`, `prewitt_x`, `prewitt_y`, `laplacian`). Doc-only fixes; no code behavior should change.

First, none of the five docstrings has an Examples section. There isn't a single `>>>` block in the module. Parameters and Returns are present and accurate.

Second, the gradient docstrings say e.g. "Detects vertical edges by convolving with the Sobel-X kernel" and show the kernel, but the implementation goes through `convolve_2d`, which applies kernels by cross-correlation (same as `scipy.ndimage.correlate`; the `convolution_2d` docstring says so explicitly). The Sobel and Prewitt kernels are antisymmetric, so convolution and correlation differ by sign. Someone who hand-computes the documented convolution gets the negated result:

```python
from xrspatial import sobel_x
from scipy import ndimage

data = np.arange(30, dtype=np.float64).reshape(5, 6) % 6  # increases left->right
K = np.array([[-1, 0, 1], [-2, 0, 2], [-1, 0, 1]], float)

sobel_x(xr.DataArray(data, dims=['y', 'x'])).data[2, 2]  # 8.0
ndimage.correlate(data, K)[2, 2]                         # 8.0
ndimage.convolve(data, K)[2, 2]                          # -8.0
```

The code matches the usual Sobel/Prewitt convention (positive where values increase toward +x / +y), so the wording is what needs fixing: say the kernel is applied by cross-correlation and state the sign convention. This affects the four gradient operators. `laplacian` has a symmetric kernel, so both operations agree there.

Third, a NaN anywhere in the input propagates to its whole 3x3 neighborhood in the output, independent of the `boundary` mode. `TestNanHandling` in test_edge_detection.py pins this, but the docstrings only describe the `boundary` edge fill and never mention it.

Two smaller things, worth folding into the same pass or skipping: the module has no module-level docstring, and integer input comes back float32 (`convolve_2d` promotes). No docstring makes a dtype claim, so nothing is wrong today.

The rest of the audit came back clean: docstring parameters match the signatures, all five functions are listed in `docs/source/reference/focal.rst`, and the "Supports NumPy, CuPy, Dask+NumPy, and Dask+CuPy" claim checks out on all four backends (run on a CUDA host).

brendancol:
Closes #3674

- Adds an Examples section to each of the five public functions (`sobel_x`, `sobel_y`, `prewitt_x`, `prewitt_y`, `laplacian`). The module had no `>>>` blocks at all; every example here was executed and is doctest-verified.
- Rewords the four gradient docstrings to say the kernel is applied by cross-correlation (`scipy.ndimage.correlate` semantics) and states the sign convention. The old "convolving with" wording predicted the negated result for the antisymmetric Sobel/Prewitt kernels: on a left-to-right ramp, `sobel_x` returns +8 where convolving the documented kernel gives -8. The code matches the standard Sobel/Prewitt convention, so only the wording changes.
- Adds drift-guard tests modeled on `test_docstring_params_match_signature` and `test_docstring_states_all_backends`: documented parameters must match the signatures, Examples sections must exist, and the cross-correlation wording is pinned.

Doc-only; no behavior changes. Also updates the documentation sweep state CSV.

Overlap note: #3678 (error-handling sweep) documents NaN propagation in these same docstrings via Notes sections. I dropped my NaN wording to stay out of its way, but both PRs insert text right after Returns, so whichever merges second will need a trivial conflict resolution (numpydoc order puts Notes before Examples).

Test plan:
- [x] `python -m doctest` on the module: 31 examples pass
- [x] `pytest xrspatial/tests/test_edge_detection.py`: 101 passed
- [x] flake8: no new findings (the 4 E201s in the kernel literals and the E302 in the test file predate this PR)


brendancol (commented):
## PR Review: Add Examples sections and fix cross-correlation wording in edge_detection docstrings

None.

- [ ] Each gradient docstring now says "cross-correlating" twice: once in the lead sentence ("Detects vertical edges by cross-correlating with...") and again in the paragraph right after the kernel ("The kernel is applied by cross-correlation, as in ``scipy.ndimage.correlate``..."), e.g. xrspatial/edge_detection.py:34 and :39. The second paragraph earns its keep by naming the scipy equivalent and the sign convention, but the two could be merged into one sentence if you want the docstrings tighter.
- [ ] Coordination, not a defect: #3678 inserts Notes sections into these same five docstrings right after Returns, where this PR inserts Examples. Whichever merges second gets a conflict at every insertion point. The PR body already says this; resolution is mechanical (numpydoc order puts Notes before Examples).

- [ ] `laplacian`'s docstring skips the cross-correlation paragraph. Correct, since the kernel is symmetric and the two operations agree, but a reader comparing the five docstrings may wonder why it's absent. A short clause ("the kernel is symmetric, so correlation and convolution agree") would preempt that. Fine to leave as is.

### What looks good
- All five examples are real output: they pass bare `doctest.testmod` (31 examples) and `pytest --doctest-modules` on this host.
- The sign-convention fix is backed by measurement, not reading: sobel_x on a left-to-right ramp returns +8, matching `scipy.ndimage.correlate`, while convolving the documented kernel gives -8. The old wording predicted the wrong sign.
- The drift-guard tests follow the existing patterns (`test_docstring_params_match_signature` in test_multispectral.py, phrase-pinning as in test_proximity.py) and pin "cross-correlat" / reject "by convolving" so the wording can't silently regress.
- Doc-only: no function body changed, and the full test file passes (101 tests). Flake8 findings are unchanged from baseline (4 E201 in the kernel literals, 1 E302 in the test file, all pre-existing).

### Checklist
- [x] Algorithm matches reference/paper (wording now matches the measured correlate semantics)
- [x] All implemented backends produce consistent results (verified numpy/cupy/dask+numpy/dask+cupy during the audit; unchanged by this PR)
- [x] NaN handling is correct (unchanged; NaN docs land in #3678)
- [x] Edge cases are covered by tests (unchanged coverage plus new docstring-contract tests)
- [x] Dask chunk boundaries handled correctly (n/a, doc-only)
- [x] No premature materialization or unnecessary copies (n/a, doc-only)
- [x] Benchmark exists or is not needed (not needed, doc-only)

brendancol (commented):
## PR Review follow-up (after 90592c27)

Disposition of the first review's findings:

- Duplicated "cross-correlating"/"cross-correlation" wording: fixed. The paragraph after each gradient kernel now reads "This matches ``scipy.ndimage.correlate``: the response is positive where values increase toward higher column/row index", so the operation is named once, in the lead sentence.
- Overlap with #3678: no change in this PR, by design. Whichever PR merges second resolves the insertion-point conflict; numpydoc order puts its Notes above these Examples.
- `laplacian` symmetry nit: fixed. Its docstring now says "The kernel is symmetric, so cross-correlation and convolution agree."

Re-verified after the commit: `pytest --doctest-modules` on the module plus the test file passes (106 tests, includes all 31 docstring examples), and flake8 shows only the pre-existing baseline findings. No further issues.

Issue #3675, opened 2026-07-18
Style: flake8 E201 in edge_detection kernel literals


brendancol:
Found by a /sweep-style pass over `xrspatial/edge_detection.py`. flake8 (repo config from setup.cfg) reports 4 violations, all E201:

```
xrspatial/edge_detection.py:13:22: E201 whitespace after '['
xrspatial/edge_detection.py:14:22: E201 whitespace after '['
xrspatial/edge_detection.py:22:24: E201 whitespace after '['
xrspatial/edge_detection.py:23:24: E201 whitespace after '['
```

These are the `SOBEL_Y` and `PREWITT_Y` kernel literals, where rows starting with 0 or 1 get a leading space so the columns line up under the -1 rows:

```python
SOBEL_Y = np.array([[-1, -2, -1],
                    [ 0,  0,  0],
                    [ 1,  2,  1]], dtype=np.float64)
```

`LAPLACIAN_KERNEL` in the same file already avoids this (no space after the bracket, extra spaces after commas instead, which flake8 doesn't flag since E241 is off by default). The fix is to drop the space after `[` on those four lines to match.

While in there: `xrspatial/tests/test_edge_detection.py` has one E302 (one blank line instead of two before the first fixture) that can go in the same cleanup.

isort is clean and there are no other flake8 findings on the module. No behavior change involved, kernels are numerically identical.

PR #3685, opened 2026-07-18
Add edge_detection tests for Inf, boundary NaN, degenerate shapes, and multi-chunk dask

### LINKED ISSUE #3682: edge_detection tests never see Inf, boundary NaN, degenerate shapes, or multi-chunk dask
test_edge_detection.py measures well on paper (87 tests, 100% branch coverage of xrspatial/edge_detection.py with JIT disabled) but several input classes never appear in it. Found during a test-coverage audit of the module. Related sibling issues from the same audit: #3672 (benchmarks), #3674/#3676 (docstrings), #3675 (style), #3677 (scipy golden values). This one is about input-space gaps only.

Inf input: `grep -ci inf xrspatial/tests/test_edge_detection.py` returns 0. An inf cell propagates inf and NaN (from inf minus inf) through the 3x3 kernels. Verified all backends agree on this; nothing pins it.

NaN placement and NaN on GPU/dask: NaN input is only tested at an interior cell, only on the numpy backend. No test puts a NaN at the raster edge where it interacts with 'nearest'/'reflect'/'wrap' padding, and the cross-backend equivalence tests (cupy, dask+numpy, dask+cupy) all run on NaN-free ramp data. NaN-input backend divergence has shipped elsewhere in the library before (kde, #3628), so the parity tests should include NaN.

Degenerate shapes: the smallest raster tested is 3x3. 1x1, 1xN and Nx1 never appear. With boundary='nan' a 1x1 input comes back all-NaN; with 'reflect' it comes back 0. A (0, N) empty raster passes through and returns an empty result. None of this is asserted.

Invalid boundary: aspect, morphology, diffusion and bilateral each have a test that boundary='invalid' raises ValueError. edge_detection has none.

Dask chunking: both dask tests use a single chunk equal to the full array (chunks=(5, 6) on a 5x6 raster, chunks=(8, 10) on 8x10), so chunk-boundary stitching in the map_overlap path is never exercised from this module's tests, on dask+numpy or dask+cupy. The accuracy audit probed multi-chunk grids and found them correct, so this is a test gap, not a bug.

Dim names: outputs pass through agg.dims, but only the default y/x dims are tested. A regression that hardcoded dims=['y', 'x'] would pass today's suite.

I probed every case above on a CUDA host: numpy, cupy, dask+numpy and dask+cupy all agree, including NaN-at-corner under all four boundary modes and multi-chunk grids. So the source is fine today and this is test-only work: pin the current behavior so regressions in convolve_2d or the wrappers get caught here.

Probe excerpt (numpy vs dask+numpy chunks=(3, 4), NaN at corner and interior):

```
sobel_x boundary=nan: numpy==dask(multichunk) -> True
sobel_x boundary=nearest: numpy==dask(multichunk) -> True
... (all 20 func x mode combinations True, same for cupy and dask+cupy)
sobel_x 1x1 numpy: [[nan]]
sobel_x 1x1 boundary=reflect: [[0.]]
sobel_x boundary='bogus': ValueError: boundary must be one of ('nan', 'nearest', 'reflect', 'wrap'), got 'bogus'
```

brendancol:
Closes #3682

Test-only change from the test-coverage audit of edge_detection. The source is untouched; every new test pins behavior that was verified correct on a CUDA host but had no coverage.

- Inf and -Inf input: exact expected 3x3 neighborhood per operator (inf where the kernel weight has one sign, NaN where inf meets -inf or a zero weight), unaffected cells stay 0.
- NaN at the raster corner under all four boundary modes, with numpy vs dask+numpy (multi-chunk), cupy, and dask+cupy parity. Earlier NaN tests only used an interior cell on numpy; NaN-input backend divergence has shipped before (kde, #3628).
- Degenerate shapes: 1x1 (all-NaN with boundary='nan', 0 with 'reflect'), 1x6 and 6x1 strips, (0, 5) empty raster.
- Multi-chunk dask equivalence (chunks smaller than the array) for dask+numpy and dask+cupy; the existing dask tests all used one chunk, so chunk stitching was never exercised from this file.
- boundary='invalid' raises ValueError (same test the other convolution-backed modules have).
- lat/lon dim names pass through.

Backends: numpy, cupy, dask+numpy, dask+cupy all exercised; the GPU tests ran locally, not just in CI.

Test plan:
- [x] `pytest xrspatial/tests/test_edge_detection.py` on a CUDA host: 222 passed (was 87)
- [x] flake8 on the test file: no new errors (line 22 E302 pre-exists, tracked by #3675)
- [x] branch coverage of xrspatial.edge_detection stays 100%

Also carries the sweep state CSV row for this audit.


brendancol (commented):
## PR Review: Add edge_detection tests for Inf, boundary NaN, degenerate shapes, and multi-chunk dask

Test-only PR, source untouched. Ran the new suite locally on a CUDA host: 222 passed, GPU tests executed rather than skipped.

None.

- [ ] Integer dtype input is still untested. #3682 calls out input-space gaps, and `convolve_2d` promotes int32 to float32; a small test pinning the promoted dtype and a known value would close that corner (new sections of `xrspatial/tests/test_edge_detection.py`).
- [ ] `TestInfHandling.test_negative_inf` (test_edge_detection.py, ~line 396) only asserts an inf-or-nan mask. The -inf case is the +inf expected block negated with NaN kept, so it could pin exact values the way `test_inf_neighborhood` does.

- [ ] The module-level `inf = np.inf` / `nan = np.nan` aliases (test_edge_detection.py, ~line 358) leak into the module namespace for the sake of one parametrize table. Inlining `np.inf`/`np.nan` in the table would keep the namespace clean, at the cost of longer lines.

### What looks good

- `test_inf_neighborhood` pins exact per-operator 3x3 expectations instead of a loose isfinite check, including the 0-weight -> NaN cells.
- NaN-at-corner parity runs all 4 boundary modes on all 4 backends; this is the combination that has caught real divergence elsewhere (kde #3628).
- Multi-chunk dask tests (chunks=(4,5) on 9x11, chunks=(3,4) on 8x10) finally exercise chunk stitching from this file; previous dask tests were all single-chunk.
- `test_invalid_boundary_raises` matches the pattern aspect/morphology/diffusion/bilateral already use.
- The pre-existing E302 at line 22 was left alone (tracked by #3675), no drive-by edits.

### Checklist

- [x] Algorithm matches reference/paper (n/a, test-only; expectations match probed behavior)
- [x] NaN handling is correct
- [ ] Benchmark exists or is not needed (missing, tracked separately in #3672)
- [x] Docstrings present and accurate (n/a; docstring gaps tracked in #3674/#3676)

brendancol (commented):
## PR Review: follow-up after 65687d4a

Re-checked the branch after the review fixes.

### Blockers

None.

### Suggestions

None. Both earlier suggestions are addressed:

- `TestIntegerDtype.test_int32_promotes_to_float32` pins the float32 promotion and compares against the float64-input result for all five operators.
- `test_negative_inf` now asserts the exact negated neighborhood via the shared `INF_NEIGHBORHOODS` table instead of an inf-or-nan mask.

The nit is also done: the module-level `inf`/`nan` aliases are gone, with the expected blocks built from `np.inf`/`np.nan` directly.

### Verification

- 227 tests pass locally on a CUDA host (was 222 at first review, 87 before the PR).
- flake8 on the file reports only the pre-existing E302 at line 22 (#3675).

Nothing further from me.

Issue #3684, opened 2026-07-18
_promote_float casts 32/64-bit integer rasters to float32, losing integer precision above 2**24


brendancol:
## Description

`_promote_float` in `xrspatial/convolution.py` maps every integer dtype to float32:

```python
def _promote_float(dtype):
    """Return at least float32; preserve float64."""
    if np.issubdtype(dtype, np.floating):
        return dtype
    return np.float32
```

float32 has a 24-bit mantissa, so it cannot represent integers above 2**24. Any consumer that feeds it a 32- or 64-bit integer raster with values past ~16.7M loses the low bits before the kernel runs. This produces silent wrong results through at least `convolve_2d` / `convolution_2d` and `focal.mean` (and `apply` / `focal_stats`, which follow the same contract per #2769).

#3680 hit this through edge_detection and scoped its fix to that module (pre-cast wide ints to float64 before `convolve_2d`). The shared helper still has the problem, and it is a contract question rather than a one-line patch: the int-to-float32 promotion was ratified three times (#1096, #2769, #3214), there is an explicit contract test (`test_mean_int_promotes_to_float32_3214`), and widening int32/int64 to float64 changes output dtype for integer inputs and doubles working memory in the focal paths whose budget guards use `_promote_float(agg.dtype).itemsize`.

## Reproduction

```python
import numpy as np
import xarray as xr
from xrspatial.convolution import convolution_2d
from xrspatial.focal import mean

base = 100_000_000
data = (np.arange(30).reshape(5, 6) % 6 + base).astype(np.int32)
kernel = np.array([[-1, 0, 1], [-2, 0, 2], [-1, 0, 1]], dtype=np.float64)

print(convolution_2d(xr.DataArray(data), kernel).data[2, 2])
# 0.0        (float64 reference: 8.0)

print(mean(xr.DataArray(data)).data[2, 2])
# 1e+08      (float64 reference: 100000002.0)
```

## Options

1. Widen `_promote_float`: int8/int16/uint8/uint16 to float32 (exact), int32/uint32/int64/uint64 to float64 (exact for 32-bit, exact up to 2**53 for 64-bit). Update the contract test, the focal memory-guard comments, and docstring examples that show float32 output for int input.
2. Keep the contract and document the 2**24 limit loudly in `convolve_2d`, `convolution_2d`, `focal.mean`, `apply`, `focal_stats`.
3. Keep float32 but warn at runtime when an integer input's max abs value exceeds 2**24.

[...]

PR #3686, opened 2026-07-18
Promote wide integer rasters to float64 in edge_detection (#3680)

### LINKED ISSUE #3680: edge_detection: integer rasters with values above 2**24 give silently zeroed gradients
## Description

All five edge_detection functions (`sobel_x`, `sobel_y`, `prewitt_x`, `prewitt_y`, `laplacian`) hand `agg.data` to `convolve_2d`, which casts every integer dtype to float32 (`_promote_float` in `xrspatial/convolution.py`). float32 has a 24-bit mantissa, so integers above 2**24 (about 16.7 million) lose their low bits in that cast. Unit-scale steps between neighboring cells disappear before the kernel runs, and the gradient comes back as exactly 0.0. No warning, no error, on any backend.

int32 holds values up to ~2.1e9 and int64 goes far beyond that, but both get squeezed into a type that can only separate integers below 16.7M. Integer rasters at that magnitude are not exotic: cumulative counts, integer-packed IDs, measurements stored in small units.

Same bug class as #2769 and #3214 (float32 precision collapse on large-magnitude values), this time for integer inputs.

## Reproduction

Run on this host, xrspatial main:

```python
import numpy as np
import xarray as xr
from xrspatial.edge_detection import sobel_x

base = 100_000_000
data = (np.arange(30).reshape(5, 6) % 6 + base).astype(np.int32)

print(sobel_x(xr.DataArray(data)).data[2, 2])                     # 0.0
print(sobel_x(xr.DataArray(data.astype(np.float64))).data[2, 2])  # 8.0
```

All four backends agree on the wrong answer:

```
numpy      : 0.0 float32
dask+numpy : 0.0 float32
cupy       : 0.0 float32
dask+cupy  : 0.0 float32
expected   : 8.0
```

Same result with int64 input at base 10_000_000_000.

## Proposed fix

In edge_detection, promote 32- and 64-bit integer inputs to float64 before calling `convolve_2d`. 8- and 16-bit integers keep the existing float32 path since float32 represents them exactly. This makes gradients of int32 input exact, and int64 exact up to 2**53.

The int-to-float32 promotion inside `convolve_2d` is a documented contract shared with focal (#1096, #2769, #3214 all ratified it), so this issue keeps the fix scoped to edge_detection. Whether `_promote_float` itself should widen large integer dtypes is a separate design question for the convolution module; filing that as a follow-up.

Found by /sweep-security (Cat 6, dtype confusion). Severity: HIGH (silent wrong result on realistic inputs).

brendancol:
Closes #3680

`convolve_2d` casts every integer dtype to float32, whose 24-bit mantissa cannot separate integers above 2**24. Sobel, Prewitt, and Laplacian on int32/int64 rasters with large values (offset 1e8 in the repro) returned all-zero gradients on every backend, with no warning.

- Add `_promote_wide_int` in `edge_detection.py`: 32/64-bit integer inputs are cast to float64 before `convolve_2d`. 8/16-bit integers are exact in float32 and keep the existing path. `astype` is lazy on dask, so no materialization.
- Note the resulting output dtypes in the five docstrings.
- Fix stays scoped to edge_detection; the `_promote_float` contract shared with convolution/focal is #3684.

Backend coverage: numpy, cupy, dask+numpy, dask+cupy (all four executed locally, CUDA host).

Test plan:
- [x] New `TestWideIntegerPrecision`: int32 ramp at offset 1e8 matches the float64 reference on all four backends; interior gradients nonzero (the exact silent failure); int64 at offset 1e10; output-dtype table for 8 input dtypes.
- [x] Full `test_edge_detection.py` passes: 117 tests (87 existing + 30 new).
- [x] flake8 on touched files: no new violations (the 4 E201s are pre-existing, tracked in #3675).


brendancol (commented):
## PR Review: Promote wide integer rasters to float64 in edge_detection (#3680)

None.

- [ ] `xrspatial/tests/test_edge_detection.py` (`test_output_dtype_by_input_dtype`): the dtype table covers int8/int16/uint16/int32/uint32/int64 plus the two floats but skips uint8 and uint64. Both hit the same branches, so it's two cheap rows that make the table actually exhaustive over the kinds the helper switches on.
- [ ] `xrspatial/edge_detection.py` docstrings: "32/64-bit integers as float64 so that large values keep unit precision" overpromises slightly for int64/uint64, where float64 is only exact up to 2**53. The `_promote_wide_int` helper docstring states the 2**53 bound; the five public docstrings could carry the same qualifier since that's what users read.

- [ ] `_to_numpy` in the test file re-implements the backend extraction that `general_output_checks` does with an `ArrayTypeFunctionMapping`. Fine as a local helper, but if a third test file grows one of these it belongs in `general_checks.py`.

### What looks good

- The fix targets the actual defect: pre-promoting only 32/64-bit ints keeps 8/16-bit inputs on the established float32 contract (#1096/#2769/#3214) instead of quietly widening everything.
- `astype` on the DataArray's `.data` stays lazy on dask and works unchanged on cupy, so all four backends are covered by one code path; the parametrized regression test executes all four.
- `test_large_int32_gradient_nonzero` pins the exact silent-failure mode from #3680 (all-zero interior), not just closeness to a reference.
- Scoping the shared `_promote_float` question to #3684 rather than reaching into convolution/focal from this PR is the right call given the contract test in test_focal.py.

### Checklist

- [x] Algorithm matches reference/paper (no kernel math changed; promotion arithmetic checked against float32/float64 mantissa widths)
- [x] All implemented backends produce consistent results (parametrized test, all four executed on this host)
- [x] NaN handling is correct (untouched; existing NaN tests still pass)
- [x] Edge cases are covered by tests (int32 above 2**24, int64 at 1e10, 8-dtype output table)
- [x] Dask chunk boundaries handled correctly (no depth/boundary changes)
- [x] No premature materialization or unnecessary copies (lazy astype; one cast only for wide-int inputs)
- [ ] Benchmark exists or is not needed — none exists for edge_detection at all; tracked separately in #3672, not this PR's job
- [x] README feature matrix updated (n/a: no new function, no backend change)
- [x] Docstrings present and accurate (modulo the 2**53 qualifier above)

brendancol (commented):
## PR Review follow-up: Promote wide integer rasters to float64 in edge_detection (#3680)

Disposition of the first-pass findings, re-checked against 033fda80:

- Suggestion (dtype table missing uint8/uint64): fixed. The table now has all eight integer dtypes plus float32/float64; 119 tests pass.
- Suggestion (docstrings overpromise for 64-bit ints): fixed. All five public docstrings now carry the "exact up to 2**53 for 64-bit integers" qualifier, matching the helper docstring.
- Nit (`_to_numpy` duplicates general_checks extraction): dismissed for this PR. It is a four-line test-local helper; promoting it to `general_checks.py` is only worth doing when a second consumer appears.

No new findings in the follow-up diff. Nothing blocking.

PR #3690, opened 2026-07-22
Compare nearest-target distances at float32 on every backend (#3689)

### LINKED ISSUE #3689: allocation/direction: nearest-target tie evaluated at different float precision per backend, diverging on non-lattice grids
**Describe the bug**

`allocation` and `direction` pick the nearest target with an argmin whose comparison precision differs by backend:

- `_process_numpy_bruteforce` (proximity.py:542) compares distances that `_distance` has already rounded to float32 (proximity.py:261). Two targets whose float64 distances differ only past the float32 mantissa count as a tie, and the lowest-flat-index rule from #2856 picks the winner.
- `_proximity_cuda_kernel` (proximity.py:631-635) runs the same loop but compares float64 distances; float32 rounding only happens afterwards for the `max_distance` test (#3389). No tie, so the float64-closer target wins.
- Both cKDTree paths compare float64 too: `cKDTree.query` works in float64 and `_kdtree_query_lowest_index` (proximity.py:875) detects ties with exact float64 equality.

On a grid whose coordinates are exact float64 lattice points (integer coords, res 0.5, 0.25, ...) equidistant targets tie exactly in float64 as well, every backend applies the lowest-flat-index rule, and nothing diverges. That is the only kind of grid the tie-break tests use. On any other grid (res 0.1, `np.linspace`, reprojected coordinates) a geometric tie shows up as float64 distances separated by rounding noise around 1e-13 but identical in float32. The numpy backend calls it a tie and takes the lowest flat index; cupy and the dask cKDTree paths each take whichever target their own float64 arithmetic puts first. The backends then return different allocation values and direction angles for the same input, which the Tie-breaking paragraph in both docstrings promises cannot happen. In a scan of random two-target layouts on a 40x40 res-0.1 grid, 92 of 300 layouts had at least one diverging pixel.

`proximity` itself is unaffected: the returned distance is float32 either way, so both members of a float32 tie produce the same output value.

**Reproduce**

Plain `np.linspace` coordinates, res 0.1, two targets, no perturbation anywhere:

```python
import numpy as np, xarray as xr, dask.array as da, cupy
from xrspatial import allocation, direction

H, W, res = 40, 40, 0.1
data = np.zeros((H, W))
data[27, 35] = 1.0   # target A, lower flat index
data[37, 25] = 2.0   # target B

r['y'] = np.linspace((H - 1) * res, 0, H)
r['x'] = np.linspace(0, (W - 1) * res, W)

alloc_np = allocation(r).data

rd = r.copy(); rd.data = da.from_array(data, chunks=(16, 16))
alloc_da = allocation(rd).data.compute()

rc = r.copy(); rc.data = cupy.asarray(data)
alloc_cp = allocation(rc).data.get()

print(np.argwhere(alloc_np != alloc_da))   # [[10  8]]
print(np.argwhere(alloc_np != alloc_cp))   # [[15 13] [29 27]]
print(alloc_np[10, 8], alloc_da[10, 8])    # 1.0 2.0
```

`direction` diverges at the same pixels (57.80 vs 32.20 degrees at (10, 8)). Note the cupy and dask results also disagree with each other: cKDTree compares squared distances while the CUDA kernel compares square roots, so their float64 orderings differ at 1-ulp separations. Three backends, three different rasters.

**Expected behavior**

One definition of "equidistant", applied by every backend. The outputs are float32, and exact float64 agreement between cKDTree (which compares squared distances) and the kernels (which compare square roots) is not achievable at 1-ulp separations, so the workable convention is the one the numpy backend already uses: targets whose distances round to the same float32 value are tied, and the lowest flat (row-major) index wins. Concretely:

- round the candidate distance to float32 inside the CUDA kernel's argmin loop before comparing (the kernel already rounds for the `max_distance` test);
- detect ties in `_kdtree_query_lowest_index` at float32 precision instead of exact float64 equality;
- state the float32 tie precision in the Tie-breaking paragraphs of the `allocation`/`direction` docstrings;
- pin with a cross-backend test on a non-lattice (res 0.1) grid.

**Additional context**

Related earlier fixes: #2856 introduced the lowest-flat-index rule for exact ties, #3389 aligned the `max_distance` range-test precision to float32, #3090 fixed the target collection order for the cKDTree tie-break. All the existing tie tests use integer-lattice coordinates, which is why this never fired.

Found during a numerical-accuracy sweep of the proximity module. Repro run on Linux with CUDA; numpy vs dask+numpy divergence reproduces without a GPU.

brendancol:
Closes #3689

`allocation` and `direction` evaluated "nearest target" at a different float precision per backend: the numpy brute force compares float32-rounded distances, the CUDA kernel compared float64, and the cKDTree paths detected ties with exact float64 equality. On grids whose coordinates are not exact float64 lattice points (res 0.1, `np.linspace`, reprojected coords), a geometric tie shows up as float64 distances separated by ~1e-13 of rounding noise that vanishes at float32. The numpy backend called it a tie and applied the documented lowest-flat-index rule; cupy and the dask cKDTree paths each returned whichever target their own float64 arithmetic favored, so three backends could produce three different rasters. `proximity` itself was unaffected (the output distance is float32 either way).

Changes:

- `_proximity_cuda_kernel` rounds each candidate distance to float32 inside the argmin loop, so the winning target (and the existing float32 `max_distance` test from #3389) matches the CPU brute force.
- `_kdtree_query_lowest_index` detects ties at float32 instead of exact float64 equality, covering both the global and tiled dask KDTree paths.
- The Tie-breaking paragraphs in the `allocation`/`direction` docstrings now state that distances are compared at float32 precision, the precision of the output.

Backend coverage: numpy (already correct, unchanged), cupy, dask+numpy (bounded map_overlap path was already correct via the brute-force chunk kernel; unbounded KDTree path fixed), dask+cupy (bounded GPU kernel fixed; unbounded routes through the fixed KDTree path).

Also updates the accuracy-sweep state CSV row for proximity.

Test plan:

- [x] New `test_tie_break_float32_precision_nonlattice_grid`: allocation and direction on a 40x40 res-0.1 grid, all four backends, bounded and unbounded `max_distance`, asserting float32-tied pixels resolve to the lowest flat index everywhere. 10 of 16 parametrizations fail without the fix.
- [x] Full proximity suite: 595 passed locally with CUDA (cupy and dask+cupy paths executed, not skipped).
- [x] Reference checks from the sweep: results match `scipy.ndimage.distance_transform_edt` and a float64 brute force for all three metrics and modes.


brendancol (commented):
## PR Review: Compare nearest-target distances at float32 on every backend (#3689)

None.

- [ ] `xrspatial/tests/test_proximity.py:705-706`: the fixture sanity check computes distances with `np.hypot`, but the implementation computes `np.sqrt(x*x + y*y)` (`euclidean_distance`, proximity.py:106-108). The two can differ by one float64 ulp, and this test lives exactly in the regime where one ulp decides whether a pixel counts as float32-tied. Use the same `sqrt(dx*dx + dy*dy)` formula so the oracle and the code under test agree bit-for-bit on every platform.

- [ ] `xrspatial/proximity.py:863-868`: the updated 3+-way-tie caveat in `_kdtree_query_lowest_index` is honest about the k=2 limitation, but now that ties are float32-wide it would be worth stating the practical consequence in one sentence: a 3-way float32 tie whose lowest-flat-index member is the float64-largest resolves differently on the KDTree path than on the kernels. Rare (three targets within one float32 ulp), but the docstring is the only place a future reader will learn it.

### What looks good

- The three backends genuinely disagreed before this change (issue #3689 repro shows numpy, cupy, and dask+numpy each returning a different raster), and the fix picks the only convention all of them can implement exactly: float32, the output precision.
- The CUDA change keeps the strict `<` so the lowest flat index wins a tie without any extra bookkeeping, and the float32 rounding now covers both the argmin and the existing #3389 range test with one cast.
- `isfinite` still gates the KDTree tie test, so inf sentinels from bounded queries cannot alias into float32 ties.
- The regression test self-validates its geometry (asserts at least one float32-tied pixel whose float64 noise favors the higher index) instead of hard-coding pixel positions, and 10 of its 16 parametrizations fail on the pre-fix code.
- Bounded dask paths were left alone, correctly: their chunk function is the brute-force kernel that already compared float32.

### Checklist

- [x] Algorithm matches reference/paper (float32 tie convention now documented in both docstrings)
- [x] All implemented backends produce consistent results (verified on CUDA host: numpy, cupy, dask+numpy, dask+cupy)
- [x] NaN handling is correct (untouched by this change; isfinite gates preserved)
- [x] Edge cases are covered by tests (bounded and unbounded max_distance; pre-fix failure confirmed)
- [x] Dask chunk boundaries handled correctly (bounded map_overlap path unchanged, chunks=(16,16) exercised)
- [x] No premature materialization or unnecessary copies (two `.astype(np.float32)` on k=2 query columns; negligible)
- [x] Benchmark exists or is not needed (behavior fix; one extra cast per candidate in the GPU loop is noise)
- [x] Docstrings present and accurate (Tie-breaking paragraphs updated in allocation and direction)

brendancol (commented):
## PR Review follow-up: commit eabb349f

- The test oracle now computes `np.sqrt(dx*dx + dy*dy)`, the same expression `euclidean_distance` uses, with a comment explaining why `np.hypot` was not safe here. The fixture sanity check and the code under test now agree bit-for-bit.
- `_kdtree_query_lowest_index` spells out the 3-way float32 tie consequence and what it takes to hit it (three targets within one float32 ulp of the same distance).

No new findings. Tie-break tests pass (30/30 locally, CUDA backends included).

PR #3694, opened 2026-07-23
hydro: drop twi dask+cupy host round-trip, vectorize watershed init loops

### LINKED ISSUE #3692: hydro: twi_d8 dask+cupy round-trips chunks through host; watershed init loops run in pure Python
**Describe the bug**

Two performance problems in the hydro subpackage, found during a performance audit on 2026-07-23.

**1. `twi_d8` dask+cupy path round-trips every chunk through the host**

`_twi_dask_cupy` (`xrspatial/hydro/twi_d8.py:116-130`) pulls each chunk to the host with `b.get()`, runs the elementwise math (`tan`, `where`, `log`) on numpy, then copies the result back to the device with `cp.asarray`. TWI has no neighborhood or graph structure, so the same expression runs natively on cupy-backed dask arrays via `_twi_dask`. The host detour looks copied from the D8 iterative drivers, which actually need it (their kernels are numba CPU code). TWI does not.

Measured on this host (8192x8192 float64, 2048-square chunks, warm runs, best of 5):

| scheduler | current round-trip | direct dask+cupy | ratio |
|---|---|---|---|
| threads | 0.672 s | 0.038 s | 17.7x |
| synchronous | 2.170 s | 0.035 s | 62x |

Max abs difference between the two paths: 1.8e-15.

**2. Watershed family initializes labels/state with plain-Python full-grid loops**

The numpy paths of `watershed_d8`, `watershed_dinf`, and `watershed_mfd` (and the cupy fallback paths of the dinf/mfd variants) build the `labels`/`state` arrays with nested Python loops over every cell before handing off to the jitted trace kernel:

- `xrspatial/hydro/watershed_d8.py:1048-1056` (numpy path)
- `xrspatial/hydro/watershed_dinf.py:704-712` (numpy) and `224-232` (cupy)
- `xrspatial/hydro/watershed_mfd.py:706-714` (numpy) and `232-240` (cupy)

On a 2000x2000 grid the init loop takes 0.95 s while the jitted trace kernel takes 0.069 s, so roughly 90% of the numpy-backend runtime is Python loop overhead. A vectorized `np.where` init produces identical `labels`/`state` arrays (checked with `np.array_equal`, NaN positions included) and runs 17x faster, which makes the whole call about 6x faster.

**Expected behavior**

`twi(..., routing='d8')` on dask+cupy input keeps chunks on the device. Watershed init is vectorized so the trace kernel dominates the runtime.

**Additional context**

Found by /sweep-performance (hydro pass). The fix should extend the asv `TWI` benchmark with a dask+cupy case so a reintroduced round-trip shows up in benchmarks. The existing `Watershed` numpy benchmark already covers the init path.

Environment: Linux (WSL2), CUDA available, cupy installed.

brendancol:
Closes #3692

Two performance fixes in the hydro subpackage, from the 2026-07-23 /sweep-performance pass (OOM verdict RISKY unchanged, bottleneck memory-bound).

- `twi_d8` dask+cupy: removed `_twi_dask_cupy`, which pulled every chunk to the host with `.get()`, ran the elementwise math on numpy, and pushed the result back with `cp.asarray`. The dask+cupy branch now calls `_twi_dask` directly; `tan`/`where`/`log` run natively on cupy chunks. Measured on 8192x8192 with 2048-square chunks: 0.672 s before, 0.033 s after (20x). Output matches to 1.8e-15.
- Watershed family: the numpy paths of `watershed_d8`/`watershed_dinf`/`watershed_mfd` and the cupy fallback paths of the dinf/mfd variants built their `labels`/`state` init arrays with nested Python loops over the full grid. Replaced with `np.where` expressions that produce identical arrays. On 2000x2000 the loop was ~90% of the runtime; `watershed_d8` drops from 0.79 s to 0.104 s.

Backend coverage: numpy and cupy paths of the watershed family, dask+cupy path of twi. The dask+numpy paths and all tile-sweep drivers are untouched.

Benchmarks: the asv `TWI` benchmark gains `cupy` and `dask+cupy` params (skipped via `NotImplementedError` when CUDA is unavailable), so a reintroduced host round-trip shows up as a regression. The existing `Watershed` numpy benchmark already covers the init path.

Also includes the sweep-performance state CSV row update for the hydro module.

Test plan:
- [x] `pytest xrspatial/hydro/tests/` (954 passed, CUDA host, cupy and dask+cupy lanes executed)
- [x] New regression test: twi dask+cupy result keeps a cupy meta and adds the same graph layers as the plain dask path
- [x] New test: pour point on a NaN flow_dir cell stays nodata (the exact semantics the old init loop encoded)
- [x] flake8 on touched files introduces no new findings


brendancol (commented):
## PR Review: hydro: drop twi dask+cupy host round-trip, vectorize watershed init loops

None.

- [ ] `xrspatial/hydro/twi_d8.py:67-76`: the dask+cupy branch and the plain dask branch now execute the same call (`_twi_dask(fa_data, sl_data, cellsize)`). The `is_dask_cupy` branch can be removed so the generic `isinstance(fa_data, da.Array)` branch handles both, with the "purely elementwise, no host round-trip" comment moved there. One branch is harder to drift than two.

- [ ] `xrspatial/hydro/watershed_d8.py:51-59`: the memory-budget comment counts 33 B/px but the new init adds two transient boolean masks (about 2 B/px) plus the `fd_valid & pp_valid` temporary. Well inside the 50% guard headroom, so just a comment accuracy point.
- [ ] `xrspatial/hydro/tests/test_twi_d8.py:198-216`: the regression test reads `result.data._meta` and `HighLevelGraph.layers`. Both are stable in practice but not public dask API; if a future dask release reworks graph internals this test will need a touch-up. A code comment already explains the intent, which helps.

### What looks good

- Before/after timings are in the PR body and were measured on real hardware (twi dask+cupy 0.672 s to 0.033 s; watershed numpy 0.79 s to 0.104 s on 2000x2000).
- The vectorized init was checked for exact equality against the old loop output (`np.array_equal`, NaN positions included) before the loop was removed, and a new test pins the one subtle semantic: a pour point on a NaN flow_dir cell stays nodata.
- The graph-layer regression test compares layers added by twi rather than absolute graph size, so it is robust to differences in how the test inputs are constructed.
- The asv `TWI` benchmark gains cupy and dask+cupy params with a `NotImplementedError` guard in setup, matching the repo's convention for skipping GPU benchmarks on CPU-only hosts.
- The dask+numpy paths and all tile-sweep drivers are untouched; the diff stays inside the two audited findings.

### Checklist

- [x] Algorithm matches reference/paper (no algorithm change; init semantics preserved exactly)
- [x] All implemented backends produce consistent results (954 hydro tests pass on a CUDA host, cupy and dask+cupy lanes executed)
- [x] NaN handling is correct (pour-point-on-NaN test added)
- [x] Dask chunk boundaries handled correctly (elementwise only; no overlap involved)
- [x] No premature materialization or unnecessary copies (the PR removes one)
- [x] Benchmark exists (TWI benchmark extended to the fixed path)
- [x] README feature matrix updated: not applicable, tiers unchanged
- [x] Docstrings present and accurate (`_twi_dask` gained a one-liner)

brendancol (commented):
## PR Review: follow-up pass after 38bf988b

Re-checked the branch after the review follow-ups.

- Suggestion (redundant twi dask+cupy branch): fixed in 38bf988b. The `is_dask_cupy` branch is gone, the generic `isinstance(fa_data, da.Array)` branch handles both chunk backends, and the now-unused `is_dask_cupy` import was removed. Dispatch order stays safe: `is_cupy_array` is False for dask-wrapped cupy, so dask+cupy input still reaches `_twi_dask`.
- Nit (memory-budget comment in watershed_d8): fixed in 38bf988b; the comment now mentions the ~3 B/px transient masks and that they fit in the guard headroom.
- Nit (private dask API in the twi regression test): dismissed with reason. The `_meta`/`layers` reads are deliberate, commented in the test, and there is no public API that exposes the same signal. Revisit only if a dask upgrade breaks it.

`flake8` is clean on the touched files and the twi/watershed test files pass (47 tests). No new findings. Nothing blocking.

PR #3696, opened 2026-07-27
Take color_ramp statistics from the materialised COG buffer instead of re-running the source graph

### LINKED ISSUE #3695: to_geotiff with cog=True and color_ramp runs the source dask graph twice
### What happened

`to_geotiff(dask_backed_da, path, cog=True, color_ramp=...)` runs the source dask graph twice. Everything upstream of the write (the read, plus any chained computation) executes a second time just to produce the four numbers that go into the symbology sidecars.

### Reproduction

A 1024x1024 float32 dask source chunked (256, 256) (16 chunks), wrapped in a counting `map_blocks` layer so chunk executions can be counted:

```python
import dask.array as dsa, numpy as np, xarray as xr

CALLS = {"n": 0}

def _count(block):
    CALLS["n"] += 1
    return block

base = np.linspace(0, 100, 1024 * 1024, dtype="float32").reshape(1024, 1024)
coords = {"y": np.linspace(50, 40, 1024), "x": np.linspace(-100, -90, 1024)}

def make():
    arr = dsa.from_array(base, chunks=(256, 256)).map_blocks(_count, dtype="float32")
    return xr.DataArray(arr, dims=["y", "x"], coords=coords)

CALLS["n"] = 0
to_geotiff(make(), "out.tif", cog=True, color_ramp="viridis",
           compression="deflate", tiled=True, tile_size=256)
print(CALLS["n"])   # 33, for a 16-chunk source
```

Measured on this host:

| call | chunk executions | ratio | wall |
|---|---|---|---|
| `cog=False` + `color_ramp` | 17 | 1.06x | 105 ms |
| `cog=True` + `color_ramp` | **33** | **2.06x** | 197 ms |
| `cog=True`, no `color_ramp` | 17 | 1.06x | 43 ms |
| `cog=False`, no `color_ramp` | 17 | 1.06x | 27 ms |

Only the COG-plus-symbology combination pays double.

### What I expected

One execution, the same as every other combination. The write already materialises every pixel; the statistics should come from that buffer.

### Root cause

In `xrspatial/geotiff/_writers/eager.py` the streaming dask path is gated on:

```python
if hasattr(raw, 'dask') and not cog and not _path_is_file_like:
```

That branch carries the fix from #3597: it builds a `StreamingStats` accumulator and threads `chunk_observer=` into `write_streaming`, so the `color_ramp` statistics are folded in from the buffers the write materialises anyway.

`cog=True` deliberately falls through to the eager path, because COG overviews need the full array. There, `arr = raw.compute()` materialises everything, and then `_write_sidecars()` calls `write_symbology_sidecars(path, _sym_data, ..., stats=_sym_stream_stats)` with `_sym_stream_stats` still `None` and `_sym_data` still the original dask-backed DataArray. That reaches `_symbology._finite_stats` -> `_dask_finite_stats`, which issues a fresh `dask.compute(...)` over the same upstream graph.

For a raster read off HTTP or cold storage, or sitting at the end of a long chain, that is the entire pipeline run a second time.

This is the third recurrence of the same class: #3117 (per-tile-row recompute), #3235 (eager pack NaN guard), #3597 (color_ramp stats after the streaming write). #3597 fixed this exact symptom for the streaming path and left the COG branch behind. The `to_geotiff` docstring names only the GPU and VRT paths as doing a second execution, so the COG path is undocumented as well as avoidable.

### Suggested fix

Reuse the plumbing #3597 already added. In the eager branch, right after the array is materialised from dask and before the sentinel-restore rewrite, build a `StreamingStats(nodata=...)` and feed it the materialised buffer in one `update(...)` call, then pass it as `_sym_stream_stats`. `write_symbology_sidecars` already prefers `stats.result()` over `_finite_stats` when `stats` is not None, so no signature changes and the sidecar contents stay identical.

Snapshotting before the sentinel restore matters: the statistics need to describe the same logical values `_finite_stats` would have seen.

### Environment

- Affected backends: dask+numpy and dask+cupy (both reach the same eager COG fallthrough). Plain numpy and cupy are unaffected, since there is no graph to re-execute.
- Found by `/sweep-performance` (pass 19). Module OOM verdict SAFE, bottleneck IO-bound.

brendancol:
Closes #3695.

`to_geotiff(dask_data, path, cog=True, color_ramp=...)` ran the caller's source graph twice. `cog=True` skips the streaming writer (COG overviews need the full array), so the `chunk_observer` accumulation added in #3597 never fired, and `_write_sidecars` fell back to `_finite_stats` on the still-lazy DataArray. Four statistics numbers cost a second full execution of whatever was upstream: an HTTP COG read, cold storage, a long chain of operations.

## Change

- Fold the buffer the eager path has already materialised into the same `StreamingStats` accumulator the streaming path uses, and pass it through the existing `stats=` parameter of `write_symbology_sidecars`. No signature changes.
- The snapshot is taken before the NaN-to-sentinel restore, so the statistics describe the same logical values `_finite_stats` would have seen. The tests assert that against the reduction being replaced.
- One behaviour note: `StreamingStats` accumulates in float64 where `_finite_stats` accumulated at the input's native width, so a float32 source's mean and stddev shift by ~1e-7 relative (float16 ~1e-4). `min` and `max` are exact either way, and integer and float64 sources land within 1e-10. This makes a `cog=True` write agree with the `cog=False` write on the same data instead of diverging, since `cog=False` has used the float64 accumulator since #3597. A test pins that agreement.
- Docstring updated: the `cog=True` write now runs the source once, so only the GPU and VRT paths still take a second pass.

Measured on a 16-chunk source with a counting `map_blocks` layer:

| call | before | after |
|---|---|---|
| `cog=True` + `color_ramp` | 33 | 17 |
| `cog=False` + `color_ramp` | 17 | 17 |
| `cog=True`, no `color_ramp` | 17 | 17 |

This is the fourth time this class has come up: #3117, #3235, #3597, and now this one. Each fix landed on the streaming path and a sibling branch kept the second pass.

## Backend coverage

- dask+numpy and dask+cupy: fixed. Both reach the same eager COG fallthrough.
- numpy and cupy: unaffected, no graph to re-execute. Covered by tests to pin that.

## Benchmark

`benchmarks/benchmarks/geotiff.py` gains `WriteCOGSymbology`, with `time_write_cog_color_ramp` and a `time_write_cog_no_ramp` baseline over numpy and dask at 512 and 2048. On dask at 2048 it reads 65.8 ms before the fix and 48.5 ms after. That is the floor of the win: the benchmark's source is a cheap in-memory `from_array`, so the second execution is nearly free. A real pipeline pays closer to the full 2x.

## Test plan

New module `xrspatial/geotiff/tests/write/test_color_ramp_cog_single_pass_3695.py`, 15 tests. Seven of them fail without the fix.

- [x] Execution counters: `cog=True` + ramp is single-pass; no-ramp COG and the streaming path stay single-pass
- [x] Sidecar values match `_eager_finite_stats`, the reduction the accumulator replaces
- [x] Chunking invariance across `(16,16)`, `(64,64)`, `(8,32)`, `(7,13)`, since Chan combine has to be layout-independent
- [x] nodata exclusion (float and integer), integer sources
- [x] `color_ramp_range` escape hatch still skips the statistics
- [x] Gating unchanged: multiband gets no symbology, single-band 3D does, all-NaN writes no sidecar, constant raster writes stats but no `.qml`
- [x] dask+cupy and cupy legs, run on-device
- [x] Full geotiff suite: 6132 passed, 68 skipped, 1 xfailed

Found by `/sweep-performance` pass 19. Module OOM verdict SAFE, bottleneck IO-bound; the state CSV update is in this branch.


brendancol (commented):
## PR Review: Take color_ramp statistics from the materialised COG buffer

Reviewed the full `_writers/eager.py` context around the change, `_symbology.py` (both `_finite_stats` variants and `StreamingStats`), and the new test module. The fix is small and reuses machinery that already exists, which is the right shape for the fourth recurrence of this bug class.

None.

- [x] `_writers/eager.py:1237` -- the accumulator changes float32 output slightly. `StreamingStats.update` reduces with `dtype=np.float64`; `_eager_finite_stats` reduces at the input's native width. Measured on this branch: float32 mean/stddev move ~1.8e-7 relative, float16 ~1.9e-4. `min`/`max` are exact either way, and integer and float64 sources land within 1e-10. The PR body originally said sidecar contents were unchanged, which is not true for narrow float dtypes. Not a defect (the accumulator is the more accurate one, and it aligns `cog=True` with the `cog=False` path that has used float64 since #3597) but it needed to be stated rather than implied. Fixed in 09b3dca5: added `test_cog_float32_matches_streaming_write` asserting the two dask write paths now emit identical sidecars, plus a code comment and a corrected PR body.

- [x] `_writers/eager.py:1227` -- the comment said `cog=True` "(and a file-like destination)" skips the streaming writer. True of the writer, but misleading here: `_sym_stops` is only assigned under `isinstance(path, str)`, so a file-like destination never reaches this branch with symbology pending. Reworded in 09b3dca5.

### Things I checked that are fine

- Closure binding. `_write_sidecars` is defined at line 549 and reads `_sym_stream_stats` from the enclosing scope; the assignment at 1237 happens before the call at 1313, so it sees the accumulator. The tests would catch a rebinding mistake, and they pass.
- Snapshot ordering. The buffer is captured before the float16/bool promotion, before the band-first `moveaxis`, and before the NaN-to-sentinel restore at 1273. That last one matters: restoring first would fold sentinel values into the statistics. Reductions are order-invariant so the `moveaxis` is irrelevant.
- Semantic parity with what it replaces. NaN nodata collapses to `None` in both `StreamingStats.__init__` and `_finite_stats`. Integer sources with and without a sentinel take matching branches. An empty valid set returns `None` from both, so `write_symbology_sidecars` still writes nothing.
- Gating is unchanged. `color_ramp_range` short-circuits before the accumulator is built, and `_is_single_band` is the same predicate `write_symbology_sidecars` applies, so multiband behaviour cannot drift between the two.
- No new materialisation. `update` uses `buf.ravel()` (a view) on the common all-valid buffer and only compacts with `buf[mask]` when something is masked, which is what the reduction it replaces did anyway.
- Backends. dask+numpy and dask+cupy both funnel through this branch; `arr` is already on the host by line 1226 so `update` always receives numpy. numpy and plain cupy skip the branch entirely and are pinned by tests.
- No CUDA kernels, no `map_overlap`, no dispatch table touched, so the usual neighbourhood and register-pressure checks do not apply.

### Test coverage

Sixteen tests. Seven fail with the fix reverted, which is the useful property. The execution counters use a locked `map_blocks` layer that skips dask's zero-size meta call, matching the #3597 module's approach. Chunking invariance is exercised over ragged layouts (`(7,13)`, `(8,32)`), which is the right check for a Chan combine. The dask+cupy and cupy legs ran on-device here rather than skipping.

One gap worth naming: no test writes a COG through a source whose graph is genuinely expensive, so the tests prove single execution but not the wall-clock win. The asv benchmark covers that, though its in-memory source means it reports the floor of the improvement rather than the typical case.

### Checklist

- [x] Algorithm matches the reduction it replaces (verified numerically per dtype, divergence documented)
- [x] NaN handling correct (NaN nodata, all-NaN input, NaN-bearing float32)
- [x] Edge cases covered (constant raster, all-NaN, single-band 3D, multiband, integer sentinels)
- [x] Dask chunk boundaries handled (chunking-invariance test over ragged layouts)
- [x] Benchmark added (`WriteCOGSymbology`)
- [ ] README feature matrix -- not applicable, no new public function
- [x] Docstrings accurate (`color_ramp` text updated to say `cog=True` now runs the source once)

PR #3698, opened 2026-07-27
Stop the numba fast path claiming CRSs it cannot reproduce (#3697)

### LINKED ISSUE #3697: reproject: numba fast path accepts CRS definitions it cannot reproduce (LCC lat_2, non-metre units, +axis=, GRS80 datums)
## What happened

The numba fast path in `xrspatial/reproject/_projections.py` claims CRS pairs
whose PROJ definition it does not model. When it claims one it projects with the
wrong parameters, and `reproject()` returns a raster whose pixels sit anywhere
from tens of metres to thousands of kilometres from where the target CRS puts
them. There is no warning, and the output metadata looks fine.

I swept all 5363 projected CRSs in the EPSG database. The fast path accepts 2685
of them; 220 of those land more than 5 m from pyproj. Four separate defects add
up to that 220.

### 1. `lat_2` is converted to radians twice (185 CRSs, up to 3.8 km)

`_lcc_params` and `_aea_params` read the second standard parallel like this:

```python
lat_1 = math.radians(d.get('lat_1', d.get('lat_0', 0.0)))
lat_2 = math.radians(d.get('lat_2', lat_1))
```

When `lat_2` is absent the fallback is `lat_1`, which is already in radians, so
`math.radians` runs on it a second time. For a 47 degree standard parallel the
result is 0.82 degrees. The `abs(lat_1 - lat_2) > 1e-10` test then passes and
the two-parallel branch computes a cone constant from a parallel pair that does
not exist:

```
EPSG:8325 (NAD83(2011) / Oregon Mitchell zone)
  n computed = 0.4173471497
  n expected = 0.7313537016  (= sin(lat_1))
  forward error vs pyproj = 3289.174 m
  same call with lat_2 = lat_1 = 2.38e-06 m
```

PROJ writes a single-parallel LCC as `+lat_1=X +lat_0=X +k_0=...` with no
`lat_2`, so every one-standard-parallel Lambert Conformal Conic hits this. Mostly
that is the NAD83(2011), NAD83(HARN) and NATRF2022 low-distortion zone families:
185 EPSG codes, median error 50 m, worst 3819 m.

### 2. Non-metre units are ignored by six parameter extractors (up to 6594 km)

`_lcc_params` and `_tmerc_params` read `units` and return a `to_meter` factor.
`_aea_params`, `_cea_params`, `_laea_params`, `_stere_params`, `_sinu_params`
and `_sterea_params` never look at `units` at all, so a CRS in feet is treated
as metres:

```
EPSG:10481 (NAD83 / TWDB GM, +proj=aea +units=us-ft) -> 6594341.2 m off
```

### 3. `+axis=` is ignored (10 CRSs, up to 7085 km)

`_tmerc_params` accepts `EPSG:2048` (`+proj=tmerc +axis=wsu`, the South African
Lo grid). The kernel emits easting/northing while the CRS stores west/south, so
both components come out negated and swapped. The ten Hartebeesthoek94 Lo zones
(EPSG:2046-2055) are affected, median error 6487 km.

### 4. A non-WGS84 datum on a GRS80/WGS84 ellipsoid passes the guard (24 CRSs, up to 325 m)

`_is_wgs84_compatible_ellipsoid` decides from `to_dict()`:

```python
if (ellps in ('WGS84', 'GRS80', '') and datum in ('WGS84', 'NAD83', '')):
    return True
```

pyproj only emits `+datum=` for datums that have a PROJ alias. Everything else
comes back with no datum key, which this reads as "WGS84". A CRS on the GRS80
ellipsoid with a real shift to WGS84 therefore takes the fast path and the shift
is dropped:

```
EPSG:2100  GGRS87 / Greek Grid        datum=''  ellps='GRS80'  -> 324.7 m off
EPSG:2039  Israel 1993 / Israeli TM   datum=''  ellps='GRS80'  ->  78.9 m off
EPSG:9207  VN-2000 / TM-3 104-30      datum=''  ellps='WGS84'  -> 226.1 m off
```

This is the same bug class as #2651 and #3275: the guard admits a CRS the
kernels cannot reproduce.

## Reproduction

Run on this host, pyproj 3.7.2 / PROJ 9.7.1. The source raster stores its own
longitude as the pixel value, the output grid is pinned explicitly, and each
output pixel is compared against the longitude pyproj assigns to that pixel's
target coordinates. No resampling ambiguity: the source is a smooth ramp sampled
far finer than the output.

```python
import warnings
import numpy as np
import pyproj
import xarray as xr

warnings.simplefilter('ignore')
WGS84 = pyproj.CRS.from_epsg(4326)


def lon_error(label, crs_spec, lon_c, lat_c, span=0.5):
    tgt = pyproj.CRS.from_user_input(crs_spec)
    fwd = pyproj.Transformer.from_crs(WGS84, tgt, always_xy=True)
    inv = pyproj.Transformer.from_crs(tgt, WGS84, always_xy=True)

    n = 401
    lons = np.linspace(lon_c - span, lon_c + span, n)
    lats = np.linspace(lat_c + span, lat_c - span, n)
    src = xr.DataArray(np.tile(lons, (n, 1)), dims=['y', 'x'],
                       coords={'y': lats, 'x': lons},
                       attrs={'crs': 'EPSG:4326'})

    cx, cy = fwd.transform(lon_c, lat_c)
    ex, ey = fwd.transform(np.array([lon_c - span / 2, lon_c + span / 2]),
                           np.array([lat_c - span / 2, lat_c + span / 2]))
    half = max(abs(ex[1] - ex[0]), abs(ey[1] - ey[0])) / 2
    out = reproject(src, tgt, bounds=(cx - half, cy - half, cx + half, cy + half),
                    width=32, height=32, resampling='bilinear')

    GX, GY = np.meshgrid(out.coords['x'].values, out.coords['y'].values)
    true_lon, _ = inv.transform(GX.ravel(), GY.ravel())
    got_lon = np.asarray(out.data).ravel()
    m = np.isfinite(got_lon) & np.isfinite(true_lon)
    metres = np.abs(got_lon[m] - np.asarray(true_lon)[m]) * 111320.0 * np.cos(np.radians(lat_c))
    print(f"  {label:46s} max {metres.max():10.1f} m   median {np.median(metres):8.1f} m")


lon_error("EPSG:8325 Oregon Mitchell zone (LCC 1SP)", "EPSG:8325", -120.25, 44.75)
lon_error("EPSG:9549 LTF2004(C) (LCC 1SP)", "EPSG:9549", 6.82, 45.18)
lon_error("EPSG:2100 GGRS87 / Greek Grid", "EPSG:2100", 24.0, 38.0)
lon_error("EPSG:2039 Israel 1993 / Israeli TM", "EPSG:2039", 35.2, 31.7)
print("  -- control --")
lon_error("EPSG:32633 WGS 84 / UTM zone 33N", "EPSG:32633", 15.0, 50.0)
lon_error("EPSG:2154 RGF93 / Lambert-93 (lat_2 present)", "EPSG:2154", 3.0, 46.5)
lon_error("EPSG:3035 ETRS89 / LAEA Europe", "EPSG:3035", 10.0, 52.0)
```

Output:

```
  EPSG:8325 Oregon Mitchell zone (LCC 1SP)       max      520.4 m   median    243.1 m
  EPSG:9549 LTF2004(C) (LCC 1SP)                 max      212.6 m   median     80.2 m
  EPSG:2100 GGRS87 / Greek Grid                  max      150.7 m   median    149.4 m
  EPSG:2039 Israel 1993 / Israeli TM             max       66.2 m   median     66.0 m
  -- control --
  EPSG:32633 WGS 84 / UTM zone 33N               max        0.0 m   median      0.0 m
  EPSG:2154 RGF93 / Lambert-93 (lat_2 present)   max        0.0 m   median      0.0 m
  EPSG:3035 ETRS89 / LAEA Europe                 max        0.0 m   median      0.0 m
```

Causes 2 and 3 push the data so far off the pinned grid that no output pixel
overlaps the source at all, so they show up directly against the transform
instead:

```python
from xrspatial.reproject import _projections as P
c = pyproj.CRS.from_epsg(2048)
print(P._crs_to_dict(c)['axis'])            # 'wsu'
print(P._tmerc_params(c) is not None)       # True -- accepted anyway

c = pyproj.CRS.from_epsg(10481)
print(P._crs_to_dict(c)['units'])           # 'us-ft'
print(len(P._aea_params(c)))                # 6 -- no unit factor returned
```

## Affected backends

All of them. `try_numba_transform` (numpy, dask+numpy, streaming) and
`try_cuda_transform` (cupy, dask+cupy) share these parameter extractors, and
`transform_points` uses them for output-bounds estimation, so the bounds and the
pixel data are wrong together and stay self-consistent.

## Expected

A CRS the kernels cannot reproduce should fall back to pyproj, the way #2651 and
#3275 made non-WGS84 datums and spherical ellipsoids fall back. A
one-standard-parallel LCC should use `n = sin(lat_1)`.

## Environment

- xarray-spatial main @ 286095b
- pyproj 3.7.2, PROJ 9.7.1
- Linux, Python 3.14

brendancol:
Closes #3697

The fast-path parameter extractors in `_projections.py` accepted CRS
definitions whose PROJ parameters they do not model, then projected them with
the wrong constants. Sweeping the EPSG database, 220 of the 2685 projected CRSs
the fast path accepted came back more than 5 m from pyproj, with no warning.

## What changed

- `_lcc_params` / `_aea_params`: take the standard parallels in degrees before
  converting. The `lat_2` fallback was `lat_1`, already in radians, so
  `math.radians` ran over it twice; a 47 degree parallel became 0.82 degrees and
  the two-parallel branch built a cone constant for a parallel pair that does
  not exist. PROJ writes a single-parallel LCC with no `lat_2` at all, so every
  1SP definition hit it (185 EPSG codes, worst 3819 m).
- `_aea_params`, `_cea_params`, `_laea_params`, `_stere_params`, `_sinu_params`,
  `_sterea_params`: decline a non-metre CRS. These build metre-denominated
  constants and have no `to_meter` factor to return, unlike `_lcc_params` and
  `_tmerc_params`, which keep their foot support.
- `_is_wgs84_compatible_ellipsoid`: reject a non-`enu` `+axis=`, and measure the
  datum offset from WGS84 rather than matching datum names. PROJ only emits
  `+datum=` for datums it has an alias for, so GGRS87, Israel 1993 and VN-2000
  arrived with the key missing and were read as WGS84. The 5 m threshold sits
  above the NAD83-as-WGS84 approximation this module already makes on purpose
  and far below every datum the guard has to catch.

## Result

Same EPSG-wide sweep, before and after:

| | fast path accepted | >5 m from pyproj |
|---|---|---|
| before | 2685 | 220 |
| after | 2649 | 0 |

The 36 CRSs that stopped taking the fast path are the ones it was getting
wrong. They now go through pyproj, which is slower for those users; that is
the intended trade.

## Cost

The datum measurement is one pyproj `Transformer` build per CRS, cached with
`lru_cache`. Measured on the dev box: 9.4 ms on the first call for a given CRS,
about 9.6 us on cache hits. The dispatcher can hit the cache up to seven times
per chunk as it tries each projection family, so roughly 70 us per chunk on top
of a 512x512 kernel.

## Backends

`try_numba_transform` (numpy, dask+numpy, streaming) and `try_cuda_transform`
(cupy, dask+cupy) share these extractors, and `transform_points` uses them for
output-bounds estimation, so all four backends and the bounds math are covered
by the same change. Tests assert both dispatchers decline, and that dask+numpy
matches numpy on a repaired CRS.

## Test plan

- [x] New `xrspatial/tests/test_reproject_fast_path_guard_3697.py`, 36 passed /
      1 skipped, covering all four causes plus the CPU and CUDA dispatchers
- [x] Existing reproject suite green: 622 passed across `test_reproject.py`,
      `test_lite_crs.py` and the #3274 / #3275 files, plus 55 more in the
      #3141 / #3276 / cupy / streaming files
- [x] Kernel-vs-pyproj sweep unchanged for every CRS that still takes the fast
      path (merc, emerc, UTM, tmerc incl. the `lat_0 != 0` Zb path, 2SP LCC,
      AEA, CEA, sinusoidal, LAEA oblique/equatorial/polar, polar stereographic,
      oblique stereographic)
- [x] Resampling invariants re-checked on numpy, cupy, dask+numpy and dask+cupy:
      constant raster stays constant, same-CRS nearest is bitwise identity, no
      NaN leak into fully valid stencils, backends agree to 1.2e-13
- [x] flake8 and isort clean
- [x] CUDA available on the dev box, so the cupy and dask+cupy paths ran for real


brendancol (commented):
## PR Review: Stop the numba fast path claiming CRSs it cannot reproduce (#3697)

Reviewed the full `_projections.py` around the diff, not just the changed
hunks, and re-ran the EPSG-wide differential on the branch to check the
claimed numbers.

None.

- `_projections.py` `_is_wgs84_compatible_ellipsoid` now rejects on three
  unrelated grounds: ellipsoid axes, `+axis=` orientation, and datum offset.
  The name only describes the first. Every caller wants "can the kernels
  reproduce this CRS", so either rename it to something like
  `_fast_path_supports_crs` or say plainly in the docstring that it is the
  general fast-path precondition. A future reader adding a projection will
  otherwise wonder why an ellipsoid predicate is checking axis order.

- The axis check reads `d.get('axis')`, which only exists when PROJ emits
  `+axis=` in the proj4 export. Several CRSs declare north/east ordering in
  WKT and still come back with `axis=None` (EPSG:2193, EPSG:3346). Those are
  fine, because every transformer in this package uses `always_xy=True`,
  which normalizes ordering but not direction, and direction differences do
  surface as `+axis=`. That reasoning is load bearing and is not written
  down anywhere. Add it to the comment so nobody later "fixes" the check by
  widening it to `axis_info` and breaks the two CRSs above.

- `_datum_offset_from_wgs84` probes at the CRS's `lon_0`/`lat_0`. For a few
  CRSs that origin sits outside the area of use. Helmert shifts vary slowly,
  so the measured offset is representative, and the EPSG-wide check backs
  that up. Worth one sentence in the docstring saying the probe point is
  deliberately approximate.

- Bare `except Exception` in `_datum_offset_from_wgs84`. Consistent with the
  surrounding `_crs_to_dict` callers, so not worth changing on its own, but
  it will swallow a genuine pyproj bug as "no shift detected", which fails
  open toward the fast path. Failing closed would be safer given this
  function is a correctness guard.

- The measured cost is 9.4 ms on the first call per CRS and about 9.6 us on
  cache hits, and the dispatcher can call it up to seven times per chunk
  while it tries each projection family. That is noise next to a 512x512
  kernel, but the PR body would be more useful with the number in it, since
  the change adds pyproj work to a path whose whole purpose is avoiding
  pyproj.

- 36 CRSs stop taking the fast path and now go through pyproj, which is
  slower for those users. That is the correct trade, they were getting wrong
  answers, but it is a performance characteristic worth stating in the PR
  body rather than only the correctness win.

### What looks good

- The `lat_2` fix is the right shape: convert the degree values once rather
  than paper over it with an `abs(lat_1 - lat_2)` tolerance bump. Keeping
  `lat_1_deg` and `lat_2_deg` as named locals makes the double conversion
  hard to reintroduce silently.
- Measuring the datum offset instead of extending the datum-name whitelist
  is the fix that keeps working as PROJ adds datums. The previous two
  attempts at this guard (#2651, #3275) both grew name lists and both leaked
  again; this one does not have that failure mode.
- Declining non-metre CRSs in the six extractors that have no `to_meter` to
  return, while leaving `_lcc_params` and `_tmerc_params` foot support
  intact, is the minimal correct split. The parametrized test asserting
  metres-accepted and feet-declined for all six pins it.
- The threshold is chosen against a real number rather than a round guess,
  and the comment records why 5 m sits where it does relative to the
  deliberate NAD83-as-WGS84 approximation.
- The test file covers all four causes, both dispatchers, and dask-vs-numpy
  parity, and the assertions are lower bounds on measured offsets rather
  than exact PROJ-version-specific values.

### Verification I ran

- EPSG-wide differential on this branch: 5304 projected CRSs checked, 2649
  take the fast path, 0 land more than 5 m from pyproj. Matches the PR body.
- Kernel-vs-pyproj forward and inverse sweep across every supported family:
  unchanged from before the PR, the only residuals being the documented
  1.6 mm authalic series truncation and the deliberate 0.3 to 1.8 m NAD83
  approximation.
- Resampling invariants on numpy, cupy, dask+numpy and dask+cupy: constant
  raster stays constant, same-CRS nearest is a bitwise identity, no NaN leak
  into fully valid stencils, backends agree to 1.2e-13. Unaffected by this
  PR, which confirms the change is confined to CRS acceptance.

### Checklist

- [x] Algorithm matches reference (PROJ `lcc` and `aea` setup, `n = sin(lat_1)` for 1SP)
- [x] NaN handling is correct (untouched by this PR, re-verified)
- [x] Dask chunk boundaries handled correctly (parity test included)
- [ ] Benchmark exists or is not needed. No benchmark covers CRS dispatch;
      the added pyproj work is bounded by the cache but unmeasured in asv
- [x] README feature matrix updated (not applicable, no new public function)

brendancol (commented):
## PR Review (follow-up pass): fast-path guard #3697

Re-reviewed after `b7d5789` and `6190122`. Disposition of every finding from
the first pass, then a fresh look at the new code.

### Disposition of the previous findings

| Finding | Verdict |
|---|---|
| `_is_wgs84_compatible_ellipsoid` name no longer describes what it does | Fixed, partially. Docstring now states plainly that it is the general fast-path precondition and asks future checks to live there. Kept the name: nine call sites plus the #3275 tests reference it, and a rename buys nothing the docstring does not. |
| Axis check reads the PROJ dict, not `axis_info`; reasoning undocumented | Fixed. The comment now explains that only direction matters because `always_xy=True` normalizes ordering, names EPSG:2193 and EPSG:3346 as the CRSs a widened check would break, and says not to widen it. `test_axis_order_only_crs_keeps_fast_path` pins it. |
| Probe point may sit outside the area of use | Fixed. Docstring says the probe point is deliberately approximate and why a slowly-varying Helmert shift makes that safe. |
| Bare `except Exception` fails open toward the fast path | Fixed. Now returns `inf` on a failed measurement so the caller rejects, while still returning `None` when there is genuinely nothing to measure (no pyproj, or a lite CRS with no geodetic base). `test_measurement_failure_fails_closed` pins it. |
| Timing numbers absent from the PR body | Fixed. PR body now carries the 9.4 ms cold / 9.6 us warm figures and the per-chunk estimate. |
| 36 CRSs losing the fast path not stated as a perf characteristic | Fixed. PR body now says so explicitly. |
| No asv benchmark covers CRS dispatch | Deferred, not fixed. Adding one means a new benchmark class and a CRS fixture set, which is a bigger diff than the fix it would measure. The cost is bounded by the cache and measured by hand above. Worth a follow-up if the dispatch path gets touched again. |

### New review of the follow-up commits

No blockers.

One thing I checked specifically, because the fail-closed change is the
riskiest of the four: returning `inf` on a measurement failure could in
principle strip the fast path from CRSs that were fine, if some common CRS
happens to throw inside the probe. Re-ran the EPSG-wide differential on
`6190122` to rule that out:

```
checked=5304  fast-path-taken=2649  >5m-off=0
```

Identical to the pre-follow-up run, so nothing new is being rejected. The
change is a pure safety improvement rather than a behaviour change on real
CRSs.

Two smaller observations, neither worth blocking:

- The `None` versus `inf` split now carries real meaning, and the two cases
  are only a few lines apart. The docstring distinguishes them clearly, but
  it is the kind of contract that decays. The new test locks the `inf` half;
  `test_lite_crs_has_no_geodetic_base` locks the `None` half. Between them
  the contract is pinned, so this is an observation, not a request.

- `test_measurement_failure_fails_closed` monkeypatches
  `pyproj.Transformer.from_crs` globally and clears the cache in a `finally`.
  That is correct as written, and `monkeypatch` undoes the patch itself, but
  it does mean the test is order-sensitive if the cache were ever populated
  by an earlier test with the patch still live. The `cache_clear()` on both
  sides handles it.

### What still looks good

Everything from the first pass holds. The follow-up did not weaken any of
it, and the fail-closed change makes the guard behave the way a correctness
guard should: an unverifiable datum now goes to pyproj rather than to
kernels that would silently skip its shift.

### Checklist

- [x] All first-pass findings dispositioned, six fixed and one deferred with a reason
- [x] Follow-up commits re-verified against the EPSG-wide differential
- [x] New behaviour covered by tests
- [x] Full reproject suite still green (622 passed, plus 36 in the new file)
- [x] flake8 and isort clean

PR #3700, opened 2026-07-28
Make the gpu extra pip-installable: drop cuspatial, use cupy-cuda12x/13x wheels

brendancol:
Closes #3699.

`pip install 'xarray-spatial[gpu]'` could not succeed on any machine. Both
packages in the extra break the install:

- `cuspatial` on PyPI is a 2020 name-holder sdist (0.0.1) whose `setup.py`
  overrides the install command to raise "Please install cuspatial via the
  rapidsai conda channel". It contains no module. Nothing in `xrspatial`
  imported it either; the only reference was a guarded no-op in
  `rasterize.py` marked "reserved for future GPU geometry parsing".
- `cupy` on PyPI is sdist-only at every release from 11.6.0 to 14.1.1, so pip
  compiles CuPy from source and needs a full nvcc toolchain. The wheels are
  published per CUDA major as `cupy-cuda12x` / `cupy-cuda13x`.

## Changes

- `setup.cfg`: `gpu` now installs `cupy-cuda12x >=12.3`, and a new
  `gpu-cuda13` extra installs `cupy-cuda13x >=13.6`. `cuspatial` is gone.
- `xrspatial/rasterize.py`: dropped the dead `cuspatial` import.
- `docs/source/getting_started/installation.rst`: the extras table and the
  GPU notes now match what the extras install, and warn against the plain
  `cupy` PyPI project.
- `.github/workflows/test.yml`: new `gpu-extras-resolve` job that resolves
  both GPU extras against PyPI on every PR. CI only ever installed
  `.[tests]`, which is why this sat unnoticed.
- `xrspatial/tests/test_gpu_extras_metadata.py`: offline guards that no extra
  reintroduces `cuspatial` or bare `cupy`, and that each GPU extra names a
  CUDA-suffixed wheel with a floor.

## Evidence for the version floors

`setup.cfg` sets `python_requires = >=3.12`. From PyPI release metadata, the
CPython tags each release ships wheels for:

```
### cupy-cuda12x
  12.0.0: ['cp310', 'cp311', 'cp38', 'cp39']
  12.3.0: ['cp310', 'cp311', 'cp312', 'cp38', 'cp39']    <- first cp312
  13.6.0: ['cp310', 'cp311', 'cp312', 'cp313', 'cp39']
  14.1.1: ['cp310', 'cp311', 'cp312', 'cp313', 'cp314']

### cupy-cuda13x
  13.6.0: ['cp310', 'cp311', 'cp312', 'cp313', 'cp39']    <- first release
[...]

PR #3702, opened 2026-07-28
Declare dependency version bounds and test them in CI

### LINKED ISSUE #3701: Dependency floors are undeclared and untested: numba's numpy ceiling is not mirrored, pandas is missing, zstandard is over-declared
## What's wrong

`setup.cfg` declares six required dependencies, every one of them a bare name with no version bound:

```ini
[options]
install_requires =
    numba
    scipy
    xarray
    numpy
    urllib3
    zstandard
```

Four problems come out of that block.

### 1. numba's numpy ceiling is not mirrored

numba caps numpy in its own metadata and enforces the cap again at import time. Current numba releases:

| numba | declared numpy range |
|-------|----------------------|
| 0.66.0 | `numpy<2.5,>=1.22` |
| 0.65.1 | `numpy<2.5,>=1.22` |
| 0.64.0 | `numpy<2.5,>=1.22` |
| 0.63.1 | `numpy<2.4,>=1.22` |
| 0.61.2 | `numpy<2.3,>=1.24` |

numpy 2.5.1 is the current release on PyPI and no numba supports it yet. Because xarray-spatial declares a bare `numpy`, its metadata claims compatibility with a numpy where `import xrspatial` cannot work, and pip's diagnostics are useless when the constraint bites.

### 2. pandas is a direct dependency declared nowhere

`pandas` is imported unguarded at module top level in three files, and it shows up in the public API: `zonal_stats` and `crosstab` are annotated `-> pd.DataFrame` and build their results with `pd.DataFrame(...)`.

- `xrspatial/zonal.py:26`
- `xrspatial/dasymetric.py:35`
- `xrspatial/focal.py:11`

It resolves today only because xarray depends on pandas transitively. Nothing pins that relationship, and nothing constrains which pandas arrives: a fresh resolve currently pulls pandas 3.0.5.

### 3. zstandard is declared required, but the code treats it as optional

`xrspatial/geotiff/_compression.py:1402-1407` guards the import and sets a capability flag:

```python
ZSTD_AVAILABLE = False
try:
    import zstandard as _zstd
    ZSTD_AVAILABLE = True
except ImportError:
    _zstd = None
```

`zstd_decompress` and `zstd_compress` then raise `ImportError("zstandard is required to read ZSTD-compressed TIFFs. Install it with: pip install zstandard")`. That is the code declaring zstandard optional while `install_requires` declares it mandatory. The same GeoTIFF-codec role is already handled by the `geotiff` extra for `deflate` and `pyproj`, so zstandard is the odd one out.

### 4. Nothing tests a minimum-version set

Every workflow installs the newest of everything:

- `.github/workflows/test.yml:50` — `pip install -e .[tests]`
- `.github/workflows/docs.yml:45` — `pip install -e '.[doc,tests]'`
- `.github/workflows/test-geotiff-corpus.yml:68` — `pip install -e .[tests]`
- `.github/workflows/test-cog-validator.yml:98` — `pip install -e .[tests]`

`grep -rln 'min-deps\|minimum\|oldest\|floor' .github/workflows/` returns nothing. So whatever floors get declared would be guesses unless a job installs them.

## Reproduction

Linux, CPython 3.14.2, pip 25.3, numba 0.65.0, numpy 2.3.2.

**a. Metadata permits a numpy that cannot import.** Simulating the numpy version that pip will happily install today:

```console
$ python -c "
import numpy as np
np.__version__ = '2.5.1'
import xrspatial
"
ImportError: Numba needs NumPy 2.4 or less. Got NumPy 2.5.
```

**b. pip will install that numpy without complaint from xarray-spatial.** In an environment that already holds xarray-spatial and numba:

```console
$ python -m pip install --dry-run --upgrade numpy
Requirement already satisfied: numpy in .../site-packages (2.3.2)
Collecting numpy
  Using cached numpy-2.5.1-cp314-cp314-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl.metadata
Would install numpy-2.5.1
```

**c. When the constraint does bite, pip blames 2020-era releases.** Because no published xarray-spatial names its real numpy ceiling, the resolver walks the entire release history:

```console
$ pip install --dry-run --ignore-installed --target ./t --only-binary=:all: 'xarray-spatial' 'numpy==2.5.1'
ERROR: Cannot install numpy==2.5.1, xarray-spatial==0.2.8, xarray-spatial==0.2.9,
xarray-spatial==0.3.0, xarray-spatial==0.3.1 and xarray-spatial==0.3.7 because
these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested numpy==2.5.1
    xarray-spatial 0.3.7 depends on numpy<=1.23.4
    xarray-spatial 0.3.1 depends on numpy<=1.20 and >=1.7
    ...
ERROR: ResolutionImpossible
```

The message never mentions numba, which is the package that actually caps numpy.

**d. Blocking each declared dependency, to separate the required from the optional.** `sys.modules[pkg] = None`, then `import xrspatial`:

```
block pandas      -> ModuleNotFoundError: import of pandas halted
block scipy       -> ModuleNotFoundError: No module named 'scipy.ndimage'
block urllib3     -> ModuleNotFoundError: import of urllib3 halted
block numba       -> ModuleNotFoundError: import of numba halted
block xarray      -> ModuleNotFoundError: import of xarray halted
block numpy       -> ModuleNotFoundError: import of numpy halted
block zstandard   -> IMPORT OK
```

pandas is required but undeclared; zstandard is declared but not required.

## Evidence for the floors

CPython 3.12 is the minimum interpreter (`python_requires = >=3.12`), so the compiled dependencies cannot go below their first cp312 wheel. Queried from the PyPI JSON API:

| package | first release shipping a cp312 wheel |
|---------|--------------------------------------|
| numpy   | 1.26.0 |
| numba   | 0.59.0 |
| scipy   | 1.11.2 |
| pandas  | 2.1.1 |

xarray is pure Python and has no wheel anchor; 2023.10.0 is the first release published after CPython 3.12.0 (2023-10-02).

For urllib3, OSV reports that every release before 2.7.0 carries at least one unfixed advisory against the streaming read path: GHSA-2xpw-w6gg-jr37, GHSA-gm62-xv2j-4w53 and GHSA-38jv-5279-wg99, all decompression handling in the streaming API. `xrspatial/geotiff/_sources.py:1507` streams responses with `preload_content=False` and a byte cap, so that is the code path in use. The redirect-related urllib3 CVEs do not apply here: `_sources.py:445` builds the pool with `redirect=False` and follows redirects by hand through `_validate_http_url`.

The numpy ceiling `<2.5` mirrors numba 0.64 through 0.66 and numba's runtime check in `_ensure_critical_deps`.

## Proposed fix

1. Declare bounds in `setup.cfg`: `numba>=0.59`, `scipy>=1.11.2`, `xarray>=2023.10.0`, `numpy>=1.26,<2.5`, `urllib3>=2.7`, and add the missing `pandas>=2.1.1`.
2. Move `zstandard` out of `install_requires` into the `geotiff` extra alongside `deflate` and `pyproj`, and add it to the `tests` extra so CI keeps exercising the ZSTD codec paths.
3. Add a `min-deps` job to `.github/workflows/test.yml` that pins every floor and runs the fast lane, so the numbers stay honest.

brendancol:
Closes #3701.

`install_requires` listed six bare package names with no version bound, and nothing in CI ever installed an old one. This declares bounds, adds the dependency that was missing, moves out the one that was never required, and adds a job that installs the floors so the numbers stay honest.

## What changed

```ini
install_requires =
    numba >=0.59
    scipy >=1.11.2
    xarray >=2025.11
    numpy >=1.26,<2.5
    pandas >=2.2
    urllib3 >=2.7
```

`zstandard >=0.22` moves to the `geotiff` extra (beside `deflate` and `pyproj`) and to `tests`, so the ZSTD codec tests keep running in CI instead of silently skipping.

## Where each bound comes from

**numba >=0.59, scipy >=1.11.2, numpy >=1.26** are the first releases of each that ship a CPython 3.12 wheel, and `python_requires = >=3.12` already rules out anything older. Checked against the PyPI JSON API rather than guessed:

| package | first release with a cp312 wheel |
|---------|----------------------------------|
| numpy   | 1.26.0 |
| numba   | 0.59.0 |
| scipy   | 1.11.2 |
| pandas  | 2.1.1 |
| zstandard | 0.22.0 |

**numpy <2.5** mirrors numba. Every current numba declares a numpy ceiling in its own metadata, and re-checks it at import in `numba._ensure_critical_deps`:

| numba | declared |
|-------|----------|
| 0.66.0 | `numpy<2.5,>=1.22` |
| 0.65.1 | `numpy<2.5,>=1.22` |
| 0.64.0 | `numpy<2.5,>=1.22` |
| 0.63.1 | `numpy<2.4,>=1.22` |

numpy 2.5.1 is on PyPI now and no numba supports it, so the old bare `numpy` claimed support for a version where the package cannot import:

```console
$ python -c "
import numpy as np
np.__version__ = '2.5.1'
import xrspatial
"
ImportError: Numba needs NumPy 2.4 or less. Got NumPy 2.5.

$ python -m pip install --dry-run --upgrade numpy
Would install numpy-2.5.1
```

**xarray >=2025.11** is the one bound that is not a wheel-availability artifact, and it is the interesting one. Running the suite in a Python 3.12 virtualenv pinned to the older candidate floors produced 7 failures, all the same assertion (`output_agg.attrs == input_agg.attrs`):

```
FAILED xrspatial/tests/test_corridor.py::test_precomputed_keeps_source_attrs_not_friction
FAILED xrspatial/tests/test_mcda.py::TestMetadataPreservation::test_combine
FAILED xrspatial/tests/test_mcda.py::TestMetadataPreservation::test_sensitivity_oat
FAILED xrspatial/tests/test_morphology_derived.py::test_gradient_output_metadata
FAILED xrspatial/tests/test_morphology_derived.py::test_white_tophat_output_metadata
FAILED xrspatial/tests/test_morphology_derived.py::test_black_tophat_output_metadata
FAILED xrspatial/tests/test_morphology_derived.py::test_default_kernel
```

Those functions build their result with a DataArray binary op (`morph_white_tophat` is `agg - opened`), and older xarray drops attrs across those, so `res` and `crs` never reach the output. Bisecting with everything else held fixed:

```
xarray==2023.10.0 -> 7 failed
xarray==2024.1.0  -> 7 failed
xarray==2024.7.0  -> 7 failed
xarray==2025.1.0  -> 7 failed
xarray==2025.7.0  -> 7 failed
xarray==2025.10.0 -> 7 failed
xarray==2025.11.0 -> passed
```

Ruled out a co-installed package: `xarray==2025.10.0 + pandas==3.0.5` still fails, `xarray==2025.12.0 + pandas==2.2.3` passes.

**pandas >=2.2** follows from that, since xarray 2025.11.0 declares `pandas>=2.2`. pandas needed declaring regardless: `xrspatial/zonal.py:26`, `xrspatial/dasymetric.py:35` and `xrspatial/focal.py:11` import it unguarded at module scope, and `zonal_stats` and `crosstab` are annotated `-> pd.DataFrame`. It was resolving purely because xarray drags it in.

**urllib3 >=2.7** is the first release OSV reports with no open advisory against the streaming read path. Everything earlier carries at least one of GHSA-2xpw-w6gg-jr37, GHSA-gm62-xv2j-4w53 or GHSA-38jv-5279-wg99, all decompression handling in the streaming API, and `xrspatial/geotiff/_sources.py:1507` streams responses with `preload_content=False` under a byte cap. The redirect-related urllib3 CVEs do not apply here: `_sources.py:445` builds the pool with `redirect=False` and validates each `Location` by hand.

**zstandard** was never required. `xrspatial/geotiff/_compression.py:1402` guards the import behind `ZSTD_AVAILABLE`, and `zstd_compress` / `zstd_decompress` raise a pip-install hint when it is missing. Blocking it with `sys.modules['zstandard'] = None` leaves `import xrspatial` working, unlike the other five:

```
block pandas      -> ModuleNotFoundError
block scipy       -> ModuleNotFoundError
block urllib3     -> ModuleNotFoundError
block numba       -> ModuleNotFoundError
block xarray      -> ModuleNotFoundError
block numpy       -> ModuleNotFoundError
block zstandard   -> IMPORT OK
```

## Verification

Fast lane in a Python 3.12 virtualenv pinned to exactly the declared floors (`numba==0.59.0 scipy==1.11.2 xarray==2025.11.0 numpy==1.26.0 pandas==2.2.0 urllib3==2.7.0 zstandard==0.22.0`):

```
10797 passed, 2992 skipped, 1 xfailed, 147 warnings in 241.08s
```

The new `min-deps` job in `.github/workflows/test.yml` runs that same install and suite on every PR, so a floor that stops being true shows up as a red check rather than as a user bug report.

`xrspatial/tests/test_dependency_metadata_3701.py` reads `setup.cfg` from the source checkout (skipped when absent, e.g. from an installed wheel) and asserts pandas is declared, zstandard is not, every requirement carries a lower bound, numpy carries an upper bound, and the installed numpy satisfies both setup.cfg and numba. The upper-bound test fails against the old bare `numpy` declaration.

Backend coverage: metadata and CI only, no compute paths touched. The floor run above exercised numpy and dask+numpy; the cupy paths skip in that virtualenv and are unaffected by the change.

## Test plan

- [x] `pytest -m "not slow"` at the declared floors on Python 3.12: 10797 passed, 0 failed
- [x] `pytest xrspatial/tests/test_dependency_metadata_3701.py` on both Python 3.12 at the floors and Python 3.14 at current versions
- [x] flake8 and isort clean on the new test file
- [x] `setup.cfg` parses via `setuptools.config.setupcfg.read_configuration`
- [x] `.github/workflows/test.yml` parses as YAML with both jobs present

## For the maintainer

Two calls in here want an explicit yes rather than a silent merge:

- **`zstandard` leaves the default install.** A plain `pip install xarray-spatial` no longer reads ZSTD-compressed TIFFs; the failure surfaces at decode time as `ImportError: zstandard is required to read ZSTD-compressed TIFFs. Install it with: pip install zstandard`, not at import. `pip install 'xarray-spatial[geotiff]'` restores it. This needs a line in the release notes at tag time (not touched here, since parallel branches all editing CHANGELOG.md conflict).
- **`urllib3 >=2.7` is the tightest bound in the set** and the only one driven by advisories rather than by an API the code calls. It is free on PyPI (2.7.0 is what pip resolves today), but conda-forge and distro packagers lag, so anyone on 2.6.x gets a resolver error instead of a working install with an older HTTP stack. Happy to relax it to a functionality-only floor if you would rather carry the advisories.


brendancol (commented):
## PR Review: Declare dependency version bounds and test them in CI

No compute code changed, so most of the usual correctness checks (algorithm, NaN semantics, dask chunk boundaries, CUDA bounds) do not apply. This is a packaging and CI review.

None.

- [ ] `xrspatial/tests/test_dependency_metadata_3701.py:11` imports `packaging.requirements`, but `packaging` is not in the `tests` extra. It arrives only because xarray declares `packaging>=24.1`. That is the exact pattern this PR is fixing for pandas, so the test that enforces the rule should not break it. Add `packaging` to the `tests` extra in `setup.cfg:111`.

- [ ] `setup.cfg:35` sets `urllib3 >=2.7`, which is the tightest bound in the set and the only one motivated by advisories rather than by an API the code needs. 2.7.0 is what pip resolves today so it costs nothing on PyPI, but conda-forge and distro packagers lag, and a user on urllib3 2.6.x will now get a resolver error rather than a working install with a stale HTTP stack. That may still be the right call given `_sources.py:1507` streams under a byte cap; it is worth a maintainer's explicit sign-off rather than sliding in with the wheel-availability floors.

- [ ] Dropping `zstandard` from `install_requires` (`setup.cfg:98`) is user-visible: anyone on a plain `pip install xarray-spatial` who reads a ZSTD-compressed TIFF goes from working to `ImportError`, and the error only appears at decode time, not at import. The message is actionable and the extras convention is right, but this belongs in the release notes at tag time.

- [ ] `.github/workflows/test.yml:75` gives `min-deps` no `if:` guard, so it runs identically on pull_request, push-to-main, the nightly cron, and `workflow_dispatch`, always on the fast lane. The `run` job distinguishes those (fast lane on PRs, full suite otherwise). Keeping min-deps cheap everywhere is defensible; just note it is a deliberate divergence rather than an oversight.

- [ ] `setup.cfg:32` pins `numpy <2.5` to mirror numba, and `test_numpy_has_an_upper_bound` only checks that some ceiling exists. When numba widens to `<2.6` there is nothing that surfaces the stale cap, so xarray-spatial silently blocks a numpy it would otherwise support. Not worth building tooling for, but the comment at `setup.cfg:26` should probably say it needs a look on every numba bump.

- [ ] `test_dependency_metadata_3701.py:22` re-reads and re-parses `setup.cfg` on each helper call (five times per run). Harmless at this size.

### What looks good

- The floors are derived rather than rounded. First-cp312-wheel for the compiled packages is a rule that regenerates the same answer next year, and the xarray bound comes from an actual bisect rather than from picking a recent-looking date.

- The xarray finding is the substantive one. `morph_white_tophat` is `agg - opened`, so on any xarray before 2025.11 the output silently loses `res` and `crs`. That is a real metadata bug reachable through today's declared metadata, and the seven failing tests are the evidence.

- `test_numpy_has_an_upper_bound` fails against the pre-change bare `numpy` declaration, so the guard is not vacuous. Reading `setup.cfg` from the source tree instead of `importlib.metadata` avoids reporting on a stale editable install, and the skip when `setup.cfg` is absent keeps it from failing inside a wheel.

- Adding `zstandard` to the `tests` extra alongside the `geotiff` move is the part that is easy to forget; without it the ZSTD codec tests would have skipped silently and nobody would have noticed for months.

### Checklist

- [x] Algorithm matches reference/paper (n/a, no compute code)
- [x] All implemented backends produce consistent results (n/a; numpy and dask+numpy exercised by the floor run)
- [x] Edge cases are covered by tests (metadata guards cover the pre-change declaration)
- [x] README feature matrix updated (not applicable, no new functions)
- [x] Docstrings present and accurate (test module documents why it reads setup.cfg rather than installed metadata)

brendancol (commented):
## PR Review: Declare dependency version bounds and test them in CI (follow-up pass)

Second pass over `62e3300b`, which addresses the first review.

None.

None outstanding. Disposition of the first pass:

- `packaging` missing from the `tests` extra: fixed at `setup.cfg:120`, with a comment saying why a test file should not lean on a transitive dependency.
- `urllib3 >=2.7` wanting a maintainer decision: not changed. The advisory class (GHSA-2xpw-w6gg-jr37 and friends, streaming decompression) lands on the byte-capped streaming reads at `_sources.py:1507`, and 2.7.0 is what pip resolves today, so the bound is free on PyPI. Surfaced explicitly in the PR body under "For the maintainer" rather than decided unilaterally.
- ZSTD leaving the default install: not changed, and cannot be, since `CHANGELOG.md` is off limits on a feature branch. Written into the PR body so it reaches the release notes at tag time.

- [ ] `xrspatial/tests/test_dependency_metadata_3701.py:15` caches the parsed config with `lru_cache`, so a test that edits `setup.cfg` mid-session would read stale values. Nothing does that today and these tests are read-only, so this is only worth remembering if someone later adds a test that writes the file.

### What looks good

- The `packaging` fix is the right shape. A guard against undeclared transitive imports that itself relied on an undeclared transitive import would have been an easy thing to shrug at.

- `setup.cfg:29-31` now says the numpy ceiling needs re-checking on every numba bump. That is the failure mode a mirrored bound actually has: it goes stale quietly and blocks a numpy that works fine.

- The min-deps comment at `.github/workflows/test.yml:78-80` explains the fast-lane-everywhere choice instead of leaving it looking like a copy-paste omission.

### Checklist

- [x] Algorithm matches reference/paper (n/a, no compute code)
- [x] All implemented backends produce consistent results (n/a; numpy and dask+numpy exercised by the floor run)

brendancol:
CI note: everything in the `pytest` workflow is green on `93f76b5`, including the new `min-deps` job (8m6s) and all three `3.14t` runners. `docs-build` on GitHub Actions passed in 4m54s.

The one red check is `docs/readthedocs.org:xarray-spatial`, and it is not this PR. Pulling the build record for 33792678:

```
0 | python -m pip install --upgrade --upgrade-strategy only-if-needed --no-cache-dir .
0 | pip install '.[doc,tests]'
0 | cat docs/source/conf.py
None | python -m sphinx -T -b html -d _build/doctrees -D language=en . $READTHEDOCS_OUTPUT/html
```

Both install steps exit 0 under the new bounds, so the metadata resolves fine on RTD. The sphinx step has no exit code and no captured output, and the build length is 902 seconds against RTD's 900-second cap, so it was killed on the time limit. For comparison, #3700's builds came in at 595s and 569s, already close to the ceiling; this branch merges #3698 and #3700 on top, and the merged tree tips over.

The diff here is `setup.cfg`, one CI job, one test file and the sweep state CSV. Nothing sphinx reads. This is the docs build-time problem, which the GitHub Actions job papers over with `-j auto` and RTD cannot.

PR #3706, opened 2026-08-16
Clear the pathfinding LOW backlog: perf nits, empty-raster error, test and bench gaps

### LINKED ISSUE #3705: pathfinding: LOW-severity cleanup backlog from the 2026-07-08 sweep round
**Is your feature request related to a problem?**

The 2026-07-08 sweep round over `xrspatial/pathfinding.py` (issues #3644, #3645, #3649, #3651, #3652, #3655, #3657, #3660) fixed everything MEDIUM and above but left a trail of LOW findings in the sweep notes. None of them justifies a PR on its own, so this bundles them into one cleanup pass. Every item below was re-verified against current `main` on 2026-08-16. Items already fixed since the sweeps (TSP solver unit tests, anisotropic-cellsize golden tests) are not listed.

**Performance nits** (documented in the #3660 sweep, not fixed)

- `_nearest_neighbor_2opt` calls `_tour_cost` twice per 2-opt candidate, recomputing the full tour sum each time instead of the O(1) delta for a segment reversal. Only matters for hundreds of waypoints (`_MAX_WAYPOINTS` is 1000).
- `_a_star_dask` re-fetches the popped node's friction value (`f_u_val`, invariant across the loop) inside the 8-neighbor loop (`pathfinding.py:752`). Hoist above the loop.
- `parent_ys`/`parent_xs` are built with `np.ones((h, w), dtype=np.int64) * NONE` (`pathfinding.py:286-287`); `np.full` avoids the multiply pass.
- The dask start/goal crossability check runs two separate scalar `.compute()` calls (`pathfinding.py:1069,1072`); one `dask.compute(start_val, goal_val)` halves the scheduler round-trips.
- `f_min` for dask friction is recomputed with a full `nanmin` scan per `a_star_search` call (`pathfinding.py:1130`), so an N-waypoint `multi_stop_search` pays N-1 redundant scans. Flagged as a follow-up in #3660.

**Error message** (documented in the #3649 sweep, not fixed)

- A (0, 0)-size raster fails with `ValueError: numpy.nanmin raises on a.size==0 and axis=None; So Bottleneck too.` out of the resolution/f_min machinery. A shape check in `_validate_surface_dims` with a message naming the function and the empty input would match the other validators added in #3649. Reproduced on current main.

**Test gaps** (the "documented but not fixed" list from #3657)

- `_ChunkCache` eviction: no test exercises the `maxsize` overflow path, so the LRU `popitem` branch has zero coverage.
- `_held_karp` n==2 shortcut (`pathfinding.py:1300-1301`): the TSP unit tests added since use 3-4 cities only.
- The 2-opt improvement branch: `test_nearest_neighbor_2opt_vs_held_karp`'s distance matrix never produces an improving swap (nearest-neighbor already returns the 2-opt-stable tour), so `improved = True` is never hit.
- `_get_pixel_id` default `xdim`/`ydim` arguments (`pathfinding.py:103-106`): every caller in the test suite passes explicit dims.

**Benchmark gaps** (documented in the #3645 sweep, not fixed)

- The asv pathfinding benchmarks route an open grid with no barriers and no friction, which is A*'s best case; regressions in barrier/friction handling or frontier growth won't show. Add a maze or barrier-field variant.
- `dask+cupy` is not a parameterizable backend anywhere in the suite (`benchmarks/benchmarks/common.py` `get_xr_dataarray` has no such type). This is suite-wide rather than pathfinding-specific; fixing it here would benefit every benchmark, but it can also be split out if it grows.

**Describe the solution you'd like**

One cleanup PR (or one PR plus a split-out for the suite-wide `dask+cupy` benchmark type):

- the five small performance fixes, none of which change results
- a zero-size raster validation error
- four targeted tests for the uncovered branches
- a barrier/friction benchmark variant

Excluded on purpose: the `snap_start`/`snap_goal` vs `snap` naming split (intentional per #3644, per-endpoint vs per-waypoint semantics) and the library-wide `surface`/`raster`/`agg` first-parameter naming (cross-cutting, tracked in #3644's notes).

Found by /sweeps-report rollup of the 2026-07-08 sweep round; re-verified 2026-08-16.

brendancol:
Closes #3705.

Works through the bundled LOW backlog from the 2026-07-08 sweep round in one pass.

**Performance**
- `_nearest_neighbor_2opt` now scores each 2-opt candidate in O(1) from prefix sums over the current tour's forward and backward edge costs, instead of re-summing the whole tour twice per candidate. Infinite edges (unreachable waypoint pairs) are tracked as counts next to the finite sums, since subtracting two prefix sums that both contain inf gives nan and would silently reject valid moves. One behavior change in the degenerate case: a tour stuck with an unavoidable inf edge previously could never improve (`inf < inf` is always false); it now still applies finite local improvements. Termination holds because each accepted move strictly decreases (inf-edge count, finite cost) lexicographically.
- `_a_star_dask` fetches the popped node's friction value once per node instead of once per neighbor.
- The dask start/goal crossability checks run as a single `dask.compute` instead of two scheduler round-trips.
- Dask friction `f_min` is cached in a small LRU keyed by the dask array's graph token, so an N-waypoint `multi_stop_search` scans the friction raster once instead of N-1 times (the follow-up noted in #3660). The token is content-derived, so a different or rebuilt raster gets a fresh entry.
- The numba kernel's parent index arrays use `np.full` instead of `np.ones * NONE`.

**Error message**
- A zero-size surface now raises `ValueError: ... surface is empty (shape (0, 0))` from `_validate_surface_dims` in both public functions, instead of leaking `numpy.nanmin raises on a.size==0` from the resolution machinery.

**Tests** (11 new, file went from 141 to 152 passing locally on a CUDA host)
- `_ChunkCache` LRU eviction and refresh-on-hit
- `_held_karp` n==2 shortcut, both directions
- 2-opt improvement branch: a hand-built matrix where greedy nearest-neighbor is wrong and one reversal fixes it; seeded asymmetric matrices asserting the returned cost matches the recomputed tour cost and never exceeds the greedy tour; an inf-edge case pinning the nan-poisoning guard
- `_get_pixel_id` default dim arguments
- empty surfaces for `a_star_search` and `multi_stop_search`
- `f_min` cache behavior plus a dask-vs-numpy `multi_stop_search` friction parity test that asserts all three segments resolve through one cache entry

**Review follow-up**: the posted review's suggestion (assert cache reuse across segments) and docstring nit are applied; the pre-existing `common.py` guard bug is filed as #3707.

**Benchmarks**
- New `AStarSearchObstacles`: serpentine barrier walls plus a friction gradient, since the existing open-grid benchmark is A*'s best case and hides regressions in barrier handling and frontier growth.
- `common.get_xr_dataarray` gains a `dask+cupy` type (verified cupy-backed via `is_dask_cupy`), wired into `AStarSearch`. Other benchmark classes can opt in later.

Not fixed here, spotted next to the change: `common.py`'s cupy branch tests `if not has_cuda_and_cupy:` without calling the function, so it never raises `NotImplementedError` on non-GPU hosts and falls through to an ImportError instead. Pre-existing, left alone to keep the diff scoped.

Backend coverage: no dispatch changes; numpy/cupy/dask+numpy/dask+cupy all exercised by the existing suite plus the new tests, run locally on a CUDA host.

Test plan:
- [x] `pytest xrspatial/tests/test_pathfinding.py` — 152 passed (CUDA host, GPU tests executed)
- [x] New benchmark classes instantiated and run once per backend, including `dask+cupy`
- [x] flake8/isort clean on the added code (pre-existing findings in the test file untouched)

### DIFF
--- a/benchmarks/benchmarks/common.py
+++ b/benchmarks/benchmarks/common.py
@@ -58,6 +58,13 @@ def get_xr_dataarray(
     elif type == "dask":
         import dask.array as da
         z = da.from_array(z, chunks=(max(1, ny // 2), max(1, nx // 2)))
+    elif type == "dask+cupy":
+        if not has_cuda_and_cupy():
+            raise NotImplementedError()
+        import cupy
+        import dask.array as da
+        z = da.from_array(
+            cupy.asarray(z), chunks=(max(1, ny // 2), max(1, nx // 2)))
     else:
         raise RuntimeError(f"Unrecognised type {type}")
 
--- a/benchmarks/benchmarks/pathfinding.py
+++ b/benchmarks/benchmarks/pathfinding.py
@@ -2,6 +2,7 @@ import numpy as np
 import xarray as xr
 
 from xrspatial.pathfinding import a_star_search, multi_stop_search
+from xrspatial.utils import has_cuda_and_cupy
 
 from .common import get_xr_dataarray
 
@@ -36,12 +37,12 @@ class MultiStopSearchDaskMemory:
 
 
 class AStarSearch:
-    params = ([100, 300, 1000], [4, 8], ["numpy", "cupy", "dask"])
+    params = ([100, 300, 1000], [4, 8], ["numpy", "cupy", "dask", "dask+cupy"])
     param_names = ("nx", "connectivity", "type")
 
     def setup(self, nx, connectivity, type):
-        if type == "dask" and nx > 300:
-            # The dask backend is a pure-Python sparse A* that loads
+        if type in ("dask", "dask+cupy") and nx > 300:
+            # The dask backends run a pure-Python sparse A* that loads
             # chunks on demand; at nx=1000 a single call takes ~4 s,
             # which would dominate the suite's runtime.
             raise NotImplementedError()
@@ -50,7 +51,7 @@ class AStarSearch:
         self.start = self.agg.y[0], self.agg.x[0]
         self.goal = self.agg.y[-1], self.agg.x[-1]
         # snap_start/snap_goal raise on dask-backed arrays by design
-        self.snap = type != "dask"
+        self.snap = type not in ("dask", "dask+cupy")
 
     def time_a_star_search(self, nx, connectivity, type):
         a_star_search(
@@ -58,6 +59,64 @@ class AStarSearch:
             connectivity=connectivity,
             snap_start=self.snap, snap_goal=self.snap
         )
+
+
+class AStarSearchObstacles:
+    """A* through a barrier field with friction (issue #3705).
+
+    The open-grid AStarSearch benchmark is A*'s best case: the heuristic
+    walks almost straight to the goal.  Walls with alternating gaps force
+    a serpentine route, so the frontier and visited set actually grow,
+    and the friction surface exercises the friction-weighted cost path.
+    Regressions in barrier handling or frontier bookkeeping are invisible
+    to the open-grid benchmark but show up here.
+    """
+
+    params = ([100, 300], ["numpy", "cupy", "dask"])
+    param_names = ("nx", "type")
+
+    def setup(self, nx, type):
+        ny = nx // 2
+        z = np.ones((ny, nx), dtype=np.float64)
+        # Vertical walls (value 0) with a one-cell gap alternating
+        # between the top and bottom row
+        step = max(2, nx // 10)
+        for wi, col in enumerate(range(step, nx - 1, step)):
+            z[:, col] = 0.0
+            if wi % 2 == 0:
+                z[0, col] = 1.0
+            else:
+                z[-1, col] = 1.0
+        # Left-to-right friction gradient
+        f = 1.0 + np.linspace(0.0, 4.0, nx)[np.newaxis, :] * np.ones(
+            (ny, 1), dtype=np.float64)
+
+        if type == "cupy":
+            if not has_cuda_and_cupy():
+                raise NotImplementedError()
+            import cupy
+            z = cupy.asarray(z)
+            f = cupy.asarray(f)
+        elif type == "dask":
+            import dask.array as da
+            chunks = (max(1, ny // 2), max(1, nx // 2))
+            z = da.from_array(z, chunks=chunks)
+            f = da.from_array(f, chunks=chunks)
+
+        y = np.linspace(ny - 1, 0, ny)
+        x = np.linspace(0, nx - 1, nx)
+        self.agg = xr.DataArray(z, coords=dict(y=y, x=x), dims=["y", "x"])
+        self.friction = xr.DataArray(
+            f, coords=dict(y=y, x=x), dims=["y", "x"])
+        self.start = float(y[-1]), float(x[0])
+        self.goal = float(y[0]), float(x[-1])
+
+    def time_a_star_search_barriers(self, nx, type):
+        a_star_search(self.agg, self.start, self.goal, barriers=[0])
+
+    def time_a_star_search_barriers_friction(self, nx, type):
+        a_star_search(self.agg, self.start, self.goal, barriers=[0],
+                      friction=self.friction)
 
 
 class MultiStopSearch:
--- a/xrspatial/pathfinding.py
+++ b/xrspatial/pathfinding.py
@@ -35,6 +35,11 @@ def _validate_surface_dims(surface, x, y, func_name):
             f"got {surface.dims}. Pass the actual dimension names via the "
             f"`x=` and `y=` parameters."
         )
+    if 0 in surface.shape:
+        raise ValueError(
+            f"{func_name}(): `surface` is empty (shape {surface.shape}); "
+            f"pathfinding requires at least one cell."
+        )
 
 
 def _validate_barriers(barriers):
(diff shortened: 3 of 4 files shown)


brendancol (commented):
## PR Review: Clear the pathfinding LOW backlog

None.

- [ ] `xrspatial/tests/test_pathfinding.py` (`test_multi_stop_dask_friction_matches_numpy`): the test proves dask/numpy parity but never asserts the f_min cache was actually hit across segments, which is the point of routing 3 segments through one friction raster. Clearing `_dask_fmin_cache` before the dask run and asserting `len(_dask_fmin_cache) == 1` afterward would pin the cross-segment reuse; today only the graph-token unit test covers it, and a future refactor that silently changes the token per call (e.g. rechunk to different chunks) would regress the optimization with no test failing.

- [ ] `xrspatial/pathfinding.py:1444` (`_nearest_neighbor_2opt`): the behavior change for inf-edge tours (previously frozen because `inf < inf` never accepted a move; now finite local improvements still apply) lives in a code comment and the PR body. One sentence in the function docstring would make it visible to the next reader without archaeology.
- [ ] `benchmarks/benchmarks/common.py:49`: the pre-existing cupy branch tests `if not has_cuda_and_cupy:` without calling the function, so it can never raise `NotImplementedError` and a non-GPU host gets an ImportError instead. The PR body flags it and correctly calls the function in the new `dask+cupy` branch; fine to leave for a separate fix, noting it here for the trail.

### What looks good

- The 2-opt rewrite handles the case a naive prefix-sum delta gets wrong: with unreachable waypoint pairs, `inf - inf = nan` would silently reject valid moves. The inf-count-plus-finite-sum bookkeeping keeps range sums exact, and the inf-edge test pins exactly that failure mode. Termination still holds since each accepted move strictly decreases (inf-edge count, finite cost) lexicographically.
- The asymmetric consistency test asserts the returned cost equals the cost recomputed from the returned order, which is the invariant a delta-bookkeeping bug would break, and it replicates the greedy construction to assert 2-opt never worsens it.
- The `f_u_val` hoist in `_a_star_dask` is safe: the goal check returns before the fetch, and any node in the heap already passed the positive-finite friction check when it was enqueued as a neighbor (or is the start, checked before the loop).
- The f_min cache key is the dask graph token, which is content-derived for both `da.from_array` (numpy friction) and `rechunk` (dask friction), so a rebuilt or different raster gets a fresh entry rather than a stale hit. Capped at 16 entries with LRU eviction.
- The empty-surface check sits in `_validate_surface_dims`, which both public functions call before any resolution math, so `(0, 0)`, `(0, N)`, and `(N, 0)` all get the clear message; all three shapes are parametrized in the test.
- `AStarSearchObstacles` fixes a real blind spot: the open-grid benchmark lets the heuristic walk nearly straight to the goal, so frontier-growth regressions were invisible. The serpentine wall layout forces exploration, and the friction variant exercises the weighted-cost path.
- The new `dask+cupy` benchmark type was verified genuinely cupy-backed (`is_dask_cupy` true, cupy `_meta`), not silently coerced to numpy blocks by `from_array`.

### Checklist

- [x] Algorithm matches reference: 2-opt/Held-Karp semantics preserved; delta scoring verified equal to full recomputation by test
- [x] All implemented backends produce consistent results: 152 existing + 10 new tests pass on a CUDA host (cupy and dask+cupy executed)
- [x] NaN handling is correct: no changes to NaN semantics; inf handling in the TSP path improved and tested
- [x] Edge cases are covered by tests: empty surfaces, two-city TSP, inf edges, cache eviction, default dims
- [x] Dask chunk boundaries handled correctly: no chunking changes; crossability reads batched into one compute
- [x] No premature materialization: the batched compute and f_min cache reduce materialization; nothing new added
- [x] Benchmark exists: two new benchmark surfaces added
- [x] README feature matrix: not applicable, no new public functions or backend-support changes
- [x] Docstrings: no public signatures changed; new private helpers have docstrings

No blockers. The one suggestion is a cheap test hardening; the nits are optional.

brendancol (commented):
## PR Review: follow-up on 73b9a18e

Re-review of the changes since the first review.

None.

### Suggestions / Nits

None new.

### Disposition of the first review's findings

- Suggestion (cache-reuse assertion): **fixed**. `test_multi_stop_dask_friction_matches_numpy` now clears `_dask_fmin_cache` before the dask run and asserts exactly one entry after routing three segments, which pins the cross-segment reuse the optimization exists for. A future change that varies the graph token per call now fails this test.
- Nit (inf-edge docstring): **fixed**. `_nearest_neighbor_2opt`'s docstring states that inf-edge tours still receive finite local improvements and keep an inf total.
- Nit (pre-existing `common.py` guard bug): **deferred** to #3707, as the first review itself recommended.

### One note on the commit's diff shape

The commit shows ~195 removed/re-added lines in `test_pathfinding.py` that are pure line-ending normalization: the file is CRLF on `main`, the first commit appended the new tests with LF endings, and this commit's edit rewrote them as CRLF. The net PR diff against `main` is 202 pure additions with zero deletions and consistent CRLF, so nothing to fix; noting it so the per-commit history doesn't read as a rewrite.

Full pathfinding suite re-run after the changes: 152 passed on a CUDA host. No blockers; the PR is in good shape from this reviewer's side.

Issue #3713, opened 2026-08-16
surface_direction() returns wrong bearings on dask rasters: block-local pixel indices subtracted from global source indices


brendancol:
## What happens

`surface_direction()` returns wrong compass bearings when the input is a
dask-backed DataArray and the run takes the iterative tile-Dijkstra path
(the default, since `max_distance` defaults to `np.inf`). The numpy and
cupy backends are correct; so is the bounded dask path that goes through
`map_overlap`.

On a 12x12 raster with two sources and `chunks=(6, 6)`, 105 of 144 pixels
disagree with the numpy answer.

## Why

`_run_tile()` records **global** source indices:

```python
src_row[r, c] = r + row_offset
src_col[r, c] = c + col_offset
```
(`xrspatial/surface_distance.py:963-964`)

`_finalize_direction()` then subtracts **block-local** pixel indices:

```python
row_idx, col_idx = np.meshgrid(np.arange(H), np.arange(W), indexing='ij')
dx = (src_col.astype(np.float64) - col_idx) * cellsize_x
dy = (src_row.astype(np.float64) - row_idx) * cellsize_y
```
(`xrspatial/surface_distance.py:366-370`, reached from `_assemble_sd` ->
`_extract_output` at `:1177`)

`H, W` there are the chunk's shape, so every chunk except the top-left one
measures the bearing from the wrong origin. The offset error equals the
chunk's `(row_offset, col_offset)`.

The eager numpy path is unaffected because `_surface_distance_numpy()`
seeds `src_row`/`src_col` through `_seed_sources()`, which stores plain
`r`/`c` — global and local coincide when there is one block. The bounded
dask path is unaffected for the same reason: `_make_sd_chunk_func()` calls
`_surface_distance_numpy()` on the padded block, so both sides of the
subtraction are block-local.

## Reproduction

```python
import warnings
import numpy as np
import xarray as xr
import dask.array as da

from xrspatial import surface_direction

source = np.zeros((3, 3))
source[1, 1] = 1.0
elevation = np.zeros((3, 3))


def _da(arr, chunks=None):
    n, m = arr.shape
    d = xr.DataArray(arr, dims=['y', 'x'])
    d['y'] = np.arange(n)[::-1]
    d['x'] = np.arange(m)
    if chunks is not None:
        d.data = da.from_array(d.data, chunks=chunks)
    return d


print(surface_direction(_da(source), _da(elevation)).values)
with warnings.catch_warnings():
[...]

PR #3718, opened 2026-08-16
Accept target_values=None in the surface_* trio like the proximity trio

### LINKED ISSUE #3712: surface_distance trio rejects target_values=None while proximity, allocation, direction and cost_distance accept it
`surface_distance`, `surface_allocation` and `surface_direction` declare `target_values: list = []`. Every other function in the same family declares `target_values: list = None` and normalizes it to `[]` in the body:

| function | file | default |
|---|---|---|
| `proximity` | `xrspatial/proximity.py:1607` | `None` |
| `allocation` | `xrspatial/proximity.py:1767` | `None` |
| `direction` | `xrspatial/proximity.py:1928` | `None` |
| `cost_distance` | `xrspatial/cost_distance.py:1219` | `None` |
| `surface_distance` | `xrspatial/surface_distance.py:1424` | `[]` |
| `surface_allocation` | `xrspatial/surface_distance.py:1485` | `[]` |
| `surface_direction` | `xrspatial/surface_distance.py:1527` | `[]` |

Two consequences.

The first is that passing `target_values=None` explodes. The proximity trio and `cost_distance` accept it because they do `if target_values is None: target_values = []`. `surface_distance` has no such guard, so `None` reaches `np.asarray(None, dtype=np.float64)`, which produces a 0-d array, which is then handed to the `_seed_sources` numba kernel and dies in type inference. The traceback runs about 40 lines about `getitem(array(float64, 0d, C), int64)` and never mentions `target_values`.

That is not a contrived call. Wrapper code routinely threads `target_values` through from its own optional argument, which is `None` when the caller did not set it. The pattern works against four functions in this family and fails against three.

The second is that `[]` as a default is a mutable default argument. Nothing mutates it today, since `_compute` immediately does `np.asarray`, so there is no live aliasing bug. It is still the anti-pattern the sibling functions already avoid.

## Reproduction

```python
import numpy as np
import xarray as xr
from xrspatial import (proximity, allocation, direction,
                       surface_distance, surface_allocation, surface_direction)

H = W = 8
src = np.zeros((H, W)); src[0, 0] = 1.0
elev = np.zeros((H, W))
coords = {'y': np.arange(H)[::-1].astype(float), 'x': np.arange(W).astype(float)}
raster = xr.DataArray(src, dims=['y', 'x'], coords=coords)
elevation = xr.DataArray(elev, dims=['y', 'x'], coords=coords)

for name, fn, args in [
    ("proximity", proximity, (raster,)),
    ("allocation", allocation, (raster,)),
    ("direction", direction, (raster,)),
    ("cost_distance", cost_distance, (raster, elevation)),
    ("surface_distance", surface_distance, (raster, elevation)),
    ("surface_allocation", surface_allocation, (raster, elevation)),
    ("surface_direction", surface_direction, (raster, elevation)),
]:
    try:
        fn(*args, target_values=None)
        print(f"{name}(target_values=None) -> OK")
    except Exception as e:
        print(f"{name}(target_values=None) -> {type(e).__name__}")
```

Observed on this host (Python 3.14.2, numba 0.65.0, numpy 2.3.2):

```
proximity(target_values=None) -> OK
allocation(target_values=None) -> OK
direction(target_values=None) -> OK
cost_distance(target_values=None) -> OK
surface_distance(target_values=None) -> TypingError
surface_allocation(target_values=None) -> TypingError
surface_direction(target_values=None) -> TypingError
```

The `TypingError` body:

```
No implementation of function Function(<built-in function getitem>) found for signature:
 >>> getitem(array(float64, 0d, C), int64)
...
File "xrspatial/surface_distance.py", line 193:
                for k in range(n_values):
                    if val == target_values[k]:
                    ^
```

## Proposed fix

Change the three signatures to `target_values: list = None` and add the same `if target_values is None: target_values = []` normalization the siblings use. This is not a breaking change: `[]` and `None` both mean "treat every non-zero finite pixel as a source", callers passing an explicit list are unaffected, and callers passing `None` go from a numba traceback to working code. No deprecation shim is needed because no parameter is renamed and no accepted input changes meaning.

## Notes

Found by `/sweep-api-consistency` on `surface_distance` (category 1 and 4, MEDIUM).

`balanced_allocation` (`xrspatial/balanced_allocation.py`) carries the same `target_values: list = []` default. Out of scope for this issue, but worth a separate look to see whether it fails the same way.

brendancol:
Closes #3712

`surface_distance`, `surface_allocation` and `surface_direction` declared `target_values: list = []`. The four sibling functions in the same family (`proximity`, `allocation`, `direction`, `cost_distance`) declare `None` and normalize it to `[]` in the body, so `target_values=None` worked against four of the seven and raised a 40-line numba `TypingError` against the other three.

- Change the three defaults to `None` and normalize once in `_compute`.
- Update the `target_values` docstring so it matches the new default.

Not a breaking change. `None` and `[]` both mean "treat every non-zero finite pixel as a source", so callers passing an explicit list see no difference and callers passing `None` stop getting a traceback. No parameter was renamed, so no deprecation shim is needed.

## Backend coverage

The normalization sits in `_compute`, ahead of the backend dispatch, so all four backends are covered. The new parity test runs against numpy, cupy, dask+numpy and dask+cupy; it was run on a CUDA host, so the two GPU variants executed rather than skipped.

## Test plan

- [x] `test_target_values_none_matches_empty_list` — `None` and `[]` produce identical output for all three functions on all four backends (12 cases)
- [x] `test_target_values_default_is_not_mutable` — the public signatures no longer carry a mutable default
- [x] `pytest xrspatial/tests/test_surface_distance.py` — 50 passed (37 before, 13 new), GPU tests included
- [x] `flake8` clean on `xrspatial/surface_distance.py`

Found by `/sweep-api-consistency` on the `surface_distance` module.

One note on something this PR does not touch: `xrspatial/tests/test_surface_distance.py:234` has a pre-existing `F841` (unused `sd` in `test_allocation_consistency`). It is unrelated to this change, so I left it alone.


brendancol (commented):
## PR Review: Accept target_values=None in the surface_* trio like the proximity trio

Small, well-targeted change. The normalization lands in `_compute` (`xrspatial/surface_distance.py:1300`), which is the single dispatcher ahead of the numpy / cupy / dask+numpy / dask+cupy branch, so one guard genuinely covers all four backends and the two accessor classes, which forward `**kwargs`. No numeric path is touched.

Three things to tighten before merge, none of them blocking correctness.

None.

- `xrspatial/surface_distance.py:1426`, `:1487`, `:1529` -- the annotation is now `target_values: list = None`, which the default contradicts. Before this PR `list = []` was at least type-correct; the PR is what makes it wrong, so it should carry the fix. The file already has `from __future__ import annotations` (line 28) and the repo already uses PEP 604 in `xrspatial/hydro/stream_order_d8.py` (`method: str | None = None`), so `target_values: list | None = None` costs nothing and needs no import. `proximity` and `cost_distance` have the same wrong annotation, but copying their mistake is not what "match the siblings" should mean here.

- `xrspatial/tests/test_surface_distance.py:362` -- the parity test only exercises one of the two dask branches. `_make_raster` defaults to `chunks=(3, 3)` and `res=1.0`, so `max_distance=2.0` gives `pad = int(2.0 / 1.0) + 1 = 3`, which fails the `pad < max(chunks_y) and pad < max(chunks_x)` test at `surface_distance.py:1206` and routes to `_sd_dask_iterative`. The bounded `map_overlap` route in `_surface_distance_dask_bounded` never runs. Parametrizing `max_distance` over a bounded value below 2.0 and `np.inf` would cover both.

- `xrspatial/tests/test_surface_distance.py:373-375` -- the fixture is thin. With that elevation ramp (0.5 per cell, so a cardinal step costs `sqrt(1 + 0.25) = 1.118`) and `max_distance=2.0`, only 4 of the 36 pixels come back finite; the other 32 comparisons are NaN against NaN. It does assert something real, but a wider budget would make the None-vs-`[]` claim much harder to satisfy by accident.

- `test_target_values_default_is_not_mutable` imports `inspect` inside the function body. That matches the surrounding file, which does the same for `cost_distance` and `unittest.mock`, so this is fine as-is; noting it only so the next reader does not flag it.

### What looks good

- The guard sits before `_validate_raster`, so `None` is resolved before anything else can trip over it. No per-backend duplication.
- Non-breaking, and correctly identified as such: `None` and `[]` both mean "every non-zero finite pixel is a source", so no deprecation shim is warranted. Good call not adding one.
- The docstring was updated alongside the default rather than left to drift, which is how this class of issue gets created in the first place.
- The issue carries a runnable reproduction with real observed output, and the same script now prints OK for all seven functions.

### Note on something this PR does not cause

While reading the dask path I saw `surface_direction` return different bearings on numpy `{0, 90, 270}` and dask+numpy `{0, 126.87, 135, 270}` for identical input. That is the block-local-vs-global source index bug in `_finalize_direction`, already filed as #3713 and #3719. It predates this PR and does not affect the new test, which only compares `None` against `[]` within a single backend. Flagging it so nobody blames this change for it.

### Checklist

- [x] Algorithm matches reference -- no algorithm changed
- [x] All implemented backends produce consistent results -- new test covers all four, executed on a CUDA host
- [x] NaN handling is correct -- unchanged
- [ ] Edge cases are covered by tests -- see the two test suggestions above
- [x] Dask chunk boundaries handled correctly -- unchanged
- [x] Benchmark exists or is not needed -- not needed, no perf change

brendancol (commented):
## PR Review (follow-up pass): Accept target_values=None in the surface_* trio

Re-reviewed after `ca17a3fa`. All three suggestions from the first pass are addressed.

### Disposition of the earlier findings

- Annotation contradicts the default -- **fixed**. All three signatures now read `target_values: list | None = None` (`xrspatial/surface_distance.py:1426`, `:1487`, `:1529`). Confirmed at runtime: `inspect.signature` reports `list | None` with default `None` for each. `proximity` and `cost_distance` still carry the older `list = None` spelling; that is their modules' problem, not this PR's.
- Only one dask route exercised -- **fixed**. `max_distance` is now parametrized over `1.5` and `np.inf` (`xrspatial/tests/test_surface_distance.py:366`). At `res=1.0` and `chunks=(3, 3)`, `1.5` gives `pad = 2`, which passes the `pad < max(chunks)` gate and routes through `_surface_distance_dask_bounded`; `np.inf` skips the finite check entirely and routes through `_sd_dask_iterative`. Both branches now run for all three functions.
- Thin fixture, mostly NaN-vs-NaN -- **fixed**, and better than asked. Rather than only widening the budget, the test now asserts `np.isfinite(with_none).any()` before comparing, so a future change that quietly turns the output all-NaN fails instead of passing vacuously.
- The `import inspect` nit was raised as already-fine and needed no change.

### Blockers

None.

### Suggestions

None.

### Nits

None.

### Verification

- `pytest xrspatial/tests/test_surface_distance.py` -- 62 passed, up from 37 on `main`. The 24 new parity cases (3 functions x 4 backends x 2 budgets) all ran; this is a CUDA host, so cupy and dask+cupy executed rather than skipped.
- `flake8 xrspatial/surface_distance.py` -- clean.
- The pre-existing `F841` at `xrspatial/tests/test_surface_distance.py:234` is untouched and predates this branch (confirmed against `HEAD`).

The `surface_direction` numpy-vs-dask bearing divergence noted in the first pass is still there and still belongs to #3713 / #3719, not here.

PR #3728, opened 2026-08-16
Grow the surface_distance Dijkstra heap on demand instead of capping it at height*width (#3723)

### LINKED ISSUE #3723: surface_distance Dijkstra heap is sized height*width and overflows on dense target rasters, corrupting memory
## Description

`surface_distance()`, `surface_allocation()` and `surface_direction()` can write past the end of their Dijkstra heap arrays and corrupt the process heap. On a 48x48 raster with 40% target pixels over rugged terrain the call aborts the interpreter with `double free or corruption (!prev)`.

`_dijkstra()` and `_dijkstra_geodesic()` in `xrspatial/surface_distance.py` size the heap at `height * width`:

```python
max_heap = height * width
h_keys = np.empty(max_heap, dtype=np.float64)
h_rows = np.empty(max_heap, dtype=np.int64)
h_cols = np.empty(max_heap, dtype=np.int64)
```

That is not the right bound. Both kernels use a lazy-deletion min-heap: a pixel is pushed again every time its tentative distance improves, and stale entries stay in the heap until they are popped and skipped by the `visited` check. Live heap occupancy is therefore bounded by the seed count plus the number of improving relaxations, not by the pixel count.

`_heap_push` is `@ngjit` with no bounds checking, so once `h_size` reaches `height * width` the next push writes into whatever allocation follows `h_keys` / `h_rows` / `h_cols`.

This is the same defect that was fixed in `cost_distance` (`xrspatial/cost_distance.py:161`), which now sizes its heap `height * width * (n_neighbors + 1)` with a comment explaining that the old `height * width` sizing "underflows that bound and let `_heap_push` write past the end of the arrays, corrupting memory". `surface_distance` was never updated.

## Affected backends

- numpy: `_surface_distance_numpy` calls `_dijkstra` / `_dijkstra_geodesic` directly.
- dask+numpy: the bounded `map_overlap` path calls `_surface_distance_numpy` per chunk, and `_run_tile` in the iterative path calls `_dijkstra` per tile.
- cupy and dask+cupy bounded are not affected; they use the parallel relaxation kernel, which has no heap.

## Reproduction

```python
import numpy as np
import xarray as xr
from xrspatial.surface_distance import surface_distance

N = 48
rng = np.random.default_rng(1)
src = np.zeros((N, N), dtype=np.float64)
src.flat[rng.choice(N * N, size=921, replace=False)] = 1.0   # 40% targets
elev = rng.random((N, N)) * 200.0                            # rugged relief

coords = {'y': np.arange(N, dtype=np.float64),
          'x': np.arange(N, dtype=np.float64)}
raster = xr.DataArray(src, dims=['y', 'x'], coords=coords,
                      attrs={'res': (1.0, 1.0)})
elevation = xr.DataArray(elev, dims=['y', 'x'], coords=coords,
                         attrs={'res': (1.0, 1.0)})

surface_distance(raster, elevation, connectivity=8)
```

Observed on this host (numba 0.62, Python 3.14):

```
$ python repro.py
double free or corruption (!prev)
Aborted
```

With bounds checking on, the out-of-bounds write is reported instead of corrupting memory:

```
$ NUMBA_BOUNDSCHECK=1 python repro.py
  File "xrspatial/surface_distance.py", line 474, in _surface_distance_numpy
    _dijkstra(elev_data, H, W, max_distance,
IndexError: index is out of bounds
```

## How close normal inputs get

Instrumenting `_dijkstra` to record peak heap occupancy against the `height * width` cap, over random scenes varying grid size, relief amplitude and target density:

| N | relief | targets | cap (H*W) | peak heap | peak / cap |
|---|--------|---------|-----------|-----------|------------|
| 48 | 200 | 46 (2%) | 2304 | 1314 | 0.57 |
| 48 | 200 | 115 (5%) | 2304 | 1697 | 0.74 |
| 48 | 200 | 230 (10%) | 2304 | 1940 | 0.84 |
| 48 | 200 | 460 (20%) | 2304 | 2259 | 0.98 |
| 48 | 200 | 921 (40%) | 2304 | 2519 | 1.09 -- overflow |

Total pushes are already over the cap at 2% target density (1.6x at the top of this table). Only the interleaved pops keep live occupancy under it in the lower rows, and that margin disappears as target density or relief goes up. Dense target rasters are not exotic here: a landcover raster where roads or built-up cells are the targets easily reaches 20-40%.

## Suggested fix

Grow the heap on demand instead of pre-allocating a fixed cap. Starting at `height * width` and doubling keeps the common case at today's footprint and removes the cap entirely.

The static bound `height * width * (n_neighbors + 1)` that `cost_distance` uses would also be correct, but for 8-connectivity it multiplies the heap allocation by 9, from 24 to 216 bytes per pixel. `surface_distance` documents an 80 bytes-per-pixel working set in `_BYTES_PER_PIXEL` and enforces it in `_check_memory`, so that constant would have to rise to roughly 272 and the guard would start refusing rasters about 3.4x smaller than it does today.

(Related, not fixed here: `cost_distance._BYTES_PER_PIXEL` is still 40, which was the pre-fix figure and no longer covers its own 9x heap.)

brendancol:
Closes #3723

`_dijkstra` and `_dijkstra_geodesic` sized their lazy-deletion min-heap at `height * width`. That bound is wrong: a pixel is pushed again every time its tentative distance improves, and stale entries stay in the heap until a pop skips them, so live occupancy tracks the number of improving relaxations rather than the pixel count. `_heap_push` is `@ngjit` with no bounds checking, so passing the cap writes into whatever allocation follows `h_keys` / `h_rows` / `h_cols`.

On a 48x48 raster with 40% target pixels over 200 m of relief, peak occupancy hit 1.09x the cap and the call aborted the interpreter with `double free or corruption (!prev)`. With `NUMBA_BOUNDSCHECK=1` the same input raises `IndexError: index is out of bounds` from `_surface_distance_numpy`.

## What changed

- The heap starts at one entry per pixel, which the seeding loop can never exceed, and `_heap_grow` doubles it when a relaxation would overflow.
- Regression tests for the numpy, dask iterative and geodesic paths, checking the dense-target scene against a pure-Python `heapq` Dijkstra reference.
- A `SurfaceDistanceDenseTargets` asv benchmark over 5%, 20% and 40% target density, with a `peakmem` variant. The existing `SurfaceDistance` benchmark cannot catch a regression here: its integer source raster makes nearly every pixel a target, so almost no relaxation ever improves a distance and the heap barely fills.

## Why not the cost_distance approach

`cost_distance` fixed this defect at `xrspatial/cost_distance.py:161` with a static `height * width * (n_neighbors + 1)` bound. That is also correct, but for 8-connectivity it multiplies the heap allocation by 9, from 24 to 216 bytes per pixel. This module documents an 80 bytes-per-pixel working set in `_BYTES_PER_PIXEL` and enforces it in `_check_memory`, so the constant would have to rise to roughly 272 and the guard would start refusing rasters about 3.4x smaller than it does today. Growing on demand keeps the documented footprint accurate and still removes the cap.

## Backends

- numpy: fixed. `_surface_distance_numpy` calls both kernels directly.
- dask+numpy: fixed. The bounded `map_overlap` path calls `_surface_distance_numpy` per chunk and `_run_tile` calls `_dijkstra` per tile.
- cupy and dask+cupy bounded: not affected. They use the parallel relaxation kernel, which has no heap. The dask+cupy unbounded path converts to numpy and picks up the fix.

## Test plan

- [x] `pytest xrspatial/tests/test_surface_distance.py` -- 41 passed (37 before, 4 new)
- [x] The `#3723` reproduction under `NUMBA_BOUNDSCHECK=1` returns a full finite grid instead of raising
- [x] The same reproduction without bounds checking no longer aborts
- [x] Dense-target result matches a pure-Python `heapq` Dijkstra to `rtol=1e-5`
- [x] `flake8` clean on the changed files (the one `F841` in the test file predates this branch)
- [x] `SurfaceDistanceDenseTargets` runs end to end at both grid sizes and all three densities


brendancol (commented):
## PR Review: Grow the surface_distance Dijkstra heap on demand instead of capping it at height*width (#3723)

Self-review from the /sweep-performance agent that wrote the patch, posted as the audit trail. Read it with that bias in mind.

None.

- [ ] `xrspatial/surface_distance.py:80-92` -- the `_BYTES_PER_PIXEL = 80` comment still itemizes `h_keys` / `h_rows` / `h_cols` at 8 bytes each, which was exact when the heap was a fixed `height * width`. It no longer is. After one doubling the heap is 48 bytes per pixel rather than 24, and `_heap_grow` allocates the replacement arrays before the originals go out of scope, so a growth event transiently holds three times the current heap. On the dense-target scenes this PR is about, `_check_memory` can therefore under-report peak by roughly 60%. The guard was advisory before and is not made worse by this change (the old code corrupted memory instead of over-allocating), but the constant should say the heap can grow so the next person reading it is not misled.

- [ ] `xrspatial/surface_distance.py:246-251` and the geodesic equivalent -- the seeding loop pushes without a capacity check. The invariant holding it up is real: the loop visits each pixel once and pushes at most once per pixel, into a heap sized `max(height * width, 1)`. It is stated in the comment at line 236. But it is exactly the kind of invariant that a later change to seeding would break silently, and the failure mode is the out-of-bounds write this PR exists to remove.

- [ ] `xrspatial/surface_distance.py:174-192` -- `_heap_grow`'s docstring is longer than its body, and the part that matters (why the old bound was wrong) belongs where the capacity is chosen, in `_dijkstra`. Line 236 has a short version already; consider making that the canonical explanation and trimming the helper to one line.

- [ ] `xrspatial/tests/test_surface_distance.py` -- none of the four new tests assert that the heap actually grew. They assert the call completes and matches a `heapq` reference, which is what a user cares about, but if someone later changed the growth trigger to fire only above, say, four times capacity, the tests would still pass while the bug returned for denser inputs. Checking that the pre-fix code fails all four (verified: `IndexError` under `NUMBA_BOUNDSCHECK=1`) is decent evidence the tests bite today; it says nothing about tomorrow.

### What looks good

- The bound in the issue is right, and the reasoning holds for both kernels. A lazy-deletion heap's live occupancy tracks improving relaxations, not pixels, and `_heap_push` has no bounds check because `@ngjit` does not add one.
- Rejecting `cost_distance`'s static `height * width * (n_neighbors + 1)` bound is the right call here. It is correct but multiplies the heap by 9 for 8-connectivity, and this module, unlike `cost_distance`, enforces a documented per-pixel budget that would have to move from 80 to about 272.
- The regression tests were checked against the pre-fix source and all four fail there, including the geodesic one, so the geodesic kernel is genuinely covered rather than just touched.
- The new benchmark targets the regime the old one structurally could not reach. The existing `SurfaceDistance` class feeds `get_xr_dataarray(is_int=True)`, which makes almost every pixel a target, so relaxations almost never improve anything and the heap never fills.
- Backend claims in the PR body check out: `_surface_distance_cupy` uses parallel relaxation with no heap, and the dask+cupy unbounded branch converts to numpy and inherits the fix.

### Coverage the PR does not claim, and does not have

- cupy and dask+cupy are untested for this scenario because there is nothing to test; neither allocates a heap.
- The bounded `map_overlap` dask branch is not covered by a dense-target test. It calls `_surface_distance_numpy` per chunk, so it is fixed, but only the iterative branch has a test.

### Checklist

- [x] Algorithm matches reference/paper -- output verified against a pure-Python `heapq` multi-source Dijkstra
- [x] All implemented backends produce consistent results -- dask iterative matches numpy at `rtol=1e-5`
- [x] NaN handling is correct -- untouched by this change; barrier and unreachable tests still pass
- [x] Edge cases are covered by tests -- dense targets at 40%, geodesic, dask tiles
- [x] Dask chunk boundaries handled correctly -- `_run_tile` seeds boundaries before `_dijkstra`, so seeds stay within one entry per pixel
- [x] No premature materialization or unnecessary copies -- the only new allocation is the heap doubling
- [x] Benchmark exists -- `SurfaceDistanceDenseTargets` added
- [x] README feature matrix updated -- not applicable, no new function and no backend change

brendancol (commented):
## PR Review: follow-up pass after 4afb223d

Second review pass over the same PR, covering the commit that responded to the first one.

### Disposition of the first pass

- Fixed: the `_BYTES_PER_PIXEL` comment now says the heap terms are a starting capacity rather than a bound, so 80 reads as a floor.
- Fixed: both seeding loops now check capacity before pushing, so every push site in both kernels is guarded. The invariant that made the unguarded version safe was real but sat only in a comment, and the failure mode it protected was the out-of-bounds write this PR removes.
- Fixed: the capacity rationale moved from `_heap_grow` into `_dijkstra` at the point the capacity is chosen; `_heap_grow` is back to a one-line docstring.
- Fixed: `test_dense_targets_dask_bounded_does_not_overflow` covers the bounded `map_overlap` branch at `max_distance=10.0` with 16x16 chunks, matching numpy at `rtol=1e-5`.
- Dismissed: asserting that the heap actually grew would mean reaching into `@ngjit` internals or exporting occupancy from the kernel, which costs more than it buys. The pre-fix check (all regression tests fail with `IndexError` under `NUMBA_BOUNDSCHECK=1`) stands as the evidence that the tests bite.

### Blockers

None.

### Suggestions

None outstanding.

### Nits

- [ ] The seeding guard is dead code on every input that exists today, since a per-pixel loop cannot exceed a per-pixel capacity. It is one comparison inside a loop that already does a `log n` sift, so the cost is not the concern; someone reading it may just wonder why it is there. The comment above `max_heap` explains the class of bug, which should be enough.

### Verification

- `pytest xrspatial/tests/test_surface_distance.py` -- 42 passed
- Same file under `NUMBA_BOUNDSCHECK=1` -- 42 passed
- `pytest xrspatial/tests/test_surface_distance.py xrspatial/tests/test_cost_distance.py` -- 133 passed, so the shared `_heap_push` / `_heap_pop` are unaffected
- `flake8` clean on `xrspatial/surface_distance.py` and `benchmarks/benchmarks/surface_distance.py`

PR #3731, opened 2026-08-16
Fix surface_direction bearings on north-up rasters and on chunked dask input

### LINKED ISSUE #3719: surface_direction() returns mirrored bearings on north-up rasters and per-chunk-local bearings on the dask iterative path
## Description

`surface_direction()` returns wrong compass bearings in two unrelated
situations. `surface_distance()` and `surface_allocation()` are correct in
both cases, so only the direction output is affected. Found by the accuracy
sweep against `xrspatial/surface_distance.py`.

### Bug 1: the bearing comes from row/column indices, not coordinates

`_finalize_direction()` (xrspatial/surface_distance.py:359-380) builds the
offset to the source pixel out of array indices scaled by an absolute cell
size:

```python
dx = (src_col.astype(np.float64) - col_idx) * cellsize_x
dy = (src_row.astype(np.float64) - row_idx) * cellsize_y
```

`_compute()` has already dropped the axis direction at line 1315-1316
(`cellsize_y = abs(float(cellsize_y))`), so the row index stands in for the y
coordinate. On a north-up raster, which is what every GeoTIFF this library
reads produces, y descends as the row index rises. Every bearing then comes
out mirrored across the east-west axis, and north reads as south.

`proximity.direction()` documents the identical compass convention and uses
real coordinate values (`_vectorized_calc_direction(query_x, target_x,
query_y, target_y)`, xrspatial/proximity.py:931). The two functions disagree
on the same input. Same class of defect as #2896, where geodesic slope/aspect
used index coordinates in place of lat/lon.

### Bug 2: the dask iterative path mixes global source indices with chunk-local pixel indices

`_run_tile()` records source positions as global indices at line 963-964
(`src_row[r, c] = r + row_offset`). `_assemble_sd()` hands those straight to
`_extract_output()` at line 1177, and `_finalize_direction()` builds its
`row_idx`/`col_idx` from `np.arange` over the tile shape. Global minus local
leaves the chunk offset sitting in the answer.

Every chunk except (0, 0) then reports the bearing it would have had as a
standalone raster. The output is visibly a tiled repeat of the first chunk.

This is the default path. `max_distance` defaults to `np.inf`, and
`_surface_distance_dask()` routes anything unbounded (or with an implied
radius larger than the chunk) to `_sd_dask_iterative()`. The bounded
`map_overlap` path keeps indices local on both sides and is unaffected.

## Reproduction

Both run on CPU, no GPU needed.

### Bug 1

```python
import numpy as np
import xarray as xr
from xrspatial.surface_distance import surface_direction
from xrspatial.proximity import direction

def make(data, y):
    return xr.DataArray(
        data.astype(float), dims=['y', 'x'],
        coords={'y': y, 'x': np.arange(data.shape[1], dtype=float)})

src = np.zeros((3, 3)); src[1, 1] = 1.0
elev = np.zeros((3, 3))
y = np.array([2.0, 1.0, 0.0])   # descending y, i.e. north-up

print(np.asarray(surface_direction(make(src, y), make(elev, y)).data))
print(np.asarray(direction(make(src, y)).data))
```

```
surface_direction:            proximity.direction:
[[135. 180. 225.]             [[ 45. 360. 315.]
 [ 90.   0. 270.]              [ 90.   0. 270.]
 [ 45. 360. 315.]]             [135. 180. 225.]]
```

Row 0 sits north of the source. `proximity.direction` says 360 (north),
`surface_direction` says 180 (south). Rerun the same call with an ascending y
coordinate and `surface_direction` prints byte-identical output. It never
looks at the coordinate.

### Bug 2

```python
from xrspatial.surface_distance import surface_direction

def make(d, chunks=None):
    a = xr.DataArray(
        d.astype(float), dims=['y', 'x'],
        coords={'y': np.arange(d.shape[0], dtype=float),
                'x': np.arange(d.shape[1], dtype=float)},
        attrs={'res': (1.0, 1.0)})
    if chunks:
        a.data = da.from_array(a.data, chunks=chunks)
    return a

src = np.zeros((8, 10)); src[2, 3] = 1.0
elev = np.random.default_rng(7).uniform(0, 5, (8, 10))

base = np.asarray(surface_direction(make(src), make(elev)).data)
dk = surface_direction(make(src, (4, 5)), make(elev, (4, 5))).data.compute()
```

```
numpy:                                     dask+numpy, 4x5 chunks:
[[123.7 135.  153.4 180.  206.6 225. ...   [[123.7 135.  153.4 180.  206.6 123.7 ...
 [108.4 116.6 135.  180.  225.  243.4 ...   [108.4 116.6 135.  180.  225.  108.4 ...
 [ 90.   90.   90.    0.  270.  270. ...    [ 90.   90.   90.    0.  270.   90.  ...
 ...                                        ...
 [ 31.   21.8  11.3 360.  348.7 338.2 ...   [ 71.6  63.4  45.  126.9 135.  113.2 ...

direction max|diff| vs numpy: 360.0, 58 of 80 pixels differ
distance  max|diff| vs numpy: 0.0
```

## Expected behaviour

`surface_direction()` should report the bearing using the raster's real y/x
axis orientation and agree with `proximity.direction()` on the same input.
The dask result should match numpy on every chunk, bounded or unbounded.

## Environment

xarray-spatial `main` @ 069fc13, Python 3.14, dask installed. Bug 1 affects
every backend (the cupy branch at line 602-616 repeats the same index math).
Bug 2 is specific to dask+numpy and dask+cupy on the unbounded path.

brendancol:
Closes #3719

`surface_direction()` was returning wrong bearings in two unrelated ways, and
fixing the second one uncovered a third. `surface_distance()` and
`surface_allocation()` were correct throughout.

## What changed

- `_finalize_direction()` measured the offset to the source in row/column
  indices scaled by an absolute cell size, so the row index stood in for y.
  On a north-up raster (y descending as the row index rises) every bearing
  came out mirrored across the east-west axis, and `surface_direction()`
  disagreed with `proximity.direction()` on identical input. `_compute()` now
  reads the sign of each axis off the coordinate and passes the backends
  signed cell sizes. Every edge-cost site either squares the cell size or
  takes `abs()` of it first, so distances and allocations are byte-identical
  to before.

- `_assemble_sd()` handed `_finalize_direction()` the global source indices
  that `_run_tile()` records, while the bearing was measured against the
  block's own `np.arange` grid. The chunk offset stayed in the answer and
  every chunk except (0, 0) reported the bearing it would have had as a
  standalone raster. It now rebases to block-local indices. This is the
  default path: `max_distance` defaults to `np.inf`, which routes to
  `_sd_dask_iterative()`.

- Rebasing broke the no-source mask, which tested `src_row < 0` — a source
  in the chunk above rebases to a legitimately negative local index. A finite
  in-budget distance always comes with an assigned source, so the mask now
  keys on `dist` alone. Direction ends up with the same NaN mask as distance
  and allocation, which is what the docstring already promised.

## Backends

numpy, cupy, dask+numpy, dask+cupy. All four checked against the numpy
baseline on a descending-y raster, bounded and unbounded, with zero NaN-mask
mismatches and zero numeric difference. The cupy eager path shares
`_finalize_direction`'s index math and was affected by the orientation bug
too; dask+cupy inherits the tiled fix through the CPU iterative path.

## Test plan

- [x] `pytest xrspatial/tests/test_surface_distance.py` — 46 passed (37 before)
- [x] `pytest xrspatial/tests/test_proximity.py xrspatial/tests/test_cost_distance.py` — 686 passed
- [x] `flake8` clean on both changed files (the F841 at
      `test_surface_distance.py:234` predates this branch and is untouched)
- [x] New: `test_direction_follows_y_axis_orientation` pins both axis
      directions; `test_direction_matches_proximity_direction` pins agreement
      with the sibling API; `test_dask_direction_matches_numpy` and
      `test_direction_nan_mask_matches_distance` cover bounded and unbounded
      dask; `test_cupy_direction_matches_numpy` closes the GPU gap
- [x] Checked out `main`'s `surface_distance.py` under the new tests:
      `test_direction_follows_y_axis_orientation`,
      `test_direction_matches_proximity_direction`, and both
      `test_dask_direction_matches_numpy` cases fail there and pass here.
      `test_direction_nan_mask_matches_distance` and
      `test_cupy_direction_matches_numpy` pass on `main` as well — they pin
      invariants this change had to preserve, and the mask one is what caught
      the third defect mid-fix.

Found by `/sweep-accuracy` against `xrspatial/surface_distance.py`. The state
CSV row for the module is in the second commit.


brendancol (commented):
## PR Review: Fix surface_direction bearings on north-up rasters and on chunked dask input

Reviewed the full `surface_distance.py` and `test_surface_distance.py` at the
head of `deep-sweep-accuracy-surface_distance-2026-08-16-01`, not just the
diff. All three defects are real and the reproductions in #3719 replay against
`main`. Distances and allocations are untouched, which the existing 37 tests
confirm.

None.

- **`xrspatial/surface_distance.py:642` — the cupy DIRECTION branch still
  masks on `srow < 0`.** `_finalize_direction` dropped that clause at line
  399-406 because a rebased block-local index can legitimately be negative.
  The eager cupy path never rebases, so `srow` is always non-negative wherever
  `dist` is finite and the two masks agree today. But leaving one backend
  keying on the source index and the other on `dist` is the asymmetry that
  produced this bug in the first place. Make it `cp.isinf(dist) | (dist >
  max_distance)` so both paths state the same rule.

- **`xrspatial/surface_distance.py:1355-1359` — the signed cell size relies on
  an unwritten invariant.** Passing signed values through the existing
  `cellsize_x` / `cellsize_y` parameters is correct only because every other
  consumer squares them (`_compute_seeds_sd` builds `diag_dist` from
  `cellsize_x ** 2 + cellsize_y ** 2`, `_edge_seeds` uses `cardinal_dist **
  2`) or takes `abs()` first (`_surface_distance_dask`,
  `_surface_distance_dask_bounded`, `_surface_distance_dask_cupy`). I walked
  every call site and the invariant holds, but nothing enforces it, and the
  next person to add a cell-size consumer has no way to know. A test asserting
  that a descending-y raster produces the same DISTANCE and ALLOCATION output
  as an ascending-y one would pin it cheaply.

- **`xrspatial/surface_distance.py:344` — `_coord_step_sign` reads the span,
  not a step.** `values[-1] - values[0]` gives the right answer for any
  monotonic axis and matches how `calc_res` already derives resolution, but
  `surface_distance` never checks monotonicity the way
  `proximity._check_monotonic_coords` does. On a scrambled coordinate the sign
  is whatever the endpoints happen to say. Not worth a validation pass, but
  worth a word in the docstring that the axis is assumed monotonic.

- **Geodesic direction is still index-scaled.** In `method='geodesic'` mode
  `_finalize_direction` scales index offsets by a cell size in degrees, so
  bearings drift away from the equator where a degree of longitude is shorter
  than a degree of latitude. This PR fixes the sign but not the scale. It
  predates the change and is not made worse by it; worth a separate issue
  rather than growing this diff.

### What looks good

- The three defects are separated cleanly, and the commit message explains how
  fixing the second exposed the third rather than presenting it as one insight.
- Distances and allocations come out byte-identical, which is the right
  outcome for a direction-only fix and is what makes the signed-cell-size
  approach defensible.
- `test_direction_matches_proximity_direction` pins agreement with the sibling
  API instead of hard-coding numbers, so the two functions cannot drift apart
  again silently.
- `test_direction_nan_mask_matches_distance` covers bounded and unbounded dask
  and is what caught the third defect mid-fix. Keeping it in the suite is more
  valuable than the bearing assertions.
- The verification ran all four backends on a descending-y raster rather than
  reasoning about the GPU path from source.

### Checklist

- [x] Algorithm matches reference — bearings now agree with
      `proximity.direction`, the in-repo reference for this convention
- [x] All implemented backends produce consistent results — numpy, cupy,
      dask+numpy, dask+cupy checked bounded and unbounded, zero difference
- [x] NaN handling is correct — direction now shares the distance NaN mask
- [x] Edge cases covered — both axis directions, no-source raster, chunks
      whose source lives in a neighbouring chunk
- [x] Dask chunk boundaries handled correctly — the tiled path was the bug
- [x] No premature materialization — `srow - row_offset` is a per-block numpy
      op inside an existing `map_blocks` chunk function
- [x] Benchmark exists — `benchmarks/benchmarks/surface_distance.py`, and this
      change has no throughput effect
- [ ] README feature matrix — not applicable, no new function and no backend
      support change
- [x] Docstrings present and accurate — the `surface_direction` compass
      convention in the public docstring is now what the function returns

brendancol (commented):
## PR Review: follow-up pass after da98601

Re-read the changed hunks and re-ran the suite. Disposition of the first pass:

**Fixed**

- The cupy DIRECTION branch no longer masks on `srow < 0`. Both it and
  `_finalize_direction` now key on `dist`, so the two backends state the same
  rule and the GPU side cannot reintroduce the asymmetry.
- `test_distance_and_allocation_ignore_axis_direction` pins the invariant the
  signed cell size leans on: flip the y axis and distance and allocation come
  back unchanged. Parametrized over numpy and dask+numpy, and over
  `surface_distance` and `surface_allocation`. 50 tests pass (37 on `main`).
- `_coord_step_sign`'s docstring now says the axis is assumed monotonic and
  points at `calc_res` making the same assumption.

**Deferred, with a linked issue**

- Geodesic bearings still scale index offsets by a cell size in degrees, so
  they tilt with `cos(lat)` away from the equator. Filed as #3733. It predates
  this PR, and the fix needs a decision this one should not make: whether to
  scale the longitude leg by `cos(lat)` at the pixel, or compute a forward
  azimuth from the two lat/lon pairs, which changes what "bearing" means over
  a long path since a great circle's initial azimuth is not its final one.

No new findings. Nothing outstanding from my side.

PR #3732, opened 2026-08-16
Let the GPU surface-distance relaxation run to convergence

### LINKED ISSUE #3721: cupy surface_distance stops relaxing after H+W passes and reports reachable pixels as NaN
## Description

The cupy backend of `surface_distance()` solves the shortest-path problem by
repeated parallel relaxation (`_sd_relax_kernel`), and caps the number of
passes at `H + W`:

```python
# xrspatial/surface_distance.py:583
max_iterations = H + W
for _ in range(max_iterations):
    changed[0] = 0
    _sd_relax_kernel[griddim, blockdim](...)
    if int(changed[0]) == 0:
        break
```

One pass propagates distance information roughly one pixel hop. The number of
passes needed is therefore the hop count of the longest shortest path, not the
raster's width plus height. Those two numbers agree on open terrain, but they
come apart as soon as NaN barriers force paths to wind, and the worst case for
a grid graph is `H * W`, not `H + W`.

When the cap is reached the loop exits with no warning and no signal in the
output. Pixels the CPU backend reaches are reported as NaN, meaning
unreachable. A user reading the result cannot tell the difference between "no
path exists" and "we stopped looking."

This affects all three public functions on the cupy backend
(`surface_distance`, `surface_allocation`, `surface_direction`) and the
bounded dask+cupy path, where `_surface_distance_cupy` runs per chunk and the
cap is the chunk's `H + W` rather than the raster's.

## Reproduction

Ran on this host with CUDA available.

```python
from xrspatial.surface_distance import surface_distance

def serpentine(n):
    """NaN everywhere except one winding open corridor from (0, 0)."""
    elev = np.full((n, n), np.nan)
    for r in range(0, n, 2):
        elev[r, :] = 0.0
    for r in range(1, n, 2):
        elev[r, n - 1 if (r // 2) % 2 == 0 else 0] = 0.0
    return elev

def make(data, gpu=False):
    a = xr.DataArray(
        data.astype(float), dims=['y', 'x'],
        coords={'y': np.arange(data.shape[0], dtype=float),
                'x': np.arange(data.shape[1], dtype=float)},
        attrs={'res': (1.0, 1.0)})
    if gpu:
        a.data = cupy.asarray(a.data)
    return a

for N in (8, 12, 16, 24):
    elev = serpentine(N)
    src = np.zeros((N, N)); src[0, 0] = 1.0
    cpu = np.asarray(surface_distance(make(src), make(elev)).data)
    gpu = surface_distance(make(src, True), make(elev, True)).data.get()
    print(f"N={N:3d} cap={2*N:3d}  reachable cpu={int(np.isfinite(cpu).sum()):4d}"
          f" gpu={int(np.isfinite(gpu).sum()):4d}"
          f"  wrongly NaN on gpu={int((np.isfinite(cpu) & ~np.isfinite(gpu)).sum()):4d}")
```

```
N=  8 cap= 16  reachable cpu=  36 gpu=  21  wrongly NaN on gpu=  15
N= 12 cap= 24  reachable cpu=  78 gpu=  29  wrongly NaN on gpu=  49
N= 16 cap= 32  reachable cpu= 136 gpu=  37  wrongly NaN on gpu=  99
N= 24 cap= 48  reachable cpu= 300 gpu=  53  wrongly NaN on gpu= 247
```

At N=16 the GPU finds 37 of the 136 reachable pixels and calls the other 99
unreachable. The last corridor row is the clearest case:

```
row 15, numpy: [126.21   nan   nan ...  nan]
row 15, cupy : [   nan   nan   nan ...  nan]
```

Distances that both backends do report agree exactly (max abs diff 0.0), so
this is purely truncated propagation rather than arithmetic drift.

A serpentine grid is an easy way to make the failure obvious, but nothing
about it is special. Any NaN-masked terrain whose shortest paths wind more
than `H + W` hops (a river network, a canyon system, a coastline with inlets)
hits the same cap.

The numpy and dask+numpy paths run an exact Dijkstra and are correct here. I
also checked the dask iterative sweep cap (`max(n_tile_y, n_tile_x) + 10`)
against the same mazes at several chunk sizes and could not make it produce a
wrong answer, so this report is about the GPU cap only.

## Expected behaviour

The cupy backend should relax until nothing changes, and its result should
match numpy. If some bound must remain as a safety net, it should be the
Bellman-Ford bound for the graph (`H * W`) rather than `H + W`, and reaching
it should warn rather than pass off a truncated answer as a finished one.

The early-exit check on `changed` already means well-behaved terrain costs the
same number of passes as it does today, so raising the ceiling is not a
throughput change for normal input.

## Environment

xarray-spatial `main` @ 069fc13, Python 3.14, cupy with CUDA available.

brendancol:
Closes #3721

## What changed

`_surface_distance_cupy()` solves the shortest-path problem by repeated
parallel relaxation and capped itself at `H + W` passes. One pass moves
distance information roughly one pixel hop, so what the loop actually needs is
the hop count of the longest shortest path. The two numbers agree on open
terrain and come apart as soon as NaN barriers make paths wind; the
Bellman-Ford bound for a grid graph is the pixel count, not the perimeter.

Reaching the cap exited the loop with no warning and nothing in the output to
say so. Pixels the CPU backend reaches came back NaN, which a caller reads as
"no path exists". On a 16x16 serpentine corridor the GPU found 37 of the 136
reachable pixels; at 24x24 it found 53 of 300.

The ceiling is now `H * W`, and reaching it warns instead of quietly handing
back a truncated answer.

## Throughput

The `changed` flag still ends the loop as soon as the solution settles, so
ordinary terrain runs exactly as many passes as it did before. Timed on this
host against `main`:

| raster | main | this branch |
|--------|------|-------------|
| 256x256 random elevation | 0.136 s | 0.135 s |
| 512x512 random elevation | 0.326 s | 0.313 s |

Both inside run-to-run noise.

## Backends

cupy, and the bounded dask+cupy path, which calls this function per chunk
against that chunk's `H` and `W` (so the old cap bit even harder there).
numpy and dask+numpy run an exact Dijkstra and were already correct.

## Test plan

- [x] `pytest xrspatial/tests/test_surface_distance.py` — 39 passed (37 before)
- [x] `flake8 xrspatial/surface_distance.py` clean
- [x] New `test_cupy_long_path_matches_numpy`, parametrized over
      `surface_distance` and `surface_allocation`, builds a serpentine
      corridor and asserts cupy matches numpy. It fails against `main`'s
      `surface_distance.py` and passes here.
- [x] Checked the repro from the issue at N = 8, 12, 16, 24: the GPU now
      reports the same reachable set as numpy at every size, with zero
      difference on the distances.

Found by `/sweep-accuracy` against `xrspatial/surface_distance.py`.

### DIFF
--- a/xrspatial/surface_distance.py
+++ b/xrspatial/surface_distance.py
@@ -608,7 +608,7 @@ def _surface_distance_cupy(source_data, elev_data, cellsize_x, cellsize_y,
             f"and was stopped.  Some pixels may be reported as unreachable "
             f"when a path exists.  Please report this raster upstream.",
             UserWarning,
-            stacklevel=2,
+            stacklevel=4,
         )
 
     # Extract output
--- a/xrspatial/surface_distance.py
+++ b/xrspatial/surface_distance.py
@@ -580,7 +580,15 @@ def _surface_distance_cupy(source_data, elev_data, cellsize_x, cellsize_y,
     changed = cp.zeros(1, dtype=cp.int32)
     griddim, blockdim = cuda_args((H, W))
 
-    max_iterations = H + W
+    # One pass moves distance information roughly one pixel hop, so the
+    # number of passes needed is the hop count of the longest shortest
+    # path.  On open terrain that is about H + W, but NaN barriers make
+    # paths wind, and the Bellman-Ford bound for a grid graph is the
+    # pixel count.  The `changed` check below exits as soon as the
+    # solution settles, so the ceiling only costs anything on the
+    # pathological inputs that need it.
+    max_iterations = H * W
+    converged = False
     for _ in range(max_iterations):
         changed[0] = 0
         _sd_relax_kernel[griddim, blockdim](
@@ -590,7 +598,18 @@ def _surface_distance_cupy(source_data, elev_data, cellsize_x, cellsize_y,
             np.float64(max_distance),
         )
         if int(changed[0]) == 0:
+            converged = True
             break
+
+    if not converged:
+        warnings.warn(
+            f"surface_distance: the GPU relaxation was still improving "
+            f"distances after {max_iterations} passes on a {H}x{W} raster "
+            f"and was stopped.  Some pixels may be reported as unreachable "
+            f"when a path exists.  Please report this raster upstream.",
+            UserWarning,
+            stacklevel=2,
+        )
 
     # Extract output
     if mode == DISTANCE:
--- a/xrspatial/tests/test_surface_distance.py
+++ b/xrspatial/tests/test_surface_distance.py
@@ -543,6 +543,46 @@ def test_cupy_matches_numpy():
                                equal_nan=True)
 
 
+def _serpentine_elevation(n):
+    """NaN barriers everywhere except one winding open corridor.
+
+    The corridor from (0, 0) is about n*n/2 pixel hops long, far more than
+    the n+n the GPU relaxation used to allow itself.
+    """
+    elev = np.full((n, n), np.nan, dtype=np.float64)
+    for r in range(0, n, 2):
+        elev[r, :] = 0.0
+    for r in range(1, n, 2):
+        elev[r, n - 1 if (r // 2) % 2 == 0 else 0] = 0.0
+    return elev
+
+
+@pytest.mark.skipif(not has_cuda_and_cupy(), reason="cupy/cuda not available")
+@pytest.mark.parametrize("mode", [surface_distance, surface_allocation])
+def test_cupy_long_path_matches_numpy(mode):
+    """CuPy must relax until it converges, not for a fixed H+W (#3721).
+
+    A winding corridor needs far more relaxation passes than the raster is
+    wide plus tall. Stopping early makes reachable pixels come back NaN,
+    which reads as "no path exists".
+    """
+    n = 16
+    elev = _serpentine_elevation(n)
+    source = np.zeros((n, n), dtype=np.float64)
+    source[0, 0] = 1.0
+
+    np_result = _compute(mode(_make_raster(source), _make_raster(elev)))
+    cp_result = _compute(mode(_make_raster(source, backend='cupy'),
+                              _make_raster(elev, backend='cupy')))
+
+    # The corridor is genuinely reachable end to end on the CPU.
+    assert np.isfinite(np_result[n - 1, 0])
+    assert int(np.isfinite(np_result).sum()) > 4 * n
+
+    np.testing.assert_allclose(cp_result, np_result, rtol=1e-5,
+                               equal_nan=True)
+
+
 @pytest.mark.skipif(not has_cuda_and_cupy(), reason="cupy/cuda not available")
 def test_cupy_returns_cupy_array():
     """CuPy input should produce CuPy output."""
--- a/xrspatial/tests/test_surface_distance.py
+++ b/xrspatial/tests/test_surface_distance.py
@@ -558,7 +558,8 @@ def _serpentine_elevation(n):
 
 
 @pytest.mark.skipif(not has_cuda_and_cupy(), reason="cupy/cuda not available")
-@pytest.mark.parametrize("mode", [surface_distance, surface_allocation])
+@pytest.mark.parametrize(
+    "mode", [surface_distance, surface_allocation, surface_direction])
 def test_cupy_long_path_matches_numpy(mode):
     """CuPy must relax until it converges, not for a fixed H+W (#3721).
 


brendancol (commented):
## PR Review: Let the GPU surface-distance relaxation run to convergence

Read the whole of `_surface_distance_cupy` and `_sd_relax_kernel` rather than
just the diff, and replayed the issue repro at N = 8, 12, 16, 24 on this host.
The diagnosis holds: `H + W` is the perimeter, not the hop count of the
longest shortest path, and `H * W` is the right ceiling for a grid graph with
`H * W` nodes. Since the in-place relaxation converges at least as fast as
Jacobi Bellman-Ford, the new bound is provably sufficient, which makes the
warning unreachable in theory. That is the correct shape for a safety net.

None.

- **`xrspatial/tests/test_surface_distance.py:565` — the long-path test skips
  DIRECTION.** It is parametrized over `surface_distance` and
  `surface_allocation`, but `_surface_distance_cupy` runs the same relaxation
  loop for all three modes and DIRECTION reads `srow`/`scol`, which the
  truncated loop leaves unset just as it leaves `dist` at infinity. Adding
  `surface_direction` to the parametrize list costs one word and covers the
  third mode.

- **`xrspatial/surface_distance.py:608` — `stacklevel=2` points inside the
  library.** From a user's `surface_distance()` call the chain is the
  `supports_dataset` wrapper, `_compute`, then `_surface_distance_cupy`, so
  level 2 lands on `_compute`. The module's other warning
  (`_surface_distance_dask`) uses `stacklevel=4` for the same call depth.
  Worth matching so the warning points at the caller.

- **Runtime on a genuinely pathological raster is now unbounded in practice.**
  A raster whose shortest paths really do need `H * W` passes will run `H * W`
  kernel launches, and on anything large that is a hang rather than an answer.
  The old cap made that case fast and wrong, which is worse, so this is the
  right trade. But nothing tells the caller they are in it. A cheap
  improvement would be a one-time note once the pass count passes, say,
  `4 * (H + W)`, so a long run is at least explicable. I would not hold the PR
  for it.

- **`int(changed[0])` syncs the device to the host every pass.** Pre-existing,
  and the pass count for converging input is unchanged, so this PR does not
  make it worse. Mentioning it only because raising the ceiling makes the
  per-pass cost matter more if the pathological case ever shows up.

- **The warning may not surface from a dask worker.** The bounded dask+cupy
  path calls this function inside a `map_overlap` task, where `warnings.warn`
  can be swallowed depending on the scheduler. Same caveat applies to the
  module's existing warnings, so no action here.

### Coverage gap worth recording

There is no dask+cupy long-path test, and I do not think a meaningful one is
writable today: the unbounded dask+cupy path converts to numpy and runs the
CPU Dijkstra, and the bounded path caps path length by `max_distance` well
below the point where the pass count matters. The per-chunk cap is still real
(a chunk's `H + W` is smaller than the raster's), so the fix matters there
even though a test cannot reach it directly.

### What looks good

- The failure is characterized by counts, not adjectives: 37 of 136 reachable
  pixels at 16x16, 53 of 300 at 24x24, and the distances that both backends
  did report agreed exactly. That separates truncated propagation from
  arithmetic drift, which is the distinction that matters here.
- The throughput claim is measured against `main` on the same host rather than
  argued from the early-exit check.
- The comment above `max_iterations` explains why the old bound looked
  plausible, which is more useful to the next reader than only stating the new
  one.
- `_serpentine_elevation` is a compact, deterministic way to build a corridor
  much longer than the raster's perimeter, and the test asserts the corridor
  really is reachable on the CPU before comparing, so it cannot pass
  vacuously.

### Checklist

- [x] Algorithm matches reference — `H * W` is the Bellman-Ford bound for a
      grid graph of `H * W` nodes
- [x] All implemented backends produce consistent results — cupy now matches
      numpy on the maze at every size tested
- [x] NaN handling is correct — the truncation was producing spurious NaN;
      that is what the fix removes
- [x] Edge cases covered by tests — long winding path, the case the old cap
      could not reach
- [ ] Dask chunk boundaries — not applicable, no change to the tiling
- [x] No premature materialization — no new host transfers
- [x] Benchmark exists — `benchmarks/benchmarks/surface_distance.py`; timings
      against `main` are in the PR body
- [ ] README feature matrix — not applicable, no new function
- [x] Docstrings — internal function, the reasoning is in the code comment

brendancol (commented):
## PR Review: follow-up pass after 1240e3ab

Re-read the changed hunks and re-ran the suite. Disposition of the first pass:

**Fixed**

- The long-path test now parametrizes over `surface_direction` as well as
  `surface_distance` and `surface_allocation`. All three modes share the
  relaxation loop and DIRECTION reads `srow`/`scol`, so it was the mode most
  worth adding. 40 tests pass (37 on `main`).
- The convergence warning moved from `stacklevel=2` to `4`, matching
  `_surface_distance_dask`'s choice for the same call depth, so it points at
  the caller rather than at `_compute`.

**Dismissed, with reasons**

- *Warn partway through a long run.* Any threshold I picked (`4 * (H + W)`, or
  anything else) would be a number with nothing behind it, and it would fire
  on input that is slow but correct. A warning nobody can act on is worse than
  no warning. The real answer for a raster that deep is to route it to the CPU
  Dijkstra, which is a design change and does not belong in a bug fix.
- *`int(changed[0])` syncs the device to the host every pass.* Pre-existing,
  and the pass count for converging input is identical to `main`, so this PR
  does not touch it. Changing it means restructuring the termination check.
- *The warning may not surface from a dask worker.* Applies equally to the
  module's existing warnings; nothing specific to this change to fix.

**Recorded, not fixed**

- No dask+cupy long-path test. The unbounded dask+cupy path converts to numpy
  and runs the CPU Dijkstra, and the bounded path caps path length by
  `max_distance` well below where the pass count bites, so I could not write
  one that actually reaches the code. The per-chunk cap is still real and the
  fix still applies there.

No new findings. Nothing outstanding from my side.

PR #3735, opened 2026-08-26
Add weight raster to flow accumulation (#3734)

### LINKED ISSUE #3734: weights for flow accumulation
## Reason or Problem
`flow_accumulation` only computes a count of upstream cells draining through each cell. There's no way to weight that accumulation by another raster (e.g. precipitation or snowmelt), so getting a physically meaningful accumulated flux currently requires reimplementing the traversal outside the library.

## Proposal
Add an optional `weight` parameter to `flow_accumulation_d8` and the other flow_accumulation functions.

**Design:**
```python
def flow_accumulation_d8(
    flow_dir: xr.DataArray,
    weight: xr.DataArray | None = None,
    name: str = 'flow_accumulation',
) -> xr.DataArray:
```
When `weight` is `None`, behavior is unchanged. When provided, each output cell equals the sum of `weight` over itself and all upstream cells draining through it, instead of a cell count. I.s. weight acts like a precipitation or surface melt field. 

**Usage:**
```python
melt_accum = flow_accumulation_d8(flow_dir, weight=melt)
```
`weight` must be on the same grid as `flow_dir`.

## Stakeholders and Impacts
Anyone using xarray-spatial for hydrology. 

## Unresolved Questions
How NaN values in `weight` should be handled (zero contribution vs. propagating as no-data downstream).

brendancol:
Closes #3734

Adds an optional `weight` raster to `flow_accumulation_d8`, `flow_accumulation_dinf` and `flow_accumulation_mfd`. `flow_accumulation(routing=...)` and the `.xrs` accessor forward kwargs, so they pick it up too.

With `weight`, each output cell is the sum of `weight` over itself and everything upstream of it (split by the D-inf / MFD fractions for those routers). Without it nothing changes: same unit count, same kernels, same memory footprint. The kernels take `(weight, has_weight)` and the unweighted path passes a 1x1 dummy.

NaN handling, answering the open question on the issue: a NaN weight contributes 0 and the output NaN mask still follows `flow_dir`. Propagating NaN downstream would blank the whole basin below one missing pixel, and treating it as a barrier would silently drop upstream flow, so neither seemed usable. Anyone wanting a different convention can `fillna` first.

Other details:

- `weight` sits between `flow_dir` and `name`, as proposed on the issue, so `name` is no longer the second positional argument. Nothing in the repo passes `name` positionally (accessor and notebooks use the keyword) and a stray string now raises a clear `TypeError` from `_validate_raster`.

- `weight` goes through `_validate_raster` + `_validate_matching_shape`; a shape mismatch raises `ValueError`.
- Memory guards add 8 bytes/pixel when a weight is present.
- Dask paths carry the weight tile through the boundary sweep and the lazy `map_blocks` assembly. A numpy weight, or a dask weight chunked differently from `flow_dir`, is rechunked to match.
- Docstrings updated, plus a short "Weighted accumulation" section in the hydrology user guide notebook.

Backends: numpy, cupy, dask+numpy, dask+cupy for all three routers.

Test plan:
- [x] Known-value tests per router (D8 chain, D-inf 8-neighbour pit and a pi/8 proportional split, MFD plane)
- [x] `weight=1` reproduces the unweighted count
- [x] NaN weight contributes 0; NaN mask equals `isnan(flow_dir)`
- [x] Shape mismatch and non-DataArray weight raise
- [x] dask parity across chunk layouts, including a weight chunked differently from `flow_dir`
- [x] cupy and dask+cupy parity (run locally on a CUDA box)
- [x] Physical accuracy checks on a `generate_terrain` DEM with an orographic melt field (`test_flow_accumulation_weight_accuracy.py`): mass balance per router and per basin (input weight = pit totals + leak off the valid grid, leak computed from `flow_dir` independently of the kernels), watershed cross-check at channel and hillslope pour points with melt and a glacier indicator, linearity in `weight`, and downstream monotonicity for non-negative weight. Mass balance runs on all four backends.
- [x] Existing `xrspatial/hydro/tests` and accessor tests still pass (1083 passed)

### DIFF
--- a/examples/user_guide/11_Hydrology.ipynb
+++ b/examples/user_guide/11_Hydrology.ipynb
@@ -418,6 +418,50 @@
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Weighted accumulation\n",
+    "\n",
+    "Counting cells assumes every cell contributes the same amount of water. Passing a `weight` raster on the same grid changes the quantity being accumulated: each output cell becomes the sum of `weight` over itself and everything upstream of it. With a precipitation or snowmelt field as the weight, the result is an accumulated flux rather than a contributing-cell count.\n",
+    "\n",
+    "The example below builds a synthetic orographic precipitation field (wetter at higher elevation and toward the west) and compares the weighted result with the plain count. NaN cells in `weight` contribute zero and do not change the output NaN mask, which still follows `flow_dir`. Fill NaN explicitly if you want a different convention.\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Synthetic precipitation (mm): rises with elevation and toward the west.\n",
+    "elev_norm = (dem - float(np.nanmin(dem.values))) / float(np.ptp(dem.values[~np.isnan(dem.values)]))\n",
+    "west_east = xr.DataArray(np.linspace(1.0, 0.4, W)[None, :].repeat(H, axis=0),\n",
+    "                         dims=dem.dims, coords=dem.coords)\n",
+    "precip = (400 + 1200 * elev_norm) * west_east\n",
+    "precip.name = 'precip'\n",
+    "\n",
+    "precip_accum = xrspatial.flow_accumulation(flow_dir, weight=precip)\n",
+    "\n",
+    "print(f\"Cell count at outlet: {np.nanmax(flow_accum.values):.0f}\")\n",
+    "print(f\"Accumulated precipitation at outlet: {np.nanmax(precip_accum.values):.3g} mm-cells\")\n",
+    "\n",
+    "fig, axes = plt.subplots(1, 2, figsize=(14, 5))\n",
+    "fig.patch.set_facecolor('white')\n",
+    "for ax, data, label in zip(axes, [flow_accum, precip_accum],\n",
+    "                           ['Upstream cell count', 'Accumulated precipitation']):\n",
+    "    ax.set_facecolor('white')\n",
+    "    data.plot.imshow(ax=ax, cmap=water_cmap,\n",
+    "                     norm=LogNorm(vmin=float(np.nanmin(data.values[data.values > 0])),\n",
+    "                                  vmax=float(np.nanmax(data.values))),\n",
+    "                     add_colorbar=True, cbar_kwargs={'label': f'{label} (log scale)',\n",
+    "                                                     'shrink': 0.7})\n",
+    "    ax.set_title(label)\n",
+    "    ax.set_axis_off()\n",
+    "plt.tight_layout()\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "id": "httgn03vcco",
    "metadata": {},
    "source": "## Flow Accumulation (MFD)\n\n`flow_accumulation` with `routing='mfd'` routes upstream contributing area through all downslope paths at once, using the fractional weights from `flow_direction(routing='mfd')`. Where D8 accumulation produces sharp single-pixel drainage lines, MFD accumulation spreads flow and produces smoother contributing-area fields.\n\nThe side-by-side comparison shows D8 (left) with crisp channels and MFD (right) with diffuse drainage patterns."
--- a/xrspatial/hydro/flow_accumulation_d8.py
+++ b/xrspatial/hydro/flow_accumulation_d8.py
@@ -32,8 +32,9 @@ except ImportError:
 
 from xrspatial.dataset_support import supports_dataset
 from xrspatial.hydro._boundary_store import BoundaryStore
-from xrspatial.utils import (_dask_task_name_kwargs, _validate_raster, cuda_args,
-                             has_cuda_and_cupy, is_cupy_array, is_dask_cupy, ngjit)
+from xrspatial.utils import (_dask_task_name_kwargs, _validate_matching_shape,
+                             _validate_raster, cuda_args, has_cuda_and_cupy,
+                             is_cupy_array, is_dask_cupy, ngjit)
 
 # =====================================================================
 # Memory guards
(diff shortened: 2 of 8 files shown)


brendancol (commented):
## PR Review: Add weight raster to flow accumulation (#3734)

Verified in a worktree on the PR head. The three test files pass (121 passed, including the cupy and dask+cupy cases on a CUDA box). Also ran ad-hoc probes for every backend pairing of `flow_dir` and `weight` (numpy, cupy, dask+numpy, dask+cupy on both sides), int/inf/3-D weights, Dataset input, `routing=` dispatch and positional `name`.

None.

- [ ] `xrspatial/hydro/flow_accumulation_mfd.py:714-716` and `:836`: the MFD dask+numpy path converts the weight tile with a bare `np.asarray(...)`, so a cupy-backed `weight` with a dask+numpy `flow_dir_mfd` raises "Implicit conversion to a NumPy array is not allowed" (reproduced). The D8 and D-inf paths route the same tile through `_to_numpy_f64`, which handles `.get()`, and every other mixed pairing works (cupy fd + numpy w, dask+cupy fd + numpy w, dask+numpy fd + cupy w for D8/D-inf, and all MFD pairings except this one). Either use `_to_numpy_f64` here too, or reject mismatched backends up front in `_validate_weight`. Low priority since it is a mixed-backend corner, but it is the one hole in the parity matrix.
- [ ] `xrspatial/hydro/flow_accumulation_d8.py:1071-1073` (and dinf `:985-987`, mfd `:879-881`): `weight` is inserted before `name`, so `flow_accumulation_d8(fd, 'myname')` now raises `TypeError: weight must be an xarray.DataArray, got builtins.str` instead of naming the output. No positional `name` callers in the repo (accessor and notebooks use kwargs), and the error message is clear, so this is a judgment call. If you want to keep the old positional contract, append `weight` after `name` or make both keyword-only. Worth a line in the PR description either way.
- [ ] Tests do not cover: an integer-dtype weight (works via `np.asarray(dtype=float64)`), a Dataset `flow_dir` with a DataArray `weight` through `@supports_dataset` (works, each variable gets the same weight), and the `.xrs` accessor / `flow_accumulation(routing=...)` forwarding path the PR description relies on (works). One small parametrised test for the first two would lock in the coercion behaviour.

- [ ] `xrspatial/hydro/tests/test_flow_accumulation_dinf.py:321`: docstring says "flowing at pi/4 is split 50/50" but the case uses pi/8 (the comment on the next line is correct). Fix the docstring.
- [ ] `xrspatial/hydro/tests/test_flow_accumulation_mfd.py:530-541`: `test_mass_conserved_on_plane` docstring describes a bottom-row column-sum check, but the assertion is `accum == count * 2.5` (uniform-weight scaling). Either rename or make the docstring match.
- [ ] `xrspatial/hydro/tests/test_flow_accumulation_dinf.py:365-368`: every parametrised chunking uses a weight chunked `(2, 7)`, so the "weight already chunked like flow_dir" branch of `_weight_as_dask` is only exercised by the D8 and MFD tests. Fine as-is since the helper is shared.
- [ ] `xrspatial/hydro/flow_accumulation_d8.py:806-810`: the weight tile is recomputed from the dask graph on every sweep pass. This mirrors how `flow_dir_da.blocks[iy, ix].compute()` is already handled, so it is not a regression, but a weight backed by an expensive graph (e.g. a lazy reprojection) pays that cost twice per sweep. Not worth changing here.

### What looks good

- Kernel indexing is correct everywhere checked: `_cell_weight` (`d8.py:154`) and the three `@cuda.jit` init kernels index `weight[r, c]` / `weight[i, j]` with the same local tile coordinates used for `flow_dir`, and `_NO_WEIGHT` is never indexed when `has_weight == 0`.
- NaN policy is implemented identically on all four backends for all three routers: NaN weight becomes 0.0, NaN mask stays `isnan(flow_dir)` (or `isnan(fractions[0])` for MFD). Negative and inf weights propagate arithmetically, no clamping, matching the documented contract.
- Dask tile alignment: `_weight_as_dask` (`d8.py:177`) rechunks to `flow_dir.chunks` before any `.blocks[iy, ix]` access, so the D8/D-inf tile lookups and the MFD cumulative-offset slicing (`mfd.py:697-716`) always see the same tile. Verified with `(2,5)` / `(4,2)` weight chunks against `(3,3)` flow chunks and with `neighbor` chunked on axis 0.
- `map_blocks` with two inputs: `block_info[0]` still refers to the flow array in both `_assemble_result*` variants, and the 2-D weight alongside the 3-D fractions with `drop_axis=0` (`mfd.py:840-846`) pairs correctly by trailing axes. dask+cupy with a numpy weight still yields cupy blocks.
- No eager materialisation: a numpy weight is wrapped with `da.from_array`, a dask weight stays lazy, tiles are computed one at a time.
- Memory guards: 8 bytes/pixel for the float64 weight copy is right for the conversion cases and slightly conservative when the weight is already float64, which is the safe direction.
- Notebook cells come after `dem`, `H, W`, `flow_dir`, `flow_accum`, `water_cmap` and `LogNorm` are defined; the LogNorm `vmin` guard is fine because precip is strictly positive.
- Docstrings are accurate for all three routers.

### Checklist

- [x] Algorithm matches reference (sum of weight over self + upstream; fraction-split for D-inf/MFD)
- [x] Backends consistent (one mixed-backend corner in MFD noted above)
- [x] NaN handling
- [x] Edge cases tested (int dtype and Dataset input untested)
- [x] Dask chunk boundaries
- [x] Benchmark exists or not needed (weighted path only changes per-cell init)
- [x] README updated if applicable (n/a)
- [x] Docstrings

brendancol (commented):
## PR Review: follow-up after 1d802e0c

Re-checked the review findings against the follow-up commit.

None.

None remaining.

- Fixed: MFD dask+numpy path with a cupy weight. `xrspatial/hydro/flow_accumulation_mfd.py:715`, `:836` and `:953` now go through `_to_numpy_f64`, which calls `.get()` on cupy input and is a no-op on numpy. Reproduced the failure before the change and confirmed it passes after.
- Fixed: added `test_weight_integer_dtype`, `test_weight_dataset_input` and `test_weight_forwarded_by_accessor_and_routing` in `test_flow_accumulation_d8.py`. 124 tests pass across the three files.
- Fixed: both test docstrings (`test_flow_accumulation_dinf.py` pi/8 wording; MFD test renamed to `test_uniform_weight_scales_count` with a matching docstring), and the D-inf dask test now also exercises a weight chunked identically to `flow_dir`.
- Dismissed with reason: positional `name` ordering. Kept `weight` before `name` to match the signature proposed on #3734; no in-repo positional callers and the failure mode is a clear `TypeError`. Noted in the PR description.
- Dismissed with reason: weight tile recomputed per sweep pass. Mirrors the existing `flow_dir` handling; reviewer agreed it is not worth changing here.

### Checklist

- [x] Backends consistent (mixed-backend MFD corner now covered)
- [x] Tests pass locally including cupy and dask+cupy

PR #3741, opened 2026-09-04
Replace the hotspots confidence ladder with boolean arithmetic (#3737)

### LINKED ISSUE #3737: hotspots: replace the nine-branch confidence ladder with boolean arithmetic
## Reason or Problem

`_calc_hotspots_numpy` in `xrspatial/focal.py` classifies each Gi* z-score with a nine-branch `if/elif` ladder. It first maps `abs(zscore)` to one of four p-values, then tests `abs(zscore)` again against three confidence thresholds combined with `p_value < ...`, and finally picks the sign in a third pair of branches. `abs(zscore)` gets evaluated up to six times per cell, and the data-dependent branches keep the loop from vectorizing.

The p-value step does nothing. Each `p_value < threshold` test is already implied by the `abs(zscore)` test sitting next to it:

- `|z| > 2.58` implies `|z| >= 2.33`, so `p_value = 0.0099 < 0.01`
- `|z| > 1.96` implies `|z| >= 1.65`, so `p_value <= 0.0495 < 0.05`
- `|z| > 1.65` implies `|z| >= 1.65`, so `p_value <= 0.0495 < 0.1`

The 1.29 threshold never reaches the output. The whole ladder reduces to: confidence 99 when `|z| > 2.58`, 95 when `1.96 < |z| <= 2.58`, 90 when `1.65 < |z| <= 1.96`, otherwise 0.

## Proposal

Replace the ladder with boolean arithmetic:

```python
az = abs(zscore)
confidence = 90 * (az > 1.65) + 5 * (az > 1.96) + 4 * (az > 2.58)
hot_cold = (zscore > 0) - (zscore < 0)
out[y, x] = hot_cold * confidence
```

NaN z-scores compare False everywhere and classify to 0, same as today. `+inf` and `-inf` classify to `+99` and `-99`, same as today. Output stays `int8` and the signature does not change.

A spike measured the rewrite at roughly 0.65x to 0.72x of the current wall time on a 2000x4000 float64 z-score array (random normal, smooth ramp, and 10% NaN inputs), with bitwise identical int8 output.

**Design:** Put a short comment above the new code stating the collapsed thresholds so nobody has to re-derive them from the old ladder. The CUDA device function `_gpu_hotspots` carries the same ladder. Leave it alone for now: branch cost on the GPU is a different question and that kernel is not on the timed CPU path.

**Value:** The classification runs once per cell after the convolution on the numpy and dask+numpy backends, so a cheaper classification shows up directly in `hotspots()` wall time on large rasters.

## Stakeholders and Impacts

Users of `hotspots()` on numpy and dask+numpy. No API or output change.

## Drawbacks

The arithmetic form is less literal than the ladder. That is the reason for the comment documenting the thresholds.

## Alternatives

Keep the ladder but hoist `abs(zscore)` into a local. That removes the repeated calls but keeps the branches, so it recovers only part of the speedup.

Verification plan: extract the current function from `origin/main` into a scratch module and assert `np.array_equal` against the new one on a large random array, a dense `linspace(-4, 4)` sweep that crosses every threshold at both signs, the exact threshold values and their float64 neighbours, and an array with NaN, +inf, -inf and signed zeros. Add the boundary and non-finite cases as a pytest test. The asv class `FocalHotspots` already benchmarks this path.

brendancol:
Closes #3737

- `_calc_hotspots_numpy` classified each Gi* z-score through a nine-branch `if/elif` ladder that recomputed `abs(zscore)` up to six times per cell. The p-value half of that ladder never changed the output: each `p_value < threshold` test is implied by the `|z|` test beside it (`|z| > 2.58` forces `p = 0.0099`, `|z| > 1.96` and `|z| > 1.65` both force `p <= 0.0495`). The whole thing reduces to 99 for `|z| > 2.58`, 95 for `1.96 < |z| <= 2.58`, 90 for `1.65 < |z| <= 1.96`, else 0.
- The classifier now computes `confidence = 90*(az > 1.65) + 5*(az > 1.96) + 4*(az > 2.58)` and `hot_cold = (z > 0) - (z < 0)`. A comment above the code states the collapsed thresholds so nobody has to re-derive them. Output dtype (int8) and signature are unchanged.
- The CUDA device function `_gpu_hotspots` still uses the ladder. It is not on the timed CPU path and branch cost on the GPU is a separate question, so it is left alone here.

## Timings

`_calc_hotspots_numpy` alone, 2000x4000 float64 z-scores, median of 5 after warmup, 20-core host:

| input | before | after |
|---|---|---|
| random normal (sd 1.5) | 18.1 ms | 11.6 ms |
| smooth ramp | 18.3 ms | 12.6 ms |
| random normal, 10% NaN | 18.9 ms | 12.1 ms |

Identical int8 results in every case.

## Verification

The original function was extracted from `origin/main` into a scratch module and compared with `np.array_equal` against the new one on:

- 2000x4000 `rng.normal(0, 1.5)` in float64 and float32
- a dense `np.linspace(-4, 4, 2_000_001)` sweep that crosses every threshold at both signs
- the exact values `+/-1.29, 1.65, 1.96, 2.33, 2.58` and their float64 neighbours via `np.nextafter`
- NaN, +inf, -inf, +0.0, -0.0

All equal. The threshold-neighbour and non-finite cases are now pytest tests (`test_hotspots_classifier_thresholds_3737`, `test_hotspots_classifier_nonfinite_3737`) with hand-derived expected values.

Backends: numpy and dask+numpy share this kernel and both benefit. cupy and dask+cupy use the untouched CUDA kernel.

## Test plan

- [x] `pytest xrspatial/tests/test_focal.py -x -q` (353 passed, GPU tests included on a CUDA box)
- [x] A/B equality against the `origin/main` implementation on the inputs listed above
- [x] Before/after timing on 2000x4000 float64

No benchmark change: `FocalHotspots` in `benchmarks/benchmarks/focal.py` already covers `hotspots()`.

### DIFF
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -1469,29 +1469,16 @@ def _calc_hotspots_numpy(z_array):
         for x in prange(cols):
             zscore = z_array[y, x]
 
-            # find p value
-            p_value = 1.0
-            if abs(zscore) >= 2.33:
-                p_value = 0.0099
-            elif abs(zscore) >= 1.65:
-                p_value = 0.0495
-            elif abs(zscore) >= 1.29:
-                p_value = 0.0985
-
-            # confidence
-            confidence = 0
-            if abs(zscore) > 2.58 and p_value < 0.01:
-                confidence = 99
-            elif abs(zscore) > 1.96 and p_value < 0.05:
-                confidence = 95
-            elif abs(zscore) > 1.65 and p_value < 0.1:
-                confidence = 90
-
-            hot_cold = 0
-            if zscore > 0:
-                hot_cold = 1
-            elif zscore < 0:
-                hot_cold = -1
+            # Confidence is 99 for |z| > 2.58, 95 for 1.96 < |z| <= 2.58,
+            # 90 for 1.65 < |z| <= 1.96, else 0. The GPU twin
+            # (_gpu_hotspots) spells this out as a p-value / confidence
+            # ladder; the p-value tests there are implied by the |z|
+            # tests, so the ladder collapses to these three thresholds.
+            # NaN compares False everywhere and classifies to 0.
+            az = abs(zscore)
+            confidence = (90 * (az > 1.65) + 5 * (az > 1.96)
+                          + 4 * (az > 2.58))
+            hot_cold = (zscore > 0) - (zscore < 0)
 
             out[y, x] = hot_cold * confidence
     return out
--- a/xrspatial/focal.py
+++ b/xrspatial/focal.py
@@ -1470,14 +1470,12 @@ def _calc_hotspots_numpy(z_array):
             zscore = z_array[y, x]
 
             # Confidence is 99 for |z| > 2.58, 95 for 1.96 < |z| <= 2.58,
-            # 90 for 1.65 < |z| <= 1.96, else 0. The GPU twin
-            # (_gpu_hotspots) spells this out as a p-value / confidence
-            # ladder; the p-value tests there are implied by the |z|
-            # tests, so the ladder collapses to these three thresholds.
-            # NaN compares False everywhere and classifies to 0.
+            # 90 for 1.65 < |z| <= 1.96, else 0. This is the p-value /
+            # confidence ladder collapsed to the three thresholds that
+            # reach the output. NaN compares False everywhere and
+            # classifies to 0.
             az = abs(zscore)
-            confidence = (90 * (az > 1.65) + 5 * (az > 1.96)
-                          + 4 * (az > 2.58))
+            confidence = 90 * (az > 1.65) + 5 * (az > 1.96) + 4 * (az > 2.58)
             hot_cold = (zscore > 0) - (zscore < 0)
 
             out[y, x] = hot_cold * confidence
--- a/xrspatial/tests/test_focal.py
+++ b/xrspatial/tests/test_focal.py
@@ -1569,6 +1569,31 @@ def test_hotspots_classifier_thresholds_3737():
     np.testing.assert_array_equal(out, expected)
 
 
+def test_hotspots_classifier_float32_3737():
+    # The production callers feed the classifier float32 z-scores
+    # (_hotspots_numpy casts before classifying). The thresholds are
+    # float64 literals, so a float32 input is promoted before comparison:
+    # np.float32(1.65) sits just below 1.65 and np.float32(1.96) just
+    # above it. Pin that the float32 result equals the classification of
+    # the same values widened to float64, so a dtype-specific comparison
+    # cannot creep in.
+    from xrspatial.focal import _calc_hotspots_numpy
+
+    thresholds = np.array([1.29, 1.65, 1.96, 2.33, 2.58])
+    z = np.stack([np.nextafter(thresholds, -np.inf), thresholds,
+                  np.nextafter(thresholds, np.inf)])
+    z = np.concatenate([z, -z]).astype(np.float32)
+    z = np.concatenate([np.nextafter(z, np.float32(-np.inf)), z,
+                        np.nextafter(z, np.float32(np.inf))])
+
+    out32 = _calc_hotspots_numpy(z)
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Replace the hotspots confidence ladder with boolean arithmetic (#3737)

None.

- [ ] `xrspatial/tests/test_focal.py:1544` (`test_hotspots_classifier_thresholds_3737`) runs the classifier on float64, but every production caller feeds it float32 (`_hotspots_numpy` casts the z-scores at `focal.py:1553`). Under numba the float32 value is promoted to float64 before the comparison, so `np.float32(1.65)` sits just below `1.65` and `np.float32(1.96)` just above it. That is the same behaviour as the old ladder, and the PR's A/B covered float32, but no test pins it. Add a float32 check, for example assert that `_calc_hotspots_numpy(z.astype(np.float32))` equals `_calc_hotspots_numpy(z.astype(np.float32).astype(np.float64))` for the same threshold grid, so a future dtype-specific change to the comparison cannot slip through.

- [ ] `xrspatial/focal.py:1479` wraps `confidence = (90 * (az > 1.65) + 5 * (az > 1.96) + 4 * (az > 2.58))` across two lines with parentheses. The expression fits on one line under the 100-column limit and reads better that way.
- [ ] `xrspatial/focal.py:1472` describes `_gpu_hotspots` as still spelling out the ladder. If the GPU kernel is collapsed later, this comment goes stale; a shorter comment that states the three thresholds and leaves the GPU cross-reference out would survive that change.

### What looks good
- The derivation in the PR body is right: `|z| > 2.58` forces `p = 0.0099`, and `|z| > 1.96` or `|z| > 1.65` force `p <= 0.0495`, so each `p_value < ...` test in the old ladder was already true whenever its `|z|` test was. The 1.29 and 2.33 thresholds never reached the output.
- `hot_cold * confidence` stays in int64 inside the kernel and is cast to int8 on store, and the largest magnitude is 99, so there is no overflow path.
- NaN and +/-inf behaviour is identical to the ladder and now pinned by `test_hotspots_classifier_nonfinite_3737`.
- `test_hotspots_dask_cupy_matches_numpy` compares the untouched CUDA ladder against the rewritten CPU classifier on standard-normal data, so the two implementations are cross-checked in the 90 and 95 bands on a CUDA box.
- The labeler applied `performance`, and `FocalHotspots` in `benchmarks/benchmarks/focal.py` already times this path.

### Checklist
- [x] Algorithm matches reference (same classification as the previous ladder, verified by A/B and by the derivation)
- [x] All implemented backends produce consistent results (numpy/dask+numpy share the kernel; cupy/dask+cupy unchanged and parity-tested)
- [x] NaN handling is correct
- [x] Edge cases are covered by tests (threshold neighbours, NaN, inf, signed zero); float32 pin suggested above
- [x] Dask chunk boundaries handled correctly (no change to the map_blocks wrapper)
- [x] Benchmark exists
- [x] README feature matrix: not applicable, no public API change
- [x] Docstrings: not applicable, private kernel

brendancol (commented):
## PR Review: follow-up on b7d04469 (#3737)

Second pass after the review fixes.

None.

None.

None.

- Float32 pin: fixed. `test_hotspots_classifier_float32_3737` (`xrspatial/tests/test_focal.py:1572`) builds the threshold grid, casts to float32, widens it by one float32 ulp on each side, and asserts the classification equals the float64 widening of the same values. It also asserts all seven output bands appear, so the float32 grid really crosses every threshold at both signs.
- One-line confidence expression: fixed at `xrspatial/focal.py:1478`, 78 columns.
- Comment cross-reference to `_gpu_hotspots`: removed. The comment now states only the three thresholds and the NaN behaviour.

### What looks good
- The A/B against the `origin/main` kernel was re-run after the reformat and stays bitwise equal on every input class; the timing is unchanged (about 0.61x to 0.66x of the old kernel on 2000x4000 float64).
- Full `test_focal.py` passes (354 tests, GPU cases included).

No further changes requested.

PR #3742, opened 2026-09-04
Drop the per-cell NaN early-out in the planar CPU slope kernel

### LINKED ISSUE #3739: slope: drop the per-cell NaN early-out in the planar CPU kernel
## Reason or Problem

The planar CPU slope kernel (`_cpu` in `xrspatial/slope.py`) starts every interior cell with

```python
if np.isnan(data[y, x]):
    continue
```

On DEMs with scattered nodata (masked water, cloud holes, void-filled SRTM edges) the branch outcome depends on the data and the CPU can't predict it. What the skip saves is a few float32 multiplies and one `arctan`, and that is cheaper than the mispredicts.

A spike removed the branch and timed a 2000x4000 float32 DEM (Gaussian bump plus `default_rng(71942).normal(0, 2)` noise). With 30% random NaN it went from 69 ms to 33 ms (0.47x). With the left half NaN, a contiguous block the predictor handles well, it was 0.87x. With no NaN at all it was within noise, 0.95x to 1.0x. Results were bitwise identical in every case.

## Proposal

Drop the early-out. `out` is pre-filled with NaN, and the 3x3 Horn stencil's arithmetic already carries any neighbour NaN into `p`. The only case the branch actually protects is a NaN centre with eight valid neighbours (the #2761 regression), and a select after the arithmetic handles that:

```python
r = np.arctan(p) * 57.29578
ctr = data[y, x]
out[y, x] = r if ctr == ctr else np.nan
```

No fastmath, no dtype change, no reordering of the neighbour loads. The GPU and geodesic kernels are untouched.

Usage: nothing changes for callers. Same `slope()` signature, same results.

Value: about 2x on speckled-nodata DEMs for the numpy and dask+numpy backends.

## Stakeholders and Impacts

Anyone calling `slope()` on CPU. The dask+numpy path calls the same kernel per chunk, so it gets the same win.

This should also come with a NaN-bearing asv benchmark (`SlopeNaN`) so the nodata path has a timing from now on. One catch: `get_xr_dataarray(include_nan=True)` in `benchmarks/benchmarks/common.py` only sets the single corner cell `[0, 0]` to NaN, which sits on the border the kernel never visits. The benchmark has to add its own speckle on top or it measures the NaN-free path again.

## Drawbacks

Cells with a NaN centre now run the stencil and an `arctan` before being masked. That is the cost being traded away, and the numbers above say it is smaller than the branch.

## Alternatives

Leave the branch in and accept the mispredict cost. A per-row "any NaN" pre-check was also considered; it adds code and I could not see it beating the plain select.

## Unresolved Questions

None.

Gates before merge: `pytest xrspatial/tests/test_slope.py`, a direct A/B against the current kernel with `np.array_equal(..., equal_nan=True)` on the three regimes above plus 3x3, 1xN and int16 inputs, and a test that pins NaN to exactly the centre-NaN cells and their 8-neighbours on a speckled raster.

brendancol:
Closes #3739

- Drop the `if np.isnan(data[y, x]): continue` early-out at the top of the planar CPU slope kernel `_cpu`. Neighbour NaN already propagates through the Horn stencil, and `out` is pre-filled with NaN, so the only case the branch covered was a NaN centre with valid neighbours. That is now handled by a select after the arithmetic (`out[y, x] = r if ctr == ctr else np.nan`). No fastmath, no dtype change, neighbour loads in the same order.
- Add a `SlopeNaN` asv benchmark (numpy and dask, same `nx` grid as `Slope`) so the nodata path gets timed from now on.
- Add tests that pin the NaN footprint on a speckled raster: NaN at every centre-NaN cell and its 8-neighbours, finite everywhere else in the interior, on numpy and dask+numpy.

Backends: numpy and dask+numpy share `_cpu`, so both get the change. The cupy and dask+cupy kernels and the geodesic kernels are untouched.

## Why

On DEMs with scattered nodata the early-out is a data-dependent branch the CPU cannot predict, and the mispredicts cost more than the few float32 ops and one `arctan` the skip avoids.

## Timings

2000x4000 float32 DEM (Gaussian bump plus `default_rng(71942).normal(0, 2)` noise), `_cpu` called directly, `time.perf_counter`, median of 9 after warmup. Three separate runs on a box that had other test suites running at the same time, so the absolute numbers wobble; the ratios hold up.

| NaN pattern | before | after | ratio |
|---|---|---|---|
| none | 102 to 107 ms | 92 to 101 ms | 0.90x to 0.98x |
| 30% random NaN | 90 to 94 ms | 53 to 67 ms | 0.58x to 0.72x |
| left half NaN | 79 to 81 ms | 70 to 77 ms | 0.86x to 0.97x |

Identical results in every regime: `np.array_equal(old, new, equal_nan=True)` holds against the `_cpu` from `origin/main` for the three rasters above plus 3x3, 1x50, 50x1, 2x50, a 3x3 with a NaN centre, and an int16 200x300 input.

The NaN-free case is unchanged within noise. A contiguous NaN block is only a small win because the predictor learns it. The gain is specific to speckled nodata such as masked water or cloud holes, where the kernel is roughly 1.5x to 2x faster. The spike that motivated this measured 0.47x on a quiet machine; I could not get the box that quiet with the sibling runs going.

## On the benchmark's NaN pattern

`get_xr_dataarray(include_nan=True)` in `benchmarks/benchmarks/common.py` sets exactly one cell, `z[0, 0]`, to NaN. That is a border cell the kernel never visits, so on its own it times the NaN-free path again. `SlopeNaN.setup` calls it with `include_nan=True` and then masks 30% of cells at random (fixed seed) with `DataArray.where`, which keeps the dask array lazy for the dask case. Verified the resulting NaN fraction is 0.301 on both backends. `common.py` is not modified. The dask case forces `.compute()` so asv times the kernel rather than graph construction, following the twi and convolution benchmarks.

## Test plan

- [x] `pytest xrspatial/tests/test_slope.py -x -q`: 137 passed
- [x] Direct A/B against `origin/main`'s `_cpu` on the nine inputs listed above, all bitwise identical
- [x] `SlopeNaN.setup` and `time_slope_nan` run for numpy and dask at nx=1000
- [x] flake8 on the three touched files reports only the pre-existing aligned rows in the QGIS fixture (same 30 hits before and after)

### DIFF
--- a/benchmarks/benchmarks/slope.py
+++ b/benchmarks/benchmarks/slope.py
@@ -29,10 +29,13 @@ class SlopeNaN(Benchmarking):
         ny = nx // 2
         agg = get_xr_dataarray((ny, nx), type, include_nan=True)
         rng = np.random.default_rng(71942)
-        speckle = rng.random((ny, nx)) < 0.3
+        speckle = rng.random((ny, nx), dtype=np.float32) < 0.3
         self.xr = agg.where(~speckle)
 
     def time_slope_nan(self, nx, type):
+        # Force the compute so the dask case times the kernel rather than
+        # graph construction. Slope.time_slope does not, so the dask numbers
+        # of the two classes are not directly comparable.
         result = self.func(self.xr)
         if hasattr(result.data, "compute"):
             result.data.compute()
--- a/benchmarks/benchmarks/slope.py
+++ b/benchmarks/benchmarks/slope.py
@@ -1,6 +1,8 @@
+import numpy as np
+
 from xrspatial import slope
 
-from .common import Benchmarking
+from .common import Benchmarking, get_xr_dataarray
 
 
 class Slope(Benchmarking):
@@ -9,3 +11,28 @@ class Slope(Benchmarking):
 
     def time_slope(self, nx, type):
         return self.time(nx, type)
+
+
+class SlopeNaN(Benchmarking):
+    # Speckled nodata: the planar CPU kernel's cost depends on whether the
+    # NaN pattern is predictable, so time it separately from Slope.
+    # get_xr_dataarray(include_nan=True) only sets the [0, 0] corner to NaN,
+    # which is on the border the kernel never visits, so add 30% random NaN
+    # over the interior on top of it.
+    params = ([100, 300, 1000, 3000, 10000], ["numpy", "dask"])
+    param_names = ("nx", "type")
+
+    def __init__(self):
+        super().__init__(func=slope)
+
+    def setup(self, nx, type):
+        ny = nx // 2
+        agg = get_xr_dataarray((ny, nx), type, include_nan=True)
+        rng = np.random.default_rng(71942)
+        speckle = rng.random((ny, nx)) < 0.3
+        self.xr = agg.where(~speckle)
+
+    def time_slope_nan(self, nx, type):
+        result = self.func(self.xr)
+        if hasattr(result.data, "compute"):
+            result.data.compute()
--- a/xrspatial/slope.py
+++ b/xrspatial/slope.py
@@ -54,8 +54,11 @@ def _cpu(data, cellsize_x, cellsize_y):
     rows, cols = data.shape
     for y in range(1, rows - 1):
         for x in range(1, cols - 1):
-            if np.isnan(data[y, x]):
-                continue
+            # No early-out on a NaN centre: the branch is unpredictable on
+            # speckled nodata and costs more than the arithmetic it skips.
+            # Neighbour NaN already propagates through the stencil; the centre
+            # is folded back in with a select so out[y, x] is NaN whenever
+            # data[y, x] is.
             a = data[y + 1, x - 1]
             b = data[y + 1, x]
             c = data[y + 1, x + 1]
@@ -67,7 +70,9 @@ def _cpu(data, cellsize_x, cellsize_y):
             dz_dx = ((c + 2 * f + i) - (a + 2 * d + g)) / (8 * cellsize_x)
             dz_dy = ((g + 2 * h + i) - (a + 2 * b + c)) / (8 * cellsize_y)
             p = (dz_dx * dz_dx + dz_dy * dz_dy) ** .5
-            out[y, x] = np.arctan(p) * 57.29578
+            r = np.arctan(p) * 57.29578
+            ctr = data[y, x]
+            out[y, x] = r if ctr == ctr else np.nan
     return out
 
 
--- a/xrspatial/slope.py
+++ b/xrspatial/slope.py
@@ -43,7 +43,7 @@ def _geodesic_cuda_dims(shape):
 
 
 # =====================================================================
-# Planar backend functions (unchanged)
+# Planar backend functions
 # =====================================================================
 
 @ngjit
--- a/xrspatial/tests/test_slope.py
+++ b/xrspatial/tests/test_slope.py
@@ -625,3 +625,47 @@ def test_units_overridden_to_degrees_geodesic():
     result = slope(raster, method='geodesic')
     assert result.attrs['units'] == 'degrees'
     assert raster.attrs['units'] == 'm'
+
+
+# The planar CPU kernel has no per-cell NaN early-out (#3739): neighbour NaN
+# flows through the Horn stencil and the centre is masked with a select. Pin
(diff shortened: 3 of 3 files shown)


brendancol (commented):
## PR Review: Drop the per-cell NaN early-out in the planar CPU slope kernel

None.

- [ ] `benchmarks/benchmarks/slope.py:35-38`: `time_slope_nan` forces `.compute()` for dask, but the existing `Slope.time_slope` goes through `Benchmarking.time`, which does not. So for `type="dask"` the two classes time different things (graph build vs actual kernel work) and a side-by-side read of `Slope` vs `SlopeNaN` is misleading. The compute is the right call for the new class. Either add a one-line comment saying the dask numbers are not comparable with `Slope`, or leave `Slope` as is and accept it; changing `Slope` is out of scope here.
- [ ] `xrspatial/tests/test_slope.py:642-648`: `_expected_nan_footprint` builds the expected mask by hand. Worth a sentence in the comment block that these tests also pass on the pre-PR kernel; they pin behaviour rather than reproduce a regression, so nobody reverts the kernel expecting them to go red. The bitwise A/B in the PR body is what proves equivalence.

- [ ] `benchmarks/benchmarks/slope.py:31`: `rng.random((ny, nx))` at `nx=10000` allocates a 50M-element float64 array (400 MB) just to derive a bool mask. `rng.random((ny, nx), dtype=np.float32)` halves that. Setup time is not timed, so this only matters for peak memory on the asv runner.
- [ ] `xrspatial/slope.py:46`: the section banner above `_cpu` still reads "Planar backend functions (unchanged)". It predates this PR and is now wrong twice over. Not this PR's doing, but the diff sits directly under it.

### What looks good
- The select semantics are exactly what the old branch gave. Any NaN in the 3x3 window makes `p` NaN and `arctan(NaN)` is NaN; the only case that needed help was a finite window with a NaN centre, and `ctr == ctr` covers it. `inf` behaves the same on both sides too (`inf - inf` is NaN in the stencil, an `inf` centre still computes).
- `out` is pre-filled with NaN so the border rows and columns stay NaN without any change to the loop bounds.
- The A/B in the PR body covers the degenerate shapes (3x3, 1x50, 50x1, 2x50) and the int16 cast, and the numpy and dask+numpy paths share `_cpu` so both are covered by one comparison.
- The benchmark reads `include_nan` correctly: `common.py` sets exactly `z[0, 0]`, a border cell, so the class adds its own 30% speckle. Verified the mask keeps the dask array lazy and chunked.
- GPU and geodesic kernels untouched, as intended.

### Checklist
- [x] Algorithm matches reference: Horn 3x3 stencil unchanged, same coefficients and load order
- [x] All implemented backends produce consistent results: numpy and dask+numpy share the kernel; cupy paths not modified
- [x] NaN handling is correct: select on the centre, stencil propagates neighbours
- [x] Edge cases are covered by tests: speckled footprint on numpy and dask, existing centre-NaN and 1xN/2xN tests still pass
- [x] Dask chunk boundaries handled correctly: no change to `map_overlap` depth or boundary
- [x] No premature materialization or unnecessary copies: benchmark compute is deliberate
- [x] Benchmark exists: `SlopeNaN` added
- [ ] README feature matrix: not applicable, no new function or backend
- [x] Docstrings: no public signature change

brendancol (commented):
## PR Review: Drop the per-cell NaN early-out in the planar CPU slope kernel (follow-up)

Second pass over b2c5a226, which responds to the first review.

None.

None.

None.

- `SlopeNaN` vs `Slope` dask comparability: fixed, comment added at `benchmarks/benchmarks/slope.py:36-38`.
- Footprint tests pass on the old kernel too: fixed, stated in the comment block at `xrspatial/tests/test_slope.py:634-636`.
- float64 draw for the speckle mask: fixed, `dtype=np.float32` at `benchmarks/benchmarks/slope.py:32`.
- Stale "(unchanged)" banner: fixed at `xrspatial/slope.py:46`.

### What looks good
- The kernel itself is unchanged since the first pass; `pytest xrspatial/tests/test_slope.py` still reports 137 passed and the benchmark class still runs on numpy and dask.
- The float32 draw does not change the mask: `rng.random(dtype=np.float32) < 0.3` selects with the same seed, so the NaN fraction stays at 0.30.

### Checklist
- [x] NaN handling is correct
- [x] Benchmark exists
- [ ] README feature matrix: not applicable
- [x] Docstrings: no public signature change

PR #3743, opened 2026-09-04
Replace the D8 code-to-offset if/elif chain with a lookup table (#3738)

### LINKED ISSUE #3738: hydro: replace the D8 code-to-offset if/elif chain with a lookup table
## Reason or Problem

`_code_to_offset(code)` in `xrspatial/hydro/flow_accumulation_d8.py` maps a D8 direction code to a `(dy, dx)` neighbour offset with an eight-way `if/elif` chain. `_flow_accum_cpu` calls it twice per cell (in-degree pass, then the BFS drain), and stream_link_d8, flow_length_d8, flow_path_d8, stream_order_d8 and hand_d8 import the same function for their own per-cell loops. On a 1000x2000 raster that is a few million data-dependent branch chains per accumulation call.

## Proposal

Replace the chain with two module-level lookup arrays indexed by the code:

```python
_D8_DY = np.zeros(129, dtype=np.int64)
_D8_DX = np.zeros(129, dtype=np.int64)
# populate entries 1, 2, 4, 8, 16, 32, 64, 128

@ngjit
def _code_to_offset(code):
    c = int(code)
    if 0 <= c <= 128:
        return _D8_DY[c], _D8_DX[c]
    return 0, 0
```

numba freezes module-level numpy arrays as compile-time constants, so this compiles to a range check and two loads.

**Design:** The guard has to be written as an inside-the-box test (`0 <= c <= 128`). `int(nan)` in numba is `INT64_MIN` and numba does no bounds checking, so an unguarded `_D8_DY[c]` on a NaN code reads out of bounds. Same bug class as the `erode()` fix in #3703. Non-power-of-two codes in range (3, 5, ...) and anything outside 0..128 return `(0, 0)` as the chain does today. Return types stay integer so `r + dy` in the callers is unchanged. `_code_to_offset_py` keeps its current behaviour.

**Usage:** No public API change. Modules that import `_code_to_offset` pick up the new version without edits.

**Value:** A spike on `_flow_accum_cpu` with a 1000x2000 D8 raster measured about 0.79x of baseline wall time with bitwise identical output.

## Stakeholders and Impacts

Users of `flow_accumulation`, `flow_length`, `stream_link`, `stream_order`, `flow_path` and `hand` on the numpy and dask+numpy backends. GPU kernels have their own device-side dispatch and are not touched. The `FlowAccumulation` asv class in `benchmarks/benchmarks/flow_accumulation.py` already covers the CPU path.

## Drawbacks

Two small module-level arrays in the module namespace. The guard adds a compare per call; the measurement above includes it.

## Alternatives

Compacting the BFS queue seeding into the in-degree pass measured only 0.93x on its own and is left out.

## Unresolved Questions

None.

`watershed_d8.py` and `basin_d8.py` carry private copies of `_code_to_offset`. Out of scope here; they can be pointed at the shared table later.

brendancol:
Closes #3738

`_code_to_offset` in `flow_accumulation_d8.py` mapped a D8 code to its `(dy, dx)` offset with an eight-way `if/elif` chain. `_flow_accum_cpu` calls it twice per cell, and stream_link_d8, flow_length_d8, flow_path_d8, stream_order_d8 and hand_d8 import it for their own per-cell loops.

- Replace the chain with two module-level `int64` lookup arrays indexed by the code. numba freezes module-level arrays as compile-time constants, so the call is now a range check and two loads.
- The range check is deliberately an inside-the-box test, `0 <= c <= 128`. `int(nan)` is `INT64_MIN` inside numba and there is no bounds checking, so a rejection-form guard would let a NaN code index out of bounds (same class as the `erode()` fix in #3703). Codes like 3 or 5, anything above 128, negative codes and NaN keep returning `(0, 0)`, and the returns stay integer so `r + dy` in the callers is unchanged.
- `_code_to_offset_py` is untouched.

On a 1000x2000 float64 flow-direction raster (Gaussian bump plus `default_rng(71942).normal(0, 2)` noise), `_flow_accum_cpu` went from 56.3 ms to 39.7 ms, median of 7 after warmup, with identical results. The output is also identical on a raster built from the same DEM with 30% random NaN, checked with `np.array_equal(..., equal_nan=True)` against the function pulled from `origin/main`.

Backends: numpy and dask+numpy share this CPU kernel. cupy and dask+cupy use the device-side dispatch and are not touched.

Test plan
- [x] New parametrised test for `_code_to_offset` covering the eight codes, 0, 3, 5, 129, 255, -1, 1e9, float codes and NaN, asserting the same tuples the old chain returned.
- [x] That test run under `NUMBA_BOUNDSCHECK=1`, plus a negative control confirming the flag raises `IndexError` on an unguarded lookup with a NaN code.
- [x] `pytest xrspatial/hydro/tests -k "d8 or flow_accum or flow_length or stream_link or stream_order or watershed or flow_path or hand or basin"`: 872 passed.
- [x] A/B against `origin/main` on the clean and 30% NaN rasters: bitwise identical.

The `FlowAccumulation` asv class in `benchmarks/benchmarks/flow_accumulation.py` already covers this path, so no benchmark change.

### DIFF
--- a/xrspatial/hydro/flow_accumulation_d8.py
+++ b/xrspatial/hydro/flow_accumulation_d8.py
@@ -192,26 +192,30 @@ def _no_weight_cupy():
 # Direction helpers
 # =====================================================================
 
+# (dy, dx) row/col offset per D8 code, indexed by the code itself.  Only
+# the eight power-of-two entries are non-zero; every other in-range index
+# stays (0, 0) so codes like 3 or 5 keep meaning "no flow".  numba freezes
+# module-level arrays as compile-time constants, so the lookup below is a
+# range check plus two loads instead of an eight-way branch chain.
+_D8_DY = np.zeros(129, dtype=np.int64)
+_D8_DX = np.zeros(129, dtype=np.int64)
+for _code, (_dy, _dx) in ((1, (0, 1)), (2, (1, 1)), (4, (1, 0)), (8, (1, -1)),
+                          (16, (0, -1)), (32, (-1, -1)), (64, (-1, 0)),
+                          (128, (-1, 1))):
+    _D8_DY[_code] = _dy
+    _D8_DX[_code] = _dx
+del _code, _dy, _dx
+
+
 @ngjit
 def _code_to_offset(code):
     """Return (dy, dx) row/col offset for a D8 direction code."""
     c = int(code)
-    if c == 1:
-        return 0, 1
-    elif c == 2:
-        return 1, 1
-    elif c == 4:
-        return 1, 0
-    elif c == 8:
-        return 1, -1
-    elif c == 16:
-        return 0, -1
-    elif c == 32:
-        return -1, -1
-    elif c == 64:
-        return -1, 0
-    elif c == 128:
-        return -1, 1
+    # Inside-the-box guard, not a rejection test: int(nan) is INT64_MIN in
+    # numba and there is no bounds checking, so ``c < 0 or c > 128`` style
+    # guards would not stop a NaN code from indexing out of bounds.
+    if 0 <= c <= 128:
+        return _D8_DY[c], _D8_DX[c]
     return 0, 0
 
 
--- a/xrspatial/hydro/flow_accumulation_d8.py
+++ b/xrspatial/hydro/flow_accumulation_d8.py
@@ -210,10 +210,15 @@ del _code, _dy, _dx
 @ngjit
 def _code_to_offset(code):
     """Return (dy, dx) row/col offset for a D8 direction code."""
+    # NaN never reaches int(): the float-to-int conversion of NaN is
+    # undefined (INT64_MIN on x86, 0 on aarch64) and numba does no bounds
+    # checking, so the table index must never depend on it.
+    if code != code:
+        return 0, 0
     c = int(code)
-    # Inside-the-box guard, not a rejection test: int(nan) is INT64_MIN in
-    # numba and there is no bounds checking, so ``c < 0 or c > 128`` style
-    # guards would not stop a NaN code from indexing out of bounds.
+    # Guard on the converted integer with an inside-the-box test.  A
+    # rejection test on the float (``code < 0 or code > 128``) is False for
+    # NaN and would fall through to the lookup.
     if 0 <= c <= 128:
         return _D8_DY[c], _D8_DX[c]
     return 0, 0
--- a/xrspatial/hydro/tests/test_flow_accumulation_d8.py
+++ b/xrspatial/hydro/tests/test_flow_accumulation_d8.py
@@ -626,3 +626,56 @@ def test_weight_dataset_accessor():
     expected = flow_accumulation(agg, weight=w).data
     for var in ('a', 'b'):
         np.testing.assert_allclose(out[var].data, expected, equal_nan=True)
+
+
+# ---------------------------------------------------------------------------
+# D8 code -> (dy, dx) lookup (#3738)
+# ---------------------------------------------------------------------------
+
+@pytest.mark.parametrize("code, expected", [
+    # the eight valid codes: E, SE, S, SW, W, NW, N, NE
+    (1, (0, 1)),
+    (2, (1, 1)),
+    (4, (1, 0)),
+    (8, (1, -1)),
+    (16, (0, -1)),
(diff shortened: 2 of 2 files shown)


brendancol (commented):
## PR Review: Replace the D8 code-to-offset if/elif chain with a lookup table (#3738)

None.

- [ ] `xrspatial/hydro/flow_accumulation_d8.py:214-216`: the comment says a `c < 0 or c > 128` guard would not stop a NaN code. That is not what happens here. `c` is already an integer at that point, and on x86 `int(nan)` lands on `INT64_MIN`, which `c < 0` does reject. The hazard that actually bites is comparing the float before the conversion (`code < 0 or code > 128` is False for NaN, so it falls through to `int(nan)`), or relying on what `fptosi` does with NaN at all, since LLVM treats that as poison and other targets return a different value (aarch64 `fcvtzs` gives 0). Reword the comment to describe that, so the next reader does not learn the wrong rule from it.
- [ ] `xrspatial/hydro/flow_accumulation_d8.py:213`: given the above, an explicit `if code != code: return 0, 0` before `int(code)` would make NaN handling independent of the conversion semantics instead of depending on the integer guard catching whatever `fptosi` produced. The bounds-checked test proves the current x86 behaviour but not the portable one. Worth a quick timing check that the extra compare is free on the 1000x2000 case; if it is, add it.

- [ ] `xrspatial/hydro/tests/test_flow_accumulation_d8.py:664-668` and `:677`: the two new tests import `_code_to_offset`, `_code_to_offset_py`, `_D8_DX` and `_D8_DY` inside the function bodies. The rest of the file imports at module level; move these up to match.
- [ ] The PR touches the kernel behind `FlowAccumulation` in `benchmarks/benchmarks/flow_accumulation.py`, but the path-based labeler may not tag hydro as `performance`. Add the label by hand so the asv job runs on this PR.

### What looks good
- The guard is written as `0 <= c <= 128` rather than a rejection test, and the PR includes a negative control showing `NUMBA_BOUNDSCHECK=1` raises on the unguarded form with a NaN code.
- Returns stay `int64` from the tables, so `r + dy` arithmetic in the seven importing modules is unchanged. The old literal tuples also typed as `int64`, so no recompilation surprises for callers.
- The A/B against `origin/main` covers both a clean raster and one with 30% NaN, using `array_equal(equal_nan=True)` rather than `allclose`.
- `_code_to_offset_py` is left alone, so the boundary-stitching code in watershed_d8, flow_length_d8 and hand_d8 that uses it is unaffected.
- The parametrised test pins the mapping explicitly (not by comparing to the Python version), so a future edit to either table is caught.

### Checklist
- [x] Algorithm matches reference (same mapping as the chain, verified for every probe)
- [x] All implemented backends produce consistent results (numpy and dask+numpy share the kernel; GPU untouched)
- [x] NaN handling is correct (see suggestion on the comment wording)
- [x] Edge cases are covered by tests (0, 3, 5, 129, 255, -1, 1e9, floats, NaN)
- [x] Dask chunk boundaries handled correctly (no change to tiling)
- [x] Benchmark exists (`FlowAccumulation`)
- [x] README feature matrix: not applicable, private helper

brendancol (commented):
## PR Review: Replace the D8 code-to-offset if/elif chain with a lookup table (#3738), follow-up

Second pass after 2405f0cf.

None.

None.

None.

- Guard comment wording: fixed. `flow_accumulation_d8.py:213-221` now names the real hazard (a rejection test on the float is False for NaN) instead of claiming an integer rejection test would miss it.
- Explicit NaN check before `int()`: fixed. `if code != code: return 0, 0` sits ahead of the conversion, so the table index no longer depends on what `fptosi` does with NaN on a given target. Timed at 36.5 ms and 37.1 ms on the 1000x2000 case versus 39.7 ms without it, so the compare is in the noise, and the A/B against `origin/main` is still bitwise identical on both rasters.
- Module-level test imports: fixed. `test_flow_accumulation_d8.py:6-11`.
- `performance` label: added to the PR.

### What looks good
- The NaN path is now tested three ways: the parametrised case, the same test under `NUMBA_BOUNDSCHECK=1`, and the A/B on a 30% NaN raster.
- The bench script's `xrspatial.__file__` assertion guards against the worktree-vs-main editable install trap, so the numbers are for the branch under review.

### Checklist
- [x] NaN handling is correct
- [x] Benchmark exists

Issue #3710, opened 2026-08-16
surface_distance.py fails the project isort check


brendancol:
`xrspatial/surface_distance.py` fails the project's isort check. `setup.cfg` configures isort with `line_length = 100`, but no CI job runs it, so the drift went unnoticed.

## Reproduction

```
$ python -m isort --check-only --diff xrspatial/surface_distance.py
ERROR: xrspatial/surface_distance.py Imports are incorrectly sorted and/or formatted.
--- xrspatial/surface_distance.py:before
+++ xrspatial/surface_distance.py:after
@@ -47,15 +47,12 @@
-from xrspatial.cost_distance import _heap_push, _heap_pop
+from xrspatial.cost_distance import _heap_pop, _heap_push
+from xrspatial.dataset_support import supports_dataset
 from xrspatial.proximity import _vectorized_calc_direction
-from xrspatial.utils import (
-    _validate_raster,
-    cuda_args, get_dataarray_resolution, ngjit,
-    has_cuda_and_cupy, is_cupy_array, is_dask_cupy,
-)
-from xrspatial.dataset_support import supports_dataset
-from xrspatial.utils import _dask_task_name_kwargs
+from xrspatial.utils import (_dask_task_name_kwargs, _validate_raster, cuda_args,
+                             get_dataarray_resolution, has_cuda_and_cupy, is_cupy_array,
+                             is_dask_cupy, ngjit)
```

Three things are off:

1. `_heap_push, _heap_pop` are out of alphabetical order.
2. `xrspatial.utils` is imported twice, in two statements separated by an unrelated import.
3. The wrapped `xrspatial.utils` block uses a hanging-indent style that its nearest neighbours do not. `cost_distance.py`, `proximity.py`, `focal.py` and `zonal.py` all already carry the exact layout isort wants here, so this file is the outlier among the modules it borrows from.

flake8 is clean on this file (0 violations), so this is purely the isort side.

## Scope

Import reordering only. No runtime behaviour change, no names added or removed.

Found by a style sweep of `surface_distance`. Two smaller things turned up that are not worth a code change on their own, noted here so nobody has to rediscover them:

- `_surface_distance_dask` unpacks `height, width = source_da.shape` and never uses either name. pyflakes skips F841 for tuple unpacking, so flake8 misses it.
- `_precompute_dd_grid` imports `_available_memory_bytes` from `xrspatial.zonal` even though this module defines its own copy of that helper at the top. Both work; the local one is right there.

PR #3746, opened 2026-09-10
Accept target_values=None in surface_distance trio (#3712)

### LINKED ISSUE #3712: surface_distance trio rejects target_values=None while proximity, allocation, direction and cost_distance accept it

shaikn6:
Fixes #3712

## Proposed Changes

  - `surface_distance`, `surface_allocation`, `surface_direction` now accept `target_values=None` (the value wrapper code passes when its own optional arg is unset), matching `proximity` / `allocation` / `direction` / `cost_distance`.
  - Changed the three signatures from `target_values: list = []` to `target_values: list = None` and added the sibling `if target_values is None: target_values = []` guard in the shared `_compute` dispatcher.
  - Added a regression test (`test_target_values_none_matches_empty`) parametrized over all three functions.

## Why / evidence

The proximity trio and `cost_distance` all declare `target_values: list = None` and normalize to `[]` in the body. The three `surface_*` functions declared `target_values: list = []` with no `None` guard, so `target_values=None` fell through to `np.asarray(None, dtype=np.float64)` (a 0-d array) and blew up before the numba kernel. On numpy 2.3 it surfaces as the existing `ValueError: target_values must be a 1-D sequence`; the issue reports a ~40-line numba TypingError on older numpy. Either way it is broken. `[]` as a default is also the mutable-default anti-pattern the siblings already avoid.

Confirmed with the reproduction from #3712 on this checkout: pre-fix, `surface_distance/allocation/direction(target_values=None)` raise while the other four return normally; post-fix all seven succeed.

## Change

`[]` and `None` both mean "treat every non-zero finite pixel as a source", so callers passing an explicit list are unaffected and callers passing `None` go from a traceback to working code. No parameter renamed, no accepted input changes meaning, so no deprecation shim is needed.

Note: `balanced_allocation` carries the same `target_values: list = []` default (flagged in #3712 as out of scope) and is left untouched here.

## Verified

```
$ python -m pytest xrspatial/tests/test_surface_distance.py -q
81 passed, 19 skipped

$ python -m pytest xrspatial/tests/test_surface_distance.py -q -k target_values
6 passed, 94 deselected

$ python repro_3712.py   # reproduction script from the issue
proximity(target_values=None) -> OK
allocation(target_values=None) -> OK
direction(target_values=None) -> OK
cost_distance(target_values=None) -> OK
surface_distance(target_values=None) -> OK
surface_allocation(target_values=None) -> OK
surface_direction(target_values=None) -> OK
```

flake8 / isort: no new violations introduced by this change (the pre-existing `surface_distance.py` isort drift is tracked separately in #3710).

### DIFF
--- a/xrspatial/surface_distance.py
+++ b/xrspatial/surface_distance.py
@@ -1383,6 +1383,9 @@ def _compute(raster, elevation, x, y, target_values, max_distance,
     signed_cellsize_x = cellsize_x * _coord_step_sign(raster, x)
     signed_cellsize_y = cellsize_y * _coord_step_sign(raster, y)
 
+    if target_values is None:
+        target_values = []
+
     target_values = np.asarray(target_values, dtype=np.float64)
     if target_values.ndim != 1:
         raise ValueError(
@@ -1528,7 +1531,7 @@ def surface_distance(
     elevation: xr.DataArray,
     x: str = "x",
     y: str = "y",
-    target_values: list = [],
+    target_values: list = None,
     max_distance: float = np.inf,
     connectivity: int = 8,
     method: str = 'planar',
@@ -1584,7 +1587,7 @@ def surface_allocation(
     elevation: xr.DataArray,
     x: str = "x",
     y: str = "y",
-    target_values: list = [],
+    target_values: list = None,
     max_distance: float = np.inf,
     connectivity: int = 8,
     method: str = 'planar',
@@ -1621,7 +1624,7 @@ def surface_direction(
     elevation: xr.DataArray,
     x: str = "x",
     y: str = "y",
-    target_values: list = [],
+    target_values: list = None,
     max_distance: float = np.inf,
     connectivity: int = 8,
     method: str = 'planar',
--- a/xrspatial/tests/test_surface_distance.py
+++ b/xrspatial/tests/test_surface_distance.py
@@ -613,6 +613,16 @@ def test_invalid_target_values_shape(bad):
         surface_distance(source, elev, target_values=bad)
 
 
+@pytest.mark.parametrize("func", [surface_distance, surface_allocation, surface_direction])
+def test_target_values_none_matches_empty(func):
+    """target_values=None must behave like the [] default (issue #3712)."""
+    source = _make_raster(np.array([[0.0, 1.0, 0.0, 2.0, 0.0]], dtype=np.float64))
+    elev = _make_raster(np.zeros((1, 5), dtype=np.float64))
+    expected = _compute(func(source, elev))
+    got = _compute(func(source, elev, target_values=None))
+    np.testing.assert_array_equal(np.asarray(got), np.asarray(expected))
+
+
 # ---------------------------------------------------------------------------
 # Tests — dask-specific
 # ---------------------------------------------------------------------------


brendancol:
@shaikn6 Hey hey, what did you think of earlier PR here: https://github.com/xarray-contrib/xarray-spatial/pull/3718

Can you help me understand what is different about this PR.  Thanks!

shaikn6:
Thanks for flagging — on comparing, #3746 is functionally the same fix as #3718 (same three signatures, same `_compute` normalization), so I should have checked for an existing PR on #3712 first. The only addition is a regression test parametrized over all three functions (`test_target_values_none_matches_empty`); feel free to lift it into #3718. Closing this one in favor of yours.

brendancol:
@shaikn6 cool cool, thanks for the contribution.  I could help guide you a bit to other areas if youre interested in other contributions.  Any existing tools (or new tools) you'd be interested in working on?

shaikn6:
Appreciate it — yes, interested. I've mostly been running `/sweep-api-consistency`-style passes looking for signature/behavior drift within a module, but I'd rather point that at something you actually need than keep freelancing. What's open — a specific tool that needs work, or an area (e.g. more of the proximity/surface_distance family, a different module entirely) you'd want eyes on first?

